PR #1, opened 2020-05-29
Token program

jackcmay:
Moved and modified from: https://github.com/solana-labs/example-token


mvines:
I'd squash the history

jackcmay:
@mvines I made a point to move the history from the example-token repo over for posterity, you think we should ditch that history?

Issue #11, opened 2020-06-02
Token program needs a Burn instruction


mvines:
Sometimes users want to burn some tokens.  Let's add the token program equivalent of `1nc1nerator11111111111111111111111111111111`

jackcmay:
@mvines I see a few options here:
- Re-use the `1nc1nerator11111111111111111111111111111111` account but only use it as a place holder in the `Transfer` instruction. The Token program doesn't actually modify that account but identifies it by key and "burns" the tokens
- Same as above using the `Transfer` instruction but use a different key (don't re-use `1nc1nerator11111111111111111111111111111111`
- Create a `Burn` instruction that burns tokens

I'm planning on doing the first, any objections?

mvines:
Sorry for the delayed reply, I sorta feel like an explicit `Burn` instruction would be better.   Reusing `1nc1nerator11111111111111111111111111111111` feels a little too sneaky to me

PR #15, opened 2020-06-02
Add incinerator

jackcmay:
Added a `Burn` instruction

fixes: #11


mvines:
Hmm, I think I may have found an argument for the Burn instruction.   With Transfer, we have no way to decrease the supply by the burn amount:
https://github.com/solana-labs/solana-program-library/blob/0c3842a07889a7de0384f096993574a84c821563/token/src/program/src/state.rs#L14

Issue #13, opened 2020-06-02
Token program should support to tokens with unlimited supply


mvines:
Currently, the total supply of tokens is created upon the `NewToken` instruction.  But there are some tokens that require an unlimited supply and the ability to mint new tokens at any time (BRRR).

One suggestion is that at token creation time, an admin key is stored in the token account. Only the admin key can then invoke a Mint instruction to create new supply, perhaps with another instruction to replace the admin key

PR #19, opened 2020-06-05
Add MintTo instruction

jackcmay:
Allow Token owners to mint new tokens.  If no owner is supplied during the creation of the token than that token's supply is fixed.

fixes: #13

### DIFF
--- a/token/src/program/src/error.rs
+++ b/token/src/program/src/error.rs
@@ -17,8 +17,8 @@ pub enum TokenError {
     NotDelegate,
     #[error("no owner")]
     NoOwner,
-    #[error("minting not allowed")]
-    MintingNotAllowed,
+    #[error("fixed supply")]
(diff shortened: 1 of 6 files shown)


token/src/program/src/error.rs
```diff
@@ -41,6 +43,7 @@ impl PrintProgramError for TokenError {
             TokenError::TokenMismatch => info!("Error: token mismatch"),
             TokenError::NotDelegate => info!("Error: not a delegate"),
             TokenError::NoOwner => info!("Error: no owner"),
+            TokenError::MintingNotAllowed => info!("Error: this token does not allow new mintings"),
```
mvines, line 46: How about:
```suggestion
            TokenError::FixedSupply => info!("Error: the total supply of this token is fixed"),
```

PR #20, opened 2020-06-05
Reorganize repository in preparation for more programs

### LINKED ISSUE #18: Developers not familiar with node/js frustrated by it being required to build programs
Reisen from Chrous One provided the following feedback (continued from https://github.com/solana-labs/solana-com/issues/390)

> 7.  Opening both of these have decent README's but include a lot of web stuff, where I only want to build a contract.
> 8.  Looking at build:bpf-rust to see how I can create a new project with JUST the BPF compilation step lands me here: https://github.com/solana-labs/example-tictactoe/tree/master/program-bpf-rust
> 9.  I see do.sh, realise I need node/npm installed, and that this script runs a script located under node_modules.
> 10. I didn't want to dig further.
> 
> At this point I just cd'd into the C BPF folder and ran make and embraced the darkness of C.

The only dependency required to build the program is the `bpf-sdk` which node currently downloads, we could obtain the `bpf-sdk` other ways and give developers a non-node path to program development.  For end-to-end testing/development, the client is still based on `Solana-web3.js` which requires node/js

CriesofCarrots:
Currently, the token program is a self-contained unit. These means that there are various scripts and features that would need to be duplicated for any other program added to the repo.

This PR reorganizes the token program, pulling out some resources that will be shared with other programs. It also breaks the dependency on `npm` for builds.

Here is an outline of the new org:
```
solana-program-library
├── program
│   ├── js
│   │   ├── cli -- morph into test?
│   │   ├── client -- morph into src?
│   │   ├── package.json
│   │   ├── package-lock.json
│   │   └── README.md
│   ├── src
│   │   └── lib.rs, etc
│   ├── address.md -- this does not exist for token yet
│   ├── Cargo.lock
│   ├── Cargo.toml
│   ├── README.md
│   └── Xargo.toml
├── .gitignore
├── .travis.yml
├── do.sh
├── LICENSE
└── README.md
```

Further questions/issues:
1. CI needs some work. I'm inclined to switch to buildkite so that we can build/test only those program directories with changes; I don't think this is possible with travis.
2. I envision publishing the program's on-chain address in an address.md file, as included above. But we need some authoritative way to couple a release of a particular program/source with the address.
3. Maybe paired with (2), it'd be great to have a tool for users to confirm that the program at the declared address on each of our clusters aligns with the source code published here.
4. Token js: this directory could use some more work, especially if we're inclined to publish the interface to npm

Fixes #18


mvines:
>  I'm inclined to switch to buildkite so that we can build/test only those program directories with changes; I don't think this is possible with travis.

I'd like to stick with Travis if we can, less infra for us to manage.  We can run the programs in parallel using Travis's build stages feature:  https://docs.travis-ci.com/user/build-stages/

jackcmay:
This is a good first step.  We also need a process to release programs individually and tie program ids to program releases

CriesofCarrots:
> This is a good first step. We also need a process to release programs individually and tie program ids to program releases

Agreed, related to number 2 above.

I'll put some initial CI build-stages in place and then I'd like to get this merged so most of the moving-things-around churn is out of the way.

CriesofCarrots:
Right now, CI is building/testing the token program twice. This is a check for this PR's `do.sh` all-repo actions, and sets up the framework for building/testing multiple programs in parallel. As soon as there are additional programs in the repo, it should be modified.

PR #25, opened 2020-06-05
Add memo program

CriesofCarrots:
Comprises the Rust program. JS utilities to come separately.

Also, I included the program-id, but if it makes more sense to add that as part of a process to release a program version and load it to cluster(s), I'll pull that out.


memo/src/instruction.rs
```diff
@@ -0,0 +1,49 @@
+use crate::error::MemoError;
+use solana_sdk::{entrypoint::ProgramResult, info, program_error::ProgramError};
+use std::{mem::size_of, str::from_utf8};
+
+/// Instructions supported by the memo program
+#[repr(C)]
+#[derive(Clone, Debug, PartialEq)]
```
garious, line 7: Why the handwritten serialize/deserialize? Seems like Serde and Bincode would work fine.

memo/src/processor.rs
```diff
@@ -0,0 +1,67 @@
...
+    Ok(())
+}
+
+pub fn process<'a>(
+    _program_id: &Pubkey,
+    _accounts: &'a [AccountInfo<'a>],
+    input: &[u8],
+) -> ProgramResult {
+    let _command = Instruction::deserialize(input)?;
+    Ok(())
```
mvines, line 27: maybe just this?
```suggestion
    Instruction::deserialize(input)
```

memo/src/processor.rs
```diff
@@ -0,0 +1,67 @@
...
+
+entrypoint!(process_instruction);
+fn process_instruction<'a>(
+    program_id: &Pubkey,
+    accounts: &'a [AccountInfo<'a>],
+    instruction_data: &[u8],
+) -> ProgramResult {
+    if let Err(error) = process(program_id, accounts, instruction_data) {
+        // catch the error so we can print it
+        error.print::<MemoError>();
```
mvines, line 15: Printing seems unnecessary here to me.  The program is trivial and it returns nice custom error codes

memo/src/instruction.rs
```diff
@@ -0,0 +1,49 @@
...
+pub enum Instruction<'a> {
+    /// Validate UTF-8 encoded characters
+    Utf8(&'a str),
```
garious, line 10: Can you follow the verb-subject naming convention that the other programs use?  Maybe `AddMemo`?

mvines, line 10: This'll cause a bincode prefix to the valid UTF-8 encoded characters right?  That's going to feel clunky for non-rust users.    The instruction data can just be the raw UTF-8

memo/README.md
```diff
@@ -0,0 +1,4 @@
+# Memo Program
+
+A simple program that accepts a string of encoded characters and verifies that
+it parses. Currently handles UTF-8.
```
mvines, line 4: Maybe a little more here that describes the instructions that the program supports?

memo/src/instruction.rs
CriesofCarrots: Yes. This was @garious 's idea, with the thought that we might want to support different encoding types in future versions

CriesofCarrots: `AddUtf8` or `AddUtf8Memo`? The whole point of having different variants is the encoding type

mvines: Bumping the program id would be a nice mechanism for that too

mvines: Otherwise we're asking all non-Rust users to append 8 zero bytes in front of their UTF-8 string right?  That's just weird IMO

garious: Right, thought we might want to add things like `AddEncryptedMemo` or `AddUtf16Memo` or `AddGifMemo`.  Just never know.  And currently we use Bincode everywhere.  If we're going to break away from Bincode, I'd like to make it for the purpose of supporting some language-agnostic RPC, like protobuf.  @jstarry is looking into it.

garious: How about `AddUtf8String` so that it reads `memo_instruction::AddUtf8String` without duplicating the word "memo".

memo/src/instruction.rs
CriesofCarrots: Yeah, it could. I followed the Token program's lead on this to shrink the instruction size to 1 byte. Serde would make it 4 bytes.

memo/src/instruction.rs
CriesofCarrots: It's just one byte, but yeah

garious: I'm also okay with a Program ID per memo type, such that the `Instruction::data` can be the UTF-8 string with no header.  But so far, `Config` program is the only program that broke from the `enum Instruction` convention. Personally, I find that program to be one of the most confusing ones.

garious: That does sound appealing though:

```rust
let memo_instruction = Instruction {
    program_id: "Memo1UhkJRfHyvLMcVucJwxXeuD728EqVDDwQDxFMNo".parse().unwrap(),
    accounts: vec![],
    data = b"Hello, world",
};
```

Wouldn't even need a custom error, could just use `InstructionError::InvalidInstructionData`.

@CriesofCarrots, what do you think about throwing together a second PR with this approach so that we can compare and contrast?

CriesofCarrots: Sure, I can do that. To be clear, then, each encoding/content type would need a separate program, so this becomes the `utf8-memo` program or some such?

garious: or just keep it `memo` and force the next one (if it ever exists) to add precision. fine to let simple things be simple

garious:
Winner https://github.com/solana-labs/solana-program-library/pull/27

PR #27, opened 2020-06-05
Add utf8 memo program

CriesofCarrots:
New, simpler version, without Instruction.
Resulting program file is about 2kb smaller.


memo/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+version = "0.1.0"
+description = "Solana Program Library Memo"
+authors = ["Solana Maintainers <maintainers@solana.com>"]
+repository = "https://github.com/solana-labs/solana-program-library"
+license = "Apache-2.0"
+edition = "2018"
+
+[dependencies]
+num-derive = "0.2"
+num-traits = "0.2"
```
garious, line 15: Can you review these dependencies? Seems like `num-derive`, `num-traits`, and `thiserror` might no longer be needed.

memo/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+solana-sdk = { version = "=1.2.0", default-features = false, features=["program"] }
```
garious, line 16: Why the hard `=`?

memo/program-id.md
```diff
@@ -0,0 +1 @@
+Memo1UhkJRfHyvLMcVucJwxXeuD728EqVDDwQDxFMNo
```
mvines, line 1: for the convenience of Rust clients, maybe the program-id should live in a .rs file too?  I guess this is a more general SPL setup comment that we can deal with later

memo/Cargo.toml
CriesofCarrots: The precedent here (Token) is to pin to a particular version of the solana-sdk. We still need to work out release/deployment processes for the programs in this repo. But I do think the `=` is helpful from a readability perspective, assuming what appears here is meant to represent what is published to the chain.

garious: We'll need to add release tags any time we're publishing to the chain.  Definitely can't assume what's in the master branch is meaningful. Can't really assume what's outside the Cargo.lock is too meaningful either.  `=` deps are bad for downstream consumers.

memo/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+version = "0.1.0"
+description = "Solana Program Library Memo"
+authors = ["Solana Maintainers <maintainers@solana.com>"]
+repository = "https://github.com/solana-labs/solana-program-library"
+license = "Apache-2.0"
+edition = "2018"
+
+[dependencies]
+solana-sdk = { version = "=1.2.0", default-features = false, features=["program"] }
+solana-sdk-bpf-test = { path = "../bin/bpf-sdk/rust/test", default-features = false }
```
garious, line 15: Just looking at this name here, with the `test` suffix. Can this be put in `[dev-dependencies]`?

memo/src/processor.rs
```diff
@@ -0,0 +1,48 @@
...
+};
+use std::str::from_utf8;
+
+entrypoint!(process_instruction);
+fn process_instruction<'a>(
+    _program_id: &Pubkey,
+    _accounts: &'a [AccountInfo<'a>],
+    instruction_data: &[u8],
+) -> ProgramResult {
+    from_utf8(instruction_data)
```
garious, line 13: Seems more readable to me:

```rust
from_utf8(instruction_data).map_err(|_| ProgramError::InvalidInstructionData)?;
Ok(())
```

Maybe @mvines didn't notice the return value?

memo/Cargo.toml
CriesofCarrots: Ah, actually memo doesn't need that one at all either

memo/program-id.md
CriesofCarrots: Yeah, good idea. I'll leave as-is for now, but will move around when we rethink how we want to publish rust assets.

memo/src/processor.rs
CriesofCarrots: I prefer that as well. I'll try to sneak it by him again ;)

memo/src/processor.rs
```diff
@@ -0,0 +1,47 @@
...
+    use solana_sdk::{program_error::ProgramError, pubkey::Pubkey};
+
+    #[test]
+    fn test_utf8_memo() {
+        let program_id = Pubkey::new(&[0; 32]);
+
+        let string = b"letters and such";
+        assert_eq!(
+            Ok(()),
+            process_instruction(&program_id, &mut vec![], string)
```
garious, line 29: `cargo clippy --tests`

garious:
:shipit:

memo/src/processor.rs
mvines: 🙈 ok do it now!

PR #37, opened 2020-06-16
Make do.sh more generic

jackcmay:
- `do.sh` checks for specific directories when it attempts to build all the programs.  Make it more generic to try to build any project that contains a `Cargo.toml` file in its root.
- Add support for generating docs

(Also my shell didn't like the form of the test syntax)


CriesofCarrots:
Ah, much better, thanks!

PR #39, opened 2020-06-16
Add rustdocs

jackcmay:
- Documents
- Move Token's `Instruction` into its own file

### DIFF
--- a/ci/memo.sh
+++ b/ci/memo.sh
@@ -4,5 +4,7 @@ cd "$(dirname "$0")/.."
 
 set -e
 
+./do.sh update
 ./do.sh build memo
+./do.sh doc memo
 ./do.sh test memo
--- a/ci/token.sh
+++ b/ci/token.sh
@@ -1,11 +1,21 @@
 #!/usr/bin/env bash
 
-cd "$(dirname "$0")/../token/js"
-
 set -e
 
-npm install
-npm run build:program
-npm run test
-npm run cluster:devnet
-npm run start
+(
+    cd "$(dirname "$0")/.."
+
+    ./do.sh update
+    ./do.sh build token
+    ./do.sh doc token
+    ./do.sh test token
+)
+
+(
+    cd "$(dirname "$0")/../token/js"
+
+    npm install
+    npm run build:program
+    npm run cluster:devnet
+    npm run start
+)
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 bin
 */target/
 .vscode
+package-lock.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ install:
   - clang-7 --version
   - nvm install node
   - node --version
-  - ./do.sh update
 
 jobs:
   include:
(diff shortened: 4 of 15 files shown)


ci/memo.sh
```diff
@@ -4,5 +4,7 @@ cd "$(dirname "$0")/.."
 
 set -e
 
+./do.sh update
```
CriesofCarrots, line 7: We'll want to remove this line: https://github.com/solana-labs/solana-program-library/blob/f57149340401051b2e1c365410d48ea39b3eb09d/.travis.yml#L27

ci/token.sh
```diff
@@ -4,6 +4,11 @@ cd "$(dirname "$0")/../token/js"
 
 set -e
 
+./do.sh update
+./do.sh build memo
+./do.sh doc memo
+./do.sh test memo
```
CriesofCarrots, line 10: You've probably already noticed the dir location at the top of this file, which is why the `./do.sh` calls are failing.
Do you think we need to test all these bash pathways as well as the js analogs below?

jackcmay: Yeah, those ci changes are wip since I ran into issues with missing bin and node_modules.  I'm still seeing some issues and I'll update this pr when I have those worked out.  I do think we want to test those routes or maybe better yet remove those routes from npm since not all programs will support them.  Having those routes but not testing them are an untested source of failure over time.

CriesofCarrots: Removal from npm was the way I was leaning as well.

ci/memo.sh
jackcmay: Why?

CriesofCarrots: Well, alternately, remove `./do.sh update` from the beginning of each ci script.
As it is, `./do.sh update` will run twice in a row for each job.

ci/token.sh
```diff
@@ -1,11 +1,21 @@
...
+    ./do.sh build token
+    ./do.sh doc token
+    ./do.sh test token
+)
+
+(
+    cd "$(dirname "$0")/../token/js"
+
+    npm install
+    npm run build:program
```
CriesofCarrots, line 18: I'm not crazy about building a 2nd time.
I think if we remove this from package.json: https://github.com/solana-labs/solana-program-library/blob/f57149340401051b2e1c365410d48ea39b3eb09d/token/js/package.json#L31
we can remove this 2ndary build. Wdyt?

jackcmay: We could remove the program stuff entirely from the npm flow and instead toss up a warning message if the program does not exist and force them to go through `do.sh` for al program building/updating, etc...

CriesofCarrots: I would prefer that. (I assumed since you didn't do that since last night, you found a reason to keep it)

jackcmay: I'll do a follow-up pr for that, was focusing on just getting CI to work and sufficiently cover

CriesofCarrots: I'm also happy to do that, if you prefer

PR #41, opened 2020-06-17
Memo: add Rust program id and bump version

CriesofCarrots:


jackcmay:
@CriesofCarrots lgtm, were you also planning to do the same to `token`?

memo/Cargo.toml
```diff
@@ -3,15 +3,15 @@
 
 [package]
 name = "spl-memo"
-version = "0.1.0"
+version = "1.0.0"
 description = "Solana Program Library Memo"
 authors = ["Solana Maintainers <maintainers@solana.com>"]
```
mvines, line 8: Oh, we should mirror this commit in SPL: https://github.com/solana-labs/solana/commit/48febf153bfb176e048b727be35966075da10dd6#diff-474f9d160624516e523edbad6af0815f

jackcmay: @mvines Good point, I'll make those changes.  We should also update the other examples as well correct?

jackcmay: https://github.com/solana-labs/solana-program-library/pull/42

CriesofCarrots:
> @CriesofCarrots lgtm, were you also planning to do the same to `token`?

@jackcmay , yep! As soon as you land your changes to token today, I'll bump version and add program id

memo/src/lib.rs
```diff
@@ -3,3 +3,6 @@
 //! A simple program that accepts a string of encoded characters and verifies that it parses. Currently handles UTF-8.
 
 pub mod processor;
+
+/// The spl-memo program's on-chain program id
+pub const PROGRAM_ID: &str = "Memo1UhkJRfHyvLMcVucJwxXeuD728EqVDDwQDxFMNo";
```
mvines, line 8: Is this intentionally not a Pubkey type?

CriesofCarrots: Nope. Replaced with the `declare_id!` macro. However, CI will fail until this is released: https://github.com/solana-labs/solana/pull/10671
Hoping to get it in 1.2.4

CriesofCarrots:
@jackcmay @mvines , I think this is ready now. Planning to use this commit to release memo. Any objections?

PR #47, opened 2020-06-18
Token: add Rust program id and bump version

CriesofCarrots:


jackcmay:
@CriesofCarrots What do you think about merging this PR with a place holder id for v0.1.0 for now so that @aeyakovenko  can plumb his swap contract?  Then later when we bump the version we can put in a real program id?

PR #50, opened 2020-06-21
Token swap

aeyakovenko:
Uniswap like contract.  

Creates a liquidity pool out token out of a 2 token swap market.

https://uniswap.org/docs/v2/

tag: @jackcmay @carllin


token-swap/src/lib.rs
```diff
@@ -0,0 +1,321 @@
...
+         token_account: &AccountInfo,
+     ) -> ProgramResult {
+         // create token A account
+         let instruction_data = vec![];
+         let instruction = token::instruction::Instruction::NewAccount;
+         instruction.serialize(&mut instruction_data)?;
+ 
+         let account_addr = Pubkey::create_program_address(&[Self::to_str(instance_id), kind], program_id)?;
+         let authority = Pubkey::create_program_address(&[Self::to_str(instance_id), "authority"], program_id)?;
+         let invoked_instruction = create_instruction(
```
jackcmay, line 133: This is not a generalized helper to create instructions, as above you are referencing the `invoked` program's specific instruction creation function.  If you want to create an instruction you need to use the sdk's helpers.

token-swap/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+    ///   4. `[writable]` pool_mint Account. Must be empty, owned by $authority.
+    ///   5. `[writable]` Pool Account to deposit the generated tokens, user is the owner.
+    ///   userdata: fee rate as a ratio
+    Init((u64, u64)),
+
+    ///   Swap the tokens in the pool.
+    ///
+    ///   0. `[]` Token-swap
+    ///   1. `[]` $authority
+    ///   2. `[writable]` token_(A|B) SOURCE Account, amount is transfarable by $authority,
```
mvines, line 38:
```suggestion
    ///   2. `[writable]` token_(A|B) SOURCE Account, amount is transferable by $authority,
```

token-swap/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+    ///   6. `[writable]` Pool MINT account, $authority is the owner.
+    ///   7. `[writable]` Pool Account to deposit the generated tokens, user is the owner.
+    ///   userdata: token_a amount to transfer.  token_b amount is set by the current exchange rate.
+    Deposit(u64),
+
+    ///   Withdraw the token from the pool at the current ratio.
+    ///   
+    ///   0. `[]` Token-swap
+    ///   1. `[]` $authority
+    ///   2. `[writable]` SOURCE Pool Account, amount is transfarable by $authority.
```
mvines, line 63:
```suggestion
    ///   2. `[writable]` SOURCE Pool Account, amount is transferable by $authority.
```

token-swap/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+        }
+        match self {
+            Self::Unallocated => output[0] = 0,
+            Self::Init(swap) => {
+                if output.len() < size_of::<u8>() + size_of::<TokenSwap>() {
+                    return Err(ProgramError::InvalidAccountData);
+                }
+                output[0] = 1;
+                #[allow(clippy::cast_ptr_alignment)]
+                let value = unsafe { &mut *(&mut output[1] as *mut u8 as *mut TokenSwap) };
```
hendrikhofstadt, line 311: See pack comment above

token-swap/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+    }
+
+    /// Serializes an [SwapInstruction](enum.SwapInstruction.html) into a byte buffer
+    pub fn serialize(self: &Self) -> Result<Vec<u8>, ProgramError> {
+        let mut output = vec![0u8; size_of::<SwapInstruction>()];
+        match self {
+            Self::Init(fees) => {
+                output[0] = 0;
+                #[allow(clippy::cast_ptr_alignment)]
+                let value = unsafe { &mut *(&mut output[1] as *mut u8 as *mut (u64, u64)) };
```
hendrikhofstadt, line 146: Why not have a generic pack function like this (with proper bounds checks and maybe taking the id as argument and returning a vec):

```rust
pub fn pack<T>(input: T,  output: &mut Vec<u8>) {
    #[allow(clippy::cast_ptr_alignment)]
    let value = unsafe { &mut *(&mut output[1] as *mut u8 as *mut T) };
    *value = input;
}
```

token-swap/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+            3 => {
+                let fee: &u64 = unpack(input)?;
+                Self::Withdraw(*fee)
+            }
+            _ => return Err(ProgramError::InvalidAccountData),
+        })
+    }
+
+    /// Serializes an [SwapInstruction](enum.SwapInstruction.html) into a byte buffer
+    pub fn serialize(self: &Self) -> Result<Vec<u8>, ProgramError> {
```
hendrikhofstadt, line 140: Shouldn't the whole ser/de logic be implemented as derive macro?

token-swap/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+            .checked_div(self.pool?)?;
+        let token_b = self
+            .token_b
+            .checked_mul(user_pool)?
+            .checked_div(self.pool?)?;
+        Some((token_a, token_b))
+    }
+}
+
+impl State {
```
hendrikhofstadt, line 284: same comment on a custom ser/de macro

token-swap/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+        if let Some(spl_token::state::State::Token(token)) =
+            spl_token::state::State::deserialize(&info.data.borrow()).ok()
+        {
+            Ok(token)
+        } else {
+            Err(Error::ExpectedToken)
+        }
+    }
+
+    pub fn authority_id(program_id: &Pubkey, my_info: &Pubkey) -> Result<Pubkey, Error> {
```
hendrikhofstadt, line 348: This looks like it should rather become an SDK function.

token-swap/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+    pub fn token_mint_to(
+        accounts: &[AccountInfo],
+        swap: &Pubkey,
+        authority: &Pubkey,
+        token: &Pubkey,
+        destination: &Pubkey,
+        amount: u64,
+    ) -> Result<(), ProgramError> {
+        let swap_string = swap.to_string();
+        let signers = &[&[&swap_string[..32]][..]];
```
hendrikhofstadt, line 382: This is unintuitive and clutters the code. What about making this a function in the SDK taking n seed strings to output signers?

hendrikhofstadt: Generally the whole aspect of signer seeds is undocumented and rather unintuitive. I see why it's there and what it does but it required me to read the actual syscall implementation.

token-swap/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+    }
+}
+
+// Test program id for the swap program
+#[cfg(not(target_arch = "bpf"))]
+const SWAP_PROGRAM_ID: Pubkey = Pubkey::new_from_array([2u8; 32]);
+
+/// Routes invokes to the token program, used for testing
+#[cfg(not(target_arch = "bpf"))]
+pub fn invoke_signed<'a>(
```
hendrikhofstadt, line 710: ideally make this a generic SDK function that takes a processor function.

Generally I have the impression that the sdk needs a multi-program testing framework that registers program_ids (via a macro?) and automatically routes instructions if the target is not bpf. That would also solve this issue.

PR #53, opened 2020-06-23
Token: add Rust program id and version a couple things

CriesofCarrots:
Placeholder for #47


token/program-id.md
```diff
@@ -0,0 +1 @@
+TokenSVp5gheXUvJ6jGWGeCsgPKgnE3YgdGKRVCMY9o
```
jackcmay, line 1: What's the purpose of this file and should it match what's in `lib.rs`?

CriesofCarrots: It's just a human-readable (easy copy-pastable) version of the program id. I thought it was easier to find than the rust macro call, but we may decide to strip it out. In the meantime, yep, it should match! Sorry about that!

CriesofCarrots:
It doesn't seem like this is needed anymore.

Issue #66, opened 2020-06-30
Add rent awareness to spl-token


mvines:
Current spl-token is completed rent unaware, which leads to a couple problems:
1) If the `Token` account is not rent-exempt,  https://github.com/solana-labs/solana-program-library/blob/f8f51c13fa5dcddaf7823c33de6b7c574ee8d500/token/src/state.rs#L19-L26 and if it's rent collected then all the basic information about the token is lost, as well as the ability to mint new tokens or burn existing tokens
2) If the `Account` account  is not rent-exempt, https://github.com/solana-labs/solana-program-library/blob/f8f51c13fa5dcddaf7823c33de6b7c574ee8d500/token/src/state.rs#L41-L52 and if it's rent collected then the supply field in the `Token` account is now wrong.  

(1) is probably not a big issue, but (2) certainly is.

Possible solution:
a) Require the `Token` account to be rent-exempt
b) Remove the `supply` field from `Token` account so that an `Account` account that is not rent-exempt cannot poison the supply.   This increases the burden on Rpc and other clients, as the token supply must now be calculated by enumerating all the `Account` accounts for a given token

mvines:
An alternative to `Possible solution (b)` is to redefine the `supply` field to be "total minted supply" rather than "current supply".  Under this scheme the `Burn` instruction doesn't need the `Token` account anymore.

jackcmay:
Why do we care about total minted supply?  Isn't that something we could harvest from history?

mvines:
> Why do we care about total minted supply? Isn't that something we could harvest from history?

Tyera and I brainstormed on this a little earlier.  I initially was arguing for removing the supply entirely if it wasn't going to be able to reflect tokens that had been implicitly burned by rent delinquency.   However the utility of keeping it around as the "total minted supply" seemed to be:
1) At least you get a sense of "no worse than"
2) Tokens "burned" by transferring them to an address with no private key already show up in the "supply", so it's not exactly super accurate already

One argument I can now see for not including "total minted supply" at all, so that some tokens will be used to reflect off-chain assets.   So if tokens are being minted/burned regularly, we'll see "total minted supply" continually growing and that could confuse the situation if somebody incorrectly assumes that "total minted supply" is the actual supply of that off-chain asset.

jackcmay:
IMO the arguments for total minted supply are weak and its definition and usage will most likely be misunderstood and sew confusion.

mvines:
[...]

PR #67, opened 2020-06-30
Allow native SOL to be represented as a token

mvines:
TODO:
* [ ] Implementation and tests
* [ ] JS bindings
* [ ] Remove Burn instruction (file a follow-up issue)
* [x] Remove signer from `NewAccount` (file a follow-up issue)

### DIFF
--- a/token/src/instruction.rs
+++ b/token/src/instruction.rs
@@ -85,6 +85,24 @@ pub enum TokenInstruction {
     ///   2. `[writable]` Token being burned.
     ///   3. Optional: `[writable]` Source account if key 1 is a delegate account.
     Burn(u64),
+
+    /// Initialize a wrapped SOL token account.  The full SOL balance in the uninitialized account
+    /// will be wrapped.
+    ///
+    /// The `InitializeWrappedAccount` instruction requires no signers and MUST be included within
+    /// the Transaction that creates the uninitialized account with the system program.  Otherwise
+    /// another party can acquire ownership of the uninitialized token account.
+    ///
+    ///   0. `[writable]` An uninitialized token account
+    ///   1. `[]` Initial owner for the token account.
+    ///
+    InitializeWrappedAccount,
+
+    /// Deallocate and release all native SOL held by a token account
+    ///   0. `[signer]` Owner of the token account.
+    ///   1. `[writable]` Token account to release.
+    ///   2. `[writable]` Destination account to receive the full SOL balance of the account.
+    ReleaseAccount,
 }
 impl TokenInstruction {
     /// Deserializes a byte buffer into an [TokenInstruction](enum.TokenInstruction.html)
--- a/token/src/native_token.rs
+++ b/token/src/native_token.rs
@@ -1,0 +1,3 @@
+
+/// The token type for wrapped native SOL
+solana_sdk::declare_id!("So11111111111111111111111111111111111111111");
--- a/token/src/lib.rs
+++ b/token/src/lib.rs
@@ -7,3 +7,4 @@ pub mod instruction;
 pub mod option;
 pub mod processor;
 pub mod state;
+pub mod native_token;


token/src/instruction.rs
```diff
@@ -85,6 +85,17 @@ pub enum TokenInstruction {
...
+    /// Wrap SOL into a token account.  The full lamport balance in the new account will be wrapped
+    ///   0. `[writable]` New wrapped account being created.
+    ///   1. `[]` Owner of the new account.
+    CreateWrappedAccount,
+
+    /// Deallocate and release all native SOL held by a token account
+    ///   0. `[signer]` Owner of the token account.
+    ///   1. `[writable]` Token account to release.
+    ///   2. `[writable]` Destination account to receive the full SOL balance of the account.
+    ReleaseAccount,
```
mvines, line 98: `ReleaseAccount` works for non-wrapped token accounts as well, and thus is effectively a replacement for `Burn(u64)` but also:
1) Prevents a zero-balance token account that floats around forever
2) Assumes that we'll redefine the `supply` field according to https://github.com/solana-labs/solana-program-library/issues/66#issuecomment-651928634 such that it never decreases

token/src/instruction.rs
```diff
@@ -85,6 +85,17 @@ pub enum TokenInstruction {
     ///   2. `[writable]` Token being burned.
     ///   3. Optional: `[writable]` Source account if key 1 is a delegate account.
     Burn(u64),
+
+    /// Wrap SOL into a token account.  The full lamport balance in the new account will be wrapped
+    ///   0. `[writable]` New wrapped account being created.
```
mvines, line 90: Note: I dropped the signer requirement (`NewAccount` should do the same IMO).   Requiring a signer here means the token account can't be derived account (aka `CreateAccountWithSeed`).

token/src/native_token.rs
```diff
@@ -0,0 +1,3 @@
+
+/// The token type for wrapped native SOL
+solana_sdk::declare_id!("So11111111111111111111111111111111111111111");
```
mvines, line 3: There's no `Mint` account for wrapped SOL.   But the `Account::token` value is set to `So11111111111111111111111111111111111111111` so that the `Transfer` instruction can do the right thing if it's wrapped and for RPC inspection.

token/src/instruction.rs
jackcmay: You can wrap my account without permission?

mvines: Yes but only if you leave it in as `State::Unallocated` (ie, you don't create and initialize your account in an atomic transaction)

jackcmay: That footgun should probably be well documented

token/src/instruction.rs
jackcmay: If we don't have the requirement to partial burn tokens from an account then we should remove the `burn` instruction :-)

token/src/instruction.rs
mvines: Yeah, I'll add a comment here.   The Rust/JS bindings should do the right thing.   Note that Vote and Stake account initialization behave this way too.

token/src/instruction.rs
mvines: Yeah, totally agree!   And you can still do a partial burn by transferring to a new token account then `ReleaseAccount`ing it.

token/src/instruction.rs
mvines: I added a comment, also renamed to `InitializeWrappedAccount` since it's not actually creating the account but only initializing it

token/src/instruction.rs
jackcmay: Though that is a two-step process (one transaction), will have a performance impact depending on how fast folks want to burn stuff

mvines: I feel this is unlikely to be an issue, but let's see what comes out of that slack thread.

token/src/instruction.rs
CriesofCarrots: +1 for Initialize. This feels a lot more consistent with our native programs.

jackcmay: https://github.com/solana-labs/solana-program-library/pull/70

token/src/instruction.rs
mvines: Ok fine, no changes to `Burn` since partial burns seem to be a little popular.

For a normal non-wrapped token, we could:
1. Consider `ReleaseAccount` to also act as a full `Burn`
2. Require the non-wrapped token have a zero token balance before `ReleaseAccount` can be called on it
3. Only allow `ReleaseAccount` on wrapped tokens, which means a rent-exempt non-wrapped token account will always live forever.

I prefer 1.

CriesofCarrots: I prefer 1 also

jackcmay:
Replaced by #112

PR #68, opened 2020-06-30
Oracle pair

aeyakovenko:
# Problem

Pair for tokens that result in a binary decision from an oracle.

# Solution


oracle-token/src/lib.rs
```diff
@@ -0,0 +1,1170 @@
...
+        colleteral: u64,
+        synthetic: u64,
+        oracle_values: &OracleValues,
+    ) -> Option<u64> {
+        let (slot, price) = oracle_values.values.get(oracle_values.latest as usize)?;
+        if *slot < self.slot + info.min_slots && clock.slot < self.slot + info.max_slots {
+            return None;
+        }
+        let obligation = synthetic as f64 * price.0 as f64 / price.1 as f64;
+        let percentage = self.amount as f64 / pool as f64;
```
hendrikhofstadt, line 317: Assuming that there are 100 pool tokens and 100 collateral but no synths.
If I withdraw 100 pool tokens they are burned in `init_withdraw` which means that here the pool supply will be 0 which would result in an error.
The calculation would also yield wrong results when withdrawing 99 pool tokens etc.

Solution: use `pool + self.amount`

oracle-token/src/lib.rs
```diff
@@ -0,0 +1,1170 @@
...
+        pool: u64,
+        colleteral: u64,
+        synthetic: u64,
+        oracle_values: &OracleValues,
+    ) -> Option<u64> {
+        let (slot, price) = oracle_values.values.get(oracle_values.latest as usize)?;
+        if *slot < self.slot + info.min_slots && clock.slot < self.slot + info.max_slots {
+            return None;
+        }
+        let obligation = synthetic as f64 * price.0 as f64 / price.1 as f64;
```
hendrikhofstadt, line 346: Same here with synth balance

oracle-token/src/lib.rs
```diff
@@ -0,0 +1,1170 @@
...
+    fn calculate(
+        &self,
+        clock: &Clock,
+        info: &OracleTokenInfo,
+        pool: u64,
+        colleteral: u64,
+        synthetic: u64,
+        oracle_values: &OracleValues,
+    ) -> Option<u64> {
+        let (slot, price) = oracle_values.values.get(oracle_values.latest as usize)?;
```
hendrikhofstadt, line 342: Assuming a pool with 1:1 price
```
Pool supply: 100
Collateral: 200
Synth: 100
```

A user wants to swap 100 synth to collateral/base.

As in the comment before, `synthetic` would already be `100-100=0` making obligation `0` and percentage `100/100=1`, making amount `collateral=200`.

Applying the same solution as in the previous comment this would yield a conversion of 100->100 (before fees) which is expected.

aeyakovenko: I should stop coding at night ;).  Obviously unit tests are still TBD

PR #70, opened 2020-06-30
Token type renaming

jackcmay:
TODO: update js bindings

### DIFF
--- a/token/src/instruction.rs
+++ b/token/src/instruction.rs
@@ -22,17 +22,18 @@ pub struct TokenInfo {
 #[repr(C)]
 #[derive(Clone, Debug, PartialEq)]
 pub enum TokenInstruction {
-    /// Creates a new token and deposit all the newly minted tokens in an account.
+    /// Initializes a new mint and deposits all the newly minted tokens in an account.
     ///
     /// # Accounts expected by this instruction:
     ///
-    ///   0. `[writable, signer]` New token to create.
+    ///   0. `[writable, signer]` New mint to create.
     ///   1.
     ///      * If supply is non-zero: `[writable]` Account to hold all the newly minted tokens.
-    ///      * If supply is zero: `[]` Owner of the token.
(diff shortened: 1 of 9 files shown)


jackcmay:
@mvines inline with what you were thinking too?

token/src/instruction.rs
```diff
@@ -22,7 +22,7 @@ pub struct TokenInfo {
 #[repr(C)]
 #[derive(Clone, Debug, PartialEq)]
 pub enum TokenInstruction {
-    /// Creates a new token and deposit all the newly minted tokens in an account.
+    /// Initializes a new token and deposit all the newly minted tokens in an account.
```
CriesofCarrots, line 25:
```suggestion
    /// Initializes a new token and deposits all the newly minted tokens in an account.
```

mvines:
Nice.  How about going a little further where:
1. state::TokenMint becomes just Mint
2. state::Account now becomes Token.

jackcmay:
@mvines Looks like you are referring to the generated C file, what do you think about the Rust naming?

jackcmay:
Could rename `State::Account(Account)` to `State::Account(Token)` but that doesn't seem better

mvines:
Maybe this?
```rust
pub enum State {
    /// Unallocated state, may be initialized into another state.
    Unallocated,
    /// A token type and associated mint.
    TokenType(Mint),
    /// Some amount of tokens
    Token(TokenInfo),
    /// Invalid state, cannot be modified by the token program.
    Invalid,
}
```

I also kinda like how the ERC20 standard has a token name and token symbol strings in it (https://eips.ethereum.org/EIPS/eip-20).  There's certainly room in the `State::TokenType` enum for those strings, since the `Mint` struct is much smaller than the `TokenInfo`.   Name/symbol a little toyish, but it does allow for a generic explorer interface to display all known tokens -- which I think people in the space tend to like.

jackcmay:
@mvines @CriesofCarrots Post commit reviews welcome :-)

PR #71, opened 2020-07-01
Add in syscall stubs to fix non-bpf spl-memo builds

jackcmay:


memo/Cargo.toml
```diff
@@ -12,6 +12,7 @@ edition = "2018"
 
 [dependencies]
 solana-sdk = { version = "1.2.4", default-features = false, features=["program"] }
+solana-sdk-bpf-test = { path = "../bin/bpf-sdk/rust/test", default-features = false }
```
mvines, line 15: I think this'll break `cargo publish` for spl-memo as crates.io doesn't accept relative path dependencies

CriesofCarrots: Yeah, it will. This is what I had lined up for spl-token: https://github.com/solana-labs/solana-program-library/pull/47/commits/a64252468286b2cb06769c0d30a50402af6d7adf

PR #73, opened 2020-07-01
Version solana-sdk-bpf-test and spl-memo for crates.io

CriesofCarrots:
Cargo.toml changes to enable `cargo publish`

@mvines @jackcmay While the program itself hasn't changed, do you think I should go all the way and change the program id and republish to the clusters, since this is a version number change?

### DIFF
--- a/memo/Cargo.toml
+++ b/memo/Cargo.toml
@@ -3,7 +3,7 @@
 
 [package]
 name = "spl-memo"
-version = "1.0.0"
+version = "1.0.1"
 description = "Solana Program Library Memo"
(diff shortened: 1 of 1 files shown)


mvines:
> @mvines @jackcmay While the program itself hasn't change, do you think I should go all the way and change the program id and republish to the clusters, since this is a version number change?

IMO, no.   We'd only need to republish if there are changes to the on-chain program

CriesofCarrots:
I'm going to wait and rebase this on #74 , since the spl-memo CI technically didn't run.

PR #74, opened 2020-07-01
Fix CI

CriesofCarrots:
CI is only running token.sh. Need bullets for the additional jobs in order to run in parallel.
Also guessing the duplicate token.sh was supposed to be token-swap.sh

CC: @jackcmay


jackcmay:
Yup, thanks!

Issue #49, opened 2020-06-19
Token Multisignature Support


CriesofCarrots:
Downstream users would like to be able to ultilize multisignature on Token Accounts. A Multisig program would be an option down the road, when cross-program invocations are enabled.
In the meantime, the shortest path seems to be to add M-of-N multisig support in Token.

Instead of just one account owner, enable configuration of each Token account to have N authorities (1 <= N <= 11) and M required signers for transfers and updates to authorities and num-required-signers.

jackcmay:
@mvines @CriesofCarrots 
Rather than bloat the existing token `State` accounts with M `Pubkey`s, how about a new `MultiSig` account that contains a list of M signers?  When doing an existing token operation, check to see if the "owner" account is a `MultiSig` account and if so, validate the instruction's signers against the list of `Pubkey`s in the `MultiSig` account.
Checking for `MultiSig` would entail checking the "owner" account:
- is owned by the token program
- is account data non-zero
- does account data deserialize to a `MultiSig`

Supporting multi-signature will entail up to M optional signer accounts...

Thoughts?

CriesofCarrots:
I like it.

> Supporting multi-signature will entail up to M optional signer accounts...

So use a byte in the Token account to mark it as supporting multisig?
We also need a byte somewhere to store how many of those stored authorities need to sign a given operation (M).

mvines:
Currently we use the same account data size (`State`) for all the token program account types, so the additional M pubkeys in the `Multisig` account would still bloat up the other two account types.   Do we want to move to variable-sized token program accounts instead of the state enum?   Otherwise a new `Multisig` account type seems of limited value in terms of reducing bloat.

jackcmay:
I was thinking `Multisig` would not be part of `State`, but it's own
structure.

mvines:
We'd differentiate `State` from `Multisig` based on account data size?   I think that's probably ok (and cheap).   We might as well do the same for `Mint` as well then

jackcmay:
`Mint` because its so much smaller than the `Account` enum variant?

mvines:
Yeah, the Mint account then doesn't need to pay as much rent as an Account-sized account.   But also for consistency with the new `Multisig` account data

PR #76, opened 2020-07-02
Add Multisignture support

jackcmay:
Downstream users would like to be able to ultilize multisignature on Token Accounts. A Multisig program would be an option down the road, when cross-program invocations are enabled.
In the meantime, the shortest path seems to be to add M-of-N multisig support in Token.

Instead of just one account owner, enable configuration of each Token account to have N authorities (1 <= N <= 11) and M required signers for transfers and updates to authorities and num-required-signers.

WIP TODO:
- More tests
- Update js bindings

Fixes: #49

### DIFF
--- a/token/src/instruction.rs
+++ b/token/src/instruction.rs
@@ -8,6 +8,11 @@ use solana_sdk::{
 };
 use std::mem::size_of;
 
+/// Maximum number of multisignature signers (max N)
+pub const MAX_SIGNERS: usize = 11;
+/// Minimum number of multisignature signers (max N)
+pub const MIN_SIGNERS: usize = 1;
+
 /// Specifies the financial specifics of a token.
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Default, PartialEq)]
@@ -26,11 +31,11 @@ pub enum TokenInstruction {
     ///
     /// # Accounts expected by this instruction:
     ///
-    ///   0. `[writable, signer]` New mint to create.
+    ///   0. `[writable, signer]` Mint account to initialize
     ///   1.
     ///      * If supply is non-zero: `[writable]` Account to hold all the newly minted tokens.
     ///      * If supply is zero: `[]` Owner of the mint.
-    ///   2. Optional: `[]` Owner of the mint if supply is non-zero, if present then the 
+    ///   2. Optional: `[]` Owner of the mint if supply is non-zero, if present then the
     ///      token allows further minting of tokens.
     InitializeMint(TokenInfo),
     /// Initializes a new account.  The new account can either hold tokens or be a delegate
@@ -38,11 +43,21 @@ pub enum TokenInstruction {
     ///
     /// # Accounts expected by this instruction:
     ///
-    ///   0. `[writable, signer]`  New account being created.
+    ///   0. `[writable, signer]`  Account to initialize
     ///   1. `[]` Owner of the new account.
     ///   2. `[]` Token this account will be associated with.
     ///   3. Optional: `[]` Source account that this account will be a delegate for.
     InitializeAccount,
+    /// Initializes a multisignature account with N provided signers.
+    /// Multisignature accounts can take the place of an "Owner" account in any token instructions that
+    /// require an owner to be a signer.  The variant field represents the number of required
+    /// signers (M).
+    ///
+    /// # Accounts expected by this instruction:
+    ///
+    ///   0. `[signer]` Multisignature account to initialize
+    ///   1-11. `[]` Signer accounts, must equal to N where 1 <= N <= 11
+    InitializeMultisig(u8),
     /// Transfers tokens from one account to another either directly or via a delegate.
     ///
     /// # Accounts expected by this instruction:
@@ -51,6 +66,7 @@ pub enum TokenInstruction {
     ///   1. `[writable]` Source/Delegate account.
     ///   2. `[writable]` Destination account.
     ///   3. Optional: `[writable]` Source account if key 1 is a delegate account.
+    ///   4-14. Optional: `[Signer]` M multisignature Signer accounts
     Transfer(u64),
     /// Approves a delegate.  A delegate account is given the authority to transfer
     /// another accounts tokens without the other account's owner signing the transfer.
@@ -60,6 +76,7 @@ pub enum TokenInstruction {
     ///   0. `[signer]` Owner of the source account.
     ///   1. `[]` Source account.
     ///   2. `[writable]` Delegate account.
+    ///   3-13. Optional: `[Signer]` M multisignature Signer accounts
     Approve(u64),
     /// Sets a new owner of a token or account.
     ///
@@ -68,6 +85,7 @@ pub enum TokenInstruction {
     ///   0. `[signer]` Current owner of the token or account.
     ///   1. `[writable]` token or account to change the owner of.
     ///   2. `[]` New owner
+    ///   2-12. Optional: `[Signer]` M multisignature Signer accounts
     SetOwner,
     /// Mints new tokens to an account.
     ///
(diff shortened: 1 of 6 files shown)


jackcmay:
@mvines @CriesofCarrots First stab, thoughts?

token/src/instruction.rs
```diff
@@ -8,6 +8,11 @@ use solana_sdk::{
 };
 use std::mem::size_of;
 
+/// Maximum number of multisignature signers (max N)
+pub const MAX_SIGNERS: usize = 11;
+/// Minimum number of multisignature signers (max N)
```
mvines, line 13: maybe this?
```suggestion
/// Minimum number of multisignature signers (min N)
```

token/src/instruction.rs
```diff
@@ -26,23 +31,33 @@ pub enum TokenInstruction {
...
     ///   3. Optional: `[]` Source account that this account will be a delegate for.
     InitializeAccount,
+    /// Initializes a multisignature account with N provided signers.
+    /// Multisignature accounts can take the place of an "Owner" account in any token instructions that
+    /// require an owner to be a signer.  The variant field represents the number of required
+    /// signers (M).
+    ///
+    /// # Accounts expected by this instruction:
+    ///
+    ///   0. `[signer]` Multisignature account to initialize
```
mvines, line 58: I don't think we need this account to sign (InitializeAccount/Mint eventually too).  Same logic as https://github.com/solana-labs/solana-program-library/pull/67#discussion_r447915150.  usecase:  if the account is a derived address, that address literally can't be a signer

token/src/instruction.rs
```diff
@@ -51,6 +66,7 @@ pub enum TokenInstruction {
     ///   1. `[writable]` Source/Delegate account.
     ///   2. `[writable]` Destination account.
     ///   3. Optional: `[writable]` Source account if key 1 is a delegate account.
+    ///   4-14. Optional: `[Signer]` M multisignature Signer accounts
```
mvines, line 69: Doesn't the `0. `[signer]` Owner of the source account.` account disappear?
The single owner concept is implicit 1x1 multisigner account, so it could just be included as the first multisignature Signer account.

That `3. Optional: `[writable]` Source account if key 1 is a delegate account.` account is now clunky.  I guess it goes away entirely when we drop from supporting N delegate accounts to just 1.

token/src/instruction.rs
```diff
@@ -26,23 +31,33 @@ pub enum TokenInstruction {
...
+    /// Initializes a multisignature account with N provided signers.
+    /// Multisignature accounts can take the place of an "Owner" account in any token instructions that
+    /// require an owner to be a signer.  The variant field represents the number of required
+    /// signers (M).
+    ///
+    /// # Accounts expected by this instruction:
+    ///
+    ///   0. `[signer]` Multisignature account to initialize
+    ///   1-11. `[]` Signer accounts, must equal to N where 1 <= N <= 11
+    InitializeMultisig(u8),
```
mvines, line 60: wdyt about including N in the instruction data: 
```suggestion
    InitializeMultisig {n: u8, m: u8},
```
N is implicit by the number of signer accounts provided but for the cost of an extra 8 bits, adding an explicit N seems to be more readable

token/src/instruction.rs
```diff
@@ -68,6 +85,7 @@ pub enum TokenInstruction {
     ///   0. `[signer]` Current owner of the token or account.
     ///   1. `[writable]` token or account to change the owner of.
     ///   2. `[]` New owner
```
mvines, line 87: Document that the new owner can be a `Multisig` account?

token/src/instruction.rs
jackcmay: Yeah, was thinking about that and planning to address that together with the other instructions as a separate change

token/src/instruction.rs
jackcmay: I was planning to keep the single owner support as is which has a shorter code path and less required keys then having to implore the multi-sig route.

Yeah, I thought clunky too, and left it to get suggestions.  It would go away if we do single delegate but it's indicative of a clunkiness we will have with other programs in the future especially one's that support a group of optional sigs as multisig support does.

token/src/instruction.rs
jackcmay: Yeah, I debated that, the `InitializeMultisig` instruction documentation mentions that the multisig can take the place of any owner, I was avoiding adding a bunch of duplicate verbiage.  I'm not opposed to adding it if you think it's worth it.

jackcmay:
@mvines In general, what do you think of this approach?  In particular, identifying the multisig account type by owner and size?  Still think we should apply that to `Mint`?

token/src/instruction.rs
mvines: Keeping single owner support makes sense to me.   I'd think that from the interface perspective, using a single owner keypair would look identical to using a `Multisig` account configured at 1x1

mvines:
> what do you think of this approach? In particular, identifying the multisig account type by owner and size? Still think we should apply that to `Mint`?

sgtm.   Yeah, I feel like the Mint should support multisig too.  `MintTo` is a superuser command, great use case for multisig

token/src/instruction.rs
mvines: Ah, I guess we just need one location that says, "owner could be either a multisig address, in which case M of the N signers must be present, or any other address where the matching signer must be present"

token/src/instruction.rs
CriesofCarrots: Hmm, it looks like with this approach, the multisig account itself always has to be a signer (in order for it to appear as the first account in the metas list and be marked as the owner). That doesn't seem right to me.

CriesofCarrots: Should the accounts maybe be something more like this?
```
*  (Multiple) `[signer]` Owner of the source account or M multisignature Signer accounts
1. `[writable]` Source/Delegate account.
2. `[writable]` Destination account.
3. (Optional) `[writable]` Source account if key 1 is a delegate account.
4. (Optional) `[]` Multisignature data account
```

mvines: I was thinking:
```
    ///   0. `[writable]` Source/Delegate account.
    ///   1. `[writable]` Destination account.
    ///   2. Optional: `[writable]` Source account if key 1 is a delegate account.
    ///   2-12, or 3-13. `[Signer]` M multisignature Signer accounts (1 if a single owner address)
```

mvines: Oh wait that doesn't work

CriesofCarrots: Yeah, so the complication of storing the multisig data in a separate account is that its address has to be passed in every time

mvines: Assuming we drop the N delegate support, and each token only supports one delegate:

Single owner/delegate:
```
    ///   0. `[writable]` Source account.
    ///   1. `[writable]` Destination account.
    ///   2. `[Signer]` Owner/delegate address
```

Multisig owner/delegate:
```
    ///   0. `[writable]` Source account.
    ///   1. `[writable]` Destination account.
    ///   2. `[]` Multisig account
    ///   3-M. `[Signer]` M multisignature Signer accounts
```

jackcmay: Good point @CriesofCarrots, if its a multisig owner then the multisig does not need to be a signer.  I think this is only an issue in the instruction constructors functions as currently implemented.  Do you see the issue manifest beyond those constructors and the current instruction docs?

jackcmay: The single delegate simplifies things a lot.  I like your suggestion @mvines, removing the "optional" from the docs makes things more clear.  For Multisig owner/delegate we should to clarify that the multisig account can either be the owner or the delegate multisig account since the behavior is different between the two
`2. '[]' Owner/delegate Multisig account`

token/src/instruction.rs
jackcmay: If we do it's adding a possible failure point and more work internally to check that the user didn't mess it up (though on the other hand, it does add redundancy to make sure the user included the intended number of singers to the instruction

jackcmay:
@mvines 

> sgtm. Yeah, I feel like the Mint should support multisig too. MintTo is a superuser command, great use case for multisig

Mint already supports multisignature since its owner can be a multisignature account.  Later `SetOwner` and `MintTo` will use multisignature to validate.

My question was what do you think about moving away from `enum` for `Mint` and to a standalone structure like is proposed here for `MultiSig`?

mvines:
> My question was what do you think about moving away from `enum` for `Mint` and to a standalone structure like is proposed here for `MultiSig`?

oh sry.   Yep I think so.   Ditch the enum entirely.  It's weird to use it for some but not all the token accounts

jackcmay:
@mvines @CriesofCarrots 

This PR is rendered obsolete by these changes: https://github.com/jackcmay/solana-program-library/pull/1

PR #77, opened 2020-07-03
Single delegate support

jackcmay:
Problem:
The Token program currently supports unlimited delegates associated with a single token account.
Doing so:
- Overly complicates both the interface and implementation of the token program
- Requires each delegate to be its own token account, adding unnecessary steps to the process
- Any instruction that that does delegation must include both the delegate account and the associated source account.  This is especially painful for programs like token-swap where it operates on a number of delegated accounts at the same time

Proposed solution:
Switch the token program to supporting one delegate per token account.

(Removes ~1000 lines of implementation and test code 🎉 )

### DIFF
--- a/token/src/instruction.rs
+++ b/token/src/instruction.rs
@@ -22,7 +22,7 @@ pub struct TokenInfo {
 #[repr(C)]
 #[derive(Clone, Debug, PartialEq)]
 pub enum TokenInstruction {
-    /// Initializes a new mint and deposits all the newly minted tokens in an account.
+    /// Initializes a new mint and optionally deposits all the newly minted tokens in an account.
     ///
     /// # Accounts expected by this instruction:
     ///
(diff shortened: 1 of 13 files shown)


token/src/instruction.rs
```diff
@@ -22,69 +22,66 @@ pub struct TokenInfo {
...
+    /// delegate is zero then delegation is rescinded
     ///
     /// # Accounts expected by this instruction:
     ///
-    ///   0. `[signer]` Owner of the source account.
-    ///   1. `[]` Source account.
-    ///   2. `[writable]` Delegate account.
+    ///   0. `[writable]` The source account.
+    ///   1. Optional: `[writable]` The delegate if amount is non-zero.
+    ///   2. `[signer]`The source account owner address
```
mvines, line 60: Just a nit:  make the optional account the 3rd account?  I think that's generally easier to code up, and also express in the documentation.    
As a motivating example, the comment "2. `[signer]`The source account owner address" is incomplete because the source account could also be at position 1 if no delegate is present.

token/src/state.rs
```diff
@@ -17,38 +17,29 @@ use std::mem::size_of;
...
     pub amount: u64,
-    /// If `delegate`  is None, `amount` belongs to this account.
-    /// If `delegate` is Option<_>, `amount` represents the remaining allowance
-    /// of tokens this delegate is authorized to transfer from the `source` account.
-    pub delegate: COption<AccountDelegate>,
+    /// If `delegate` is `Some` then `delegated_amount` represents
+    /// the amount authorized by the delegate
+    pub delegate: COption<Pubkey>,
+    /// The amount delegated
+    pub delegated_amount: u64,
```
mvines, line 42: wdyt about wrapping the amount up in the `COption` as well?
```suggestion
    /// If `delegate` is `Some` then `delegated_amount` represents
    /// the amount authorized by the delegate and the amount delegated
    pub delegate: COption<(Pubkey, u64)>,
```

token/src/instruction.rs
jackcmay: @mvines Yeah, approve was the weird one trying to support "Revoke" because of the optional parameter but I left it this way since I was about to add multisig which would result in a similar "motivating example".  Also, if the accounts are 1, 2, 3, and 2 is optional it's probably fair to expect folks to understand that 3 moves to 2 if optional 2 is not included.  I think the real answer here is we should avoid optional as much as possible and a separate revoke instruction makes more sense.

token/src/state.rs
jackcmay: I would be more inclined to do something like
```rust
put struct DelegateInfo {
    pub key: Pubkey,
    pub amount: u64,
};
pub struct Account {
...
    pub delegate: COption<DelegateInfo>
}
```

mvines: Yeah that's definitely better than an unnamed tuple

Issue #75, opened 2020-07-01
Token-swap CI js fails


CriesofCarrots:
The js commands in token-swap.sh are not successful:
![image](https://user-images.githubusercontent.com/3030561/86274793-740dcb80-bb8f-11ea-9442-4bbfd81daedd.png)

PR #81, opened 2020-07-07
Fix automation

jackcmay:
Fixes: #75


jackcmay:
The first time through this PR should fail, checking that failures are actually caught

PR #82, opened 2020-07-07
Remove supply from Token

jackcmay:
- Tracking total supply may be misinterpreted and lead to confusion If accounts are removed due to rent delinquency.
- Removing supply reduces a required key in the `Burn` instruction


jstarry:
If the pool is measured solely on token a, then I think fees could only accumulate in token b's balance

jackcmay:
Hehe, yup, I threw this up late last night when I got to that point to
garner input, looks like it worked 😁

jstarry:
Hehe, I'm just catching up to all the awesome work you've been doing!

jackcmay:
Looks like the swap math might be off and I don't think token-swap needs supply, there are other ways to track it (like swap tracks it itself).  More work to do on token-swap.  Any objections to removing supply from the Token program?

PR #90, opened 2020-07-07
Use variant fields

jackcmay:
Unnamed fields are confusing to document and not conducive to auto-generation by procedural macros


jackcmay:
@CriesofCarrots Does this also help your procedural macro work?

CriesofCarrots:
> @CriesofCarrots Does this also help your procedural macro work?

Yes, this is great!

Issue #86, opened 2020-07-07
Other token accounts can be mistaken for multisig accounts


jstarry:
#### Problem
Accounts are determined to be multisig if their data size equals the size of a multisig account. There is no restriction on non-multisig accounts that they must not equal the size of a multisig account so other token account types could be mistaken as a multisig account if they are big enough and are set as a token account owner

Doesn't seem likely that this would happen in practice, but if it did, it could result in an owner account that has always passes the owner validation check if `m` is 0.

#### Proposed Change
* Assert that token account types are different sizes
* Require token accounts to have a specific size

mvines:
I think we just need a unit test that asserts the sizes of the three token account types are not the same

jstarry:
We don't enforce the data length of token accounts when they are initialized though

mvines:
Ha, yeah we might need to add handling to reject a multisig account where N == 0 as well.  It looks like we might accept that

jstarry:
There is a check for that: `is_valid_signer_index`

mvines:
Oh good so then we just need this unit test then: https://github.com/solana-labs/solana-program-library/issues/86#issuecomment-654977104

jackcmay:
Do you mean a test like this? ;-)
https://github.com/solana-labs/solana-program-library/blob/07d249b6908421ab340c13f5bba290bbcc023309/token/src/state.rs#L597

I'll add the account size assertions too as I'm looking into breaking up the 'State' enum

mvines:
Yep!

PR #93, opened 2020-07-08
Whittle down the token program across the board

jackcmay:
The changes in this PR reduce the size of the token program:

- Split the `State` enum into individual structures so as not to incur largest variant size of all variants
- Generalize `State` unpack
- Remove `State` serialize and instead operate directly on `&mut`
- Enforce that only initialized accounts are used across the board

Reduces:
- BPF instructions for each `TokenInstruction` by an average of 30%, all instructions saw reductions
- Stripped BPF shared object file size from 186 to 86 kB
- Reduces the Mint data size from 128 to 40 bytes
- Reduces the Account data size from 128 to 120 bytes

Fixes: #86

### DIFF
--- a/token/src/state.rs
+++ b/token/src/state.rs
@@ -21,7 +21,7 @@ pub struct Mint {
     pub owner: COption<Pubkey>,
     /// Number of base 10 digits to the right of the decimal place.
     pub decimals: u8,
-    /// Is `true` if this mint has been initialized
+    /// Is `true` if this structure has been initialized
     pub is_initialized: bool,
 }
 impl IsInitialized for Mint {
@@ -43,7 +43,7 @@ pub struct Account {
     /// If `delegate` is `Some` then `delegated_amount` represents
     /// the amount authorized by the delegate
     pub delegate: COption<Pubkey>,
-    /// Is `true` if this mint has been initialized
+    /// Is `true` if this structure has been initialized
     pub is_initialized: bool,
     /// The amount delegated
     pub delegated_amount: u64,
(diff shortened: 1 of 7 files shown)


jackcmay:
@mvines fyi. Splitting the enum up worked out very nicely

token/src/state.rs
```diff
@@ -11,21 +11,26 @@ use solana_sdk::{
...
-pub struct Token {
+pub struct Mint {
     /// Optional owner, used to mint new tokens.  The owner may only
     /// be provided during mint creation.  If no owner is present then the mint
     /// has a fixed supply and no further tokens may be minted.
     pub owner: COption<Pubkey>,
     /// Number of base 10 digits to the right of the decimal place.
     pub decimals: u8,
+    /// Is `true` if this mint has been initialized
+    pub is_initialized: bool,
```
mvines, line 25: Nit: I'd move this to the front of the struct

token/src/state.rs
```diff
@@ -38,59 +43,38 @@ pub struct Account {
     /// If `delegate` is `Some` then `delegated_amount` represents
     /// the amount authorized by the delegate
     pub delegate: COption<Pubkey>,
+    /// Is `true` if this mint has been initialized
```
mvines, line 46: "mint" - is that a copy-paste bug?

token/src/state.rs
CriesofCarrots: And same for the other structs?

jackcmay: Yeah, I had it there for aesthetic reasons originally but it packs tighter to have the `u8`'s where they are now

jackcmay: Originally I didn't have these and was going to just check some known value in the structure but that's pretty weak and for `Mint` there isn't anything definitive that I can check.  The 'Multisig' structure increased by one byte but I poked the `is_initialized` member into structure padding for `Mint` and `Account`.  

On a separate note, we could use packed structs to save a few bytes, I can run a few tests to see how much of a hit that is performance-wise

mvines: > I had it there for aesthetic reasons originally but it packs tighter to have the `u8`'s where they are now

Ah, ok that explanation is good enough for me :)

PR #105, opened 2020-07-08
Version C header

jackcmay:
Add the rust crate version information to the C headers


jackcmay:
@t-nelson How about this?

token/inc/token.h
```diff
@@ -7,6 +7,10 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+#define TOKEN_MAJOR_VERSION 0
+#define TOKEN_MINOR_VERSION 1
+#define TOKEN_PATCH_VERSION 0
+
```
CriesofCarrots, line 13: Just to be clear, when we're ready for this (https://github.com/solana-labs/solana-program-library/pull/47), I just need to build token to update these, right?

jackcmay:
@CriesofCarrots That's correct, they be updated to the latest rust package version during the header generation

PR #138, opened 2020-07-21
Fix C header to include state structures

jackcmay:
Auto-generated headers did not include the Token state definitions


jackcmay:
@CriesofCarrots Doesn't change the program, just exports the token states to C, we probably don't need to push a new version

PR #147, opened 2020-07-22
Move targetDir inside perform_action

### LINKED ISSUE #146: do.sh build all doesn't strip .so files
Because targetDir is defined outside `perform_action`, it keeps the positional argument `$2` from the original command. In the case of `./do.sh build all`, this is `all`. Therefore this check fails: https://github.com/solana-labs/solana-program-library/blob/2c82a97909e249a1e42d25ad4c6a4c1f563d1870/do.sh#L40
because `~/solana-program-library/all/target/bpfel-unknown-unknown/release` does not exist.

CriesofCarrots:
Fixes #146

### DIFF
--- a/do.sh
+++ b/do.sh
@@ -37,10 +37,8 @@ perform_action() {
 
         so_path="$targetDir/$profile"
         so_name="spl_${3%/}"
-        if [ -f "$so_path/${so_name}.so" ]; then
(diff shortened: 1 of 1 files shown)


jackcmay:
does `targetDir` need to reference `$3` instead of `$2`?

jackcmay:
Also, we probably don't need or want that if statement, should always strip

`if [ -f "$so_path/${so_name}.so" ]; then`

PR #165, opened 2020-07-30
Prep token/js for publishing

jackcmay:


token/js/client/token.js
```diff
@@ -163,13 +163,12 @@ type MultisigInfo = {|
...
   signer3: PublicKey,
   signer4: PublicKey,
-  signer4: PublicKey,
   signer5: PublicKey,
   signer6: PublicKey,
   signer7: PublicKey,
   signer8: PublicKey,
   signer9: PublicKey,
-  signer10: PublicKey,
+  signer20: PublicKey,
```
jstarry, line 171: 👀

token/js/package.json
```diff
@@ -1,11 +1,28 @@
...
   },
+  "bugs": {
+    "url": "https://github.com/solana-labs/solana-program-library/issues"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "lib/index.cjs.js",
+  "module": "lib/index.esm.js",
+  "types": "lib/index.d.ts",
```
jstarry, line 20: You'll need to setup a build command to create these files. Looks like you'll also want a `token/js/index.js` file to export everything that you've added types for in `module.d.ts`

token/js/package.json
```diff
@@ -1,13 +1,30 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "client/token.js",
+  "types": "module.d.ts",
+  "files": [
+    "/cli",
+    "/client",
+    "/module.flow.js"
```
jstarry, line 23: Oh I think you need to add `/module.d.ts` here as well. I don't think npm is smart enough to include due to the `"types"` field

PR #169, opened 2020-07-30
token: Add default/program crate features to match solana-sdk

mvines:
The spl-token crate is not useful for Rust clients because it forces the "program" feature in solana-sdk, which turns off a variety of things.   We have yet to see this issue in the monorepo by fluke.

#### Simple STR:
1. `cargo init`
2. Add `solana-sdk = "1.2"` and `spl-token = "1.0"` to Cargo.toml
3. `use solana_sdk::signature::Signature;` in `src/main.rs`.   

#### Result:
```
error[E0432]: unresolved import `solana_sdk::signature`
 --> src/main.rs:1:17
  |
1 | use solana_sdk::signature::Signature;
  |                 ^^^^^^^^^ could not find `signature` in `solana_sdk`
```

Depends-on:

### DIFF
--- a/token/src/state.rs
+++ b/token/src/state.rs
@@ -6,9 +6,13 @@ use crate::{
     option::COption,
 };
 use solana_sdk::{
-    account_info::next_account_info, account_info::AccountInfo, entrypoint::ProgramResult, info,
+    account_info::next_account_info, account_info::AccountInfo, entrypoint::ProgramResult,
     program_error::ProgramError, pubkey::Pubkey,
 };
+#[cfg(feature = "program")]
+use solana_sdk::info;
+#[cfg(not(feature = "program"))]
+use log::info;
 use std::mem::size_of;
 
 /// Mint data.
@@ -549,6 +553,7 @@ pub trait IsInitialized {
 }
 
 // Pull in syscall stubs when building for non-BPF targets
+#[cfg(feature = "program")]
 #[cfg(not(target_arch = "bpf"))]
 solana_sdk::program_stubs!();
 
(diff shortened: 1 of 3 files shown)


ci/token.sh
```diff
@@ -13,8 +14,12 @@ cd "$(dirname "$0")/.."
 ./do.sh test token
 cc token/inc/token.h -o token/target/token.gch
 
-cd "$(dirname "$0")/../token/js"
+# Rust API
+cd "$(dirname "$0")/../token"
+cargo test
```
jackcmay, line 19: this should be the same as `./do.sh build-native token`

token/src/state.rs
```diff
@@ -549,6 +553,7 @@ pub trait IsInitialized {
 }
 
 // Pull in syscall stubs when building for non-BPF targets
+#[cfg(feature = "program")]
```
jackcmay, line 556: If these don't exist in the non-bpf, non-program configuration there will be unresolved symbols, hence the stubs

token/src/state.rs
```diff
@@ -6,9 +6,13 @@ use crate::{
     option::COption,
 };
 use solana_sdk::{
-    account_info::next_account_info, account_info::AccountInfo, entrypoint::ProgramResult, info,
+    account_info::next_account_info, account_info::AccountInfo, entrypoint::ProgramResult,
     program_error::ProgramError, pubkey::Pubkey,
 };
+#[cfg(feature = "program")]
```
jackcmay, line 12: Side change to help logging in non-bpf environments?

ci/token.sh
mvines: ah, ok.  removed.

token/src/state.rs
mvines: Yes I'll make the solana-sdk changes so that cb5f1e245fe285cbd688f1cd79c2f51f4bcae27a won't need to be merged

jackcmay:
@mvines How about this instead?

PR #181, opened 2020-07-31
Programs can be client dependencies

jackcmay:
The spl-token crate is not useful for Rust clients because it forces the "program" feature in solana-sdk, which turns off a variety of things. We have yet to see this issue in the monorepo by fluke.

Simple STR:
cargo init
Add solana-sdk = "1.2" and spl-token = "1.0" to Cargo.toml
use solana_sdk::signature::Signature; in src/main.rs.
Result:
error[E0432]: unresolved import `solana_sdk::signature`
 --> src/main.rs:1:17
  |
1 | use solana_sdk::signature::Signature;
  |       

**Note: These changes depend on https://github.com/solana-labs/solana/pull/11296 and CI will fail until its pulled in**


mvines:
Nice

PR #187, opened 2020-07-31
Consistency nits

CriesofCarrots:
- Spell out destination
- `authority` is used when either owner or delegate may exercise an action. In this case, only `owner` is authorized.


jackcmay:
Thanks!

PR #203, opened 2020-08-03
Return signature from sendAndConfirmTransaction

jstarry:
#### Problem
sendAndConfirmTransaction doesn't return the signature which is being confirmed. Callers may wish to receive the signature to fully confirm the transaction

#### Changes
- Return signature
- Remove unused yaml notify stuff


jackcmay:
Looks good, wanna also apply this to token-swaps copy (or better yet combine the two :-) )

jstarry:
> wanna also apply this to token-swaps copy

done!

> or better yet combine the two 

later 😝

Issue #200, opened 2020-08-02
spl-token should use 256 bits instead of 64 bits to represent an amount


mvines:
ERC20 uses 256 bits right now, so spl-token can't map 1:1 to ERC20.  This seems like a general usability problem of spl-token.

Along the way let's make 18 decimals the (strongly recommended) default for spl-token

mvines:
This will have ripple effects though the `getToken*` RPC methods as well.  (cc: @CriesofCarrots)

CriesofCarrots:
Also affects `account-decoder/parse_token.rs` and `transaction-status/parse_token.rs`

mvines:
> Along the way let's make 18 decimals the (strongly recommended) default for spl-token

Wrapped SOL should still have 9 decimals though - largely just a documentation update since a native mint account doesn't actually exist

mvines:
Alternative path: a spl-erc20-token or spl-wrapped-erc20 program that supports the 256bit amount needed for 1:1 erc20 compatibility.   Ideally such a program reuses virtually everything from spl-token, parameterizing the amount

mvines:
Better alternative path:  spl-token 1.1 can switch between u64 and u256 depending on the mint configuration

mvines:
Better better alternative path: 👢.  Back to the "alternative path".   There can exist different token standard that the Eth bridge can deal with.  Native SPL-based marketplaces don't need incur the cost of Eth compatibility.

mvines:
I think we've finally settled this.  64bits only

PR #208, opened 2020-08-04
Use 256-bit amounts

jackcmay:
Switch Token to 256-bit amounts to help interoperability between other chains

256-bit support is provided by the `primitive-types` crate.  That crate's implementation of `overflow_pow` uses a lot of `inline` and macros which result is very large stack usage.  The high stack usage overflows the current BPF stack size of 4k.  The use of `primitive-types` in these changes is safe because Token does not use `oveflow_pow`.  To accommodate we may have to increase the BPF stack size.

Fixes: #200


mvines:
Best plan I've heard so far: allow the Mint to indicate if it's a u64 or u256 token, so we can support both amount sizes using the same spl-token 1.1 program.

jackcmay:
Shelved

ZhengYuTay:
@mvines  Is there still a plan for Mint to support u256

PR #211, opened 2020-08-05
spl-token: Add ui amount conversion helpers

mvines:
@CriesofCarrots - you _might_ find these useful in RPC


CriesofCarrots:
> @CriesofCarrots - you _might_ find these useful in RPC

For sure, once they're in a published crate, I can use here: https://github.com/solana-labs/solana/blob/a11a25fd9087d6747f11ccdb1a5993615646fa02/core/src/rpc.rs#L1132

PR #228, opened 2020-08-06
Safely handle ptr alignment

jackcmay:
Token program is performing misaligned read/writes.

This PR changes the read/writes to either use `read/write_unaligned` or to enforce a pointer is aligned before reading/writing it from a pointer.

Depends on: https://github.com/solana-labs/solana/pull/11384 and requires BPFLoader2 as a program account owner


jackcmay:
Fixed via new BPFLoader, local program heap, and new pack/unpack trait implementations

PR #241, opened 2020-08-08
Avoid generating header files when they already exist

mvines:


ci/token-swap.sh
```diff
@@ -5,11 +5,14 @@ set -ex
...
 cd "$(dirname "$0")/.."
 ./do.sh update
-./do.sh build token
 ./do.sh fmt token-swap --all -- --check
-./do.sh build-lib token-swap -D warnings
-./do.sh build token-swap
 ./do.sh clippy token-swap -- --deny=warnings
+
+SPL_CBINDGEN=1 ./do.sh build-lib token-swap -D warnings
+git diff --exit-code token-swap/inc/token-swap.h
```
jackcmay, line 12: Does this have to be disabled when interface changes are required?

mvines: You'd just build `token-swap/inc/token-swap.h` locally and commit it as a part of the PR

PR #255, opened 2020-08-09
Add budget

### LINKED ISSUE #7: Missing smart contracts example
#### Problem

Solana distinguishes *programs* from *smart contracts*, but the SPL doesn't contain an example of that.

#### Proposed Solution

Add `Budget` from the Solana repo.

garious:
#### Problem

No way to transfer funds without risking loss of tokens when recipient doesn't hold the private key.

#### Proposed changes

Add Budget from the monorepo, which allows the user to put tokens into a smart contract and only transfer them to the recipient once they send the contract a signed instruction, which proves they hold the private key.

Fixes #7


mvines:
@garious - I'll look at this CI failure, it's obviously not related to this PR.   Overall I still need to hacking on `ci/`, it's not in a state that I'm ok with yet.  I'd suggest hanging on until I complete that work before trying to bolt Budget up to CI for real

garious:
No hurry, thanks

jackcmay:
@garious Nice to have different eyes on this stuff, how did the port go?

garious:
Not ported to BPF yet. Still native

mvines:
The underlying ci/ infra should be more ready for this PR now, when you care to rebase Greg

PR #271, opened 2020-08-14
Rework ci/script.sh to reduce duplication and unnecessary rebuilds 

mvines:


mvines:
This seemed to have shaved off 10minutes from PR CI, so a step in the right direction!

PR #277, opened 2020-08-14
Add token/program2

### LINKED ISSUE #212: spl-token: Ensure that `spl_token::native_mint::id()` is changed whenever the spl-token program id is changed
The native_mint id, like all other mints, should always be changed whenever the spl-token program id is changed.  RPC, in particular, makes the valid assumption that this is the case.

A human solution: document and remember to change it
A programatic solution: change the last ~32 digits of the native_mint id to be the same as the spl-token program id

### LINKED ISSUE #230: Reorganize source tree layout based on program version
The current program source layout is inadequate for working on newer version of a SPL program while maintaining the existing deployed versions.  

Proposal is to refactor up the current `token/` directory into:
```
token-1.0/  <-- represents the 1.0.0 token that has already been deployed to mainnet-beta
token-dev/ <-- the 1.1.0 token that will be deployed sometime in the future.
```

When `token-dev/` is ready to be deployed, it's copied into `token-1.1/` and then `token-dev/` moves up to "1.2.0".  

The benefit of this approach is a clear representation of the on-chain source code for a given token version **while** retaining the ability to ship new patch releases for each token crate.  Of course since token 1.0.0 program is deployed, it can't be changed but the client portion of the 1.0.0 token may still need to change from time to time (such as if we ship a breaking change in the solana-sdk crate).

The same model applies to the other SPL programs.  For `token-swap` there would only be a `token-swap-dev` since 1.0.0 has not shipped yet.

mvines:
As mentioned at https://github.com/solana-labs/solana-program-library/issues/230#issuecomment-674283283, `token/program` is now the first version of the token program (previously called "spl-token 1.0") that has already been deployed to mainnet-beta.  `token/program/src/processor.rs` in particular should now be considered immutable.

`token/program2` is the second version of the token program (previously called "spl-token 1.1") where new development occurs and will be deployed to mainnet-beta at some point in the future.

This PR also backs out #210 from `token/program`, which does not exist in the deployed program.  This will be a feature of program2.

Fixes #230
Fixes #212


mvines:
@jackcmay - wdyt?

Issue #289, opened 2020-08-19
token: Consider adding an expected decimals input to all token instructions that deal with an amount


mvines:
The user of a token program, such as a generic wallet or the `spl-token` command-line, doesn't generically have trusted way to determine the correct decimals value for a given token.   RPC will tell it, but maybe RPC is lying.  The user might know and can provide it, but maybe they are mistaken.  

As a safety/sanity check, if the decimals are provided as input into each instruction alongside the amount, the program itself can confirm the decimals are correct before performing the operation.

PR #295, opened 2020-08-20
Add Transfer2 instruction

mvines:
Here's what I'm thinking for how to add decimal validation to the Token instructions.   I only plumbed Transfer2 for now, if this direction looks good then all the other existing instructions with an `amount` input will get the same `2` treatment.

Towards https://github.com/solana-labs/solana-program-library/issues/289

### DIFF
--- a/token/program2/src/instruction.rs
+++ b/token/program2/src/instruction.rs
@@ -187,6 +187,31 @@ pub enum TokenInstruction {
     ///   2. `[]` The account's multisignature owner.
     ///   3. ..3+M '[signer]' M signer accounts.
     CloseAccount,
+
+    /// Transfers tokens from one account to another either directly or via a delegate.  If this
+    /// account is associated with the native mint then equal amounts of SOL and Tokens will be
+    /// transferred to the destination account.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[]` The mint for this account
+    ///   1. `[writable]` The source account.
+    ///   2. `[writable]` The destination account.
+    ///   3. '[signer]' The source account's owner/delegate.
+    ///
+    ///   * Multisignature owner/delegate
+    ///   0. `[]` The mint for this account
+    ///   1. `[writable]` The source account.
+    ///   2. `[writable]` The destination account.
+    ///   3. '[]' The source account's multisignature owner/delegate.
+    ///   4. ..3+M '[signer]' M signer accounts.
+    Transfer2 {
+        /// The amount of tokens to transfer.
+        amount: u64,
+        /// Number of base 10 digits to the right of the decimal place.
+        decimals: u8,
+    },
 }
 impl TokenInstruction {
     /// Unpacks a byte buffer into a [TokenInstruction](enum.TokenInstruction.html).
@@ -249,6 +274,16 @@ impl TokenInstruction {
                 Self::Burn { amount }
             }
             9 => Self::CloseAccount,
+            10 => {
+                if input.len() < size_of::<u8>() + size_of::<u64>() + size_of::<u8>() {
+                    return Err(TokenError::InvalidInstruction.into());
+                }
+                #[allow(clippy::cast_ptr_alignment)]
+                let amount = unsafe { *(&input[size_of::<u8>()] as *const u8 as *const u64) };
+                let decimals =
+                    unsafe { *(&input[size_of::<u8>() + size_of::<u64>()] as *const u8) };
+                Self::Transfer2 { amount, decimals }
+            }
             _ => return Err(TokenError::InvalidInstruction.into()),
         })
     }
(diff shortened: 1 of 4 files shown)


mvines:
@CriesofCarrots / @jackcmay - wdyt?

do.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+CALLER_PWD=$PWD
```
jackcmay, line 3: Just curious, why do it this way?

token/program2/src/instruction.rs
```diff
@@ -187,6 +187,32 @@ pub enum TokenInstruction {
     ///   2. `[]` The account's multisignature owner.
     ///   3. ..3+M '[signer]' M signer accounts.
     CloseAccount,
+
```
jackcmay, line 190: Is this at the end in an attempt to be backward compatible with spl-token 1.0?

token/program2/src/instruction.rs
```diff
@@ -187,6 +187,32 @@ pub enum TokenInstruction {
...
+    /// Transfers tokens from one account to another either directly or via a delegate.  If this
+    /// account is associated with the native mint then equal amounts of SOL and Tokens will be
+    /// transferred to the destination account.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[]` The mint for this account
+    ///   1. `[writable]` The source account.
```
jackcmay, line 199: Two source accounts?

token/program2/src/instruction.rs
```diff
@@ -187,6 +187,32 @@ pub enum TokenInstruction {
...
+    ///   2. `[writable]` The source account.
+    ///   3. `[writable]` The destination account.
+    ///   4. '[signer]' The source account's owner/delegate.
+    ///
+    ///   * Multisignature owner/delegate
+    ///   0. `[]` The mint for this account
```
jackcmay, line 205: Could stick the u8 decimals in the account structure rather than the overhead of loading and spending cycles working with the mint

token/program2/src/processor.rs
```diff
@@ -127,8 +127,16 @@ impl Processor {
         program_id: &Pubkey,
         accounts: &[AccountInfo],
         amount: u64,
+        decimals: Option<u8>,
     ) -> ProgramResult {
         let account_info_iter = &mut accounts.iter();
+
+        let mint_account_info = if let Some(decimals) = decimals {
```
jackcmay, line 134: `decimals.is_some()`?

token/program2/src/processor.rs
```diff
@@ -149,6 +157,18 @@ impl Processor {
             return Err(TokenError::MintMismatch.into());
         }
 
+        if let Some((mint_account_info, decimals)) = mint_account_info {
```
jackcmay, line 160: This could be a u8 check between the passed valuea nd the src account since we don't really need to validate mint, we just want to express the decimals in the instruction data

do.sh
mvines: Do you see a better way?  One day I'll murder `do.sh`

token/program2/src/instruction.rs
mvines: It means I don't need to make a million changes so TokenInstruction::unpack/pack

token/program2/src/instruction.rs
mvines: There's a couple problems with that:
1. We got strong encouragement today to avoid changing the Account structure if at all possible
2. I actually do want to validate the mint too.  More on this below...

token/program2/src/processor.rs
mvines: I need the unwrapped decimals below, and am trying to avoid an explicit .unwrap()

do.sh
jackcmay: Didn't understand the intent, just noticed your first commit description

token/program2/src/processor.rs
mvines: Why we need `mint` in `Transfer2`:
When signing a `Transfer2` on a ledger device, the ledger program has no access to the on-chain accounts.   The ledger program may only choose to support well-known mints (at which point it knows `decimals` and technically doesn't need to include them in `Transfer2`).   But it definitely wants to assert the `mint` of the the to/from token accounts.

Why we need `decimals` in  `Transfer2`:
You try to generically sign a token transfer based on mint data received over RPC.   But you can't trust that RPC isn't lying to you and turning that 0.1 token into a 100000 token.   So in this case, you need to assert both the mint and decimals.

token/program2/src/instruction.rs
jackcmay: For #1 we would stick it at the end

do.sh
mvines: Let me break the `do.sh` change into a different PR, it doesn't belong in here :)

token/program2/src/instruction.rs
mvines: Yeah I know.  We could argue for that if we need to.  I don't think it helps in this case though because I do want to verify the mint

mvines:
Oh also I went for `Transfer2` instead of just replacing `Transfer` since I felt there would be times when the caller just doesn't want to bother with the extra verification.

token/program2/src/processor.rs
jackcmay: Yeah, makes sense, writing a client helps expose these use cases

mvines:
I'll probably keep this PR parked until we have a resolution on #315, to avoid adding a bunch of pack/unpack code that will just need to be rewritten

mvines:
Replaced by https://github.com/solana-labs/solana-program-library/pull/343

PR #297, opened 2020-08-21
Add capability to freeze token Accounts

### LINKED ISSUE #229: No mechanism for freezing token accounts
An authority may want the ability to "freeze" specific accounts.

Add a new authority to the mint and a new flag in accounts.  The authority is able to toggle that flag.  Setting the flag to "frozen" disables the account owner and delegate's authority.

CriesofCarrots:
An authority may want the ability to "freeze" specific accounts.

Add a new authority to the Mint and a new initialized state variant to Account. The authority is able to toggle that variant for an associated account. Setting the variant to "frozen" disables any changes from being made to the account, ie transfer in or out, approve, revoke, set-authority, burn.

Fixes #229

### DIFF
--- a/token/program2/src/instruction.rs
+++ b/token/program2/src/instruction.rs
@@ -140,7 +140,10 @@ pub enum TokenInstruction {
     ///   1. `[]` The new owner/delegate/multisignature.
     ///   2. `[]` The mint's or account's multisignature owner.
     ///   3. ..3+M '[signer]' M signer accounts
-    SetOwner,
+    SetAuthority {
+        /// The type of authority to update.
+        authority_type: AuthorityType,
+    },
     /// Mints new tokens to an account.  The native mint does not support minting.
     ///
     /// Accounts expected by this instruction:
@@ -250,7 +253,17 @@ impl TokenInstruction {
                 Self::Approve { amount }
             }
             5 => Self::Revoke,
-            6 => Self::SetOwner,
+            6 => {
+                if input.len() < size_of::<u8>() + size_of::<u8>() {
+                    return Err(TokenError::InvalidInstruction.into());
+                }
+                let authority_type = match input[1] {
+                    0 => AuthorityType::Owner,
+                    1 => AuthorityType::Freezer,
+                    _ => return Err(TokenError::InvalidInstruction.into()),
+                };
+                Self::SetAuthority { authority_type }
+            }
             7 => {
                 if input.len() < size_of::<u8>() + size_of::<u64>() {
                     return Err(TokenError::InvalidInstruction.into());
@@ -338,8 +351,15 @@ impl TokenInstruction {
                 output[output_len] = 5;
                 output_len += size_of::<u8>();
             }
-            Self::SetOwner => {
+            Self::SetAuthority { authority_type } => {
                 output[output_len] = 6;
+                output_len += size_of::<u8>();
+
+                let byte = match authority_type {
+                    AuthorityType::Owner => 0,
+                    AuthorityType::Freezer => 1,
+                };
+                output[output_len] = byte;
                 output_len += size_of::<u8>();
             }
             Self::MintTo { amount } => {
@@ -369,6 +389,16 @@ impl TokenInstruction {
         output.truncate(output_len);
         Ok(output)
     }
+}
+
+/// Specifies the authority type for SetAuthority instructions
+#[repr(u8)]
+#[derive(Clone, Debug, PartialEq)]
+pub enum AuthorityType {
+    /// General authority, valid for Account and Mint
+    Owner,
+    /// Freeze authority, only valid for Mint
+    Freezer,
 }
 
 /// Creates a 'InitializeMint' instruction.
@@ -553,15 +583,16 @@ pub fn revoke(
     })
 }
 
-/// Creates a `SetOwner` instruction.
-pub fn set_owner(
+/// Creates a `SetAuthority` instruction.
+pub fn set_authority(
     token_program_id: &Pubkey,
     owned_pubkey: &Pubkey,
     new_owner_pubkey: &Pubkey,
+    authority_type: AuthorityType,
     owner_pubkey: &Pubkey,
     signer_pubkeys: &[&Pubkey],
 ) -> Result<Instruction, ProgramError> {
-    let data = TokenInstruction::SetOwner.pack()?;
+    let data = TokenInstruction::SetAuthority { authority_type }.pack()?;
 
     let mut accounts = Vec::with_capacity(3 + signer_pubkeys.len());
     accounts.push(AccountMeta::new(*owned_pubkey, false));
@@ -723,9 +754,11 @@ mod test {
         let unpacked = TokenInstruction::unpack(&expect).unwrap();
         assert_eq!(unpacked, check);
 
-        let check = TokenInstruction::SetOwner;
+        let check = TokenInstruction::SetAuthority {
+            authority_type: AuthorityType::Freezer,
+        };
         let packed = check.pack().unwrap();
-        let expect = Vec::from([6u8]);
+        let expect = Vec::from([6u8, 1]);
         assert_eq!(packed, expect);
         let unpacked = TokenInstruction::unpack(&expect).unwrap();
         assert_eq!(unpacked, check);
--- a/token/program2/src/instruction.rs
+++ b/token/program2/src/instruction.rs
@@ -1,6 +1,6 @@
 //! Instruction types
 
-use crate::error::TokenError;
+use crate::{error::TokenError, option::COption};
 use solana_sdk::{
     instruction::{AccountMeta, Instruction},
     program_error::ProgramError,
@@ -26,22 +26,18 @@ pub enum TokenInstruction {
     /// Accounts expected by this instruction:
     ///
     ///   0. `[writable]` The mint to initialize.
-    ///   1.
-    ///      * If supply is non-zero: `[writable]` The account to hold all the newly minted tokens.
-    ///      * If supply is zero: `[]` The owner/multisignature of the mint.
-    ///   2. `[]` (optional) The owner/multisignature of the mint, if supply is non-zero and:
-    ///                     (a) further minting is supported; or
-    ///                     (b) the mint freeze authority is enabled.
+    ///   1. If supply is non-zero: `[writable]` The account to hold all the newly minted tokens.
     ///
     InitializeMint {
         /// Initial amount of tokens to mint.
         amount: u64,
         /// Number of base 10 digits to the right of the decimal place.
         decimals: u8,
-        /// Whether further minting is supported
-        fixed_supply: bool,
-        /// Whether the mint freeze_authority should be enabled to freeze associated token accounts
-        can_freeze: bool,
+        /// The owner/multisignature of the mint if supply is non-zero. If present, further minting
+        /// is supported.
+        owner: COption<Pubkey>,
+        /// The freeze authority/multisignature of the mint.
+        freeze_authority: COption<Pubkey>,
     },
     /// Initializes a new account to hold tokens.  If this account is associated with the native mint
     /// then the token balance of the initialized account will be equal to the amount of SOL in the account.
@@ -237,14 +233,43 @@ impl TokenInstruction {
                 input_len += size_of::<u64>();
                 let decimals = unsafe { *(&input[input_len] as *const u8) };
                 input_len += size_of::<u8>();
-                let fixed_supply = unsafe { *(&input[input_len] as *const u8 as *const bool) };
-                input_len += size_of::<u8>();
-                let can_freeze = unsafe { *(&input[input_len] as *const u8 as *const bool) };
+
+                let owner = match input[input_len] {
+                    0 => {
+                        input_len += size_of::<u8>();
+                        COption::None
+                    }
+                    1 => {
+                        input_len += size_of::<u8>();
+                        #[allow(clippy::cast_ptr_alignment)]
+                        let owner = unsafe { *(&input[input_len] as *const u8 as *const Pubkey) };
+                        input_len += size_of::<Pubkey>();
+                        COption::Some(owner)
+                    }
+                    _ => {
+                        return Err(TokenError::InvalidInstruction.into());
+                    }
+                };
+
+                let freeze_authority = match input[input_len] {
+                    0 => COption::None,
+                    1 => {
+                        input_len += size_of::<u8>();
+                        #[allow(clippy::cast_ptr_alignment)]
+                        let freeze_authority =
+                            unsafe { *(&input[input_len] as *const u8 as *const Pubkey) };
+                        COption::Some(freeze_authority)
+                    }
+                    _ => {
+                        return Err(TokenError::InvalidInstruction.into());
+                    }
+                };
+
                 Self::InitializeMint {
+                    owner,
+                    freeze_authority,
                     amount,
                     decimals,
-                    fixed_supply,
-                    can_freeze,
                 }
             }
             1 => Self::InitializeAccount,
@@ -319,10 +344,10 @@ impl TokenInstruction {
         let mut output_len = 0;
         match self {
             Self::InitializeMint {
+                owner,
+                freeze_authority,
                 amount,
                 decimals,
-                fixed_supply,
-                can_freeze,
             } => {
                 output[output_len] = 0;
                 output_len += size_of::<u8>();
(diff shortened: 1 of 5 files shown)


CriesofCarrots:
@mvines @jackcmay , wdyt?

Also:
- Currently, it is the same instruction to freeze and account as it is to unfreeze: ToggleFreeze. We could split into two, or ask the user to pass a parameter to indicate which action is intended, if we want extra assurance that the account is intended to be frozen (or unfrozen). Thoughts?
- Currently, only the freeze_authority can sign to update the freeze_authority on the Mint. I considered giving this power to the Mint owner as well, but it seemed a little unnecessary to me. Does that sound okay?

token/program2/src/instruction.rs
```diff
@@ -187,6 +192,23 @@ pub enum TokenInstruction {
...
+    ///   0. `[writable]` The account to freeze.
+    ///   1. '[]' The token mint.
+    ///   2. `[signer]` The mint freeze authority.
+    ///
+    ///   * Multisignature owner
+    ///   0. `[writable]` The account to freeze.
+    ///   1. '[]' The token mint.
+    ///   2. `[]` The mint's multisignature freeze authority.
+    ///   3. ..3+M '[signer]' M signer accounts.
+    ToggleFreeze,
```
mvines, line 211: Making the operation more explicit, `Freeze(bool)`, would be better IMO.

token/program2/src/instruction.rs
```diff
@@ -187,6 +192,23 @@ pub enum TokenInstruction {
     ///   2. `[]` The account's multisignature owner.
     ///   3. ..3+M '[signer]' M signer accounts.
     CloseAccount,
+    /// Freeze an Initialized account or unfreeze a Frozen account, using the Mint's
+    /// freeze_authority (if set).
+    /// Native accounts cannot be frozen
```
mvines, line 197: I don't think this needs to be called out here.  The native mint will just not support freezing, so that documentation belongs with the native mint definition

token/program2/src/instruction.rs
```diff
@@ -37,6 +37,8 @@ pub enum TokenInstruction {
         amount: u64,
         /// Number of base 10 digits to the right of the decimal place.
         decimals: u8,
+        /// Whether the mint freeze_authority should be enabled to freeze associated token accounts
+        can_freeze: bool,
```
mvines, line 41: No freeze support for a token with fixed supply?

mvines:
> * Currently, it is the same instruction to freeze and account as it is to unfreeze: ToggleFreeze. We could split into two, or ask the user to pass a parameter to indicate which action is intended, if we want extra assurance that the account is intended to be frozen (or unfrozen). Thoughts?

I far prefer passing a parameter.

> * Currently, only the freeze_authority can sign to update the freeze_authority on the Mint. I considered giving this power to the Mint owner as well, but it seemed a little unnecessary to me. Does that sound okay?

Keeping the authorities completely separate seems ok to me.  Freeze is arguably more powerful than Mint, but I like the idea of not having to order them

token/program2/src/instruction.rs
CriesofCarrots: That's supported, just have to pass an authority as the option 2nd account. I obv need to update the docs there.

token/program2/src/processor.rs
```diff
@@ -51,17 +52,29 @@ impl Processor {
...
+                }
                 COption::None
             }
         } else if let Ok(owner_info) = next_account_info(account_info_iter) {
             COption::Some(*owner_info.key)
         } else {
             return Err(TokenError::OwnerRequiredIfNoInitialSupply.into());
         };
+        if can_freeze {
+            assert!(owner.is_some());
```
mvines, line 66:
```suggestion
            if owner.is_none() {
                return Err(TokenError::OwnerRequiredIfCanFreeze.into());
            }
```

token/program2/src/instruction.rs
mvines: ah, got it.   the `process_initialize_mint()` logic based on the presence of the optional owner is a little complex, more so now.   I think that's going to be my main argument for dropping the owner as an account and passing the Owner/Freeze pubkey as optional fields to the instruction data

CriesofCarrots: Okay, I pushed what this would properly look like -- we need to independently declare mint fixed supply and/or can_freeze. But I hate it. I'll rework back on https://github.com/solana-labs/solana-program-library/issues/229#issuecomment-677865126 in the morn.

CriesofCarrots:
Bump @jackcmay @mvines

mvines:
Thanks for the 🥾

token/program2/src/state.rs
```diff
@@ -36,16 +38,42 @@ pub struct Account {
     /// If `delegate` is `Some` then `delegated_amount` represents
     /// the amount authorized by the delegate
     pub delegate: COption<Pubkey>,
-    /// Is `true` if this structure has been initialized
-    pub is_initialized: bool,
+    /// Whether the account has been initialized or frozen
+    pub is_initialized: AccountState,
```
mvines, line 42: Perhaps?
```suggestion
    pub account_state: AccountState,
```

token/program2/src/instruction.rs
```diff
@@ -121,34 +122,37 @@ pub enum TokenInstruction {
...
-    ///   2. `[]` The mint's or account's multisignature owner.
+    ///   * Multisignature authority
+    ///   0. `[writable]` The mint or account to change the authority of.
+    ///   1. `[]` The new authority/multisignature.
+    ///   2. `[]` The mint's or account's multisignature authority.
     ///   3. ..3+M '[signer]' M signer accounts
-    SetOwner,
+    SetAuthority {
+        /// The type of authority to update.
+        authority_type: AuthorityType,
```
mvines, line 141: I don't like the inconsistently between `InitializeMint` and `SetAuthority`.  `InitializeMint` accepts a `COption<Pubkey>` in the instruction data for the authority, here it's passed as an account key.

I think `SetAuthority` should accept a `COption<Pubkey>`.  This allows us to very clearly remove the mint/freeze authority by providing `None`

token/program2/src/instruction.rs
```diff
@@ -26,17 +26,18 @@ pub enum TokenInstruction {
     /// Accounts expected by this instruction:
     ///
     ///   0. `[writable]` The mint to initialize.
-    ///   1.
-    ///      * If supply is non-zero: `[writable]` The account to hold all the newly minted tokens.
-    ///      * If supply is zero: `[]` The owner/multisignature of the mint.
-    ///   2. `[]` (optional) The owner/multisignature of the mint if supply is non-zero, if
-    ///                      present then further minting is supported.
+    ///   1. If supply is non-zero: `[writable]` The account to hold all the newly minted tokens.
```
mvines, line 29: wdyt about removing `amount: u64` and the optional key 1?   Now with `MintTo`, it's trivial to add a `MintTo` instruction immediately following `InitializeMint`.  And then for a fixed supply, the third instruction is a `SetAuthority(AuthorityType::MintTokens, None)`.   A little more instructions but much more clear and will reduce the complexity of `InitializeMint`

token/program2/src/instruction.rs
```diff
@@ -187,6 +191,26 @@ pub enum TokenInstruction {
...
+    ///   0. `[writable]` The account to freeze.
+    ///   1. '[]' The token mint.
+    ///   2. `[signer]` The mint freeze authority.
+    ///
+    ///   * Multisignature owner
+    ///   0. `[writable]` The account to freeze.
+    ///   1. '[]' The token mint.
+    ///   2. `[]` The mint's multisignature freeze authority.
+    ///   3. ..3+M '[signer]' M signer accounts.
+    FreezeAccount {
```
mvines, line 209: Not opposed to `FreezeAccount`/`ThawAccount` without the boolean.  ERC115 uses `Pause`/`Resume`.   I don't have strong opinions here though.  Just surfacing some other ideas

token/program2/src/instruction.rs
```diff
@@ -196,14 +220,58 @@ impl TokenInstruction {
         }
         Ok(match input[0] {
```
mvines, line 221: @jackcmay - all this `unsafe` is so icky.  There's really no other way?

token/program2/src/instruction.rs
```diff
@@ -196,14 +220,58 @@ impl TokenInstruction {
...
-                let amount = unsafe { *(&input[size_of::<u8>()] as *const u8 as *const u64) };
-                let decimals =
-                    unsafe { *(&input[size_of::<u8>() + size_of::<u64>()] as *const u8) };
-                Self::InitializeMint { amount, decimals }
+                let amount = unsafe { *(&input[input_len] as *const u8 as *const u64) };
+                input_len += size_of::<u64>();
+                let decimals = unsafe { *(&input[input_len] as *const u8) };
+                input_len += size_of::<u8>();
+
+                let mint_authority = match input[input_len] {
```
mvines, line 237: A single inline function for unpacking/packing a `COption<Pubkey>` would be nice.   This unpack/pack code is hard to read *and* unsafe.    The more we can reduce it the better

token/program2/src/instruction.rs
jackcmay: I went with the most efficient.  Even when I tried a hand written ser/de, the number of overhead went up considerably.  I think the right answer here is to jump into the compiler/linker/intrinsics and optimize that stuff.

token/program2/src/instruction.rs
jackcmay: I also like this idea, simple composable instructions.  The risks of not doing these operations atomically should be well documented so its clear they should be in the same tx.

jackcmay: Also, seems like this idea is a seperate effort for a different pr

token/program2/src/instruction.rs
```diff
@@ -121,34 +122,37 @@ pub enum TokenInstruction {
     ///   1. '[]' The source account's multisignature owner.
```
jackcmay, line 122: Shoudl `s/owner/authority/` ?

token/program2/src/instruction.rs
jackcmay: The more complex the token program gets the more unsavory this stuff gets

token/program2/src/instruction.rs
```diff
@@ -231,7 +299,18 @@ impl TokenInstruction {
                 Self::Approve { amount }
             }
             5 => Self::Revoke,
-            6 => Self::SetOwner,
+            6 => {
+                if input.len() < size_of::<u8>() + size_of::<u8>() {
+                    return Err(TokenError::InvalidInstruction.into());
+                }
+                let authority_type = match input[1] {
```
jackcmay, line 306: Can this be an into()?

token/program2/src/instruction.rs
```diff
@@ -249,6 +328,14 @@ impl TokenInstruction {
                 Self::Burn { amount }
             }
             9 => Self::CloseAccount,
+            10 => {
+                if input.len() < size_of::<u8>() + size_of::<u8>() {
+                    return Err(TokenError::InvalidInstruction.into());
+                }
+                #[allow(clippy::cast_ptr_alignment)]
+                let freeze = unsafe { *(&input[size_of::<u8>()] as *const u8 as *const bool) };
```
jackcmay, line 336: Seems like this could be simpler, its just a byte, read and convert safely?

token/program2/src/state.rs
jackcmay: And `s/Whether the account has been initialized or frozen/The account's state/` or something like that

token/program2/src/state.rs
```diff
@@ -36,16 +38,42 @@ pub struct Account {
...
     /// The amount delegated
     pub delegated_amount: u64,
 }
+impl Account {
+    /// Checks if account is frozen
+    pub fn is_frozen(&self) -> bool {
+        self.is_initialized == AccountState::Frozen
+    }
+}
 impl IsInitialized for Account {
```
jackcmay, line 54: Do we need this trait any longer?  Could be a member function of `Account` like `is_frozen`?

token/program2/src/processor.rs
```diff
@@ -233,41 +239,70 @@ impl Processor {
...
-                COption::Some(ref owner) => {
-                    Self::validate_owner(
-                        program_id,
-                        owner,
-                        authority_info,
-                        account_info_iter.as_slice(),
-                    )?;
+            match authority_type {
+                AuthorityType::MintTokens => {
+                    match mint.mint_authority {
```
jackcmay, line 278: Could this be an `ok_or`?

token/program2/src/processor.rs
```diff
@@ -391,14 +438,78 @@ impl Processor {
...
+        let mut source_data = source_account_info.data.borrow_mut();
+        let source_account: &mut Account = state::unpack(&mut source_data)?;
+
+        if source_account.is_native {
+            return Err(TokenError::NativeNotSupported.into());
+        }
+        if mint_info.key != &source_account.mint {
+            return Err(TokenError::MintMismatch.into());
+        }
+        if freeze && source_account.is_initialized != AccountState::Initialized
```
jackcmay, line 461: `!source_account.is_initialized()`?

token/program2/src/processor.rs
```diff
@@ -391,14 +438,78 @@ impl Processor {
...
+                    authority_info,
+                    account_info_iter.as_slice(),
+                )?;
+            }
+            COption::None => {
+                return Err(TokenError::MintCannotFreeze.into());
+            }
+        }
+
+        let new_state = if freeze {
```
jackcmay, line 484: nit?
```rust
source_accont.is_initialized = if freeze {
  AccountState::Frozen
} else {
  AccountState::Initialized
};
```

token/program2/src/instruction.rs
mvines: https://github.com/solana-labs/solana-program-library/issues/308

token/program2/src/instruction.rs
CriesofCarrots: 2 instructions does feel a little nicer to me, as there's a bit of a mind twist needed to get from Freeze(false) to Thaw. I'll work that up.

token/program2/src/instruction.rs
```diff
@@ -121,34 +122,37 @@ pub enum TokenInstruction {
...
     /// Mints new tokens to an account.  The native mint does not support minting.
     ///
     /// Accounts expected by this instruction:
     ///
-    ///   * Single owner
+    ///   * Single authority
     ///   0. `[writable]` The mint.
     ///   1. `[writable]` The account to mint tokens to.
-    ///   2. `[signer]` The mint's owner.
+    ///   2. `[signer]` The mint's mint-tokens authority.
```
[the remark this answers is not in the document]
CriesofCarrots: Ah, thank you, that's much better!

token/program2/src/processor.rs
CriesofCarrots: For ease of unpacking(checked), I have `Account::is_initialized()` return true if the account is Initialized or Frozen. So this needs to be more specific.
Open to other ideas, though.

token/program2/src/instruction.rs
CriesofCarrots: An account's authority is still called `owner`, so I decided to leave that term in the docs.

token/program2/src/processor.rs
jackcmay: Why not have `is_initialized() return true if not `AccountState::Initialized`?, I think I might be missing something.

CriesofCarrots: Because then calling `unpack()` on a frozen Account will fail with `TokenError::UninitializedState`. That doesn't feel right to me for frozen Accounts, as they aren't Uninitialized.

token/program2/src/instruction.rs
CriesofCarrots: Attempt: https://github.com/solana-labs/solana-program-library/pull/311

token/program2/src/processor.rs
CriesofCarrots: As discussed in DM, will update to `source_account.is_frozen()`

token/program2/src/instruction.rs
```diff
@@ -332,23 +463,59 @@ impl TokenInstruction {
...
     token_program_id: &Pubkey,
     mint_pubkey: &Pubkey,
     account_pubkey: Option<&Pubkey>,
     owner_pubkey: Option<&Pubkey>,
+    freeze_pubkey: Option<&Pubkey>,
     amount: u64,
     decimals: u8,
 ) -> Result<Instruction, ProgramError> {
-    let data = TokenInstruction::InitializeMint { amount, decimals }.pack()?;
+    let owner = if let Some(owner) = owner_pubkey {
```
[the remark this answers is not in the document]
CriesofCarrots: Done

token/program2/src/instruction.rs
CriesofCarrots: Done

token/program2/src/state.rs
CriesofCarrots: It's still used for Mint and Multisig unpack, so I'm leaving for now. We can consider removing in a separate pr.

CriesofCarrots:
@mvines @jackcmay got the COption clean up in and CI is happy. How is this looking?

token/program2/src/processor.rs
```diff
@@ -397,14 +447,74 @@ impl Processor {
         Ok(())
     }
 
+    /// Processes a [FreezeAccount](enum.TokenInstruction.html) or a
+    /// [ThawAccount](enum.TokenInstruction.html) instruction.
+    pub fn process_toggle_freeze_account(
+        program_id: &Pubkey,
+        accounts: &[AccountInfo],
+        desired_account_state: AccountState,
```
mvines, line 455: I think I'd prefer this to be a boolean: freeze/unfreeze, to dumb down the implementation.   `AccountState`  as an input technically allows the caller to pass in `Uninitialized` and we don't handle that (not do we need to)

CriesofCarrots: I had it as a boolean at first, but I didn't like the ambiguity of `false`.
So I replaced it with an enum (something like ToogleAction::Freeze / ToggleAction::Unfreeze), which then seemed superfluous, since the AccountState enum already essentially does this.

But, meh. Pushed the boolean.

PR #306, opened 2020-08-24
Correctly fail CI on JS failures

mvines:


mvines:
Caught again by my inexplicable love for `set -e`.   As my punishment, I need to read 5 five times: http://mywiki.wooledge.org/BashFAQ/105

mvines:
Yay, it fails now.  Landing because this is the correct behaviour and CI *should* be red right now

PR #311, opened 2020-08-25
Add COption helpers

### LINKED ISSUE #300: No conversion between `COption` and `Option`

CriesofCarrots:
No way to covert between rust Option and COption.
Also, packing of COptions is hard to read in the context of token Instruction::pack and Instruction::unpack, so separate into helpers.

Fixes #300 
Also: https://github.com/solana-labs/solana-program-library/pull/297#discussion_r475993617


jackcmay:
Woot!

PR #327, opened 2020-08-26
token: Simplify initialize_mint

### LINKED ISSUE #308: token: Simplify InitializeMint instruction
1. Remove `amount: u64` argument.  Instead the user can issue a `MintTo` instruction immediately after `InitializeMint`.  This allows the removal of the first account key
2. Remove the second account key.  The supply can be fixed by issuing a `SetAuthority(AuthorityType::MintTokens, None)` instruction after `MintTo`

These changes have no affect on expressibility and will reduce a number of branches from `InitializeMint`

CriesofCarrots:
InitializeMint is more complicated than it needs to be; all the same functionality is available by chaining other methods.

Namely, an InitializeMint instruction that mints a fixed supply to an account becomes:
- InitializeMint with mint_authority
- Use mint_authority to MintTo account
- Use mint_authority to set mint authority to None

Fixes #308

### DIFF
--- a/token/program2/inc/token2.h
+++ b/token/program2/inc/token2.h
@@ -118,7 +118,6 @@ typedef enum Token_TokenInstruction_Tag {
      * Accounts expected by this instruction:
      *
      *   0. `[writable]` The mint to initialize.
-     *   1. If supply is non-zero: `[writable]` The account to hold all the newly minted tokens.
      *
      */
     Token_TokenInstruction_InitializeMint,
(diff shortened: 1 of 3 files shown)


CriesofCarrots:
Having to regenerate the C headers on every change in order to get past CI is killing me.

mvines:
> Having to regenerate the C headers on every change in order to get past CI is killing me.

Yeah it's bugged me too.

Issue #333, opened 2020-08-27
token: InitializeMint's mint_authority should not be optional


mvines:
Now that it's no longer possible to mint tokens with `InitializeMint`, it doesn't make sense that the mint_authority be optional:
https://github.com/solana-labs/solana-program-library/blob/ea524284757c859c7c6be1af8fa038bb278ef193/token/program2/src/instruction.rs#L30-L37
What's the point of creating a mint account that can't mint tokens.

Issue #315, opened 2020-08-25
token: Reconsider unsafe usage for instruction pack/unpack


mvines:
The use of unsafe typecasting in instruction pack/unpack significantly reduces the instruction count over alternatives but is also scary to look at and is more likely it contain bugs (or have bugs accidentally introduced).  Further this is not a general pattern we want to see in all Solana contracts.

Consider alternatives that either make the unsafe go away entire, or pushes the unsafe into macros or generic libraries where the possibility of human error is reduced.

mvines:
https://crates.io/crates/byteorder has been used by some

mvines:
Also @jackcmay raised the point that since the instruction data is now coming in aligned in the new BPF loader that using byteorder may just be as quick as unsafe casts/copies

jackcmay:
Byte order will not be as quick and most likely significantly slower due to copies both into and back out to account data vs cast and use in place

mvines:
Fixed by https://github.com/solana-labs/solana-program-library/pull/349

PR #335, opened 2020-08-27
token: InitializeMint's mint authority is no longer optional

mvines:
Fixes https://github.com/solana-labs/solana-program-library/issues/333

### DIFF
--- a/token/program/src/instruction.rs
+++ b/token/program/src/instruction.rs
@@ -30,8 +30,8 @@ pub enum TokenInstruction {
     InitializeMint {
         /// Number of base 10 digits to the right of the decimal place.
         decimals: u8,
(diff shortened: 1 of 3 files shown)


token/program/src/instruction.rs
```diff
@@ -226,7 +226,7 @@ impl TokenInstruction {
         }
         Ok(match input[0] {
             0 => {
-                if input.len() < size_of::<u8>() + size_of::<u8>() + size_of::<bool>() {
+                if input.len() < size_of::<u8>() + size_of::<Pubkey>() + size_of::<bool>() {
```
mvines, line 229: This math seems sketchy.  But I figure it'll get cleaned up by @jackcmay as he works through https://github.com/solana-labs/solana-program-library/issues/315

Issue #328, opened 2020-08-26
token: require all token accounts be rent-exempt and restore total supply field in the mint


mvines:
We lost the total supply field in the mint when we realized that if token accounts aren't created rent-exempt, they could just disappear and cause the total supply to be inflated.

However we've failed to discover a compelling reason to permit non-rent-exempt token accounts.  Additionally with the new `CloseAccount` instruction, you can reclaim your SOL from a token account at any time.

Requiring rent exemption also simplifies wallets and other token clients, as they no longer need to be considered about a token account disappearing because it was underfunded, either by accident or maliciously.

CriesofCarrots:
Mints too

mvines:
and multisigs!

CriesofCarrots:
One effect of this change is that wrapped SOL accounts cannot hold less wrapped SOL than the account rent-exempt minimum balance. Any concerns about this?

mvines:
Seems fine to me.  Technically not necessary but the consistency is nice.  Here's a fun edge case:
1. Wrap the rent-exempt amount of SOL
2. Transfer 1 lamport to another token account
3. Now the source account is no longer rent-exempt.

mvines:
I think we need a `rent_exempt_reserve` field, like Stake accounts:
https://github.com/solana-labs/solana/blob/57174cdabed9821b529ed359eab759f7b28f52c4/programs/stake/src/stake_state.rs#L130

CriesofCarrots:
> Technically not necessary but the consistency is nice

Allowing non-rent-exempt wrapped SOL accounts requires a bit of special casing, since rent debits will not be reflected in the account's `amount` field. So yay, glad that seems fine to you.

> I  think we need a `rent_exempt_reserve` field, like Stake accounts:

Agreed, we'll need that for wrapped SOL, so `Account` structs. I think we can leave that out of `Mint` and `Multisig`, since there's no way to withdraw from those accounts.

Issue #303, opened 2020-08-24
Tokens can be initialized with invalid mint accounts


jstarry:
#### Problem
`InitializeAccount` doesn't check if `mint_info` is valid (exists, initialized, and owned by token program) and so tokens can be created with invalid mints. These token accounts are useless because without a valid mint, they can never hold a token balance.

#### Suggested Changes
- For developer experience, we should fail fast on token creation to avoid later confusion when token accounts are attempted to be minted to.
- Consider forcing mint accounts rent exempt so that token accounts are always valid

jstarry:
Here is an example on-chain where this happened on testnet
InitializeAccount: https://explorer.solana.com/tx/5Ep4euatukeYjqFSwaPfrNT8FTfkgeYkp5boCEBPNygb9kACU11Nn7Kpzrjru3VxyFWSqMyv5jYfim6Wp9VVYceT?cluster=testnet
Invalid mint account: https://explorer.solana.com/address/3ZE2ei22iLgivkeLZkmNVcTx3zPfkN7ALLx5MdeEJ2ss?cluster=testnet

jstarry:
@jackcmay @CriesofCarrots @mvines thoughts?

mvines:
> These token accounts are useless because without a valid mint, they can never hold a token balance.

The account can still be closed though.  I feel like the client-side libraries could mostly mitigate this.   I'm not opposed to adding a mint validity check in InitializeAccount.

> Consider forcing mint accounts rent exempt so that token accounts are always valid

This should be client side IMO.   I don't feel like we need to enforce that in the program.

mvines:
@CriesofCarrots - anything left to do here?

CriesofCarrots:
Yes, still need the check for valid mint on InitializeAccount

PR #336, opened 2020-08-27
token: Require that Account, Mint, and Multisig be rent exempt

CriesofCarrots:
Toward #328 
Separately: restore total_supply to Mint, and adjust on burns

Toward #303 
Separately: require valid Mint exist on InitializeAccount

### DIFF
--- a/token/program/src/processor.rs
+++ b/token/program/src/processor.rs
@@ -78,11 +78,11 @@ impl Processor {
         account.delegated_amount = 0;
         account.state = AccountState::Initialized;
         if *mint_info.key == crate::native_mint::id() {
-            account.is_native = true;
-            account.amount = new_account_info.lamports();
-            account.rent_exempt_reserve = rent.minimum_balance(new_account_info_data_len);
+            let rent_exempt_reserve = rent.minimum_balance(new_account_info_data_len);
+            account.is_native = COption::Some(rent_exempt_reserve);
+            account.amount = new_account_info.lamports() - rent_exempt_reserve;
         } else {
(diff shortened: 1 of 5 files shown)


token/program/src/state.rs
```diff
@@ -46,6 +46,9 @@ pub struct Account {
     pub delegated_amount: u64,
     /// Optional authority to close the account.
     pub close_authority: COption<Pubkey>,
+    /// An Account is required to be rent-exempt. This value logs the reserve required to be
+    /// rent-exempt so that wrapped SOL accounts do not drop below this threshold.
+    pub rent_exempt_reserve: u64,
```
mvines, line 51: Dare we combine this with `is_native`?

`pub is_native: COption<u64>`?  Where the held u64 is the reserve?

CriesofCarrots: Ooh, I like it. Same # of bytes, but makes it more explicit that the reserve only affects wrapped SOL accounts

token/program/src/processor.rs
```diff
@@ -67,6 +80,7 @@ impl Processor {
         if *mint_info.key == crate::native_mint::id() {
             account.is_native = true;
             account.amount = new_account_info.lamports();
```
mvines, line 82: I think we want to exclude `rent_exempt_reserve` from the account amount.  Generically a client is going to look at the `amount` field to determine the max amount that can be transferred.  Since the `rent_exempt_reserve` can't be transferred, if not then the client is going to need special casing for wrapped SOL

PR #342, opened 2020-08-27
Reinstate Mint supply

### LINKED ISSUE #328: token: require all token accounts be rent-exempt and restore total supply field in the mint

CriesofCarrots:
Reinstate the `supply` field to the `Mint` state, and adjust on MintTo and Burn instructions. Reinstate passing the mint account into Burn instructions.

Also cleaned up a couple account ordered lists in docs

Fixes #328

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -20,7 +20,7 @@ use spl_token::{
     self,
     instruction::*,
     native_mint,
-    state::{Account, Mint},
+    state::{self, Account, Mint},
 };
 use std::{mem::size_of, process::exit};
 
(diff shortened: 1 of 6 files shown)


token/program/src/state.rs
```diff
@@ -12,6 +12,8 @@ pub struct Mint {
     /// mint creation. If no mint authority is present then the mint has a fixed supply and no
     /// further tokens may be minted.
     pub mint_authority: COption<Pubkey>,
+    /// Total supply of tokens.
+    pub supply: u64,
```
mvines, line 16: So happy to see this back

token/program/src/processor.rs
```diff
@@ -439,6 +451,7 @@ impl Processor {
         }
 
         source_account.amount -= amount;
+        mint.supply -= amount;
```
mvines, line 454: In a follow-up PR, can you please replace all the `-=` with `checked_sub()`.  Let's just use checked math everywhere in the program, even when it's obviously not necessary

mvines (dismissed):

token/cli/src/main.rs
```diff
@@ -213,12 +213,20 @@ fn command_burn(config: &Config, source: Pubkey, ui_amount: f64) -> CommmandResu
         .rpc_client
         .get_token_account_balance_with_commitment(&source, config.commitment_config)?
         .value;
+    let source_account = config
```
mvines, line 216: `source_token_balance` is only used to figure decimals, which we can get from the mint.  So I think we can remove the `get_token_account_balance_with_commitment()` RPC call entirely

CriesofCarrots: Still two calls, since we'll have to pull the mint.

mvines: ah, yeah ok.  wfm

PR #347, opened 2020-08-27
token: Check mint on InitializeAccount

### LINKED ISSUE #303: Tokens can be initialized with invalid mint accounts

CriesofCarrots:
Tokens can be created with invalid mints. These token accounts are useless because without a valid mint, they can never hold a token balance.

- Add check for Mint validity in process_initialize_account

Fixes #303

### DIFF
--- a/token/program/inc/token.h
+++ b/token/program/inc/token.h
@@ -123,8 +123,10 @@ typedef enum Token_TokenInstruction_Tag {
      */
     Token_TokenInstruction_InitializeMint,
     /**
-     * Initializes a new account to hold tokens.  If this account is associated with the native mint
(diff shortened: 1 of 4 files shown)


mvines:
landing so I can rebase on it

PR #349, opened 2020-08-28
Safer pack/unpack

jackcmay:
Get rid of the scary-looking unsafe code.  This new code at least triples the number of instructions required but will make auditors a lot more happy

### DIFF
--- a/token/program/src/pack.rs
+++ b/token/program/src/pack.rs
@@ -1,0 +1,79 @@
+//! State transition types
+
+use crate::error::TokenError;
+use solana_sdk::program_error::ProgramError;
+
+/// Check is a token state is initialized
+pub trait IsInitialized {
+    /// Is initialized
+    fn is_initialized(&self) -> bool;
+}
+
+/// Depends on Sized
+pub trait Sealed: Sized {}
+
+/// Safely and efficiently (de)serialize account state
+pub trait Pack: Sealed {
+    /// The length, in bytes, of the packed representation
+    const LEN: usize;
+    #[doc(hidden)]
+    fn pack_into_slice(&self, dst: &mut [u8]);
+    #[doc(hidden)]
+    fn unpack_from_slice(src: &[u8]) -> Result<Self, ProgramError>;
+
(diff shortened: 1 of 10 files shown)


CriesofCarrots:
Remove these? https://github.com/solana-labs/solana-program-library/blob/74bf8a1f587692a9d14e2afaad2a455de475e308/token/program/src/option.rs#L681

jackcmay:
> Remove these?
> 
> https://github.com/solana-labs/solana-program-library/blob/74bf8a1f587692a9d14e2afaad2a455de475e308/token/program/src/option.rs#L681

They are still being used, why remove?

CriesofCarrots:
> They are still being used, why remove?

Where are `COption::pack` and `COption::unpack_or` still being used, aside from the unit test? This pr removes the usages I'm aware of.

token/program/src/pack.rs
```diff
@@ -0,0 +1,79 @@
...
+    fn pack_into_slice(&self, dst: &mut [u8]);
+    #[doc(hidden)]
+    fn unpack_from_slice(src: &[u8]) -> Result<Self, ProgramError>;
+
+    /// Borrow `Self` from `input` for the duration of the call to `f`, but first check that `Self`
+    /// is initialized
+    #[inline(never)]
+    fn unpack_mut<U>(
+        input: &mut [u8],
+        f: &mut dyn FnMut(&mut Self) -> Result<U, ProgramError>,
```
garious, line 29: Why the trait object here and not `&mut F where F: FnMut(&mut Self) -> Result<U, ProgramError>`?

garious:
A shame this doesn't compile down to something just as fast. A lot of that code looks equivalent to me. Oh well - definitely easier to audit this way.

token/program/src/pack.rs
jackcmay: That'll work too, updated

jackcmay:
@garious Part of it I'm sure is the copy into and out of that's happening in these changes that was not happening before.

PR #350, opened 2020-08-28
Move c header generation out of build.rs

mvines:
Drop the cbindgen integration into build.rs, it's just too hard to make work right.  Instead just run `./cbindgen.sh` from the root to update all the C headers


mvines:
fyi @CriesofCarrots / @jackcmay.   Let's try it this way for a while

mvines:
Something is still funky here, I think cbindgen is pulling spl-token from crates.io.  But IMO this is an improvement regardless so landing it

PR #353, opened 2020-08-28
Token nits

garious:
* Update native mint public key in the docs
* Fix `revoke` instruction constructor; account state was readonly, but needs to be writable
* Allow more parallelization in multisig transactions; signers weren't readonly
* Fix the AlreadyInUse error display message


CriesofCarrots:
There's a stuck 2nd CI job, for some reason, which is preventing the automerge. Merging since the other CI job passed.

PR #355, opened 2020-08-28
Don't repack if not needed

jackcmay:
There are cases where the token program only needed to unpack to read data but is instead also repacking that unchanged data

- Don't repack if we don't need to
- Ensure that pack function check initialization
- Cleanup the pack trait by encapsulating stray functions
- Fix some token-swap rot

### DIFF
--- a/token-swap/program/src/processor.rs
+++ b/token-swap/program/src/processor.rs
@@ -69,19 +69,13 @@ impl State {
     pub fn token_account_deserialize(
         info: &AccountInfo,
     ) -> Result<spl_token::state::Account, Error> {
-        spl_token::state::Account::unpack_from_slice(&info.data.borrow_mut())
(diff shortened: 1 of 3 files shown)


CriesofCarrots:
Ugh, cbindgen. Fyi @mvines I was seeing this locally running `cbindgen.sh` as well. Seems to only be from the token_swap build, afaict. When I commented out https://github.com/solana-labs/solana-program-library/blob/6761f440ab98159f53bbda9658edc685bc11aceb/utils/cgen/src/main.rs#L66 I got the token.h I wanted.

mvines:
There's some really weird bugs in cbindgen :(

CriesofCarrots:
Retry ftw

jackcmay:
Restarted the job and its past cbindgen now

PR #358, opened 2020-08-28
Update token js bindings

jstarry:
#### Changes
- Update mint / account state and layout
- Rename setOwner to setAuthority
- Update mint creation to no longer require initial account and supply
- Add static methods for initializing mints and accounts
- Fix bug where `Token` methods called static methods with `PublicKey` when `Account` was expected (for non multisig accounts)


mvines:
CI was being silly so I landed manually

PR #373, opened 2020-08-31
Fix token-swap tests

joncinque:
**Problem**

Token-swap tests currently do not pass, even with the linking
issue resolved on `create_program_address`, failing with key errors
since there are incorrect hard-coded program ids.

**Solution**

To fix this, some more conditional compilation is done to create valid
program ids for token and token-swap, followed by properly adding
the program as a signer on its addresses, mimicking what happens in the
real runtime's syscalls.

Note: In order to get valid `Keypair`s as they are used by the BPF
loader, the `program` feature is removed from the test build, since it
requires `sdk::signature::Keypair`, which is not allowed for
`feature = program`.  I considered adding the ed25519-dalek library as
a dev dependency, but it seemed better to use the same code as the sdk.
Let me know if there's a better way to generate these while also
maintaining `feature = program`.

### DIFF
--- a/token-swap/program/src/processor.rs
+++ b/token-swap/program/src/processor.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use num_traits::FromPrimitive;
 #[cfg(not(target_arch = "bpf"))]
-use solana_sdk::instruction::Instruction;
+use solana_sdk::instruction::AccountMeta;
 #[cfg(target_arch = "bpf")]
 use solana_sdk::program::invoke_signed;
 use solana_sdk::{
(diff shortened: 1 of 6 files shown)


token-swap/program/src/processor.rs
```diff
@@ -103,11 +107,26 @@ impl State {
             &[],
             amount,
         )?;
-        invoke_signed(&ix, accounts, signers)
+        #[cfg(target_arch = "bpf")]
+        {
+            invoke_signed(&ix, accounts, signers)
+        }
+        #[cfg(not(target_arch = "bpf"))]
```
jackcmay, line 114: Be nice to keep these featurizations out of the main code, can these be in an invoke_signed stub instead?

joncinque:
Gonna close this, I came up with a better way to avoid this issue by generating better pubkeys on the testing side.

PR #379, opened 2020-09-01
Fix up spl-token docs

### LINKED ISSUE #377: token2: document that the supply for the native_mint is always 0
Unlike all other token mints, the supply for the native mint is always 0.

This limitation is intended to be fixed for token3.  See https://github.com/solana-labs/solana-program-library/issues/374

CriesofCarrots:
Document that the supply for the native mint is always 0.
Also a couple small fixups

Fixes #377

### DIFF
--- a/docs/src/token.md
+++ b/docs/src/token.md
@@ -254,8 +254,8 @@ These accounts have a few unique behaviors
 - Burning is not supported
 - When closing an Account the balance may be non-zero.
 
-In spl-token v2.0, the Native Mint supply will always report 0, regardless of
-how much SOL is currently wrapped. This is expect to be fixed in spl-token v3.
+In spl-token v2, the Native Mint supply will always report 0, regardless of
+how much SOL is currently wrapped.
 
 ### Rent-exemption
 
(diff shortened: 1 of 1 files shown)


docs/src/token.md
```diff
@@ -254,6 +254,9 @@ These accounts have a few unique behaviors
 - Burning is not supported
 - When closing an Account the balance may be non-zero.
 
+In spl-token v2.0, the Native Mint supply will always report 0, regardless of
```
mvines, line 257:
```suggestion
In spl-token v2, the Native Mint supply will always report 0, regardless of
```

docs/src/token.md
```diff
@@ -254,6 +254,9 @@ These accounts have a few unique behaviors
...
+how much SOL is currently wrapped. This is expect to be fixed in spl-token v3.
```
mvines, line 258:
```suggestion
how much SOL is currently wrapped.
```
Extraneous IMO.  We can point folks at the issue if they raise this, for now

mvines (dismissed):

PR #380, opened 2020-09-01
Add back token perf monitor

jackcmay:
Token performance may change unexpectedly over time

Adding a performance monitor test that performs some basic operations and fails if the performance degrades significantly.  The performance monitor is run during CI and should help us keep tabs on the token program's performance.

The performance benchmarks are based on current performance.  It's sad to see the number move from where they were to the new numbers due to the serialization:
```
    InitializeMint 1340 => 4000
    InitializeAccount 1317 => 6445
    Tranfer 2161 => 7609
```


mvines:
Cool!

PR #389, opened 2020-09-03
Update logos

rmshea:


CriesofCarrots (approved):
Doesn't matter for this pr, but just fyi: the CI failure would be fixed if you rebased on master.

PR #399, opened 2020-09-08
Fix deposit instruction

joncinque:
* Refactor tests for easier composibility
* Add deposit instruction and test
* Deposit JS test works again


CriesofCarrots (dismissed):
r+ pending CI.
This is looking nice!
At some point soon-ish we should extend the processor tests to cover the various program-error cases; will probably involve manhandling the accounts in SwapAccountInfo ;)

joncinque:
absolutely, I'm building up a list of tests to add, using what's present currently in `token` as inspiration

PR #405, opened 2020-09-09
token-swap: Fix withdraw instruction

joncinque:
* Include pool mint in instruction, required for burning
* Add simple Rust test
* Fix js withdraw test


CriesofCarrots (approved):
lgtm!

PR #408, opened 2020-09-09
token-swap: Swap instruction fix and update libs for CI

joncinque:
This is two separate problems to some extent, but it all came up while trying to run the JS test suite on token-swap.  Let me know if you prefer to split this up.

* Fix the swap instruction and add a test for A -> B and B -> A swapping
* Update JS and Rust libs to new versions, and correctly use the new version of the web3 SDK, using Jack's PR on hello world for inspiration: https://github.com/solana-labs/example-helloworld/commit/0510fcc777a3a4cbbd37e54d09aa806feb128457
* Re-enable token-swap in CI

### DIFF
--- a/token-swap/program/src/instruction.rs
+++ b/token-swap/program/src/instruction.rs
@@ -53,10 +53,10 @@ pub enum SwapInstruction {
     ///   1. `[]` $authority
     ///   2. `[writable]` token_(A|B) SOURCE Account, amount is transferable by $authority,
     ///   3. `[writable]` token_(A|B) Base Account to swap INTO.  Must be the SOURCE token.
-    ///   4. `[writable]` token_(A|B) Base Account to swap FROM.  Must be the OUTPUT token.
-    ///   5. `[writable]` token_(A|B) OUTPUT Account assigned to USER as the owner.
+    ///   4. `[writable]` token_(A|B) Base Account to swap FROM.  Must be the DESTINATION token.
+    ///   5. `[writable]` token_(A|B) DESTINATION Account assigned to USER as the owner.
     ///   6. '[]` Token program id
-    ///   userdata: SOURCE amount to transfer, output to OUTPUT is based on the exchange rate
+    ///   userdata: SOURCE amount to transfer, output to DESTINATION is based on the exchange rate
     Swap(u64),
 
     ///   Deposit some tokens into the pool.  The output is a "pool" token representing ownership
@@ -160,7 +160,7 @@ pub fn initialize(
     token_a_pubkey: &Pubkey,
     token_b_pubkey: &Pubkey,
     pool_pubkey: &Pubkey,
-    user_output_pubkey: &Pubkey,
+    destination_pubkey: &Pubkey,
     nonce: u8,
     fee: Fee,
 ) -> Result<Instruction, ProgramError> {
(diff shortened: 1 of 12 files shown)


joncinque:
I got the "abuse detected" problem again 😢

CriesofCarrots:
And it didn't pick up CI from my close/reopen this time 😢

mvines:
I emailed Travis CI support again 👎

joncinque:
They got back to me really quickly, so it could be worse.  Thanks for the help @mvines !

mvines:
Great, thanks.  We sadly just need to continue to bug Travis CI until they fix it.  It's been months now ... :(

joncinque:
There's some weirdness with the token and token-swap containers in the testing environment.  For some reason, `npm run localnet:down` doesn't seem to pick up the running solana-localnet container.
```
+docker run --detach --name solana-localnet --rm --publish 8899:8899 --publish 8900:8900 --publish 9900:9900 --tty --ulimit nofile=500000 --env RUST_LOG=solana=info solanalabs/solana:v1.3.9

697b5039acaa47781294118d908391ac6ad92cf14afb5997bb331b41cc29ce3f
```
And later:
```
+npm run localnet:down
> @solana/spl-token@0.0.7 localnet:down /home/travis/build/solana-labs/solana-program-library/token/js
> solana-localnet down
Docker version 18.06.0-ce, build 0ffa825
++docker ps --filter 'name=^solana-localnet$' -q
+[[ -n '' ]]
```
Normally, the id of the container will show up in that last line, but it isn't here.  Because of that, when trying to run for the token-swap tests, we get an error.  I'll disable the token-swap js tests again so we can get this through, and investigate separately.

token-swap/program/src/instruction.rs
```diff
@@ -52,11 +52,11 @@ pub enum SwapInstruction {
...
-    ///   4. `[writable]` token_(A|B) Base Account to swap INTO.  Must be the SOURCE token.
-    ///   5. `[writable]` token_(A|B) Base Account to swap FROM.  Must be the DEST token.
-    ///   6. `[writable]` token_(A|B) DEST Account assigned to USER as the owner.
-    ///   7. '[]` Token program id
-    ///   userdata: SOURCE amount to transfer, output to DEST is based on the exchange rate
+    ///   3. `[writable]` token_(A|B) Base Account to swap INTO.  Must be the SOURCE token.
+    ///   4. `[writable]` token_(A|B) Base Account to swap FROM.  Must be the OUTPUT token.
+    ///   5. `[writable]` token_(A|B) OUTPUT Account assigned to USER as the owner.
+    ///   6. '[]` Token program id
+    ///   userdata: SOURCE amount to transfer, output to OUTPUT is based on the exchange rate
```
CriesofCarrots, line 59: What motivated the change to `OUTPUT` here?
If we want to go this route, let's use output consistently. As it is, the js still uses `destination` and the rust `swap()` instruction method uses `dest` (which we should probably spell out, even though we do the same thing in spl-token)

CriesofCarrots (commented):
Just one verbiage/consistency nit. Thanks for rolling in the upgrade to solana-web3.js v0.75.0! That was on my list for today.

token-swap/program/src/instruction.rs
joncinque: That's a good point.  I was mostly getting confused with the `into` and `from` accounts, and in the process I made those changes as well to clarify which was which.  I'll revert `output` to `destination` and expand where we only have `dest`.  From there, you'll be able to enjoy 0.75.0 :smile:

CriesofCarrots (approved):

PR #411, opened 2020-09-10
Transfer to account who owns self

jackcmay:
In the transfer instruction if the source account owns itself the instruction panics trying to double mutably borrow

Add test and re-arrange to avoid duplicate borrow

These changes are incomplete, more duplicate key tests should be added and more instruction will probably have to be updated

### DIFF
--- a/token/program/src/processor.rs
+++ b/token/program/src/processor.rs
@@ -161,82 +161,82 @@ impl Processor {
             return Ok(());
         }
 
-        let mut source_data = source_account_info.data.borrow_mut();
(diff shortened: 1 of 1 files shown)


mvines:
We should consider token/program immutable for functional changes like this.  It's time to bring token/program2 back I think, as a place to land fixes and new features for the next token program

jackcmay:
@mvines Yup, wasn't planning to commit this yet, just highlighting how we can test for and fix this kind of issue

jackcmay:
Addressed in #433

PR #412, opened 2020-09-10
Add token/program2 for next changeset

CriesofCarrots:


mvines (commented):
Thanks.  

Should we call this `token/program3` to align with the spl-token 3.x versioning

CriesofCarrots:
> Thanks.
> 
> Should we call this `token/program3` to align with the spl-token 3.x versioning

Something like that. I was toying with program-v3/spl-token-v3 to move away from the index feel. Does that work for you?

mvines:
Sure, so `token/program-v3`?

PR #416, opened 2020-09-10
Wrapped SOL support

dankelleher:
Add support for wrapped native token account generation in the Token JS client.

Usage:

```
const native = await Token.createWrappedNativeAccount(
    connection,
    programId,
    owner.publicKey,
    payer,
    lamportsToWrap,
  );
const accountInfo = await token.getAccountInfo(native);
assert(accountInfo.isNative);

// check that the new account has wrapped native tokens.
assert(accountInfo.amount.toNumber() === lamportsToWrap);
```

Token.createAccount function performs the systemProgram account creation and token account initialisation steps in one transaction, without first passing lamports to the created account. So the balance is 0 (rent notwithstanding) when initializeAccount is called. This means that, in the native token case, the SOL cannot be wrapped

This function allows you to transfer lamports on creation. The alternative is to create the account, fund it, and then call initialise in separate transactions, or not use the Token JS client at all.

Why is this function static? It can only be run on the "special case" native token mint, not on any token mint, like createAccount can. 

This PR also exposes the NATIVE_MINT constant, referring to this special case mint. 

There is a little duplicated code here between createAccount and createWrappedNativeAccount. I'd be open to refactoring this if this is seen as a problem.

### DIFF
--- a/token/js/cli/token-test.js
+++ b/token/js/cli/token-test.js
@@ -4,7 +4,7 @@ import fs from 'mz/fs';
 import {Account, Connection, BpfLoader, PublicKey} from '@solana/web3.js';
 import semver from 'semver';
 
-import {Token} from '../client/token';
+import {Token, NATIVE_MINT} from '../client/token';
 import {url} from '../url';
 import {newAccountWithLamports} from '../client/util/new-account-with-lamports';
 import {sleep} from '../client/util/sleep';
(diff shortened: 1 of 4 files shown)


token/js/cli/token-test.js
```diff
@@ -417,18 +417,29 @@ export async function multisig(): Promise<void> {
 export async function nativeToken(): Promise<void> {
   const connection = await getConnection();
 
-  const mintPublicKey = new PublicKey(
-    'So11111111111111111111111111111111111111112',
-  );
+  const mintPublicKey = NATIVE_MINT;
```
mvines, line 420: Would you mind just removing `mintPublicKey` entirely?  We can use `NATIVE_MINT` instead

token/js/client/token.js
```diff
@@ -427,6 +432,76 @@ export class Token {
...
+   * This function sends lamports to the new account before initializing it.
+   *
+   * @param connection A solana web3 connection
+   * @param programId The token program ID
+   * @param owner The owner of the new token account
+   * @param payer The source of the lamports to initialize, and payer of the initialization fees.
+   * @param amount The amount of lamports to wrap
+   * @return {Promise<PublicKey>} The new token account
+   */
+  static async createWrappedNativeAccount(
```
mvines, line 450: I'm curious why you went the `static` route instead of adding a method?

mvines (commented):
Thanks for doing this @dankelleher!

token/js/client/token.js
dankelleher: It can only be run on the "special case" native token mint, not on any token mint, like createAccount can.

To elaborate, it would be weird to do something like.
```
const srmToken = new Token(connection, serumPubkey, programId, payer)
srmToken.createWrappedNativeAcccount(owner, amount)
```

CriesofCarrots:
Thanks @dankelleher !

PR #420, opened 2020-09-10
Pack instructions and state more efficiently

joncinque:
* SwapInfo and SwapInstruction are more memory-efficient, following the
  example from SPL Token
* Remove unnecessary helper structs to make instructions clearer
* Error -> SwapError
* Update JS tests to reflect the change

### DIFF
--- a/token-swap/js/client/token-swap.js
+++ b/token-swap/js/client/token-swap.js
@@ -97,7 +97,7 @@ type TokenSwapInfo = {|
  * @private
  */
 const TokenSwapLayout = BufferLayout.struct([
-  BufferLayout.u8('is_initialized'),
+  BufferLayout.u8('isInitialized'),
   BufferLayout.u8('nonce'),
   Layout.publicKey('tokenAccountA'),
   Layout.publicKey('tokenAccountB'),
@@ -270,7 +270,7 @@ export class TokenSwap {
 
     const data = Buffer.from(accountInfo.data);
     const tokenSwapInfo = TokenSwapLayout.decode(data);
-    if (!tokenSwapInfo.is_initialized) {
+    if (!tokenSwapInfo.isInitialized) {
       throw new Error(`Invalid token swap state`);
     }
     // already properly filled in
--- a/token-swap/js/client/token-swap.js
+++ b/token-swap/js/client/token-swap.js
@@ -97,15 +97,13 @@ type TokenSwapInfo = {|
  * @private
  */
 const TokenSwapLayout = BufferLayout.struct([
-  BufferLayout.u8('state'),
+  BufferLayout.u8('is_initialized'),
   BufferLayout.u8('nonce'),
   Layout.publicKey('tokenAccountA'),
   Layout.publicKey('tokenAccountB'),
   Layout.publicKey('tokenPool'),
-  BufferLayout.blob(7, 'padding'),
-  Layout.uint64('feesDenominator'),
   Layout.uint64('feesNumerator'),
-  BufferLayout.blob(7, 'padding'),
+  Layout.uint64('feesDenominator'),
 ]);
 
 /**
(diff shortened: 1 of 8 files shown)


token-swap/program/src/instruction.rs
```diff
@@ -305,29 +311,53 @@ mod tests {
...
             fee_numerator,
             fee_denominator,
             nonce,
-        });
-        let packed = check.serialize().unwrap();
-        let unpacked = SwapInstruction::deserialize(&packed).unwrap();
+        };
+        let packed = check.pack();
+        println!("{:?}", packed);
+        let unpacked = SwapInstruction::unpack(&packed).unwrap();
```
CriesofCarrots, line 325: It doesn't look like we compare this `packed` against the expected `data` below. I like the patterns for the other instructions better.
Also axe that `println` 😄

token-swap/program/src/processor.rs
```diff
@@ -34,19 +34,23 @@ const TOKEN_PROGRAM_ID: Pubkey = Pubkey::new_from_array([1u8; 32]);
 pub struct Processor {}
 impl Processor {
     /// Deserializes a spl_token `Account`.
-    pub fn token_account_deserialize(data: &[u8]) -> Result<spl_token::state::Account, Error> {
-        spl_token::state::Account::unpack(data).map_err(|_| Error::ExpectedAccount)
+    pub fn token_account_deserialize(data: &[u8]) -> Result<spl_token::state::Account, SwapError> {
```
CriesofCarrots, line 37: Now that we're using `unpack` everywhere, maybe rename these methods?
Suggest: `unpack_token_account` and `unpack_mint`

token-swap/program/src/processor.rs
```diff
@@ -218,23 +227,26 @@ impl Processor {
         let destination_info = next_account_info(account_info_iter)?;
         let token_program_info = next_account_info(account_info_iter)?;
 
-        let token_swap = State::deserialize(&swap_info.data.borrow())?.token_swap()?;
+        let token_swap = SwapInfo::unpack(&swap_info.data.borrow())?;
+        if !token_swap.is_initialized {
+            return Err(SwapError::AlreadyInUse.into());
+        }
```
CriesofCarrots, line 233: We do this check several times; maybe time for an `unpack_checked` method? (Or an `unpack_unchecked` method for the init instruction, if you prefer)

CriesofCarrots: Also, should this be `SwapError::InvalidState`?

token-swap/js/client/token-swap.js
```diff
@@ -97,15 +97,13 @@ type TokenSwapInfo = {|
  * @private
  */
 const TokenSwapLayout = BufferLayout.struct([
-  BufferLayout.u8('state'),
+  BufferLayout.u8('is_initialized'),
```
CriesofCarrots, line 100: Camel-case in js, please

token-swap/js/client/token-swap.js
```diff
@@ -273,7 +270,7 @@ export class TokenSwap {
 
     const data = Buffer.from(accountInfo.data);
     const tokenSwapInfo = TokenSwapLayout.decode(data);
-    if (tokenSwapInfo.state !== 1) {
+    if (!tokenSwapInfo.is_initialized) {
```
CriesofCarrots, line 273: Ditto (camel-case)

CriesofCarrots (commented):
Looking good! Just a few small suggestions

token-swap/js/client/token-swap.js
joncinque: Oops, thanks for the catch

token-swap/program/src/instruction.rs
joncinque: That's embarrassing :sweat_smile: and that's right, the check was never done.  I'll make it look like the other tests

token-swap/program/src/processor.rs
joncinque: Definitely

token-swap/program/src/processor.rs
joncinque: I'll follow the `unpack_unchecked` model from `Token`, and absolutely, `InvalidState` is a better error

CriesofCarrots (approved):
Nice!

PR #426, opened 2020-09-11
WIP BUT NEEDS TRAVIS: Fix solana-localnet docker ps command in Travis

joncinque:
**Problem**

We can't enable the token-swap js tests in Travis because the solana-localnet container isn't properly stopped during the token js tests, so this will investigate and fix that by triggering minimal Travis runs with more output.


joncinque:
Looks like the solution is to add a slash to the start of the name, so that'll get done from the web3 side.  Closing

PR #433, opened 2020-09-11
Allow source account to be its own owner in Transfer and Transfer2

### LINKED ISSUE #427: `process_transfer`'s error checking is inconsistent with the rest of the program
Inconsistent error check order, 'process_transfer` checks for the `MintMismatch` case before the `AccountFrozen` one, most other functions checks for `AccountFrozen` first.

CriesofCarrots:
#411 for /program-v3

In the transfer instruction, if the source account owns itself the instruction panics trying to double mutably borrow.

Add test and re-arrange to avoid duplicate borrow

These changes are incomplete: more duplicate key tests should be added and more instructions will probably have to be updated. To follow in separate PRs

Also, since I was in there:
Fixes #427

### DIFF
--- a/token/program-v3/src/processor.rs
+++ b/token/program-v3/src/processor.rs
@@ -1170,7 +1170,7 @@ mod tests {
                 &account2_key,
                 &account1_key,
                 &[],
-                1000,
+                500,
             )
             .unwrap(),
             vec![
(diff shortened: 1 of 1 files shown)


jackcmay (approved):

CriesofCarrots:
Fyi @jackcmay , I added a few more test cases here to ensure that other duplicates (destination-owner and source-multisig-signer) cause no issues

jackcmay:
Looks like this same issue also applies to most of the other instructions

CriesofCarrots:
> Looks like this same issue also applies to most of the other instructions

Yep, expect more PRs shortly!

PR #434, opened 2020-09-11
Allow source account to be its own owner in Approve, Approve2, and Revoke

CriesofCarrots:
As in the transfer instruction (cc #433 ), if the source account of an approve or revoke instruction owns itself, the instruction panics trying to double mutably borrow.

Add tests and re-order process_approve and process_revoke to allow source-owner

### DIFF
--- a/token/program-v3/src/processor.rs
+++ b/token/program-v3/src/processor.rs
@@ -294,26 +294,27 @@ impl Processor {
         let account_info_iter = &mut accounts.iter();
         let source_account_info = next_account_info(account_info_iter)?;
 
-        let mut source_data = source_account_info.data.borrow_mut();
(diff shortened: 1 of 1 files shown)


jackcmay:
I think using `unpack` rather than `unpack_mut` is clearer anyway, maybe we get rid of `unpack_mut` altogether?

CriesofCarrots:
> I think using `unpack` rather than `unpack_mut` is clearer anyway, maybe we get rid of `unpack_mut` altogether?

That sounds like a good idea. After I get through testing/fixing all the instructions, I'll take a peek and see if we really need it anywhere. There are probably some test usages, but nothing that can't be rewritten.

PR #455, opened 2020-09-16
Token-swap: add tests for all instructions

joncinque:
Testing a program with a lot of parameters gets verbose very quickly, so this aims to reduce copy-pasta while also meaningfully testing the conditions.

One part to consider integrating in the `token` program is the addition to `do_process_instruction` which "sandboxes" the CPI calls by only applying the changes to accounts if the whole instruction succeeds.

Let me know if you have any suggestions for additional tests!


token-swap/program/src/processor.rs
```diff
@@ -163,28 +163,42 @@ impl Processor {
...
         if token_a.delegate.is_some() {
             return Err(SwapError::InvalidDelegate.into());
         }
         if token_b.delegate.is_some() {
             return Err(SwapError::InvalidDelegate.into());
         }
+        if pool_mint.supply != 0 {
+            return Err(SwapError::InvalidSupply.into());
+        }
+        // double check to be totally sure -- TODO check if this is necessary
```
joncinque, line 198: Is it worth having this second check?  I'm hoping that just checking the pool mint supply is enough.

CriesofCarrots (approved):
@joncinque this is awesome! I like the pattern of using separate blocks to set specific error cases and then reset; it's very easy to follow. I think I'll steal it to improve the spl-token test suite ;)
Thanks for the error improvements. Super easy to see what cases we've covered.

token-swap/program/src/processor.rs
joncinque: Great idea -- I'll add something to SwapInstruction::Initialize then

PR #462, opened 2020-09-17
token-swap: Remove tokenSwapAccount as signer

joncinque:
**Problem**

We unnecessarily set `tokenSwapAccount` as a signer of the initialize transaction, and it's not required.  Please correct me if I've misunderstood how this should work!

**Solution**

Remove the parameter.

### DIFF
--- a/token-swap/js/client/token-swap.js
+++ b/token-swap/js/client/token-swap.js
@@ -211,7 +211,7 @@ export class TokenSwap {
     );
 
     let keys = [
-      {pubkey: tokenSwapAccount.publicKey, isSigner: true, isWritable: true},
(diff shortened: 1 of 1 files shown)


jstarry:
@joncinque you're correct!

This is a good point because it reveals a security issue in the current client implementation. As it stands, the client creates the account and then someone else could front run and initialize it to something malicious. It would be better if the account creation and initialization was done in a single atomic instruction.

jstarry (approved):
Do you mind creating an issue for the point I raised? It doesn't need to block this small fix but it should be fixed as well

joncinque:
Sure, will do, that's an excellent point.

PR #463, opened 2020-09-17
Token-swap ts, flow bindings and build using rollup.

dankelleher:
Added token-swap ts and flow bindings and build pure JS versions using rollup, to match the token program


joncinque:
This looks great -- thanks for the work!

joncinque:
Do you mind if I merge?

dankelleher:
Go for it!

PR #475, opened 2020-09-18
token-swap js: Prepare package for npm

joncinque:
* Run `pretty` properly
* Add `build:program` like for spl-token
* Add TypeScript to CI
* snake_case -> camelCase variables
* Combine transactions on swap initialization
* Make instruction functions static


token-swap/js/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "description": "SPL Token Swap JavaScript API",
   "license": "MIT",
-  "author": "Solana Maintainers <maintainers@solana.com>",
+  "author": "Solana Maintainers <maintainers@solana.foundation>",
```
joncinque, line 6: Is this correct? I noticed the packages in the monorepo used this address

joncinque:
Please let me know if there are any other things required to get this package out.  I aimed to model this after the token package.

token-swap/js/package.json
mvines: yep!  it's correct

mvines:
Do we just want to drop the flow binding?  I'm not opposed

mvines (approved):
lgtm overall!

CriesofCarrots:
🐫

PR #484, opened 2020-09-20
Bump solana-sdk to v1.3.12

### LINKED ISSUE #465: token / token-swap: Use COption and Pack from sdk instead of spl_token
**Problem**

Once https://github.com/solana-labs/solana/pull/12294 is released, token-swap should use the SDK variants for `Pack` and `COption`.

**Solution**

Use them.

jackcmay:
Bump solana-sdk to v.1.3.12 and pull COption and Pack from the sdk for `token/program-v3`

Fixes #465


jackcmay:
@joncinque I marked this PR as fixing #465 but it doesn't switch `token-swaps` use of `COption` or `Pack` to the sdk because token-swap depends on token v2.  We are not making program code changes to `spl-token-v2`.  Maybe we should split #465 into a new issue that covers updating `token-swap` to `spl-token-v2`?

CriesofCarrots (approved):

joncinque:
@jackcmay Sure no problem. I'll update #465 for just token, close it, then create a new one for token-swap

PR #493, opened 2020-09-21
Token stake pool

aeyakovenko:
A pool program for SoMs.  User's can deposit into the pool, and the SoM collects a fee as a portion of a the pool.


stake-pool/program/src/state.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Pool Mint
+    pub pool_mint: Pubkey,
+    /// Owner fee account
+    pub owner_fee_account: Pubkey,
+    /// Pool token program id
+    pub token_program_id: Pubkey,
+    /// total stake under management
+    pub stake_total: u64,
+    /// total pool
+    pub pool_total: u64,
```
joncinque, line 30: You don't need to hold onto this, since the pool mint gives you the total supply

stake-pool/program/src/state.rs
```diff
@@ -0,0 +1,102 @@
...
+pub struct StakePool {
+    /// Owner authority
+    /// allows for updating the staking authority
+    pub owner: Pubkey,
+    /// Deposit authority nonce
+    /// for `create_program_address(&[state::StakePool account, "deposit"])`
+    pub deposit_nonce: u8,
+    /// Withdrawal authority nonce
+    /// for `create_program_address(&[state::StakePool account, "withdrawal"])`
+    pub withdraw_nonce: u8,
```
joncinque, line 20: Why are there two different authorities?  It seems like this could all be done with just one

stake-pool/program/src/state.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub pool_total: u64,
+    /// Fee applied to deposits
+    pub fee: Fee,
+}
+impl StakePool {
+    /// calculate the pool tokens that should be minted
+    pub fn calc_pool_amount(&self, stake_lamports: u64) -> Option<u128> {
+        (stake_lamports as u128)
+            .checked_mul((self.pool_total as u128).checked_add(1)?)?
+            .checked_div((self.stake_total as u128).checked_add(1)?)
```
joncinque, line 39: Maybe the pool could be initialized with a stake, this way you wouldn't need to do all of these adds.

stake-pool/program/src/state.rs
```diff
@@ -0,0 +1,102 @@
...
+pub enum State {
+    /// Unallocated state, may be initialized into another state.
+    Unallocated,
+    /// Initialized state.
+    Init(StakePool),
+}
+
+impl State {
+    /// Deserializes a byte buffer into a [State](struct.State.html).
+    pub fn deserialize(input: &[u8]) -> Result<State, ProgramError> {
```
joncinque, line 61: You'll probably want to do some packing on these later

stake-pool/program/src/instruction.rs
```diff
@@ -0,0 +1,176 @@
...
+    Deposit,
+
+    ///   Withdraw the token from the pool at the current ratio.
+    ///   The amount withdrawn is the MIN(u64, stake size)
+    ///   
+    ///   0. `[]` StakePool
+    ///   1. `[]` withdraw  authority
+    ///   2. `[w]` SOURCE Pool account, amount is transferable by authority
+    ///   3. `[w]` Pool MINT account, authority is the owner
+    ///   4. `[w]` Stake SOURCE owned by the withdraw authority
```
joncinque, line 67: Double-checking how this works -- the user needs to get all of the stake accounts controlled by the withdraw authority and choose one with enough tokens for this, correct?  That could cause some issues for people trying to withdraw a large amount, and makes me think it would be better for the StakePool to control one stake account, that it merges other stakes into.  The drawback with that approach is that then the StakePool can only delegate to one staking authority.  How do you envision this being used?

stake-pool/program/src/error.rs
```diff
@@ -0,0 +1,53 @@
...
+    CalculationFailure,
+}
+impl From<Error> for ProgramError {
+    fn from(e: Error) -> Self {
+        ProgramError::Custom(e as u32)
+    }
+}
+impl<T> DecodeError<T> for Error {
+    fn type_of() -> &'static str {
+        "Swap Error"
```
joncinque, line 51: nit: "Token stake error" or something like that

stake-pool/program/src/instruction.rs
```diff
@@ -0,0 +1,176 @@
...
+    ///   0. `[w]` StakePool
+    ///   1. `[s]` Owner
+    ///   2. '[]` New owner pubkey
+    ///   3. '[]` New owner fee account
+    SetOwner,
+}
+
+impl StakePoolInstruction {
+    /// Deserializes a byte buffer into an [StakePoolInstruction](enum.StakePoolInstruction.html).
+    pub fn deserialize(input: &[u8]) -> Result<Self, ProgramError> {
```
joncinque, line 92: You'll eventually want to do some more efficient packing here as needed

stake-pool/program/src/processor.rs
```diff
@@ -0,0 +1,598 @@
...
+const MY_PROGRAM_ID: Pubkey = Pubkey::new_from_array([2u8; 32]);
+
+/// Routes invokes to the token program, used for testing.
+#[cfg(not(target_arch = "bpf"))]
+pub fn invoke_signed<'a>(
+    instruction: &Instruction,
+    account_infos: &[AccountInfo<'a>],
+    signers_seeds: &[&[&[u8]]],
+) -> ProgramResult {
+    let mut new_account_infos = vec![];
```
joncinque, line 386: For your own testing, you'll want to add an additional check to make sure that the token or staking program id is passed in, which will let you choose which `process` to call -- see https://github.com/solana-labs/solana-program-library/blob/master/token-swap/program/src/processor.rs#L500

joncinque (commented):
Mainly a lot of questions, but this is looking on track for the most part.  It also helped me learn a lot about the staking program.

stake-pool/program/src/state.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// for `create_program_address(&[state::StakePool account, "withdrawal"])`
+    pub withdraw_nonce: u8,
+    /// Pool Mint
+    pub pool_mint: Pubkey,
+    /// Owner fee account
+    pub owner_fee_account: Pubkey,
+    /// Pool token program id
+    pub token_program_id: Pubkey,
+    /// total stake under management
+    pub stake_total: u64,
```
joncinque, line 28: Will this be updated somehow from the outside as stakes grow in value?

stake-pool/program/src/instruction.rs
aeyakovenko: @joncinque that's ok, they can aggregate on the UI side.

stake-pool/program/src/state.rs
aeyakovenko: Yea, we can a separate instruction to update the total if a stake gets slashed.

stake-pool/program/src/state.rs
aeyakovenko: I need a way to track which stakes have been deposited and which have not.

stake-pool/program/src/state.rs
aeyakovenko: There is no way to guarantee stake_total will never be 0, might be simpler to just error out here though.

stake-pool/program/src/state.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// calculate the pool tokens that should be minted
+    pub fn calc_pool_amount(&self, stake_lamports: u64) -> Option<u128> {
+        (stake_lamports as u128)
+            .checked_mul((self.pool_total as u128).checked_add(1)?)?
+            .checked_div((self.stake_total as u128).checked_add(1)?)
+    }
+    /// calculate the fee in pool tokens that goes to the owner
+    pub fn calc_fee_amount(&self, pool_amount: u128) -> Option<u128> {
+        pool_amount
+            .checked_mul((self.fee.numerator as u128).checked_add(1)?)?
```
joncinque, line 44: Actually, just noticing this more, this will cause a problem if someone wants a 0 fee

stake-pool/program/src/state.rs
```diff
@@ -0,0 +1,102 @@
...
+            .checked_div((self.fee.denominator as u128).checked_add(1)?)
```
joncinque, line 45: And you should probably have a check on init that this is not 0

stake-pool/program/src/state.rs
joncinque: In that case, you may prefer special logic for an empty pool or stake, where you default to 1:1 if either pool_total or stake_total is 0.  Otherwise do this calculation without the addition

stake-pool/program/src/processor.rs
```diff
@@ -29,89 +30,94 @@ impl Processor {
...
     /// Issue a stake_split instruction.
-    pub fn stake_split(
-        _accounts: &[AccountInfo],
-        _my_info: &Pubkey,
-        _authority: &Pubkey,
-        _nonce: u8,
-        _stake: &Pubkey,
-        _amount: u64,
+    pub fn stake_split<'a>(
+        me: &Pubkey,
```
joncinque, line 34: Nit: Can you rename this to `stake_pool`?

stake-pool/program/src/processor.rs
```diff
@@ -29,89 +30,94 @@ impl Processor {
...
     /// Issue a stake_set_owner instruction.
-    pub fn stake_set_stake_auth(
-        _accounts: &[AccountInfo],
-        _my_info: &Pubkey,
-        _authority: &Pubkey,
-        _nonce: u8,
-        _stake: &Pubkey,
-        _owner: &Pubkey,
+    pub fn stake_authorize<'a>(
+        me: &Pubkey,
```
joncinque, line 52: Nit: ditto here

stake-pool/program/src/processor.rs
```diff
@@ -29,89 +30,94 @@ impl Processor {
...
     /// Issue a spl_token `Burn` instruction.
-    pub fn token_burn(
-        accounts: &[AccountInfo],
-        my_info: &Pubkey,
-        token_program_id: &Pubkey,
-        burn_account: &Pubkey,
-        mint: &Pubkey,
-        authority: &Pubkey,
+    pub fn token_burn<'a>(
+        me: &Pubkey,
```
joncinque, line 70: Nit: and here

stake-pool/program/src/processor.rs
```diff
@@ -29,89 +30,94 @@ impl Processor {
...
     /// Issue a spl_token `MintTo` instruction.
-    pub fn token_mint_to(
-        accounts: &[AccountInfo],
-        my_info: &Pubkey,
-        token_program_id: &Pubkey,
-        mint: &Pubkey,
-        destination: &Pubkey,
-        authority: &Pubkey,
+    pub fn token_mint_to<'a>(
+        me: &Pubkey,
```
joncinque, line 100: Nit: last one here

stake-pool/program/src/processor.rs
```diff
@@ -0,0 +1,609 @@
...
+                    let signer = Pubkey::create_program_address(seeds, &MY_PROGRAM_ID).unwrap();
+                    if *account_info.key == signer {
+                        new_account_info.is_signer = true;
+                    }
+                }
+                new_account_infos.push(new_account_info);
+            }
+        }
+    }
+    spl_token::processor::Processor::process(
```
joncinque, line 412: What's the plan to call the stake program in testing?  Is it waiting for https://github.com/solana-labs/solana/pull/12653 to land?

joncinque (commented):
Looking good, just a few nits. it'll need tests on all the instructions and JS bindings for end-to-end testing.  Do you want to commit this as is and wait for Jack's PR to make staking available in SDK?

joncinque:
Close / re-open to kick off CI

joncinque:
Looks like the PR is getting bounced from CI for "abuse detected", I'll message Travis to get that resolved

joncinque:
CI working correctly again

PR #507, opened 2020-09-21
Backport v3 to v2

jackcmay:
All commits in this PR will be retained (no squash)


token/program/src/error.rs
```diff
@@ -34,9 +34,6 @@ pub enum TokenError {
     /// Invalid number of required signers.
     #[error("Invalid number of required signers")]
     InvalidNumberOfRequiredSigners,
-    /// State is uninitialized.
```
mvines, line 37: This appears to be an ABI change

jackcmay: True, in terms of client handling, but not an interface change ;-). We can leave this one out

jackcmay: @mvines Do you see an issue with returning different values (UninitializedState has been replaced with ProgramError::UninitializedAccount)

mvines: I guess my preference would have been to leave the old value in the enum to avoid shifting the ordinal values of the subsequent errors around -- that's probably ok but why even care to think about it

jackcmay: @mvines Yeah, that's what we are doing now, I left the enum as it was in v2.  The reason for this change was pack/unpack was returning a `TokenError` and in order to mainline into the sdk we needed it to return a `ProgramError`.  v2.1 now retains this unused `TokenError` variant but the processor will now return the `ProgramError` from `Pack`

PR #509, opened 2020-09-21
Standardize failure case ordering

### LINKED ISSUE #500: solana_token_audit#10: Inconsistent error check order
Description
Method process_transfer checks for the MintMismatch case before the AccountFrozen one,
most other functions checks for AccountFrozen first.

Proof of Issue
Method process_transfer
Filename: token/program/src/processor.rs
Beginning Line Number: 171

Severity and Impact Summary
The inconsistency in checking for a frozen account could lead to undefined results if something
will interfere with the execution. Keeping a consistent structure on the way that the checks are
made make the implementation more predictable in its behavior.

Recommendation
Change the order so that AccountFreeze is checked before MintMismatch

CriesofCarrots:
While relevant failure cases vary from instruction to instruction, there are some that are consistent across many. It makes code easier to understand and test if these consistent failure cases are checked in the same order across instructions.

Here is the general order implemented:
1. Frozen
2. Account type is wrong in some way (native vs. non-native, authority type not applicable)
3. Insufficient funds
4. Mint mismatch
5. Decimal mismatch
6. Delegated checks

Fixes #500 

- [x] Rebase on #507

### DIFF
--- a/token-swap/program/src/processor.rs
+++ b/token-swap/program/src/processor.rs
@@ -2086,7 +2086,7 @@ mod tests {
             ) = accounts.setup_token_accounts(
                 &user_key,
                 &withdrawer_key,
-                initial_a,
+                withdraw_amount,
                 initial_b,
                 withdraw_amount,
             );
--- a/token/program/src/processor.rs
+++ b/token/program/src/processor.rs
@@ -161,11 +161,11 @@ impl Processor {
         let mut source_account = Account::unpack(&source_account_info.data.borrow())?;
         let mut dest_account = Account::unpack(&dest_account_info.data.borrow())?;
 
-        if source_account.amount < amount {
-            return Err(TokenError::InsufficientFunds.into());
-        }
         if source_account.is_frozen() || dest_account.is_frozen() {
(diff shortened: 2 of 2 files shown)


jackcmay:
@CriesofCarrots Feel free to rebase now that the backports have gone in

token-swap/program/src/processor.rs
```diff
@@ -2086,7 +2086,7 @@ mod tests {
             ) = accounts.setup_token_accounts(
                 &user_key,
                 &withdrawer_key,
-                initial_a,
+                withdraw_amount,
```
jackcmay, line 2089: This change seems unrelated, in here by accident?

jackcmay (approved):

token-swap/program/src/processor.rs
CriesofCarrots: No, CI fails without this change on `Err(Custom(1))` instead of the expected `Err(Custom(3))`. This is because I changed  the order of error-checks. Updating the initial_a amount ensures that we are making it to the mint check.

Issue #502, opened 2020-09-21
solana_token_audit#12: 2 CLI not supporting all instructions


jackcmay:
Description
The CLI is missing support for the following instructions:
• Approve
• CloseAccount
• FreezeAccount (covered by #365)
• Revoke
• SetAuthority (FreezeAccount). 
• SetAuthority (MintToken),
• ThawAccount (covered by #365)

Proof of Issue
Filename: token/cli/src/main.rs
Beginning Line Number: N/A

Severity and Impact Summary
The expected functionality is not present.

Recommendation
Either implement these instructions or mention why they are not to be implemented.

- [x] Approve (#540)
- [ ] CloseAccount
- [x] FreezeAccount (covered by #365)
- [x] Revoke (#540)
- [x] SetAuthority (FreezeAccount).  (#531)
- [x] SetAuthority (MintToken), (#531)
- [x] ThawAccount (covered by #365)

PR #511, opened 2020-09-22
Add freeze and thaw functionality to spl-token cli

### LINKED ISSUE #365: spl-token-cli:  Add support for account freeze/thawing
spl-token 2.0 added account freeze/thawing as a new feature.  This is not yet exposed in the `spl-token` command line.

CriesofCarrots:
- Add ability to initialize mint with freeze authority (set to same as mint authority for now; SetAuthority work as part of #502 will allow greater control)
- Add freeze and thaw subcommands to cli app
- Add "frozen" marker to accounts output, when relevant

Fixes #365 
Toward #502

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2267,6 +2267,7 @@ name = "spl-token-cli"
 version = "2.0.1"
 dependencies = [
  "clap",
+ "console",
  "serde_json",
  "solana-account-decoder",
  "solana-clap-utils",
--- a/token/cli/Cargo.toml
+++ b/token/cli/Cargo.toml
@@ -10,6 +10,7 @@ version = "2.0.1"
 
 [dependencies]
 clap = "2.33.3"
+console = "0.11.3"
 serde_json = "1.0.57"
 solana-account-decoder = { version = "1.3.12" }
 solana-clap-utils = { version = "1.3.12"}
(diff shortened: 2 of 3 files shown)


mvines (approved):
The only minor nit I have is wrt. the user needing to provide the mint address to freeze/thaw.  It's an extra RPC call but from the token account address itself the mint can be determined without the user -- `solana freeze blah account` then gets reduced to `solana freeze account`

CriesofCarrots:
> The only minor nit I have is wrt. the user needing to provide the mint address to freeze/thaw. It's an extra RPC call but from the token account address itself the mint can be determined without the user -- `solana freeze blah account` then gets reduced to `solana freeze account`

Completely fair, will update. I was trying to stay consistent with mint-to... I'm a little torn about the need for the mint address param in mint-to subcommand, but it definitely makes more sense in that case.

CriesofCarrots:
> The only minor nit I have is wrt. the user needing to provide the mint address to freeze/thaw. It's an extra RPC call but from the token account address itself the mint can be determined without the user -- `solana freeze blah account` then gets reduced to `solana freeze account`

Actually, this will be much easier once solana-client v1.3.12+ is published, using: https://github.com/solana-labs/solana/blob/64af7127234ee0510cae3adfa18271ba437672bd/client/src/rpc_client.rs#L734

Publishing an updated solana-client depends on a new spl-token on crates.io that reverts the multisig Instruction names (and also accommodates the new xChecked Instruction names). So I'm going to merge this as-is with a note to update/simplify when solana-client is published.

PR #517, opened 2020-09-22
Expose layouts and instruction creation function

dankelleher:
Expose Layouts and instruction creation functions to allow more flexible use of the Token and TokenSwap clients.

This PR does two things:

- expose the Mint, Account and TokenSwap layouts, so that you can TransactionInstructions to create system program accounts for each of them. The alternative is using the client functions, which send a transaction immediately and therefore require passing an account (with secret key).
- expose a static function to create a TransactionInstruction that initialises the token swap, again to avoid having to pass an account..


joncinque:
Looks great -- thanks for the work!

PR #519, opened 2020-09-23
Fix lint check in CI

joncinque:
* `npm run lint` now checks that code is properly formatted, instead of
  running the formatter
* Add extra commands to run the linting, include ts definitions file
* Update flow-bin, fix type errors that came up

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -62,7 +62,7 @@ cargo test --manifest-path=token/perf-monitor/Cargo.toml -- --nocapture
 js_token() {
   cd token/js
   time npm install || exit $?
-  time npm run lint || exit $?
+  time npm run lint:check || exit $?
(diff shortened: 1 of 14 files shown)


jstarry:
Looks great but it would be nice to keep the npm command in-line with the web3 sdk

For example, I've gotten used to using `npm run lint:fix` to both fix/check lint errors and apply formatting

joncinque:
No problem.  There's some inconsistency though, since `lint` runs `prettier --write`, which is not a check.  I'll update it so `lint` is a pure check and `lint:fix` fixes / writes.  I can do the same for web3.js.  what do you think?

jstarry:
that sounds awesome 👍

jstarry (approved):

PR #537, opened 2020-09-25
Upgrade web3 to 0.78.0 and sdk to 1.3.13

joncinque:
Also, update `sendAndConfirmTransaction` to use `preflightCommitment` of `recent`. For the token tests, on my machine, this setup took roughly 18.5 seconds, and with `skipPreflight: true` it was about 21.5 seconds.

The main question: should the commitments move to `singleGossip`?

cc @jstarry for awareness -- thanks for the announcement!

### DIFF
--- a/bpf-sdk-install.sh
+++ b/bpf-sdk-install.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-channel=${1:-v1.3.12}
+channel=${1:-v1.3.13}
 installDir="$(dirname "$0")"/bin
 cacheDir=~/.cache/solana-bpf-sdk/"$channel"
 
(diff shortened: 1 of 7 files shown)


jstarry:
Yeah `singleGossip` is preferred but I think there is a bug where RPC requests treat `singleGossip` as `single` right now which will cause races

joncinque:
I tested around a bunch of different options for `preflightCommitment` on localnet, and I could only get `recent` to work -- `singleGossip`, `single`, `max`, and `root` all failed with:
`Error: failed to send transaction: Transaction simulation failed: Attempt to debit an account but found no record of a prior credit.`
Looking at the code quickly, it seems really strange that this would fail.  I can update the normal `commitment` to `singleGossip` and create an issue for the `preflightCommitment`.

CriesofCarrots (approved):
Wfm to merge this for now and add an issue for `preflightCommitment`; this is a step in the right direction, at least!

PR #541, opened 2020-09-25
token js: Reduce airdrop amounts

CriesofCarrots:
The token js tests (`npm run start`) timeout against devnet and testnet because they request airdrops larger than the caps in place. This isn't a permanent fix against very low caps or non-existent faucets in the future, but might as well facilitate these tests for now.


CriesofCarrots:
Not receiving travis-ci updates, but CI passed

PR #554, opened 2020-09-30
token-swap: Use sdk version of Pack

### LINKED ISSUE #492: token-swap: Update to COption and Pack from SDK instead of spl_token
**Problem**

Once solana-labs/solana#12294 is released, token-swap should use the SDK variants for Pack and COption.

**Solution**

Use them.

joncinque:
**Problem**

Token-swap is using an ad-hoc version of Pack which doesn't exactly follow the traits.

**Solution**

Use the sdk version of Pack and update other bits as needed.

Fixes #492

### DIFF
--- a/token-swap/program/src/processor.rs
+++ b/token-swap/program/src/processor.rs
@@ -615,7 +615,7 @@ mod tests {
             token_b_amount: u64,
         ) -> Self {
             let swap_key = pubkey_rand();
-            let swap_account = Account::new(0, SwapInfo::LEN, &SWAP_PROGRAM_ID);
+            let swap_account = Account::new(0, SwapInfo::get_packed_len(), &SWAP_PROGRAM_ID);
             let (authority_key, nonce) =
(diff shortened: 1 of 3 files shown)


token-swap/program/src/processor.rs
```diff
@@ -617,7 +615,7 @@ mod tests {
             token_b_amount: u64,
         ) -> Self {
             let swap_key = pubkey_rand();
-            let swap_account = Account::new(0, size_of::<SwapInfo>(), &SWAP_PROGRAM_ID);
+            let swap_account = Account::new(0, SwapInfo::LEN, &SWAP_PROGRAM_ID);
```
CriesofCarrots, line 618:
```suggestion
            let swap_account = Account::new(0, SwapInfo::get_packed_len(), &SWAP_PROGRAM_ID);
```

CriesofCarrots (dismissed):
One nit, take it or leave it.
Lgtm!

Issue #88, opened 2020-07-07
token-swap: Add constraint to liquidity deposits / withdraws


jstarry:
#### Problem
When users deposit / withdraw liquidity they have no assurance that the ratio they see in the UI is the same as what's executed on-chain

#### Proposed Change
TBD

joncinque:
Fixed with #560

PR #560, opened 2020-10-01
token-swap: Add slippage to swap / withdraw / deposit

### LINKED ISSUE #87: token-swap: Add constraint to swap instruction to prevent too much slippage
#### Problem
Users don't have any guarantee that the token ratio they submit a swap at will be the same as what is executed on-chain

#### Proposed Changes
* Add a min output to swap instruction

joncinque:
**Problem**

When performing swap, withdraw, or deposit, there's a chance that we'll receive much less than expected due to slippage.

**Solution**

Add minimum / maximum amount parameters for all commands, along with updates to JS bindings.

Fixes #87 and #88

### DIFF
--- a/token-swap/js/cli/token-swap-test.js
+++ b/token-swap/js/cli/token-swap-test.js
@@ -37,8 +37,9 @@ let tokenAccountB: PublicKey;
 
 // Initial amount in each swap token
 const BASE_AMOUNT = 1000;
-// Amount passed to instructions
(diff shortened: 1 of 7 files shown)


CriesofCarrots (dismissed):
Nice and simple. Lgtm! (Looks like just some prettier issues in CI)

joncinque:
Ah, right.  Thanks for the quick review!

PR #562, opened 2020-10-02
token-swap: Assess swap fee on input token

### LINKED ISSUE #341: token-swap: Swap fee calculation only assessed on returned token
**Problem**

This could actually be a desired feature.  According to the uniswap fee calculation https://hackmd.io/@HaydenAdams/HJ9jLsfTz#ETH-%E2%87%84-ERC20-Trades, the fee is assessed from the token being put in.  For example, if you are interacting with an A to B token swap, you are putting in 1 A and expecting some amount of B.  When calculating the fee, uniswap takes out the fee from the 1 A being put in, and gives the equivalent B token from (1 - fee), essentially "spreading" the fee across both sides of the liquidity pool.

In our implementation, we put the 1 A in full into the A liquidity, and then we assess the fee entirely on B.  The overall effect on the pool value is the same, but when we withdraw from the pool, the gains from fees are slightly unbalanced.

**Solution**

Update the swap fee calculation to line up with uniswap

joncinque:
**Problem**

In our implementation of fee calculation, we assess the fee entirely on the output token, but Uniswap and Balancer assess on the input. The overall effect on the pool value is the same, but when we withdraw from the pool, the gains from fees are slightly unbalanced.

**Solution**

Update the swap fee calculation to line up with uniswap.  Also, update the SDK version everywhere.

Fixes #341

### DIFF
--- a/token-swap/program/src/curve.rs
+++ b/token-swap/program/src/curve.rs
@@ -35,7 +35,7 @@ impl SwapResult {
         let new_source_amount_less_fee = swap_source_amount
             .checked_add(source_amount)?
             .checked_sub(fee)?;
-        let new_destination_amount = invariant.checked_div(new_source_amount)?;
+        let new_destination_amount = invariant.checked_div(new_source_amount_less_fee)?;
         let amount_swapped = swap_destination_amount.checked_sub(new_destination_amount)?;
 
         // actually add the whole amount coming in
--- a/token-swap/program/src/curve.rs
+++ b/token-swap/program/src/curve.rs
@@ -27,14 +27,19 @@ impl SwapResult {
         fee_denominator: u64,
     ) -> Option<SwapResult> {
         let invariant = swap_source_amount.checked_mul(swap_destination_amount)?;
-        let new_source_amount = swap_source_amount.checked_add(source_amount)?;
-        let new_destination_amount = invariant.checked_div(new_source_amount)?;
-        let remove = swap_destination_amount.checked_sub(new_destination_amount)?;
-        let fee = remove
+
+        // assess the fee to calculate the amount swapped
+        let fee = source_amount
             .checked_mul(fee_numerator)?
             .checked_div(fee_denominator)?;
-        let new_destination_amount = new_destination_amount.checked_add(fee)?;
-        let amount_swapped = remove.checked_sub(fee)?;
+        let new_source_amount = swap_source_amount
+            .checked_add(source_amount)?
+            .checked_sub(fee)?;
+        let new_destination_amount = invariant.checked_div(new_source_amount)?;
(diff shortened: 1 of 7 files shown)


token-swap/program/src/curve.rs
```diff
@@ -27,14 +27,19 @@ impl SwapResult {
...
-        let remove = swap_destination_amount.checked_sub(new_destination_amount)?;
-        let fee = remove
+
+        // assess the fee to calculate the amount swapped
+        let fee = source_amount
             .checked_mul(fee_numerator)?
             .checked_div(fee_denominator)?;
-        let new_destination_amount = new_destination_amount.checked_add(fee)?;
-        let amount_swapped = remove.checked_sub(fee)?;
+        let new_source_amount = swap_source_amount
```
CriesofCarrots, line 35:
```suggestion
        let new_source_amount_less_fee = swap_source_amount
```
...or similar distinct variable name makes the math (and policy) a little clearer here. Will make it easier to write up docs (something like the link in the issue?) later on.

token-swap/program/src/curve.rs
```diff
@@ -27,14 +27,19 @@ impl SwapResult {
         fee_denominator: u64,
     ) -> Option<SwapResult> {
         let invariant = swap_source_amount.checked_mul(swap_destination_amount)?;
-        let new_source_amount = swap_source_amount.checked_add(source_amount)?;
-        let new_destination_amount = invariant.checked_div(new_source_amount)?;
-        let remove = swap_destination_amount.checked_sub(new_destination_amount)?;
-        let fee = remove
+
+        // assess the fee to calculate the amount swapped
```
CriesofCarrots, line 31:
```suggestion
        // debit the fee to calculate the amount swapped
```
Assess is correct, of course, but its other meaning makes it ambiguous

CriesofCarrots (approved):
Just nits for clarity. Actual math looks good :)

joncinque:
Makes sense, thanks @CriesofCarrots !

token-swap/program/src/curve.rs
```diff
@@ -27,14 +27,19 @@ impl SwapResult {
...
+        // debit the fee to calculate the amount swapped
+        let fee = source_amount
             .checked_mul(fee_numerator)?
             .checked_div(fee_denominator)?;
-        let new_destination_amount = new_destination_amount.checked_add(fee)?;
-        let amount_swapped = remove.checked_sub(fee)?;
+        let new_source_amount_less_fee = swap_source_amount
+            .checked_add(source_amount)?
+            .checked_sub(fee)?;
+        let new_destination_amount = invariant.checked_div(new_source_amount)?;
```
CriesofCarrots, line 38:
```suggestion
        let new_destination_amount = invariant.checked_div(new_source_amount_less_fee)?;
```

PR #568, opened 2020-10-06
Add JS test for creating accounts and swapping in one transaction

joncinque:
This is WIP, but it can showcase how to reduce the number of separate roundtrips required to do many things at once.  It may be worth porting all of the token-swap tests to look like this to speed them considerably.

### DIFF
--- a/token-swap/js/cli/token-swap-test.js
+++ b/token-swap/js/cli/token-swap-test.js
@@ -6,11 +6,14 @@ import {
   Connection,
   BpfLoader,
   PublicKey,
+  SystemProgram,
+  Transaction,
   BPF_LOADER_PROGRAM_ID,
 } from '@solana/web3.js';
 
-import {Token} from '../../../token/js/client/token';
+import {AccountLayout, Token} from '../../../token/js/client/token';
 import {TokenSwap, CurveType} from '../client/token-swap';
+import {sendAndConfirmTransaction} from '../client/util/send-and-confirm-transaction';
 import {Store} from '../client/util/store';
 import {newAccountWithLamports} from '../client/util/new-account-with-lamports';
 import {url} from '../url';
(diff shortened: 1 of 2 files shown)


token-swap/js/cli/token-swap-test.js
```diff
@@ -325,6 +328,74 @@ export async function withdraw(): Promise<void> {
...
+  // Send the instructions
+  console.log('sending big instruction');
+  await sendAndConfirmTransaction(
+    'do a whole lot',
+    connection,
+    transaction,
+    owner,
+    newAccount,
+    mintA.payer,
+    mintB.payer,
```
rawfalafel, line 395: Does this transaction require 4 signatures? I'm not following why all these signers are required.

joncinque: You're absolutely right, I messed up one the first instruction which makes this require `mintB.payer` -- lemme get an update to you

token-swap/js/cli/token-swap-test.js
```diff
@@ -325,6 +328,72 @@ export async function withdraw(): Promise<void> {
...
 
+export async function swapOneTransaction(): Promise<void> {
+  console.log('Creating swap token a account');
+  let userAccountA = await mintA.createAccount(owner.publicKey);
+  await mintA.mintTo(userAccountA, owner, [], SWAP_AMOUNT_IN);
+
+  const balanceNeeded = await Token.getMinBalanceRentForExemptAccount(
+    connection,
+  );
+  const newAccount = new Account();
```
rawfalafel, line 339: For my own understanding, does this create a new keypair? In this function, it looks like `newAccount` gets passed in as a signer. If so, does the secret key for `newAccount` have to be stored so it can be reused later?

rawfalafel: My guess is that it's required for `SystemProgram.createAccount` but isn't required afterwards since `owner` has control over it. Feels weird to generate a private key and discard it afterwards, but this kind of makes sense.

joncinque: That's exactly right -- the code is really simple over at https://github.com/solana-labs/solana/blob/master/web3.js/src/account.js#L22
And yes, since `newAccount` is your new token B account, you will need access to at least its public key.  Otherwise, since it's a `Token` account, it's actually owned by the `Token` program, and `owner` has the right to issue certain instructions to the `Token` program to make changes to the account.

rawfalafel: Got it. Thanks for the explanation :+1:

PR #570, opened 2020-10-06
Remove skip-no-mangle entirely

jackcmay:
Programs now use other features to decide if the program's entrypoint should be pulled in or not.  Remove it entirely.

### DIFF
--- a/memo/program/Cargo.toml
+++ b/memo/program/Cargo.toml
@@ -12,7 +12,6 @@ edition = "2018"
 
 [features]
 no-entrypoint = []
-skip-no-mangle = []
 program = ["solana-sdk/program"]
 default = ["solana-sdk/default"]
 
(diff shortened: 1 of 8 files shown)


CriesofCarrots:
Great! If we publish new spl-token or spl-memo crates with this change, it will break solana v1.3. If we don't want to remove the tag/feature from v1.3, perhaps we should peg specific spl-token and spl-memo versions?

jackcmay:
@CriesofCarrots Are you thinking we leave token v2 and memo out of this or and remove those features from memo later if there is a version 2?

CriesofCarrots:
> Are you thinking we leave token v2 and memo out of this?

That's an option, but I was thinking we go ahead with token v2 and memo as part of this and either: 1. remove `skip-no-mangle` from solana v1.3, or 2. peg spl-token to `version = "=2.0.6"` and spl-memo to `version = "=1.0.7"` in solana v1.3

jackcmay:
I like the peg option, I'll add that to the solana pr

PR #574, opened 2020-10-07
token-swap: Add weights to swap contract

joncinque:
**Problem**

Weights for the tokens in a swap contract are very useful to come to a correct price without too much liquidity needed. For example, if token A is weighted at 10%, and token B is weighted at 90%, if you have equal liquidity at 5000: 5000, swapping 100 token A for token B will give you (according to (15) in https://balancer.finance/whitepaper/):

5000 * (1 - ( 5000 / (5000 + 100) ) ^ (1 / 9)) = 10.989

To do that, we need to be able to do fractional exponents!

**Solution**

Add some "precise" math which uses u128 under the hood, similar to the Balancer implementation: https://github.com/balancer-labs/balancer-core/blob/master/contracts/BNum.sol -- with more comments.  Note that these calculations are meant to operate on numbers between epsilon and 2, or the series won't converge.

For the power approximation, a Taylor Series did better than Newtonian root-finding.  It's high school calculus all over again!

This is still WIP, and the parts to add the weights and integrate these calculations will come tomorrow, but I wanted to get the math out there for discussion.  I first kept the numerator and denominator separate as we do elsewhere, but you very quickly blow out a u128 for a fraction like (40 / 3), which is a reasonable set of weights in a pool.


joncinque:
This is all ready to be reviewed now -- fire away!

token-swap/program/src/processor.rs
```diff
@@ -605,6 +624,7 @@ impl PrintProgramError for SwapError {
             SwapError::ExceededSlippage => {
                 info!("Error: Swap instruction exceeds desired slippage limit")
             }
+            SwapError::InvalidWeight => info!("Provided weight outside of valid range (1-99)"),
```
CriesofCarrots, line 627: Isn't max weight 100?

token-swap/program/src/math.rs
```diff
@@ -0,0 +1,265 @@
...
+
+impl PreciseNumber {
+    const ONE: Self = Self { value: ONE };
+
+    const ROUNDING_CORRECTION: u128 = ONE / 2;
+    const POW_PRECISION: u128 = 100;
+    const APPROXIMATION_ITERATIONS: u64 = 100_000;
+
+    const MIN_POW_BASE: u128 = 1;
+    const MAX_POW_BASE: u128 = 2 * ONE;
```
CriesofCarrots, line 26: Can you please add some comments about these consts?

token-swap/program/src/math.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// Wrapper over the inner value, which is multiplied by ONE
+    pub value: u128,
+}
+
+impl PreciseNumber {
+    const ONE: Self = Self { value: ONE };
+
+    const ROUNDING_CORRECTION: u128 = ONE / 2;
+    const POW_PRECISION: u128 = 100;
+    const APPROXIMATION_ITERATIONS: u64 = 100_000;
```
CriesofCarrots, line 23: What do the instruction counts on Swap execution look like with this limit?

CriesofCarrots (commented):
Phew, that was a lot of math! Just a few questions where I think we could clarify/explain more fully.
It generally looks solid, but I'm a little concerned about instruction count and the approximation iterations. Maybe time to add something like this? https://github.com/solana-labs/solana-program-library/tree/master/token/perf-monitor

token-swap/program/src/curve.rs
```diff
@@ -18,25 +20,51 @@ pub struct SwapResult {
 
 impl SwapResult {
     /// SwapResult for swap from one currency into another, given pool information
-    /// and fee
+    /// and fee.
+    ///              /          B_i       W_i / W_o \
+    /// A_o = B_o * | 1 - ---------------            |
+    ///              \    B_i + A_i - fee           /
+    ///
```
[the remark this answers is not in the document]
joncinque: I was mainly following the example at Balancer, which I found pretty neat, but it's true that having it one line is clearer -- will make that change

token-swap/program/src/math.rs
```diff
@@ -0,0 +1,265 @@
+//! Math utilities for swap calculations
+
+/// The representation of the number one as a precise number
+pub const ONE: u128 = 10_000_000_000;
+
+/// Maximum weight for token in swap
+pub const MAX_WEIGHT: u8 = 100;
```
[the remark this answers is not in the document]
joncinque: It's partly a balancer precedent, motivated by the problems when calculating huge powers.  For example, if a swap specifies weights of `10000000` and `1`, that's an exponent that we can't calculate.  I'll clarify in the comment

token-swap/program/src/math.rs
```diff
@@ -0,0 +1,265 @@
...
+        difference.value < precision
+    }
+
+    /// Floors a precise value to a precision of ONE
+    pub fn floor(&self) -> Option<Self> {
+        let value = self.value.checked_div(ONE)?.checked_mul(ONE)?;
+        Some(Self { value })
+    }
+
+    /// Performs a checked division on two "precise" unsigned integers
```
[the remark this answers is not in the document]
joncinque: Thanks for catching these -- will update.  They were from an earlier version before encapsulating `PreciseNumber`

token-swap/program/src/processor.rs
joncinque: It is, thanks for the catch.  I've actually been wondering, why do we specify the same error message in two different places?  It would be easier to just take the error message defined in `error.rs` and log that instead to avoid copy-pasting.  I'll put that in so you can see, but feel free to reject it :-)

token-swap/program/src/math.rs
joncinque: Let me add a perf monitor as you've recommended and get a sense of it.  These numbers were tuned for accuracy based on the unit tests, but could potentially be relaxed to get under any thresholds.

joncinque:
I'll add in the perf monitor as you recommend, and otherwise I'll update all of the bits of documentation that you recommended.

token-swap/program/src/math.rs
```diff
@@ -16,13 +17,32 @@ pub struct PreciseNumber {
 }
 
 impl PreciseNumber {
+    /// The number 1 as a PreciseNumber, used for easier calculations.
     const ONE: Self = Self { value: ONE };
 
+    /// Correction to apply to avoid truncation errors on division.  Since
+    /// integer operations will always floor the result, we artifically bump it
+    /// up by one half to get the expect result.
```
CriesofCarrots, line 25:
```suggestion
    /// up by one half to get the expected result.
```

token-swap/program/src/processor.rs
CriesofCarrots: No good reason that I know of -- I went through the token errors to see if there were any substantive differences and didn't find any.
So much simpler with your change 😍

joncinque:
Too much has changed since the last update on this, namely the refactoring of instruction tests, all the new fees / parameters in token swap, and even the building of programs, so I'll close this and start fresh somewhere else.

While this PR is still active though, I wanted to publish the instruction counts from preliminary testing.  To do this, I used the existing `token/perf-monitor`, adapted it to token-swap, and stubbed out the calls to `invoke_signed` in the token swap program.  Here are the numbers, calculated using the test results and adding the tolerances from `token/perf-monitor`:

```
Initialize: 17_000; // stubbed: 9735, add 6500 for mint to
Deposit: 33_000; // stubbed: 10102, add 6500 for mint to, 8000 for each transfer
Withdraw: 33_000; // stubbed: 10103, add 6500 for burn, 8000 for each transfer
Swap from weight 100 to weight 3: 55_000; // stubbed: 36678, add 8000 for each transfer
Swap from weight 3 to weight 100: 40_000; // stubbed: 20636, add 8000 for each transfer
```

I also played with the iterations in the pow calculation and removed the `break` to see how many instructions it generated.  For 100 iterations, here are the results:

```
Swap high to low weight: 108,869
Swap low to high weight: 92,070
```

For a compute budget is 200,000, this seemed like a reasonable limit.

joncinque:
Going to close this -- too many changes since this happened, but at least the math got in eventually.  We might want to create a "WeightedCurve" in order to make this happen.

Issue #581, opened 2020-10-08
Migrate from using Flow to TypeScript for JS clients


jstarry:
The SPL Token Lending project now uses TypeScript and it would be great if all SPL clients moved away from Flow and used TypeScript

joncinque:
This is done! No more Flow or even plain JS in any of the client libraries, we're all TypeScript

PR #582, opened 2020-10-08
SPL token lending scaffolding

jstarry:
A bit of a copy / paste job but with TypeScript instead of Flow


jstarry:
Follow up issues:
Migrate from using Flow to TypeScript for JS clients: https://github.com/solana-labs/solana-program-library/issues/581
Too much duplicated JS code across different SPL program clients: https://github.com/solana-labs/solana-program-library/issues/583

Issue #406, opened 2020-09-09
token-swap: Reduce parameter passing in JS client


joncinque:
**Problem**

Brought up by @dankelleher .  Many of the functions in `token-swap.js` require parameters that are already contained within the `TokenSwap` object.

**Solution**

Remove those parameters to make the JS client easier to use.

joncinque:
This was taken care of #586 -- thanks for the work @rawfalafel !

PR #586, opened 2020-10-09
Pass TokenSwap state into the constructor to simplify function calls

rawfalafel:
Addresses #406

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -561,13 +561,12 @@ dependencies = [
 [[package]]
 name = "curve25519-dalek"
 version = "2.1.0"
-source = "git+https://github.com/garious/curve25519-dalek?rev=60efef3553d6bf3d7f3b09b5f97acd54d72529ff#60efef3553d6bf3d7f3b09b5f97acd54d72529ff"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d85653f070353a16313d0046f173f70d1aadd5b42600a14de626f0dfb3473a5"
 dependencies = [
- "borsh",
  "byteorder",
  "digest 0.8.1",
  "rand_core",
- "serde",
  "subtle 2.2.3",
  "zeroize",
 ]
@@ -575,12 +574,13 @@ dependencies = [
 [[package]]
 name = "curve25519-dalek"
 version = "2.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5d85653f070353a16313d0046f173f70d1aadd5b42600a14de626f0dfb3473a5"
+source = "git+https://github.com/garious/curve25519-dalek?rev=60efef3553d6bf3d7f3b09b5f97acd54d72529ff#60efef3553d6bf3d7f3b09b5f97acd54d72529ff"
 dependencies = [
+ "borsh",
  "byteorder",
  "digest 0.8.1",
  "rand_core",
+ "serde",
  "subtle 2.2.3",
  "zeroize",
 ]
--- a/token-swap/js/client/token-swap.js
+++ b/token-swap/js/client/token-swap.js
@@ -16,6 +16,7 @@ import {
 
 import * as Layout from './layout';
 import {sendAndConfirmTransaction} from './util/send-and-confirm-transaction';
+import {loadAccount} from './util/account';
 
 /**
  * Some amount of tokens
@@ -61,6 +62,8 @@ type TokenSwapInfo = {|
    */
   nonce: number,
 
+  tokenProgramId: PublicKey,
+
   /**
    * Token A. The Liquidity token is issued against this value.
    */
@@ -100,6 +103,7 @@ export const TokenSwapLayout: typeof BufferLayout.Structure = BufferLayout.struc
   [
     BufferLayout.u8('isInitialized'),
     BufferLayout.u8('nonce'),
+    Layout.publicKey('tokenProgramId'),
     Layout.publicKey('tokenAccountA'),
     Layout.publicKey('tokenAccountB'),
     Layout.publicKey('tokenPool'),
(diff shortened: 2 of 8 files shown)


token-swap/js/client/token-swap.js
```diff
@@ -328,23 +380,16 @@ export class TokenSwap {
   }
 
   /**
-   * Swap the tokens in the pool
+   * Swap token A for token B
```
joncinque, line 383: The next step of token-swap is to expand to pools, so the function will need to take in the accounts of the swap no matter what -- meaning this can keep the same signature as before, minus the `authority` and `tokenProgramId`.  Going along with that, you can get rid of the second function `swapInverse`

token-swap/js/client/token-swap.js
```diff
@@ -328,23 +380,16 @@ export class TokenSwap {
...
-   * @param authority Authority
-   * @param source Source account
-   * @param swapSource Base account to swap into, must be a source token
-   * @param swapDestination Base account to swap from, must be a destination token
+   * @param source Source token account
    * @param destination Destination token account
-   * @param tokenProgramId Token program id
-   * @param amount Amount to transfer from source account
+   * @param amountIn Amount to transfer from source account
+   * @param minimumAmountOut Minimum amount to send
```
joncinque, line 388: This param should be "Minimum amount to receive"

joncinque (commented):
Looks good for the most part, just the one part about the `swap` function which needs to be updated.  Otherwise, it looks like this failed the lint test, so be sure to run `npm run lint:fix` as well.

joncinque:
And also, could you update the signatures in `module.flow.js` and `module.d.ts`?

token-swap/js/client/token-swap.js
rawfalafel: Isn't `amountIn` the amount the program's receiving, and `minimumAmountOut` the amount the program's sending?

token-swap/js/client/token-swap.js
rawfalafel: Gotcha. Since devs can fetch the token mint accounts with `getSwapInfo`, requiring them to pass in the mint accounts here seems more reasonable.

Cargo.lock
```diff
@@ -561,26 +561,26 @@ dependencies = [
 [[package]]
 name = "curve25519-dalek"
 version = "2.1.0"
-source = "git+https://github.com/garious/curve25519-dalek?rev=60efef3553d6bf3d7f3b09b5f97acd54d72529ff#60efef3553d6bf3d7f3b09b5f97acd54d72529ff"
```
rawfalafel, line 564: I couldn't figure out what was causing this file to automatically change for me. Any idea?

rawfalafel:
@joncinque this PR now changes the program's state to include `tokenProgramId`. In the process, I added a `loadTokenSwap` static method and removed the `TokenSwapInfo` struct entirely. Let me know if it makes sense to you!

token-swap/js/client/token-swap.js
joncinque: Ah, my mistake, I misunderstood.  Yes, `minimumAmountOut` is how much the program sends out, but I thought the comment implied how much the user was sending in.  Since this `token-swap` js package is user-facing, the parameters should be framed from the point of view of the user, and not the program.  The user is sending in `amountIn`, and receiving `minimumAmountOut`.  Does that make sense?

token-swap/js/client/token-swap.js
```diff
@@ -342,54 +359,21 @@ export class TokenSwap {
...
-
   /**
    * Swap token A for token B
    *
-   * @param source Source token account
-   * @param destination Destination token account
+   * @param userSource User's source token account
+   * @param poolSource Pool's source token account
+   * @param userSource User's destination token account
+   * @param poolSource Pool's destination token account
```
joncinque, line 368: Comments here reference the wrong variables `userSource` -> `userDestination` and `poolSource` -> `poolDestination`

token-swap/js/client/token-swap.js
```diff
@@ -342,54 +359,21 @@ export class TokenSwap {
...
    * @param amountIn Amount to transfer from source account
    * @param minimumAmountOut Minimum amount to send
    */
   async swap(
-    source: PublicKey,
-    destination: PublicKey,
+    userSource: PublicKey,
+    poolSource: PublicKey,
+    userDestination: PublicKey,
+    poolDestination: PublicKey,
```
joncinque, line 376: Could you invert these two parameters to match `swapInstruction`?

token-swap/js/client/token-swap.js
```diff
@@ -452,10 +401,10 @@ export class TokenSwap {
...
     tokenSwap: PublicKey,
     authority: PublicKey,
-    source: PublicKey,
-    swapSource: PublicKey,
-    swapDestination: PublicKey,
-    destination: PublicKey,
+    userSource: PublicKey,
+    poolSource: PublicKey,
+    userDestination: PublicKey,
+    poolDestination: PublicKey,
```
joncinque, line 407: same here

token-swap/js/client/token-swap.js
```diff
@@ -480,10 +429,10 @@ export class TokenSwap {
...
       {pubkey: tokenSwap, isSigner: false, isWritable: false},
       {pubkey: authority, isSigner: false, isWritable: false},
-      {pubkey: source, isSigner: false, isWritable: true},
-      {pubkey: swapSource, isSigner: false, isWritable: true},
-      {pubkey: swapDestination, isSigner: false, isWritable: true},
-      {pubkey: destination, isSigner: false, isWritable: true},
+      {pubkey: userSource, isSigner: false, isWritable: true},
+      {pubkey: poolSource, isSigner: false, isWritable: true},
+      {pubkey: userDestination, isSigner: false, isWritable: true},
+      {pubkey: poolDestination, isSigner: false, isWritable: true},
```
joncinque, line 435: last here

token-swap/js/module.d.ts
```diff
@@ -16,23 +16,20 @@ declare module '@solana/spl-token-swap' {
     static fromBuffer(buffer: Buffer): Numberu64;
   }
 
-  export type TokenSwapInfo = {
```
joncinque, line 19: nice job removing this!

token-swap/js/module.d.ts
```diff
@@ -70,26 +74,22 @@ declare module '@solana/spl-token-swap' {
...
-      authority: PublicKey,
-      source: PublicKey,
-      swapSource: PublicKey,
-      swapDestination: PublicKey,
-      destination: PublicKey,
-      tokenProgramId: PublicKey,
+      userSource: PublicKey,
+      poolSource: PublicKey,
+      userDestination: PublicKey,
+      poolDestination: PublicKey,
```
joncinque, line 81: these should also be put back as they were

token-swap/js/module.flow.js
```diff
@@ -66,26 +70,22 @@ declare module '@solana/spl-token-swap' {
...
-      authority: PublicKey,
-      source: PublicKey,
-      swapSource: PublicKey,
-      swapDestination: PublicKey,
-      destination: PublicKey,
-      tokenProgramId: PublicKey,
+      userSource: PublicKey,
+      poolSource: PublicKey,
+      userDestination: PublicKey,
+      poolDestination: PublicKey,
```
joncinque, line 77: same here

token-swap/js/module.flow.js
```diff
@@ -66,26 +70,22 @@ declare module '@solana/spl-token-swap' {
...
       tokenSwap: PublicKey,
       authority: PublicKey,
-      source: PublicKey,
-      swapSource: PublicKey,
-      swapDestination: PublicKey,
-      destination: PublicKey,
+      userSource: PublicKey,
+      poolSource: PublicKey,
+      userDestination: PublicKey,
+      poolDestination: PublicKey,
```
joncinque, line 88: same here

joncinque (commented):
just a few more small things, then this should be good to go!

token-swap/js/client/token-swap.js
rawfalafel: yep!

rawfalafel:
good catch on the parameters names. this should be ready for another look.

joncinque:
Looks great!  It looks like CI didn't get triggered, so I'm going to close / re-open this to see if that gets it going

joncinque:
All looks good -- thanks for the contribution

PR #594, opened 2020-10-09
Standardise amount datatypes in token client

dankelleher:
Standardise datatypes of amount parameters to accept u64s across the board in the token program JS client.


joncinque:
Looks great, nice catch!

PR #595, opened 2020-10-09
token-swap: Add checks for 0 token output and tests

joncinque:
**Problem**

Pointed out by @dankelleher -- if working with very small amounts of pool tokens, we can get in situations where we deposit / withdraw / swap 0 tokens, which is a big problem!

**Solution**

Return `None` from the calculations if 0 is invalid, triggering `CalculationFailure`.  The other approach is to add checks wherever these functions are used, but I thought it would be clearer for `None` to mean that the calculation has no good result.


CriesofCarrots (approved):
Nice catch

Issue #398, opened 2020-09-08
token-swap: Incorporate more invariants / curves, e.g. curve.fi


joncinque:
**Problem**

The curve defined by Uniswap: `x * y = k` is useful for many applications, but can be difficult to work with in the stablecoin space, where slippage needs to be minimal.  For example, if trading USDC for DAI, which are both USD stablecoins, you can have 4% slippage because of harsh curve movements.

**Solution**

Allow for different curves to be set for the token swap.  For example, curve.fi uses a special kind of "varying leverage" curve.  More can be found at their whitepaper: https://www.curve.fi/stableswap-paper.pdf  -- they suggest an invariant of the form:
```
A * n ^ n * sum(x_i) + D = A * D * n ^ n + (D ^ (n + 1)) / (n ^ n * product(x_i))
```
where n is the number of currencies in the pool, D is the sum of all coins, x_i is an individual coin amount, and A is some constant.  This provides for dynamic leverage while calculating the invariant -- it essentially flattens the function when balances are close to equal, and then creates huge prices movements at the tails.

Their "optimal" recommendation is A = 85, fee = 0.06%, providing 312% APR for liquidity providers.

joncinque:
Curve variants have been added, but the stable curve implementation is not totally there, so I'll leave this issue open

rawfalafel:
@joncinque which parts are in complete?

joncinque:
@rawfalafel the withdrawal / deposit calculation doesn't line up perfectly with stable swap, so it could be possible to leak value out in a fee-less environment using combinations of deposit on one side and swap.  That's why the stable curve does not have the same proptests as the other curves

joncinque:
Closing this in favor of more specific issues, since the stable curve has been included for awhile

PR #624, opened 2020-10-14
token-swap: Add curve and pool token trait and integrate in processor

joncinque:
To easily move between different implementations of swap curves as mentioned in #398, we define a SwapCurve through a trait.

Based on discussions with other AMM providers, this will definitely need to be enhanced for deposit and withdraw, but this gives us a baseline to work from.

Additionally, we will likely need to expand the curve creator to take parameters which would vary depending on the curve type.  For example, as mentioned in #398, the StableSwap curve has an input of A, the "amp" coefficient.  We can save that for another round of changes though.

WIP: need to integrate the curve type as a field in `SwapInfo` and the `Initialize` instruction then add tests and JS support


token-swap/program/src/curve.rs
```diff
@@ -1,11 +1,93 @@
...
+}
+
+impl CurveType {
+    /// Create a swap curve corresponding to the enum
+    pub fn swap_curve(
+        &self,
+        swap_source_amount: u64,
+        swap_destination_amount: u64,
+        fee_numerator: u64,
+        fee_denominator: u64,
```
CriesofCarrots, line 47: Can you explain more how this will work for different curves with different inputs? My initial take is that this doesn't seem like the right abstraction, if the inputs will vary.

token-swap/program/src/processor.rs
```diff
@@ -262,28 +259,14 @@ impl Processor {
...
-                token_b: source_account.amount,
-                fee_numerator: token_swap.fee_numerator,
-                fee_denominator: token_swap.fee_denominator,
-            };
-            invariant
-                .swap_b_to_a(amount_in)
-                .ok_or(SwapError::CalculationFailure)?
-        };
-        if amount_out < minimum_amount_out {
+        let curve = token_swap.curve_type.swap_curve(
```
CriesofCarrots, line 262: I have been trying to think of a way to avoid so much dynamic dispatch, but I do like how simple this section becomes.

token-swap/js/client/token-swap.js
```diff
@@ -123,6 +124,13 @@ export class TokenSwap {
    */
   feeDenominator: Numberu64;
 
+  /**
+   * Curve type, current options are:
+   * 0 = Constant product curve, Uniswap-style
+   * 1 = Flat curve, always 1:1 trades
```
CriesofCarrots, line 130: Make an "enum" type for this, perhaps?

CriesofCarrots (commented):
This seems pretty close! Just a couple questions.

token-swap/js/client/token-swap.js
joncinque: Good idea -- I'm never sure of the best way to do JS enums though.  Do you mean something like this?
```js
const CurveType = {
    ConstantProduct: 0,
    Flat: 1
};
```

token-swap/program/src/processor.rs
joncinque: I wasn't sure how to avoid it either.  We need to encode it in the state of the swap account somehow, so this seemed like the smallest way to do it.  Another thought could be to have the curve trait implement `Pack` so that we can encode the whole thing into the swap account.  Or, all of this could be delegated to some SwapCurve program which just does the calculations.  Does the dynamic dispatch bloat the instructions a lot?

token-swap/program/src/curve.rs
joncinque: You're totally right, I wasn't sure how to cover many different types of curves, especially if the curve itself requires instantiation parameters.  Just the `swap_source_amount` and `swap_destination_amount` should be the main inputs, and everything else is encoded in the curve, including fees and other parameters that might be needed.

Here's what I propose: 
* `SwapInfo` includes some extra space (64 bytes?) for the curve parameters, also passed on init
* `SwapCurve` implements `Pack`
* include `deposit` and `withdraw` in the `SwapCurve` interface -- for example, some extra parameters are needed for stable swap.
* the call to `swap` passes in `swap_source_amount` and `swap_destination_amount`

what do you think?

token-swap/program/src/processor.rs
CriesofCarrots: Based on my understanding, it shouldn't have much of an impact on the instruction count, but will be a bit slower at runtime. I think the flexibility and simplicity is rather worth giving it a go, though. We can rework if performance benchmarks aren't where we'd like them.

token-swap/js/client/token-swap.js
CriesofCarrots: Yep, that's what I was thinking, except:
```javascript
const CurveType = Object.freeze({
    ConstantProduct: 0,
    Flat: 1
});
```

token-swap/program/src/processor.rs
joncinque: Ah I see, that is definitely something to keep in mind

token-swap/program/src/curve.rs
CriesofCarrots: The `Pack` implementation is to easily go from `SwapCurve` to the parameters stored in `SwapInfo` and back? If I've got that right, this sounds pretty good to me!

joncinque: Exactly, I'll do that then!

joncinque:
This gets a little tricky because `Pack` wasn't meant to be used with dynamic dispatch.  I'll have to look into a bit more because it always trips me up, but we might want to relax the `Sized` requirement to `?Sized` -- something doesn't necessarily need a fixed size to be packed.  Or we can keep it the way it is, since trait objects make things tricky.  Anyway, this is ready now!

CriesofCarrots (approved):

PR #636, opened 2020-10-15
Upgrade solana-sdk in themis

garious:


garious:
@jackcmay, token CLI is bound to a specific version of `solana-sdk`, which made it so I couldn't upgrade the themis dependency independently. Any way we can make it more flexible?

jackcmay:
Probably doesn't need to be pinned

mvines:
We pinned to avoid random `cargo install spl-token-cli` bustage as we ship new solana crates that are not source compatible with older versions.   But now that we ship `spl-token-cli` with the main solana release artifacts we'll detect issues due to lack of pinning much quicker

PR #647, opened 2020-10-17
Changes to stake pool program address generation

ysavchenko:
After a discussion with @jackcmay we decided to:
- Use different seeds to generate deposit and withdraw authorities (added "deposit" and "withdraw" prefixes)
- Rename nonce used for program address generation to 'bump seed'

### DIFF
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -26,10 +26,10 @@ pub struct InitArgs {
     /// Fee paid to the owner in pool tokens
     pub fee: Fee,
     /// Nonce used for the deposit program address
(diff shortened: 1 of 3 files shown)


joncinque:
This looks good, and it's a great first experience with the CI.  If you look through the ci script https://github.com/solana-labs/solana-program-library/blob/master/ci/script.sh#L36, you'll see that we're also running `cargo fmt` and `cargo clippy` along with building and testing each program.  This is currently failing CI because of fmt issues.  Be sure to run those tools and then commit the updates, and once it passes CI, we'll integrate!

joncinque:
Looks good, thanks!

PR #648, opened 2020-10-19
Set correct read/write permissions on token-swap instruction calls

rawfalafel:
As @joncinque already mentioned, this will only become relevant for cross-program invocations.


joncinque:
Looks great, thanks for the work

PR #649, opened 2020-10-19
Update SDK version to v1.3.17

jstarry:


jstarry:
Main intended change was to update the `testnetDefaultChannel` so that when web3.js is updated, base64 transaction sending will work properly

PR #665, opened 2020-10-19
Stake pool improvements and fixes

ysavchenko:
- Fixed invalid program address derivation in mint and burn commands
- Moved dump seeds generation into contract initialization, removed from params
- Added claim method (to claim whole address without splitting it)
- Added constant for stake pool storage length (StakePool::LEN)
- Added check (and error message) for fee > 1.0


stake-pool/program/src/processor.rs
```diff
@@ -40,57 +46,52 @@ impl Processor {
...
 
     /// Issue a stake_split instruction.
     pub fn stake_split<'a>(
-        stake_pool: &Pubkey,
         stake_account: AccountInfo<'a>,
         authority: AccountInfo<'a>,
-        nonce: u8,
         amount: u64,
         split_stake: AccountInfo<'a>,
+        signers: &[&[&[u8]]],
```
joncinque, line 66: Note that this is a problem with signers in general, but the `signers` parameter for these functions is confusing -- it's unclear how to construct it.  Instead of passing signers in here, could you pass all of the pubkeys and bump seed required and then construct it in this function?  That way the caller doesn't worry about how to create `signers`

stake-pool/program/src/processor.rs
```diff
@@ -40,57 +46,52 @@ impl Processor {
...
 
     /// Issue a stake_set_owner instruction.
     pub fn stake_authorize<'a>(
-        stake_pool: &Pubkey,
         stake_account: AccountInfo<'a>,
         authority: AccountInfo<'a>,
-        nonce: u8,
         new_staker: &Pubkey,
         staker_auth: stake::StakeAuthorize,
+        signers: &[&[&[u8]]],
```
joncinque, line 79: Same here

stake-pool/program/src/processor.rs
```diff
@@ -40,57 +46,52 @@ impl Processor {
...
     /// Issue a spl_token `Burn` instruction.
     pub fn token_burn<'a>(
-        stake_pool: &Pubkey,
         token_program: AccountInfo<'a>,
         burn_account: AccountInfo<'a>,
         mint: AccountInfo<'a>,
         authority: AccountInfo<'a>,
-        nonce: u8,
         amount: u64,
+        signers: &[&[&[u8]]],
```
joncinque, line 93: Same here

stake-pool/program/src/processor.rs
```diff
@@ -109,17 +110,13 @@ impl Processor {
...
     /// Issue a spl_token `MintTo` instruction.
     pub fn token_mint_to<'a>(
-        stake_pool: &Pubkey,
         token_program: AccountInfo<'a>,
         mint: AccountInfo<'a>,
         destination: AccountInfo<'a>,
         authority: AccountInfo<'a>,
-        nonce: u8,
         amount: u64,
+        signers: &[&[&[u8]]],
```
joncinque, line 118: Same here

stake-pool/program/src/instruction.rs
```diff
@@ -71,6 +66,18 @@ pub enum StakePoolInstruction {
...
     Withdraw(u64),
 
+    ///   Claim ownership of a whole stake account.
+    ///   Also burns enough tokens to make up for the stake account balance
+    ///   
+    ///   0. `[w]` StakePool
+    ///   1. `[]` withdraw  authority
+    ///   2. `[w]` Stake account, its withdraw authority will be changed to the destination account
+    ///   3. `[w]` Pool MINT account, authority is the owner
+    ///   6. `[]` New stake owner account
```
joncinque, line 76: Nit: the numbering here is off

stake-pool/program/src/instruction.rs
```diff
@@ -71,6 +66,18 @@ pub enum StakePoolInstruction {
...
+    ///   7. `[]` Token program id
+    ///   userdata: amount to withdraw
```
joncinque, line 78: Nit: there isn't any userdata in this instruction

stake-pool/program/src/processor.rs
```diff
@@ -291,32 +313,99 @@ impl Processor {
...
+        stake_pool.stake_total -= stake_amount;
+        State::Init(stake_pool).serialize(&mut stake_pool_info.data.borrow_mut())?;
+        Ok(())
+    }
+    /// Processes an [Claim](enum.Instruction.html).
+    pub fn process_claim(program_id: &Pubkey, accounts: &[AccountInfo]) -> ProgramResult {
+        let account_info_iter = &mut accounts.iter();
+        let stake_pool_info = next_account_info(account_info_iter)?;
+        let withdraw_info = next_account_info(account_info_iter)?;
+        let source_info = next_account_info(account_info_iter)?;
```
joncinque, line 359: You'll need an additional parameter of the user's pool token account to burn from.  `source_info` is currently referenced as both the stake account to reassign *and also* the pool token account to burn from

stake-pool/program/src/processor.rs
```diff
@@ -291,32 +313,99 @@ impl Processor {
...
+        let withdraw_bump_seed = [stake_pool.withdraw_bump_seed];
+        let stake_pool_key_bytes = stake_pool_info.key.to_bytes();
+        let withdraw_signers = prepare_signers!(
+            stake_pool_key_bytes,
+            Self::AUTHORITY_WITHDRAW,
+            withdraw_bump_seed
+        );
+
+        Self::stake_authorize(
+            source_info.clone(),
```
joncinque, line 395: For reference, `source_info` used here once

stake-pool/program/src/processor.rs
```diff
@@ -291,32 +313,99 @@ impl Processor {
...
+            withdraw_info.clone(),
+            dest_user_info.key,
+            stake::StakeAuthorize::Withdrawer,
+            withdraw_signers,
+        )?;
+
+        Self::token_burn(
+            token_program_info.clone(),
+            source_info.clone(),
```
joncinque, line 404: `source_info` referenced here again

stake-pool/program/src/state.rs
```diff
@@ -67,6 +67,8 @@ pub enum State {
 }
 
 impl State {
+    /// Length of state data when serialized
+    pub const LEN: usize = size_of::<u8>() + size_of::<StakePool>();
```
joncinque, line 71: Why is the extra `u8` needed here?

joncinque (commented):
There's one main question here, otherwise a lot of smaller things.  The instructions definitely need some tests soon, but it can be done in a separate PR.

stake-pool/program/src/processor.rs
ysavchenko: Previously it was exactly like you describe, however with new params for constructing program addresses (AUTHORITY_DEPOSIT or AUTHORITY_WITHDRAW) number of params in mint and burn functions reached 8. And cargo clippy did not like that (7 params per function maximum), this is why I refactored all signers construction code to happen outside of these functions.

Also there is a bonus now: we need to construct signers only once and then pass to each function when we chain several requests in one smart contract method.

If you think this is confusing, please tell me if you have any ideas on how to improve the situation without moving signers construction outside of those functions. Thanks.

stake-pool/program/src/processor.rs
ysavchenko: Yes, this is a bug, I believe there was a similar bug in the original split method, will fix both

stake-pool/program/src/state.rs
ysavchenko: Please check several lines below: the first byte is used to encode account status (initialized or not) and the rest is used for the stake pool status struct.

stake-pool/program/src/processor.rs
joncinque: Ah gotcha, I see why you did it that way.  I think it's ok to reach 8 parameters in this case, so feel free to use `#[allow(clippy::too_many_arguments)]` before the functions.  To me, it'll make the interface simpler to use despite the higher number of arguments.

stake-pool/program/src/state.rs
joncinque: Ah gotcha, I was only looking at the changed code.  Thanks for the clarification

stake-pool/program/src/state.rs
```diff
@@ -67,6 +67,8 @@ pub enum State {
...
     /// Deserializes a byte buffer into a [State](struct.State.html).
     /// TODO efficient unpacking here
     pub fn deserialize(input: &[u8]) -> Result<State, ProgramError> {
```
joncinque, line 74: I can't comment further down, but the deserialization logic is probably still broken on line 81, since `unpack` is still receiving the entire input, and not `input[1..]`

ysavchenko: Right, this was here before me, will fix it.

joncinque:
Look good -- next it'll need some tests to make sure that everything works!  We'll also need to figure out a way to handle the calls to the stake program

PR #672, opened 2020-10-20
Add Token class properties to spl-token library module declaration files

paul-schaaf:
Add Token class properties to spl-token library module declaration files so downstream users can access them without having to bypass type checking


paul-schaaf:
@jstarry

Issue #646, opened 2020-10-16
perf-monitor: build.rs isn't building shared object


joncinque:
**Problem**

In the perf-monitor, build.rs https://github.com/solana-labs/solana-program-library/blob/master/token/perf-monitor/build.rs is currently failing, and the error returned by `success()` is getting swallowed.  The error is `error: unknown codegen option: embed-bitcode`, which is a codegen flag that was added in rust 1.45.  You can update the file to read:
```rs
fn main() {
    println!("cargo:warning=(not a warning) Building SPL Token shared object");
    let success = Command::new(canonicalize("../../do.sh").unwrap())
        .current_dir("../..")
        .arg("build")
        .arg("token/program")
        .status()
        .expect("Failed to build token program")
        .success();
    assert!(success);
}
```
And the command to run is `cargo test --manifest-path=token/perf-monitor/Cargo.toml` to trigger this.  We haven't noticed this because in CI we build all of the shared objects before running the perf-monitor tests.

**Solution**

No idea to be honest.  There may be some strange rustc version mismatch between `./do.sh` and `cargo`, or running `do.sh` under `cargo` passes some additional flags while compiling dependencies.  I'd be very interested to learn about what's going on!

PR #675, opened 2020-10-20
Bail on build script error

jackcmay:
Reports of failures building the shared objects from within a build script.  Unable to reproduce myself and CI has been covering this up since it builds the shared objects first and independently.

Build the test upfront and assert if the build script fails.

Reported in: #646


jackcmay:
Unsetting `CARGO` succeeds on linux, closing this PR, will update to a newer bpf-sdk later that does the unset

PR #676, opened 2020-10-21
Program update proposal

### LINKED ISSUE #482: Add a token-upgrade program
When token v3 ships, there may be v2 tokens that want to upgrade themselves to v3.   This can be done by a fully automated 1:1 swap program.   

This could be a variant of the existing token-swap program, but it might make sense to keep this independent as we might eventually have to support various token migrations: v2 -> v3,  v3 -> v4, v2 -> v4

jackcmay:
Add a proposal on how program updates can work and an example token update example.

Fixes #482

### DIFF
--- a/docs/src/token.md
+++ b/docs/src/token.md
@@ -466,29 +466,34 @@ intermediary.  The migration program will be provided by the exchange to the
 wallet.  The wallet will issue a single instruction to the migration program,
 providing no instruction data but the following accounts:
 
-0. The v2 account to migrate.
-1. The v2 account's authority.  This authority must sign the migration
+0. The Payer for the additional space required by the new v3 account.  v2 and v3
+   accounts are the same size so this isn't actually neccissary, but likely will
+   be needed for most all migrations.
+1. The v2 account to migrate.
+2. The v2 account's authority.  This authority must sign the migration
    transaction and be an authority that can both burn and close the account.
    This authority will become the new v3 account's authority.
 3. The v2 token mint.
 4. The pubkey of the new v3 account that will be created and initialized.  Must
    be a signer to allow for `SystemInstruction::CreateAccount`.
-6. A v3 mint to be used as a source of tokens for the new account. In some
+5. A v3 mint to be used as a source of tokens for the new account. In some
    scenarios a mint may be a fixed supply.  In lieu of a mint, the program could
    be modified to instead take v3 account could be provided that contains
    sufficient tokens for the migration.
-7. The mint's multisignature authority where at least one of the signers is a
-   program address associated with the migration program.  Using a multisignture
-   here allows an external entity to also retain ownership of the mint or source
-   account.
-8. The Rent sysvar
+6. The mint's 1xN multisignature authority where at least one of the signers is
+   the program address associated with the migration program.  Using a
+   multisignture here allows an external entity to also retain ownership of the
+   mint or source account.
+7. The Rent sysvar
+8. A valid signer, must be one of the N multisignture authorities.
 
 The update program will perform the following:
 
 - Query the v2 account's token balance.
-- Issue a `Burn` instruction on the v2 account for the entire balance.
+- Calculate the difference between the old and new account sizes.
 - Issue a `SystemInstruction::CreateAccount` instruction using the new v3
   account's pubkey.
+- Issue a `Burn` instruction on the v2 account for the entire balance.
 - Issue a `CloseAccount` instruction on the v2 account, transferring all
   lamports to the new v3 account.
 - Issue an `InitializeAccount` instruction on the new v3 account.
@@ -499,22 +504,36 @@ The update program will perform the following:
 The migration program's `process_instruction` may look like this:
 
 ```rust
-pub fn process(program_id: &Pubkey, accounts: &[AccountInfo], _input: &[u8]) -> ProgramResult {
+    pub fn process(_program_id: &Pubkey, accounts: &[AccountInfo], input: &[u8]) -> ProgramResult {
+        if input.is_empty() || input[0] != 0 {
+            return Err(ProgramError::InvalidArgument);
+        }
+
         let account_info_iter = &mut accounts.iter();
+        let payer_info = next_account_info(account_info_iter)?;
         let v2_account_info = next_account_info(account_info_iter)?;
         let v2_account_authority_info = next_account_info(account_info_iter)?;
         let v2_mint_info = next_account_info(account_info_iter)?;
         let v3_account_info = next_account_info(account_info_iter)?;
         let v3_mint_info = next_account_info(account_info_iter)?;
         let v3_mint_authority_info = next_account_info(account_info_iter)?;
+        let rent_info = next_account_info(account_info_iter)?;
+        let program_signer_info = next_account_info(account_info_iter)?;
 
         let amount = spl_token::state::Account::unpack(&v2_account_info.data.borrow())?.amount;
 
+        // TODO v2 and v3 Account is same size so this is unnecessary, but maybe
+        // leave it in here as an example for other migrations?
+        let rent = Rent::from_account_info(rent_info)?;
+        let additional_amount_to_fund = rent
+            .minimum_balance(spl_token::state::Account::get_packed_len())
+            .saturating_sub(rent.minimum_balance(spl_token::state::Account::get_packed_len()));
+
         let instruction = system_instruction::create_account(
-            v2_account_info.key,
+            payer_info.key,
             v3_account_info.key,
-            0, // TODO lamports from the v2 account transfered during `CloseAccount` may not be sufficient to cover the space required for a v3 account.
-            spl_token_v3::state::Account::get_packed_len(),
+            additional_amount_to_fund,
+            spl_token_v3::state::Account::get_packed_len() as u64,
             &spl_token_v3::id(),
         );
         invoke(&instruction, accounts)?;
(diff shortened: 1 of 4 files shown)


jackcmay:
@mvines fyi

docs/src/token.md
```diff
@@ -449,3 +450,134 @@ have a balance of zero to be closed.
...
+
+As the SPL Token program evolves the accounts owned by older version may need to
+be migrated to newer versions.  This can be done by proxy via a migration
+program.  For more information on program updates in general, refer to
+[program updates](program-updates.md).
+
+The following is an example of how a wallet may migrate accounts owned by SPL
+Token to accounts owned by SPL Token v3.
+
+Updating the accounts will require a migration program that serves as an
```
mvines, line 464: Here's the instructions that I'm thinking of for the v2 -> v3 token upgrade program.  The minter of the v3 token assigns a batch of v3 tokens to the program with `OpenReserveAccount`.  It may need to add more v3 tokens as the users upgrade.

Each user, or their wallet, uses `Upgrade`.

The minter can reclaim any remaining v3 tokens from the program with `CloseReserveAccount`.

```rust
    /// Create a v3 token account to hold funds used when upgrading v2 token accounts (if
    /// necessary), and transfer `amount` tokens into it.
    ///
    ///   0. `[writeable]` v3 token account program address
    ///   1. `[]` The v3 mint this account will be associated with.
    ///   2. `[writeable,signer]` The system account that will fund the v3 token account
    ///   3. `[]` Rent sysvar
    ///   4. `[writeable]` funding v3 token account program address.  `amount` tokens will be
    ///                    transferred from this account
    ///   5. `[signer]` owner of the v3 token account (TODO: support multisig owners too...)
    ///
    OpenReserveAccount { amount: u64 },

    /// Close the v3 token account held by the program to upgrade v2 token accounts
    ///
    ///   0. `[writeable]` v3 token account program address
    ///   1. `[]` The v3 mint this account is associated with.
    ///   2. `[writeable]` v3 token account address that will receive any remaining tokens in the
    ///                    program's token account
    ///   3. `[writeable]` The system account that will receive the rent lamports after
    ///                    closing the account
    ///
    CloseReserveAccount,

    /// Update the source v2 token account into the destination v3 token account.  All tokens in
    /// the v2 account will be burned, and equivalent tokens will be credited to the v3 account
    /// from the program's v3 token reserve.
    ///
    /// It's the caller's responsibility to close the v2 account after this instruction completes
    /// if they wish.
    ///
    ///   0. `[writeable]` source v2 token account
    ///   1. `[writeable]` destination v3 token account
    ///   2. `[writeable]` v3 token account program address
    ///   3. `[]` The v3 mint this account will be associated with.
    ///   4. `[signer]` owner of v2 token account (TODO: support multisig owners...)
    ///
    Upgrade,
```

docs/src/token.md
```diff
@@ -449,3 +450,134 @@ have a balance of zero to be closed.
 
 ### Non-Fungible tokens
 An NTF is simply a token type where only a single token has been minted.
+
+### Program updates
```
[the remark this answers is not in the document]
jackcmay: Ok, sounds good, yes I have a working upgrade program that I'll post.  Wanted to get some review first before spending that time.

docs/src/token.md
jackcmay: Splitting the signing authorities (wallet, dex) apart into different instructions simplifies things.  It also complicates some things.

Some questions/comments
- Who creates the new v3 account, if done by `Upgrade` then the lamport funding account must also be a signer.
- If account #2 in `OpenReserveAccount` is used to fund the new v3 token account then that's potentially a lot of lamports locked up.  Transfering the lamports from the v2 to v3 account directly in the `upgrade` instruction eliminates that.
- The `upgrade` instruction would be called by the wallet, they may or may not be able to call `CloseAccount` which means the dex would have to monitor what accounts have been upgraded and perform and clean-up.  The dex would also have to actively monitor the reserve account if it did not fully fund it to account for all upgrades.
- Why Account #3 in `Upgrade`?  Wouldn't this be already be known by looking at the mint associated with the funding reserve account?
- `Upgrade` would need the v2 mint account to `Burn`
- If `Upgrade` creates and initializes the v3 token account then account #1 will have to be a signer and `Upgrade` will also need the v3 token account authority account.
- The wallet would be paying the tx fee for the upgrade

I was thinking that the wallet would be given a template upgrade tx that it would fill in, sign, and send to the dex, the dex would then finishing signing and submit to the chain.  The dex could respond to these upgrade requests directly to refund the funding account, payer, etc...  Upgrade would look like this:

```rust
    ///   0. `[signer]` Payer for any additional account space.  The v2
    ///      account's lamports are transferred to the new v3 account but if the
    ///      v3 account is larger then additional lamports will be required.
    ///      This payer provides those additional lamports.
    ///   1. `[writable]` The v2 account to migrate.
    ///   2. `[writable]` The v2 token mint.
    ///   3. '[writable, signer]' The new v3 account's pubkey.
    ///   4. `[writable]` The v3 token mint.
    ///   5. `[writable]` The v3 funding account
    ///   6. `[signer]` The v3 funding account authority
    ///   7. `[]` Rent sysvar
    ///   8. `[signer]` the V2 token signer (or multiple for multisignature)
```

mvines: > * Who creates the new v3 account, if done by `Upgrade` then the lamport funding account must also be a signer.

A prior instruction does it, to avoid baking in assumptions about how it's created.  This 

> * If account #2 in `OpenReserveAccount` is used to fund the new v3 token account then that's potentially a lot of lamports locked up.  Transfering the lamports from the v2 to v3 account directly in the `upgrade` instruction eliminates that.

s/lamports/tokens/ but yes.  But this is a better trade-off than requiring the v3 mint authority sign each `Upgrade`, or requiring the v3 mint authority be assigned to the upgrade program itself.

> * The `upgrade` instruction would be called by the wallet, they may or may not be able to call `CloseAccount` which means the dex would have to monitor what accounts have been upgraded and perform and clean-up.  The dex would also have to actively monitor the reserve account if it did not fully fund it to account for all upgrades.

Yep.   

> * Why Account #3 in `Upgrade`?  Wouldn't this be already be known by looking at the mint associated with the funding reserve account?

The funding reserve account is a program account that's generated from provided v3 mint address.  I was thinking this program could support upgrading accounts from any v2 mint to v3, not just one particular mint.

> * `Upgrade` would need the v2 mint account to `Burn`

Ah, ok.   

> * If `Upgrade` creates and initializes the v3 token account then account #1 will have to be a signer and `Upgrade` will also need the v3 token account authority account.

Na, the v3 token account should already exist before invoking `Update`

> * The wallet would be paying the tx fee for the upgrade

It could.  If there's somebody who's motivated to cause an upgrade, they can pay the fee on behalf of the wallet though (with some coordination):
* 3rd party could create the v3 token account for the user
* 3rd party could fund the transaction with the `Upgrade` instruction in it, once the wallet provides the v2 token owner signature


> ... Upgrade would look like this:

This signature is problematic: `3. '[writable, signer]' The new v3 account's pubkey.`.  Can't assume the v3 account will be on the curve, it might be (and probably will be once #626 finishes) a program address

jackcmay: > > * Who creates the new v3 account, if done by `Upgrade` then the lamport funding account must also be a signer.
> 
> A prior instruction does it, to avoid baking in assumptions about how it's created.

Yeah, I guess it doesn't have to be a brand new v3 account :-)
> 
> > * If account #2 in `OpenReserveAccount` is used to fund the new v3 token account then that's potentially a lot of lamports locked up.  Transfering the lamports from the v2 to v3 account directly in the `upgrade` instruction eliminates that.
> 
> s/lamports/tokens/ but yes. But this is a better trade-off than requiring the v3 mint authority sign each `Upgrade`, or requiring the v3 mint authority be assigned to the upgrade program itself.

I meant lamports working with the idea that `Upgrade` would create the new v3 account, it would also be funded by the reserver account both for tokens and lamports.
> 
> > * The `upgrade` instruction would be called by the wallet, they may or may not be able to call `CloseAccount` which means the dex would have to monitor what accounts have been upgraded and perform and clean-up.  The dex would also have to actively monitor the reserve account if it did not fully fund it to account for all upgrades.
> 
> Yep.
> 
> > * Why Account #3 in `Upgrade`?  Wouldn't this be already be known by looking at the mint associated with the funding reserve account?
> 
> The funding reserve account is a program account that's generated from provided v3 mint address. I was thinking this program could support upgrading accounts from any v2 mint to v3, not just one particular mint.

There would be a 3rd mint besides the one's reference in the src or dst accounts?

> 
> > * `Upgrade` would need the v2 mint account to `Burn`
> 
> Ah, ok.
> 
> > * If `Upgrade` creates and initializes the v3 token account then account #1 will have to be a signer and `Upgrade` will also need the v3 token account authority account.
> 
> Na, the v3 token account should already exist before invoking `Update`
> 
> > * The wallet would be paying the tx fee for the upgrade
> 
> It could. If there's somebody who's motivated to cause an upgrade, they can pay the fee on behalf of the wallet though (with some coordination):
> 
>     * 3rd party could create the v3 token account for the user
> 
>     * 3rd party could fund the transaction with the `Upgrade` instruction in it, once the wallet provides the v2 token owner signature
> 
> 
> > ... Upgrade would look like this:
> 
> This signature is problematic: `3. '[writable, signer]' The new v3 account's pubkey.`. Can't assume the v3 account will be on the curve, it might be (and probably will be once #626 finishes) a program address

If the v3 account is created before the `Upgrade` instruction then this signer is not required.  

I like the idea of a single instruction upgrade, the lamports flowing from the v2 to v3 account, and then the result of the `Upgrade` being a closed v2 account.  But I guess I was trying to fit too much into one instruction 😉

PR #686, opened 2020-10-22
Check for close and freeze authorities in token-swap

jstarry:
#### Problem
- Swap pool mints can have a freeze authority which can be used to freeze liquidity provider's pool tokens to lock in liquidity
- Swap pool liquidity token accounts can have a close authority. If the token accounts reach zero tokens, they can be closed and deleted and replaced with new accounts which aren't restricted by the normal initialization instruction.

#### Changes
- Prohibit token accounts from having a close authority
- Prohibit mint accounts from having a freeze authority

### DIFF
--- a/token-swap/program/src/error.rs
+++ b/token-swap/program/src/error.rs
@@ -58,6 +58,12 @@ pub enum SwapError {
     /// Swap instruction exceeds desired slippage limit
     #[error("Swap instruction exceeds desired slippage limit")]
     ExceededSlippage,
+    /// The provided token account has a close authority.
+    #[error("Token account has a close authority")]
+    InvalidCloseAuthority,
+    /// The pool token mint has a freeze authority.
+    #[error("Pool token mint has a freeze authority")]
+    InvalidFreezeAuthority,
 }
 impl From<SwapError> for ProgramError {
     fn from(e: SwapError) -> Self {
--- a/token-swap/program/src/processor.rs
+++ b/token-swap/program/src/processor.rs
@@ -173,6 +173,7 @@ impl Processor {
         if COption::Some(*authority_info.key) != pool_mint.mint_authority {
             return Err(SwapError::InvalidOwner.into());
         }
+
         if token_a.mint == token_b.mint {
             return Err(SwapError::RepeatedMint.into());
         }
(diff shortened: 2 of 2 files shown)


jstarry:
@joncinque can you confirm whether token accounts can reach `amount == 0`?

joncinque:
@jstarry the liquidity token accounts can reach amount == 0 if all of the pool tokens are withdrawn, effectively killing the swap.  Are you thinking of a check to make sure that doesn't happen?  Or do you intend a possibility of depositing when liquidity accounts are at 0?

joncinque (approved):
Great catch and test coverage -- thanks!

jstarry:
> @jstarry the liquidity token accounts can reach amount == 0 if all of the pool tokens are withdrawn. 

Interesting, I just wanted to check that checking `CloseAuthority` is actually needed. Having a close authority in that case doesn't actually seem like a bad thing. I was actually asking if token A could be depleted to 0 if swaps go in one direction, because then the swap would still be active and closing a token account could have bad consequences

> effectively killing the swap

How is it effectively killed? Is it not possible to add liquidity again if all pool tokens are withdrawn?

> Are you thinking of a check to make sure that doesn't happen? Or do you intend a possibility of depositing when liquidity accounts are at 0?

Wasn't thinking about that actually. Seems valid to withdraw all liquidity

joncinque:
>  I was actually asking if token A could be depleted to 0 if swaps go in one direction, because then the swap would still be active and closing a token account could have bad consequences

That depends on the curve being used to calculate transactions.  In most curves, it isn't possible to deplete one of the accounts through swapping since the price tends to infinity, but we should account for the possibility in our code in case someone wants to do it.  For example, a "curve" which is just a line could have a token account go to 0, so we probably want the ability to avoid closing accounts.

> How is it effectively killed? Is it not possible to add liquidity again if all pool tokens are withdrawn?

Currently, no, since all new deposits are weighted as a fraction of the total pool token supply.  If there's no pool token supply, the calculation will fail, and it's impossible to do anything with the swap.  A possibility for "resurrecting" a depleted swap is to mint the initial pool amount on the next deposit though.  It's equivalent to recreating the swap.  It deserves an issue to discuss futher either way :-)

jstarry:
Makes sense, thanks for the explanation!

PR #695, opened 2020-10-22
token-swap: Add fee account to receive withdraw / trading fees, trading token mints

joncinque:
The current fee model only provides fees to pool token holders, but most other DeFi protocols also include an additional fee on trades, like Uniswap, or on pool token withdrawals, like Balancer.

This adds the ability to specify trade and withdraw fees to some other pool token account, along with the necessary calculations.

As an added bonus, @bartosz-lipinski asked for the mints to be directly present in the swap account for ease of fetching data.

Most of this is straightforward adding of fields, and most of the math uses calculations that already exist.

The tricky portion is how to assess fees on trades in pool tokens, since trades are done using only the trading tokens.  The general formula is:

fee_in_pool_tokens = fee_in_trading_tokens * pool_supply / trading_tokens / tokens_in_pool

that last `tokens_in_pool` is important because otherwise we'll give too much in fees.  `fee_in_pool_tokens` would be redeemable for `fee_in_trading_tokens` and also the fee for the other trading token.


token-swap/js/cli/token-swap-test.js
```diff
@@ -37,14 +38,25 @@ let tokenAccountB: PublicKey;
 // curve type used to calculate swaps and deposits
 const CURVE_TYPE = CurveType.ConstantProduct;
 // Initial amount in each swap token
-const BASE_AMOUNT = 1000;
+let currentSwapTokenA = 1000;
```
joncinque, line 41: There's a lot of state sharing in these tests, which has gotten more confusing with fees

token-swap/program/src/curve.rs
```diff
@@ -50,6 +50,16 @@ impl Default for SwapCurve {
     }
 }
 
+/// Clone takes advantage of pack / unpack to get around the difficulty of
```
joncinque, line 53: Not sure if this is too hacky -- it's mainly used for testing purposes

token-swap/program/src/curve.rs
```diff
@@ -134,18 +144,47 @@ pub trait CurveCalculator: Debug + DynPack {
...
+    }
 
-    /// Get the amount of liquidity tokens for pool tokens given the total amount
-    /// of liquidity tokens in the pool
-    fn liquidity_tokens(
+    /// Get the amount of trading tokens for the given amount of pool tokens,
+    /// provided the total trading tokens and supply of pool tokens.
+    /// The default implementation is a simple ratio calculation for how many
+    /// trading tokens correspond to a certain number of pool tokens
+    fn pool_tokens_to_trading_tokens(
```
joncinque, line 177: Justin and I were discussing the terminology to use here, since "liquidity token" is ambiguous.  I've switched to "trading token" in this PR

token-swap/program/src/instruction.rs
```diff
@@ -83,6 +88,7 @@ pub enum SwapInstruction {
     ///   6. `[writable]` token_a user Account to credit.
     ///   7. `[writable]` token_b user Account to credit.
     ///   8. '[]` Token program id
+    ///   9. `[writable]` Fee account, to receive withdrawal fees
```
joncinque, line 91: This parameter is last per Bartosz's request

token-swap/program/src/instruction.rs
```diff
@@ -203,25 +209,11 @@ pub fn initialize(
     token_a_pubkey: &Pubkey,
     token_b_pubkey: &Pubkey,
     pool_pubkey: &Pubkey,
+    fee_pubkey: &Pubkey,
     destination_pubkey: &Pubkey,
     nonce: u8,
-    curve_type: CurveType,
-    fee_numerator: u64,
-    fee_denominator: u64,
+    swap_curve: SwapCurve,
```
joncinque, line 215: This makes things *a lot* neater!

token-swap/program/src/curve.rs
CriesofCarrots: Is it used exclusively in tests? If so, perhaps rename or at least comment to that effect.

token-swap/program/src/instruction.rs
CriesofCarrots: Any particular reason to not have `Token program id` last, as in the other instructions?
Edit: noticed that token program id is actually not last in swap, just in swap docs. I still rather like program id last.

token-swap/program/src/instruction.rs
```diff
@@ -42,7 +45,9 @@ pub enum SwapInstruction {
     ///   3. `[writable]` token_(A|B) Base Account to swap INTO.  Must be the SOURCE token.
     ///   4. `[writable]` token_(A|B) Base Account to swap FROM.  Must be the DESTINATION token.
     ///   5. `[writable]` token_(A|B) DESTINATION Account assigned to USER as the owner.
-    ///   6. '[]` Token program id
+    ///   6. `[writable]` Pool token mint, to generate trading fees
+    ///   7. `[writable]` Fee account, to receive trading fees
+    ///   8. '[]` Token program id
```
CriesofCarrots, line 50: `process_swap` actually expects this order: 6. Token program id; 7. Pool token mint; 8. Fee account
Either fixup docs or implementation. Fwiw, I kind of like Token program id last, but your call.

CriesofCarrots (commented):
Seems pretty close; couple small things.
I like the renaming you included; brings a lot of clarity!

Have you run this by downstream users to ensure it meets their needs?

token-swap/program/src/curve.rs
joncinque: I'll put in a comment and `#[cfg(test)]` to make it as clear as possible

token-swap/program/src/instruction.rs
joncinque: Great catch -- @bartosz-lipinski asked for them at the end for his integration, but I think he ended up using the JS package that we include.  I do also prefer the token program id at the end.  @bartosz-lipinski do you think we could switch them back?

joncinque:
I've been working closely with Bartosz to make sure it works with his frontend, but we should probably have a demo with Serum to be sure it's what they expect.

joncinque:
I've put in a commit to update the order to get CI working on it in case we want to go that way.  I can always revert if preferred

CriesofCarrots (approved):
Nice, looks good from my end!

PR #696, opened 2020-10-22
Introduce SPL Token Lending

jstarry:
SPL Token Lending: an Aave-inspired token lending program

Reference Doc: https://docs.google.com/document/d/1wp70o3fehQfRqP2FcR-Ek5vCLpI0pO_ayu-NQd4WK74/edit

Todo:
- [x] Update reference doc
- [x] Clean up tests
- [x] Tokenized obligations

Implemented instructions:
- [x] InitPool
- [x] InitReserve
- [x] Deposit
- [x] Withdraw
- [x] Repay
- [x] Liquidate

Implemented states:
- [x] Pool
- [x] Reserve
- [x] Obligation


aeyakovenko:
Tag @bartosz-lipinski @joncinque 

Can we implement the liqdation as a token-swap compatible pool?  Basically automatically offer the collateral for sale at the liquidation oracle price

joncinque:
great coincidence @aeyakovenko -- @bartosz-lipinski had the same idea, and the three of us had an initial discussion about it last week.  here's a proposed diagram of what that could look like: https://github.com/joncinque/solana-program-library/blob/lending-diagram/token-lending/spl-lending.svg

aeyakovenko:
Curve implementation that uses serum as the oracle. So the swap is always at the spread + slippage

jstarry:
> WVAP for the last hour?

that feels like a long time.. that means that prices need to be recorded continuously (let's say once every minute) for any markets that the lending protocol depends on, right? I feel like 5min is enough time for arbitrage bots to kick in and is short enough that users record prices for a few minutes to liquidate if they needed to take action

aeyakovenko:
@jstarry VWAP for the last hour can be updated on every tick, so it's instant.

PR #711, opened 2020-10-23
Use u128 for all the math but store in u64

aeyakovenko:


token-swap/program/src/processor.rs
```diff
@@ -672,6 +676,15 @@ impl PrintProgramError for SwapError {
     }
 }
 
+fn to_u128(val: u64) -> Result<u128, SwapError> {
+    val.try_into().map_err(|_| SwapError::CalculationFailure)
```
joncinque, line 680: Is it ever possible for this to fail?

token-swap/program/src/processor.rs
```diff
@@ -672,6 +676,15 @@ impl PrintProgramError for SwapError {
...
+}
+
+fn to_u64(val: u128) -> Result<u64, SwapError> {
+    val.try_into().map_err(|_| SwapError::CalculationFailure)
```
joncinque, line 684: A different error like `OverflowError` would make things clearer.  Could anything else make this fail?

joncinque (commented):
Looks great, just a couple of questions

token-swap/program/src/processor.rs
aeyakovenko: Probably not, the compiler should figure it out though.

aeyakovenko: id rather not "unwrap" or throw a random exception since the exception makes it hard to reason about all the exit conditions.

PR #718, opened 2020-10-25
Unit tests refactoring and success deposit test added to stake pool program

ysavchenko:
Before this commit stake pool had only a simple initialization test for the stake pool program.

This PR adds checking newly created stake pool state after initialization and, after some refactoring, uses this state to test successful deposit command.

For now tests ignore stake account instructions processing (they always succeed), this will be improved once there is a direct interface for the stake account program processor.

Please check if everything is ok with such test organization and we'll continue using the similar structure for the remaining tests.


stake-pool/program/src/processor.rs
```diff
@@ -560,79 +561,198 @@ impl PrintProgramError for Error {
...
+                            }
+                        }
+                        new_account_infos.push(new_account_info);
+                    }
+                }
+            }
+
+            match instruction.program_id {
+                TOKEN_PROGRAM_ID => invoke_token(&new_account_infos, &instruction.data),
+                pubkey => {
```
joncinque, line 618: Could you have this match arm work like the token one? then you could have something like:
```rust
let stake_id = stake_program_id();
match instruction.program_id {
    TOKEN_PROGRAM_ID => ...,
    stake_id => ...,
    _ => Err
}
```

stake-pool/program/src/processor.rs
```diff
@@ -560,79 +561,198 @@ impl PrintProgramError for Error {
...
+    }
+
+    /// Mocks token instruction invocation
+    pub fn invoke_token<'a>(account_infos: &[AccountInfo<'a>], input: &[u8]) -> ProgramResult {
+        spl_token::processor::Processor::process(&TOKEN_PROGRAM_ID, &account_infos, &input)
+    }
+
+    /// Mocks stake account instruction invocation
+    pub fn invoke_stake<'a>(_account_infos: &[AccountInfo<'a>], _input: &[u8]) -> ProgramResult {
+        // For now always return ok
```
joncinque, line 636: we'll have this working for you soon!

joncinque (approved):
Just one small suggestion, which you can take or leave.  Looks good!

stake-pool/program/src/processor.rs
ysavchenko: Stake id is not a constant, so I could not really do this in the same way as with token program id, may be it will be better after direct access to stake program is added for testing (the one promised by mvines)

PR #719, opened 2020-10-26
Prevent swap accounts being used as user accounts

jstarry:
#### Problem
No validation check to prevent swap info token accounts being used as user inputs

#### Changes
- Prevent pool accounts from being used as swap inputs, deposit inputs, and withdraw outputs

### DIFF
--- a/token-swap/program/src/processor.rs
+++ b/token-swap/program/src/processor.rs
@@ -266,6 +266,12 @@ impl Processor {
         if *swap_source_info.key == *swap_destination_info.key {
             return Err(SwapError::InvalidInput.into());
         }
+        if swap_source_info.key == source_info.key {
+            return Err(SwapError::InvalidInput.into());
+        }
+        if swap_destination_info.key == destination_info.key {
+            return Err(SwapError::InvalidInput.into());
+        }
         if *pool_mint_info.key != token_swap.pool_mint {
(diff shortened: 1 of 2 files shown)


token-swap/program/src/processor.rs
```diff
@@ -266,6 +266,12 @@ impl Processor {
         if *swap_source_info.key == *swap_destination_info.key {
             return Err(SwapError::InvalidInput.into());
         }
+        if swap_source_info.key == source_info.key {
+            return Err(SwapError::InvalidInput.into());
+        }
+        if swap_destination_info.key == destination_info.key {
+            return Err(SwapError::InvalidInput.into());
+        }
```
joncinque, line 274: Great catch, this could be super destructive

token-swap/program/src/processor.rs
```diff
@@ -2215,20 +2233,20 @@ mod tests {
...
                 accounts.deposit(
                     &depositor_key,
-                    &pool_key,
-                    &mut pool_account,
                     &token_a_key,
                     &mut token_a_account,
                     &token_b_key,
                     &mut token_b_account,
+                    &pool_key,
+                    &mut pool_account,
```
joncinque, line 2241: Good catch, these were sloppy copy / paste on my part

joncinque (approved):
Looks good, thanks for noticing this

PR #728, opened 2020-10-26
Refer to CreateAccount

jackcmay:
There is no `CreateInstruction` System instruction, should be referring to `CreateAccount`


jackcmay:
@CriesofCarrots Maybe I'm misunderstanding the current reference?

CriesofCarrots (approved):
Spot on, good catch! Looks like that verbiage came in here: https://github.com/solana-labs/solana-program-library/pull/89

PR #729, opened 2020-10-26
Add Search functionality

rmshea:
* Install `lunar-search` component

* Add lightweight search

### DIFF
--- a/docs/src/theme/SearchBar/index.js
+++ b/docs/src/theme/SearchBar/index.js
@@ -1,0 +1,106 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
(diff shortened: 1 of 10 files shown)


docs/src/theme/SearchBar/index.js
```diff
@@ -0,0 +1,106 @@
+/**
```
mvines, line 1: There's no way to just pull this in from npm or something instead of duplicating?

rmshea: @mvines nope, have to add it to the repo

mvines: sad

mvines (approved):

PR #730, opened 2020-10-26
Add freeze and thaw account to module declaration in spl token js library

### LINKED ISSUE #697: [spl-token-js] thaw account and freeze account are currently not exposed in module
I am bypassing the type checks and they both seem to work, guessing those were overlooked?

paul-schaaf:
closes #697


jstarry:
Thank you!

PR #731, opened 2020-10-26
Token-swap: Add "production mode" hard-coded fee constraints

joncinque:
In the dex situation, the token swap program may have multiple frontends interacting with it, but the program creator would like to be the default destination for trading fees, along with enforcing certain fees on its exchange.  This provides a little framework for setting and enforcing these constants, enabled using the "production" feature.

Currently, only the owner key is configured through an environment variable.  I wanted to statically create a Pubkey based on &'static str using a const function, but the only const function for creating is from an array, so I'm not sure if we can go the full way to creating a Pubkey at this point.  We could eventually also include the fee parameters as env variables, but the conversion &'static str -> u64 wasn't clear either.

Additionally, there's an optional "host account" parameter added to the swap instruction.  If provided, the program creator shares some of the trading fee with this host account.

I had some trouble working with and without the constraints in testing, but I was pretty happy with the idea of just factoring out the constraints from the normal `process` function in the end.  Any better ideas are absolutely appreciated!

Note that a JS end-to-end test for production mode still needs to be added for CI.

### DIFF
--- a/token-swap/program/src/constraints.rs
+++ b/token-swap/program/src/constraints.rs
@@ -1,0 +1,49 @@
+//! Various constraints as required for production environments
+
+#[cfg(feature = "production")]
+use std::env;
+
+/// Encodes fee constraints, used in multihost environments where the program
+/// may be used by multiple frontends, to ensure that proper fees are being
+/// assessed.
+pub struct FeeConstraints<'a> {
+    /// Owner of the program
+    pub owner_key: &'a str,
+    /// Fee numerator
+    pub trade_fee_numerator: u128,
+    /// Fee denominator
+    pub trade_fee_denominator: u128,
+    /// Owner trade fee numerator
+    pub owner_trade_fee_numerator: u128,
+    /// Owner trade fee denominator
+    pub owner_trade_fee_denominator: u128,
+    /// Host fee numerator (e.g. 20 / 100 for host to receive 20% of owner trade fees)
+    pub host_fee_numerator: u128,
+    /// Host fee denominator
+    pub host_fee_denominator: u128,
+}
+
+#[cfg(feature = "production")]
+const OWNER_KEY: &'static str = env!("SWAP_PROGRAM_OWNER_FEE_ADDRESS");
+
+/// Fee structure defined by program creator in order to enforce certain
+/// fees when others use the program.  Adds checks on pool creation and
+/// swapping to ensure the correct fees and account owners are passed.
+pub const FEE_CONSTRAINTS: Option<FeeConstraints> = {
+    #[cfg(feature = "production")]
+    {
+        Some(FeeConstraints {
+            owner_key: OWNER_KEY,
+            trade_fee_numerator: 25,
+            trade_fee_denominator: 10000,
+            owner_trade_fee_numerator: 5,
+            owner_trade_fee_denominator: 10000,
+            host_fee_numerator: 20,
+            host_fee_denominator: 100,
+        })
+    }
+    #[cfg(not(feature = "production"))]
+    {
+        None
+    }
+};
--- a/token-swap/program/src/constraints.rs
+++ b/token-swap/program/src/constraints.rs
@@ -10,17 +10,17 @@ pub struct FeeConstraints<'a> {
     /// Owner of the program
     pub owner_key: &'a str,
     /// Fee numerator
-    pub trade_fee_numerator: u128,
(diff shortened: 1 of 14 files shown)


token-swap/program/src/constraints.rs
```diff
@@ -0,0 +1,184 @@
...
+/// Since this struct needs to be created at compile-time, we only have access
+/// to const functions and constructors. Since SwapCurve contains a Box, it
+/// cannot be used, so we have to split the curves based on their types.
+pub struct FeeConstraints<'a> {
+    /// Owner of the program
+    pub owner_key: &'a str,
+    /// Valid flat curves for the program
+    pub valid_flat_curves: &'a[FlatCurve],
+    /// Valid constant product curves for the program
+    pub valid_constant_product_curves: &'a[ConstantProductCurve],
```
joncinque, line 25: I'm not super happy about having to split these up, but we can't really have `dyn` things in const structure

token-swap/program/src/constraints.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+impl<'a> FeeConstraints<'a> {
+    /// Checks that the provided curve is valid for the given constraints
+    pub fn validate_curve(&self, swap_curve: &SwapCurve) -> Result<(), ProgramError> {
+        let valid_swap_curves: Vec<SwapCurve> = match swap_curve.curve_type {
```
joncinque, line 31: This part's a bit redundant, but `valid_flat_curves` and `valid_constant_product_curves` have different types, so we can't treat them the same

token-swap/program/src/processor.rs
```diff
@@ -328,6 +342,33 @@ impl Processor {
             )
             .ok_or(SwapError::FeeCalculationFailure)?;
         if pool_token_amount > 0 {
+            if let Some(_fee_constraints) = &fee_constraints {
```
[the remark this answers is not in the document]
CriesofCarrots: Can we do this check based on whether the host fee values are set in swap_curve and avoid passing in the constraints?

token-swap/program/src/processor.rs
```diff
@@ -203,6 +205,17 @@ impl Processor {
             return Err(SwapError::IncorrectPoolMint.into());
         }
 
+        if let Some(fee_constraints) = fee_constraints {
```
CriesofCarrots, line 208: Hmm, what happens if I just pass `None` into Initialize? Can't I bypass the constraints?
I was expecting to see a `production` feature-wrapped check here, instead of optional passing of the fee_constraints. What am I missing about the way this works?

token-swap/program/src/processor.rs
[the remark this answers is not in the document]
joncinque: Great point -- yes, absolutely

token-swap/program/src/processor.rs
joncinque: You could potentially find a way to pass `None` into initialize, but only if you can directly call `process_initialize` or `process_with_fee_constraints`.  `process` is the main entrypoint into the program, and it passes the hard-coded `FEE_CONSTRAINTS` into `process_with_fee_constraints`.  I might be missing something about how programs can be used -- is it possible to get around calling `process` in a production context?

CriesofCarrots: OH. Apologies, I misread the `process` edit. Disregard!

token-swap/program/src/curve.rs
```diff
@@ -315,12 +315,11 @@ impl CurveCalculator for FlatCurve {
     /// Calculate the host fee based on the owner fee, only used in production
     /// situations where a program is hosted by multiple frontends
     fn host_fee(&self, owner_fee: u128) -> Option<u128> {
-        let numerator = u128::try_from(self.host_fee_numerator).ok()?;
-        let denominator = u128::try_from(self.host_fee_denominator).ok()?;
-        owner_fee
-            .checked_mul(numerator)?
-            .checked_div(denominator)
-            .and_then(map_zero_to_none)
+        calculate_fee(
```
CriesofCarrots, line 318: This is nice cleanup

CriesofCarrots (approved):
r+ with whatever prettier wants fixed

joncinque:
Great thanks!  Prettier and eslint clashed on how a line should look for a long ternary operator, so they would keep clobbering each other's changes and then reporting an error on check.  I updated the problematic line to avoid a ternary operator.

PR #753, opened 2020-10-29
Document about the known fake rent issue

ryoqun:
Document another known issue.

sibling: #743

### DIFF
--- a/docs/src/token.md
+++ b/docs/src/token.md
@@ -389,6 +389,16 @@ instruction from within their program.
 The `Burn` instruction decreases an Account's token balance without transferring
 to another Account, effectively removing the token from circulation permanently.
 
+There is no other way to reduce supply on chain. This is similar to transferring
+to an account with unknown private key or burning the private key. But the act
+of burning by using `Burn` instructions is more explicit and can be confirmed on
+chain by any parties.
+
+Note that, there is a known issue where a malicious and determined account owner
+can silently burn their tokens without updating supply on chain by making an
+account be removed because of overdue rent because of
+[the known issue](#Rent-exemption).
+
 ### Authority delegation
 
 Account owners may delegate authority over some or all of their token balance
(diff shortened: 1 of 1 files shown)


docs/src/token.md
```diff
@@ -449,6 +458,17 @@ consistently valid Multisig accounts all Solana accounts holding a Account,
 Mint, or Multisig must contain enough SOL to be considered [rent
 exempt](https://docs.solana.com/implemented-proposals/rent)
```
mvines, line 460: Let's put this under a new subsection, and then link to it from "the known issue" above:
`#### Rent-exemption loophole`

docs/src/token.md
```diff
@@ -449,6 +458,17 @@ consistently valid Multisig accounts all Solana accounts holding a Account,
...
 
+However note that there is currently a loophole to escape from the rent-exemption
+rule. So it isn't technically impossible to create spl-token Solana accounts
```
mvines, line 462:
```suggestion
rule. So it isn't technically impossible to create SPL Token accounts
```

docs/src/token.md
```diff
@@ -449,6 +458,17 @@ consistently valid Multisig accounts all Solana accounts holding a Account,
...
+which aren't actually rent-exempt. This loophole was unintentionally created
+because the token program is susceptible to sysvar spoofing. Specifically, there
```
mvines, line 464: Can you link to the PR in the monorepo that fixed the issue please

docs/src/token.md
```diff
@@ -449,6 +458,17 @@ consistently valid Multisig accounts all Solana accounts holding a Account,
...
+was an insufficient check for accounts passed as sysvar while executing in
+solana-program-sdk, which is statically linked into the token program.
```
mvines, line 466: Can remove "while executing in
solana-program-sdk, which is statically linked into the token program." I think

CriesofCarrots: How about just: "It is possible to create SPL Token accounts that are not rent exempt by spoofing the Rent sysvar, since there are insufficient sysvar checks in the program. This could be abused to burn tokens by transferring tokens to a non-exempt Account that is subsequently rent-collected out of existence."

ryoqun:
@mvines @CriesofCarrots Thanks for all suggestions! I think I've addressed all comments.

CriesofCarrots (approved):

mvines (approved):

PR #755, opened 2020-10-30
Port to tip of master/v1.4 solana branches

mvines:
The movement of Accounts back into `solana-sdk` (https://github.com/solana-labs/solana/pull/13198) caused some bustage in  the SPL tests.  

Once 1.4.4 ships, this PR will be needed to get SPL building again.  You also need this PR if you want to `./patch.crates-io.sh` in a local solana monorepo


mvines:
@jstarry - help.  This builds locally but I can't debug the github action thing

mvines:
The problem is that the "test" job also needs the solana cli installed

PR #763, opened 2020-10-30
Added missing accounts to the interface to call stake account program

ysavchenko:
Calls from stake pool program to stake accounts were not working unless these two accounts (sysvar clock and stake account program id) were added.


joncinque (commented):
Just a nit on the order of accounts.  Can you rebase your branch again to pick up all of the CI updates to see if that fixes things?

Issue #584, opened 2020-10-08
CI jobs take an hour to complete


jstarry:
#### Problem
CI is very slow and creates a bottleneck for PRs

#### Proposed Change
More caching?
Use Github actions for more parallel jobs?

PR #768, opened 2020-11-01
Switch to Github Actions for CI

jstarry:
Fixes: https://github.com/solana-labs/solana-program-library/issues/584


jstarry:
Down from 1hr+ to 7.5min

mvines:
.github/workflows/pull-request.yml is pretty gross, I'm failing in some random yaml world and have no ability to reproduce locally.  Debugging via CI is the worst 💩 
I guess it's worth the speed up overall 😭

jstarry:
@mvines oops I went overboard with the config, I can either reintroduce the PR for full review or simplify in a follow up

mvines:
Simplifying in a follow-up wfm.   Minimal CI-specific crap, keep things as much as possible in shell scripts that can be also run and debug locally is the only way to keep CI from ruling you.

PR #769, opened 2020-11-02
Simplify github action with scripts

jstarry:


joncinque:
We had an issue with mergify when everything was updated, since it was just checking the Travis CI job, causing #775 to be incorrectly merged in.  According to the mergify documentation regarding GitHub Actions, we'll need more checks: https://docs.mergify.io/conditions.html?highlight=check%20success#github-actions

For this PR, that will mean changing the squash and rebase actions conditions to:
```yaml
conditions:
  - check-success=Travis CI - Pull Request
  - check-success=rustfmt
  - check-success=clippy
  - check-success=build_test
  - check-success=js_token
  - check-success=js_token_swap
  - check-success=js_token_lending
```

I couldn't find anything about the `status-success` condition that we use currently in .mergify.yml, so perhaps we should use `check-success` instead?

.github/workflows/pull-request.yml
```diff
@@ -63,63 +47,37 @@ jobs:
...
-  build_bpf_programs:
-    name: Build Program ${{ matrix.path }}
-    needs: list_bpf_programs
+  build_test:
     runs-on: ubuntu-latest
-    strategy:
-      matrix: ${{fromJson(needs.list_bpf_programs.outputs.matrix)}}
+    env:
+      SOLANA_VERSION: v1.4.4
+      RUST_VERSION: 1.47.0 # MSRV
```
mvines, line 63: Can we get these out of yaml and into a script?

ci/build_test_script.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
mvines, line 1: nit: `ci/build-test-script.sh` for consistency

ci/install-program-deps.sh
```diff
@@ -0,0 +1,15 @@
...
+set -e
+
+cargo --version
+cargo install rustfilt || true
+
+if [[ -n $SOLANA_VERSION ]]; then
+  sh -c "$(curl -sSfL https://release.solana.com/$SOLANA_VERSION/install)"
+fi
+
+export PATH=/home/runner/.local/share/solana/install/active_release/bin:"$PATH"
```
mvines, line 12: Does `/home/runner` exist as $HOME in github actions?

ci/js_token_swap_script.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -e
+
+(cd ../../token/js && npm install)
```
mvines, line 5: Let's not assume a PWD, that'll break when a user runs it.

Also more kabab case - `ci/js-token-swap-script.sh`.   The `-script` suffix seems unnecessary too

ci/js_token_swap_script.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -e
```
mvines, line 3: `set -ex` -- `-x` is super useful when something fails

.github/workflows/pull-request.yml
```diff
@@ -37,22 +37,6 @@ jobs:
           profile: minimal
```
mvines, line 37: Is .github/workflows/pull-request.yml composable?  ie, can we have one in each subdirectory and have those included individually by the main `.github/workflows/pull-request.yml`?

jstarry: > You must store workflow files in the .github/workflows directory of your repository.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows

We can split up workflows into separate files but right now at least the JavaScript tests rely on the programs being built in the same workflow.

jstarry:
@mvines feedback addressed. Test run: https://github.com/jstarry/solana-program-library/actions/runs/343669245

mvines (approved):
Thanks, looks much better!  Where does GITHUB_ENV get used?

jstarry:
> Where does GITHUB_ENV get used?

It's just a magic env var that is used to set env vars for later action steps

PR #777, opened 2020-11-02
Add more checks to mergify due to GitHub Actions migration

joncinque:
Let's see if this gets merged in automatically once all of the CI is passed!

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -9,7 +9,7 @@ env:
   SOLANA_VERSION: v1.4.4
 
 jobs:
-  github-checks:
+  all_github_checks:
     runs-on: ubuntu-latest
     needs:
       - clippy
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -9,6 +9,17 @@ env:
   SOLANA_VERSION: v1.4.4
 
 jobs:
+  github-checks:
+    runs-on: ubuntu-latest
+    needs:
+      - clippy
+      - list_bpf_programs
+      - cargo_test
+      - js_token
+      - js_token_swap
+      - js_token_lending
+      - build_bpf_programs
+      - excluded_project_tests
   rustfmt:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -20,6 +20,8 @@ jobs:
       - js_token_lending
       - build_bpf_programs
       - excluded_project_tests
+    steps:
+      - run: echo "Done"
   rustfmt:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


joncinque:
Mergify doesn't allow PRs that modify its configuration to make use of mergify, but it says "The new Mergify configuration is valid"

joncinque:
Mainly for a sanity check @mvines -- note that this will need to be updated in #769

.mergify.yml
```diff
@@ -6,7 +6,23 @@
 pull_request_rules:
   - name: automatic merge (squash) on CI success
     conditions:
-      - status-success=Travis CI - Pull Request
+      - check-success=Travis CI - Pull Request
+      - check-success=rustfmt
```
mvines, line 10: Ouch.  There's not a single roll-up status for ALL the github actions?  Listing out each individual action like this will cause automerge to break all the time

joncinque: Unfortunately no -- they even say it's not possible in their docs: https://docs.mergify.io/conditions.html?highlight=check%20success#validating-all-status-check

> There is no such thing as “every status check” in GitHub.

joncinque: on the flip side, we can do something like `check-success~=build` but they recommend against it

mvines: That's really shitty.  Buildkite solves this by having a roll-up build status that only goes green when the associated jobs go green.  It's disappointing that github actions didn't do that as as well

mvines: Can we have a single github action that depends on all the other actions?

joncinque: that's a good idea -- let me check and see

.github/workflows/pull-request.yml
```diff
@@ -9,6 +9,19 @@ env:
...
+      - clippy
+      - list_bpf_programs
+      - cargo_test
+      - js_token
+      - js_token_swap
+      - js_token_lending
+      - build_bpf_programs
+      - excluded_project_tests
+    steps:
+      - run: echo "Done"
```
joncinque, line 24: Although their docs don't mention that steps are required, they are :-\

joncinque:
It works!  Although we're just shifting the responsibility from changing the mergify file to the github actions, I hope that people will more readily remember to update the github actions file as needed.

Let me know if the names are ok and I'll merge in

.mergify.yml
```diff
@@ -15,7 +16,8 @@ pull_request_rules:
   # Join the dont-squash-my-commits group if you won't like your commits squashed
   - name: automatic merge (rebase) on CI success
     conditions:
-      - status-success=Travis CI - Pull Request
+      - check-success=Travis CI - Pull Request
+      - check-success=all_github_checks
```
mvines, line 20:
```suggestion
      - check-success=all_github_action_checks
```

mvines (approved):
nice! Just a minor name nit

PR #778, opened 2020-11-03
SPL Associated Account program

mvines:
A generalization of #626.  This program defines how to associate an account with a primary account, while allowing anybody to create the associated account.  Essentially `SystemInstruction::CreateAccount` that anybody can use to create an account for ***you***.

Also features a `token` module that's targeted for the SPL token use case.


mvines:
Ack, 🚳.   I really want to generalize this but nope.

PR #779, opened 2020-11-03
SPL Associated Token Account program

mvines:
A generalization of #626 but not as general as #778. This program defines how to associate a token account with a primary account, while allowing anybody to create that associated token account.


mvines:
I ripped out Assert!   Per the SPL Token Wallet Integration Guide (https://github.com/solana-labs/solana-program-library/pull/612, happy to take review on that too), the wallet app shouldn't permit Grandma to alter the authority of her associated token account **ever**.  So as long as that's respected, the Assert instruction is not required

mvines:
Deployed on both devnet and testnet to start with.  Will save deploying to mainnet-beta for later, after more client integration testing

jstarry (commented):
Looks good, thanks for looking out for grandma!

PR #788, opened 2020-11-03
Added positive tests in stake pool and fixed several bugs in instructions

ysavchenko:
Positive tests for:
- withdraw
- claim
- set staking authority
- set owner

Fixed bugs when forming claim and set owner instructions.

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -607,7 +607,7 @@ impl PrintProgramError for Error {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::instruction::{deposit, initialize, Fee, InitArgs};
+    use crate::instruction::*;
     use solana_program::{
         instruction::Instruction, native_token::sol_to_lamports, program_pack::Pack, program_stubs,
         rent::Rent,
@@ -702,10 +702,19 @@ mod tests {
         pub withdraw_authority_key: Pubkey,
         pub fee: Fee,
         pub owner_key: Pubkey,
+        pub owner_account: Account,
         pub owner_fee_key: Pubkey,
         pub owner_fee_account: Account,
         pub mint_key: Pubkey,
         pub mint_account: Account,
+    }
+
+    struct DepositInfo {
+        stake_account_key: Pubkey,
+        stake_account_account: Account,
+        token_receiver_key: Pubkey,
+        token_receiver_account: Account,
+        token_owner_key: Pubkey,
     }
 
     fn do_process_instruction(
@@ -759,26 +768,34 @@ mod tests {
         Rent::default().minimum_balance(SplMint::get_packed_len())
     }
 
+    struct TokenInfo {
+        key: Pubkey,
+        account: Account,
+        owner: Pubkey,
+    }
+
     fn create_token_account(
         program_id: &Pubkey,
         mint_key: &Pubkey,
         mint_account: &mut Account,
-    ) -> (Pubkey, Account) {
-        let account_key = Pubkey::new_unique();
-        let mut account_account = Account::new(
-            account_minimum_balance(),
-            SplAccount::get_packed_len(),
-            &program_id,
-        );
+    ) -> TokenInfo {
+        let mut token = TokenInfo {
+            key: Pubkey::new_unique(),
+            account: Account::new(
+                account_minimum_balance(),
+                SplAccount::get_packed_len(),
+                &program_id,
+            ),
+            owner: Pubkey::new_unique(),
+        };
         let mut rent_sysvar_account = create_account(&Rent::free(), 1);
-        let owner_key = Pubkey::new_unique();
         let mut owner_account = Account::default();
 
         // create account
         do_process_instruction(
-            initialize_account(&program_id, &account_key, &mint_key, &owner_key).unwrap(),
+            initialize_account(&program_id, &token.key, &mint_key, &token.owner).unwrap(),
             vec![
-                &mut account_account,
+                &mut token.account,
                 mint_account,
                 &mut owner_account,
                 &mut rent_sysvar_account,
@@ -786,7 +803,7 @@ mod tests {
         )
         .unwrap();
 
-        (account_key, account_account)
+        token
     }
 
     fn _mint_token(
@@ -796,15 +813,14 @@ mod tests {
         authority_key: &Pubkey,
         amount: u64,
     ) -> (Pubkey, Account) {
-        let (account_key, mut account_account) =
-            create_token_account(program_id, mint_key, mint_account);
+        let mut token_account = create_token_account(program_id, mint_key, mint_account);
         let mut authority_account = Account::default();
 
         do_process_instruction(
             spl_token::instruction::mint_to(
                 &program_id,
                 &mint_key,
-                &account_key,
+                &token_account.key,
                 &authority_key,
                 &[],
                 amount,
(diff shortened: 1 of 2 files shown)


stake-pool/program/src/processor.rs
```diff
@@ -941,70 +1047,241 @@ mod tests {
...
-        let mut stake_account_account = Account::new(stake_balance, 100, &stake_program_id());
-        // TODO: Set stake account Withdrawer authority to pool_info.deposit_authority_key
+        let deposit_info = do_deposit(&mut pool_info, stake_balance);
 
-        // Create account to receive minted tokens
-        let (token_receiver_key, mut token_receiver_account) = create_token_account(
+        // Test stake pool balance
+        let state = State::deserialize(&pool_info.pool_account.data).unwrap();
+        match state {
+            State::Unallocated => panic!("Stake pool state is not initialized after deposit"),
```
joncinque, line 1055: nit: Prefer to use the `matches!` macro instead https://doc.rust-lang.org/std/macro.matches.html

stake-pool/program/src/processor.rs
```diff
@@ -941,70 +1047,241 @@ mod tests {
...
+                .expect("User token account is not initialized after deposit");
+        assert_eq!(user_token_state.amount, user_token_balance);
+        let fee_token_state = SplAccount::unpack_from_slice(&pool_info.owner_fee_account.data)
+            .expect("Fee token account is not initialized after deposit");
+        assert_eq!(fee_token_state.amount, fee_token_balance);
+
+        // Test mint total issued tokens
+        let mint_state = SplMint::unpack_from_slice(&pool_info.mint_account.data)
+            .expect("Mint account is not initialized after deposit");
+        assert_eq!(mint_state.supply, stake_balance);
```
joncinque, line 1074: It may be useful to add a TODO to check the results of the stake instruction too

stake-pool/program/src/processor.rs
```diff
@@ -941,70 +1047,241 @@ mod tests {
...
                 &mut Account::default(),
                 &mut Account::default(),
                 &mut Account::default(),
             ],
         )
-        .expect("Error on stake pool deposit");
+        .expect("Error on withdraw");
 
-        // Test stake pool balance
+        let fee_amount = stake_balance * FEE_DEFAULT.numerator / FEE_DEFAULT.denominator;
```
joncinque, line 1129: You can use the `fee` on `pool_info` to calculate this instead

stake-pool/program/src/processor.rs
```diff
@@ -941,70 +1047,241 @@ mod tests {
...
+        assert_eq!(
+            user_token_state.amount,
+            stake_balance - fee_amount - withdraw_amount
+        );
+
+        // Check stake pool token amounts
         let state = State::deserialize(&pool_info.pool_account.data).unwrap();
         match state {
-            State::Unallocated => panic!("Stake pool state is not initialized after deposit"),
+            State::Unallocated => panic!("Stake pool state is not initialized after withdraw"),
```
joncinque, line 1142: You can do `assert!(matches!(...` here too

stake-pool/program/src/processor.rs
```diff
@@ -941,70 +1047,241 @@ mod tests {
...
+    #[test]
+    fn test_withdraw() {
+        let mut pool_info = create_stake_pool_default();
+
+        let user_withdrawer_key = Pubkey::new_unique();
+
+        let stake_balance = sol_to_lamports(20.0);
+        let mut deposit_info = do_deposit(&mut pool_info, stake_balance);
+
+        let withdraw_amount = sol_to_lamports(5.0);
```
joncinque, line 1085: nit: since `withdraw_amount` is in pool tokens and not lamports, could you hard-code a number instead?

stake-pool/program/src/processor.rs
```diff
@@ -941,70 +1047,241 @@ mod tests {
...
 
-        // Test mint total issued tokens
-        let mint_state = SplMint::unpack_from_slice(&pool_info.mint_account.data)
-            .expect("Mint account is not initialized after deposit");
-        assert_eq!(mint_state.supply, stake_balance);
+        let stake_balance = sol_to_lamports(20.0);
+        let mut deposit_info = do_deposit(&mut pool_info, stake_balance);
+
+        // Need to deposit more to cover deposit fee
+        let fee_amount = stake_balance * FEE_DEFAULT.numerator / FEE_DEFAULT.denominator;
```
joncinque, line 1159: nit: You can use the fee on `pool_info` here as well

stake-pool/program/src/processor.rs
```diff
@@ -941,70 +1047,241 @@ mod tests {
...
+                &mut Account::default(),
+                &mut Account::default(),
+            ],
+        )
+        .expect("Error on claim");
+
+        let user_token_state =
+            SplAccount::unpack_from_slice(&extra_deposit_info.token_receiver_account.data)
+                .expect("User token account is not initialized after withdraw");
+        assert_eq!(user_token_state.amount, 0);
```
joncinque, line 1211: nit: you can add a TODO here as well to check the stake account ownership

stake-pool/program/src/processor.rs
```diff
@@ -941,70 +1047,241 @@ mod tests {
...
+                &mut pool_info.owner_account,
+                &mut new_owner_account,
+                &mut new_owner_fee.account,
+            ],
+        )
+        .expect("Error on set_owner");
+
+        let state = State::deserialize(&pool_info.pool_account.data).unwrap();
+        match state {
+            State::Unallocated => panic!("Stake pool state is not initialized after set owner"),
```
joncinque, line 1280: nit: `assert!(matches!(...` here too

joncinque (commented):
Looks great, mainly just a few nits

stake-pool/program/src/processor.rs
```diff
@@ -892,20 +943,75 @@ mod tests {
...
+            ],
+        )
+        .expect("Error on stake pool deposit");
+
+        DepositInfo {
+            stake_account_key,
+            stake_account_account,
+            token_receiver_key: token.key,
+            token_receiver_account: token.account.clone(),
+            token_owner_key: token.owner,
```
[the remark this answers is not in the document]
ysavchenko: Actually there are two deposit functions: one accepts token account as a parameter and another creates new token account and passes it to the first function (`do_deposit` and `do_deposit_custom`). We can make `do_deposit_custom` return only stake account info (key/account) but it is better to have them both return the same DepositInfo struct (to make code calling these two functions similar).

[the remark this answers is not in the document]
joncinque: I see what you mean.  To be honest, it was a bit hard to follow some of the tests because of the account creation in `do_deposit`, so I would explicitly create the account outside, and only have `do_deposit_custom`.  What do you think?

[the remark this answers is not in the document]
ysavchenko: Actually it was the other way round :) We did do_deposit, which is used in deposit/withdraw/claim tests. And then realized that we need to collect tokens for two deposits to the same accounts, so we've added do_deposit_custom to allow that. And I think in remaining tests we will mostly use do_deposit which will create token account by itself.

But you of course have the last word, if you think it will be clearer if token receiver was created outside of the deposit we will do it this way :)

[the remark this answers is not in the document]
joncinque: Creating an account in `do_deposit` still seems confusing to me, so I'd prefer that we only have the version that takes `&mut TokenInfo`, and then also removing the `token_*` fields from `DepositInfo`.  That way, you don't need to worry about data duplication.

joncinque:
Looks great, thanks for the refinements!

joncinque (approved):

PR #809, opened 2020-11-09
Fix type definitions for TokenSwap class

rawfalafel:
This changeset fixes some inconsistencies between the type definitions and the actual class.


joncinque:
Sorry that I missed this yesterday -- thanks for the PR!

PR #813, opened 2020-11-09
token-swap: Fix zero fee calculation

joncinque:
If a swap has a trading fee and a zero owner fee, when calculating the owner
fee to pool token conversion, the simulated trading fee will be non-zero
because we floor the trading fee to 1 token.  If the input to trading
fee calculation is 0, then we will try to do `0 - 1` on an unsigned int,
which causes an underflow error.  The fix is to simply return a 0 fee
for a 0 input.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -684,13 +684,12 @@ dependencies = [
 [[package]]
 name = "curve25519-dalek"
 version = "2.1.0"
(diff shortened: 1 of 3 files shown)


CriesofCarrots (approved):
Good catch!

joncinque:
It was all thanks to @rawfalafel :smile:

PR #815, opened 2020-11-10
Add Rust examples

jackcmay:
Add example rust programs that demonstrate various program functionality.


jackcmay:
@mvines Planning to link to these examples, using the new test framework, any comments before I commit them?

Cargo.toml
```diff
@@ -1,6 +1,11 @@
 [workspace]
 members = [
   "associated-token-account/program",
+  "examples/rust-cross-program-invocation",
```
mvines, line 4: Drop the `rust-` prefix?   ie, default to Rust and then add a `cpp-` prefix for the outliers?

examples/rust-cross-program-invocation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "spl-example-cross-program-invocation"
+version = "1.0.1"
```
mvines, line 3: why 1.0.1?  Keep it 1.0.0?

examples/rust-cross-program-invocation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+description = "Solana Program Library Cross Program Invocation Example"
+authors = ["Solana Maintainers <maintainers@solana.foundation>"]
+repository = "https://github.com/solana-labs/solana-program-library"
+license = "Apache-2.0"
+edition = "2018"
```
mvines, line 8: Let's add `publish = false` to all the examples

examples/rust-cross-program-invocation/src/entrypoint.rs
```diff
@@ -0,0 +1,16 @@
+//! Program entrypoint
+
+#![cfg(all(target_arch = "bpf", not(feature = "no-entrypoint")))]
```
mvines, line 3: I think we can get away with this:
```suggestion
#![cfg(feature = "no-entrypoint")]
```

Cargo.toml
jackcmay: There are going to be both Rust and C versions, are we officially making C a second citizen?

examples/rust-cross-program-invocation/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! Rust example demonstrating invoking another program
+#![deny(missing_docs)]
+#![forbid(unsafe_code)]
+
+mod entrypoint;
+pub mod processor;
+
+solana_program::declare_id!("invoker111111111111111111111111111111111111");
```
mvines, line 8: I feel like we should remove the `solana_program::declare_id!()` from all the examples because they setup the user for failure.  
* If they attempt to deploy the example to this address, it'll fail. 
* If they deploy the example to another address and accidentally use this address in their code, it'll fail. 
So this address only works in the context of a unit test

Cargo.toml
mvines: From an SPL perspective I don't see why not.

examples/rust-cross-program-invocation/src/lib.rs
jackcmay: Yeah, I originally had that, I'll go back,

examples/rust-sysvar/src/processor.rs
```diff
@@ -0,0 +1,37 @@
...
+    // Create in iterator to safety reference accounts in the slice
+    let account_info_iter = &mut accounts.iter();
+
+    // As part of the program specification the first account is the clock
+    // sysvar
+    let clock_sysvar_info = next_account_info(account_info_iter)?;
+
+    if *clock_sysvar_info.key != sysvar::clock::id() {
+        // first account is not the clock sysvar
+        return Err(ProgramError::InvalidArgument);
```
mvines, line 27: `Clock::from_account_info()` now performs this check right?   So this condition can be removed?

mvines (commented):
Sweet, most of my comments apply to all the example programs looks like

Cargo.toml
jackcmay: I'm in favor of keeping both `Rust` and `C` prefix since they align with the docs and it makes it immediately clear from the first glance what language the example is.  We probably won't have other SPL programs written in both languages but the examples we will (and hopefully more in the future like solidity).  Could also do separate directories but that hasn't been a liked approach in the past ;-)

examples/rust-sysvar/src/processor.rs
jackcmay: true dat

Cargo.toml
jackcmay: Now I'm leaning toward a `c` and `rust` directory for the examples, prefixes are verbose when scaled up in the number of examples and number of languages, plus any language-specific readme can be in the corresponding directory

examples/rust-cross-program-invocation/src/entrypoint.rs
jackcmay: Not happy with just `not(feature = "no-entrypoint")`, `global_allocator` conflicts with the one in solana's `measure` crate.  Seems like the checks in the entrypoint which already test for bpf are not working...

mvines: ah, right.  Under `cargo test` measure gets sucked in.  Something to fix later... :(

PR #827, opened 2020-11-11
Added set-staking-auth and set-owner commands to stake pool CLI

ysavchenko:


stake-pool/cli/src/main.rs
```diff
@@ -562,6 +562,104 @@ fn command_withdraw(
...
+        .stake_pool()
+        .unwrap();
+
+    // If new accounts are missing in the arguments use the old ones
+    let new_owner: Pubkey = match new_owner {
+        None => pool_data.owner,
+        Some(value) => *value,
+    };
+    let new_fee_receiver: Pubkey = match new_fee_receiver {
+        None => pool_data.owner,
```
joncinque, line 624: Be careful here -- the `new_fee_receiver` is meant to be a new fee account, so this should be `pool_data.owner_fee_account`.  I looked back at `process_set_owner`, and we don't do any checks on the new fee account.  We should add a check that it's actually a valid pool token account to avoid causing a problem later.

stake-pool/cli/src/main.rs
```diff
@@ -562,6 +562,104 @@ fn command_withdraw(
...
+                match TokenAccount::unpack_from_slice(account_data.as_slice()) {
+                    Ok(data) => data,
+                    Err(_) => {
+                        println!("{} is not a token account", value);
+                        return Ok(None);
+                    }
+                };
+            if account_data.mint != pool_data.pool_mint {
+                println!("Fee receiver account belongs to a different mint");
+                return Ok(None);
```
joncinque, line 638: Instead of doing a `println!` and `return Ok(None)` we'd be better off returning an `Err` with the message, and then catching / printing the error at the processor level

stake-pool/cli/src/main.rs
```diff
@@ -562,6 +562,104 @@ fn command_withdraw(
...
+        )?],
+        Some(&config.fee_payer.pubkey()),
+    );
+
+    let (recent_blockhash, fee_calculator) = config.rpc_client.get_recent_blockhash()?;
+    check_fee_payer_balance(config, fee_calculator.calculate_fee(&transaction.message()))?;
+    let mut signers = vec![config.fee_payer.as_ref(), config.owner.as_ref()];
+    unique_signers!(signers);
+    transaction.sign(&signers, recent_blockhash);
+    Ok(Some(transaction))
```
joncinque, line 603: nit: this predates this PR, but a lot of this logic can probably be factored out since it's done on every command. You can see a slick of way of refactoring it in the token cli at https://github.com/solana-labs/solana-program-library/blob/master/token/cli/src/main.rs#L1926

You can save it for a separate PR though

stake-pool/cli/src/main.rs
```diff
@@ -719,6 +817,67 @@ fn main() {
...
+                    .help("Stake account address to change staking authority."),
+            )
+            .arg(
+                Arg::with_name("new_staker")
+                    .long("new-staker")
+                    .validator(is_pubkey)
+                    .value_name("ADDRESS")
+                    .takes_value(true)
+                    .required(true)
+                    .help("Public key of the new staker account."),
```
joncinque, line 846: The CLI document mentions that this is optional and defaults to the owner, which makes sense to me.  Do we no longer want to do that?

stake-pool/cli/src/main.rs
```diff
@@ -719,6 +817,67 @@ fn main() {
...
+                    .validator(is_pubkey)
+                    .value_name("ADDRESS")
+                    .takes_value(true)
+                    .help("Public key for the new account to set as the stake pool fee receiver."),
+            )
+            .group(ArgGroup::with_name("new_accounts")
+                .arg("new_owner")
+                .arg("new_fee_receiver")
+                .required(true)
+            )
```
joncinque, line 879: In case someone wants to update both at once, this should include `multiple(true)`.  If someone wants to reassign both, we have a footgun if they do it in the wrong order, e.g.:

case 1:
* reassign fee account -> succeeds
* reassign owner -> succeeds

case 2:
* reassign owner -> succeeds
* reassign fee account -> fails, no longer the owner

So let's have both be possible at once.

joncinque (commented):
A few little things, it's mostly there

stake-pool/cli/src/main.rs
ysavchenko: We do have a check for the new fee account to have a proper mint assigned to it.
Using owner as default is nasty, great catch.

stake-pool/cli/src/main.rs
ysavchenko: Yes, this was bothering me as well

stake-pool/cli/src/main.rs
ysavchenko: Agreed and yes, I'll leave it for one of the next PRs, it is better to refactor this in all CLIs at once.

stake-pool/cli/src/main.rs
ysavchenko: Since our discussion on how to deal with Staker authority (which prevents withdraw from happening if changed) did not finish I decided to keep it required, so the authority will not be changed accidentally.

stake-pool/cli/src/main.rs
joncinque: Yes true, you're doing it here, which is great.  We also need a check at https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/src/processor.rs#L561 though -- could you add it with this PR?

stake-pool/cli/src/main.rs
joncinque: Ok that's fine, we can always change that later if we need to.

stake-pool/cli/src/main.rs
```diff
@@ -361,16 +361,16 @@ fn command_list(config: &Config, pool: &Pubkey) -> CommandResult {
...
-        println!("No accounts found.")
-    } else {
-        let mut total_balance: u64 = 0;
-        for (pubkey, account) in accounts {
-            let balance = account.lamports;
-            total_balance += balance;
-            println!("{}\t{} SOL", pubkey, lamports_to_sol(balance));
-        }
-        println!("Total: {} SOL", lamports_to_sol(total_balance));
+        return Err("No accounts found.".to_string().into());
```
joncinque, line 364: Nice

stake-pool/cli/src/main.rs
ysavchenko: Ok, got it, I guess we should add it to the stake pool initialization as well.

joncinque: Yes, please do

ysavchenko: Done :)

joncinque:
Great, thanks for everything!

PR #828, opened 2020-11-11
stake-pool: Add end-to-end testing using ProgramTest

joncinque:
stake pool unit tests can only cover the local logic, since it isn't possible to extract the stake program processor like we can with spl_token.  This adds the capability to run tests against the real stake program processor using ProgramTest.

### DIFF
--- a/stake-pool/program/Cargo.toml
+++ b/stake-pool/program/Cargo.toml
@@ -22,7 +22,9 @@ spl-token = { path = "../../token/program", features = [ "no-entrypoint" ] }
 thiserror = "1.0"
 
 [dev-dependencies]
+solana-program-test = "1.4.7"
 solana-sdk = "1.4.7"
+tokio = { version = "0.3", features = ["macros"]}
 
 [lib]
 crate-type = ["cdylib", "lib"]
--- a/stake-pool/program/Cargo.toml
+++ b/stake-pool/program/Cargo.toml
@@ -22,6 +22,7 @@ spl-token = { path = "../../token/program", features = [ "no-entrypoint" ] }
 thiserror = "1.0"
 
 [dev-dependencies]
+bincode = "1.3.1"
 solana-program-test = "1.4.7"
 solana-sdk = "1.4.7"
 tokio = { version = "0.3", features = ["macros"]}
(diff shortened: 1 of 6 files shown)


stake-pool/program/src/stake.rs
```diff
@@ -102,6 +103,55 @@ pub enum StakeInstruction {
     AuthorizeWithSeedNOTUSED,
 }
 
+/// FIXME copied from the stake program
```
joncinque, line 106: I hope that all of this can get removed soon :-)

CriesofCarrots: 🙏

stake-pool/program/Cargo.toml
```diff
@@ -22,7 +23,10 @@ spl-token = { path = "../../token/program", features = [ "no-entrypoint" ] }
 thiserror = "1.0"
 
 [dev-dependencies]
+bincode = "1.3.1"
+solana-program-test = "1.4.5"
```
CriesofCarrots, line 27: Bump this to v1.4.7 to match solana-sdk (happened today :) )

stake-pool/program/src/stake.rs
```diff
@@ -102,6 +103,55 @@ pub enum StakeInstruction {
...
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Copy)]
+#[allow(clippy::large_enum_variant)]
+pub enum StakeState {
+    /// FIXME copied from the stake program
```
CriesofCarrots, line 110: Just curious: why the FIXME comments on some variants and fields as well as the enum/struct?

CriesofCarrots (approved):
A nice intro to the new test framework; CI failure looks unrelated

stake-pool/program/src/stake.rs
joncinque: Clippy requires public enums / structs / fields / functions to have an annotation, so we just throw this one on usually

stake-pool/program/Cargo.toml
joncinque: Will do, thanks!

stake-pool/program/src/stake.rs
CriesofCarrots: Ah thanks, forgot about that setting on this repo

PR #830, opened 2020-11-11
Fix u64.fromBuffer return type to u64

rawfalafel:
`fromBuffer` expects the return type to be `u64` or `Numberu64` but returns `BN` instead. This causes errors downstream if a user tries to call `toBuffer`, which only exists on `u64`.


joncinque:
Looks great, thanks for looking into this

PR #824, opened 2020-11-11
Feature Proposal program

mvines:
The Feature Proposal program will allow for Solana runtime features to be activated based on community vote.  The feature proposer is responsible for:
* Submitting the code for the feature itself into the Solana code base, and gating the feature by a feature id
* Once all validators have updated to a version of the software with the new feature, submitting a transaction to initiate the on-chain voting.
* Distributing the SPL Tokens that will be used to vote for the feature using `solana-tokens` (fyi @CriesofCarrots)
* Informing the community that a vote is active, and sharing the SPL Token activation address that used to denote "yes".

Community members then use `spl-token transfer` to vote for the feature activation (or another SPL Token-enabled wallet of their choosing).

Finally the permissionless `Tally` instruction is invoked by anybody to confirm the vote count and affect the runtime feature activation.

TODO:
* [ ] Documentation
* [ ] Comprehensive unit tests
* [ ] CLI integration

PR #836, opened 2020-11-12
Upgrade to Solana 1.4.7

mvines:
https://github.com/solana-labs/solana-program-library/pull/824 needs `solana-program-test` features added in 1.4.7


mvines:
hmm, token-lending test failed.  I think it's unrelated to the upgrade, merging on red 💥

PR #838, opened 2020-11-13
Added stable curve invariant to the token swap smart contract

ysavchenko:


token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+
+    // sum' = prod' = x
+    // c =  D ** (n + 1) / (n ** (2 * n) * prod' * A)
+    let c = checked_pow(d_val, 3)?.checked_div(
+        new_source_amount.checked_mul(checked_pow(n_coins, 2)?.checked_mul(leverage)?)?,
+    )?;
+    // b = sum' - (A*n**n - 1) * D / (A * n**n)
+    let b = new_source_amount.checked_add(d_val.checked_div(leverage)?)?; // d is subtracted on line 82
+
+    // Solve for y by approximating: y**2 + b*y = c
```
joncinque, line 111: I wonder why they don't just use the quadratic equation to solve this

token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+    let n_coins: u128 = 2;
+    let leverage = amp.checked_mul(n_coins)?; // A * n
+
+    // sum' = prod' = x
+    // c =  D ** (n + 1) / (n ** (2 * n) * prod' * A)
+    let c = checked_pow(d_val, 3)?.checked_div(
+        new_source_amount.checked_mul(checked_pow(n_coins, 2)?.checked_mul(leverage)?)?,
+    )?;
+    // b = sum' - (A*n**n - 1) * D / (A * n**n)
+    let b = new_source_amount.checked_add(d_val.checked_div(leverage)?)?; // d is subtracted on line 82
```
joncinque, line 109: line number is off in this comment -- also, why don't we just subtract it right away?

token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+/// y**2 + y * (sum' - (A*n**n - 1) * D / (A * n**n)) = D ** (n + 1) / (n ** (2 * n) * prod' * A)
+/// y**2 + b*y = c
+fn compute_dest(amp: u128, new_source_amount: u128, d_val: u128) -> Option<u128> {
+    // XXX: Curve uses u256
+    let n_coins: u128 = 2;
+    let leverage = amp.checked_mul(n_coins)?; // A * n
+
+    // sum' = prod' = x
+    // c =  D ** (n + 1) / (n ** (2 * n) * prod' * A)
+    let c = checked_pow(d_val, 3)?.checked_div(
```
joncinque, line 105: better to use `checked_pow(d_val, n_coins.checked_add(1)?)` to avoid an issue later

token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+        new_source_amount.checked_mul(checked_pow(n_coins, 2)?.checked_mul(leverage)?)?,
```
joncinque, line 106: This is also weird -- in the original code, their `Ann` is just `amp * n_coins` when it should be `amp * n_coins ^ n_coins`.  Either way, this should be `checked_pow(n_coins, n_coins)`

token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+            trade_fee_denominator,
+            owner_trade_fee_numerator,
+            owner_trade_fee_denominator,
+            owner_withdraw_fee_numerator,
+            owner_withdraw_fee_denominator,
+            host_fee_numerator,
+            host_fee_denominator,
+            amp,
+        };
+        assert_eq!(calculator.new_pool_supply(), INITIAL_SWAP_POOL_AMOUNT);
```
joncinque, line 299: I think we'll need to customize this as well

token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+            owner_withdraw_fee_denominator,
+            host_fee_numerator,
+            host_fee_denominator,
+            amp,
+        };
+        let result = curve
+            .swap(source_amount, swap_source_amount, swap_destination_amount)
+            .unwrap();
+        assert_eq!(result.new_source_amount, 1_100);
+        assert_eq!(result.amount_swapped, 2_063);
```
joncinque, line 367: Where do these numbers come from? Is there a reference calculation somewhere?

joncinque (commented):
Almost all of the comments are related to the math, which you don't need to worry about for your implementation.  I wanted to get them down for later discussions though.  It's just a few small nits for you :smile: nice job!

token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+    fn swap(
+        &self,
+        source_amount: u128,
+        swap_source_amount: u128,
+        swap_destination_amount: u128,
+    ) -> Option<SwapResult> {
+        let trade_fee = self.trading_fee(source_amount)?;
+        let owner_fee = self.owner_trading_fee(source_amount)?;
+
+        let new_source_amount_less_fee = swap_source_amount
```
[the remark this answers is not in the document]
ysavchenko: Yes, this was intentional and we changed the fee to be taken from the source, like with the constant product or flat curve in the same contract. If we would take the fee from the swap destination we would need to change the way it is calculated into tokens in processor.rs.

token-swap/program/src/curve/stable.rs
ysavchenko: This is probably because curve.fi is about n-tokens so the equation would be in the higher order than 2.

token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+    sum_x: u128,
+    d_p: u128,
+    n_coins: u128,
+) -> Option<u128> {
+    let leverage_mul = leverage.checked_mul(sum_x)?;
+    let d_p_mul = d_p.checked_mul(n_coins)?;
+
+    let l_val = leverage_mul.checked_add(d_p_mul)?.checked_mul(initial_d)?;
+
+    let leverage_sub = leverage.checked_sub(1)?.checked_mul(initial_d)?;
```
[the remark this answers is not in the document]
michaelhly: In my Implementation, I assumed that all coins in the same pool will have the same decimals. All instances of `A_PRECISION` are replaced with `1`.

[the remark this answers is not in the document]
michaelhly: Also here is the convergence formula from Curve.fi. 

https://github.com/curvefi/curve-contract/blob/295e7daaad0654a6c7a233f77e82a01fb78d85b4/tests/simulation.py#L39

This is also the `compute_d` I check mine against.

token-swap/program/src/curve/stable.rs
michaelhly: Subtracting the invariant `d` here can lead to a lot of underflow errors. `d` can be quite large. 

Curve also delays subtracting `d` in the iteration step:
https://github.com/curvefi/curve-contract/blob/295e7daaad0654a6c7a233f77e82a01fb78d85b4/contracts/pool-templates/base/SwapTemplateBase.vy#L638

token-swap/program/src/curve/stable.rs
michaelhly: To be honest, I am also confused about this when I was implementing this. I just kind of assumed that the power of n gets applied in the interaction step? Don't know if this is the correct assumption.

token-swap/program/src/curve/stable.rs
michaelhly: It looks like these tests are just taken from `contant_product.rs` and the expected values are hardcoded to whatever values the `swap` function produces.

Math for this curve invariant is a lot more complicated compared to the constant product curve invariant. I still recommend porting over the thin simulation wrapper I wrote and add tests to check math against the simulation script provided by Curve.fi.

The range of numbers tested in these tests is really small and does not account for things like very big numbers, etc.

token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+    let amount_b_times_coins = amount_b.checked_mul(n_coins)?;
+    let sum_x = amount_a.checked_add(amount_b)?; // sum(x_i), a.k.a S
+    if sum_x == 0 {
+        Some(0)
+    } else {
+        let mut d_prev: u128;
+        let mut d = sum_x;
+        let leverage = amp.checked_mul(n_coins)?; // A * n
+                                                  // Newton's method to approximate D
+        for _ in 0..128 {
```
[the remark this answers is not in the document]
michaelhly: @joncinque how did you check this? Do you still have the script for checking this? In my implementation, I actually blew it up more by converting all numbers to u256 😅 .

token-swap/program/src/curve/stable.rs
michaelhly: I don't think ^ is correct. But I'm not sure why they didn't use the quadratic equation here.

token-swap/program/src/curve/stable.rs
michaelhly: +1

token-swap/program/src/curve/stable.rs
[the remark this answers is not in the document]
joncinque: @michaelhly you have to hack around a bit to get it to work unfortunately :disappointed:  we'll be better off using `solana-program-test` to count the number of instructions used with some tests.  Actually, are you running this in your stable swap hackathon project?  If you can get it to work there, we might be ok.  Maybe these end up converging sooner than expected.

token-swap/program/src/curve/stable.rs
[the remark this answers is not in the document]
joncinque: Ok great, that works

token-swap/program/src/curve/stable.rs
joncinque: I thought about this some more, and the quadratic equation would also require some approximation since we need a square root, so we're all good here

token-swap/program/src/curve/stable.rs
joncinque: Mm maybe.  Either way, let's just start with whatever they already have, and I'll go to them with the math questions to be sure we've understood everything.

token-swap/program/src/curve/stable.rs
joncinque: Gotcha, yeah that would be great.  I wasn't sure if there were some published reference calculation numbers that we use to be sure we're doing the right thing.

token-swap/program/src/constraints.rs
```diff
@@ -87,6 +89,18 @@ const VALID_FLAT_CURVES: &[FlatCurve] = &[FlatCurve {
     host_fee_numerator: 20,
     host_fee_denominator: 100,
 }];
+#[cfg(feature = "production")]
+const VALID_STABLE_CURVES: &[StableCurve] = &[StableCurve {
```
joncinque, line 93: Let's keep this empty for the moment since I'm not sure if our production users want to allow for a `StableCurve`

joncinque (commented):
Just one last comment.  For the CI failures, you'll need to run `cargo fmt`.  For the other CI failure, you'll need to update the token-swap JS to account for the larger swap curve by adding 8 bytes of padding after the last field at https://github.com/solana-labs/solana-program-library/blob/master/token-swap/js/client/token-swap.js#L78 and https://github.com/solana-labs/solana-program-library/blob/master/token-swap/js/client/token-swap.js#L312

token-swap/program/src/curve/stable.rs
ysavchenko: I propose to leave it as it is now and use the next PR to increase number of test cases and update to u256 math if needed (which is most likely)

token-swap/program/src/curve/stable.rs
ysavchenko: Do you mean add it to the pool params?

token-swap/program/src/curve/stable.rs
ysavchenko: Are there any outstanding issues here?

token-swap/program/src/curve/stable.rs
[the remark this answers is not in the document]
ysavchenko: I propose leaving it to the next PR where we will have more tests and can do some optimization.

token-swap/program/src/curve/stable.rs
```diff
@@ -0,0 +1,467 @@
...
+    let n_coins_sum = n_coins.checked_add(1)?.checked_mul(d_p)?;
+
+    let r_val = leverage_sub.checked_add(n_coins_sum)?;
+
+    l_val.checked_div(r_val)
+}
+
+/// Compute stable swap invariant (D)
+/// Equation:
+/// A * sum(x_i) * n**n + D = A * D * n**n + D**(n+1) / (n**n * prod(x_i))
```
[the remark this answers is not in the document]
ysavchenko: As I understand, the math is fine for now, right?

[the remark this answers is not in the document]
joncinque: Yes all good

token-swap/program/src/curve/stable.rs
joncinque: nope all good

token-swap/program/src/curve/stable.rs
joncinque: yes eventually, doesn't need to be done in this round though

joncinque:
Great, thanks for the work!

PR #840, opened 2020-11-14
Vendor ouroboros 0.4.1

mvines:
Author of the ouroboros crate yanked 0.4.1 and broke us.  See https://github.com/solana-labs/solana/pull/13555

Once the monorepo recovers, we can bring new Solana crates into SPL and revert this PR


mvines:
grrr, fine.  I'll just wait for the v1.4.8 crates to ship

PR #842, opened 2020-11-14
spl-token-cli: Use config commitment for balance checks

### LINKED ISSUE #837: spl-token-cli: inconsistent commitment
The spl-token Cli defaults to singleGossip commitment for transaction confirmation. However, balance checks and such are made using the solana-client default (max) commitment. This is a bit annoying when running a string of cli commands, since you have to remember which ones to wait until max on.

Proposed Solution:
Audit all client requests, and ensure commitment is consistent

CriesofCarrots:
SPL Token cli uses max commitment for balance checks. This is annoying and unexpected, since everything else defaults to singleGossip commitment.

Fixes #837


CriesofCarrots:
spl-token-lending failure seems unrelated to this PR

PR #846, opened 2020-11-16
[js-token-library] Fix create set authority ix module files

paul-schaaf:


paul-schaaf:
@jstarry hi :) pls review

jstarry (approved):
Thanks!

PR #857, opened 2020-11-17
Added negative tests for stake pool

ysavchenko:
Here we add a bunch of negative tests (where we expect errors to happen) to the stake pool smart contract.

This change set was waiting for its turn before interface to stake account program was added for test purposes, so it does not incorporate those changes.

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -645,6 +645,7 @@ mod tests {
         instruction::{initialize_account, initialize_mint},
         processor::Processor as TokenProcessor,
         state::{Account as SplAccount, Mint as SplMint},
+        error::TokenError,
     };
 
     /// Test program id for the stake-pool program.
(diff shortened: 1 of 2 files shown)


stake-pool/program/src/processor.rs
```diff
@@ -1048,57 +1170,211 @@ mod tests {
...
+        let withdraw_info = do_withdraw(&mut test_data);
+
+        assert_eq!(
+            withdraw_info.result,
+            Err(Error::InvalidProgramAddress.into())
+        );
+    }
+    #[test]
+    fn negative_test_withdraw_all() {
+        //
```
joncinque, line 1345: nit: empty comment

stake-pool/program/src/processor.rs
```diff
@@ -1107,42 +1383,40 @@ mod tests {
...
+            stake_balance,
         );
 
-        let stake_to_receive_key = Pubkey::new_unique();
-        let mut stake_to_receive_account =
-            Account::new(stake_balance, STAKE_ACCOUNT_LEN, &stake_program_id());
-
         let _result = do_process_instruction(
-            withdraw(
+            claim(
```
joncinque, line 1404: nit: The rest of the tests using `do_withdraw` and `do_deposit` are really easy to read -- any chance you could make a `do_claim` function to refactor a lot of this logic?

joncinque (commented):
Looks really good!  Just two tiny nits which you can take or leave

joncinque:
Looks great! I've kicked off the tests again -- once they pass I'll merge.  Thanks for the work

PR #875, opened 2020-11-19
token-swap: Add fuzzer for swap / withdraw / deposit

joncinque:
Using libfuzzer / cargo-fuzz (https://github.com/rust-fuzz/cargo-fuzz), this adds fuzz instruction testing in token-swap, following the example over at serum (https://github.com/project-serum/serum-dex/tree/master/dex/fuzz).

Still left to do:

- [x] add limited runs to CI using the `-max_total_time` flag
- [x] debug owner mismatch crashes, not sure what's there yet


token-swap/program/src/instruction.rs
```diff
@@ -29,13 +82,7 @@ pub enum SwapInstruction {
...
     ///   supply.  Must be empty, not owned by $authority.
     ///   7. '[]` Token program id
-    Initialize {
-        /// nonce used to create valid program address
-        nonce: u8,
-        /// swap curve info for pool, including CurveType, fees, and anything
-        /// else that may be required
-        swap_curve: SwapCurve,
-    },
+    Initialize(Initialize),
```
mvines, line 85: Kinda sad to see this.  Arbitrary doesn't work over enums?

joncinque: I totally agree.  This is actually long-standing rust issue that enum variants don't count as types.  I didn't follow the whole discussion over here, but I know there were efforts to make that happen: https://github.com/rust-lang/rust/issues/40666

joncinque: the other option is to have the fuzz instructions live separately, but I preferred having them re-use the existing instructions.  I can be overruled on that though

mvines: Separately sounds a lot worse.   Maybe there'll be other options when using Borsh encoding since there we have the ability to programmatically figure the types, even from an enum, and could use that information to construct the fuzzer inputs

joncinque: Oh yeah that's a great point.  I'll look into possibilities for how to integrate borsh with the libfuzzer inputs and ditch Arbitrary

mvines: sweet, `feature-proposal` is already borshed up and its instruction interface is really small.  It'd be a good guinea pig for Borsh-based fuzzing.  (cc:  https://github.com/solana-labs/solana-program-library/blob/master/feature-proposal/program/src/borsh_utils.rs).   Might be slightly quicker to iterate on than token-swap for experiments.

joncinque: good point, I'll take a look at it next then :-)

joncinque:
I'll have this run somewhere over the weekend, but so far, results are good.  Here's 4 threads running for nearly 3 hours without any problems:
```
stat::number_of_executed_units: 39009
stat::average_exec_per_sec:     243
stat::new_units_added:          405
stat::slowest_unit_time_sec:    0
stat::peak_rss_mb:              451
INFO: exiting: 2 time: 9609s
```

token-swap/program/fuzz/fuzz_targets/instructions.rs
```diff
@@ -0,0 +1,258 @@
...
+
+/// Use u8 as an account id to simplify the address space and re-use accounts
+/// more often.
+type AccountId = u8;
+
+const INITIAL_SWAP_TOKEN_A_AMOUNT: u64 = 100_000_000_000;
+const INITIAL_SWAP_TOKEN_B_AMOUNT: u64 = 300_000_000_000;
+
+const INITIAL_USER_TOKEN_A_AMOUNT: u64 = 1_000_000_000;
+const INITIAL_USER_TOKEN_B_AMOUNT: u64 = 3_000_000_000;
```
joncinque, line 57: I'm not sure if these numbers should be higher to allow for more extensive testing -- when doing random numbers for swapping and depositing, they're much higher than these

token-swap/program/fuzz/fuzz_targets/instructions.rs
```diff
@@ -0,0 +1,258 @@
...
+            trade_fee_numerator,
+            trade_fee_denominator,
+            owner_trade_fee_numerator,
+            owner_trade_fee_denominator,
+            owner_withdraw_fee_numerator,
+            owner_withdraw_fee_denominator,
+            host_fee_numerator,
+            host_fee_denominator,
+        }),
+    };
```
joncinque, line 84: We can eventually add more curves or even fuzz the curves, but that would be an initialize test as well

token-swap/program/fuzz/fuzz_targets/instructions.rs
```diff
@@ -0,0 +1,258 @@
...
+        .map_err(|e| {
+            if e == SwapError::CalculationFailure.into() {
+            } else if e == SwapError::ConversionFailure.into() {
+            } else if e == SwapError::FeeCalculationFailure.into() {
+            } else if e == SwapError::ExceededSlippage.into() {
+            } else if e == SwapError::ZeroTradingTokens.into() {
+            } else if e == TokenError::InsufficientFunds.into() {
+            } else {
+                Err(e).unwrap()
+            }
```
joncinque, line 231: This allows certain errors to slide, since they don't mean anything truly wrong has happened.  Note that these checks do run the risk of overlap, since `ProgramError::Custom(u32)` can re-use the same value for `TokenError` and `SwapError`.  We should consider changing it to `ProgramError::Custom(Pubkey, u32)` to separate better

token-swap/program/fuzz/fuzz_targets/instructions.rs
```diff
@@ -0,0 +1,258 @@
...
+    },
+    error::SwapError,
+    instruction::{Deposit, Swap, Withdraw},
+};
+
+use spl_token::error::TokenError;
+
+use arbitrary::Arbitrary;
+use libfuzzer_sys::fuzz_target;
+use std::collections::{HashMap, HashSet};
```
joncinque, line 18: Other than these dependencies, which can be removed if desired, I think this could actually run on bpf as well, so we can make a `fuzz-bpf`command too!

token-swap/program/fuzz/src/lib.rs
```diff
@@ -0,0 +1,586 @@
...
+            }
+        }
+
+        spl_token::processor::Processor::process(
+            &instruction.program_id,
+            &new_account_infos,
+            &instruction.data,
+        )
+    }
+}
```
joncinque, line 57: This is lifted directly from the processor test code, not sure how we can share this better.

token-swap/program/fuzz/src/lib.rs
```diff
@@ -0,0 +1,586 @@
...
+
+#[derive(Clone)]
+pub struct AccountData {
+    pub key: Pubkey,
+    pub lamports: u64,
+    pub data: Vec<u8>,
+    pub program_id: Pubkey,
+    pub is_signer: bool,
+}
```
joncinque, line 124: This is a bit hacky, but encompasses account data in an easy-to-use way.  Is there a better solution for this?

joncinque:
The concept of these fuzzing tests is to run fuzzed withdraw / swap / deposit instructions on a hard-coded curve, allow for certain errors (like Insufficient funds and calculation errors), and at the end check that we haven't created or deleted any tokens, and that we can also withdraw everything from the token swap. also tagging @aeyakovenko if he has any additional ideas for conditions to check.

joncinque:
I'd like to leave the CI for a follow-up PR, and I'll keep running the fuzzing over the weekend

joncinque:
After running this over the whole weekend, it only caught an error in the fuzzer logic!

joncinque:
I've tried getting this to run in bpf, but we'll need nightly for that, specifically for `Arbitrary`, which uses the `non_exhaustive` feature for its errors

joncinque:
I've tried using honggfuzz  in [2997203](https://github.com/solana-labs/solana-program-library/pull/875/commits/2997203de73ecd4064ad18280d6367a86ace1029).  Specifically, a version that doesn't use `#[non_exhaustive]` from `arbitrary`, which trips up the current bpf builder.  There would still need to be some clever work to make a `fuzz-bpf` command though.  I'll do some performance diffs between libfuzzer and honggfuzz (current and old version) to see which is the winner.

joncinque:
Converting back to draft as I try honggfuzz some more

ci/install-build-deps.sh
```diff
@@ -10,4 +10,6 @@ sudo apt-get install -y openssl --allow-unauthenticated
 sudo apt-get install -y libssl-dev --allow-unauthenticated
 sudo apt-get install -y libssl1.1 --allow-unauthenticated
 sudo apt-get install -y libudev-dev
+sudo apt-get install -y binutils-dev
+sudo apt-get install -y libunwind-dev
```
joncinque, line 14: these can also get added to `fuzz.sh` if preferred

ci/install-program-deps.sh
```diff
@@ -9,6 +9,7 @@ set -x
 
 cargo --version
 cargo install rustfilt || true
+cargo install honggfuzz || true
```
joncinque, line 12: this can get added to `fuzz.sh` as well

joncinque:
A nice note for `on.schedule` workflows at https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/disabling-and-enabling-a-workflow:
> Warning: To prevent unnecessary workflow runs, scheduled workflows may be disabled automatically. When a public repository is forked, scheduled workflows are disabled by default. In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days. 
Meaning people who have forked the repo *won't* run the nightly fuzzer

shared-memory/program/tests/shared-memory.rs
```diff
@@ -26,7 +25,7 @@ fn run_program(
     program_id: &Pubkey,
     parameter_accounts: &[KeyedAccount],
     instruction_data: &[u8],
-) -> Result<u64, InstructionError> {
+) -> u64 {
```
joncinque, line 28: This came from a new clippy error

token-swap/program/fuzz/src/instructions.rs
```diff
@@ -0,0 +1,288 @@
...
+
+fn main() {
+    loop {
+        fuzz!(|fuzz_instructions: Vec<FuzzInstruction>| {
+            run_fuzz_instructions(fuzz_instructions)
+        });
+    }
+}
+
+fn run_fuzz_instructions(fuzz_instructions: Vec<FuzzInstruction>) {
```
mvines, line 69: nit: avoid requiring a `Vec` when a slice will do
```suggestion
fn run_fuzz_instructions(fuzz_instructions: &[FuzzInstruction]) {
```

token-swap/program/src/processor.rs
```diff
@@ -550,7 +553,8 @@ impl Processor {
                 to_u128(token_a.amount)?,
             )
             .ok_or(SwapError::ZeroTradingTokens)?;
-        if a_amount < to_u128(minimum_token_a_amount)? {
+        let a_amount = to_u64(a_amount)?;
```
mvines, line 556: Was this a bug fix as a result of fuzzing?

mvines (approved):
Looks good overall!   I have some concerns about the amount of boilerplate that seems to be required to fuzz but as a v1 there's no need for this PR to continue to dangle.   This'll be a good example for other programs at least, and hopefully inspiration for future improvements to streamline fuzzing.

Errr, so did we find any actual bugs in token-swap though?

token-swap/program/src/processor.rs
joncinque: It actually wasn't, just an inconsistency in the code from when we transitioned to `u128`

joncinque:
@mvines thanks for the review.  I ended up starting on the banks version, and we can definitely refactor some logic that's also used in the stake pool end-to-end testing in a `token/client` library.  In the end, no bugs were discovered in token-swap from a few days of fuzzing, but maybe the nightly runs will discover something!

PR #879, opened 2020-11-20
Use RpcClient in Themis

garious:
#### Problem

Not clear if BanksClient outperforms RpcClient. Also, not sure if we want to continue supporting two RPC clients.

#### Proposed changes

* Reimplement the Themis client using RpcClient
* So far TPS is ~3 times *lower* than BanksClient. More tuning to come.


joncinque:
Closing this since SPL hasn't had themis for some time

PR #886, opened 2020-11-23
token-swap stable curve: 256-bit math added with tests

ysavchenko:
- 256 bit math added to stable curve calculation
- Reference curve.fi simulation python script added (with interface by @michaelhly)
- Range tests added to check math against reference simulation


token-swap/program/src/curve/stable.rs
```diff
@@ -82,53 +107,50 @@ fn compute_d(amp: u128, amount_a: u128, amount_b: u128) -> Option<u128> {
...
-                if d.checked_sub(d_previous)? <= 1 {
-                    break;
-                }
-            } else if d_previous.checked_sub(d)? <= 1 {
+            if almost_equal(&d, &d_previous) {
                 break;
             }
         }
-        Some(d)
+        Some(d.as_u128())
```
joncinque, line 116: This panics in the case of an overflow, so can you use `TryFrom` instead?

token-swap/program/src/curve/stable.rs
```diff
@@ -82,53 +107,50 @@ fn compute_d(amp: u128, amount_a: u128, amount_b: u128) -> Option<u128> {
...
-            }
-        } else if y_prev.checked_sub(y)? <= 1 {
+        y = (checked_u8_power(&y, 2)?.checked_add(c)?)
+            .checked_div(checked_u8_mul(&y, 2)?.checked_add(b)?.checked_sub(d_val)?)?;
+        if almost_equal(&y, &y_prev) {
             break;
         }
     }
-    Some(y)
+    Some(y.as_u128())
```
joncinque, line 153: Please use `TryFrom` here as well

token-swap/program/src/curve/stable.rs
```diff
@@ -59,20 +86,18 @@ fn calculate_step(
...
-        let mut d_previous: u128;
-        let mut d = sum_x;
-        let leverage = amp.checked_mul(n_coins)?; // A * n
-                                                  // Newton's method to approximate D
-        for _ in 0..128 {
+        let mut d_previous: U256;
+        let mut d: U256 = sum_x.into();
+
+        // Newton's method to approximate D
+        for _ in 0..32 {
```
joncinque, line 100: Did you do some instruction number checking for this, or is it based on our previous discussions?  I'd be curious to know how `U256` performs!

token-swap/program/src/curve/stable.rs
```diff
@@ -82,53 +107,50 @@ fn compute_d(amp: u128, amount_a: u128, amount_b: u128) -> Option<u128> {
...
+    let leverage: U256 = leverage.into();
+    let new_source_amount: U256 = new_source_amount.into();
+    let d_val: U256 = d_val.into();
 
     // sum' = prod' = x
     // c =  D ** (n + 1) / (n ** (2 * n) * prod' * A)
-    let c = checked_pow(d_val, n_coins.checked_add(1)? as usize)?.checked_div(
-        new_source_amount.checked_mul(checked_pow(n_coins, 2)?.checked_mul(leverage)?)?,
-    )?;
+    let c = checked_u8_power(&d_val, N_COINS + 1)?
```
joncinque, line 136: This is pedantic, but it helps with security reviews later -- can you change `N_COINS.checked_add(1)?`?

token-swap/program/src/curve/stable.rs
```diff
@@ -35,21 +40,43 @@ pub struct StableCurve {
...
+fn calculate_step(initial_d: &U256, leverage: u64, sum_x: u128, d_product: &U256) -> Option<U256> {
+    let leverage_mul = U256::from(leverage).checked_mul(sum_x.into())?;
+    let d_p_mul = checked_u8_mul(d_product, N_COINS)?;
 
-    let l_val = leverage_mul.checked_add(d_p_mul)?.checked_mul(initial_d)?;
+    let l_val = leverage_mul.checked_add(d_p_mul)?.checked_mul(*initial_d)?;
 
-    let leverage_sub = leverage.checked_sub(1)?.checked_mul(initial_d)?;
-    let n_coins_sum = n_coins.checked_add(1)?.checked_mul(d_product)?;
+    let leverage_sub = initial_d.checked_mul((leverage - 1).into())?;
```
joncinque, line 78: This is pedantic, but it helps with security reviews later -- can you do `leverage.checked_sub(1)?`?

token-swap/program/src/curve/stable.rs
```diff
@@ -35,21 +40,43 @@ pub struct StableCurve {
...
+        result = result.checked_add(*a)?;
+    }
+    Some(result)
+}
+
+fn almost_equal(a: &U256, b: &U256) -> bool {
+    if a > b {
+        a - b <= U256::one()
+    } else {
+        b - a <= U256::one()
```
joncinque, line 67: nit: I know that you do the check earlier to make sure that isn't a problem, but for security reviews, could you do checked math here?

token-swap/program/src/curve/stable.rs
```diff
@@ -35,21 +40,43 @@ pub struct StableCurve {
...
+    let leverage_mul = U256::from(leverage).checked_mul(sum_x.into())?;
+    let d_p_mul = checked_u8_mul(d_product, N_COINS)?;
 
-    let l_val = leverage_mul.checked_add(d_p_mul)?.checked_mul(initial_d)?;
+    let l_val = leverage_mul.checked_add(d_p_mul)?.checked_mul(*initial_d)?;
 
-    let leverage_sub = leverage.checked_sub(1)?.checked_mul(initial_d)?;
-    let n_coins_sum = n_coins.checked_add(1)?.checked_mul(d_product)?;
+    let leverage_sub = initial_d.checked_mul((leverage - 1).into())?;
+    let n_coins_sum = checked_u8_mul(&d_product, N_COINS + 1)?;
```
joncinque, line 79: checked math here as well please

token-swap/program/src/curve/stable.rs
```diff
@@ -147,18 +169,17 @@ impl CurveCalculator for StableCurve {
             .checked_sub(trade_fee)?
             .checked_sub(owner_fee)?;
 
+        let leverage = self.amp * N_COINS as u64;
```
joncinque, line 172: Checked math here as well please

token-swap/program/src/curve/stable.rs
```diff
@@ -412,20 +434,84 @@ mod tests {
...
-            ..Default::default()
-        };
-
-        let result = curve
-            .swap(source_amount, swap_source_amount, swap_destination_amount)
-            .unwrap();
-        assert_eq!(result.new_source_amount, 1100);
-        assert_eq!(result.amount_swapped, 2083);
-        assert_eq!(result.new_destination_amount, 47917);
+        const POOL_AMOUNTS: &[u128] = &[
```
joncinque, line 437: this looks great! I can't wait to get this integrated in the fuzzer :-D

token-swap/program/src/curve/stable.rs
```diff
@@ -412,20 +434,84 @@ mod tests {
...
+                        };
+
+                        if *source_amount >= *swap_source_amount {
+                            continue;
+                        }
+
+                        println!(
+                            "trying: source_amount={}, swap_source_amount={}, swap_destination_amount={}", 
+                            source_amount, swap_source_amount, swap_destination_amount
+                        );
```
joncinque, line 478: nit: You've done a great job with the print in the assert, so this can be removed

token-swap/program/src/curve/stable.rs
```diff
@@ -412,20 +434,84 @@ mod tests {
...
+                            *swap_destination_amount,
+                        );
+
+                        let result = result.unwrap();
+                        let sim_result = model.sim_exchange(0, 1, *source_amount);
+
+                        println!(
+                            "result={}, sim_result={}",
+                            result.amount_swapped, sim_result
+                        );
```
joncinque, line 498: nit: Same with this

joncinque (commented):
Looks really close, just a few organization nits and checked math questions!

token-swap/program/src/curve/stable.rs
```diff
@@ -1,17 +1,22 @@
 //! The curve.fi invariant calculator.
+#![allow(clippy::assign_op_pattern)]
+#![allow(clippy::ptr_offset_with_cast)]
+#![allow(clippy::unknown_clippy_lints)]
+#![allow(clippy::manual_range_contains)]
```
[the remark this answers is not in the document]
ysavchenko: Well, all these warning come from `construct_uint!` macro, just annotating the line before the macro does not help, so I disabled these warnings for the whole file.
One solution could be to move U256 definition into a separate module and disable all those warnings there.
What do you think? Is there any other solution?

token-swap/program/src/curve/stable.rs
ysavchenko: Here is the method I used: added all those range tests, got an overflow error and moved parts of the code to U256.
Once the overflow was resolved I reduced the number of iterations from 128, to 64, then to 32, then 16. On 16 tests broke so I kept 32.

token-swap/program/src/curve/stable.rs
ysavchenko: Well, this is a constant now and its value is 2, do we really need checked_add there? No problem adding it, just curious :)

token-swap/program/src/curve/stable.rs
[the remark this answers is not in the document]
joncinque: Ah crap, that stinks.  I like the idea of moving it to another file though

token-swap/program/src/curve/stable.rs
joncinque: awesome, that sounds reasonable. we'll see how this does against the fuzzing!

token-swap/program/src/curve/stable.rs
joncinque: I know it's totally pedantic, but it will absolutely help with security reviews

token-swap/program/lib/sim/src/lib.rs
```diff
@@ -0,0 +1,207 @@
...
+        self.extract_py_ret(py, py_ret)
+    }
+
+    fn call1(
+        &self,
+        py: Python,
+        method_name: &str,
+        args: impl IntoPy<Py<PyTuple>>,
+    ) -> Result<PyObject, PyErr> {
+        let sim = PyModule::from_code(py, &self.py_src, FILE_NAME, MODULE_NAME).unwrap();
```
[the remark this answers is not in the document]
ysavchenko: I do not think so, each call it acquires interpreter lock and then initializes module from source. May be there is a space for optimization, but not much. And the calls themselves are working quite fast, so I would not worry about this.

token-swap/program/src/curve/stable.rs
```diff
@@ -412,20 +434,84 @@ mod tests {
...
+                        let sim_result = model.sim_exchange(0, 1, *source_amount);
+
+                        println!(
+                            "result={}, sim_result={}",
+                            result.amount_swapped, sim_result
+                        );
+                        let diff = (sim_result as i128 - result.amount_swapped as i128).abs();
+
+                        assert!(
+                            diff <= 1,
```
michaelhly, line 502: We should be checking `diff == 0` right? Probably should fix the off-by-one errors?

ysavchenko: Well, I think it is ok to have errors by 1, after all, this is one lamport error. And we ignore it the rest works great on a wide range of inputs. But this is just my opinion.

token-swap/program/src/curve/stable.rs
```diff
@@ -109,11 +77,11 @@ fn compute_d(leverage: u64, amount_a: u128, amount_b: u128) -> Option<u128> {
             //d = (leverage * sum_x + d_p * n_coins) * d / ((leverage - 1) * d + (n_coins + 1) * d_p);
             d = calculate_step(&d, leverage, sum_x, &d_product)?;
             // Equality with the precision of 1
-            if almost_equal(&d, &d_previous) {
+            if d.almost_equal(&d_previous)? {
                 break;
             }
         }
-        Some(d.as_u128())
+        Some(u128::try_from(d).ok()?)
```
joncinque, line 84: You can just do `u128::try_from(d).ok()` here

joncinque (commented):
One last tiny thing, then we can merge in -- thanks for the quick turnaround!

PR #895, opened 2020-11-26
Multi-account withdraw added the stake-pool CLI

ysavchenko:
Now if there is not enough funds on the single stake account CLI will withdraw from several and then join them into one single account.

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -1092,7 +1092,7 @@ mod tests {
     }
 
     fn pick_with_balance(
-        test_list: &Vec<(Pubkey, Account)>,
+        test_list: &[(Pubkey, Account)],
         balance: u64,
         withdraw: u64,
     ) -> WithdrawAccount {
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -1083,67 +1083,83 @@ mod tests {
         let mut result: Vec<(Pubkey, Account)> = Vec::with_capacity(lamports.len());
 
         for (index, balance) in lamports.iter().enumerate() {
-            result.push(
-                (
-                    Pubkey::new_from_array([index as u8; 32]),
-                    Account::new(*balance, 0, &PK_OWNER),
-                )
-            );
+            result.push((
+                Pubkey::new_from_array([index as u8; 32]),
+                Account::new(*balance, 0, &PK_OWNER),
+            ));
         }
         result
     }
 
-    fn pick_with_balance(test_list: &Vec<(Pubkey, Account)>, balance: u64, withdraw: u64) -> WithdrawAccount {
+    fn pick_with_balance(
+        test_list: &Vec<(Pubkey, Account)>,
+        balance: u64,
+        withdraw: u64,
+    ) -> WithdrawAccount {
         for (pubkey, account) in test_list {
             if account.lamports == balance {
                 return WithdrawAccount {
                     pubkey: *pubkey,
                     account: account.clone(),
-                    amount: withdraw
+                    amount: withdraw,
                 };
             }
         }
 
         panic!();
     }
-    
+
     #[test]
     fn test_pick_withdraw_accounts() {
         let mut test_list = lamports_to_accounts(&[2, 7, 8, 5, 3, 10]);
-        assert_eq!(pick_withdraw_accounts(&mut test_list, 5).unwrap(), vec![
-            pick_with_balance(&test_list, 5, 5),
-        ]);
+        assert_eq!(
+            pick_withdraw_accounts(&mut test_list, 5).unwrap(),
+            vec![pick_with_balance(&test_list, 5, 5),]
+        );
 
         let mut test_list = lamports_to_accounts(&[2, 7, 8, 5, 3, 10]);
-        assert_eq!(pick_withdraw_accounts(&mut test_list, 13).unwrap(), vec![
-            pick_with_balance(&test_list, 10, 10),
-            pick_with_balance(&test_list, 3, 3),
-        ]);
+        assert_eq!(
+            pick_withdraw_accounts(&mut test_list, 13).unwrap(),
+            vec![
+                pick_with_balance(&test_list, 10, 10),
+                pick_with_balance(&test_list, 3, 3),
+            ]
+        );
 
         let mut test_list = lamports_to_accounts(&[2, 7, 8, 5, 3, 10]);
-        assert_eq!(pick_withdraw_accounts(&mut test_list, 14).unwrap(), vec![
-            pick_with_balance(&test_list, 5, 4),    // Partial should always be the first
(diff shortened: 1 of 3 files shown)


stake-pool/cli/src/main.rs
```diff
@@ -391,12 +391,80 @@ fn pool_tokens_to_stake_amount(pool_data: &StakePool, tokens: u64) -> u64 {
...
+        // Those accounts will be withdrawn completely with `claim` instruction
+        withdraw_from.push(WithdrawAccount {
+            pubkey: *pubkey,
+            account: account.clone(),
+            amount: account.lamports,
+        });
+        remaining_amount -= account.lamports;
+        accounts.pop();
+    }
+    if remaining_amount > 0 {
```
joncinque, line 429: Won't this always be true?

stake-pool/cli/src/main.rs
```diff
@@ -391,12 +391,80 @@ fn pool_tokens_to_stake_amount(pool_data: &StakePool, tokens: u64) -> u64 {
...
+    }
+    // Sort from lowest to highest balance
+    accounts.sort_by(|a, b| a.1.lamports.cmp(&b.1.lamports));
+
+    // Prepare the list of accounts to withdraw from
+    let mut withdraw_from: Vec<WithdrawAccount> = vec![];
+    let mut remaining_amount = amount;
+
+    // First use largest amounts
+    while !accounts.is_empty() && accounts.last().unwrap().1.lamports < remaining_amount {
```
joncinque, line 417: I'm finding the two loops a bit confusing, so maybe we could add some comments about what's being done. Here we iterate from the back to get large accounts, then we iterate from the front to get small accounts. What's the purpose of each one?  Perhaps we could get it all done in one iteration?

stake-pool/cli/src/main.rs
```diff
@@ -433,131 +501,124 @@ fn command_withdraw(
...
                 &spl_token::id(),
                 &stake_program_id(),
-                amount,
             )?);
-        }
 
-        let mut transaction =
-            Transaction::new_with_payer(&instructions, Some(&config.fee_payer.pubkey()));
+            // Merge into stake_receiver
+            if let Some(merge_into) = stake_receiver {
```
joncinque, line 600: If, for example, we don't provide `stake_receiver`, and our withdraw results in claiming two accounts and withdrawing from one, will this work as intended?  `stake_receiver` might not be defined at this point, but we might still want to merge into the final account.  Is there something in the sorting that makes this work out?  If so, we should clarify that somewhere

stake-pool/program/src/processor.rs
```diff
@@ -410,6 +410,18 @@ impl Processor {
             stake_program_info.clone(),
         )?;
 
+        Self::stake_authorize(
```
joncinque, line 413: nice catch -- did you notice this from testing?

joncinque (commented):
looks really close, just a couple of questions to make sure we're on the same page

stake-pool/cli/src/main.rs
ysavchenko: Yes, the error was `<` instead of `<=` in the condition above, then it could be zero. As per your next comment I'll better join two loops together.

stake-pool/cli/src/main.rs
ysavchenko: Mostly because we already had the second loop and added first one to get the rest of the accounts. :). You're right it is better to make them into one loop.

stake-pool/cli/src/main.rs
ysavchenko: Ok, this is a bit confusing.
If we have at lease one `withdraw` command then it will be first in the list (as per this comment) and it will definitely create `stake_receiver`:
`// This account will be withdrawn partially (if remaining_amount < account.lamports), so we put it first`
However if we do only `claim`s and do not specify `stake_receiver` then the stakes will not be joined.

So probably we need to rework params a bit. May be make this a required param? Or if we are in `claim`-only situation and have more than one account then merge all of them to the first one?

stake-pool/program/src/processor.rs
ysavchenko: Yes and then verified with the stake program source code. We need Staker permission for a lot of things...

stake-pool/cli/src/main.rs
ysavchenko: I thought about it for a while and will do the second option (if there are only bunch of `claim`s then the first one of them is used to merge everything to).

stake-pool/cli/src/main.rs
ysavchenko: This one it outdated now, will describe what was done in the next comment.

stake-pool/cli/src/main.rs
ysavchenko: So now there is only one loop and it seems like there is more code now, but I think it looks prettier and with more comments now.
So the goal is to catch the following:

For accounts [10, 8, 7, 5, 3, 2]:
Withdraw 5 will return single claim of [5]
Withdraw 13 returns [10, 3]
Withdraw 14 returns [10, 4 of 5]
Withdraw 19 returns [10, 8, 1 of 2]

So it is not the most effective algorithm now, but should be enough.

stake-pool/cli/src/main.rs
```diff
@@ -406,47 +406,59 @@ fn prepare_withdraw_accounts(
...
+            });
+            remaining_amount -= account.lamports;
+            if remaining_amount == 0 {
+                // We filled all the balance, ignore split candidate
+                split_candidate = None;
+                break;
+            }
+        } else {
+            // Save the last account with balance more than remaining as a candidate for split
+            split_candidate = Some(WithdrawAccount {
```
joncinque, line 446: For your example of "Withdraw 19 returns [10, 8, 1 of 2]", wouldn't `split_candidate` get overwritten on the second iteration by the account with 8?  Does the logic from 433-443 need to get done here too?

Actually, could you add some tests for all the cases you've laid out?  Then we'll be sure that everything works as intended

joncinque (commented):
Looks close -- I'm thinking some small test cases will help stabilize this

stake-pool/cli/src/main.rs
ysavchenko: No, it is not, iterations will go like this:
1. Add 10 to the list, remaining is 9
2. Add 8 to the list, remaining is 8
3. Split candidate set to 7, continue
4. Split candidate is 5
5. Split candidate is 3
6. Spit candidate is 2
Now the cycle is over so we take 1 from the last split candidate.

ysavchenko: But you're right, I also thought we should cover it with tests at least for those cases I specified, will do.

joncinque: Yes, you're right, my mistake!

joncinque:
This looks great -- thanks so much!  Once it passes CI I'll merge it in :-)

Issue #693, opened 2020-10-22
token-swap: Token program id is stored in swap info but is unused


jstarry:
#### Problem
Token program id is saved on swap info state but is never used for token account validation

#### Proposed Changes
- Remove token program id from swap info state
- Check that token a and token b both belong to the specified token program id when processing the init instruction

tag @joncinque

joncinque:
Fixed with #903

PR #903, opened 2020-11-30
token-swap: Cleanup check

joncinque:


joncinque:
This now takes care of everything in #693

jstarry:
@joncinque I think a good convention would be to always check the account owner when unpacking external accounts. We have this pattern for sysvars now and we could apply it to swap's `unpack_token_account` and `unpack_mint` methods too. 

It would also be nice to establish the pattern of being careful about what accounts are passed to invoked instructions but that feels less critical

joncinque:
@jstarry that's a good point, I'll do that here.  To help other programs avoid this, what would you say to expanding `Pack` to include an `unpack` variant which takes in `AccountInfo` and an expected program id?  We can then rename the current `unpack` to `unpack_raw` or something better.

joncinque:
@jstarry this is ready for another look whenever you get the chance

jstarry (commented):
I opened a PR that applies my feedback here: https://github.com/joncinque/solana-program-library/pull/2

I thought that it was too confusing to test the curves when fees are involved so I separated that out and added more edge case tests for the rounding.

All the program id checks look good!

PR #922, opened 2020-12-03
token-swap: Add constant price and offset curves

joncinque:
This encompasses a lot of changes to properly support the constant price
curve, and fixes some old pain points to make the whole thing easier to
use, including:
* refactor fees out of the curve
* move all fee calculations into new Fees type
* modify constraints to only care about curve types and fee
* add price parameter into constant price curve (flat curve)
* add offset curve


joncinque:
It's a bit of a doozy, mainly because all of the duplicated fees have been moved out of the curve and into one place.  I was going to hold off on doing this until later, but the `FeeConstraints` became a huge mess since it was just checking for curve equality, which doesn't work when there are parameters that can be set in the curve.  I thought of some hacky workarounds, but they all seemed much worse than biting the bullet and just doing the refactor.  I'll point out some important things to look for

token-swap/js/cli/token-swap-test.js
```diff
@@ -241,6 +240,7 @@ export async function createTokenSwap(): Promise<void> {
     OWNER_WITHDRAW_FEE_DENOMINATOR,
     HOST_FEE_NUMERATOR,
     HOST_FEE_DENOMINATOR,
+    CURVE_TYPE,
```
joncinque, line 243: From the outside this is the only change -- fees come before the curve

token-swap/program/src/constraints.rs
```diff
@@ -20,44 +22,29 @@ use std::env;
...
-                    curve_type: swap_curve.curve_type,
-                    calculator: Box::new(x.clone()),
-                })
-                .collect(),
-        };
-        if valid_swap_curves.iter().any(|x| *x == *swap_curve) {
+        if self
+            .valid_curve_types
+            .iter()
+            .any(|x| *x == swap_curve.curve_type)
```
joncinque, line 37: Now we only check for the curve type, but this could be expanded later

token-swap/program/src/constraints.rs
```diff
@@ -20,44 +22,29 @@ use std::env;
 pub struct FeeConstraints<'a> {
     /// Owner of the program
     pub owner_key: &'a str,
-    /// Valid flat curves for the program
-    pub valid_flat_curves: &'a [FlatCurve],
-    /// Valid constant product curves for the program
-    pub valid_constant_product_curves: &'a [ConstantProductCurve],
-    /// Valid stable curves for the program
-    pub valid_stable_curves: &'a [StableCurve],
+    /// Valid curve types
```
joncinque, line 25: Even this part gets neater

token-swap/program/src/curve/base.rs
```diff
@@ -20,9 +24,28 @@ pub enum CurveType {
...
+    ConstantPrice,
     /// Stable, Like uniswap, but with wide zone of 1:1 instead of one point
     Stable,
+    /// Offset, Like uniswap, but with the token B side having a faked offset
+    Offset,
+}
+
+/// Encodes all results of swapping from a source token to a destination token
+#[derive(Debug, PartialEq)]
+pub struct SwapResult {
```
joncinque, line 36: This was moved, since `SwapWithoutFeesResult` covers the calculator now

token-swap/program/src/curve/base.rs
```diff
@@ -37,6 +60,47 @@ pub struct SwapCurve {
...
+impl SwapCurve {
+    /// Subtract fees and calculate how much destination token will be provided
+    /// given an amount of source token.
+    pub fn swap(
+        &self,
+        source_amount: u128,
+        swap_source_amount: u128,
+        swap_destination_amount: u128,
+        trade_direction: TradeDirection,
+        fees: &Fees,
```
joncinque, line 72: Two new parameters to watch for here -- the fees are done from the outside now, so it made more sense to just move this into the `SwapCurve` and have the calculator stay totally fee-free

token-swap/program/src/curve/base.rs
```diff
@@ -167,16 +218,124 @@ mod tests {
...
-        packed.extend_from_slice(&owner_withdraw_fee_denominator.to_le_bytes());
-        packed.extend_from_slice(&host_fee_numerator.to_le_bytes());
-        packed.extend_from_slice(&host_fee_denominator.to_le_bytes());
         packed.extend_from_slice(&0u64.to_le_bytes()); // 8 bytes reserved for larget calcs
         let unpacked = SwapCurve::unpack_from_slice(&packed).unwrap();
         assert_eq!(swap_curve, unpacked);
     }
+
+    #[test]
+    fn constant_product_trade_fee() {
```
joncinque, line 227: Since fees are outside the curve, the fee tests were moved outside

token-swap/program/src/curve/calculator.rs
```diff
@@ -176,9 +79,6 @@ pub trait CurveCalculator: Debug + DynPack {
             .and_then(map_zero_to_none)
     }
 
-    /// Calculate the host fee based on the owner fee, only used in production
-    /// situations where a program is hosted by multiple frontends
-    fn host_fee(&self, _owner_fee: u128) -> Option<u128> {
-        Some(0)
-    }
+    /// Validate that the given curve has no bad parameters
+    fn validate(&self) -> Result<(), SwapError>;
```
joncinque, line 83: Currently used for bad offsets or constant prices.  We'll have to look into stable swap for bad amp factors, but that can be for another PR

token-swap/program/src/curve/constant_price.rs
```diff
@@ -0,0 +1,215 @@
...
+                let destination_amount_swapped = source_amount.checked_div(token_b_price)?;
+                let mut source_amount_swapped = source_amount;
+
+                // if there is a remainder from buying token B, floor
+                // token_a_amount provided to avoid taking too many tokens, but
+                // don't recalculate the fees
+                let remainder = source_amount_swapped.checked_rem(token_b_price)?;
+                if remainder > 0 {
+                    source_amount_swapped = source_amount.checked_sub(remainder)?;
+                }
```
joncinque, line 46: Similar concept to the rounding done for constant product

token-swap/program/src/curve/constant_product.rs
```diff
@@ -5,31 +5,50 @@ use solana_program::{
...
-    /// Owner withdraw fee denominator
-    pub owner_withdraw_fee_denominator: u64,
-    /// Host trading fee numerator
-    pub host_fee_numerator: u64,
-    /// Host trading fee denominator
-    pub host_fee_denominator: u64,
+pub struct ConstantProductCurve {}
+
+/// The constant product swap calculation, factored out of its class for reuse
+pub fn swap(
```
joncinque, line 20: Nothing changed here, just refactored to be used by OffsetCurve

token-swap/program/src/curve/constant_product.rs
```diff
@@ -115,59 +76,18 @@ impl IsInitialized for ConstantProductCurve {
...
+    const LEN: usize = 0;
     fn pack_into_slice(&self, output: &mut [u8]) {
         (self as &dyn DynPack).pack_into_slice(output);
     }
 
-    fn unpack_from_slice(input: &[u8]) -> Result<ConstantProductCurve, ProgramError> {
-        let input = array_ref![input, 0, 64];
-        #[allow(clippy::ptr_offset_with_cast)]
-        let (
-            trade_fee_numerator,
```
joncinque, line 127: Deleting all of the `_fee_numerator` and `_fee_denominator` felt very nice

token-swap/program/src/curve/fees.rs
```diff
@@ -1,51 +1,77 @@
-//! Simple constant 1:1 swap curve
+//! All fee information, to be used for validation currently
```
joncinque, line 1: Yikes, I'm not sure why git thinks flat.rs -> fees.rs.  This file is new, and flat.rs became constant_price.rs

token-swap/program/src/curve/offset.rs
```diff
@@ -0,0 +1,194 @@
...
+        let token_b_offset = self.token_b_offset as u128;
+        let swap_source_amount = match trade_direction {
+            TradeDirection::AtoB => swap_source_amount,
+            TradeDirection::BtoA => swap_source_amount.checked_add(token_b_offset)?,
+        };
+        let swap_destination_amount = match trade_direction {
+            TradeDirection::AtoB => swap_destination_amount.checked_add(token_b_offset)?,
+            TradeDirection::BtoA => swap_destination_amount,
+        };
+        swap(source_amount, swap_source_amount, swap_destination_amount)
```
joncinque, line 42: Simply call into constant product's swap after increasing the appropriate values artificially

token-swap/program/src/curve/offset.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    fn validate(&self) -> Result<(), SwapError> {
+        if self.token_b_offset == 0 {
```
joncinque, line 46: Technically this is fine, but people should just make a normal constant product curve if they don't want any offset

token-swap/program/src/curve/stable.rs
```diff
@@ -197,61 +150,22 @@ impl IsInitialized for StableCurve {
...
-    const LEN: usize = 72;
+    const LEN: usize = 8;
     fn pack_into_slice(&self, output: &mut [u8]) {
         (self as &dyn DynPack).pack_into_slice(output);
     }
 
     fn unpack_from_slice(input: &[u8]) -> Result<StableCurve, ProgramError> {
-        let input = array_ref![input, 0, 72];
-        #[allow(clippy::ptr_offset_with_cast)]
-        let (
```
joncinque, line 208: I can't stress how nice it was to delete all this

token-swap/program/src/instruction.rs
```diff
@@ -134,8 +136,18 @@ impl SwapInstruction {
         Ok(match tag {
             0 => {
                 let (&nonce, rest) = rest.split_first().ok_or(SwapError::InvalidInstruction)?;
-                let swap_curve = SwapCurve::unpack_unchecked(rest)?;
-                Self::Initialize(Initialize { nonce, swap_curve })
+                if rest.len() >= Fees::LEN {
+                    let (fees, rest) = rest.split_at(Fees::LEN);
```
joncinque, line 140: `split_at` panics, so we do the check right before

token-swap/program/src/processor.rs
```diff
@@ -236,7 +237,10 @@ impl Processor {
                 return Err(SwapError::InvalidOwner.into());
             }
             fee_constraints.validate_curve(&swap_curve)?;
+            fee_constraints.validate_fees(&fees)?;
         }
+        fees.validate()?;
+        swap_curve.calculator.validate()?;
```
joncinque, line 243: I thought this was pretty neat

token-swap/program/src/curve/offset.rs
```diff
@@ -0,0 +1,194 @@
+//! The Uniswap invariant calculator with an extra offset
```
jstarry, line 1: Can you provide more context on why this is useful? It seems like it makes it so you don't have to provide so much liquidity on one side of the pool and the offset effectively sets a minimum exchange price.

jstarry (commented):
I think offset curves allows the B token pool supply to go to zero and constant price curves allow either side to go to zero. If that happens, withdraws and deposits become disabled because pool tokens to B tokens will map to zero and return an error. Maybe it should still be possible to deposit / withdraw if one side of the pool is zero?

token-swap/program/src/curve/fees.rs
```diff
@@ -1,51 +1,77 @@
...
-use crate::curve::calculator::{calculate_fee, CurveCalculator, DynPack, SwapWithoutFeesResult};
-use arrayref::{array_mut_ref, array_ref, array_refs, mut_array_refs};
 use std::convert::TryFrom;
 
-/// FlatCurve struct implementing CurveCalculator
+/// Encapsulates all fee information and calculations for swap operations
 #[derive(Clone, Debug, Default, PartialEq)]
-pub struct FlatCurve {
-    /// Fee numerator
+pub struct Fees {
```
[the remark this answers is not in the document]
joncinque: Very good point.  I wanted to add an additional check to make sure that `trade_fee + owner_trade_fee < 1`, and a separate type would be perfect for things like that

token-swap/program/src/curve/constant_product.rs
```diff
@@ -115,59 +76,18 @@ impl IsInitialized for ConstantProductCurve {
 }
 impl Sealed for ConstantProductCurve {}
 impl Pack for ConstantProductCurve {
```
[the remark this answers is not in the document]
joncinque: That makes sense, and definitely avoids a footgun for later

token-swap/program/src/instruction.rs
```diff
@@ -134,8 +136,18 @@ impl SwapInstruction {
...
+                    let fees = Fees::unpack_unchecked(fees)?;
+                    let swap_curve = SwapCurve::unpack_unchecked(rest)?;
```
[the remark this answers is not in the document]
joncinque: True, we can make the check in `Pack` a bit different to return `AccountDataTooSmall` or create another ProgramError for `IncorrectAccountDataSize`

token-swap/program/src/processor.rs
```diff
@@ -236,7 +237,10 @@ impl Processor {
                 return Err(SwapError::InvalidOwner.into());
             }
             fee_constraints.validate_curve(&swap_curve)?;
+            fee_constraints.validate_fees(&fees)?;
```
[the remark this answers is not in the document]
joncinque: From what I heard, they want the fees to be the same for all pools.  I'll double-check

token-swap/program/src/curve/offset.rs
joncinque: You've got it exactly right -- it's specifically geared towards new exchange offerings that don't require tons of liquidity on the other side.  For example, if I want to sell 1 million of my new token, and I think each is worth 2 USDC, I make this pool with an offset of 2 million USDC, so I can start selling the token at that rate

token-swap/program/src/processor.rs
```diff
@@ -3978,6 +4113,126 @@ mod tests {
...
+            _pool_key,
+            _pool_account,
+        ) = accounts.setup_token_accounts(&user_key, &swapper_key, initial_a, initial_b, 0);
+
+        // swap a to b way, fails, there's no liquidity
+        let a_to_b_amount = initial_a;
+        let minimum_token_b_amount = 0;
+
+        assert_eq!(
+            Err(SwapError::ZeroTradingTokens.into()),
```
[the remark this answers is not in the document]
joncinque: The problem actually came from the trading token to pool token calculation post-swap -- I wasn't calculating from the new source token amount, which meant we were producing some bad zeros.  I've made the test better to actually accept 0s now

token-swap/program/src/instruction.rs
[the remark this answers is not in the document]
jstarry: Awkward because they are unpacked for instruction data not accounts

token-swap/program/src/curve/base.rs
```diff
@@ -37,6 +60,70 @@ pub struct SwapCurve {
...
+            trade_fee,
+            owner_fee,
+        })
+    }
+
+    /// Calculate the pool token equivalent of the owner fee on trade
+    /// See the math at: https://balancer.finance/whitepaper/#single-asset-deposit
+    /// For the moment, we do an approximation for the square root.  For numbers
+    /// just above 1, simply dividing by 2 brings you very close to the correct
+    /// value.
```
jstarry, line 107: These comments seem outdated

joncinque:
Closing this since work was split up into #930 #934 #935 #936 #937

PR #930, opened 2020-12-07
token-swap: Refactor for new curves

joncinque:
This encompasses a lot of changes to properly support the constant price
curve, and fixes some old pain points to make the whole thing easier to
use, including:
* refactor fees out of the curve
* move all fee calculations into new Fees type
* modify constraints to only care about curve types and fee
* update interface for conversions between pool tokens and trading tokens

Note that this is entirely a refactor, no appreciable changes to functionality, so the risk for this is low.


token-swap/program/src/processor.rs
```diff
@@ -2057,21 +2096,27 @@ mod tests {
...
-                valid_constant_product_curves,
-                valid_flat_curves: &[],
-                valid_stable_curves: &[],
+                valid_curve_types,
+                fees: &fees,
             });
-            let mut accounts =
-                SwapAccountInfo::new(&user_key, swap_curve, token_a_amount, token_b_amount);
+            let mut bad_fees = fees.clone();
+            bad_fees.trade_fee_numerator = trade_fee_numerator;
```
jstarry, line 2112: This is a bit confusing. How about instead of decrementing the trade fee numerator for the fee constraints, you deprecate it for the bad fees object?

token-swap/program/src/curve/fees.rs
```diff
@@ -1,51 +1,77 @@
...
+        if fee == 0 {
+            Some(1) // minimum fee of one token
+        } else {
+            Some(fee)
+        }
+    }
+}
+
+fn validate_fraction(numerator: u64, denominator: u64) -> Result<(), SwapError> {
+    if numerator > denominator {
```
jstarry, line 65: Can you think of a case where fees should be 100%?

token-swap/program/src/processor.rs
```diff
@@ -3971,7 +4065,8 @@ mod tests {
         let host_fee_account = spl_token::state::Account::unpack(&pool_account.data).unwrap();
         let owner_fee_account =
             spl_token::state::Account::unpack(&accounts.pool_fee_account.data).unwrap();
-        let total_fee = host_fee_account.amount * host_fee_denominator / host_fee_numerator;
+        let total_fee = owner_fee_account.amount * host_fee_denominator
+            / (host_fee_denominator - host_fee_numerator);
```
jstarry, line 4069: This made me realize that host fees probably should never be 100%

jstarry (dismissed):
Looks good, mostly minor nits

token-swap/js/client/token-swap.js
```diff
@@ -279,7 +280,6 @@ export class TokenSwap {
     tokenProgramId: PublicKey,
     swapProgramId: PublicKey,
     nonce: number,
-    curveType: number,
```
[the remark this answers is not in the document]
joncinque: Since this is a new major version, I'd prefer to make this change now since it maps naturally to the layout of the structure

token-swap/program/src/curve/fees.rs
joncinque: Good point, most of the other logic would break because of 0s at that point anyway

token-swap/program/src/processor.rs
joncinque: haha thankfully it's just a shorthand for the test

joncinque:
I ran all of the failing CI scripts locally, and they passed, so merging this on red.

PR #931, opened 2020-12-07
Test criterion in CI

jackcmay:


jackcmay:
This was only a test, real changes here: https://github.com/solana-labs/solana/pull/14002

PR #934, opened 2020-12-08
token-swap: Add pool token conversion interface

joncinque:
In order to do proper conversions between pool tokens and trading tokens (A/B), we need the full amounts of token A and B in the swap, so this adds an interface to be used by the new curves.


joncinque:
`cargo-build-test` and `js-test-token-swap` run successfully locally

jstarry (approved):

PR #936, opened 2020-12-09
token-swap: Constant price curve

joncinque:
The constant price is a flat line, where the amount of token A required to buy 1 token B is fixed on init.  This makes the value function for the pool very simple:
`value = token_a_amount + token_b_amount * token_b_price`, so single deposit calculations are done with respect to that number.

~~Note that this depends on #934 , so the only substantial changes are in `constant_price.rs` and some more tests in `processor.rs`.~~
The PR has been rebased for easier reading.

### DIFF
--- a/token-swap/program/src/curve/base.rs
+++ b/token-swap/program/src/curve/base.rs
@@ -110,8 +110,14 @@ impl SwapCurve {
         trade_direction: TradeDirection,
         fees: &Fees,
     ) -> Option<u128> {
-        // Get the trading fee incurred if the owner fee is swapped for the other side
-        let trade_fee = fees.trading_fee(source_amount)?;
+        // Get the trading fee incurred if *half* the source amount is swapped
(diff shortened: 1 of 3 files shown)


token-swap/program/src/curve/base.rs
```diff
@@ -110,8 +110,14 @@ impl SwapCurve {
         trade_direction: TradeDirection,
         fees: &Fees,
     ) -> Option<u128> {
-        // Get the trading fee incurred if the owner fee is swapped for the other side
-        let trade_fee = fees.trading_fee(source_amount)?;
+        // Get the trading fee incurred if *half* the source amount is swapped
```
joncinque, line 113: This is unrelated, but still important to fix

CriesofCarrots (approved):

joncinque:
cargo-build-test and js-test-token-swap pass locally, merging on red

PR #943, opened 2020-12-11
token-swap: Refactor math into sqrt function and add proptest

joncinque:
In order to restrict the potential problem surface of the new math functions, we expose only a `sqrt` function from `PreciseNumber` and add a property test to be sure that the calculation works as expected.

This implements the recommended test: "To claim that the function f approximates sqrt with a relative error of at most epsilon, then your property-based test should assert that (f(x)*(1-epsilon))^2 < x < (f(x)*(1+epsilon))^2 for all x in the range of inputs that f is documented as accepting."

### DIFF
--- a/token-swap/program/src/curve/math.rs
+++ b/token-swap/program/src/curve/math.rs
@@ -51,7 +51,7 @@ pub const MAX_WEIGHT: u8 = 100;
 pub const MIN_WEIGHT: u8 = 1;
 
 /// Struct encapsulating a fixed-point number that allows for decimal calculations
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct PreciseNumber {
     /// Wrapper over the inner value, which is multiplied by ONE
     pub value: U256,
@@ -133,11 +133,6 @@ impl PreciseNumber {
         difference.value < precision
     }
 
-    /// Checks that two PreciseNumbers are equal
-    pub fn eq(&self, rhs: &Self) -> bool {
-        self.value == rhs.value
-    }
-
     /// Checks that a number is less than another
     pub fn less_than(&self, rhs: &Self) -> bool {
         self.value < rhs.value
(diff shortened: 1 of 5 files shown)


token-swap/program/src/curve/math.rs
```diff
@@ -323,11 +367,46 @@ impl PreciseNumber {
...
+        {
+            return None;
+        }
+        let two = PreciseNumber::new(2)?;
+        let one = PreciseNumber::new(1)?;
+        let guess = if self.value > one.value {
+            self.checked_div(&two)? // guess half of number
+        } else {
+            self.checked_add(&one)?.checked_div(&two)? // guess average of 1 and number
+        };
```
CriesofCarrots, line 401: Can you add some additional comments here to explain the guess-selection logic?

token-swap/program/src/curve/math.rs
```diff
@@ -345,55 +424,57 @@ mod tests {
...
+        check_pow_approximation(one / 2, one * 4 / 50, U256::from(946057646730u128));
+        // 0.94605764673
     }
 
     fn check_pow_fraction(base: U256, exponent: U256, expected: U256, precision: U256) {
         let base = PreciseNumber { value: base };
         let exponent = PreciseNumber { value: exponent };
         let power = base.checked_pow_fraction(&exponent).unwrap();
         let expected = PreciseNumber { value: expected };
+        println!("calculated: {} expected: {}", power.value, expected.value);
```
CriesofCarrots, line 443: Intended?

CriesofCarrots (approved):
Generally, looks good. Just one place I'd like a little more documentation.
Out of curiosity, how long does the prop test take to run?

token-swap/program/src/curve/math.rs
joncinque: oops

joncinque:
The prop test goes very quickly!  It only runs 256 test cases by default: https://altsysrq.github.io/proptest-book/proptest/tutorial/config.html?#configuring-the-number-of-tests-cases-requried  -- they're designed to work more on the level of unit tests, as opposed to fuzzing which runs indefinitely

PR #952, opened 2020-12-15
token-swap: Add proptests for decreasing curve value

joncinque:
This set of proptests checks that a swap never takes value out of the liquidity pools, and re-uses the successful swap conditions developed for #950.

In order to check that value has not been lost, this also adds a simple `total_value` function to the calculator, which defaults to the constant product invariant.  The offset and constant price curves have different ways of figuring out the total value.

Note that the total value result is meant to be unique to each curve type, and not to be used for comparing against other curves.

Also, I tried implementing this for `StableCurve`, but the tests failed.  It seems we need to do some more work to properly support it first.

### DIFF
--- a/token-swap/program/src/curve/calculator.rs
+++ b/token-swap/program/src/curve/calculator.rs
@@ -1,6 +1,9 @@
 //! Swap calculations
 
-use crate::{curve::math::PreciseNumber, error::SwapError};
+use crate::{
+    curve::math::{PreciseNumber, U256},
+    error::SwapError,
+};
 use std::fmt::Debug;
 
 /// Initial amount of pool tokens for swap contract, hard-coded to something
@@ -165,6 +168,16 @@ pub trait CurveCalculator: Debug + DynPack {
     fn allows_deposits(&self) -> bool {
         true
     }
+
+    /// Calculates the total value of the curve given the liquidity parameters.
+    ///
+    /// This is mostly useful for invariant testing the curves, to make sure
+    /// that value is not lost on any trade.
+    ///
+    /// The default implementation for this function gives the Uniswap invariant
+    fn total_value(&self, swap_token_a_amount: u128, swap_token_b_amount: u128) -> Option<U256> {
+        U256::from(swap_token_a_amount).checked_mul(U256::from(swap_token_b_amount))
+    }
 }
 
 /// Test helpers for curves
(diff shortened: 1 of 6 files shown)


token-swap/program/src/curve/calculator.rs
```diff
@@ -145,6 +148,16 @@ pub trait CurveCalculator: Debug + DynPack {
...
         true
     }
+
+    /// Calculates the total value of the curve given the liquidity parameters.
+    ///
+    /// This is mostly useful for invariant testing the curves, to make sure
+    /// that value is not lost on any trade.
+    ///
+    /// The default implementation for this function gives the Uniswap invariant
+    fn total_value(&self, swap_token_a_amount: u128, swap_token_b_amount: u128) -> Option<U256> {
```
jstarry, line 158: Would this be more accurately called something like "normalized value" or "squared value" or "proportional_value"?

token-swap/program/src/curve/calculator.rs
```diff
@@ -145,6 +148,16 @@ pub trait CurveCalculator: Debug + DynPack {
...
+        U256::from(swap_token_a_amount).checked_mul(U256::from(swap_token_b_amount))
```
jstarry, line 159: Looks like token amounts can only be as high as u64::MAX so I don't think u256 is needed here

token-swap/program/src/curve/offset.rs
```diff
@@ -282,4 +293,64 @@ mod tests {
...
+        fn curve_value_does_not_decrease_from_swap_a_to_b(
+            source_token_amount in 1..u64::MAX,
+            swap_source_amount in 1..u64::MAX,
+            swap_destination_amount in 1..u64::MAX,
+            token_b_offset in 1..u64::MAX,
+        ) {
+            // The invariant needs to fit in a u128, so keep token b side
+            // under u64::MAX.
+            // invariant = swap_source_amount * (swap_destination_amount + token_b_offset)
+            prop_assume!(!swap_destination_amount.overflowing_add(token_b_offset).1);
```
jstarry, line 308: Could this assumption reflect the invariant requirement directly?

token-swap/program/src/curve/calculator.rs
```diff
@@ -225,4 +238,45 @@ pub mod test {
...
+            .unwrap();
+        let (swap_token_a_amount, swap_token_b_amount) = match trade_direction {
+            TradeDirection::AtoB => (new_swap_source_amount, new_swap_destination_amount),
+            TradeDirection::BtoA => (new_swap_destination_amount, new_swap_source_amount),
+        };
+
+        let new_value = curve
+            .total_value(swap_token_a_amount, swap_token_b_amount)
+            .unwrap();
+        assert!(new_value >= previous_value);
```
jstarry, line 280: It feels worth adding an error bounds for this test too. We wouldn't want a regression that causes the invariant to bump too much because that would indicate that swaps are too lossy for traders

token-swap/program/src/curve/calculator.rs
joncinque: I didn't want to drag in `PreciseNumber` and get the normalized value, but now that you mention this, there's no reason we should be able to take a square root since this is only used in tests.  Let's go with `normalized_value` in that case

token-swap/program/src/curve/calculator.rs
joncinque: This brought up another little bug in `PreciseNumber` -- we needed a max allowable base of `u128::MAX * ONE`

jstarry (approved):

PR #953, opened 2020-12-16
Port token js tests from solana-localnet to solana-test-validator

mvines:
Replace `solana-localnet` with `solana-test-validator` in Token JS tests.

Depends-on: Solana 1.5.0


mvines:
@jstarry / @joncinque - I only ported token/js over to `solana-test-simulator`, it would be nice to drop solana-localnet from the other JS in SPL too.  Not a big rush but I think it's nicer.   The lack of macOS binaries for 1.5.0 is annoying for local dev though, hopefully I'll be able to work that out with Travis CI soon :-/

@jackcmay - the BPF VM signatures changed in 1.5 so I just commented out the instruction counting tests for now.  I feel like reworking those to be `solana-program-test`-based would be a nicer path for the future anyway.  Although `solana-program-test` doesn't allow you to get an exact instruction count currently, you can put a cap on the total compute units so I feel that is close enough.  I feel the BPF VM API is something internal as well, so by not using it in SPL we don't implicitly promote this to Solana devs looking for some examples to copy from

PR #958, opened 2020-12-16
token-swap: Add fuzz check for pool token value

joncinque:
This adds an extra check in the instruction fuzzer, where we calculate the total value of pool tokens before and after the instructions.  Their relative value should never decrease, ie. `A_total_start / P_total_start <= A_total_end / P_total_end` should always hold.  This also showed that the deposit math was incorrect, and we should *not* be adding the amount coming in on a deposit.  See https://balancer.finance/whitepaper/#all-asset-deposit-withdrawal for the applicable formula.

### DIFF
--- a/token-swap/program/src/curve/calculator.rs
+++ b/token-swap/program/src/curve/calculator.rs
@@ -1,6 +1,6 @@
 //! Swap calculations
 
-use crate::{curve::math::PreciseNumber, error::SwapError};
+use crate::{curve::math::{ceiling_division, PreciseNumber}, error::SwapError};
 use std::fmt::Debug;
 
 /// Initial amount of pool tokens for swap contract, hard-coded to something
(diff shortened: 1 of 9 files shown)


token-swap/program/src/curve/calculator.rs
```diff
@@ -99,12 +102,14 @@ pub trait CurveCalculator: Debug + DynPack {
         swap_token_a_amount: u128,
         swap_token_b_amount: u128,
     ) -> Option<TradingTokenResult> {
-        let token_a_amount = pool_tokens
-            .checked_mul(swap_token_a_amount)?
-            .checked_div(pool_token_supply)?;
-        let token_b_amount = pool_tokens
-            .checked_mul(swap_token_b_amount)?
-            .checked_div(pool_token_supply)?;
+        let (token_a_amount, _) = ceiling_division(
```
jstarry, line 105: Should ceiling division be applied to the other `pool_tokens_to_trading_tokens` methods too?

joncinque: Yes absolutely, great catch

jstarry (approved):

PR #965, opened 2020-12-21
lending: Add owner and host fees

joncinque:
In order to get a production release, we'll need owner and host fees, similar to serum and swap.  To get that done, we'll need:

- [x] Fee info in reserve packing
- [x] Initialize reserve with fee info
- [x] New accounts in instructions
- [x] Integrate fees in repay
- [x] Integrate fees functionality in liquidate


joncinque:
This is ready for integration now.  Note that it's bare bones and does not have specific checks to make sure that the fees are processed correctly.  I manually checked to make sure that the fees were sent during the tests.

The main concept is to assess fees whenever collateral is returned, on repay or liquidate.  The fee is configured on the reserve on initialization, and allows you to specify how much the owner will receive for repayment and liquidation, expressed in basis points.  There's also a host fee percentage field, which will send a portion of the fees to some other account passed in during repay and liquidate.

Main changes:
* another token account and fee info must be passed on reserve init
* two more collateral token accounts (the owner account and the host account) must be passed for liquidate and repay 

A few particularities to note:
* since the collateral mint is created during reserve initialization, we can't prepare an account beforehand for the owner fee.  To get around that, the fee account owner is the lending market -- since a signature was produced for the lending market, the reserve creator can always change the owner of the fee account after init
* since a host fee account may not always be provided, it's possible to send in `Pubkey::default()` as the host fee account to signify that there is no host fee account

jstarry (commented):
Looks really solid! Main feedback is that I'm wondering why fees are assessed on repay/liquidate vs on borrows. I think borrow might be better because it achieves the same effect, the fees will be available sooner, and you then only need to implement them in one instruction.

joncinque:
The concept of taking fees on repay / liquidate was to mirror how fees are distributed to depositors -- if you're a depositor, your tokens go up in value when a borrower repays, and not on deposit.  I'm happy to go about it either way though.  Maybe @bartosz-lipinski has some feedback as well?

token-lending/program/src/processor.rs
```diff
@@ -828,6 +858,26 @@ fn process_repay(
...
+            1,
+            collateral_withdraw_amount * repay_fee_rate / TOTAL_BASIS_POINTS,
+        )
+    } else {
+        0
+    };
+    let host_fee_rate = withdraw_reserve.config.fees.host_fee_percentage as u64;
+    let host_fee =
+        if host_fee_rate > 0 && *withdraw_reserve_collateral_host_info.key != Pubkey::default() {
+            std::cmp::max(1, repay_fee * host_fee_rate / 100)
```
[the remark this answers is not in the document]
joncinque: That's right, I wanted to prioritize getting the host a fee since theirs is smaller generally, but we can do this the other way around.

jstarry:
Since the fees are collected in the form of collateral tokens, we don't have to think about interest at all here. The collateral tokens are interest bearing and the fee collector can choose how long to wait (for interest to accrue) to withdraw just like depositors can.

Also, I think UX is better if the user knows up front how much they will be paying in fees for a loan rather than being faced with it at the end when they are forced to pay back the loan

joncinque:
Ok, that works.  Do you also want to remove the fee on liquidation?

jstarry:
Keeping fees for liquidation sounds fine to me

joncinque:
Moved to only borrow fees per our discussion, expressed as a Wad.  I also updated the fees to avoid the 0 owner fee situation, and instead we assess a minimum of 2 tokens if we have a host and owner.

jstarry:
Looks very solid, do you mind adding unit tests for the fees?

joncinque:
I added a proptest for the fee calculation, which uncovered an overflow bug!  Otherwise I added some additional init tests and a check for the fees in the main borrow test.  Let me know if you want anything else for this PR

token-lending/program/src/state.rs
```diff
@@ -525,3 +596,87 @@ fn pack_decimal(decimal: Decimal, dst: &mut [u8; 16]) {
...
+    use super::*;
+    use crate::math::WAD;
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn borrow_fee_calculation(
+            borrow_fee_wad in 0..WAD,
+            host_fee_percentage in 0..100u8,
+            borrow_amount in 0..u64::MAX,
```
jstarry, line 611:
```suggestion
            borrow_fee_wad in 0..=WAD,
            host_fee_percentage in 0..=100u8,
            borrow_amount in 0..=u64::MAX,
```

token-lending/program/tests/borrow.rs
```diff
@@ -91,10 +92,20 @@ async fn test_success() {
...
-                amount: INITIAL_COLLATERAL_RATE * USER_SOL_COLLATERAL_LAMPORTS / 2,
+                amount: borrow_amount / 2,
                 user_accounts_owner: &user_accounts_owner,
                 obligation: Some(obligation),
             },
         )
         .await;
+
+    // check that fee accounts have been properly credited
+    let (total_fee, host_fee) = TEST_RESERVE_CONFIG
```
jstarry, line 103: Can you also add a sanity test that host and protocol fees are both non-zero?

token-lending/program/src/state.rs
```diff
@@ -525,3 +596,87 @@ fn pack_decimal(decimal: Decimal, dst: &mut [u8; 16]) {
...
+            assert!(total_fee <= borrow_amount);
+            // the host fee can't be greater than the total fee
+            assert!(host_fee <= total_fee);
+
+            // for all fee rates greater than 0, we must have some fee
+            if borrow_fee_wad > 0 {
+                assert!(total_fee > 0);
+            }
+
+            let host_fee_percentage = host_fee_percentage as u64;
```
jstarry, line 632: nit: I don't think this conversion is needed

token-lending/program/src/state.rs
```diff
@@ -525,3 +596,87 @@ fn pack_decimal(decimal: Decimal, dst: &mut [u8; 16]) {
...
+            if host_fee_percentage == 100 {
+                // if the host fee percentage is maxed at 100%, it should get all the fee
+                assert_eq!(host_fee, total_fee);
+            }
+
+            // if there's a host fee, it must be greater than 0
+            if host_fee_percentage > 0 {
```
jstarry, line 639: Can you also add a test that checks that host fees aren't too high relative to total fees?

token-lending/program/src/state.rs
```diff
@@ -525,3 +596,87 @@ fn pack_decimal(decimal: Decimal, dst: &mut [u8; 16]) {
...
+                assert!(host_fee > 0);
```
jstarry, line 640: This won't pass if borrow_fee_wad == 0

token-lending/program/src/state.rs
```diff
@@ -525,3 +596,87 @@ fn pack_decimal(decimal: Decimal, dst: &mut [u8; 16]) {
...
+            let fees = ReserveFees {
+                borrow_fee_wad,
+                host_fee_percentage,
+            };
+            let (total_fee, host_fee) = fees.calculate_borrow_fees(
+                borrow_amount,
+                &Pubkey::new_unique(),
+            );
+
+            // the total fee can't be greater than the amount borrowed
```
jstarry, line 622: It actually can if borrow_amount <= 1

jstarry:
Nice, what was the overflow bug?

token-lending/program/src/state.rs
joncinque: Very good catch, I should've been doing this with all of the others too

token-lending/program/src/state.rs
joncinque: Nice catch

token-lending/program/src/state.rs
joncinque: There's a check of `assert!(host_fee <= total_fee);` a few lines above -- does that cut it?

token-lending/program/src/state.rs
joncinque: Kind of worrying that they don't perform that test case in the normal proptesting :-\

joncinque:
The host fee calculation was overflowing since it wasn't using a `Rate` and simply doing `fee * host_fee_percentage / 100`.

joncinque:
Thanks for all the feedback, I'm a bit worried that the proptesting didn't catch a few of the edge cases, so I added some specific tests for those.

token-lending/program/src/state.rs
jstarry: Yeah, I was surprised about that too. I was under the impression that it prioritized the mins and maxes

token-lending/program/src/state.rs
jstarry: The spirit of my suggestion is that if host fees were always equal to the total fee, all your tests would still pass

token-lending/program/src/state.rs
```diff
@@ -33,6 +33,57 @@ pub struct LendingMarket {
...
+        let host_fee_rate = Rate::from_percent(self.host_fee_percentage);
+        if borrow_fee_rate > Rate::zero() && collateral_amount > 0 {
+            let need_to_assess_host_fee =
+                host_fee_rate > Rate::zero() && *host_pubkey != Pubkey::default();
+            let minimum_fee = if need_to_assess_host_fee {
+                2 // 1 token to owner, 1 to host
+            } else {
+                1 // 1 token to owner, nothing else
+            };
+            let borrow_fee = std::cmp::max(
```
jstarry, line 71: I think we should return an error if `borrow_fee >= collateral_amount`

token-lending/program/src/instruction.rs
```diff
@@ -94,20 +99,24 @@ pub enum LendingInstruction {
...
-    ///   12 `[]` Dex market
-    ///   13 `[]` Dex market order book side
-    ///   14 `[]` Temporary memory
-    ///   15 `[]` Clock sysvar
-    ///   16 `[]` Rent sysvar
-    ///   17 '[]` Token program id
+    ///   4. `[writable]` Deposit reserve collateral fee receiver account.
+    ///                     Must be the fee account specified at InitReserve.
+    ///   5. `[writable]` Deposit reserve collateral host fee receiver account.
+    ///                     Can be any account or the empty account at `Pubkey::default`
```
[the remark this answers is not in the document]
joncinque: No problem, will do.  I wasn't aware about the `Pubkey::default` behavior -- thanks for flagging that!

jstarry (approved):

PR #985, opened 2020-12-30
token-swap: Add version to state

joncinque:
Since we now have an upgradeable loader, we'll need to properly handle any updates to state structures.  The approach here is to use the excellent `enum_dispatch` crate https://docs.rs/enum_dispatch/0.3.4/enum_dispatch/ to have multiple structs implement the same interface, and then use these structs as trait objects in processors.  This seemed like the best approach to minimize code bloat in the current case.  Note that this won't cover every possible case for backwards compatibility.  For example, we might end up adding a field that won't have a good backwards-compatible default, requiring us to have different versions of instruction processors.  Let's hope that doesn't happen too soon.

Since this approach requires dynamic dispatch on trait objects, I measured the number of BPF instructions before and after this PR on every swap instruction, along with the number of dynamic dispatch calls in each program instruction.  Here are the results:

|Dyn Calls|Instruction Name|BPF Instructions Before|BPF Instructions After|
| --- | --- | --- | --- |
|30|Swap|60250|60516|
|17|WithdrawSingleTokenTypeExactAmountOut|64934|65168|
|15|DepositSingleTokenTypeExactAmountIn|57298|57505|
|15|WithdrawAllTokenTypes|37469|37667|
|13|DepositAllTokenTypes|29754|29953|
|0|Init|17928|17362|

There's a small impact of ~250 BPF instructions, so this could be a viable approach for all programs! Any suggestions or comments are very appreciated.

cc @aeyakovenko @jstarry @mvines

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -934,6 +934,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "enum_dispatch"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "15b748aeb1b8c58d317739aa3171a2752ca3e76b8077c13dc748f327aebc0a72"
+dependencies = [
+ "once_cell",
+ "proc-macro2 1.0.24",
+ "quote 1.0.7",
+ "syn 1.0.48",
+]
+
+[[package]]
 name = "enumflags2"
 version = "0.6.4"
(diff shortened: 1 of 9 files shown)


bartosz-lipinski:
Awesome, I am adding that to client side right now.

PR #1039, opened 2021-01-07
lending: Fix fee collection for borrows

jstarry:
#### Problem
Fees are subtracted from `collateral_deposit_amount` before depositing the collateral into the reserve. Later, fees are paid from the reserve which causes the actual collateral deposited to be slightly less than it should.

#### Solution
- Perform collateral deposit and fee payment from the user account
- Add test

### DIFF
--- a/token-lending/program/tests/borrow.rs
+++ b/token-lending/program/tests/borrow.rs
@@ -104,11 +104,18 @@ async fn test_success() {
         .fees
         .calculate_borrow_fees(borrow_amount)
         .unwrap();
-    let sol_fee_balance =
-        get_token_balance(&mut banks_client, sol_reserve.collateral_fees_receiver).await;
+
     assert!(total_fee > 0);
     assert!(host_fee > 0);
-    assert_eq!(total_fee - host_fee, sol_fee_balance);
+
+    let sol_collateral_supply =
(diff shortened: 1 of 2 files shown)


jstarry:
@joncinque I needed to merge this to unblock another PR but could you take a look at this?

token-lending/program/tests/borrow.rs
```diff
@@ -104,11 +104,18 @@ async fn test_success() {
...
-    let sol_fee_balance =
-        get_token_balance(&mut banks_client, sol_reserve.collateral_fees_receiver).await;
+
     assert!(total_fee > 0);
     assert!(host_fee > 0);
-    assert_eq!(total_fee - host_fee, sol_fee_balance);
+
+    let sol_collateral_supply =
+        get_token_balance(&mut banks_client, sol_reserve.collateral_supply).await;
+    assert_eq!(sol_collateral_supply, borrow_amount - total_fee);
```
jstarry, line 113: nit: `borrow_amount` is misnamed, it should be `collateral_amount`. Will fix in a follow up

joncinque:
Thanks a lot for catching this.  I must have been in the mode of thinking about repay / liquidate fees, which were coming from the reserve.  How did you find it?

PR #1043, opened 2021-01-07
lending: Add version and padding to state structs

joncinque:
We haven't been able to discuss the best path for versioning structs just yet, but this is a simple alternative of just adding a version at the start of the struct and padding to be potentially used later.

Once `version` was added to `Obligation`, the build failed because of stack size violations in `process_borrow`, so I refactored the borrow amount calculation into a function on `TradeSimulator`.  I didn't see an obvious place to move the calculation -- I thought about putting it in `state.rs`, but then you pull in the TradeSimulator into state, which didn't seem like the right solution.  We can also implement it on `BorrowAmountType`.

Let me know what you think!

### DIFF
--- a/token-lending/program/src/state.rs
+++ b/token-lending/program/src/state.rs
@@ -18,6 +18,9 @@ use solana_program::{
 /// Collateral tokens are initially valued at a ratio of 5:1 (collateral:liquidity)
 pub const INITIAL_COLLATERAL_RATE: u64 = 5;
 
+/// Current version of the program and all new accounts created
+pub const PROGRAM_VERSION: u8 = 1;
+
 /// Number of slots per year
 pub const SLOTS_PER_YEAR: u64 =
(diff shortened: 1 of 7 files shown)


token-lending/program/src/state.rs
```diff
@@ -18,13 +18,18 @@ use solana_program::{
...
+
 /// Number of slots per year
 pub const SLOTS_PER_YEAR: u64 =
     DEFAULT_TICKS_PER_SECOND / DEFAULT_TICKS_PER_SLOT * SECONDS_PER_DAY * 365;
 
 /// Lending market state
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct LendingMarket {
+    /// Version of lending market
+    pub version: u8,
```
jstarry, line 32: nit: let's get rid of `is_initialized` and use version instead

jstarry (approved):
Looks good, just the one nit

PR #1047, opened 2021-01-08
swap: Move transfers into user account to the end of instructions

joncinque:
This changes the order of CPIs into the token program to always transfer towards the user at the end of instructions.

cc @jstarry


bartosz-lipinski (approved):
This looks awesome @joncinque

PR #1054, opened 2021-01-11
fuzz: Dump error file as hex instead of using cat

joncinque:
If there's a fuzz failure, the contents of the file are printed using `cat`, which isn't useful for binary files.  Instead, use `od -t x1` to get a hex output that's easy to use with `xxd -r` to reconstruct the binary file.  That way we can reliably reproduce the fuzz failure locally.  There are now more instructions included in the script as well.


joncinque:
Merging on red, error is unrelated, the fuzz part passed

PR #1055, opened 2021-01-12
Refine SPL Token wallet integration guide

mvines:
To improve transfer interoperability between wallets supporting the SPL Token transfer, the sender's wallet is now **required** to fund the recipient's SPL token account if it doesn't already exist.   

Previously this was just a recommendation, however this causes interoperability problems between wallets that choose to not follow the recommendation.


mvines:
Thanks @danpaul000, I kinda had that 🙃 didn't I

PR #1058, opened 2021-01-12
swap: Avoid fuzz 1 token withdrawal errors by using fee account

joncinque:
There are sporadic crashes on the nightly fuzzing because of the new "withdraw single token type" instruction.  In the fuzzing environment, the curve has a withdraw fee, which means we must provide a minimum of 2 pool tokens to the withdraw instruction, 1 for the fee, and 1 to withdraw token A / B.  There are checks in the other instructions to avoid getting down to 1 pool token, but "withdraw single token type" doesn't allow us to specify exactly how many pool tokens to withdraw, meaning that we can eventually have a situation where we have 1 pool token, causing a crash on the end calculation.  To get around that, this introduces a change to transfer all pool tokens to the fee account before the final withdrawal.  The fee account is not charged a withdrawal fee, so this makes the end calculation more stable.


joncinque:
Merging on red, unrelated test failure

PR #1061, opened 2021-01-13
swap: Add rounding correction for deposit / withdraw

joncinque:
The fuzzing test was improved to make sure that the value of pool tokens
(minus fees) never changed, which revealed a calc bug with deposits and
withdrawals.  If someone withdraws, they can take additional small
value out of the pool in the form of truncated value.

Before, deposit and withdrawal used the same calculation, rounding up the token
A and B amounts.  Remember that deposit / withdrawal specify the pool
token amount, and calculate the amount of A / B required to get to that
many pool tokens.  On withdrawal, we were giving back a little bit too
much.  The concept is that everything would even out since a deposit is
required for a withdrawal, which makes everything a wash.

Fuzzing found this issue very quickly! This change introduces a
rounding parameter to avoid ever giving away too much value, either
during deposit or withdrawal.


joncinque:
I tried to think of ways to break this up, but most of it hinges on the addition of the rounding correction, some refactoring to avoid duplication of logic in the offset curve, and a test for it.  I can isolate the fuzzing part in a separate PR, but it's essentially doing the same check as the withdraw test which is why I wanted to keep it in.  Let me know if you have any questions!

token-swap/program/src/curve/constant_price.rs
```diff
@@ -67,18 +67,34 @@ impl CurveCalculator for ConstantPriceCurve {
...
+                    .checked_div(pool_token_supply)?;
+                (token_a_amount, token_b_amount)
+            }
+            RoundDirection::Ceiling => {
+                let (token_a_amount, _) =
+                    ceiling_division(pool_tokens.checked_mul(total_value)?, pool_token_supply)?;
+                let (token_b_amount, _) = ceiling_division(
+                    pool_tokens
+                        .checked_mul(total_value)?
+                        .checked_div(token_b_price)?,
```
jstarry, line 92: Should this be ceiling division too?

token-swap/program/src/curve/calculator.rs
```diff
@@ -399,4 +375,57 @@ pub mod test {
...
+            )
+            .unwrap();
+        let new_swap_token_a_amount = swap_token_a_amount - withdraw_result.token_a_amount;
+        let new_swap_token_b_amount = swap_token_b_amount - withdraw_result.token_b_amount;
+        let new_pool_token_supply = pool_token_supply - pool_token_amount;
+
+        let pool_value = curve
+            .normalized_value(swap_token_a_amount, swap_token_b_amount)
+            .unwrap();
+        // since we can get rounding issues on the pool value which make it seem that the
```
jstarry, line 408: Is this because of the divide by 2? I was wondering if you could remove that and have the normalized value be the full total and then do the divide by 2 inside `pool_tokens_to_trading_tokens`. (I'm looking specifically at the constant price curve)

token-swap/program/src/curve/constant_price.rs
```diff
@@ -478,4 +505,58 @@ mod tests {
...
+                .pool_tokens_to_trading_tokens(
+                    pool_token_amount,
+                    pool_token_supply,
+                    swap_token_a_amount,
+                    swap_token_b_amount,
+                    RoundDirection::Floor,
+                )
+                .unwrap();
+            prop_assume!(withdraw_result.token_a_amount <= swap_token_a_amount);
+            prop_assume!(withdraw_result.token_b_amount <= swap_token_b_amount);
```
jstarry, line 539: Should these be asserts?

token-swap/program/fuzz/src/instructions.rs
```diff
@@ -203,20 +205,26 @@ fn run_fuzz_instructions(fuzz_instructions: Vec<FuzzInstruction>) {
...
 
-    let pool_tokens = [&token_swap.pool_token_account, &token_swap.pool_fee_account]
-        .iter()
-        .map(|&x| get_token_balance(x))
-        .sum::<u64>() as u128;
+    // Omit fees intentionally, because fees in the form of pool tokens can
+    // dilute the value of the pool.  For example, if we perform a small swap
+    // whose fee is worth less than 1 pool token, we may round up to 1 pool
+    // token and mint it as the fee.  Depending on the size of the pool, this
+    // fee can actually reduce the value of pool tokens.
```
jstarry, line 212: nit: doesn't minting fees always decrease the value of pool tokens?

token-swap/program/src/curve/constant_product.rs
```diff
@@ -43,6 +44,93 @@ pub fn swap(
...
+    let mut token_b_amount = pool_tokens
+        .checked_mul(swap_token_b_amount)?
+        .checked_div(pool_token_supply)?;
+    let (token_a_amount, token_b_amount) = match round_direction {
+        RoundDirection::Floor => (token_a_amount, token_b_amount),
+        RoundDirection::Ceiling => {
+            let token_a_remainder = pool_tokens
+                .checked_mul(swap_token_a_amount)?
+                .checked_rem(pool_token_supply)?;
+            if token_a_remainder > 0 && token_a_amount > 0 {
```
jstarry, line 71: Why check for `token_a_amount > 0` here? Isn't it valid for `token_a_amount` to be ~0.99 and round up?

token-swap/program/src/curve/math.rs
```diff
@@ -201,6 +201,16 @@ impl PreciseNumber {
...
 
+    /// Ceiling a precise value to a precision of ONE
+    pub fn ceiling(&self) -> Option<Self> {
+        let value = self
+            .value
+            .checked_add(one())?
+            .checked_div(one())?
+            .checked_mul(one())?;
+        Some(Self { value })
+    }
```
jstarry, line 212: Do you mind adding a test for `precise.ceiling() == precise.ceiling().ceiling()`? I think it will fail

jstarry:
Looking good, I keep getting inspired by your testing and math fixes and want to apply them to lending too! I imagine we'll want a shared precise math library soon 😉

token-swap/program/fuzz/src/instructions.rs
joncinque: It depends.  We're supposed to be withholding enough tokens from the trade to *not* reduce the value of pool tokens.  Ie. if we trade 10000 token A, 25 go as normal trading fees, and 5 go to the owner in the form of pool tokens.  If there's a clean conversion, ie. 5 tokens = 10 pool tokens, we're all good.  If 5 tokens = 1.6 pool tokens, and we mint 2 pool tokens, there will be value lost per token for that trade.

jstarry: Ah right, thanks for the explanation!

token-swap/program/src/curve/calculator.rs
joncinque: It's simply that we can get a situation where our calculation cuts off value from rounding, making the value per token look smaller.  For example, after a withdraw, if the normalized value goes from 100.6 to 100.4 in absolute terms, then our calculation will say it went from 101 to 100, showing that too much value per token was lost.

token-swap/program/src/curve/constant_price.rs
```diff
@@ -105,12 +122,21 @@ impl CurveCalculator for ConstantPriceCurve {
...
+            RoundDirection::Floor => Some(
+                pool_supply
+                    .checked_mul(given_value)?
+                    .checked_div(total_value)?
+                    .as_u128(),
+            ),
+            RoundDirection::Ceiling => Some(
+                pool_supply
+                    .checked_mul(given_value)?
+                    .checked_add(U256::from(1))?
```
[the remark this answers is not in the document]
joncinque: I'll add ceiling division for U256

Otherwise, there's `check_pool_token_conversion` which does a few round trips to make sure no value is lost, but I can add another one which just checks that no value is lost on one `trading_tokens_to_pool_tokens`.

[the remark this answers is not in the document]
jstarry: Up to you, this math just didn't look quite right so I figured there might be some missing test coverage

token-swap/program/src/curve/calculator.rs
jstarry: Do you think it's worth changing how the normalized value is created to avoid this? It looks doable for constant price at least

token-swap/program/src/curve/constant_price.rs
joncinque: This was a bit of a hammer solution to make sure it's a working withdrawal.  When withdrawing equally on both sides, constant price always uses the same ratio, so you can easily get into a situation where you try to overdraw one of the sides.  For example, if you have a price of 20 per token b, 1000 token A, and 50 token B, you'll never run into any issues since everything is in perfect balance.  If you try to withdraw 10% of pool tokens, you'll try to take 10% of the value of the pool in equal A and B, meaning 100 A and 5 B.  Now if your supplies are 10000 token A, and 5 token B, and you try to withdraw 10% of the value of the pool, equally split, you'll ask for 505 token A and 25 token B.  It's one of the big reasons we needed single-sided deposit and withdrawal.  Other than by doing the same math as the withdrawal, I didn't know how to make sure the withdrawal was valid.

token-swap/program/src/curve/math.rs
joncinque: Will do, it definitely will fail

token-swap/program/src/curve/constant_product.rs
joncinque: Since we're doing a ceiling and not a rounding, we can run into bad UX if someone asks for 1 pool token, which is worth 0.01 token A, and take 1 token A from them for it.  There will always be some lost value in some direction, but this check avoids the bad offenders at the lower limit

token-swap/program/src/curve/calculator.rs
joncinque: We could have the normalized value returned as a `PreciseNumber` instead of an int, otherwise you have to round at some point.  I didn't want to expose that type in the Curve interface, but there shouldn't be a problem with exposing it.

Constant price is actually the only curve that doesn't need this epsilon, which is why the test is implemented differently for it.

token-swap/program/src/curve/constant_price.rs
```diff
@@ -67,18 +67,34 @@ impl CurveCalculator for ConstantPriceCurve {
...
+                let token_b_amount = pool_tokens
+                    .checked_mul(total_value)?
+                    .checked_div(token_b_price)?
+                    .checked_div(pool_token_supply)?;
+                (token_a_amount, token_b_amount)
+            }
+            RoundDirection::Ceiling => {
+                let (token_a_amount, _) =
+                    ceiling_division(pool_tokens.checked_mul(total_value)?, pool_token_supply)?;
+                let (token_b_amount, _) = ceiling_division(
```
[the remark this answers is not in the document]
joncinque: Yeah true, it would make things nifty.  I liked all of the math traits that you added for lending :-)

token-swap/program/src/curve/constant_product.rs
jstarry: Got it, do you think that the existing slippage constraints are enough to protect against this? Client UX will typically display max slippage that the user could experience. If not, do you mind adding a comment in there with the above explanation about low number protection?

token-swap/program/src/curve/calculator.rs
jstarry: > We could have the normalized value returned as a PreciseNumber instead of an int

Ah yeah that works too! 

> Constant price is actually the only curve that doesn't need this epsilon, which is why the test is implemented differently for it.

Oh, didn't realize the difference, thanks

token-swap/program/src/curve/constant_price.rs
jstarry: I see, thanks for the explanation. The way the test is written seems fine to me

token-swap/program/src/curve/calculator.rs
joncinque: Done with `fe05a51`

token-swap/program/src/curve/math.rs
joncinque: Fixed in `95d7c7e`

token-swap/program/src/curve/constant_product.rs
joncinque: This calculation is done for a deposit, where the slippage constraints go in the opposite direction, specifying the maximum to take, not the minimum, so unfortunately it doesn't cover this situation.  I'll add the comment with this info

joncinque: Comment add in in `4a936c6`

token-swap/program/src/curve/constant_price.rs
joncinque: Good catch

token-swap/program/src/curve/constant_price.rs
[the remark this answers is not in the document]
joncinque: I ended up doing it using templates  in `99a08d5` -- let me know what you think

token-swap/program/src/curve/constant_price.rs
joncinque: Done in `99a08d5`

token-swap/program/src/curve/constant_price.rs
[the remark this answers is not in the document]
joncinque: Done now with `ee618f2`

jstarry (approved):

PR #1062, opened 2021-01-14
[WIP] Feat/better lending tests

dummytester123:
As part of learning rust, going to gradually improve the lending platform's test suite.


jstarry:
> As part of learning rust, going to gradually improve the lending platform's test suite.

Awesome! FYI the repay test was just expanded here: https://github.com/solana-labs/solana-program-library/pull/1068

dummytester123:
Hey @jstarry I feel like I just retread your exact foot steps here! I did not know you were also working on this. I learned a lot doing it though, which feels great. Is there another instruction here I could assist in further extending?

jstarry:
> Hey @jstarry I feel like I just retread your exact foot steps here! I did not know you were also working on this. I learned a lot doing it though, which feels great. Is there another instruction here I could assist in further extending?

Haha, you had good intuition, `repay` was the least tested and deserved another look. I'd recommend the liquidation instruction 👍

dummytester123:
Roger that. Will give it a whirl. Going to try to add checks on the other mutated fields!

token-lending/program/tests/liquidate.rs
```diff
@@ -146,7 +182,19 @@ async fn test_success() {
         usdc_liquidated,
         usdc_loan_state.borrowed_liquidity_wads.round_u64()
     );
+    let collateral_liquidated =
+        USDC_LOAN_SOL_COLLATERAL - usdc_loan_state.deposited_collateral_tokens;
+    assert!(collateral_liquidated > 0)
```
dummytester123, line 187: Collateral is a tough one to test exactly because the math to get to it is complex. We're either hard coding it (which is not really a good clear cut test) or replicating the logic (which isn't really testing at all.) For now, I just asserted that it changed. Not great, but better than nothing.

token-lending/program/tests/liquidate.rs
```diff
@@ -146,7 +182,19 @@ async fn test_success() {
...
+}
 
+#[tokio::test]
+async fn test_liquidate_sol_obligation() {
+    let TestReturn {
+        mut banks_client,
```
dummytester123, line 193: I separated the sol and usdc tests into two separate tests

token-lending/program/tests/liquidate.rs
```diff
@@ -101,42 +116,63 @@ async fn test_success() {
...
+                amount: sol_amount,
                 user_accounts_owner: &user_accounts_owner,
                 obligation: &sol_obligation,
+                successful_transaction,
             },
         )
         .await;
 
+    TestReturn {
+        banks_client,
```
dummytester123, line 158: I created a TestReturn struct and made a setup function so it can be reused.

token-lending/program/tests/liquidate.rs
```diff
@@ -156,4 +204,28 @@ async fn test_success() {
...
     );
+
+    let collateral_liquidated =
+        SOL_LOAN_USDC_COLLATERAL - sol_loan_state.deposited_collateral_tokens;
+    assert!(collateral_liquidated > 0)
+}
+
+#[tokio::test]
+async fn test_liquidate_healthy_obligation_failure() {
+    let TestReturn { usdc_result, .. } = setup(100, 100, false).await;
```
dummytester123, line 215: My first real contribution. I rewired the liquidate function slightly and used it to actually test that one of your validations works! What do you think?

token-lending/program/tests/helpers/mod.rs
```diff
@@ -587,7 +591,12 @@ impl TestLendingMarket {
             ],
             recent_blockhash,
         );
-        assert!(banks_client.process_transaction(transaction).await.is_ok());
+        let result = banks_client.process_transaction(transaction).await;
```
dummytester123, line 594: By telling liquidate if it expects a success or not, we can return an actual result that can be used by the test to inspect validations later.

dummytester123:
@jstarry Still very much a WIP, but I spent a good chunk of today experimenting with liquidate and I refactored liquidate.rs (the test file) a lot, and added one new test around the healthy obligation check. I was wondering if you would not mind checking out my syntax, and style, and letting me know if I'm on the right track here. I think a good approach would be to write tests around the validation checks first as this is really important and low hanging fruit, healthy obligation being my first attempt. Thanks!

token-lending/program/tests/liquidate.rs
dummytester123: I definitely think as I grow this test file I will need to re-parameterize setup to take a struct instead of comma separated parameters because it'll get too unwieldy.

dummytester123: Also will want to tell it whether I want a USDC loan or a SOL loan, not both...

token-lending/program/tests/helpers/mod.rs
```diff
@@ -520,14 +523,15 @@ impl TestLendingMarket {
         banks_client: &mut BanksClient,
         payer: &Keypair,
         args: LiquidateArgs<'_>,
-    ) {
+    ) -> Result<(), TransportError> {
```
jstarry, line 526: Could you use `TransactionError` instead? `TransportError` can be unwrapped to it

token-lending/program/tests/helpers/mod.rs
```diff
@@ -587,7 +591,12 @@ impl TestLendingMarket {
             ],
             recent_blockhash,
         );
-        assert!(banks_client.process_transaction(transaction).await.is_ok());
+        let result = banks_client.process_transaction(transaction).await;
+        if successful_transaction {
+            assert!(result.is_ok());
+        }
```
jstarry, line 597: I would just pass the result through. Whenever a function in Rust returns a `Result`, the compiler will bug you to check the result anyways so we can chat for `ok` there

token-lending/program/tests/repay.rs
```diff
@@ -140,4 +145,43 @@ async fn test_success() {
...
+
+    TestReturn {
+        banks_client,
+        obligation,
+        usdc_reserve,
+    }
+}
+
+#[tokio::test]
+async fn test_repay_obligation() {
```
jstarry, line 157: Why did you decide to split the obligation and repay reserve checks off from the success test?

token-lending/program/tests/repay.rs
```diff
@@ -28,19 +28,24 @@ const INITIAL_SOL_RESERVE_SUPPLY_LAMPORTS: u64 = 42_500 * LAMPORTS_TO_SOL;
...
         spl_token_lending::id(),
         processor!(process_instruction),
     );
 
     // limit to track compute unit increase
-    test.set_bpf_compute_max_units(80_000);
-
-    const OBLIGATION_LOAN: u64 = 1;
-    const OBLIGATION_COLLATERAL: u64 = 500;
+    test.set_bpf_compute_max_units(NUMBER_OF_TESTS * 80_000);
```
jstarry, line 48: I don't see `NUMBER_OF_TESTS` used elsewhere, what is its purpose here?

token-lending/program/tests/liquidate.rs
jstarry: Yeah it's tough. I think we'll want to mock out the dex markets to get nice numbers to work with. Do you want to give that a shot?

token-lending/program/tests/liquidate.rs
jstarry: > My first real contribution

haha, woot!

> I will need to re-parameterize setup to take a struct instead of comma separated parameters because it'll get too unwieldy. Also will want to tell it whether I want a USDC loan or a SOL loan, not both...

Those changes sound good to me

token-lending/program/tests/liquidate.rs
```diff
@@ -156,4 +204,28 @@ async fn test_success() {
...
+    let he_as_number = LendingError::HealthyObligation as u32;
+    let unwrapped = usdc_result.unwrap_err().unwrap();
+    match unwrapped {
```
jstarry, line 218: Can you try using a single `assert_eq!` rather than `match`?

jstarry (commented):
Thanks for the contribution, here's some initial feedback!

token-lending/program/tests/repay.rs
dummytester123: I noticed that this becomes a global counter even if you reset test multiple times - it's tied per-file I think, or potentially per run. So I was using it to scale up the number of maximum instructions to the number of tests you had in the file. Trying to figure out a better way to reset this each time you call "setup."

token-lending/program/tests/repay.rs
dummytester123: With unit tests, you want to have as few assertions as possible in each test, so that you aren't hunting down line numbers for each test failure and the assertion that's failed is clear and legible. I was trying to split these up a bit since they could be compartmentalized a bit more to be "object-level" checks, checking obligation in one and repay reserve in the other. I find that unit tests, when they get very verbose and "omnipotent", become very hard to maintain and fragile.

token-lending/program/tests/repay.rs
jstarry: The max units param restricts the number of compute units allowed for each transaction. It shouldn't be accumulated by number of tests. You should be able to set a new value per test if you want

token-lending/program/tests/repay.rs
jstarry: I agree on your approach to unit tests but these are integration tests which are intended to test end-to-end. Unit tests can be found in `src/state/repay.rs` and `src/state/obligation.rs` and are better suited for what I think you are aiming for

dummytester123: It still may be easier to make the integration test itself a larger setup function that runs to completion before we do any sort of assertion - that's what makes it integration. Then break the assertion set into many sub tests, or sub functions that is grouped by some organizing heuristic. Then you just either run those in series or re-run the integration setup multiple times, once for each group as a separate test. Technically takes longer to run but one set can pass while the other can fail and you then can have a more clear test suite (multiple failures and successes vs a single failure or success). These may "look" like unit tests in that they are testing a few assertions instead of many but they're still integration tests in that they're calling the same setup function that runs everything together first before asserting anything. 

All this being said, I totally plan to remove this file from this PR, as you've done a way better job in your PR. I'm focusing on Liquidate now :)

dummytester123:
Alright, some minor updates tonight - took into account your feedback @jstarry , and did a bunch of refactoring. Next up, going to try to stub out that Dex market! Thanks for all your time so far on this.

token-lending/program/tests/repay.rs
dummytester123: Hmm ok I will look into this

token-lending/program/tests/repay.rs
jstarry: > It still may be easier to make the integration test itself a larger setup function that runs to completion before we do any sort of assertion - that's what makes it integration. Then break the assertion set into many sub tests, or sub functions that is grouped by some organizing heuristic. 

This sounds good to me!

jstarry:
> Alright, some minor updates tonight - took into account your feedback @jstarry , and did a bunch of refactoring. Next up, going to try to stub out that Dex market! Thanks for all your time so far on this.

No problem, thank you!

token-lending/program/tests/liquidate.rs
dummytester123: Hey @jstarry I couldn't figure out a good way to really stub out the Dex Market here. There is this cool library called https://docs.rs/mockall/0.9.0/mockall/ that could be useful but it didn't really pan out for the job because it's pretty invasive and doesn't seem to work with the lifetime parameters in use with the TradeSimulator, and overall Rust is just not that friendly with stubbing. The other thing that's at issue is that the liquidator is instantiating the TradeSimulator on the go vs being passed one in, and the mockall library relies heavily on you mocking the object before passing it into something else. I'm not quite sure how to get around that. Do you have any good ideas here? I don't want to spin my wheels on this too much more, as it seems like a potential deep well!

token-lending/program/tests/repay.rs
dummytester123: So I did look into this, and indeed, even though each ProgramTest::new ostensibly makes a new ProgramTest object, it actually stores in memory by spl_token_lending::id() the program instruction count, so if you create it again, it doesn't reset the counter. So if I remove the number_of_tests variable, ie set it to 80,000 and just remake it each time, I run out of instructions and the tests die. However, if I just run a single test, the test is fine, because 80k for one is alright. I think this hasn't been a problem for you before because each of your files only has one test @jstarry

dummytester123:
@jstarry I removed the repay file with a couple git reverts, answered your query about NUMBER_OF_TESTS, and did a follow up on stubbing the dex market (turns out it's tougher than anticipated).

token-lending/program/tests/repay.rs
jstarry: Oh really? Nice find, that's a bug in the program test framework then. Do you mind writing an issue over in the [main monorepo project](https://github.com/solana-labs/solana) describing the problem?

token-lending/program/tests/liquidate.rs
jstarry: I was imagining that we could fill the dex market account data with specific orders dynamically. This would require deeper knowledge of how serum's data structures work and then serialize that data into the dex market order book accounts to seed the program tests.

jordansexton:
Closing after discussion with author, https://github.com/solana-labs/solana-program-library/pull/1441 changes most of this.

PR #1068, opened 2021-01-14
lending: Fix obligation interest accrual

jstarry:
#### Problem
Obligation interest calculation is overcomplicated and untested

#### Changes
- Simplify obligation interest calculation
- Add tests for obligation repayment


joncinque (approved):
Great test, and much easier to understand -- nice!  It might be worth adding a proptest for interest accrual to make sure it's behaving properly.

jstarry:
> It might be worth adding a proptest for interest accrual to make sure it's behaving properly.

Totally, I have one in the checked math PR. Will follow up there

PR #1084, opened 2021-01-18
lending: Refactor and test obligation repays

jstarry:
#### Problem
Difficult to test obligation repay math due to it being inline in the processor instruction handling

#### Changes
Factor out repay math into obligation method and add proptests. Caught a few issues:
- Obligation token repay percent should always strictly less than withdraw percent because we don't want a split obligation to be able to take advantage of rounding error to receive more collateral than they are allotted.
- Borrowers could take advantage of rounding error to pay less liquidity for more withdrawn collateral leading to lost interest returns for the pool. Changes were made to round up liquidity payments and round down collateral withdrawals.

### DIFF
--- a/token-lending/program/src/state/reserve.rs
+++ b/token-lending/program/src/state/reserve.rs
@@ -218,20 +218,19 @@ impl ReserveLiquidity {
         Ok(())
     }
 
-    /// Subtract repay amount from total borrows and return rounded repay value
-    pub fn repay(&mut self, repay_amount: Decimal) -> Result<u64, ProgramError> {
-        let rounded_repay_amount = repay_amount.try_round_u64()?;
-        if rounded_repay_amount == 0 {
-            return Err(LendingError::ObligationTooSmall.into());
-        }
-
+    /// Subtract repay amount from total borrows and add to available liquidity
+    pub fn repay(
+        &mut self,
+        integer_amount: u64,
+        decimal_amount: Decimal,
+    ) -> Result<(), ProgramError> {
         self.available_amount = self
             .available_amount
-            .checked_add(rounded_repay_amount)
+            .checked_add(integer_amount)
             .ok_or(LendingError::MathOverflow)?;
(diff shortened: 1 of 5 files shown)


token-lending/program/src/state/reserve.rs
```diff
@@ -218,20 +218,19 @@ impl ReserveLiquidity {
...
+        integer_amount: u64,
+        decimal_amount: Decimal,
+    ) -> Result<(), ProgramError> {
         self.available_amount = self
             .available_amount
-            .checked_add(rounded_repay_amount)
+            .checked_add(integer_amount)
             .ok_or(LendingError::MathOverflow)?;
-        self.borrowed_amount_wads = self.borrowed_amount_wads.try_sub(repay_amount)?;
+        self.borrowed_amount_wads = self.borrowed_amount_wads.try_sub(decimal_amount)?;
```
joncinque, line 231: If we're adding a whole amount of tokens back into the available liquidity, shouldn't we subtract the same amount of whole tokens from the borrowed amount?  Your tests show that this is working correctly and getting down to 0 on the full repay, which probably answers my question :-)

joncinque (approved):
Looks really good!  The proptests cover exactly the situations to worry about.  I just have one question, but given the working test it shouldn't be an issue, so feel free to push.

token-lending/program/src/state/reserve.rs
jstarry: Yeah, good question. The intent of my change was to avoid dust balances by having special logic for when the repay amount is higher than the borrow amount (which has a decimal balance due to interest) such that we switch to paying off the full borrow amount. When repay amount is lower than the borrow amount, the `decimal_amount` you pointed out above is just a whole number. 

> Your tests show that this is working correctly and getting down to 0 on the full repay, which probably answers my question :-)

Haha, yeah that was the goal of the proptests! Glad they were able to speak for themselves :)

PR #1088, opened 2021-01-18
lending: Add InitObligation instruction

jstarry:
#### Problem
The borrow instruction is too overloaded and consumes a lot of compute units which causes problems with composability

#### Changes
- New init obligation instruction
- Borrow now always expects initialized obligation accounts
- Interest is not accrued for the deposit reserve to save compute cost

### DIFF
--- a/token-lending/program/src/processor.rs
+++ b/token-lending/program/src/processor.rs
@@ -6,8 +6,8 @@ use crate::{
     instruction::{BorrowAmountType, LendingInstruction},
     math::{Decimal, Rate, TryAdd, TryMul, TrySub, WAD},
     state::{
-        LendingMarket, NewReserveParams, Obligation, RepayResult, Reserve, ReserveCollateral,
-        ReserveConfig, ReserveLiquidity, PROGRAM_VERSION,
+        LendingMarket, NewObligationParams, NewReserveParams, Obligation, RepayResult, Reserve,
+        ReserveCollateral, ReserveConfig, ReserveLiquidity, PROGRAM_VERSION,
     },
 };
 use num_traits::FromPrimitive;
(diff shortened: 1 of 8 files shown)


token-lending/program/src/processor.rs
```diff
@@ -575,10 +675,36 @@ fn process_borrow(
...
+    if obligation_token_output_info.owner != token_program_id.key {
+        return Err(LendingError::InvalidTokenOwner.into());
+    }
+    if &obligation_token_output.mint != obligation_token_mint_info.key {
+        return Err(LendingError::InvalidTokenMint.into());
+    }
+
     // accrue interest and update rates
     borrow_reserve.accrue_interest(clock.slot)?;
-    deposit_reserve.accrue_interest(clock.slot)?;
```
joncinque, line 580: What is the impact of not accruing interest on the deposit reserve? We need the deposit reserve for the trade simulation in order to convert between collateral tokens and liquidity tokens.  Based on our exchange in the calc PR, I thought the exchange rate between collateral tokens and liquidity tokens depended on the interest accrued on the reserve.  Digging a bit deeper, it only depends on the amount of tokens issued, which makes sense, since collateral tokens should NOT grow in value just because time has passed -- there need to be borrows / repays / liquidations.  Anyway, this was a bit of a brain dump, but either way, could you add a comment explaining why it isn't needed?

joncinque (dismissed):
Looks great!  Just one little nit / question about the deposit reserve interest accrual on borrow which you can do in your ow time.  What are the savings in instructions in the end?

token-lending/program/src/processor.rs
jstarry: Assuming the worst case of 1 year of interest accrual calculation, not accruing saves about 30k instructions. And cutting out obligation initialization saved about 10k from removing CPI calls.

> I thought the exchange rate between collateral tokens and liquidity tokens depended on the interest accrued on the reserve.

This is indeed the case.

> since collateral tokens should NOT grow in value just because time has passed -- there need to be borrows / repays / liquidations

The way the protocol is implemented allows lenders to continuously earn interest on their lent liquidity. Then, the only trick is having liquidity available to withdraw if they want to cash out. This is where the variable interest rate helps because as the interest rate increases, it incentivizes borrowers to pay back loans and new lenders to add liquidity.

My thinking was that since collateral value is only going up, the worst case with removing the interest accrual is that the loan is underwritten in a way that undervalues collateral, which leads to slightly over-collateralized positions, which didn't seem like an awful thing.

But now I'm thinking it's better to just take the computation hit of accruing the interest to give the client a better experience. In practice, I expect that the reserves will be touched enough that the amortized cost of accruing interest is pretty stable and low.

PR #1104, opened 2021-01-20
lending: Fixes and tests for borrow and liquidate

jstarry:
#### Problems
- Code for borrow and liquidate was not tested and not very suitable for rigorous prop testing
- A few configuration parameters were not validated correctly
- Rounding was not done in the correct direction in a few places

#### Changes
- Remove trade simulator orders cache
- Split bonus payout into separate instruction for higher transparency
- Return an error if liquidation will result in 0 received collateral
- Ensure LTV ratio is non-zero
- Enforce that liquidation threshold is higher than LTV
- Ensure math is rounded with ceiling / floor when appropriate:
  - Liquidations should never round to more than the close factor
  - Liquidation should never receive more collateral than the precise exchange rate
  - Bonus amount should never be higher than the bonus rate

Liquidation processing has changed quite a bit because it turns out that we don't need to do price discovery twice. Previously, price discovery was first done to determine the overall health of an obligation and then again to determine the collateral received for the amount of repaid liquidity. The intuition is that despite the token exchange price changes in relation to the amount of tokens to be converted, it's sufficient to just use the exchange price returned for the amount of repaid liquidity which is strictly smaller than the total size of an obligation. This is because the exchange price gets less favourable as amount of converted tokens increases, which means the relative value of the collateral diminishes against the loan which leads to an obligation being seen as less healthy. So, if the exchange rate used to convert a relatively small amount of tokens is low enough to make the obligation unhealthy, there's no need to also check the exchange rate for the full obligation because it will only indicate an even lower health factor.

### DIFF
--- a/token-lending/program/src/state/reserve.rs
+++ b/token-lending/program/src/state/reserve.rs
@@ -1,6 +1,7 @@
 use super::*;
 use crate::{
     error::LendingError,
+    instruction::BorrowAmountType,
     math::{Decimal, Rate, TryAdd, TryDiv, TryMul, TrySub},
 };
 use arrayref::{array_mut_ref, array_ref, array_refs, mut_array_refs};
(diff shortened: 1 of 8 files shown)


token-lending/program/src/state/reserve.rs
```diff
@@ -544,12 +723,240 @@ impl Pack for Reserve {
...
+                        obligation.deposited_collateral_tokens
+                );
+            }
+        }
+
+        #[test]
+        fn allowed_borrow_for_collateral(
+            collateral_amount in 0..=u32::MAX as u64,
+            collateral_exchange_rate in PERCENT_SCALER..=5 * WAD,
+            token_conversion_rate in 0..=u64::MAX as u128,
```
joncinque, line 849: Should this be `1..=u64::MAX` like in the other test?

token-lending/program/src/state/reserve.rs
```diff
@@ -544,12 +723,240 @@ impl Pack for Reserve {
...
+            // Assert that reversing the calculation returns approx original amount
+            assert!(collateral_amount >= collateral_amount_lower);
+            assert!(collateral_amount <= collateral_amount_upper);
+        }
+
+        #[test]
+        fn required_collateral_for_borrow(
+            borrow_amount in 0..=u32::MAX as u64,
+            collateral_exchange_rate in PERCENT_SCALER..=5 * WAD,
+            token_conversion_rate in 0..=u64::MAX as u128,
```
joncinque, line 907: Same here

joncinque (approved):
The main logic looks good, just a few questions on the test limits.  I'll approve this so you can move onto the liquidation threshold we discussed.

token-lending/program/tests/liquidate.rs
```diff
@@ -140,24 +140,24 @@ async fn test_success() {
...
-        sol_loan_state
-            .borrowed_liquidity_wads
-            .try_round_u64()
-            .unwrap()
+    assert!(
+        sol_liquidated
+            <= sol_loan_state
+                .borrowed_liquidity_wads
+                .try_floor_u64()
+                .unwrap()
```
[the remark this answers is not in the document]
jstarry: Now that interest accrual is separate, this can be exact now. Fixed!

token-lending/program/src/state/reserve.rs
jstarry: Good question. I opted to change the other to `0..=u64::MAX` with the mentality that wider ranges are better for proptests. Thanks for calling out the inconsistency

jstarry: Err, not sure how the other passed with `0..=u64::MAX` because it's not valid. I'll follow up in another PR to make these consistent in starting with `1` as the minimum exchange rate

joncinque: It's kind of worrying that the proptests don't automatically test the limits, we might have to add some proptest utilities to make sure it always goes for the top and bottom of all ranges.  Not sure if you remember, but I ran into the same issue with the swap proptests.

jstarry: Yes, I remember and agree :/

PR #1118, opened 2021-01-22
lending: Handle the lost funds from defaulted loans

jstarry:
#### Problem
When loans default (collateral is fully withdrawn with an outstanding balance), the remaining borrow amount is not taken into account in the reserve collateral calculations. When loans default, collateral tokens should lose value, but currently, the end result is a situation where the reserve thinks there's still liquidity that will be paid back and so will not allow full redemption of outstanding collateral tokens.

#### Changes
- After an obligation is liquidated and defaults, subtract the default amount from the repay reserve


joncinque (approved):
Looks good!

PR #1133, opened 2021-01-26
docs: Flesh out token swap README

joncinque:
Now that the token swap program is out in the wild, it's really time for some documentation.

The idea is to follow the docs at [token](https://github.com/solana-labs/solana-program-library/blob/master/docs/src/token.md), specifically focusing on:

* How the accounts are created and used.  Devs (myself included) have initial struggles understanding the many accounts required by Solana programs, so the idea is to also provide background information about accounts
* Curves present.  If people want to implement their own curve, they should know that it's possible and what's already in the program.

### DIFF
--- a/docs/src/token-swap.md
+++ b/docs/src/token-swap.md
@@ -1,47 +1,35 @@
----
-title: Token Swap Program
+# Token Swap Program
 ---
 
-A Uniswap-like exchange for the Token program on the Solana blockchain,
-implementing multiple automated market maker (AMM) curves.
+A Uniswap-like exchange for the Token program on the Solana blockchain, implementing multiple automated market maker (AMM) curves.
 
 ## Overview
 
-The Token Swap Program allows simple trading of token pairs without a centralized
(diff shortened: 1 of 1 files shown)


bartosz-lipinski (dismissed):
Looks awesome @joncinque I made some edits

docs/src/token-swap.md
```diff
@@ -1,7 +1,291 @@
...
+
+The Token Swap Program's source is available on
+[github](https://github.com/solana-labs/solana-program-library).
+
+## Interface
+
+[JavaScript
+bindings](https://github.com/solana-labs/solana-program-library/blob/master/token-swap/js/client/token-swap.js)
+are available that support loading the Token Swap Program on to a chain and
+issue instructions.
```
CriesofCarrots, line 45:
```suggestion
issuing instructions.
```

joncinque: Thanks for looking at it!

PR #1134, opened 2021-01-26
Stake pool: test coverage for most of the program errors

ysavchenko:
All negative outcomes covered except for the cases where it is needed to change account data (or just the system clock) for proper testing.

### DIFF
--- a/stake-pool/program/tests/deposit.rs
+++ b/stake-pool/program/tests/deposit.rs
@@ -4,16 +4,25 @@ mod helpers;
 
 use helpers::*;
 
-use solana_sdk::signature::{Keypair, Signer};
+use solana_program::hash::Hash;
+use solana_program_test::BanksClient;
+use solana_sdk::{
+    instruction::InstructionError,
+    signature::{Keypair, Signer},
+    transaction::Transaction,
+    transaction::TransactionError,
+    transport::TransportError,
+};
 use spl_stake_pool::*;
+use spl_token::error as tokenErrors;
 
-#[tokio::test]
-async fn test_stake_pool_deposit() {
+async fn setup() -> (BanksClient, Keypair, Hash, StakePoolAccounts, StakeAccount) {
     let (mut banks_client, payer, recent_blockhash) = program_test().start().await;
     let stake_pool_accounts = StakePoolAccounts::new();
     stake_pool_accounts
         .initialize_stake_pool(&mut banks_client, &payer, &recent_blockhash)
-        .await;
+        .await
+        .unwrap();
 
     let validator_stake_account: StakeAccount = simple_add_validator_stake_account(
         &mut banks_client,
@@ -22,6 +31,20 @@ async fn test_stake_pool_deposit() {
         &stake_pool_accounts,
     )
     .await;
+
+    (
+        banks_client,
+        payer,
+        recent_blockhash,
+        stake_pool_accounts,
+        validator_stake_account,
+    )
+}
+
+#[tokio::test]
+async fn test_stake_pool_deposit() {
+    let (mut banks_client, payer, recent_blockhash, stake_pool_accounts, validator_stake_account) =
+        setup().await;
 
     let user = Keypair::new();
     // make stake account
@@ -50,7 +73,8 @@ async fn test_stake_pool_deposit() {
         &stake_pool_accounts.pool_mint.pubkey(),
         &user.pubkey(),
     )
-    .await;
+    .await
+    .unwrap();
 
     // Save stake pool state before depositing
     let stake_pool_before =
@@ -81,7 +105,8 @@ async fn test_stake_pool_deposit() {
             &user_pool_account.pubkey(),
             &validator_stake_account.stake_account,
         )
-        .await;
+        .await
+        .unwrap();
 
     // Original stake account should be drained
     assert!(banks_client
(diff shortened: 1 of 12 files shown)


stake-pool/program/tests/withdraw.rs
```diff
@@ -151,3 +196,736 @@ async fn test_stake_pool_withdraw() {
...
+        .await
+        .err()
+        .unwrap();
+
+    match transaction_error {
+        TransportError::TransactionError(TransactionError::InstructionError(
+            _,
+            InstructionError::Custom(error_index),
+        )) => {
+            let program_error = error::StakePoolError::InvalidProgramAddress as u32;
```
joncinque, line 347: This should trigger an `IncorrectProgramId` error -- it looks like the `new_authority` might be throwing this off.  What is it for in this case?

stake-pool/program/tests/withdraw.rs
```diff
@@ -151,3 +196,736 @@ async fn test_stake_pool_withdraw() {
...
+            &deposit_info.user_pool_account,
+            &validator_stake_account.stake_account,
+            &new_authority,
+            lamports_to_withdraw,
+        )
+        .await
+        .err()
+        .unwrap();
+
+    match transaction_error {
```
joncinque, line 682: I might be missing something, so let me know if I have this right -- we're trying to withdraw 1/4 from the same stake twice, so it should normally succeed, but we're missing another call to delegate the pool tokens.  Is this trying to test the token approval code?  On a separate note, why does this fail with `InvalidAccountData`?

stake-pool/program/tests/withdraw.rs
```diff
@@ -151,3 +196,736 @@ async fn test_stake_pool_withdraw() {
...
+        TransportError::TransactionError(TransactionError::InstructionError(
+            _,
+            InstructionError::Custom(error_index),
+        )) => {
+            let program_error = TokenError::OwnerMismatch as u32;
```
joncinque, line 776: I believe this should have a delegate / approve instruction before this, which is why the wrong error is getting triggered -- this should probably be a `MintMismatch` error

stake-pool/program/tests/vsa_remove.rs
```diff
@@ -104,3 +148,449 @@ async fn test_remove_validator_stake_account() {
...
+        .await
+        .err()
+        .unwrap();
+
+    match transaction_error {
+        TransportError::TransactionError(TransactionError::InstructionError(
+            _,
+            InstructionError::Custom(error_index),
+        )) => {
+            let program_error = error::StakePoolError::InvalidProgramAddress as u32;
```
joncinque, line 246: Shouldn't this be `IncorrectProgramId`? What is `new_authority` for?

stake-pool/program/tests/vsa_add.rs
```diff
@@ -102,3 +143,411 @@ async fn test_add_validator_stake_account() {
...
+        .await
+        .err()
+        .unwrap();
+
+    match transaction_error {
+        TransportError::TransactionError(TransactionError::InstructionError(
+            _,
+            InstructionError::Custom(error_index),
+        )) => {
+            let program_error = error::StakePoolError::InvalidProgramAddress as u32;
```
joncinque, line 187: Why is this `InvalidProgramAddress` and not `IncorrectProgramId`?

joncinque (commented):
Lots of great stuff!  I mainly had questions about the errors returned in a few tests.  Also, will the tests with TODO be done in another PR?

ysavchenko:
> Lots of great stuff! I mainly had questions about the errors returned in a few tests. Also, will the tests with TODO be done in another PR?

The next PR is for the stuff we skipped in the previous PR :). And TODOs in the tests here wait until we can change account data from tests.

stake-pool/program/tests/deposit.rs
```diff
@@ -143,3 +168,693 @@ async fn test_stake_pool_deposit() {
...
+    create_token_account(
+        &mut banks_client,
+        &payer,
+        &recent_blockhash,
+        &user_pool_account,
+        &stake_pool_accounts.pool_mint.pubkey(),
+        &user.pubkey(),
+    )
+    .await
+    .unwrap();
```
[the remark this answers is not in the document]
ysavchenko: Yes, this is weird, looks like a copy-paste error. It is supposed to be stake account creation and the test is for the case when deposit is trying to merge with account not appended to the pool. Fixing this.

stake-pool/program/tests/deposit.rs
```diff
@@ -143,3 +168,693 @@ async fn test_stake_pool_deposit() {
...
+            &user_pool_account.pubkey(),
+            &validator_stake_account.stake_account,
+        )
+        .await
+        .err()
+        .unwrap();
+
+    match transaction_error {
+        TransportError::TransactionError(TransactionError::InstructionError(_, error)) => {
+            assert_eq!(error, InstructionError::MissingRequiredSignature);
```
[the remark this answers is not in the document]
ysavchenko: Yes, it does, I believe from this line:

https://github.com/solana-labs/solana/blob/72f10f5f29dc69c72ea4d814111fc529e47f7457/programs/stake/src/stake_state.rs#L472

The goal of this test is to check if stake program is properly called and throws an error in this case.

stake-pool/program/tests/withdraw.rs
ysavchenko: Right, the problem is in the smart contract, it throws a wrong exception and the test checks for it. Nice catch, fixing this.

stake-pool/program/tests/vsa_add.rs
ysavchenko: Same issue with the wrong error from the contract.

stake-pool/program/tests/vsa_remove.rs
ysavchenko: Same issue with the wrong error from the contract. And `new_authority` is a required param in removing stake account from the pool (it is the new pubkey to be set as Staker and Withdrawer).

stake-pool/program/tests/withdraw.rs
ysavchenko: I've looked it over and this test is totally redundant, sorry, missed it in my review. This error (insufficient funds) is already checked in the last test for `withdraw`

stake-pool/program/tests/withdraw.rs
ysavchenko: Actually this is a test of double withdrawal to the same stake account and the error comes from the stake program itself (because stake account on the second withdraw is not a new one and already initialized).

stake-pool/program/tests/deposit.rs
```diff
@@ -143,3 +168,693 @@ async fn test_stake_pool_deposit() {
...
+            &user_pool_account.pubkey(),
+            &new_validator_stake_account.stake_account,
+        )
+        .await
+        .err()
+        .unwrap();
+
+    match transaction_error {
+        TransportError::TransactionError(TransactionError::InstructionError(_, error)) => {
+            assert_eq!(error, InstructionError::IncorrectProgramId);
```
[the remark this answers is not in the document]
ysavchenko: This is actually a weird test with `IncorrectProgramId` exception for some reason. Let's remove it for now and later go through the coverage and missing test cases where needed. Ok?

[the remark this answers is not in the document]
joncinque: Works for me!

stake-pool/program/tests/withdraw.rs
joncinque: Oh nice! Right, that makes sense, thanks for clarifying

stake-pool/program/tests/vsa_remove.rs
joncinque: Oh right, I see, thanks for the explanation

joncinque:
Looks good! Thanks for quickly addressing everything

PR #1135, opened 2021-01-27
Add initial signed-memo program

### LINKED ISSUE #1071: Add a signed memo program
The current memo program only performs UTF verification of the memo data.   This facility be made more expressive with a new variant of the memo program that:
1. Requires all accounts provided to it be signers of the transaction, and logs their address after confirming to the transaction log
2. Logs the memo text to the transaction log

Now somebody can easily view the transaction log over pubsub or elsewhere, observe memos and know they where approved by zero or more addresses (assuming they trust the transaction log provider..)

Note that in this scheme, the current memo program is simply the signed memo program with no addresses passed in

CriesofCarrots:
The current memo program only performs UTF verification of the memo data. This facility is made more expressive in this program, which:

- Logs the memo text to the transaction log
- Requires all accounts provided to it be signers of the transaction and logs their address to the transaction log. For ease of debugging, this implementation logs a message containing the provided account address, regardless of whether the signature verification was successful

Fixes #1071


CriesofCarrots:
As I was writing the docs page for spl.solana.com, it occurred to me it might be useful to log the UTF-8 parsing err on failure. Thoughts?

signed-memo/program/src/processor.rs
```diff
@@ -0,0 +1,107 @@
...
+            missing_required_signature = true;
+        }
+    }
+
+    let memo = from_utf8(input).map_err(|_| ProgramError::InvalidInstructionData)?;
+    msg!("{:?}", memo);
+
+    if missing_required_signature {
+        return Err(ProgramError::MissingRequiredSignature);
+    }
```
mvines, line 31: I think we should bail on MissingRequiredSignature before verifying and printing the memo

signed-memo/program/src/processor.rs
```diff
@@ -0,0 +1,107 @@
...
+    accounts: &[AccountInfo],
+    input: &[u8],
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+    let mut missing_required_signature = false;
+    for account_info in account_info_iter {
+        if let Some(address) = account_info.signer_key() {
+            msg!("Signed by {:?}", address);
+        } else {
+            msg!("Not signed! {:?}", account_info.key);
```
mvines, line 21: I think I'd be ok with removing this log entirely.    Less chance of something writing a really dump log parser that just matches on address and considers that to be a success

mvines:
> As I was writing the docs page for spl.solana.com, it occurred to me it might be useful to log the UTF-8 parsing err on failure. Thoughts?

Yeah, sure.  Something generic like "Invalid UTF-8" sound good to help decipher an `InvalidInstructionData` error from transaction simulation

mvines:
I'm leaning towards just calling this memo v2.0.0 and overwriting `memo/`.   smemo is a superset and I'm not sure that making devs have to choose between using memo or smemo is beneficial, or having to explain to each new dev why there are two

signed-memo/program/src/processor.rs
CriesofCarrots: Should we suppress all logs unless all the signer-checks pass, then?

CriesofCarrots:
> I'm leaning towards just calling this memo v2.0.0 and overwriting `memo/`. smemo is a superset and I'm not sure that making devs have to choose between using memo or smemo is beneficial, or having to explain to each new dev why there are two

Okay, here's what that looks like @mvines

memo/program/program-id.md
```diff
@@ -1 +1 @@
-Memo1UhkJRfHyvLMcVucJwxXeuD728EqVDDwQDxFMNo
+MemoSq4gqABAXKb96qnH8TysNcWxMyWCqXgDLGmfcHr
```
mvines, line 1: nit: Do you have a `Memo2` ground out?

memo/program/src/processor.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+    if missing_required_signature {
+        return Err(ProgramError::MissingRequiredSignature);
+    }
+
+    let memo = from_utf8(input).map_err(|err| {
+        msg!("Invalid UTF-8, from byte {}", err.valid_up_to());
+        ProgramError::InvalidInstructionData
+    })?;
+    msg!("{:?}", memo);
```
mvines, line 32: Please add a prefix to this message, something like:
```suggestion
    msg!("Memo: {:?}", memo);
```
to prevent me from including an unsigned memo of "Signed by {address}" :)

docs/src/memo.md
```diff
@@ -22,9 +25,16 @@ The Memo Program's source is available on
...
+[spl-memo](https://crates.io/crates/spl-memo) and
+[docs.rs](https://docs.rs/spl-memo).
 
-## Operational overview
+The crate provides a `build_memo()` method to easily create a properly
+constructed Instruction.
 
-The Memo program attempts to UTF-8 decode the instruction data; if successfully
-decoded, the instruction is successful.
+## Operational Notes
```
mvines, line 34: Please add a section about the expected log output, since in this case logging is very much a part of the public API for the program

memo/program/src/processor.rs
mvines: log the string length as well perhaps?

mvines (commented):
> Okay, here's what that looks like @mvines
👍🏼 👍🏼

memo/program/program-id.md
CriesofCarrots: Don't like `MemoS`?
Technically, this is memo v3 (I can't recall why, but you bumped the major version in November). Happy to grind a `Memo3`, if you don't think that will be confusing.

mvines: Na, it's fine.  I'm over it

mvines (approved):
looks super!

PR #1157, opened 2021-02-04
Keep tokio consistent with solana-program-test

CriesofCarrots:
We are working on upgrading tokio upstream at solana: https://github.com/solana-labs/solana/pull/15013
Currently, this breaks the downstream-projects CI build, because of conflicts between the solana and SPL tokio versions.
This PR updates the SPL patching script to keep tokio in sync with `solana-program-test`. This will fix solana CI, but also make developing SPL against a local solana monorepo easier.

PR #1159, opened 2021-02-04
Remove target from cache

CriesofCarrots:
PRs that do not change the dependency hash run into this error on CI build:
`error[E0460]: found possibly newer version of crate `std` which `solana_program` depends on`
https://github.com/solana-labs/solana-program-library/pull/1157/checks?check_run_id=1829061379

Use a big hammer for now, and remove `target` from the cache


t-nelson (approved):
:hammer:

PR #1194, opened 2021-02-07
Add types for bnjs

dummytester123:
Ok, been working on the new yarn2-based monorepository over at https://github.com/solana-labs/oyster. Had an issue where u64 type imported from spl-token lacked any typing.

It has to do with yarn 2 being way smarter with dependencies than npm ever was. The issue is that in the spl-token cli,, we never added "@types/bn.js":"^5.1.0" to the devDependencies. In NPM projects, this is fine as long as you add this to the top level package.json (which we do in oyster-lending) but in yarn2, it's smart enough to not pull that from the outer scope and instead just has no type for the inner scope. We need to release a version with this package.json (name it 0.0.14) and then the monorepo will work.

@joncinque @bartosz-lipinski


token/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@solana/spl-token",
-  "version": "0.0.13",
+  "version": "0.0.14",
```
bartosz-lipinski, line 3: I think we could move this package to 0.1.0, since there is a lot of projects depending on it

bartosz-lipinski (approved):
Awesome great catch.

jstarry:
LGTM, add this flag to fix CI
```
tsc --esModuleInterop module.d.ts
```

dummytester123:
@jstarry where do I add this? To some script in package.json? I'm not familiar with the CI setup here yet.

jstarry:
In .Travis/script.sh

jstarry:
@dummytester123 thanks! Sorry I mixed up repo this was, glad you found the script. Could you also fix the conflicts please?

dummytester123:
Done!

token/js/package.json
```diff
@@ -42,8 +42,8 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.10.5",
-    "@solana/web3.js": "^0.91.0",
-    "bn.js": "^5.0.0",
+    "@solana/web3.js": "^0.90.0",
```
jstarry, line 45: Should be 0.91

token/js/package.json
jstarry: @dummytester123 can you bump to 0.1.0 like bartosz suggested?

token/js/package.json
dummytester123: done

jstarry:
@dummytester123 I think you missed this comment: https://github.com/solana-labs/solana-program-library/pull/1194#discussion_r572459417

token/js/package.json
dummytester123: oh, totally missed that.

PR #1199, opened 2021-02-08
lending: Add overflow check in liquidation proptest

joncinque:
This issue was discovered during CI:
https://github.com/solana-labs/solana-program-library/runs/1833196857

The approach is to do the same calc as `_liquidate_obligation` and make
sure that the `MockConverter` doesn't overflow `Decimal`.  This seemed
like a simpler approach than doing a local proptest constraint.

The actual failure is on line 150 in `_liquidate_obligation`: `obligation.max_liquidation_amount()` is always the amount we try to convert, and in this case:

conversion rate = 681
collateral = u64::MAX (roughly)
max liquidation amount = collateral / 2
collateral exchange rate = 0.01

`receive amount = conversion_rate * (collateral / 2) * collateral_exchange_rate`
`receive_amount = 681 * u64::MAX / 2 * 0.01 > u64::MAX`

On top of that, this PR includes the proptest regression to run the failing case
on every `cargo test`.


jstarry (approved):
Thanks!

PR #1200, opened 2021-02-08
Various postponed fixes and changes to the stake pool program

atticlab:
Here is a complete list of changes:

- Add check for stake to be active on `deposit`
- Update `initialize` instruction to be signed by owner (update CLI)
- Change Create VSA instruction to use a single instruction to create account
- Check if initialized accounts are rent-exempt
- Change `withdraw` instruction parameter from number of SOLs to withdraw to the number of tokens to burn
- Add extra validity checks to the pool initialization
- Update error labels
- Remove `State` enum, save `StakePool` directly to the storage
- Remove initialization bool flag in `ValidatorStakeList`, use version instead
- Name refactoring in tests


stake-pool/cli/src/main.rs
```diff
@@ -1444,6 +1419,7 @@ fn main() {
         ("withdraw", Some(arg_matches)) => {
             let pool_account: Pubkey = pubkey_of(arg_matches, "pool").unwrap();
             let burn_from: Pubkey = pubkey_of(arg_matches, "burn_from").unwrap();
+            // convert pool tokens
             let amount: u64 = sol_to_lamports(value_t_or_exit!(arg_matches, "amount", f64));
```
joncinque, line 1423: Should this just be an amount instead of lamports since we're sending in pool tokens?

stake-pool/program/src/state.rs
```diff
@@ -114,67 +125,34 @@ impl StakePool {
...
-            0 => State::Unallocated,
-            1 => {
-                // We send whole input here, because unpack skips the first byte
-                let swap: &StakePool = unpack(&input)?;
-                State::Init(*swap)
-            }
-            _ => return Err(ProgramError::InvalidAccountData),
-        })
+
+        let stake_pool: &StakePool = unsafe { &*(&input[0] as *const u8 as *const StakePool) };
```
joncinque, line 140: Not for this PR, but eventually we'll want to do some better packing with Borsh

stake-pool/program/src/processor.rs
```diff
@@ -293,70 +294,100 @@ impl Processor {
...
 
         // Check if validator stake list storage is unitialized
         let mut validator_stake_list =
             ValidatorStakeList::deserialize(&validator_stake_list_info.data.borrow())?;
-        if validator_stake_list.is_initialized {
+        if validator_stake_list.is_initialized() {
             return Err(StakePoolError::AlreadyInUse.into());
         }
-        validator_stake_list.is_initialized = true;
+        validator_stake_list.version = ValidatorStakeList::VALIDATOR_STAKE_LIST_VERSION;
```
joncinque, line 320: nit: What's the reason of having both VALIDATOR_STAKE_LIST_VERSION and PROGRAM_VERSION?  I might be missing a use case, or perhaps you want to keep things as flexible as possible.  I'm worried about handling version mismatches later on

stake-pool/program/src/processor.rs
```diff
@@ -582,7 +601,7 @@ impl Processor {
         stake_pool.pool_total += token_amount;
```
joncinque, line 601: I can't comment further up, but you'll need to add the activated check a little higher up at the TODO as well

stake-pool/program/src/processor.rs
```diff
@@ -845,7 +870,30 @@ impl Processor {
...
+
+        // TODO: remove conditional compilation when time travel in tests is possible
+        #[cfg(not(feature = "test-bpf"))]
+        {
+            let stake_acc_state: stake::StakeState =
+                deserialize(&stake_info.data.borrow()).unwrap();
+            let delegation = stake_acc_state.delegation();
+            if let Some(delegation) = delegation {
+                let target_epoch = clock.epoch;
+                let history = None;
```
joncinque, line 886: You'll also need to pass in the `StakeHistory`sysvar to calculate this correctly

stake-pool/program/src/processor.rs
```diff
@@ -845,7 +870,30 @@ impl Processor {
...
+                let fix_stake_deactivate = false;
```
joncinque, line 887: Why is this set to `false`?  I think we want the fix enabled

stake-pool/program/src/processor.rs
```diff
@@ -845,7 +870,30 @@ impl Processor {
...
+                let (effective, activating, deactivating) = delegation
+                    .stake_activating_and_deactivating(target_epoch, history, fix_stake_deactivate);
+                if activating != 0 && deactivating != 0 && effective == 0 {
```
joncinque, line 890: This should probably be 
```suggestion
                if activating != 0 || deactivating != 0 || effective == 0 {
```

stake-pool/program/src/processor.rs
```diff
@@ -845,7 +870,30 @@ impl Processor {
             return Err(ProgramError::IncorrectProgramId);
         }
 
-        let mut stake_pool = State::deserialize(&stake_pool_info.data.borrow())?.stake_pool()?;
+        let mut stake_pool = StakePool::deserialize(&stake_pool_info.data.borrow())?;
+        if !stake_pool.is_initialized() {
+            return Err(StakePoolError::InvalidState.into());
+        }
+
+        // TODO: remove conditional compilation when time travel in tests is possible
```
joncinque, line 878: The work for that is done and will be released with 1.5.6!

stake-pool/program/src/processor.rs
```diff
@@ -1213,27 +1270,30 @@ impl PrintProgramError for StakePoolError {
...
-            StakePoolError::WrongPoolMint => msg!("Error: WrongPoolMint"),
-            StakePoolError::WrongStakeState => msg!("Error: WrongStakeState"),
-            StakePoolError::ValidatorAlreadyAdded => msg!("Error: ValidatorAlreadyAdded"),
-            StakePoolError::ValidatorNotFound => msg!("Error: ValidatorNotFound"),
-            StakePoolError::InvalidStakeAccountAddress => msg!("Error: InvalidStakeAccountAddress"),
-            StakePoolError::StakeListOutOfDate => msg!("Error: StakeListOutOfDate"),
-            StakePoolError::StakeListAndPoolOutOfDate => msg!("Error: StakeListAndPoolOutOfDate"),
+            StakePoolError::AlreadyInUse => msg!("Error: The account cannot be initialized because it is already being used"),
+            StakePoolError::InvalidProgramAddress => msg!("Error: The program address provided doesn't match the value generated by the program"),
+            StakePoolError::InvalidState => msg!("Error: The token swap state is invalid"),
```
joncinque, line 1275:
```suggestion
            StakePoolError::InvalidState => msg!("Error: The stake pool state is invalid"),
```

joncinque (commented):
Looks really close!  Just a couple of comments and questions

joncinque:
Also, it looks like some of the stake pool tests are failing in the CI, so please take a look at those as well

stake-pool/cli/src/main.rs
ysavchenko: Actually pool token has the same precision as SOL, so we still have to convert it. Probably need to set up a different function (like `pool_token_float_to_int`) or just keep `sol_to_lamports`?

stake-pool/program/src/processor.rs
ysavchenko: I had in mind that both these structures might change format independently and in this case only one of them has to be migrated. For example if `ValidatorStakeList` changes there will be no need to migrate pool state.

stake-pool/program/src/processor.rs
ysavchenko: Right, this is what inability to test results in, good catch.

ysavchenko:
> Also, it looks like some of the stake pool tests are failing in the CI, so please take a look at those as well

Yes, this is what I was asking about in Discord. We already have `test-bpf` feature in Cargo.toml and it gets used in test environment. However once we try to do `cargo test-bpf` we have to specify feature once again (like this `cargo test-bpf --features test-bpf`) to make it used during compilation. This is why those tests are failing. Can you please check what I'm missing?

stake-pool/cli/src/main.rs
joncinque: Ah I see, we can keep it then -- could you just add a comment explaining the same precision then?

joncinque:
Ok, I see what you mean now -- this gets a bit complicated between cargo and program test.  `cargo test-bpf` actually does two things:
* builds the program's .so using `cargo build-bpf` 
* runs tests using `cargo test --features test-bpf`

At first, it seems like everything should be fine since you're passing in the instruction processor in your tests, so the built .so isn't used for your tests.  Unfortunately, that isn't the case because we try to find a .so when `prefer_bpf` is set to true.  This means that your tests are running against a .so that doesn't have `test-bpf` defined, mimicking a true production environment.

I tried changing the tests to not use the .so, but it triggered a program-test bug that's fixed in 1.5.6.   For now, your best option is to remove the check.  1.5.6 is now released, so in the next PR you can update the tests to properly use time travel.

joncinque:
Looks good, thanks for the work!

PR #1205, opened 2021-02-09
spl-math: Add separate package to test instruction usage

joncinque:
**Problem**

We are using math approximations to have unsigned ints behave like floats, but we don't have a good way to share the math or test the instruction usage of different calculations.

**Solution**

Add an spl-math package to address both of those issues!  The types and approximations defined in it can be reused across programs, and we can eventually publish this crate with other spl utilities.

`PreciseNumber` is directly copied from token-swap, which needs a lot of work, but this is just about isolating the PreciseNumber instruction usage.  There's no need to read through the PreciseNumber code in this review, except for copy/paste issues.  Improvements can come in later work.

Ideas for improvement:
* use the approach in lending of implementing traits like `TryAdd` and returning `Result` instead of `Option`
* change PreciseNumber to a U192 internally, which cuts down instructions by 1/4
* update swap to use the u64 sqrt approximation

Since the PreciseNumber implementation of square root is so instruction-heavy, using 350k instructions to get the root of u64::MAX (!!), this also adds a u64-only approximation which uses less than 10k instructions.  For ease of review, that change is all in [9ced577](https://github.com/solana-labs/solana-program-library/commit/9ced57750af89e9b4cb9e02d428b395b51e8ba6a) -- let me know if you prefer to move that to a separate PR


math/program/src/processor.rs
```diff
@@ -0,0 +1,31 @@
+//! Program state processor
+
+use {
+    crate::{approximations::sqrt, instruction::MathInstruction, precise_number::PreciseNumber},
+    borsh::BorshDeserialize,
+    solana_program::{account_info::AccountInfo, entrypoint::ProgramResult, msg, pubkey::Pubkey},
+};
+
+/// Instruction processor
+pub fn process_instruction(
```
jstarry, line 10: Is this intended to be used by other programs are is it just for testing? If just for testing, maybe we can `mv math/program libraries/math` and just include the entrypoint for testing instruction counts

math/program/src/uint.rs
```diff
@@ -0,0 +1,45 @@
...
+construct_uint! {
+    pub struct U256(4);
+}
+construct_uint! {
+    pub struct U192(3);
+}
+
+impl U256 {
+    /// Returns self to the power of b
+    pub fn checked_u8_power(&self, b: u8) -> Option<U256> {
```
jstarry, line 20: This feels too inefficient to add to the library, binary decomposition is always faster right?

jstarry:
I think we should scrap the general newtonian root calculations for now, they add extra bloat for square root because there's a multiplication by one and power to one that add extra instructions. Also, I think having a trait that can be applied to all unsigned integers would be best. There's a developer asking for `u128` for example, and I think we could implement that more easily with a trait and/or macro

math/program/src/uint.rs
joncinque: This is being used by the stable curve implementation, and I didn't want to change it -- I'll move it back to swap

joncinque: Done in be1c771

joncinque:
Using number generics in `8d8a8f8` and removed generic newtonian approximation.  You were right, we went down from roughly 8k instructions to 2k.  I also tested to see the difference in number of instructions between the generic implementation and the u64 implementation, and they're exactly the same -- compilers are smart!

jstarry:
Yay zero cost abstractions!

math/program/src/processor.rs
joncinque: Done in fe78b0a

jstarry:
Looks great, do you mind adding an instruction count test for u128 as well?

jstarry (dismissed):

joncinque:
I worked on a clever (read: overly-complicated) template-based approach for different instructions, but it was very messy with almost nothing to gain.  We can figure out a macro later on if needed.

PR #1226, opened 2021-02-12
CRUD program: create / update / delete

joncinque:
This shows how to create a simple program for CRUD operations using Borsh and the new utilities added in https://github.com/solana-labs/solana/pull/15290

The read part through JS is not implemented just yet.  That will either be built on top of this PR or in a separate one.

This will stay as a draft until the monorepo PR lands and the solana crates exist with the needed utilities, at which point `Cargo.toml` will be updated.


crud/program/src/instruction.rs
```diff
@@ -0,0 +1,135 @@
...
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, PartialEq)]
+pub enum CrudInstruction {
+    /// Create a new document
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    /// 0. `[writeable, signer]` Data account, must be uninitialized
+    /// 1. `[]` Document owner
+    /// 2. `[]` Rent sysvar, to check for rent exemption
+    Create {
```
mvines, line 21: We typically use `Initialize` here instead of `Create`, since the account was created by the system program.
A couple comments:
1. The data account should not be the signer, as this means this program won't be compatible with `SystemInstruction::CreateAccountWithSeed`.
2. Provide an authority address to `Initialize` for the account owner.  This doesn't need be a signer

So like other programs, `Initialize` is now a fully unsigned instruction and must be issued in the same transaction as the account creation.

I feel like `data: Data` could potentially be removed for simplicity,  the user just immediately does an `Update` after `Initialize` (potentially within the same transaction for atomicity)

crud/program/src/instruction.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// Update the provided document account
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    /// 0. `[writeable]` Document account, must be previously initialized (version != 0)
+    /// 1. `[signer]` Current owner of the document
+    Update {
+        /// Data to replace the existing document data
+        data: Data,
```
mvines, line 34: Supporting partial writes would be really nice

crud/program/src/instruction.rs
```diff
@@ -0,0 +1,135 @@
...
+    },
+
+    /// Update the provided document account
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    /// 0. `[writeable]` Document account, must be previously initialized (version != 0)
+    /// 1. `[signer]` Current owner of the document
+    Update {
```
mvines, line 32: maybe?
```suggestion
    Write {
```

crud/program/src/instruction.rs
```diff
@@ -0,0 +1,135 @@
...
+    },
+
+    /// Delete the provided document account, draining lamports to recipient account
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    /// 0. `[writeable]` Document account, must be previously initialized (version != 0)
+    /// 1. `[signer]` Owner of the document
+    /// 2. `[]` Receiver of drained lamports
+    Delete,
```
mvines, line 44: how about this, for consistency with SPL Token:
```suggestion
    CloseAccount,
```

crud/program/src/instruction.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// Accounts expected by this instruction:
+    ///
+    /// 0. `[writeable, signer]` Data account, must be uninitialized
+    /// 1. `[]` Document owner
+    /// 2. `[]` Rent sysvar, to check for rent exemption
+    Create {
+        /// Data to be filled into the account
+        data: Data,
+    },
+
```
mvines, line 25: I think we'll want a `SetAuthority` instruction as well

mvines (commented):
I mostly just reviewed the instruction interface here.  Once we're settled on this, I'll dig further into the implementation!

crud/program/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "spl-crud"
```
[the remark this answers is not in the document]
joncinque: It's a strange word to see written so much.  How about "SPL Persistent Storage"?  Otherwise I also like "SPL Record"

crud/program/src/instruction.rs
joncinque: That all makes sense.  Regarding point 2, how would that be different from the `1. Document owner` passed in?

crud/program/src/instruction.rs
joncinque: Do you have a neat way in mind to do this?  We could come up with a derivable traits on `Data` which allows for merging

crud/program/src/instruction.rs
joncinque: Done in f63408d

crud/program/src/instruction.rs
joncinque: Done in f63408d

crud/program/src/instruction.rs
joncinque: Done in f63408d

crud/program/src/instruction.rs
joncinque: Done in e9b8f98 -- let me know if something else needs to be done on the authority

crud/program/Cargo.toml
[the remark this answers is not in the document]
mvines: I think I prefer "SPL Record" over ""SPL Persistent Storage" myself

crud/program/src/state.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Struct wrapping data and providing metadata
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, BorshSchema, PartialEq)]
+pub struct AccountData {
+    /// Struct version, allows for upgrades to the program
+    pub version: u8,
+
+    /// The account allowed to update the data
+    pub authority: Pubkey,
+
+    /// The data contained by the account, could be anything or serializable
```
mvines, line 16:
```suggestion
    /// The data contained by the account, could be anything serializable
```

crud/program/src/state.rs
```diff
@@ -0,0 +1,87 @@
...
+use {
+    borsh::{BorshDeserialize, BorshSchema, BorshSerialize},
+    solana_program::{program_pack::IsInitialized, pubkey::Pubkey},
+};
+
+/// Struct wrapping data and providing metadata
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, BorshSchema, PartialEq)]
+pub struct AccountData {
+    /// Struct version, allows for upgrades to the program
+    pub version: u8,
```
mvines, line 11: How do you imagine `version` being used?  For now it would always be set to 1 on Initialize?

crud/program/src/instruction.rs
mvines: I was just thinking of `Write { offset: u32, data: Vec<u32> }`, nothing fancy

crud/program/src/instruction.rs
mvines: lgtm

crud/program/src/state.rs
joncinque: The concept is that a program developer iterates on a record, they might need to modify the record type by adding / removing fields, so the version allows the program to handle all types

joncinque: eventually there could also be an "UpdateVersion" instruction that would take a v1 and turn it into v2, and then Initialize would set new accounts to v2

mvines: I feel like I'm missing an important piece of the motivation behind this program, as it seems to me that versioning of the data should be a client concern. 

"UpdateVersion" could be either an atomic `CloseAccount`, `Initialize`, `Write`, or once we support account resizing (with a new version of this program deployed) could be an atomic `Resize` (new instruction) and `Write`.

joncinque: I see what you mean -- that certainly works too.  No matter what, we need to know if the account data is initialized with some field at the start.  By having that field as `version` instead of just `is_initialized`, it can carry additional information, to be used by the client or not.  If you think it's clearer to change to an `is_initialized` field, I don't have any objections

mvines: Got it, seems ok to me as is then :)

crud/program/src/instruction.rs
joncinque: Done in d57870f

crud/program/Cargo.toml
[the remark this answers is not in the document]
joncinque: Done in fa089d3

record/program/src/instruction.rs
```diff
@@ -0,0 +1,176 @@
...
+
+/// Instructions supported by the program
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, PartialEq)]
+pub enum RecordInstruction {
+    /// Create a new document
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    /// 0. `[writable]` Data account, must be uninitialized
+    /// 1. `[]` Document authority
```
mvines, line 19: "Document" and "Document authority" are used in a couple places.  If we're going with Record then I think these should be updated to match.

record/program/src/processor.rs
```diff
@@ -0,0 +1,118 @@
...
+            msg!("RecordInstruction::Initialize");
+
+            let data_info = next_account_info(account_info_iter)?;
+            let owner_info = next_account_info(account_info_iter)?;
+            let rent_sysvar_info = next_account_info(account_info_iter)?;
+            let rent = &Rent::from_account_info(rent_sysvar_info)?;
+
+            let mut account_data = AccountData::try_from_slice(*data_info.data.borrow())?;
+            if account_data.is_initialized() {
+                return Err(ProgramError::AccountAlreadyInitialized);
```
mvines, line 42: I've found it pretty useful to add `msg!()` statements for all the error cases, so from the transaction log it's more obvious what went wrong without having to drudge up the program source code

record/program/src/state.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub const CURRENT_VERSION: u8 = 1;
+
+    /// Start of writable account data, after version and authority
+    pub const WRITABLE_START_INDEX: usize = 33;
+}
+
+impl IsInitialized for AccountData {
+    /// Is initialized
+    fn is_initialized(&self) -> bool {
+        self.version != 0
```
mvines, line 43: nit: since we set `self.version` explicitly to CURRENT_VERSION during Initialize, aligning the initialized check here seems nice:
```suggestion
        self.version == CURRENT_VERSION
```

mvines (commented):
Looking good!

record/program/src/error.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Errors that may be returned by the program.
+#[derive(Clone, Debug, Eq, Error, FromPrimitive, PartialEq)]
+pub enum RecordError {
+    /// Incorrect owner provided on update or delete
+    #[error("Incorrect owner provided on update or delete")]
+    IncorrectOwner,
+
+    /// Calculation overflow
+    #[error("Calculation overflow")]
+    Overflow,
```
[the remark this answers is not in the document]
joncinque: I'd prefer to make a new error for `CalculationError` since `UnbalancedInstruction` isn't really the same concept.  I'm happy to do everything to remove `RecordError` though!  Also, not all `InstructionError`s are mapped to `ProgramError`, so we'd have to add those mappings -- is there a reason for that?

record/program/src/instruction.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// 2. `[]` Rent sysvar, to check for rent exemption
```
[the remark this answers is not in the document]
joncinque: The concept was to make the records persistently stored, but there's no express need for that

record/program/src/error.rs
```diff
@@ -0,0 +1,27 @@
...
+use num_derive::FromPrimitive;
+use solana_program::{decode_error::DecodeError, program_error::ProgramError};
+use thiserror::Error;
+
+/// Errors that may be returned by the program.
+#[derive(Clone, Debug, Eq, Error, FromPrimitive, PartialEq)]
+pub enum RecordError {
+    /// Incorrect owner provided on update or delete
+    #[error("Incorrect owner provided on update or delete")]
+    IncorrectOwner,
```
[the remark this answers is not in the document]
joncinque: No mapping to ProgramError though https://github.com/solana-labs/solana/blob/master/sdk/program/src/program_error.rs#L165 shall I add all of the instruction errors to program error?

record/program/src/instruction.rs
```diff
@@ -0,0 +1,176 @@
...
+    Initialize,
+
+    /// Write to the provided data account
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    /// 0. `[writable]` Document account, must be previously initialized (version != 0)
```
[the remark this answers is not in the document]
joncinque: Done in 0d65c8d

record/program/src/instruction.rs
joncinque: Done in 0d65c8d

record/program/src/state.rs
joncinque: Done in 0d65c8d

record/program/src/processor.rs
joncinque: Done in 0d65c8d

record/program/src/instruction.rs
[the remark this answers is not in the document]
joncinque: Done in d167cfd

record/program/src/error.rs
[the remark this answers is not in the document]
mvines: Ah, we only map a subset in now.  I guess we could use `ProgramError::MissingRequiredSignature` here instead

record/program/src/error.rs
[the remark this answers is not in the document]
mvines: A new `ProgramError::Overflow` sounds good now that you mention it.

I'm not sure why we don't have all `InstructionErrors`s mapped into `ProgramError`.  Might have just _happened_ unintentionally.  Do you have any thoughts here Jack?   Mapping most/all of them into `ProgramError` seems nice.  It looks like we might need to be careful with the rollout of such a change though, as it seems to me like we could risk a bank hash mismatch if a program generates a new `ProgramError` before the supermajority has upgraded to know about it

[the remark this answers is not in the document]
joncinque: tagging @jackcmay in case he didn't see ^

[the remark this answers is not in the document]
jackcmay: Some instruction errors don't make sense coming from programs but I don't see any reason not to map them anyway.  @mvines What bankhash issue are you worried about?  Something more than adding a new instruction error?

record/program/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+mod entrypoint;
+pub mod error;
+pub mod instruction;
+pub mod processor;
+pub mod state;
+
+// Export current SDK types for downstream users building with a different SDK version
+pub use solana_program;
+
+solana_program::declare_id!("CRUD3SXMKeggKN2FZePp8WLK2RpQvhp1LANGf6a5MLdu");
```
jackcmay, line 13: Update grinded id to match new name?

record/program/src/error.rs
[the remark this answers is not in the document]
mvines: Unknown program errors are mapped to/from `ProgramError::Custom` [here](https://github.com/solana-labs/solana/blob/5d31763a3b48a9199861fad09166a0d61719fd52/sdk/program/src/program_error.rs#L157).  If we add a new enum entry to ProgramError, the nodes that know about that entry will no longer map it to/from `ProgramError::Custom`.   If a program generates such an error before 66% of the cluster has upgraded to a software version that knows about it, I'm worried we'll get a bank hash mismatch (since the transaction result is encoded in the bank hash) and possibly kill the cluster

record/program/src/lib.rs
joncinque: "record" is a lot of letters, still working on it :sweat_smile:

mvines: Maybe next decade.  `Rec`? :)

joncinque: too easy! `rEcnBBXv6GqDJERL4yLVE5G2mcx7nVTkLHYeqw6mgpk` generated instantly

CriesofCarrots: Hold out for `Rec`, then ;)

joncinque: `ReciQBw6sQKH9TVVJQDnbnJ5W7FP539tPHjZhRF4E9r` :smile:

Issue #993, opened 2020-12-31
lending: Support cross collateralized loans


jstarry:
#### Problem
Loans can only be collateralized by a single currency and collateral cannot be pooled for multiple loans

#### Proposed Changes

Expand `Obligation` to include a list of borrow / collateral amounts
- Add authority which will be allowed to repay obligations (and remove obligation tokenization)
- For each borrow / collateral amount, track the quote currency value and the slot the value were last computed at.
- Reasonably high max of total currencies per obligation*

Obligation market value calculation
- Introduce `RefreshCollateralValue` instruction for computing the market value of obligation collateral
  - Requires reserve and dex accounts for collateral exchange rate and price discovery, respectively
- Introduce `RefreshBorrowValue` instruction for computing the market value of obligation borrow amount
  - Requires reserve and dex accounts for interest accrual and price discovery, respectively

Liquidation
- `Liquidation` instruction now requires an obligation with recently refreshed market values to determine loan health

joncinque:
> Expand Obligation to include a list of borrow / collateral amounts

Would it be easier to have one borrow type and multiple collateral types?  Unless this is to be more like Aave, it could be clearer for people to manage.  Then you might also be able to just have a `RefreshObligation` instruction which takes care of all calculations, rather than separating the collateral and borrow side.

> Add authority which will be allowed to repay obligations (and remove obligation tokenization)

We could also create a tokenized market for the authority, as you mentioned :smile: 

Otherwise, this approach seems good to me!

jstarry:
> Would it be easier to have one borrow type and multiple collateral types? Unless this is to be more like Aave, it could be clearer for people to manage. Then you might also be able to just have a RefreshObligation instruction which takes care of all calculations, rather than separating the collateral and borrow side.

Yeah, not opposed to combine those instructions. I think it's worth the flexibility to allow a many to many relationship for collateral to loans though to support multiple borrow currencies

> Otherwise, this approach seems good to me!

Thanks for looking it over!

PR #1229, opened 2021-02-13
lending: add DepositObligationCollateral and WithdrawObligationCollateral instructions

### LINKED ISSUE #995: lending: No way to adjust obligation's collateral
#### Problem
If an obligation starts to become unhealthy, users have no way of improving the health by increasing the collateral for the loan. Similarly, if the value of collateral goes up, there's no way for a user to withdraw excess collateral

#### Proposed Changes
Add `AdjustObligationCollateral` instruction which supports increasing and decreasing collateral (loan must remain healthy)

jordansexton:
Resolves #995

(Moving discussion from WIP @ https://github.com/jordansexton/solana-program-library/pull/1)

@jordansexton:
> A general impression -- the implementation seems like it could be simpler and have a more obvious API with two instructions (e.g. `DepositObligationCollateral` and `WithdrawObligationCollateral`) instead of one.

@joncinque:
> This is going in a great direction.
> 
> One thing to add are the obligation token account and mint. On a borrow, obligation tokens are minted 1:1 with collateral tokens, as a proof of ownership of the obligation. When adjusting collateral, we also need to mint / burn the right amount of obligation tokens.
> 
> Regarding splitting the instruction, that seems reasonable Since adding collateral can never make a loan less healthy, we can disable a lot of checks and simplify it. @jstarry did you have a reason for keeping it as one instruction?

#### Todo

- [x] Mint & burn obligation tokens
- [x] Spike out what two separate instructions would look like for comparison
- [x] Resolve general questions and remove `@TODO` comments
- [x] Add tests


jordansexton:
@joncinque Thanks for the feedback! I'll update this PR with a checklist of things I plan to do.

jordansexton:
0750108a840d82edea15d0e3f010abb35c3c156f adds obligation token minting and burning, let me know if this looks correct.

@joncinque you mentioned:

> On a borrow, obligation tokens are minted 1:1 with collateral tokens, as a proof of ownership of the obligation. When adjusting collateral, we also need to mint / burn the right amount of obligation tokens.

I see that in `process_borrow`, the amount of obligation tokens minted is the same as the collateral amount:

https://github.com/solana-labs/solana-program-library/blob/7225f0b9419dcc1b88d204f52c86c6c8e7c699b0/token-lending/program/src/processor.rs#L816-L823

And in `process_repay`, the amount of obligation tokens burned is this:

https://github.com/solana-labs/solana-program-library/blob/7225f0b9419dcc1b88d204f52c86c6c8e7c699b0/token-lending/program/src/processor.rs#L952-L959

`obligation_token_amount` comes from `Obligation::repay`:

https://github.com/solana-labs/solana-program-library/blob/7225f0b9419dcc1b88d204f52c86c6c8e7c699b0/token-lending/program/src/state/obligation.rs#L127-L138

What mechanism causes the amount of obligation tokens to not be the same when burned?

jordansexton:
`DepositObligationCollateral` and `WithdrawObligationCollateral` instructions implemented.

Pro:
- fewer code smells (eliminates conditionals and makes parameters appropriately explicit)
- easier to understand and test
- performance might be faster? (probably negligible)

Con:
- some repeated chunks of code (mostly safety checks, probably not helpful to extract to another method)

Pro or con:
- similar instructions from the caller's perspective but they have different account list signatures (`dex_market_*` accounts aren't used when depositing)

@jstarry what do you think?

jstarry:
> similar instructions from the caller's perspective but they have different account list signatures (dex_market_* accounts aren't used when depositing)
> @jstarry what do you think?

Yeah, that interface looks good. I didn't consider the different set of accounts needed when suggesting using a single instruction. Having two is easier to reason about 👍🏻

jstarry:
> What mechanism causes the amount of obligation tokens to not be the same when burned?

If an obligation is liquidated at some point, some of its collateral will be given to the liquidator but the liquidator doesn't pay with obligation tokens. So really they just represent a fractional share of the locked collateral, but in the happy case, the amount of obligation tokens and collateral tokens will match.

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,363 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+    )?;
+
+    // @TODO: this seems simpler than obligation.loan_to_value but is there a tradeoff?
+    let borrow_amount = obligation.borrowed_liquidity_wads.try_ceil_u64()?;
+    // @TODO: should this be using borrow_reserve?
+    let required_collateral_amount = withdraw_reserve.required_collateral_for_borrow(
+        borrow_amount,
+        &borrow_reserve.liquidity.mint_pubkey,
+        trade_simulator,
+    )?;
```
jstarry, line 1492: This method was intended for borrows only because it uses some special logic to see how much collateral would be needed to execute orders on the dex market to get a certain amount of currency you want to borrow. I think this withdraw logic should probably mirror what we are doing for liquidations where we use the best bid to determine the market price. I think it'd be reasonable to restrict withdraws to keeping the loan about the ltv ratio so that there's still a buffer against the liquidation threshold.

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,363 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+
+    if obligation_collateral_amount - collateral_amount < required_collateral_amount {
+        return Err(LendingError::UnhealthyObligation.into());
+    }
```
jstarry, line 1496: This is safe due to your check above but it's good practice to always use checked math to prevent against future bugs/refactors. I would recommend computing `remaining_collateral_amount` and then reusing that below instead of doing the subtraction twice

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,363 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+
+    let obligation_token_amount = {
+        let withdraw_pct =
+            Decimal::from(collateral_amount).try_div(obligation_collateral_amount)?;
+        let token_amount: Decimal = withdraw_pct.try_mul(obligation_token_mint.supply)?;
+        token_amount.try_floor_u64()?
+    };
```
jstarry, line 1503: It'd be nice to pull this out into a common method since it's used in repay as well

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,363 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+    // Check obligation health
+    let trade_simulator = TradeSimulator::new(
+        dex_market_info,
+        dex_market_orders_info,
+        memory,
+        &lending_market.quote_token_mint,
+        // @TODO: are these mint accounts correct?
+        &borrow_reserve.liquidity.mint_pubkey,
+        &withdraw_reserve.liquidity.mint_pubkey,
+    )?;
```
jstarry, line 1483: Hmm this trade simulator feels a little awkward for withdraws. Instead of trading currencies, we really only need to judge the health of the obligation by getting the best price like we do for liquidation. Probably better as a separate change, but any ideas on how we could better split the trade simulator from just fetching the best price?

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,363 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+
+    // @TODO: this seems simpler than obligation.loan_to_value but is there a tradeoff?
+    let borrow_amount = obligation.borrowed_liquidity_wads.try_ceil_u64()?;
```
jstarry, line 1486: nit: It'd be better to use the more precise `Decimal` directly for currency conversions rather than round to integer here

jstarry (commented):
Nice progress! I took a look at your inline TODO's and gave some extra context, feel free to tag me about anything that's unclear

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,363 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+    let borrow_reserve_info = next_account_info(account_info_iter)?;
+    let obligation_info = next_account_info(account_info_iter)?;
+    let obligation_token_mint_info = next_account_info(account_info_iter)?;
+    let obligation_token_output_info = next_account_info(account_info_iter)?;
+    let lending_market_info = next_account_info(account_info_iter)?;
+    let lending_market_authority_info = next_account_info(account_info_iter)?;
+    let user_transfer_authority_info = next_account_info(account_info_iter)?;
+    let memory = next_account_info(account_info_iter)?;
+    let clock = &Clock::from_account_info(next_account_info(account_info_iter)?)?;
+    let token_program_id = next_account_info(account_info_iter)?;
```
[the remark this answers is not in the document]
jordansexton: Thinking harder about what these are doing, that makes sense. Thank you!

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,363 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+        return Err(LendingError::InvalidTokenProgram.into());
+    }
+
+    let deposit_reserve = Reserve::unpack(&deposit_reserve_info.data.borrow())?;
+    // @TODO: check if this logic is valid for withdraw
+    if deposit_reserve_info.owner != program_id {
+        return Err(LendingError::InvalidAccountOwner.into());
+    }
+    // @TODO: check if & reference is valid
+    if &deposit_reserve.lending_market != lending_market_info.key {
```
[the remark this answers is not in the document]
jordansexton: I think what I was missing conceptually is how the shared borrow operator in Rust works.

Screen Shot 2021-02-14 at 11 41 35 AM

This suggests `pubkey1` and `pubkey2` are equivalent (but not equal), and `pubkeyRef1` and `pubkeyRef2` are equal.

Basically, I was thinking that `&deposit_reserve.lending_market` might be equivalent to `(&deposit_reserve).lending_market`, but instead it's `&(deposit_reserve.lending_market)`, so what you said makes sense.

token-lending/program/src/processor.rs
jordansexton: Okay, got it. I'll rework this.

token-lending/program/src/processor.rs
jordansexton: Yeah, I was wondering about this. It seemed wasteful to do it twice, but I wasn't sure if subtracting without the check was somehow going to do something different in an assignment versus a comparison. Thanks!

token-lending/program/src/processor.rs
jordansexton: Done.

token-lending/program/src/processor.rs
jordansexton: I'll probably end up removing this when I resolve https://github.com/solana-labs/solana-program-library/pull/1229#discussion_r575806271 but this is good to know.

token-lending/program/src/processor.rs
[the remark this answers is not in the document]
jstarry: Yup, looks like you got it. Worth pointing out that `Pubkey` implements the `Copy` trait so in the first two lines, you're copying the supply pubkey twice, once for `pubkey1` and once for `pubkey2`. If `Pubkey` didn't implement `Copy`, Rust would _move_ the value out of the collateral struct. And the compiler would complain because you can only do that once.

token-lending/program/src/processor.rs
jordansexton: Okay, I've refactored this in c0b9d7c554b3f4123a5692f9e3eb0664a3fd07c2. Would you check the comments there and tell me if I'm thinking about this correctly?

token-lending/program/src/processor.rs
jordansexton: Removed by c0b9d7c554b3f4123a5692f9e3eb0664a3fd07c2.

token-lending/program/src/state/reserve.rs
```diff
@@ -92,6 +92,73 @@ impl Reserve {
...
+        } else if loan_to_value_ratio.gt(&Decimal::one()) {
+            // unhealthy, at least `deposit_amount` of collateral should be deposited
+            deposit_amount = decimal_collateral
+                .try_mul(loan_to_value_ratio)?
+                .try_ceil_u64()?;
+            withdraw_amount = 0;
+        } else {
+            deposit_amount = 0;
+            withdraw_amount = 0;
+        }
```
jordansexton, line 153: @jstarry My understanding of the ratios:

If an obligation's loan to value ratio is greater than the reserve's loan to value ratio, that implies the loan is getting unhealthy, and additional collateral should be deposited before it gets to the liquidation threshold. Conversely, if an obligation's LTV is less than the reserve LTV, that implies the loan is healthy and some collateral could be withdrawn. The max collateral that should be allowed to be withdrawn is the amount that would bring the obligation's LTV to the reserve LTV.

Therefore, the health of an obligation can be thought of as the ratio of the obligation's LTV to the reserve LTV, which simplifies the math. Is that correct?

jordansexton: I asked for some feedback from @joncinque and simplified this implementation in 447c49862b0c33ed04a8559ba2777d75a978820d.

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,363 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+
+    let obligation_collateral_amount = obligation.deposited_collateral_tokens;
+    if obligation_collateral_amount == 0 {
+        return Err(LendingError::ObligationEmpty.into());
+    }
+
+    // @TODO: can these next checks be combined? e.g. obligation_collateral_amount <= collateral_amount
+    if obligation_collateral_amount < collateral_amount {
+        // @TODO: does this need a different error type?
+        return Err(LendingError::ObligationEmpty.into());
```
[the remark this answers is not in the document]
jordansexton: Done.

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,363 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+    }
+
+    if obligation_collateral_amount == collateral_amount {
+        // @TODO: do obligations ever allow withdrawing all collateral?
+        return Err(LendingError::ObligationEmpty.into());
```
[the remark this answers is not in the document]
jordansexton: 447c49862b0c33ed04a8559ba2777d75a978820d ended up replacing this with a simpler `recommended_collateral` for health of the obligation.

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,315 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+    let trade_simulator = TradeSimulator::new(
+        dex_market_info,
+        dex_market_orders_info,
+        memory,
+        &lending_market.quote_token_mint,
+        &borrow_reserve.liquidity.mint_pubkey,
+        &withdraw_reserve.liquidity.mint_pubkey,
+    )?;
+
+    let recommended_collateral = withdraw_reserve.recommended_collateral(
```
jstarry, line 1440: How about `minimum_collateral`?

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,315 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+        &obligation,
+        &borrow_reserve.liquidity.mint_pubkey,
+        trade_simulator,
+    )?;
+    if obligation_collateral_amount < recommended_collateral {
+        return Err(LendingError::ObligationUnhealthy.into());
+    }
```
jstarry, line 1447: `minimum_collateral` is determined by LTV but the liquidation threshold is what determines whether an obligation is unhealthy. Maybe the error could be more specific by saying that the collateral is below the minimum?

token-lending/program/src/processor.rs
```diff
@@ -1176,6 +1184,315 @@ fn process_accrue_interest(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pro
...
+
+    let remaining_collateral = obligation_collateral_amount
+        .checked_sub(collateral_amount)
+        .ok_or(LendingError::MathOverflow)?;
+    if remaining_collateral < recommended_collateral {
+        return Err(LendingError::UnhealthyObligation.into());
```
jstarry, line 1453: Same as above, not necessarily unhealthy

token-lending/program/src/state/reserve.rs
```diff
@@ -92,6 +92,47 @@ impl Reserve {
...
+    ) -> Result<u64, ProgramError> {
+        let borrow_token_price = token_converter.best_price(liquidity_token_mint)?;
+        let decimal_collateral = Decimal::from(obligation.deposited_collateral_tokens);
+        let collateral_value = collateral_exchange_rate
+            .decimal_collateral_to_liquidity(decimal_collateral)?
+            .try_div(borrow_token_price)?;
+
+        let loan_to_value = obligation
+            .borrowed_liquidity_wads
+            .try_div(collateral_value)?;
```
jstarry, line 126: FYI obligation has a loan_to_value method

token-lending/program/src/state/reserve.rs
```diff
@@ -376,6 +417,17 @@ pub struct LiquidateResult {
     pub repay_amount: u64,
 }
 
+/// Obligation health result
+#[derive(Debug)]
+pub struct ObligationHealthResult {
```
jstarry, line 422: nit: still used?

jstarry (commented):
looking good!

token-lending/program/src/state/reserve.rs
```diff
@@ -92,6 +92,73 @@ impl Reserve {
...
+    ) -> Result<ObligationHealthResult, ProgramError> {
+        // @TODO: does this approach make sense?
+        let borrow_token_price = token_converter.best_price(liquidity_token_mint)?;
+        let decimal_collateral = Decimal::from(obligation.deposited_collateral_tokens);
+        let collateral_value = collateral_exchange_rate
+            .decimal_collateral_to_liquidity(decimal_collateral)?
+            .try_div(borrow_token_price)?;
+        let loan_to_value = obligation
+            .borrowed_liquidity_wads
+            .try_div(collateral_value)?;
```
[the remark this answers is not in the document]
jordansexton: Haha. Right. I was finding it helpful to inline the method when I was reasoning about the code... and then didn't reason to remove it. Thanks.

token-lending/program/src/error.rs
```diff
@@ -125,6 +125,17 @@ pub enum LendingError {
...
     #[error("Token burn failed")]
     TokenBurnFailed,
+
+    // 35
+    /// Cannot withdraw from an unhealthy obligation
+    #[error("Cannot withdraw from an unhealthy obligation")]
+    ObligationUnhealthy,
+    /// Cannot make obligation unhealthy
+    #[error("Cannot make obligation unhealthy")]
+    UnhealthyObligation,
```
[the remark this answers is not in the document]
jordansexton: Agreed

token-lending/program/src/processor.rs
jordansexton: Much better

jordansexton:
@jstarry thank you for the review, all good points. Please let me know if 3d0c82850d29815f32a9b0dd5a47a3c9e6cc4570 addresses things. I'm going to work on testing next.

token-lending/program/src/error.rs
```diff
@@ -125,6 +125,17 @@ pub enum LendingError {
...
+    // 35
+    /// Invalid obligation collateral amount
+    #[error("Invalid obligation collateral amount")]
+    InvalidObligationCollateral,
+    /// Obligation collateral is below minimum
+    #[error("Obligation collateral is below minimum")]
+    ObligationCollateralBelowMinimum,
+    /// Obligation collateral cannot be withdrawn to below minimum
+    #[error("Obligation collateral cannot be withdrawn to below minimum")]
+    ObligationCollateralWithdrawMinimum,
```
jordansexton, line 138: Is it worth thinking about these with respect to https://github.com/solana-labs/solana-program-library/issues/993 or another other issues? I want to make sure the errors are sufficiently descriptive and non-overlapping with other obligation collateral changes

jstarry: I think the current descriptions are sufficient but can always update if needed when implementing that issue

jordansexton:
680bfb205ffefcd01aa8b7c0be270391bb0e4bac fixes the order of the source and destination collateral params to follow the convention of the other instructions.

jordansexton:
6d2606c46c6ea9f95e1e07c08e1eef16dde93617 adds a unit test for `Reserve::minimum_collateral`

token-lending/program/src/state/reserve.rs
```diff
@@ -92,6 +92,41 @@ impl Reserve {
...
+        let borrow_token_price = token_converter.best_price(liquidity_token_mint)?;
+        let loan_to_value =
+            obligation.loan_to_value(collateral_exchange_rate, borrow_token_price)?;
+        let reserve_loan_to_value =
+            Decimal::from_percent(collateral_reserve_config.loan_to_value_ratio);
+        let loan_to_value_ratio = loan_to_value.try_div(reserve_loan_to_value)?;
+
+        Decimal::from(obligation.deposited_collateral_tokens)
+            .try_mul(loan_to_value_ratio)?
+            .try_ceil_u64()
```
jstarry, line 127: This is overly complex. You can get by with `borrow_amount / reserve_loan_to_value` to calculate `minimum_collateral`. Can you simplify and change the signature of `minimum_collateral` to accept a borrow amount instead of an obligation?

token-lending/program/src/state/reserve.rs
```diff
@@ -953,6 +988,75 @@ mod test {
             }
         }
 
+        #[test]
+        fn minimum_collateral(
+            collateral_amount in 0..=u32::MAX as u64,
```
jstarry, line 993: Can you change this to a range over `borrow_amount` instead? That's more what's being tested

token-lending/program/src/state/reserve.rs
```diff
@@ -953,6 +988,75 @@ mod test {
...
+            // should be slightly more or equal to the given loan to value ratio.
+            obligation.deposited_collateral_tokens = minimum_collateral - 1;
+            let loan_to_value_below_minimum = obligation.loan_to_value(
+                collateral_exchange_rate,
+                borrow_token_price
+            )?;
+
+            // Assert that reversing the calculation returns approx loan to value ratio
+            assert!(loan_to_value_at_minimum <= decimal_loan_to_value_ratio);
+            assert!(loan_to_value_below_minimum >= decimal_loan_to_value_ratio);
```
jstarry, line 1057: Looks good!

jstarry (commented):
@jordansexton we don't code review drafts typically. Can you please mark your PR as ready for review?

token-lending/program/src/state/reserve.rs
jordansexton: I'm wondering in that case if this function is even needed. Wouldn't this do almost the same thing?

```rust
let required_collateral = withdraw_reserve.required_collateral_for_borrow(
    obligation.borrowed_liquidity_wads.try_ceil_u64()?,
    &borrow_reserve.liquidity.mint_pubkey,
    trade_simulator,
)?;
```

The main difference is then that `minimum_collateral` uses `token_converter.best_price` and `required_collateral_for_borrow` uses `token_converter.convert` per https://github.com/solana-labs/solana-program-library/pull/1229#discussion_r575806271

But is this distinction worth having a separate function that's so similar?

token-lending/program/src/state/reserve.rs
jordansexton: Hmm. If `borrow_amount` is what we want to test, then this tests the same inputs as https://github.com/solana-labs/solana-program-library/blob/a99e052566e52967ce0404f2e77b3799835aeb54/token-lending/program/src/state/reserve.rs#L1014-L1019

Using `borrow_amount` with `required_collateral_for_borrow` was [an earlier approach](https://github.com/solana-labs/solana-program-library/pull/1229/files/f65021b4d09088be30792deceeaeebf6b46b1983#diff-724dc1de5d8ee8e4daf8d2a6d0c0c12bde2a03921e3cc74bcb75ab0caf3d00f1R1484-R1492) we considered. If this approach is okay, then maybe I can just remove this function and test?

token-lending/program/src/state/reserve.rs
jstarry: Yeah, it's debatable. The current thought is that the token converter method gives you an impact price which is more resistant to manipulation which prevents people from taking out borrowing more than they should. On the other hand, the health of a loan is judged on the best price because we want to make liquidations fair. Otherwise, if we also used impact price for liquidations, the size of a loan would make an obligation more or less vulnerable to liquidation which didn't seem fair. More details here: https://github.com/solana-labs/solana-program-library/pull/1119

token-lending/program/src/state/reserve.rs
jstarry: Required collateral for borrow has special logic for calculating the impact price when taking out a new borrow position. We don't want to do that for withdraws because the slippage could prevent borrowers from taking out a reasonable amount of their collateral. I think I was wrong here, the withdraw should allow borrowers to take out collateral just up until the liquidation point if they want to.

We could even add a parameter for them to decide what they want the resulting LTV to be after their withdrawal. For example, if liquidation is at 80% and the user wants to live on the edge, they can specify that they want to pull out up to 78% LTV. @bartosz-lipinski do you think allowing users to specify an LTV as a constraint would be useful?

token-lending/program/src/state/reserve.rs
jordansexton: Got it, that all makes sense. In this case, since it doesn't affect liquidation, my thinking is that `withdraw_obligation_collateral` should prevent you from withdrawing more collateral than you should based on the impact price, because if you can borrow and then manipulate the best price, you can effectively take out more borrowing than you should after the fact.

token-lending/program/src/state/reserve.rs
jordansexton: > Required collateral for borrow has special logic for calculating the impact price when taking out a new borrow position. We don't want to do that for withdraws because the slippage could prevent borrowers from taking out a reasonable amount of their collateral.

What happens if you max borrow (using the impact price) then immediately withdraw collateral (using the best price)? It seems that if your max borrow is determined by your collateral, then your min collateral should be determined by your borrow using the same price.

> the withdraw should allow borrowers to take out collateral just up until the liquidation point if they want to

Similar to above, but could you take out a loan, manipulate the best price for the collateral and/or liquidity to decrease the obligation's LTV, and withdraw more collateral than you should?

jstarry: Yeah, good points. Makes the most sense to keep borrow/withdraw collateral paths the same otherwise we just have a loophole. So, I think agree with your earlier point now about just using the same method and nixing minimum_collateral..

jordansexton: Removed in dcdcf9e

token-lending/program/src/state/reserve.rs
jordansexton: Removed in dcdcf9e

jordansexton:
The test added in 24fea1ba4d69e67d83c0cb26d3540d670b489225 is failing:

```
➜  solana-program-library git:(adjust-obligation-collateral) cargo test --color=always --package spl-token-lending --test deposit_obligation_collateral --no-fail-fast --all-features
    Finished test [unoptimized + debuginfo] target(s) in 0.77s
     Running target/debug/deps/deposit_obligation_collateral-4175f2f7df25afcd

running 1 test
[2021-03-02T23:31:57.489173000Z INFO  solana_program_test] "spl_token_lending" program loaded as native code
[2021-03-02T23:31:57.618554000Z DEBUG solana_runtime::message_processor] Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [1]
[2021-03-02T23:31:57.619002000Z DEBUG solana_runtime::message_processor] Program log: Instruction: Approve
[2021-03-02T23:31:57.619908000Z DEBUG solana_rbpf::vm] BPF instructions executed: 5
[2021-03-02T23:31:57.619935000Z DEBUG solana_rbpf::vm] Max frame depth reached: 6
[2021-03-02T23:31:57.620006000Z DEBUG solana_runtime::message_processor] Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 3294 of 50000 compute units
[2021-03-02T23:31:57.620175000Z DEBUG solana_runtime::message_processor] Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
[2021-03-02T23:31:57.620563000Z DEBUG solana_runtime::message_processor] Program log: Instruction: Deposit Obligation Collateral
[2021-03-02T23:31:57.620742000Z DEBUG solana_runtime::message_processor] Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [1]
[2021-03-02T23:31:57.628181000Z DEBUG solana_runtime::message_processor] Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [2]
[2021-03-02T23:31:57.628567000Z DEBUG solana_runtime::message_processor] Program log: Instruction: Transfer
[2021-03-02T23:31:57.629992000Z DEBUG solana_rbpf::vm] BPF instructions executed: 5
[2021-03-02T23:31:57.630016000Z DEBUG solana_rbpf::vm] Max frame depth reached: 6
[2021-03-02T23:31:57.630076000Z DEBUG solana_runtime::message_processor] Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 5062 of 50000 compute units
[2021-03-02T23:31:57.630173000Z DEBUG solana_runtime::message_processor] Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
test test_success ... test test_success has been running for over 60 seconds
test test_success ... FAILED

failures:

---- test_success stdout ----
thread 'solana-bank-forks-client' panicked at 'Account data resizing not supported yet: 165 -> 0. Consider making this test conditional on `#[cfg(feature = "test-bpf")]`', /Users/jordan/.cargo/registry/src/github.com-1ecc6299db9ec823/solana-program-test-1.5.4/src/lib.rs:298:25
thread 'test_success' panicked at 'assertion failed: banks_client.process_transaction(transaction).await.is_ok()', token-lending/program/tests/deposit_obligation_collateral.rs:130:5


failures:
    test_success

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '-p spl-token-lending --test deposit_obligation_collateral'
```

Feels like it's something obvious that I'm messing up. Thoughts?

jordansexton:
> The test added in [24fea1b](https://github.com/solana-labs/solana-program-library/commit/24fea1ba4d69e67d83c0cb26d3540d670b489225) is failing:

I wasn't running it with `cargo test-bpf`. All good now. Thanks @mvines

jordansexton:
Tests added and passing. Ready for review!

token-lending/program/tests/deposit_obligation_collateral.rs
```diff
@@ -0,0 +1,140 @@
...
+    );
+
+    let (mut banks_client, payer, recent_blockhash) = test.start().await;
+
+    let initial_collateral_supply_balance =
+        get_token_balance(&mut banks_client, sol_reserve.collateral_supply).await;
+    let initial_user_collateral_balance =
+        get_token_balance(&mut banks_client, sol_reserve.user_collateral_account).await;
+    let initial_obligation_token_balance =
+        get_token_balance(&mut banks_client, obligation.token_account).await;
```
jstarry, line 83: Could you please assert that these initial balances are greater than 0? The base case of 0 is probably worth testing too but we expect deposit collateral to be used for non-zero obligations

token-lending/program/tests/withdraw_obligation_collateral.rs
```diff
@@ -0,0 +1,183 @@
...
+    math::Decimal,
+    processor::process_instruction,
+    state::{INITIAL_COLLATERAL_RATIO, SLOTS_PER_YEAR},
+};
+
+const LAMPORTS_TO_SOL: u64 = 1_000_000_000;
+const FRACTIONAL_TO_USDC: u64 = 1_000_000;
+
+#[tokio::test]
+async fn test_success() {
```
jstarry, line 25: This looks great! Could you add a failing test for this too?

jstarry (approved):
Awesome work, once you make the test adjustments, feel free to merge!

token-lending/program/tests/withdraw_obligation_collateral.rs
jordansexton: Done in 0e8810fc9feb30a819cf03675d90a69b78961db5. Also made the passing test more strict by withdrawing the maximum allowed amount.

token-lending/program/tests/deposit_obligation_collateral.rs
jordansexton: I changed this test in 1a0f35d8385549c0eae00e4926608f489247d026 to use the same (nonzero) obligation collateral and liquidity balances as the withdraw test. Does it still make sense to assert the balances since they aren't the function being tested?

token-lending/program/tests/withdraw_obligation_collateral.rs
jordansexton: What's the right convention for tests like these that repeat a bunch of setup code? I get the itch to abstract

token-lending/program/tests/deposit_obligation_collateral.rs
jstarry: It's sometimes nice to enforce the setup for what *is* being tested, in this case I don't think it matters

token-lending/program/tests/withdraw_obligation_collateral.rs
jstarry: Reducing setup code sounds great! The more setup, the more tedious it is to add more tests, which is definitely not great

PR #1277, opened 2021-02-18
stake-pool: Add usage docs

joncinque:
#### Problem

People need to know how to use stake pools!

#### Solution

Add a guide to the different instructions and examples of how to use them in the CLI.  The program is currently deployed on testnet at `poo1B9L9nR3CrcaziKVYVpRX6A9Y1LAXYasjjfCbApj`, so people can start playing with stake pools.

### DIFF
--- a/docs/src/stake-pool.md
+++ b/docs/src/stake-pool.md
@@ -1,11 +1,244 @@
 ---
-title: Stake-Pool Program
+title: Stake Pool Program
 ---
 
-A program for pooling together SOL to be staked by an off-chain agent running SoM (Stake-o-Matic).
+A program for pooling together SOL to be staked by an off-chain agent running
+SoM (Stake-o-Matic) which redistributes the stakes across the network and tries
+to maximize censorship resistance and rewards.
 
-Each SoM needs at least one pool.  Users deposit stakes into the SoM pool
-and receives a pool token minus the fee.  The SoM redistributes the stakes
-across the network and tries to maximize censorship resistance and rewards.
+## Overview
 
-The project is under construction.
+Information regarding staking and stake programming is available at:
+
+- https://solana.com/staking
+- https://docs.solana.com/staking/stake-programming
+
+
+## Background
+
+Solana's programming model and the definitions of the Solana terms used in this
+document are available at:
+
+- https://docs.solana.com/apps
+- https://docs.solana.com/terminology
+
+## Source
+
+The Stake Pool Program's source is available on
+[github](https://github.com/solana-labs/solana-program-library).
+
+## Operational overview
+
+The following explains the instructions available in the Stake Pool Program along
+with examples using the command-line utility.
+
+## Very important note concerning activated stakes
+
+In its current iteration, the stake pool only processes fully activated stakes.
+Deposits must come from fully active stakes, and withdrawals return a fully 
+active stake account.
+
+This feature maintains fungibility of stake pool tokens. Fully activated stakes
+are not equivalent to inactive, activating, or deactivating stakes due to the
+time cost of staking. Otherwise, malicious actors can deposit stake in one state
+and withdraw it in another state without waiting.
+
+Future iterations of the stake pool program aim to improve on this limitation.
+
+## Command-line Utility
+
+The `spl-stake-pool` command-line utility can be used to experiment with SPL
+tokens.  Once you have [Rust installed](https://rustup.rs/), run:
+```sh
+$ cargo install spl-stake-pool-cli
+```
+
+Run `spl-stake-pool --help` for a full description of available commands.
+
+### Configuration
+
+The `spl-stake-pool` configuration is shared with the `solana` command-line tool.
+
+#### Current Configuration
+
+```
+solana config get
+```
+
+```
+Config File: ${HOME}/.config/solana/cli/config.yml
+RPC URL: https://api.mainnet-beta.solana.com
+WebSocket URL: wss://api.mainnet-beta.solana.com/ (computed)
+Keypair Path: ${HOME}/.config/solana/id.json
+```
+
+#### Cluster RPC URL
+
+See [Solana clusters](https://docs.solana.com/clusters) for cluster-specific RPC URLs
+```
+solana config set --url https://devnet.solana.com
+```
+
+#### Default Keypair
+
+See [Keypair conventions](https://docs.solana.com/cli/conventions#keypair-conventions)
+for information on how to setup a keypair if you don't already have one.
+
+Keypair File
+```
+solana config set --keypair ${HOME}/new-keypair.json
+```
+
+Hardware Wallet URL (See [URL spec](https://docs.solana.com/wallet-guide/hardware-wallets#specify-a-keypair-url))
+```
+solana config set --keypair usb://ledger/
+```
+
+### Example: Creating your own stake pool with 3% fee
+
+```sh
+$ spl-stake-pool create-pool --fee-numerator 3 --fee-denominator 100
+Creating mint Gmk71cM7j2RMorRsQrsyysM4HsByQx5PuDGtDdqGLWCS
+Creating pool fee collection account 3xvXPfQi2SaTkqPV9A7BQwh4GyTe2ZPasfoaCBCnTAJ5
+Creating stake pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC
+Signature: 5HdDoPssqwyLjt2QvhRbnSATZqFLGKha92zMuJiBUpKeKYKGURRV41N5ydCQxqnFjCud3xv85Z6ghErppNJzaYM8
+```
+
+The unique stake pool identifier is `3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC`.
+
+The identifier for the SPL token for staking derivatives is
+`Gmk71cM7j2RMorRsQrsyysM4HsByQx5PuDGtDdqGLWCS`. The stake pool has full control
+over the mint.
+
+The pool creator's fee account identifier is
+`3xvXPfQi2SaTkqPV9A7BQwh4GyTe2ZPasfoaCBCnTAJ5`. When users deposit SOL into the
+stake pool, the program will transfer 3% of their contribution into this account.
+
+### Example: Create validator stake account
+
+In order to accommodate large numbers of user deposits into the stake pool, the
+stake pool only manages one stake account per validator. To add a new validator
+to the stake pool, we first create a validator-associated stake account.
+
+According to [validators.app](https://www.validators.app/), the top validator is
+identity `8SQEcP4FaYQySktNQeyxF3w8pvArx3oMEh7fPrzkN9pu` on vote account 
+`2HUKQz7W2nXZSwrdX5RkfS2rLU4j1QZLjdGCHcoUKFh3` so let's create a stake account
+delegated to that vote account.
+
+```sh
+$ spl-stake-pool create-validator-stake --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC --validator 2HUKQz7W2nXZSwrdX5RkfS2rLU4j1QZLjdGCHcoUKFh3
+Creating stake account FYQB64aEzSmECvnG8RVvdAXBxRnzrLvcA3R22aGH2hUN
+Signature: 4pA2WKT6d2wkXEtSpiQswv22WyoFad2KX6FdPEzwBiEquvaUBEtzenys5Jh1ABPCh7yc4w8kzqMRRCwDj6ZSUV1K
+```
+
+In order to maximize censorship resistance, we want to distribute our SOL to as
+many validators as possible, so let's add a few more.
+
+```sh
+$ spl-stake-pool create-validator-stake --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC --validator HJiC8iJ4Sj846SswQuauFJK93UvV6zp3c2T6jzGqzhhz
+Creating stake account E5KBATUd21Dnjnh5sGFw5ngp9kdVXCcAAYMRe2WsVXie
+Signature: 4pyRZzjsWG7jP3GRZeZCo2Eb2TPjHM4kAYRFMivimme6HAee1nhzoNJBe3VSt2sv7acp5fwT7J8omBM8o3niY8gu
+$ spl-stake-pool create-validator-stake --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC --validator AUCzCaGAGjL3uyjFBtJs7KuJcgQWvNZu1Z2S9G3pw77G
+Creating stake account CrStLEWfme37kDc3nubK9HsmWR5dsuVUuqEKqTR4Mc5E
+Signature: 4ZUdZzUARgUCPuY8nVsJbN6vRDbVX8sYAQGYYXj2YVvjoJ2oevq2H8uzrhYApe419uoP7QYukqNstiti5p5DDukN
+$ spl-stake-pool create-validator-stake --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC --validator 8r1f8mwrUiYdg2Rx9sxTh4M3UAUcCBBrmRA3nxk3Z6Lm
+Creating stake account FhFft7ArhZZkh6q4ir1JZMYFgXdH6wkT5M5nmDDb1Q13
+Signature: yQqXCbuA66wQsHtkziNg3XadfZF5aCmvjfentwbZJnSPeEjJwPka3M1QY5GmR1efprptqaePn71BTMSLscX8DLr
+```
+
+NOTE: These stake accounts have not been added to the stake pool yet. Stake pools
+only accept deposits from fully delegated (warmed-up) stake accounts, so we must
+first delegate these stakes.
+
+We can see the status of stake account using the Solana command-line utility.
+
+```sh
+$ solana stake-account FYQB64aEzSmECvnG8RVvdAXBxRnzrLvcA3R22aGH2hUN
+Balance: 0.002282881 SOL
+Rent Exempt Reserve: 0.00228288 SOL
+Stake account is undelegated
+Stake Authority: 4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn
+Withdraw Authority: 4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn
+```
+
+The stake pool creates these special staking accounts with 1 lamport as a
+minimum delegation.  We must delegate them ourselves to the vote account specified
+on the creation.
+
+```sh
+$ solana delegate-stake FYQB64aEzSmECvnG8RVvdAXBxRnzrLvcA3R22aGH2hUN 2HUKQz7W2nXZSwrdX5RkfS2rLU4j1QZLjdGCHcoUKFh3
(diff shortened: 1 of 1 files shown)


docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+```
+
+The unique stake pool identifier is `3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC`.
+
+The identifier for the SPL token for staking derivatives is
+`Gmk71cM7j2RMorRsQrsyysM4HsByQx5PuDGtDdqGLWCS`. The stake pool has full control
+over the mint.
+
+The pool creator's fee account identifier is
+`3xvXPfQi2SaTkqPV9A7BQwh4GyTe2ZPasfoaCBCnTAJ5`. When users deposit SOL into the
```
mvines, line 124: >  When users deposit SOL into the stake pool ...

I found this sentence a little confusing as the user deposits warmed-up stake accounts into the pool, not just SOL

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+stake pool, the program will transfer 3% of their contribution into this account.
+
+### Example: Create validator stake account
```
mvines, line 127: This is an "Administrator Example" right?

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+* size of the overall transaction, limited to roughly 1 MTU / packet
+* computation budget per instruction
+
+A stake pool may manage hundreds of staking accounts, so it is impossible to
+update the total value of the stake pool in one instruction. Thankfully, the
+command-line utility does all of the work of breaking up transactions.
+
+```sh
+$ spl-stake-pool update --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC 
+Signature: 3Yx1RH3Afqj5ckX8YvPCRt1DudVP4HuRPkh1dBPvTM9GqGxcB9ZXHGZPADVSZiaqKi166fevMG232EWxrRWswPtt
```
mvines, line 309: This confused me a little, as the section talks about hundreds of transactions but only one signature is shown here.

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+the total value managed by the stake pool every epoch.
+
+The Solana transaction processor has two important limitations:
+
+* size of the overall transaction, limited to roughly 1 MTU / packet
+* computation budget per instruction
+
+A stake pool may manage hundreds of staking accounts, so it is impossible to
+update the total value of the stake pool in one instruction. Thankfully, the
+command-line utility does all of the work of breaking up transactions.
```
mvines, line 305: I feel like this detail belongs in a design section rather than the examples section, where the user just wants to know what to do and learn what the expected output should be.

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+```
+
+If another user already updated the stake pool balance for the current epoch, we
+see a different output.
+
+```sh
+$ spl-stake-pool update --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC 
+Stake pool balances are up to date, no update required.
+```
+
```
mvines, line 319: Maybe a quick note as to what happens if `spl-stake-pool update` is not run for an epoch

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+### Example: Withdraw stake
+
+Whenever we want to recover SOL plus accrued rewards, we can provide our
```
mvines, line 322:
```suggestion
Whenever the user wants to recover SOL plus accrued rewards, they can provide their
```

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+staking derivative SPL tokens in exchange for an activated stake account.
+
+Let's withdraw 0.02 of our staking derivative tokens from the stake pool.
```
mvines, line 325:
```suggestion
Let's withdraw 0.02 of staking derivative tokens from the stake pool.
```

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+Rent Exempt Reserve: 0.00228288 SOL
+Delegated Stake: 8.867176377 SOL
+Active Stake: 8.867176377 SOL
+Delegated Vote Account Address: 2HUKQz7W2nXZSwrdX5RkfS2rLU4j1QZLjdGCHcoUKFh3
+Stake Authority: 4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn
+Withdraw Authority: 4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn
+```
+
+Note: due to truncation on SPL token to SOL conversion, we will not receive the
+exact amount calculated by the command-line utility.
```
mvines, line 350: Can the cli be changed to show the exact amount received?   This seems like an annoyance the user shouldn't need to have to deal with, and could breed mistrust

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+
+We can leave this stake account as it is, or we can deactivate it as a normal
+stake account.
+
+```sh
+$ solana deactivate-stake CZF2z3JJoDmJRcVjtsrz1BKUUGNL3VPW5FPFqge1bzmQ
+Signature: 816LjrZFPWcdvkeFTBezFxptMoNjUEyVmHQ6bpX7sdZZbzhWKchUyx2MC5VHdXFP7U41B8S7QQfb7hTqDASCqZg
+```
+
+Once the stake is no longer active, we can use it as normal SOL.
```
mvines, line 360: nit: seems unnecessary as now the user is back to the normal staking flow
```suggestion
```

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+
+### Example: Withdraw stake
+
+Whenever we want to recover SOL plus accrued rewards, we can provide our
+staking derivative SPL tokens in exchange for an activated stake account.
+
+Let's withdraw 0.02 of our staking derivative tokens from the stake pool.
+
+```sh
+$ spl-stake-pool withdraw --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC  --amount 0.02 --burn-from 34XMHa3JUPv46ftU4dGHvemZ9oKVjnciRePYMcX3rjEF
```
mvines, line 328: The `--burn-from` argument name feels unsettling.  This is "my" stake pool SPL Token account right?  `--withdraw-from`?

> Creating account to receive stake CZF2z3JJoDmJRcVjtsrz1BKUUGNL3VPW5FPFqge1bzmQ

Can the recipient stake address be passed in?  Having to parse the cli output to determine the address will be hard to use from a script

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+Signature: 2cDjHXSHjuadGQf1NQpPi43A8R19aCifsY16yTcictKPHcSAXN5TvXZ58nDJwkYs12tuZfTh5WVgAMSvptfrKdPP
+```
+
+Two epochs later, when the stake is fully active and has received one epoch of
+rewards, we can deposit the stake into the stake pool.
+
+```sh
+$ spl-stake-pool deposit --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC --stake 4F4AYKZbNtDnu7uQey2Vkz9VgkVtLE6XWLezYjc9yxZa
+Depositing into stake account FYQB64aEzSmECvnG8RVvdAXBxRnzrLvcA3R22aGH2hUN
+Creating account to receive tokens 34XMHa3JUPv46ftU4dGHvemZ9oKVjnciRePYMcX3rjEF
```
mvines, line 278: > Creating account to receive tokens 34XMHa3JUPv46ftU4dGHvemZ9oKVjnciRePYMcX3rjEF

Can the user optionally pass in the token address?  Having to parse the cli output to determine the address will be hard to use from a script

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+Rent Exempt Reserve: 0.00228288 SOL
+Delegated Stake: 1.000000001 SOL
+Active Stake: 1.000000001 SOL
+Delegated Vote Account Address: AUCzCaGAGjL3uyjFBtJs7KuJcgQWvNZu1Z2S9G3pw77G
+Stake Authority: 4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn
+Withdraw Authority: 4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn
+```
+
+The SPL token account has been debited to accommodate the removal of SOL from
+the stake pool.
```
mvines, line 391: Maybe this section could be expanded on.  It's not immediately clear to me how the system prevents a malicious administrator from running `spl-stake-pool remove-validator-stake` on all the stake pools and running away with the SOL

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+stake pool manager cannot steal funds from the stake pool.
+
+### Example: (Administrator only) Set owner
+
+The stake pool owner may pass their administrator privileges to another account.
+
+```sh
+$ spl-stake-pool --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC --new-owner 4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn
+Signature: 39N5gkaqXuWm6JPEUWfenKXeG4nSa71p7iHb9zurvdZcsWmbjdmSXwLVYfhAVHWucTY77sJ8SkUNpVpVAhe4eZ53
+```
```
mvines, line 427: Aligning this interface with the existing `solana stake-authorize` command would be good.  A single `spl-stake-pool authorize ...` that takes:
* `--new-staker` -> `--new-stake-authority <PUBKEY>`
* `--new-owner` -> `--new-withdraw-authority <PUBKEY>`

mvines (commented):
Great start!

mvines:
Also how about an example for how the staking authority for a pool would be expected to move the pool stake between the different validators?

brianlong:
This is super-helpful! Thanks for working on it.

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+and withdraw it in another state without waiting.
+
+Future iterations of the stake pool program aim to improve on this limitation.
+
+## Command-line Utility
+
+The `spl-stake-pool` command-line utility can be used to experiment with SPL
+tokens.  Once you have [Rust installed](https://rustup.rs/), run:
+```sh
+$ cargo install spl-stake-pool-cli
```
brianlong, line 66: Missing pre-requisites for this cargo command. In my case, I see this:
```
% cargo install spl-stake-pool-cli
    Updating crates.io index
error: could not find `spl-stake-pool-cli` in registry `https://github.com/rust-lang/crates.io-index` with version `*`
```

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+
+We can see the status of stake account using the Solana command-line utility.
+
+```sh
+$ solana stake-account FYQB64aEzSmECvnG8RVvdAXBxRnzrLvcA3R22aGH2hUN
+Balance: 1.002282880 SOL
+Rent Exempt Reserve: 0.00228288 SOL
+Stake account is undelegated
+Stake Authority: 4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn
+Withdraw Authority: 4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn
```
brianlong, line 172: This is the first mention of `4SnSuUtJGKvk2GYpBwmEsWG53zTurVM8yXGsoiZQyMJn` and they seem to have high-authority. Who holds this key?

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+
+Two epochs later, when the stake is fully active and has received one epoch of
+rewards, we can deposit the stake into the stake pool.
+
+```sh
+$ spl-stake-pool deposit --pool 3CLwo9CntMi4D1enHEFBe3pRJQzGJBCAYe66xFuEbmhC --stake 4F4AYKZbNtDnu7uQey2Vkz9VgkVtLE6XWLezYjc9yxZa
+Depositing into stake account FYQB64aEzSmECvnG8RVvdAXBxRnzrLvcA3R22aGH2hUN
+Creating account to receive tokens 34XMHa3JUPv46ftU4dGHvemZ9oKVjnciRePYMcX3rjEF
+Signature: 4AESGZzqBVfj5xQnMiPWAwzJnAtQDRFK1Ha6jqKKTs46Zm5fw3LqgU1mRAT6CKTywVfFMHZCLm1hcQNScSMwVvjQ
+```
```
brianlong, line 280: How does the stake pool re-allocate the stake from this account? Does the stake-pool have the authority to remove SOL from the given stake account & redistribute it to the stake accounts for other validators? It will be helpful to see some before & after examples for stake account 4F4AYKZbNtDnu7uQey2Vkz9VgkVtLE6XWLezYjc9yxZa.

brianlong:
Hi @joncinque, Thanks again for working on this. My preconceived notion about the operation of a stake pool was 100% wrong. I think you will need a very basic, high-level, overview of the process at the top of the document. The actual process is pretty complicated and I think that many readers will struggle with the concepts.

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+
+The Stake Pool Program's source is available on
+[github](https://github.com/solana-labs/solana-program-library).
+
+## Operational overview
+
+The following explains the instructions available in the Stake Pool Program along
+with examples using the command-line utility.
+
+## Very important note concerning activated stakes
```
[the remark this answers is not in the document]
joncinque: Perhaps that can go in a high-level overview portion as recommended by @brianlong -- I would say it's more important than an Appendix, and a fundamental concept of the program

docs/src/stake-pool.md
```diff
@@ -1,11 +1,427 @@
...
+```sh
+solana config set --keypair ${HOME}/new-keypair.json
+```
+
+Hardware Wallet URL (See [URL spec](https://docs.solana.com/wallet-guide/hardware-wallets#specify-a-keypair-url))
+```sh
+solana config set --keypair usb://ledger/
+```
+
+### Example: Creating your own stake pool with 3% fee
```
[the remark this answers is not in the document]
joncinque: Yes, and I'll clarify that the fees go to the owner in the form of pool tokens

docs/src/stake-pool.md
joncinque: This is currently a permissionless instruction since it isn't changing the stake pool in any way, but it makes more sense to have it be permissioned.  Added #1419 for that

docs/src/stake-pool.md
joncinque: The CLI manages to get it all done in one transaction currently, and thus one signature, but we'll need to expand it to potentially create multiple transactions.  Added #1420 for that

docs/src/stake-pool.md
joncinque: I did some refactoring and managed to fix this :-)

docs/src/stake-pool.md
joncinque: Will change the param.

And the stake address can be passed in, but it's bugged currently since it always splits.  Added #1421 for this

docs/src/stake-pool.md
joncinque: Yep, you can use the `token-receiver` param.  I'll add an example for that

docs/src/stake-pool.md
joncinque: They have to go through the same mechanism as outside users and provide pool tokens in exchange for staked SOL.  Admins can add / remove validator accounts totally though.  I'll clarify that

docs/src/stake-pool.md
joncinque: First one sounds good, but the second one doesn't quite reflect the situation, since it's setting the stake pool manager, who doesn't have withdraw authority over the account.  How about `--new-owner` -> `new-administrator`?

docs/src/stake-pool.md
joncinque: Sorry about that, the crate hasn't been published yet, so you'll have to build / run things yourself.  Is that ok?

docs/src/stake-pool.md
joncinque: I'll clarify this -- this key is the stake pool manager, which defaults to whatever is in your Solana CLI `config.yml`.  Since I was testing with one of my keys, it was chosen.

docs/src/stake-pool.md
joncinque: There was another question about how to rebalance, so I'll add an example for how to do that.  The concept is that the stake pool manager will have to run their own scripts in order to do that part.

joncinque:
@brianlong thanks for taking a look at this, I appreciate the feedback.  It's definitely a complicated system, so I aimed to write the document as my own step-by-step process for setting up the pool.  I'm happy to add in some high-level information about the stake pool operation.  Are you suggesting an outline of all the steps for administrators to setup a pool, and for users to interact with it?

joncinque:
@mvines I'll add in the example of moving stake around.  I think we need to add some more bookkeeping logic to allow the stake pool owner to manage the pool more directly.  Currently, the best way for them would be to use the same `deposit` and `withdraw` operations that outside users use.

The `set-staking-authority` command can cause some problems since the manager might merge / split staking accounts, without the stake pool knowing.  I think this can be easily fixed if we just have the update commands accept any stake account that the pool owns, regardless if it's a proper "validator stake account".  This breaks the "one stake account per validator" model, but gives a lot more flexibility for stake pool scripts.  what do you think?

docs/src/stake-pool.md
brianlong: No worries for me, I'm sure I can figure it out. A non-technical user will see this as a road-block. A quick mention of pre--reqs near the top might help the reader know what to expect.

docs/src/stake-pool.md
brianlong: Hopefully, the manager can split the active stake accounts and then delegate those to different validators. I don't think that a deactivation/activation cycle is required (or desired). This will be an interesting area for a skilled manager to maximize uptime/reward from the pool.

brianlong:
@joncinque I think my initial misunderstanding was due to context & perspective. There are two different users to consider: 1) Defi investors who want pool tokens to use in derivative financial products. Those users will spend the time required to learn & participate; and 2) Users who think that that "Defi" means "Stick it to 'The Man'" and don't participate in derivative markets.

I was reading this from point-of-view #2. I thought that I would simply send SOL to an on-chain program and get a shiny new pool token in exchange. It's actually more complicated than that. A short paragraph at the top might help some people to understand the context.

Great work so far! I've already learned a bunch about how pools work.

P.S. User group #2 doesn't need pools to in order to spread stake around -- a simple stake manager will meet their needs. Don't feel like you need to write for audience #2 since the real audience is group #1.

docs/src/stake-pool.md
joncinque: There will have to be some deactivating / reactivating unfortunately because of how the staking process works in general.  The big question is more: "how much liquidity does a pool manager need to have available to manage the pool?" And in its current form, they might need quite a bit.

joncinque:
@brianlong That's a great way of putting it.  I'll add some of the context about how stake pools are meant to fit in the ecosystem then.

The initial concept was for stake pools to work that easily, but we have the reality that staked SOL is not fungible with SOL, and in fact staked SOL is only fungible with other staked SOL on the exact same validator.

Thanks for the feedback!

mvines (approved):

PR #1347, opened 2021-03-01
Binary oracle implementation

VadimGrozinok:


binary-oracle-pair/program/src/processor.rs
```diff
@@ -0,0 +1,523 @@
...
+        let account_info_iter = &mut accounts.iter();
+        let pool_account_info = next_account_info(account_info_iter)?;
+        let authority_account_info = next_account_info(account_info_iter)?;
+        let user_token_account_info = next_account_info(account_info_iter)?;
+        let pool_deposit_token_account_info = next_account_info(account_info_iter)?;
+        let token_pass_mint_info = next_account_info(account_info_iter)?;
+        let token_fail_mint_info = next_account_info(account_info_iter)?;
+        let token_pass_destination_account_info = next_account_info(account_info_iter)?;
+        let token_fail_destination_account_info = next_account_info(account_info_iter)?;
+        let token_program_id_info = next_account_info(account_info_iter)?;
```
joncinque, line 278: This should add the check to make sure we're not past the mint end

binary-oracle-pair/program/src/processor.rs
```diff
@@ -0,0 +1,523 @@
...
+                    authority_account_info.clone(),
+                    token_program_id_info.clone(),
+                    pool_deposit_token_account_info.clone(),
+                    amount,
+                )?;
+            }
+        } else {
+            let current_slot = clock.slot;
+            if current_slot < pool.mint_end_slot || current_slot > pool.decide_end_slot {
+                let possible_withdraw_amount = cmp::min(
```
joncinque, line 409: This should probably be `amount.min(user_pass_token_account.amount).min(user_fail_token_account.amount)`, otherwise we always take all of their tokens, which will probably fail if someone only approves `amount` to be taken, and `amount < possible_withdraw_amount`

binary-oracle-pair/program/tests/tests.rs
```diff
@@ -0,0 +1,675 @@
...
+    pub pool_deposit_account: Keypair,
+    pub token_pass_mint: Keypair,
+    pub token_fail_mint: Keypair,
+    pub decider: Keypair,
+    pub mint_end_slot: u64,
+    pub decide_end_slot: u64,
+}
+
+impl TestPool {
+    pub fn create() -> Self {
```
joncinque, line 130: nit: by convention, this is typically called `new`

binary-oracle-pair/program/tests/tests.rs
```diff
@@ -0,0 +1,675 @@
...
+            &pool.pool_account.pubkey(),
+            &pool.authority,
+            &pool.pool_deposit_account.pubkey(),
+            &user_pass_account.pubkey(),
+            &user_fail_account.pubkey(),
+            &pool.token_pass_mint.pubkey(),
+            &pool.token_fail_mint.pubkey(),
+            &user_account.pubkey(),
+            &spl_token::id(),
+            deposit_amount,
```
joncinque, line 588: This test can be modified to withdraw some amount less than `deposit_amount` to test the overdraw issue pointed out before

binary-oracle-pair/program/tests/tests.rs
```diff
@@ -0,0 +1,675 @@
...
+        .unwrap();
+
+    let pool_data_before = state::Pool::unpack(pool_account_data_before.data.as_slice()).unwrap();
+
+    assert!(pool_data_before.decision.is_none());
+
+    let decision = true;
+
+    program_context
+        .warp_to_slot(pool.mint_end_slot + 1)
```
joncinque, line 641: :tada: it's useful!

binary-oracle-pair/program/tests/tests.rs
```diff
@@ -0,0 +1,675 @@
...
+        .banks_client
+        .get_account(pool.pool_account.pubkey())
+        .await
+        .unwrap()
+        .unwrap();
+
+    let pool_data_after = state::Pool::unpack(pool_account_data_after.data.as_slice()).unwrap();
+
+    assert!(pool_data_after.decision.unwrap());
+}
```
joncinque, line 675: This should include a positive test for withdrawing after a PASS and FAIL decision to make sure that we can get tokens back.

We can later add the negative tests for trying to deposit after `mint_end` and trying to withdraw during the decide time

joncinque (commented):
Really great work!  Most of the comments are pretty minor

binary-oracle-pair/program/src/instruction.rs
```diff
@@ -0,0 +1,213 @@
...
+    ///
+    ///   0. `[]` Pool
+    ///   1. `[]` Authority
+    ///   2. `[w]` Token SOURCE Account, amount is transferable by pool authority with allowances.
+    ///   3. `[w]` Deposit token account
+    ///   4. `[w]` token_P PASS mint
+    ///   5. `[w]` token_F FAIL mint
+    ///   6. `[w]` token_P DESTINATION Account
+    ///   7. `[w]` token_F DESTINATION Account
+    ///   8. `[]` Token program id
```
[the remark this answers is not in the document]
VadimGrozinok: Sure

binary-oracle-pair/program/src/instruction.rs
```diff
@@ -0,0 +1,213 @@
...
+    ///   6. `[w]` Token Fail mint. Should not be initialized
+    ///   7. `[]` Rent sysvar
+    ///   8. `[]` Token program id
+    InitPool(InitArgs),
+
+    ///   Deposit in the pool.
+    ///
+    ///   0. `[]` Pool
+    ///   1. `[]` Authority
+    ///   2. `[w]` Token SOURCE Account, amount is transferable by pool authority with allowances.
```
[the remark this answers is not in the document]
VadimGrozinok: How will it work then ?
To transfer tokens from account transaction should be signed with this authority
And when we call transfer instruction inside the program we can sign it only with program account
Or if we sign `Deposit` instruction with authority key we can call just `invoke()` inside the program to transfer tokens ?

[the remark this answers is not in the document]
joncinque: > Or if we sign Deposit instruction with authority key we can call just invoke() inside the program to transfer tokens ?

Exactly.  In this particular usecase, it should be ok, since the UI will approve exactly the amount before calling deposit or withdraw.  Either way, the best practice is to approve to an ephemeral keypair, and then sign with that.

binary-oracle-pair/program/src/instruction.rs
```diff
@@ -0,0 +1,211 @@
...
+
+/// Create `Decide` instruction
+pub fn decide(
+    program_id: &Pubkey,
+    pool: &Pubkey,
+    decider: &Pubkey,
+    decision: bool,
+) -> Result<Instruction, ProgramError> {
+    let init_data = PoolInstruction::Decide(decision);
+    let data = init_data.try_to_vec()?;
```
dzmitry-lahoda, line 200: your know the size. known size struct(it may be even on stack). unwrap borsch serde into it. there seems not reason to allocated arbitraty heap here (and hence no error code).

binary-oracle-pair/program/src/state.rs
```diff
@@ -0,0 +1,93 @@
...
+    Undecided,
+    /// Decision set at Pass
+    Pass,
+    /// Decision set at Fail
+    Fail,
+}
+
+impl Pool {
+    /// Length serialized data
+    pub const LEN: usize = size_of::<Pool>() - 13;
```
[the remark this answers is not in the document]
dzmitry-lahoda: also it may be better just plug the number (no x - 13, but whatever value N has)

binary-oracle-pair/program/src/state.rs
```diff
@@ -0,0 +1,93 @@
...
+use solana_program::pubkey::Pubkey;
+use std::mem::size_of;
+
+/// Uninitialized version value, all instances are at least version 1
+pub const UNINITIALIZED_VERSION: u8 = 0;
+/// Initialized pool version
+pub const POOL_VERSION: u8 = 1;
+
+/// Program states.
+#[repr(C)]
```
dzmitry-lahoda, line 13: consider addin here https://github.com/solana-labs/solana/blob/193e37018650a0bb4d56246dee6e2892204289bf/frozen-abi/src/abi_digester.rs#L537 .

so each time it is changed, the type system fails and must be fixed (ABI test)

binary-oracle-pair/program/src/state.rs
```diff
@@ -0,0 +1,93 @@
...
+            mint_end_slot: 433,
+            decide_end_slot: 5546,
+            decision: Decision::Fail,
+        };
+
+        let packed = p.try_to_vec().unwrap();
+
+        let unpacked = Pool::try_from_slice(packed.as_slice()).unwrap();
+
+        assert_eq!(p, unpacked);
```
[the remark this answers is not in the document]
dzmitry-lahoda: or use https://github.com/solana-labs/solana/blob/193e37018650a0bb4d56246dee6e2892204289bf/frozen-abi/src/abi_digester.rs#L537

binary-oracle-pair/program/src/instruction.rs
```diff
@@ -66,8 +68,8 @@ pub enum PoolInstruction {
     ///   5. `[w]` token_P PASS mint
```
joncinque, line 68: I didn't mention this explicitly, but this should have a user transfer authority as well, same as for deposit

binary-oracle-pair/program/src/instruction.rs
```diff
@@ -120,6 +120,8 @@ pub fn deposit(
     program_id: &Pubkey,
     pool: &Pubkey,
     authority: &Pubkey,
+    user_transfer_authority: &Pubkey,
+    is_user_authority_signer: bool,
```
joncinque, line 124: Let's go with the concept that the user transfer authority will always be signing, so this param won't be needed

binary-oracle-pair/program/src/processor.rs
```diff
@@ -349,67 +396,64 @@ impl Processor {
...
-                    &pool_account_info.key,
-                    pool.bump_seed,
-                    user_token_destination_account_info.clone(),
-                    authority_account_info.clone(),
-                    token_program_id_info.clone(),
-                    pool_deposit_token_account_info.clone(),
-                    amount,
-                )?;
-            }
+        if pool.decision == Decision::Pass {
```
joncinque, line 407: It's more idiomatic to do a `match` statement since all cases are covered

binary-oracle-pair/program/src/processor.rs
```diff
@@ -255,9 +281,12 @@ impl Processor {
         pool.decider = *decider_info.key;
         pool.mint_end_slot = mint_end_slot;
         pool.decide_end_slot = decide_end_slot;
-        pool.decision = None;
+        pool.decision = Decision::Undecided;
+
+        pool.serialize(&mut *pool_account_info.data.borrow_mut())
+            .unwrap();
```
joncinque, line 287: nit: since the Borsh error can become a ProgramError, you can actually just do, `pool.serialize(&mut *pool_account_info.data.borrow_mut())` as the last line!

binary-oracle-pair/program/src/state.rs
```diff
@@ -44,59 +41,27 @@ pub struct Pool {
...
-                None
-            } else if input[179] == 1 {
-                Some(true)
-            } else {
-                Some(false)
-            },
-        })
+impl Pool {
+    /// Length serialized data
+    pub const LEN: usize = size_of::<Pool>() - 13;
```
joncinque, line 60: It's not usable for static arrays unfortunately, but you can use `https://github.com/solana-labs/solana/blob/master/sdk/program/src/borsh.rs#L52` in a lot of places.  Also, this seems prone to breaking -- could you add the actual number directly here?

joncinque (commented):
It's really close, just a couple more points!

binary-oracle-pair/program/src/instruction.rs
VadimGrozinok: If user will choose program account as an authority how can we sign transaction in such case ?
As I see we can’t create Keypair from the program account to produce signature

ysavchenko: I think no signature will be required in this case, right? @joncinque

joncinque: The whole idea is that using the same program account to control the program and also as a signer to transfer your funds is a potentially unsafe pattern, because anyone can get the program's signature if they call into the program.  This isn't particularly an issue for this program, but in general it's better to use ephemeral keypairs as a transfer authority

binary-oracle-pair/program/src/processor.rs
VadimGrozinok: When I'm trying to do this I have such an error
```
expected enum std::result::Result<_, solana_program::program_error::ProgramError>
         found enum std::result::Result<_, std::io::Error>
```

joncinque: ah right, you still need to map the error, like `pool.serialize(...).map_err(|e| e.into())`

joncinque:
Can you rebase and try again? There was a fix to `rm -rf target/debug` during CI to avoid filling up the drive.  Otherwise, this is good to go from my end

joncinque:
Thanks for all of your work!

PR #1353, opened 2021-03-02
spl-token-cli: Use forward-compatible apis

CriesofCarrots:
Upstream solana is slated to deprecate the TokenAmount::ui_amount `f64`. In the short term, this field will become an `Option<f64>`, which breaks several spots in the spl-token cli that call this field directly.

Replace these usages with forward-compatible ones to prepare for the change.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -4,7 +4,7 @@ use clap::{
 };
 use console::Emoji;
 use solana_account_decoder::{
-    parse_token::{TokenAccountType, UiAccountState, UiTokenAmount},
+    parse_token::{TokenAccountType, UiAccountState},
     UiAccountData,
 };
 use solana_clap_utils::{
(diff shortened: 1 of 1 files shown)


CriesofCarrots:
@mvines , could I get your feedback on this approach?
1. Merge this
2. Bump spl-token-cli version and publish to crates.io
3. Move forward with https://github.com/solana-labs/solana/pull/15616, for which downstream-projects should then build.

There are a couple places in this crate that don't tolerate large Mint decimals well. Also, these token/program methods should probably be fixed to use checked_pow: https://github.com/solana-labs/solana-program-library/blob/d536460d3fd465d26ed298843a046acb64063edf/token/program/src/lib.rs#L20-L27
But I'm thinking that work can follow the solana updates.

mvines (approved):
Your approach sgtm

PR #1362, opened 2021-03-02
Update solana-program to 1.5.11

joncinque:
#### Problem

#1226 is blocked because the sdk update breaks the compute limits on a memo test.

#### Solution

Separate PR to update to 1.5.11 to update broken tests and unblock #1226


CriesofCarrots (approved):
Thank you!

PR #1368, opened 2021-03-02
token-swap: Document owner address in README

joncinque:
#### Problem

People have a hard time discovering the production parameters of token-swap, specifically who is supposed to be the fee owner address.

#### Solution

Include that info in the README along with more info about testing.  This is the documentation side of #1367

### DIFF
--- a/docs/src/token-swap.md
+++ b/docs/src/token-swap.md
@@ -5,6 +5,18 @@ title: Token Swap Program
 A Uniswap-like exchange for the Token program on the Solana blockchain,
 implementing multiple automated market maker (AMM) curves.
 
+Here is some important developer information regarding the program deployed on devnet,
+testnet, and mainnet-beta:
+
+| Information | Account Address |
+| --- | --- |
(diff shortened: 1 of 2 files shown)


docs/src/token-swap.md
```diff
@@ -53,28 +53,28 @@ The following explains the instructions available in the Token Swap Program.
 Note that each instruction has a simple code example that can be found in the
 [end-to-end tests](https://github.com/solana-labs/solana-program-library/blob/master/token-swap/js/cli/token-swap-test.js).
 
-### Creating a new token swap
+### Creating a new token swap pool
```
bartosz-lipinski, line 56: I think it will be easier to find if host fee address in this file in addition to README.md since this file is published to  spl.solana.com. Additionally, it might be worth adding a table at the top with token-swap address itself for easy of discovery.

bartosz-lipinski (commented):
Looks great @joncinque, I only added one small comment.

PR #1374, opened 2021-03-03
Revert "Document about the known fake rent issue (#753)"

ryoqun:
This reverts commit 0dc32f67dfa2d296bf48eb0b63f8869eae0d5d74.


ryoqun:
Well turned out I want to leave some text. So just reverting won't work. ;)

ryoqun:
Well turned out I want to leave some text. So just reverting won't work. ;)

PR #1422, opened 2021-03-09
stake-pool-cli: Fix withdraw calc for CLI / program consistency

joncinque:
#### Problem

The stake pool CLI shows how much we expect to receive on withdrawal in SOL, but that number is different and higher than what we actually receive.  That stinks.

#### Solution

Refactor the calcs to do everything terms of pool tokens.  This avoids unnecessary conversions which create truncations.


stake-pool/cli/src/main.rs
```diff
@@ -797,8 +801,8 @@ fn pick_withdraw_accounts(
     // Not enough stake to withdraw the specified amount
     if remaining_amount > 0 {
         return Err(format!(
-            "No stake accounts found in this pool with enough balance to withdraw {} SOL.",
-            lamports_to_sol(amount)
+            "No stake accounts found in this pool with enough balance to withdraw {} pool tokens.",
+            lamports_to_sol(pool_amount)
```
mvines, line 805: Little weird to use `lamports_to_sol()` on pool tokens here, could `spl_token::amount_to_ui_amount()` be used instead?

stake-pool/cli/src/main.rs
```diff
@@ -875,15 +876,16 @@ fn command_withdraw(
...
+            .calc_lamports_amount(withdraw_stake.pool_amount)
+            .unwrap();
+
         println!(
             "Withdrawing from account {}, amount {} SOL, {} pool tokens",
             withdraw_stake.pubkey,
-            lamports_to_sol(withdraw_stake.amount),
-            lamports_to_sol(withdraw_amount),
+            lamports_to_sol(sol_withdraw_amount),
+            lamports_to_sol(withdraw_stake.pool_amount),
```
mvines, line 888: same comment about using `amount_to_ui_amount()` here as well

stake-pool/cli/src/main.rs
```diff
@@ -875,15 +876,16 @@ fn command_withdraw(
...
+        let sol_withdraw_amount = pool_data
+            .calc_lamports_amount(withdraw_stake.pool_amount)
+            .unwrap();
+
         println!(
             "Withdrawing from account {}, amount {} SOL, {} pool tokens",
             withdraw_stake.pubkey,
-            lamports_to_sol(withdraw_stake.amount),
-            lamports_to_sol(withdraw_amount),
+            lamports_to_sol(sol_withdraw_amount),
```
mvines, line 887: Consider using the [Sol](https://github.com/solana-labs/solana/blob/3355efbb5ac0c8a1be23befa68a91da098abc851/sdk/program/src/native_token.rs#L17) type and it'll automatically add the fancy ◎ prefix
```suggestion
            solana_sdk::native_token::Sol(sol_withdraw_amount),
```

mvines (approved):
lgtm, just nits

PR #1423, opened 2021-03-09
stake-pool-cli: Rename burn -> withdraw

joncinque:
Per feedback on the stake pool docs `burn` is a bit of a scary word for withdrawals, so change the name of the param.


mvines (approved):
Thanks 👻

PR #1424, opened 2021-03-10
WIP: Implementation of stable curve methods

rawfalafel:
Implements `pool_tokens_to_trading_tokens` and `trading_tokens_to_pool_tokens` for the stable curve. 

TODO:
 - [ ] Add unit tests and prop tests to ensure invariant always holds
 - [ ] Analyze instructions to ensure that they're well below compute budget

@joncinque


token-swap/program/src/curve/stable.rs
```diff
@@ -175,16 +173,27 @@ impl CurveCalculator for StableCurve {
...
+        match round_direction {
+            RoundDirection::Floor => Some(TradingTokenResult {
+                token_a_amount: token_a_value.floor()?.to_imprecise()?,
+                token_b_amount: token_b_value.floor()?.to_imprecise()?,
+            }),
+            RoundDirection::Ceiling => Some(TradingTokenResult {
+                token_a_amount: token_a_value.ceiling()?.to_imprecise()?,
+                token_b_amount: token_b_value.ceiling()?.to_imprecise()?,
+            }),
+        }
```
joncinque, line 192: This function needs to be inverse of `trading_tokens_to_pool_tokens`, otherwise you'll create an arbitrage opportunity within your own curve.  As it stands, it doesn't seem like that would be the case, since this is the same as the constant product curve.

token-swap/program/src/curve/stable.rs
```diff
@@ -194,14 +203,36 @@ impl CurveCalculator for StableCurve {
         trade_direction: TradeDirection,
```
joncinque, line 203: `TradeDirection` is not the same thing as `deposit` vs `withdrawal`, and instead tells which token you'll be adding / removing.  The `RoundDirection` kind of tells you if you're doing a deposit or withdrawal, but isn't very clear for that.  You'll want to change `RoundDirection` to give that information instead, which means updating some of the code for the other curves.  You can change it to `LiquidityProviderOperation` with names `Withdraw` and `Deposit`

token-swap/program/src/curve/stable.rs
```diff
@@ -194,14 +203,36 @@ impl CurveCalculator for StableCurve {
...
+        let (new_swap_source_amount, new_swap_destination_amount) = match trade_direction {
+            TradeDirection::AtoB => (swap_token_a_amount + source_amount, swap_token_b_amount),
+            TradeDirection::BtoA => (swap_token_b_amount - source_amount, swap_token_a_amount),
+        };
+        let d1 = compute_d(
+            leverage,
+            new_swap_source_amount,
+            new_swap_destination_amount,
+        )?;
+        let diff = match trade_direction {
```
joncinque, line 221: This should match on round direction too

joncinque (commented):
Good start!

token-swap/program/src/curve/stable.rs
rawfalafel: Hey! My understanding is that this is the inverse of `trading_tokens_to_pool_tokens` for the sable curve, even though its logically the same as the constant product curve. This is true is because `pool_tokens_to_trading_tokens` can be calculated without knowledge of the curve (as long as the ratio between pool tokens to pool tokens supply is equal to trading tokens to trading tokens supply), whereas `trading_tokens_to_pool_tokens` is dependent on the curve. We will write tests cases to verify this.

rawfalafel:
Hi @joncinque , we addressed your comment about `RoundDirection`! Let us know what you think.

token-swap/program/src/curve/stable.rs
joncinque: They should both be dependent on the curve, since the curve defines the value of tokens contained.  You should try implementing this test https://github.com/solana-labs/solana-program-library/blob/master/token-swap/program/src/curve/calculator.rs#L183 for the stable curve and see what happens!

token-swap/program/src/curve/calculator.rs
```diff
@@ -219,7 +229,7 @@ pub mod test {
                 swap_token_b_amount,
                 pool_supply,
                 trade_direction,
-                RoundDirection::Floor,
+                LiquidityProviderOperation::Withdrawal,
```
joncinque, line 232: I'm now remembering why I kept this as `RoundDirection` and not the operation -- it's not always clear which one to use!  For example, if I'm depositing token A and B to get pool tokens, we actually specify the number of pool tokens and then calculate the amount of token A and B necessary to get that.  I call `pool_tokens_to_trading_tokens`, but since it's a deposit, I want to be sure to `ceiling` the amount of A and B that are taken.  When I deposit one side, I specify the amount of token A to deposit and then calculate the amount of pool tokens to receive, which means I want to `floor` that calculation to avoid minting too many pool tokens.  All that to say, these should all be `Deposit`.  

This will be fine for now, but it will be very confusing once we add more flexible options for deposits.  Or we'll just have to add more `LiquidityProviderOperation` options

token-swap/program/src/curve/constant_price.rs
```diff
@@ -125,14 +125,14 @@ impl CurveCalculator for ConstantPriceCurve {
             .checked_mul(token_b_price)?
             .checked_add(U256::from(swap_token_a_amount))?;
         let pool_supply = U256::from(pool_supply);
-        match round_direction {
-            RoundDirection::Floor => Some(
+        match liquidity_provider_operation {
+            LiquidityProviderOperation::Withdrawal => Some(
```
joncinque, line 129: These need to be flipped as mentioned earlier

token-swap/program/src/curve/constant_product.rs
```diff
@@ -118,9 +118,9 @@ pub fn trading_tokens_to_pool_tokens(
     let root = base.sqrt()?.checked_sub(&one)?;
     let pool_supply = PreciseNumber::new(pool_supply)?;
     let pool_tokens = pool_supply.checked_mul(&root)?;
-    match round_direction {
-        RoundDirection::Floor => pool_tokens.floor()?.to_imprecise(),
-        RoundDirection::Ceiling => pool_tokens.ceiling()?.to_imprecise(),
+    match liquidity_provider_operation {
+        LiquidityProviderOperation::Withdrawal => pool_tokens.floor()?.to_imprecise(),
```
joncinque, line 122: These need to be flipped too

token-swap/program/src/curve/stable.rs
```diff
@@ -201,37 +208,27 @@ impl CurveCalculator for StableCurve {
...
+            TradeDirection::BtoA => (swap_token_b_amount, swap_token_a_amount)
         };
         let leverage = self.amp.checked_mul(N_COINS as u64)?;
         let d0 = compute_d(leverage, swap_source_amount, swap_destination_amount)?;
-        let (new_swap_source_amount, new_swap_destination_amount) = match trade_direction {
-            TradeDirection::AtoB => (swap_token_a_amount + source_amount, swap_token_b_amount),
-            TradeDirection::BtoA => (swap_token_b_amount - source_amount, swap_token_a_amount),
+        let (new_swap_source_amount, new_swap_destination_amount) = match liquidity_provider_operation {
+            LiquidityProviderOperation::Deposit => (swap_token_a_amount + source_amount, swap_token_b_amount),
+            LiquidityProviderOperation::Withdrawal => (swap_token_b_amount - source_amount, swap_token_a_amount)
```
joncinque, line 221:
```suggestion
        let new_swap_source_amount = match liquidity_provider_operation {
            LiquidityProviderOperation::Deposit => swap_source_amount + source_amount,
            LiquidityProviderOperation::Withdrawal => swap_source_amount - source_amount,
```

token-swap/program/src/curve/stable.rs
```diff
@@ -201,37 +208,27 @@ impl CurveCalculator for StableCurve {
...
-                Some(PreciseNumber::new(final_amount)?.floor()?.to_imprecise()?)
-            }
-            RoundDirection::Ceiling => Some(
-                PreciseNumber::new(final_amount)?
-                    .ceiling()?
-                    .to_imprecise()?,
-            ),
+        match liquidity_provider_operation {
+            LiquidityProviderOperation::Deposit => Some(PreciseNumber::new(final_amount)?.ceiling()?.to_imprecise()?),
+            LiquidityProviderOperation::Withdrawal => Some(PreciseNumber::new(final_amount)?.floor()?.to_imprecise()?)
```
joncinque, line 231: `floor()` and `ceiling()` should be flipped here

token-swap/program/src/processor.rs
```diff
@@ -433,7 +433,7 @@ impl Processor {
                 swap_token_b_amount,
                 to_u128(pool_mint.supply)?,
                 trade_direction,
-                RoundDirection::Ceiling,
+                LiquidityProviderOperation::Deposit,
```
joncinque, line 436: Should be `Withdraw`

token-swap/program/src/processor.rs
```diff
@@ -784,7 +784,7 @@ impl Processor {
                 to_u128(swap_token_b.amount)?,
                 pool_mint_supply,
                 trade_direction,
-                RoundDirection::Floor,
+                LiquidityProviderOperation::Withdrawal,
```
joncinque, line 787: Should be `Deposit` -- finally something makes sense!

token-swap/program/src/processor.rs
```diff
@@ -918,7 +918,7 @@ impl Processor {
                 swap_token_b_amount,
                 pool_mint_supply,
                 trade_direction,
-                RoundDirection::Ceiling,
+                LiquidityProviderOperation::Deposit,
```
joncinque, line 921: Should be `Withdraw`

token-swap/program/src/processor.rs
```diff
@@ -5333,7 +5333,7 @@ mod tests {
                     swap_token_b.amount.try_into().unwrap(),
                     pool_mint.supply.try_into().unwrap(),
                     TradeDirection::AtoB,
-                    RoundDirection::Ceiling,
+                    LiquidityProviderOperation::Deposit,
```
joncinque, line 5336: `Withdraw`

token-swap/program/src/processor.rs
```diff
@@ -5506,7 +5506,7 @@ mod tests {
                 token_b_amount.try_into().unwrap(),
                 initial_supply.try_into().unwrap(),
                 TradeDirection::AtoB,
-                RoundDirection::Ceiling,
+                LiquidityProviderOperation::Deposit,
```
joncinque, line 5509: `Withdraw`

token-swap/program/src/processor.rs
```diff
@@ -5583,7 +5583,7 @@ mod tests {
                 token_b_amount.try_into().unwrap(),
                 initial_supply.try_into().unwrap(),
                 TradeDirection::BtoA,
-                RoundDirection::Ceiling,
+                LiquidityProviderOperation::Deposit,
```
joncinque, line 5586: `Withdraw`

joncinque (commented):
Mainly some changes for the `trading_tokens_to_pool_tokens` calc

token-swap/program/src/state.rs
```diff
@@ -40,7 +40,7 @@ pub trait SwapState {
 }
 
 /// All versions of SwapState
-#[enum_dispatch(SwapState)]
+#[enum_dispatch(SwapVersion)]
```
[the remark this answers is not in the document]
akamutsu: This is done

token-swap/program/src/processor.rs
akamutsu: done

token-swap/program/src/processor.rs
akamutsu: done

token-swap/program/src/processor.rs
akamutsu: done

token-swap/program/src/processor.rs
akamutsu: done

token-swap/program/src/processor.rs
akamutsu: done

token-swap/program/src/processor.rs
akamutsu: done

token-swap/program/src/curve/stable.rs
akamutsu: done

token-swap/program/src/curve/stable.rs
akamutsu: done

token-swap/program/src/curve/constant_product.rs
akamutsu: done

token-swap/program/src/curve/constant_price.rs
akamutsu: done

token-swap/program/src/curve/calculator.rs
akamutsu: done

akamutsu:
@joncinque could I please have a fresh review since all the new pushes?

token-swap/program/src/curve/stable.rs
```diff
@@ -216,19 +216,19 @@ impl CurveCalculator for StableCurve {
         };
         let leverage = self.amp.checked_mul(N_COINS as u64)?;
         let d0 = compute_d(leverage, swap_source_amount, swap_destination_amount)?;
-        let (new_swap_source_amount, new_swap_destination_amount) = match liquidity_provider_operation {
-            LiquidityProviderOperation::Deposit => (swap_token_a_amount + source_amount, swap_token_b_amount),
-            LiquidityProviderOperation::Withdrawal => (swap_token_b_amount - source_amount, swap_token_a_amount)
+        let new_swap_source_amount = match liquidity_provider_operation {
+            LiquidityProviderOperation::Deposit => swap_source_amount + source_amount,
+            LiquidityProviderOperation::Withdrawal => swap_source_amount - source_amount,
```
joncinque, line 221: I forgot to mention this earlier, but this should use checked math, `checked_add` and `checked_sub`

joncinque (commented):
One more little comment, otherwise this is on the right track!

rawfalafel:
Closing this PR to let @akamutsu open a new PR as the owner!

PR #1425, opened 2021-03-10
Add bpf float performance tests

jackcmay:


libraries/math/src/processor.rs
```diff
@@ -33,5 +33,34 @@ pub fn process_instruction(
             msg!("{}", result);
             Ok(())
         }
+        MathInstruction::U64Multiply {
+            multiplicand,
+            multiplier,
+        } => {
+            msg!("Calculating U64 Multiply");
+            let result = multiplicand + multiplier;
```
joncinque, line 41: Post-merge review! Should this be `*`?  Also, now I know the word `multiplicand`

PR #1433, opened 2021-03-11
Token cli: Enable short arg, monikers

CriesofCarrots:
Supports the short `-u` url arg, as well as adds parsing for public-cluster monikers


ryoqun:
@CriesofCarrots hooray! Yeah, I was very itched to do this. But you beat me to it. :)

PR #1434, opened 2021-03-11
lending: Add SetLendingMarketOwner instruction

jstarry:
#### Problem
No way to migrate an existing lending market to a new owner. In this case, I would like to move the owner authority to a multisig contract

### DIFF
--- a/token-lending/program/src/instruction.rs
+++ b/token-lending/program/src/instruction.rs
@@ -26,7 +26,10 @@ pub enum BorrowAmountType {
 /// Instructions supported by the lending program.
 #[derive(Clone, Debug, PartialEq)]
 pub enum LendingInstruction {
+    // 0
     /// Initializes a new lending market.
+    ///
+    /// Accounts expected by this instruction:
     ///
     ///   0. `[writable]` Lending market account.
     ///   1. `[]` Quote currency SPL Token mint. Must be initialized.
@@ -37,7 +40,10 @@ pub enum LendingInstruction {
         market_owner: Pubkey,
     },
 
+    // 1
     /// Initializes a new lending market reserve.
+    ///
+    /// Accounts expected by this instruction:
     ///
     ///   0. `[writable]` Source liquidity token account.  $authority can transfer $liquidity_amount
     ///   1. `[writable]` Destination collateral token account - uninitialized
@@ -63,7 +69,10 @@ pub enum LendingInstruction {
         config: ReserveConfig,
     },
 
+    // 2
     /// Initializes a new loan obligation.
+    ///
+    /// Accounts expected by this instruction:
     ///
     ///   0. `[]` Deposit reserve account.
     ///   1. `[]` Borrow reserve account.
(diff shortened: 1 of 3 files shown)


token-lending/program/src/instruction.rs
```diff
@@ -78,44 +87,53 @@ pub enum LendingInstruction {
...
+    ///
     ///   0. `[writable]` Source liquidity token account. $authority can transfer $liquidity_amount
     ///   1. `[writable]` Destination collateral token account.
     ///   2. `[writable]` Reserve account.
     ///   3. `[writable]` Reserve liquidity supply SPL Token account.
     ///   4. `[writable]` Reserve collateral SPL Token mint.
     ///   5. `[]` Lending market account.
     ///   6. `[]` Derived lending market authority.
-    ///   7. `[]` User transfer authority ($authority).
+    ///   7. `[signer]` User transfer authority ($authority).
```
jordansexton, line 103: Nice

token-lending/program/src/processor.rs
```diff
@@ -1497,6 +1501,33 @@ fn process_withdraw_obligation_collateral(
...
+        return Err(LendingError::InvalidAccountOwner.into());
+    }
+    if &lending_market.owner != lending_market_owner_info.key {
+        return Err(LendingError::InvalidMarketOwner.into());
+    }
+    if !lending_market_owner_info.is_signer {
+        return Err(LendingError::InvalidSigner.into());
+    }
+
+    lending_market.owner = new_owner;
```
jordansexton, line 1525: Should this error if the lending market owner is already `new_owner`? (I don't know how no-ops are generally handled)

token-lending/program/tests/set_lending_market_owner.rs
```diff
@@ -0,0 +1,139 @@
...
+        Some(&payer.pubkey()),
+    );
+
+    transaction.sign(&[&payer, &lending_market.owner], recent_blockhash);
+
+    banks_client
+        .process_transaction(transaction)
+        .await
+        .map_err(|e| e.unwrap())
+        .unwrap();
```
jordansexton, line 50: What does this do if no error is expected / what's the difference between this and

```rust
assert!(banks_client.process_transaction(transaction).await.is_ok());
```

jordansexton (approved):
Looks good, I like the additional instruction docs. Had two questions but non blocking.

token-lending/program/src/processor.rs
jstarry: I don't think there's any downside to allowing a no-op. The SPL token SetAuthority instruction also allows no-ops

token-lending/program/tests/set_lending_market_owner.rs
jstarry: If it errors, it will panic due to the unwrap. An assert is better, thanks!

jstarry: Oops, my patch for the assert didn't get pushed before the auto-merge label took effect. I'll clean this up later, promise 😉

PR #1461, opened 2021-03-16
Fix CI no more space on device

CriesofCarrots:


mvines:
Oh cool.  I'm pretty sure I just ran into this, and then just mashed "Rebase and merge" 🤠

CriesofCarrots:
> Oh cool. I'm pretty sure I just ran into this, and then just mashed "Rebase and merge" 🤠

Yeah, seems to be hitting all the CI runs now. Not sure what pushed us over the edge all of a sudden, but we're definitely filling up the allotted 14GB. We might want to split build/testing into multiple jobs in the future, esp as we add more programs.

PR #1456, opened 2021-03-16
 Bump Solana version to 1.5.15

mvines:
Port to Solana v1.5.15.  This release contains changes that'll ease porting to v1.6.1 next

PR #1470, opened 2021-03-18
Binary-oracle: Remove explicit tokio

CriesofCarrots:
Binary-oracle-pair program was merged after https://github.com/solana-labs/solana-program-library/pull/1456, and didn't get those changes. As a result, the program is breaking solana CI downstream-projects build on v1.5

cc @joncinque


joncinque:
@CriesofCarrots Ah, my bad -- thanks for fixing this!

PR #1472, opened 2021-03-19
wip: imporve error message of js spl-token client

ryoqun:
sadly, this isn't helpful.

```
Error: Failed to find account
     at Token.getAccountInfo (node_modules/@solana/spl-token/lib/index.cjs.js:474:13)
     at runMicrotasks (<anonymous>)
     at processTicksAndRejections (internal/process/task_queues.js:97:5)
     at async Token.getOrCreateAssociatedAccountInfo node_modules/@solana/spl-token/lib/index.cjs.js:319:16)
```

### DIFF
--- a/token/js/client/token.js
+++ b/token/js/client/token.js
@@ -34,6 +34,10 @@ export const ASSOCIATED_TOKEN_PROGRAM_ID: PublicKey = new PublicKey(
 const FAILED_TO_FIND_ACCOUNT = 'Failed to find account';
 const INVALID_ACCOUNT_OWNER = 'Invalid account owner';
 
+function formatDynamicErrorMessage(prefix, context) {
+  return `${prefix}: ${context}`;
+}
+
 /**
  * Unfortunately, BufferLayout.encode uses an `instanceof` check for `Buffer`
  * which fails when using `publicKey.toBuffer()` directly because the bundled `Buffer`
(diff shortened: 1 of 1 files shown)


token/js/client/token.js
```diff
@@ -764,13 +768,13 @@ export class Token {
   ): Promise<AccountInfo> {
     const info = await this.connection.getAccountInfo(account, commitment);
     if (info === null) {
-      throw new Error(FAILED_TO_FIND_ACCOUNT);
+      throw new Error(formatDynamicErrorMessage(FAILED_TO_FIND_ACCOUNT, account.toString());
```
ryoqun, line 771: @jstarry these days, what's most fashionable way of structured error in js land? Still, these adhoc way are rampant?

ryoqun: for two things, I want to add some `error.code = FAILED_...` and `error.context = {address}`.. For internal use.

ryoqun:
Also, I think this is pretty low-priority. I think just library consumers can print whatever arguments they gave to these functions.

token/js/client/token.js
jstarry: You can use the following syntax with backticks: `` `Failed to find account with address ${account.toString()}` ``

jstarry: I will have to do some research, this is something that could be improved in the web3 library as well!

PR #1474, opened 2021-03-21
fixed formula to match referenced doc [1] + formatting

keredson:
`P_{margin}=iRP` didn't make sense, as P isn't defined, and i is already defined as the price (from the oracle).
the referenced doc contains `P_{margin}=iR`, so I assume this is a typo.


bartosz-lipinski:
Thanks!

PR #1481, opened 2021-03-22
token/cli: Rename --allow-empty-recipient to --allow-unfunded-recipient

mvines:
Align `spl-token` with https://github.com/solana-labs/solana/pull/16060

`--allow-empty-recipient` is retained as a hidden arg for backwards compatibility.


t-nelson (approved):
:shipit:

PR #1504, opened 2021-03-24
Proposal for token-swap curve using stake pool account

bartosz-lipinski:


token-swap/proposals/StakePoolStableCurve.md
```diff
@@ -0,0 +1,9 @@
+# Combine Stable Curve with Stake Pool for price discovery
+
+Implement a curve for token-swap that accepts stake pool account and uses stable curve to swap assets. 
+[Stake pool account](https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/src/state.rs#L17) will be used for initial price discovery then stable curve would be used to derive price based on the size of the order.
```
joncinque, line 4: tiny nit: stable curve would be used to derive price based on the size of the order and available reserves

is that a little clearer?

joncinque (dismissed):
Looks good, just one tiny nit that probably isn't necessary

PR #1505, opened 2021-03-24
stake-pool: Add borsh support and specify size on creation

joncinque:
We can't specify the size in the instruction unfortunately, since we'd
only have 10kb max for the validator list. At roughly 50 bytes per
validator, that only gives us 200 validators.

On the flip side, using Borsh means we can allow the validator stake list
to be any size!


stake-pool/program/src/state.rs
```diff
@@ -185,16 +158,14 @@ pub struct ValidatorStakeInfo {
...
-
-    /// Header length
-    pub const HEADER_LEN: usize = size_of::<u8>() + size_of::<u16>();
-
     /// Version of validator stake list
     pub const VALIDATOR_STAKE_LIST_VERSION: u8 = 1;
 
+    /// Get the Borsh size for a list containing up to `max_validators`
+    pub fn size_with_max_validators(max_validators: usize) -> usize {
+        1 + 4 + get_packed_len::<ValidatorStakeInfo>() * max_validators
```
joncinque, line 166: Not sure how to do this better -- any ideas?

mvines:
> We can't specify the size in the instruction unfortunately, since we'd only have 10kb max for the validator list

I don't get what you mean by this?

stake-pool/program/src/state.rs
mvines: I'm still not a fan of the `version: u8`.  What if the first byte was a Borsh enum that indicated the kind of stake pool account?  

It looks like currently if you were to pass one kind of stake pool account into an instruction that expected another kind, it *might* just work, if we got unlucky and the deserialization succeeded.  This could potentially be exploitable

joncinque: That's a great point.  The convention, instead of `version`, should be the struct type, which allows us to version more easily in the future, as we could have variants like `StakePoolV2` as needed

joncinque:
If you perform `create_account` as a CPI, we're limited to allocating a max of 10KB.  The runtime provides a 10KB buffer for any new allocations that are performed, so that you can use it right away.  Here's the constant: https://github.com/solana-labs/solana/blob/43a116a84ea3f71007b97d5d58054e2eccf536ac/sdk/program/src/entrypoint.rs#L134

I tried a test, and if you attempt to do `create_account` from a CPI, it works for exactly `1024 * 10` and fails for `1024 * 10 + 1`.  So if we change the `initialize` instruction to take in a `size` for the validator list, and then perform `create_account` inside the instruction, the max acceptable size would be `1024 * 10` (minus the size of the stake pool).

stake-pool/program/src/state.rs
joncinque: Done in bdff716

mvines:
> If you perform `create_account` as a CPI, we're limited to allocating a max of 10KB. The runtime provides a 10KB buffer for any new allocations that are performed, so that you can use it right away. Here's the constant: https://github.com/solana-labs/solana/blob/43a116a84ea3f71007b97d5d58054e2eccf536ac/sdk/program/src/entrypoint.rs#L134
> 
> I tried a test, and if you attempt to do `create_account` from a CPI, it works for exactly `1024 * 10` and fails for `1024 * 10 + 1`. So if we change the `initialize` instruction to take in a `size` for the validator list, and then perform `create_account` inside the instruction, the max acceptable size would be `1024 * 10` (minus the size of the stake pool).

Ah, I see.   The validator list account doesn't need to be an PDA.   The address of the ValidatorList could be included in the StakePool account.  This would allow the create to be done outside of a CPI and give us immediate access to the full 10MB max account size.

stake-pool/program/src/state.rs
```diff
@@ -2,21 +2,38 @@
...
 
+/// Enum representing the account type managed by the program
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub enum AccountType {
+    /// If the account has not been initialized, the enum will be 0
+    Uninitialized,
+    /// Stake pool v1
+    StakePoolV1,
+    /// Validator stake list
+    ValidatorStakeListV1,
```
mvines, line 22: nit: v2 doesn't exist yet and might never exist, no need to allude to it
```suggestion
    /// Stake pool
    StakePool,
    /// Validator stake list
    ValidatorStakeList,
```

stake-pool/program/src/state.rs
mvines: What if `size_with_max_validators()` created a `ValidatorStakeList ` struct with `max_validators` in it, then `get_packed_len()`ed it instead?   The cost of doing so in tests or client code is no big deal

joncinque: I like the idea!  There's a slight issue here -- `get_packed_len()` works on types, not instances, which means we'd have to make another function which takes in an instance, `get_instance_packed_len()`? That could work properly on `Sequence` types, whose size is not known at compile-time, and instead looks at `capacity` for `Vec`, `HashSet`, and `HashMap`.  What do you think?

joncinque:
> Ah, I see. The validator list account doesn't need to be an PDA. The address of the ValidatorList could be included in the StakePool account. This would allow the create to be done outside of a CPI and give us immediate access to the full 10MB max account size.

Ok great, that's the way it works now!  I thought you wanted to specify the size as a parameter at init.  In this state, it just takes in a validator stake list account and initializes it, assuming it was created earlier.  Does that work?

mvines:
> Ok great, that's the way it works now! I thought you wanted to specify the size as a parameter at init. In this state, it just takes in a validator stake list account and initializes it, assuming it was created earlier. Does that work?

My motivation was to allow the stake pool manager the flexibility of determining the max number of validators in the pool, as that affects the rent cost of the account.   For stake-o-matic, we'd need to support thousands of validators.   But others might be content with a max of 20 validators.

stake-pool/program/src/state.rs
mvines: sgtm

joncinque:
Sorry, I'm misunderstanding something.  You have the ability to do that with this PR -- you `create_account` on your validator stake list with the size you want, and then the stake pool Just Works (TM) with it.  The `size_with_max_validators` helps you figure out how many bytes you'll need depending on how many validators you expect to support in your pool.  What else is needed?

mvines:
> Sorry, I'm misunderstanding something. You have the ability to do that with this PR -- you `create_account` on your validator stake list with the size you want, and then the stake pool Just Works (TM) with it. The `size_with_max_validators` helps you figure out how many bytes you'll need depending on how many validators you expect to support in your pool. What else is needed?

Close I think!   The max validators is not stored in `ValidatorStakeList`, so there's no clean error when one too many validators are added to the list via `AddValidatorStakeAccount`.   Instead we just hope that this blows up:
https://github.com/solana-labs/solana-program-library/blob/0a009baefc00b970089b1627c84e0f3d36206154/stake-pool/program/src/processor.rs#L599

Beyond that lgtm

----
I have a couple more naming nits, thanks for bearing with me!  (If you agree, this could all be a separate PR as this PR is already mushrooming on us)
* `ValidatorStakeList` -> `ValidatorList`?  
* `AddValidatorStakeAccount`/`RemoveValidatorStakeAccount` -> `AddValidatorToPool`/`RemoveValidatorFromPool`  (`CreateValidatorStakeAccount` seems fine to me though)
* `UpdateListBalance` -> `UpdateValidatorListBalance`?
* `UpdatePoolBalance` -> `UpdateStakePoolBalance`?

joncinque:
I put the renames into #1507 

As for the `max_validators`, do you imagine deducing the max validators based on the size of the account or passing it into the instruction?  My instinct is to deduce it to avoid footguns, but that requires a not-so-pretty calculation of `(size - 1 - 4) /  get_packed_len::<ValidatorStakeInfo>()`.  thoughts?

mvines:
> I put the renames into #1507

👍🏼 

> As for the `max_validators`, do you imagine deducing the max validators based on the size of the account or passing it into the instruction? My instinct is to deduce it to avoid footguns, but that requires a not-so-pretty calculation of `(size - 1 - 4) / get_packed_len::<ValidatorStakeInfo>()`. thoughts?

How about both.  User passes it in, and the program confirms that the limit that the user requests matches the not-so-pretty calculation based on the account data length.    After reviewing https://borsh.io/, I feel better about the not-so-pretty calculation since the serialization format is fully specified.

joncinque:
Should be all ready now!  In a fascinating turn of events, calling `get_packed_len` from inside a program creates an ELF issue with the runtime, so I created a minimal reproducible example at #1511, and went with a different approach to get the instance packed len, using a simple counter as a writer.

mvines (dismissed):
:shipit: !

PR #1515, opened 2021-03-29
Freeze Authorities for Token Swap/AMMs

bhgames:
As part of @bartosz-lipinski 's NFT project, this introduces Freeze Authorities for the Token Swap project and a bitmask that covers a toggle on-off switch for each of the actions available in the program. A freeze authority can be introduced on creation of a swap and then that signer (single sig only) can toggle on/off the bits with another action.

Couples with https://github.com/project-serum/oyster-swap/pull/39

### DIFF
--- a/token-swap/js/client/token-swap.js
+++ b/token-swap/js/client/token-swap.js
@@ -83,6 +83,9 @@ export const TokenSwapLayout: typeof BufferLayout.Structure = BufferLayout.struc
     Layout.uint64('hostFeeDenominator'),
     BufferLayout.u8('curveType'),
     BufferLayout.blob(32, 'curveParameters'),
+    BufferLayout.u8('freezeAuthorityOption'),
+    Layout.publicKey('freezeAuthority'),
+    BufferLayout.u8('freezeAuthorityBitMask')
   ],
 );
 
@@ -319,6 +322,9 @@ export class TokenSwap {
       BufferLayout.nu64('hostFeeDenominator'),
       BufferLayout.u8('curveType'),
       BufferLayout.blob(32, 'curveParameters'),
+      BufferLayout.u8('freezeAuthorityOption'),
+      Layout.publicKey('freezeAuthority'),
+      BufferLayout.u8('freezeAuthorityBitMask')
     ]);
     let data = Buffer.alloc(1024);
     {
--- a/token-swap/js/client/token-swap.js
+++ b/token-swap/js/client/token-swap.js
@@ -83,7 +83,7 @@ export const TokenSwapLayout: typeof BufferLayout.Structure = BufferLayout.struc
     Layout.uint64('hostFeeDenominator'),
     BufferLayout.u8('curveType'),
     BufferLayout.blob(32, 'curveParameters'),
-    BufferLayout.u8('freezeAuthorityOption'),
+    BufferLayout.u32('freezeAuthorityOption'),
     Layout.publicKey('freezeAuthority'),
     BufferLayout.u8('freezeAuthorityBitMask'),
   ],
@@ -309,6 +309,7 @@ export class TokenSwap {
       {pubkey: tokenAccountPool, isSigner: false, isWritable: true},
       {pubkey: tokenProgramId, isSigner: false, isWritable: false},
     ];
+
     const commandDataLayout = BufferLayout.struct([
       BufferLayout.u8('instruction'),
       BufferLayout.u8('nonce'),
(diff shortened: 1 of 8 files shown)


token-swap/program/src/processor.rs
```diff
@@ -345,6 +391,7 @@ impl Processor {
             return Err(ProgramError::IncorrectProgramId);
         }
         let token_swap = SwapVersion::unpack(&swap_info.data.borrow())?;
+        Self::check_allowed_to_use(token_swap.as_ref(), 0)?;
```
joncinque, line 394: take it or leave it, but a better alternative to hard-coding these bit numbers would be to use a true bitmask library.  bitflags seems to be a pretty good one https://docs.rs/bitflags/1.2.1/bitflags/

token-swap/program/src/processor.rs
```diff
@@ -149,6 +153,43 @@ impl Processor {
...
+            }
+            COption::None => Err(SwapError::UnauthorizedToFreeze.into()),
+        }
+    }
+
+    fn check_allowed_to_use(token_swap: &dyn SwapState, bit_position: usize) -> ProgramResult {
+        let bitmask: u8 = token_swap.freeze_authority_bit_mask();
+        match Self::get_bit_at(bitmask, bit_position) {
+            Ok(frozen) => {
+                if frozen {
```
joncinque, line 184: A newish best practice is to have a `msg!` explaining what's happening before throwing an error, which makes debugging from the explorer much easier.  Swap isn't doing this everywhere of course, but this new code should :smile:

token-swap/program/src/processor.rs
```diff
@@ -992,6 +1048,39 @@ impl Processor {
         Ok(())
     }
 
+    /// Processes a [SetFreezeAuthorityBitMask](enum.Instruction.html).
+    pub fn process_set_freeze_authority_bit_mask(
+        _: &Pubkey,
+        freeze_authority_bit_mask: u8,
```
joncinque, line 1054: Setting a raw bitmask is fine, but be sure there's an easy library to do it from the outside!

token-swap/program/src/processor.rs
```diff
@@ -992,6 +1048,39 @@ impl Processor {
...
+        accounts: &[AccountInfo],
+    ) -> ProgramResult {
+        let account_info_iter = &mut accounts.iter();
+        let swap_info = next_account_info(account_info_iter)?;
+        let freeze_authority_info = next_account_info(account_info_iter)?;
+        let token_swap = SwapVersion::unpack(&swap_info.data.borrow())?;
+
+        Self::check_allowed_to_freeze(token_swap.as_ref(), freeze_authority_info)?;
+
+        let clone = SwapVersion::SwapV2(SwapV2 {
```
joncinque, line 1064: This can become brittle once we add a v3 -- could you deserialize it and set the new bitmask through a function?  The v1 swap can error out for it

token-swap/program/src/processor.rs
```diff
@@ -1152,6 +1255,18 @@ impl PrintProgramError for SwapError {
...
+            SwapError::InvalidBitMaskOperation => {
+                msg!("Error: Attempted to access an invalid bit in the freeze authority bitmask")
+            }
+            SwapError::FrozenAction => {
+                msg!("Error: This action has been frozen by the Freeze Authority")
+            }
+            SwapError::UnauthorizedToFreeze => {
+                msg!("Error: Unauthorized to freeze")
+            }
+            SwapError::SwapV1UnsupportedAction => {
```
joncinque, line 1267: I might be missing something, but it doesn't seem like this is ever used

joncinque (commented):
Lots of great stuff here!  Mostly small comments to look at, but I'm liking the direction of this

token-swap/js/client/token-swap.js
```diff
@@ -83,6 +83,9 @@ export const TokenSwapLayout: typeof BufferLayout.Structure = BufferLayout.struc
     Layout.uint64('hostFeeDenominator'),
     BufferLayout.u8('curveType'),
     BufferLayout.blob(32, 'curveParameters'),
+    BufferLayout.u32('freezeAuthorityOption'),
```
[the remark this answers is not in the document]
bhgames: Yeah I'm not sure why he uses it. I just copied his pattern. The instructions seem to expect a u8 though, while the storage uses a u32.

token-swap/program/src/instruction.rs
```diff
@@ -107,7 +131,8 @@ pub enum SwapInstruction {
     ///   Must be empty, not owned by swap authority
     ///   6. `[writable]` Pool Token Account to deposit the initial pool token
     ///   supply.  Must be empty, not owned by swap authority.
-    ///   7. '[]` Token program id
+    ///   7. '[]` Freeze authority. Optional.
```
joncinque, line 134: This isn't actually passed as an account, no need to have it as a comment

bhgames: Oops!

PR #1525, opened 2021-03-30
Add spl-token transfer --no-wait

ryoqun:
#### program

Currently, `spl transfer` lacks `--no-wait` while `solana transfer` has it. This lack of feature-parity is causing some trouble for partners.

#### solution

Implement it. The code is shamelessly copied from the `solana` code for the strict same behavior, implementation and documentation.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1238,6 +1238,7 @@ impl offline::ArgsConfig for SignOnlyNeedsDelegateAddress {
 
 fn main() {
     let default_decimals = &format!("{}", native_mint::DECIMALS);
+    let mut no_wait = false;
     let app_matches = App::new(crate_name!())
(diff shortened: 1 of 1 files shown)


token/cli/src/main.rs
```diff
@@ -2044,6 +2051,7 @@ fn main() {
             let fund_recipient = matches.is_present("fund_recipient");
             let allow_unfunded_recipient = matches.is_present("allow_empty_recipient")
                 || matches.is_present("allow_unfunded_recipient");
+            no_wait = matches.is_present("no_wait");
```
ryoqun, line 2054: @CriesofCarrots as discussed on dm lightly, I restricted `no_wait` only to `spl-token transfer`.

CriesofCarrots (approved):
lgtm!

PR #1535, opened 2021-03-30
Automatically update the stake pool if needed, use the `--no-update` flag to disable

mvines:
Using a stake pool with `solana-test-validator --slots-per-epoch 32` is super annoying because you need to run `spl-stake-pool update` every 15 seconds.  Instead detect and automatically update the pool by default


joncinque:
Very slick, thanks!

PR #1560, opened 2021-04-01
stake-pool: Separate manager from owner

joncinque:
* Add manager pubkey to stake pool
* Differentiate manager functions from owner functions
* Include a `set_manager` function to be used by the owner
* Change the term `owner` to `authority` in the CLI for clarity


mvines:
I think we align the authority terminology with normal stake accounts as much as possible:  
1. The pool "manager" could be the "stake authority"/"Staker", identical to the stake account terms

2. Renaming "owner" to "authority" to me confuses the situation, since there can be multiple authorities.   However using the "withdraw authority"/"Withdrawer" from stake account terms doesn't really make sense here.   I feel like the "manage authority"/"Manager" actually might work here though.  This authority allows for the management of who participates in the pool, sets the fee, etc.

stake-pool/program/src/instruction.rs
```diff
@@ -31,12 +31,13 @@ pub enum StakePoolInstruction {
...
     ///   1. `[s]` Owner
-    ///   2. `[w]` Uninitialized validator stake list storage account
-    ///   3. `[]` pool token Mint. Must be non zero, owned by withdraw authority.
-    ///   4. `[]` Pool Account to deposit the generated fee for owner.
-    ///   5. `[]` Clock sysvar
-    ///   6. `[]` Rent sysvar
-    ///   7. `[]` Token program id
+    ///   2. `[]` Manager
+    ///   3. `[w]` Uninitialized validator stake list storage account
+    ///   4. `[]` Pool token mint. Must be non zero, owned by withdraw authority.
```
mvines, line 36: who is `withdraw authority` here?

mvines:
> Renaming "owner" to "authority" to me confuses the situation, since there can be multiple authorities. 

What is the list of things that the owner/authority can actually do?  (1) Set the fee, (2) change the manager/owner.  Looking at that full list might help us find a better name for this authority

stake-pool/program/src/instruction.rs
joncinque: That's the stake pool program derived address, the actual withdraw authority on all stake accounts in the pool.

joncinque:
You've got it all correct:

* owner (now manager): sets the fee, manager (staker), and owner
* manager (now staker): adds / removes validators, rebalance operations

I'm down for:

* owner -> manager
* manager -> staker

It's a good point, let's reuse terminology where appropriate.  And `owner` isn't a great term since it implies that they may have withdraw authority over the stakes.

The rename of `authority` in the CLI is different and entirely for cosmetics in the code. The CLI may be signing for owner (manager) or manager (staker) operations, so just having `config.owner` in the code isn't great.  Do you have a better term than `authority`?

mvines:
> The rename of `authority` in the CLI is different and entirely for cosmetics in the code. The CLI may be signing for owner (manager) or manager (staker) operations, so just having `config.owner` in the code isn't great. Do you have a better term than `authority`?

`stake_authority` and `manage_authority` perhaps?  That is, be explicit instead of trying to re-use a single authority field for both?

joncinque:
> `stake_authority` and `manage_authority` perhaps? That is, be explicit instead of trying to re-use a single authority field for both?

Makes sense to me!  At that point, we may as well mimic the stake pool struct and have them as just `staker` and `manager`

stake-pool/cli/src/main.rs
```diff
@@ -1232,12 +1283,32 @@ fn main() {
...
             .group(ArgGroup::with_name("new_accounts")
-                .arg("new_owner")
+                .arg("new_manager")
                 .arg("new_fee_receiver")
                 .required(true)
                 .multiple(true)
             )
         )
+        .subcommand(SubCommand::with_name("set-staker")
+            .about("Change staker account for the stake pool. Must be signed by the manager.")
```
mvines, line 1293: It would be nice if the current staker could change the staker as well.  This is the case for stake accounts as well.  The use case is that you don't want to bother your boss (the manager), they already delegated you the authority

mvines (approved):
Looks great, just one comment about adding a "don't disturb the boss" mode

PR #1563, opened 2021-04-01
stake-pool: Add rebalancing instruction interface

joncinque:
This can get the discussion going about the interface for the new rebalancing instructions.  I may be missing some sysvars or builtin programs, which will come out in implementation and testing.

Some open questions:
* PDA derivation for transient stake accounts: I'm not wedded to the solution of just using the stake account, and just opted for the simplest thing
* Accounting gets more complicated.  In order to properly keep track of the amount of stake under management, we need to also update the balances in transient accounts.  This proposes sending in pairs of accounts (canonical and transient) during `UpdateValidatorListBalance`.
* This adds a funder for transient stake accounts, but if we make sure that all movements are for `stake_account_rent_exemption + 1` lamport, then we can have all of the transient stake accounts self-funded (unless I'm forgetting how splitting works).

Let me know what you think and how this will work for stake-o-matic!  I'll prefer to merge this once it's ok and then implement bit by bit.

### DIFF
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -104,7 +104,7 @@ pub enum StakePoolInstruction {
     /// some amount of stake, up to the total activated stake, from the canonical
     /// validator stake account, into its "transient" stake account, defined by:
     ///
-    /// ```
+    /// ```ignore
     /// Pubkey::find_program_address(
     ///     &[&stake_account_address.to_bytes()[..32],], program_id,
     /// )
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -33,7 +33,7 @@ pub enum StakePoolInstruction {
     ///   1. `[s]` Manager
     ///   2. `[]` Staker
     ///   3. `[w]` Uninitialized validator stake list storage account
-    ///   4. `[]` Reserve stake account Must be initialized, have zero balance,
+    ///   4. `[]` Reserve stake account must be initialized, have zero balance,
     ///       and staker / withdrawer authority set to pool withdraw authority.
     ///   5. `[]` Pool token mint. Must be non zero, owned by withdraw authority.
     ///   6. `[]` Pool account to deposit the generated fee for manager.
@@ -94,7 +94,10 @@ pub enum StakePoolInstruction {
     ///  10. `[]` Stake program id,
     RemoveValidatorFromPool,
 
-    /// (Staker only) Split a validator stake account into a transient stake account.
+    /// (Staker only) Decrease active stake on a validator, eventually moving it to the reserve
+    ///
+    /// Internally, this instruction splits a validator stake account into its
+    /// corresponding transient stake account and deactivates it.
     ///
     /// In order to rebalance the pool without taking custody, the staker needs
     /// a way of reducing the stake on a stake account. This instruction splits
@@ -106,8 +109,6 @@ pub enum StakePoolInstruction {
     ///     &[&stake_account_address.to_bytes()[..32],], program_id,
     /// )
     /// ```
-    ///
-    /// After the split, the instruction automatically deactivates the stake.
     ///
     /// The instruction only succeeds if the transient stake account does not
     /// exist. The amount of lamports to move must be at least rent-exemption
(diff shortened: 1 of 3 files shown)


mvines:
This was useful to get me thinking, thanks.   Here's a slight variation of the current model that I came up with:

----
The general idea is that each validator is only allowed a single stake operation per epoch: the pool staker can _either_ add or remove some stake from the validator.
  
There also exists a new undelegated stake account associated with the pool called the _reserve stake account_.  A user cannot deposit stake into the reserve, but they may withdraw from the reserve if all validator stake accounts are depleted (to prevent stuck stake tokens).

The single staking operation per validator per epoch either adds stake from the reserve, or removes stake back into the reserve.  This uses transient stake accounts like in this PR.

`UpdateValidatorListBalance` will check the transient stake account for each validator.  If fully active, it'll merge the transient stake into the main validator stake.  If fully deactivated, it'll merge the transient stake into the reserve stake.

This means that once the stake pool is updated for the epoch, any remaining transient stake accounts are still activating/deactivating due to more than 25% stake moment and that validator's stake cannot be modified this epoch.  Otherwise all stake is either active in the validator stake accounts, or fully inactive in the reserve stake account and ready to be moved to a new validator by the pool staker.

----
The stake-o-matic implementation is then quite simple, and looks like this:
1. Figure the desired stake amount for each validator in the pool based on the stake-o-matic criteria and the current pool total stake
2. Ensure the pool is updated for the current epoch, issue `UpdateValidatorListBalance` instructions as needed.
3. For each validator in the pool:
         * If the transient stake account exists for this validator, do nothing
         * If the validator's stake is less than the desired amount, issue an `ReserveToValidator(u64)` instruction.   This instruction splits a new transient stake account from the reserve for the validator and actives it
         * If the validator's stake is more than the desired amount, issue a `ValidatorToReserve(u64)` instruction.   This instruction splits a new transient stake account from the validator's stake and deactivates it

joncinque:
This is a neat and clear model, especially the permissionless side for merging and consolidating stake.  I updated the instructions / struct to capture this new model, so let me know what you think.

There's one little wrinkle to keep in mind for stake-o-matic regarding `credits_observed` on the transient stake accounts.  You'll have to wait *at least* one full epoch to merge back in, but *at most* one epoch to deactivate.

Thanks for the feedback!  If this looks good, I'll add helper functions to create the instructions and merge this in.  That way you can start using them even if the program does nothing.

stake-pool/program/src/instruction.rs
```diff
@@ -33,11 +33,13 @@ pub enum StakePoolInstruction {
     ///   1. `[s]` Manager
     ///   2. `[]` Staker
     ///   3. `[w]` Uninitialized validator stake list storage account
-    ///   4. `[]` Pool token mint. Must be non zero, owned by withdraw authority.
-    ///   5. `[]` Pool account to deposit the generated fee for manager.
-    ///   6. `[]` Clock sysvar
-    ///   7. `[]` Rent sysvar
-    ///   8. `[]` Token program id
+    ///   4. `[]` Reserve stake account Must be initialized, have zero balance,
```
mvines, line 36:
```suggestion
    ///   4. `[]` Reserve stake account must be initialized, have zero balance,
```

stake-pool/program/src/instruction.rs
```diff
@@ -92,18 +94,79 @@ pub enum StakePoolInstruction {
...
+    ///  2. `[]` Validator list
+    ///  3. `[]` Stake pool withdraw authority
+    ///  5. `[w]` Canonical stake account to split from
+    ///  5. `[w]` Transient stake account to receive split
+    ///  6. `[]` Clock sysvar
+    ///  7. `[]` Rent sysvar
+    ///  8. `[]` System program
+    ///  9. `[]` Stake program
+    ///  userdata: amount of lamports to split
+    ValidatorToReserve(u64),
```
mvines, line 127: I don't love the instruction names `ValidatorToReserve` and `ReserveToValidator`, I was just being lazy in my initial writeup.

Maybe `IncreaseValidatorStake` and `DecreaseValidatorStake`.   Any other ideas?

mvines (commented):
Program instruction changes look nice, thanks 

> There's one little wrinkle to keep in mind for stake-o-matic regarding credits_observed on the transient stake accounts. You'll have to wait at least one full epoch to merge back in, 

Ah yes, that's slightly annoying but oh well.  Fortunately that's basically just an implementation detail of the stake program that we could potentially attempt to fix in the future with no impact here

stake-pool/program/src/instruction.rs
joncinque: D'oh, I thought you wanted those names.  I like `IncreaseValidatorStake` and `DecreaseValidatorStake`, since they're clear opposites that accurately describe what will happen in the future.  The only other options I could come up with that describe what's happening in the instruction are `DelegateReserveStakeToValidator` and `DeactivateValidatorStakeToReserve`

mvines: heh, na.  I came up with variations of `DelegateReserveStakeToValidator` as well and it seemed like a mouthful!

Issue #1567, opened 2021-04-02
stake-pool: the stake accounts passed into `UpdateValidatorListBalance` can be spoofed


mvines:
`UpdateValidatorListBalance` doesn't use `get_validator_checked()`:

https://github.com/solana-labs/solana-program-library/blob/092432f1e11b6453f3b719f225fd85cc27625fd7/stake-pool/program/src/processor.rs#L63

so I can forge my own stake accounts with less lamports than what the pool currently holds for a validator, and cause `stake_total` to be incorrect:
https://github.com/solana-labs/solana-program-library/blob/092432f1e11b6453f3b719f225fd85cc27625fd7/stake-pool/program/src/state.rs#L53-L54
This is surely exploitable in some form or another.

----
The `UpdateValidatorListBalance` instruction should be updated to include the stake pool account (read-only), so it has the ability to use `get_validator_checked()` to ensure the provided stake accounts are its own.

Along the way, it would be nice to avoid this search (which gets expensive if there's hundreds or thousands of validators in the pool):
https://github.com/solana-labs/solana-program-library/blob/092432f1e11b6453f3b719f225fd85cc27625fd7/stake-pool/program/src/processor.rs#L674

Instead the `UpdateValidatorListBalance` instruction could be provided a starting index and the next N validator_list accounts to be updated, in order.  This is easy for the client to figure out and will dumb down the program implementation nicely

mvines:
If the stake pool account is passed as writable into `UpdateValidatorListBalance` (at the cost of slowing down the updates due to the write-lock), the `UpdateStakePoolBalance` instruction can be removed entirely.   Now we end up with a single `Update` instruction that performs `UpdateValidatorListBalance` and at the end checks and performs an `UpdateStakePoolBalance`.   I think would be quite a nice simplification

joncinque:
Sorry, I'm not seeing how the attack is possible here.  How would you forge your own validator stake list?  It's being modified here, so the runtime will check ownership by the stake pool program, correct?  To make this work, you would need to initialize and modify a validator stake list which is owned by the stake pool program, which is only possible by creating a stake pool and adding validators.

I'm down to add the `N` parameter to Update though!

mvines:
> Sorry, I'm not seeing how the attack is possible here. How would you forge your own validator stake list?

[...]

PR #1568, opened 2021-04-03
stake-pool-cli: Break up the UpdateValidatorListBalance instructions over multiple transactions

mvines:
`stake-pool-cli update` will not work as is with a stake pool containing hundreds of validators.  Break up the `UpdateValidatorListBalance` instructions across multiple transactions so it will

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -678,8 +678,9 @@ fn command_update(config: &Config, stake_pool_address: &Pubkey) -> CommandResult
         })
         .collect();
 
-    let mut instructions: Vec<Instruction> = vec![];
+    println!("Updating stake pool...");
 
+    let mut instructions: Vec<Instruction> = vec![];
     for accounts_chunk in accounts_to_update.chunks(MAX_ACCOUNTS_TO_UPDATE) {
         instructions.push(spl_stake_pool::instruction::update_validator_list_balance(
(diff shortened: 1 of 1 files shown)


stake-pool/cli/src/main.rs
```diff
@@ -688,20 +689,22 @@ fn command_update(config: &Config, stake_pool_address: &Pubkey) -> CommandResult
...
         stake_pool_address,
         &stake_pool.validator_list,
     )?);
 
-    let mut transaction =
-        Transaction::new_with_payer(&instructions, Some(&config.fee_payer.pubkey()));
+    // TODO: A faster solution would be to send all the `update_validator_list_balance` instructions concurrently
+    for instruction in instructions {
+        let mut transaction =
+            Transaction::new_with_payer(&[instruction], Some(&config.fee_payer.pubkey()));
```
joncinque, line 701: I was hoping we could chunk out the instructions to fit more than one per transaction, but we can save that for another PR.

joncinque (approved):
Looks good!

stake-pool/cli/src/main.rs
mvines: Yes definitely.  I just didn't want to get ahead of  https://github.com/solana-labs/solana-program-library/issues/1567 here, figured we'd optimize the transaction backing later

PR #1581, opened 2021-04-07
stake-pool: Rework add / remove validator to not use pool tokens

### LINKED ISSUE #1501: stake-pool: Rework remove validator stake account
#### Problem

`remove-validator-stake-account` can cause issues for non-custodial stake pools, since it requires pool tokens to be burned.

#### Solution

Only allow removals for empty stake accounts.  The manager reclaims an empty stake account, which is fine because they provided the rent exempt reserve on creation and were not given stake pool tokens on that amount.  Also, rename to `remove-validator-from-pool`

joncinque:
#### Problem

Adding and removing validators from the pool involves pool tokens, which mixes the staker and manager roles.  The staker should not have to handle any pool tokens to do their work.

#### Solution

* Remove pool tokens from add to pool and remove validator from pool
* Set a required minimum for validator stake accounts (1 SOL + rent exemption)
* Make sure that on add validator / remove validator / withdraw we are respecting that limit
* Update book-keeping to not count the minimum amount

Fixes #1501

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -19,7 +19,6 @@ use {
         decode_error::DecodeError,
         entrypoint::ProgramResult,
         msg,
-        native_token::LAMPORTS_PER_SOL,
         program::{invoke, invoke_signed},
         program_error::PrintProgramError,
         program_error::ProgramError,
@@ -32,8 +31,6 @@ use {
     },
     spl_token::state::Mint,
 };
-
-const MAXIMUM_ACTIVE_STAKE: u64 = LAMPORTS_PER_SOL / 1_000;
 
 /// Deserialize the stake state from AccountInfo
 fn get_stake_state(
@@ -503,9 +500,12 @@ impl Processor {
         // Check amount of lamports
         let stake_lamports = **stake_account_info.lamports.borrow();
         let minimum_lamport_amount = minimum_stake_lamports(&meta);
-        let maximum_lamport_amount = minimum_lamport_amount.saturating_add(MAXIMUM_ACTIVE_STAKE);
-        if stake_lamports < minimum_lamport_amount || stake_lamports > maximum_lamport_amount {
-            msg!("Error: attempting to add stake with {} lamports, must have between {} and {} lamports", stake_lamports, minimum_lamport_amount, maximum_lamport_amount);
+        if stake_lamports != minimum_lamport_amount {
+            msg!(
+                "Error: attempting to add stake with {} lamports, must have {} lamports",
+                stake_lamports,
+                minimum_lamport_amount
+            );
             return Err(StakePoolError::StakeLamportsNotEqualToMinimum.into());
         }
 
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -4,7 +4,6 @@ use {
     crate::{
         borsh::try_from_slice_unchecked,
         error::StakePoolError,
-        id,
         instruction::{Fee, StakePoolInstruction},
         minimum_stake_lamports, stake_program,
         state::{AccountType, StakePool, ValidatorList, ValidatorStakeInfo},
@@ -561,7 +560,7 @@ impl Processor {
             return Err(ProgramError::IncorrectProgramId);
         }
 
-        if *stake_pool_info.owner != id() {
+        if stake_pool_info.owner != program_id {
             return Err(ProgramError::IncorrectProgramId);
         }
         let stake_pool = StakePool::try_from_slice(&stake_pool_info.data.borrow())?;
@@ -725,12 +724,14 @@ impl Processor {
         }
 
         let previous_lamports = stake_pool.total_stake_lamports;
-        let mut total_stake_lamports = 0;
+        let mut total_stake_lamports: u64 = 0;
         for validator_stake_record in validator_list.validators {
             if validator_stake_record.last_update_epoch < clock.epoch {
                 return Err(StakePoolError::StakeListOutOfDate.into());
             }
-            total_stake_lamports += validator_stake_record.stake_lamports;
+            total_stake_lamports = total_stake_lamports
+                .checked_add(validator_stake_record.stake_lamports)
+                .ok_or(StakePoolError::CalculationFailure)?;
         }
 
         stake_pool.total_stake_lamports = total_stake_lamports;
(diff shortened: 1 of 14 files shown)


stake-pool/program/src/state.rs
```diff
@@ -223,7 +223,7 @@ pub struct ValidatorList {
 #[derive(Clone, Copy, Debug, Default, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
 pub struct ValidatorStakeInfo {
     /// Validator vote account address
-    pub vote_account: Pubkey,
+    pub voter_pubkey: Pubkey,
```
mvines, line 226: Why this rename?  This is the vote account, not the voter (authorized voter?).   `vote_account_address` would be most specific, although we use `vote_address` in processor.rs as well.

stake-pool/program/src/processor.rs
```diff
@@ -19,65 +19,69 @@ use {
...
-        vote_account: &Pubkey,
-        program_id: &Pubkey,
-        stake_pool_info: &AccountInfo,
-        stake_account_info: &AccountInfo,
-    ) -> bool {
-        // Check stake account address validity
-        let (stake_address, _) =
-            crate::find_stake_program_address(&program_id, &vote_account, &stake_pool_info.key);
-        stake_address == *stake_account_info.key
+const MAXIMUM_ACTIVE_STAKE: u64 = LAMPORTS_PER_SOL / 1_000;
```
mvines, line 36: This constant seems named poor to me.  It probably could just be inlined at it's single use with a comment about why we can't use exactly 1 SOL

mvines: But thinking about this more, could we still require exactly 1 SOL by splitting off any excess after warmup, so that the account to add has exactly 1 SOL?  
Alternatively, what if the stake account had the exact rent minimum instead of 1 SOL.  I don't think fractional rewards matter for the initial validator's stake account

stake-pool/program/src/state.rs
joncinque: I renamed it to match `Delegation` which uses `voter_pubkey` as the name of the field.

https://github.com/solana-labs/solana/blob/8ec7e2e14f427dc6cfa0243a46091adddfe8d4d2/programs/stake/src/stake_state.rs#L220

Happy to change it to anything else though or revert it! Maybe just back to `vote_account`?

stake-pool/program/src/processor.rs
joncinque: I can definitely inline it, no problem.

Otherwise, we can enforce the exact 1 SOL by doing a split.  Since there's a tolerance, you can already do that if you want!  It seemed like a headache for the staker to also remember a tiny stake account, since you'll have to do:

* split tiny stake
* deactivate
* wait 1 epoch to withdraw

The stake account can't have the exact rent minimum unfortunately -- all stake accounts must have at least 1 lamport above rent-exemption in order to be initialized.  Unless you imagine that it's an uninitalized stake account, in which case on the first deposit, we can split everything from the user's stake into the validator account.  This creates another risk of someone bricking your validator account -- if someone deposits 1 lamport, it becomes essentially impossible for anyone else to deposit since that requires a merge.

stake-pool/program/src/state.rs
mvines: Oh yuck.  That's super old terminology.  `voter_pubkey` there ought to be changed as well, again to `vote_account_address` IMO

stake-pool/program/src/processor.rs
mvines: Taking a step back, I think there's a different way.
1. There'll soon exist a fully inactive "reserve" stake account that holds the stake that is not currently allocated to a specific validator in the pool.
2. SOL from the "reserve" stake account could be used by the program to create and activate the initial stake account for the newly added validator.  The staker authority no longer needs to provide any of their SOL.

This means that initially when the pool is bootstrapped there'll need to be some initial SOL allocated to the reserve, but that's ok and the pool creator should get proper pool tokens back in return.  This seems to make pool bootstrapping pretty nice, logically:
The pool creator who has a pot of SOL creates a stake pool for it, with no validators and just a reserve stake account.  Creator then finds a stake authority that will start adding validators and managing to the pool, and (possibly) soliciting for other pool deposits.

stake-pool/program/src/state.rs
joncinque: Sounds good, will do!

stake-pool/program/src/processor.rs
mvines: What about requiring a minimum balance of 1 SOL in all validator stake accounts?

joncinque: Just so I'm sure we're on the same page, the minimum requirement of 1 SOL is there, and this *also* adds the maximum check.  Since the staker isn't getting pool tokens in return, I added the maximum check to be sure they don't lose out more than needed, but I can certainly remove it.  Is that what you're suggesting?

mvines: > I can't redeem my pool tokens (except by tiny bits as rewards are accrued)

I think I'm stuck on this point. With a 1 SOL minimum, a malicious staker could only create 10 validator stake accounts from my deposit. It's slightly annoying that I'd need to execute 10 transactions to recover my deposit but this doesn't seem too bad and the cli could wallpaper over this.  I'm probably just not getting *it* yet.

joncinque: The problem is that we maintain the 1 SOL minimum on withdrawals, so a validator account can't ever go under 1 SOL.  If we allow the accounts to go below that, then deposits will be broken for everyone after that.

So given that, how would I get my 10 SOL back if everything goes through the reserve account?

mvines: What if
* Adding a new validator to the pool always requires the staker contribute the 1 SOL minimum (same for the reserve account at pool creation time).   The staker does not receive pool tokens for this 1 SOL minimum.
* Removing a validator from the pool requires that the staker draw that validator's stake balance back down to 1 SOL, at which point they can close the stake account and reclaim their 1 SOL (which did not receive any rewards).

joncinque: That's what this PR currently does :smile:

joncinque: The question seems to be this "maximum" amount, which I added as a courtesy so stakers are forbidden from putting too much SOL that they won't get pool tokens for.

mvines: > That's what this PR currently does 😄

lol :shipit: 

Maybe an exact value instead of a maximum then?

joncinque: > lol :shipit: 

To me, it's a good sign that we got to the same answer :sunglasses: 

> Maybe an exact value instead of a maximum then?

We can do that, it means that every time you want to add a validator to the pool, you'll need to do

* create validator stake account
* wait until it's activated
* split just the rewards gained into another stake account
* add the validator stake account to the pool
* reclaim the rewards later

Those extra steps to split and reclaim for tiny rewards seemed annoying, but as the client, if you're ok with it, I'm down to do an exact value.

mvines: How about this:
1. Deposit 1 SOL (without receiving pool tokens) into the reserve pool (always a fully inactive stake account)
2. Split 1 SOL from the reserve pool into a new stake account from the new validator and delegate it

joncinque: So essentially, we combine `CreateValidatorStakeAccount` and `AddValidatorToPool`, and the staker provides the funds?  That works for me.  In that case, do you mind if that's done in a later PR?  And I'll remove the maximum stake check in this one.

mvines: > do you mind if that's done in a later PR

wfm

stake-pool/program/src/processor.rs
```diff
@@ -768,6 +756,7 @@ impl Processor {
                 .checked_add(fee)
                 .ok_or(StakePoolError::CalculationFailure)?;
         }
+        stake_pool.total_stake_lamports = total_stake_lamports;
```
joncinque, line 759: A merge issue uncovered this bug in the fee calculation!

PR #1589, opened 2021-04-12
Auction Program Draft

Reisen:


auction/program/src/processor.rs
```diff
@@ -0,0 +1,299 @@
...
+    accounts: &[AccountInfo],
+    args: CreateAuctionArgs,
+) -> ProgramResult {
+    let account_iter = &mut accounts.iter();
+    let creator_act = next_account_info(account_iter)?;
+    let auction_act = next_account_info(account_iter)?;
+
+    let auction_path = [
+        PREFIX.as_bytes(),
+        program_id.as_ref(),
```
bhgames, line 142: With PDA-based accounts, the front end won't have the ability to allocate to them, since it won't have authority, you'll need to do the allocating.

auction/program/src/processor.rs
```diff
@@ -0,0 +1,152 @@
...
+pub enum WinnerLimit {
+    Unlimited,
+    Capped(usize),
+}
+
+/// Models a set of metadata for a bidder, meant to be stored in a PDA. This allows looking up
+/// information about a bidder regardless of if they have won, lost or cancelled.
+struct BidderMetadata {
+    /// Tracks the last time this user bid.
+    last_bid: UnixTimestamp,
```
Reisen, line 149: struct LastBid
- Add public key of user bidder.
- Bid amount is a better use of this variable.
- last_bid_timestamp: UnixTimestamp

Reisen: Store cancel in both meta and bistate.

bhgames:
Closing this PR as the code for it is being moved to another repository.

PR #1596, opened 2021-04-14
token-swap: Adjust fee constraints

bartosz-lipinski:
### Problem: 

Token swap currently allows only for a fixed set of fee constraints. For some pools, 25bps is too high.

### Solution: 

Production constraints are considered minimum fees charged. Specifically `trade_fee_numerator` is now set to `0` to allow protocol users to customize how high fees for the given pool should be. Note: minimum owner fee remains at 5bps of that host can claim up to 1bps


joncinque (approved):
looks good

PR #1597, opened 2021-04-14
stake-pool: Assess fee as a percentage of rewards

### LINKED ISSUE #1521: stake-pool: Change fee model to happen on "update pool balance"
#### Problem

Taking all of the fee upfront on deposit is a simple model, but unnecessarily punishes users who deposit for a short time.  Also, it's totally different from the main staking program's reward model if assessing the commission every epoch.

#### Solution

Adopt a model similar to regular staking, and have the stake pool fee paid out as pool tokens minted every epoch during "update pool balance".  Also, update the documentation to reflect this change.

joncinque:
Fixes #1521 

Instead of minting pool tokens on deposit to the manager, we mint them every epoch as a percentage of rewards perceived.  If there were no rewards or the pool lost value due to slashing, no fee is minted.  

From the associated issue:
> Problem
>
> Taking all of the fee upfront on deposit is a simple model, but unnecessarily punishes users who deposit for a short time. Also, it's totally different from the main staking program's reward model if assessing the commission every epoch.
>
> Solution
>
> Adopt a model similar to regular staking, and have the stake pool fee paid out as pool tokens minted every epoch during "update pool balance". Also, update the documentation to reflect this change.

### DIFF
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -41,7 +41,7 @@ pub enum StakePoolInstruction {
     ///   8. `[]` Rent sysvar
     ///   9. `[]` Token program id
     Initialize {
-        /// Deposit fee assessed
+        /// Fee assessed as percentage of perceived rewards
         #[allow(dead_code)] // but it's not
         fee: Fee,
         /// Maximum expected number of validators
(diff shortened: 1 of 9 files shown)


stake-pool/program/src/instruction.rs
```diff
@@ -41,7 +42,7 @@ pub enum StakePoolInstruction {
     ///   8. `[]` Rent sysvar
     ///   9. `[]` Token program id
     Initialize {
-        /// Deposit fee assessed
+        /// Fee assessed as percentage of perceived rewards
         #[allow(dead_code)] // but it's not
         fee: Fee,
```
mvines, line 47: Totally fine as a follow-up, but I think we'll want an instruction to allow the fee to be changed after initialization

mvines (approved):
lgtm!

stake-pool/program/src/instruction.rs
joncinque: Yeah definitely, I didn't want to do put too much into this PR

luciotato:
You're welcome 😜

PR #1600, opened 2021-04-15
Add naming service program

Lcchy:
Hi, 
here is a short description of a spl-program that would allow issuing, accessing, and managing ownership of: domain names, Solana Pubkeys, URLs, twitter handles, ipfs cid's etc..

The program interface is also given in the readme.
Please let me know if you have any suggestions.


naming/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+pub enum NameStatus {
+    Deleted,
+    Active,
+}
+
+pub type ValueType = u16;
+
+#[derive(BorshDeserialize, BorshSerialize)]
+pub struct NameState {
```
bartosz-lipinski, line 33: Can you include a structure that allows for easy lookup given an account name? I think wallets would have two use-cases:
1. Find address given a string
2. Find name given an address

bartosz-lipinski (commented):
Looks great, can you also include a state that will manage if the top-level domain is open or restricted to authority i.e. twitter names

antoineherzog:
Can we have multiple addresses per name? Is your values an Array of string? Being able to prove that you own your list of public addresses can be important for NFT use-cases.

naming/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+Program Interface:
+
+```
+pub enum NameStatus {
+    Deleted,
+    Active,
+}
+
+pub type ValueType = u16;
```
bhgames, line 30: wouldn't this be better off as an enum?

naming/README.md
```diff
@@ -0,0 +1,93 @@
...
+    },
+    /// Change the owner of the name.
+    /// If the owner of the given name is itself a name, the program will follow the given list of parents
+    /// until finding a pubkey owner.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   1. `[writable]` The program account derived from the name
+    ///   2. `[signer]` The current owner of the program account derived from the name
+    ///   3...N `[]` The list of accounts needed to find an owner which is a pubkey or a null
```
bhgames, line 79: what happens if this blows out the 1200-something byte transaction size limit? Very realistic if you have a long enough chain. Either we place a maximum limit on the amount of nesting allowed or you do something where you actually have to walk up the tree on the client and provide the final parent here. Question is how do you prove it's the final parent? Not sure on that one. Anyway, something to think about. There is an interesting problem to solve here that warrants more discussion.

naming/README.md
```diff
@@ -0,0 +1,93 @@
...
+    Deleted,
+    Active,
+}
+
+pub type ValueType = u16;
+
+#[derive(BorshDeserialize, BorshSerialize)]
+pub struct NameState {
+    status: u8,
+    values: Vec<(ValueType, String)>,
```
bhgames, line 35: Here's a question - is there a constraint on having multiple of the same ValueType in this vec? Like what if there are two SOL addresses in here, or two twitter handles? Is it up to the client reading it to interpret that how they will?

mvines:
Thanks for getting this started @Lcchy! I've played with this idea a little and have a sketch of a slight variation.  Here's what the instructions and account state might look like:
```rust

/// The data for a Name Registry account is always prefixed a `NameRecordHeader` structure.
///
/// The layout of the remaining bytes in the account data are determined by the record `class`
#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, BorshSchema, PartialEq)]
pub struct NameRecordHeader {
    // Names are hierarchical.  `parent_name` contains the account address of the parent
    // name, or `Pubkey::default()` if no parent exists.
    pub parent_name: Pubkey,

    // The owner of this name
    pub owner: Pubkey,

    // The class of data this account represents (DNS record, twitter handle, SPL Token name/symbol, etc)
    //
    // If `Pubkey::default()` the data is unspecified.
    pub class: Pubkey,
}

/// Instructions supported by the generic Name Registry program
#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, BorshSchema, PartialEq)]
pub enum NameRegistryInstruction {
    /// Create an empty name record
    ///
    /// The address of the name record (account #1) is a program-derived address with the following
    /// seeds to ensure uniqueness:
    ///     * SHA256(`Create::name`)
    ///     * Account class (account #3)
    ///     * Parent name record address, if present (account #4)
    ///
    /// If this is a child record, the parent record's owner must approve by signing (account #5)
    ///
    ///   0. `[writeable, signer]` Funding account (must be a system account)
    ///   1. `[writeable]` Name record to be created (program-derived address)
    ///   2. `[]` Account owner (written into `NameRecordHeader::owner`)
    ///   3. `[signer]` Account class (written into `NameRecordHeader::class`).
    ///                 If `Pubkey::default()` then the `signer` bit is not required
    ///   4. `[]` Parent name record (written into `NameRecordHeader::parent_name). Optional
    ///   5. `[signer]` Owner of the parent name record. Optional
    ///   6. `[]` System program
    ///
    Create {
        /// Name of the record to create
        name: String,

        /// Number of lamports to fund the name record with
        lamports: u64,

        /// Number of bytes of memory to allocate in addition to the `NameRecordHeader`
        space: u32,
    },

    /// Update the data in a name record
    ///
    ///   * If account class is `Pubkey::default()`:
    ///   0. `[writeable]` Name record to be updated
    ///   1. `[signer]` Account owner
    ///
    ///   * If account class is not `Pubkey::default()`:
    ///   0. `[writeable]` Name record to be updated
    ///   1. `[signer]` Account class
    ///
    Update {
        offset: u32
        data: Vec<u8>,
    }

    /// Transfer ownership of a name record
    ///
    ///
    ///   * If account class is `Pubkey::default()`:
    ///   0. `[writeable]` Name record to be transferred
    ///   1. `[signer]` Account owner
    ///
    ///   * If account class is not `Pubkey::default()`:
    ///   0. `[writeable]` Name record to be transferred
    ///   1. `[signer]` Account owner
    ///   1. `[signer]` Account class
    ///
    Transfer { new_owner: Pubkey },

    /// Delete a name record.
    ///
    /// Any lamports remaining in the name record will be transferred to the refund account (#2)
    ///
    ///   0. `[writeable]` Name record to be transferred
    ///   1. `[signer]` Account owner
    ///   2. `[]` Refund account
    ///
    Delete,
}
```

### The basics
I want the top-level name "michael".  As soon as the name registry exists, I rush to execute `NameRegistryInstruction::Create { name: "michael", ...}` with a `class` of `Pubkey::default()` and set my keypair as the owner (signing with it).  The name record data is whatever I want or even empty.  I have full control over it, and can write data into the record as I please with `NameRegistryInstruction::Update`, transfer it to somebody else with `NameRegistryInstruction::Transfer` or delete it with `NameRegistryInstruction::Delete`.

Also since I own the top-level name "michael", only I can authorize the creation of child records under "michael".

### Authority to modify the record data (aka. `class`es)
The owner of a name record can't always be trusted with the ability to modify it. For example a twitter bot that goes through an off-chain process to link an owner to a twitter handle.  Both the bot and owner need to agree on the handle that will be stored in the record.

### Wallet Name Mapping Sketch

Imagine a WalletAddressMap program that's built on top of the name registry.  The WalletAddressMap program owns the top-level "address-to-name" and "name-to-address" records.  

I want to link the name "michael" and my wallet address "mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN".

I ask WalletAddressMap to create two records with my signature:
1. ["michael", "name-to-address"] - holds my wallet address, "mvines..". 
2. ["mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN", "address-to-name"] - holds my name, "michael"

A name-registry aware user wallet can now:
a. Search for the string "michael" in name records under "name-to-address" over RPC to locate my wallet address of "mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN" 
a. Take the address, "mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN", and resolve the name record under "address-to-name" and read the name "michael" from the account over RPC

If I want to transfer the name "michael" to another user, I need the cooperation of WalletAddressMap to do so. However I can delete the name "michael" without needing WalletAddressMap

naming/README.md
t-nelson: An immutable address to the parent domain might do

naming/README.md
```diff
@@ -0,0 +1,93 @@
...
+    /// Accounts expected by this instruction:
+    ///
+    ///   1. `[writable]` The program account derived from the name
+    ///   2. `[signer]` (Optional) The domain name parent account if the given name is a sub-domain
+    ///   3...N `[]` The list of accounts needed to find an owner which is a pubkey or a null
+    AddName {
+        name: String,
+        values: Vec<(ValueType, String)>,
+        owner_pubkey: Option<Pubkey>,
+        owner_name: Option<String>,
```
t-nelson, line 56: If these are exclusive a single enum fields would make more sense
```rust
enum OwnerKind {
    Account(Pubkey),
    Name(String),
    None,
}
```

Though I'm not sure I see the value in the name variant :thinking:

naming/README.md
```diff
@@ -0,0 +1,93 @@
...
+        owner_address: Pubkey,
+    },
+    /// Delete a name (and receive the allocation rent?)
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   1. `[writable]` The program account derived from the name
+    ///   2. `[signer]` The owner of the program account derived from the name
+    ///   3...N `[]` The list of accounts needed to find an owner which is a pubkey or a null
+    DeleteName { name: String },
```
t-nelson, line 91: What happens to subdomains?

t-nelson (commented):
nifty! :slightly_smiling_face:

naming/README.md
Lcchy: Sounds good! I'll add a (parent, when it applies) name field to the NameState.

naming/README.md
Lcchy: Yes I kept a list of some example types with an enum before, but we thought that hardcoding the type spec into the program once it's deployed could be restrictive. With a u16 it can be freely reinterpreted in the js bindings or other programs and scale better? Let me know what you think

naming/README.md
Lcchy: Yes I completely agree, it is not an easy problem. I thought that limiting the dependency depth is maybe not to worrying as Dns type of trees scale more in breadth but maybe people will come up with a new and cooler use for this. Enforcing a depth limit (of 10 parents accounts?) could be a first way of handling this, I don't see anyway of proving the parent existence without walking up the chain.. maybe we could do this in chunks and hold a temporary value in an account to not lose our progress point?

naming/README.md
Lcchy: Ah yes I wanted to keep this limitation as otherwise you open up to a whole lot of prioritization problems between two values indeed. But if we agree on a natural order of interpretation (maybe just the order in the Vec?) then it sounds feasible.

naming/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+A spl program for issuing and managing ownership of: domain names, Solana Pubkeys, URLs, twitter handles, ipfs cid's etc..
+
+This program can be used for dns, pubkey etc lookups via a browser extension (wip), the goal is to create an easy way to identify Solana public keys with various links.
+
+Definitions: Names are strings that map to records which can hold multiple values.\
+ValueType defines the type (url, uri, solkey, etc..) of the value stored following the specification of the js bindings that will read from the name accounts.
+
+- For each name it's values will be stored in the data of a program account. The address of this account is derived from the name.
+In this account the pubkey of the owner (who can change the values) is also stored.
```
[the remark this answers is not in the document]
Lcchy: Sounds good!

naming/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+- Domain name declarations need to be verified by the owner associated to the parent domain. That owner can also be set to Null, removing the permissions needed for a specific domain name and its sub-domains. The declaration of a new TLD (top-level domain) name needs to be verified by the root TLD owner as it is a sub-domain of it. Therefore the owner of the root domain can administrate the TLD and sub-domain distribution.
+Similarly to DNS, we use `"."` as a root domain name seed. Nevertheless, as we need to protect the domain names from collisions with DNS names, the root tld is written out as the string `".sol"` for url lookups in the browser search bar via the extension.
+For example the name resolution of `something.sol` would use the seeds: `[["something"], ["sol"], ["."]]` (converting the strings to bytes of course) but would be needed to be written as `something.sol.sol` in the browser bar.
```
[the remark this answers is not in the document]
ellttBen: It seems like most of the time only a few bump seeds need to be tested to find a suitable one. This has the caveat that some names will take more compute to resolve than others. We just thought about using the solana-program createProgramAddress implementation as a basis for the spec (highest value resulting in a valid address IIRC)

naming/README.md
Lcchy: I talked about it with @ellttBen and we agreed that adding the `parent_name` field to the NameState means that there can be an arbitrary long chain of sub-domains but all pointing to the same owner. That way having a long chain of successive owners is a border use-case that we don't need to worry about and can limit it to a maximum depth of 10?

naming/README.md
ellttBen: The name variant is useful when changing an authority on one record with implicit propagation to other records which mention its name as their authority. Imagine having authority over a certain domain and seamlessly transferring it and all of its subdomains in one transaction. This allows for larger trustless transactions.

naming/README.md
```diff
@@ -0,0 +1,93 @@
+# Solana Naming Service [WIP]
+
+A spl program for issuing and managing ownership of: domain names, Solana Pubkeys, URLs, twitter handles, ipfs cid's etc..
+
+This program can be used for dns, pubkey etc lookups via a browser extension (wip), the goal is to create an easy way to identify Solana public keys with various links.
+
+Definitions: Names are strings that map to records which can hold multiple values.\
+ValueType defines the type (url, uri, solkey, etc..) of the value stored following the specification of the js bindings that will read from the name accounts.
```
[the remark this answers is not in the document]
ellttBen: This would be a client-side spec. The on-chain program is basically use-case agnostic for better extensibility.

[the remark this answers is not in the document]
Lcchy: Yes they could be anything and will only be interpreted by exterior applications. The specification for this could be defined in the js bindings to keep a greater flexibility after deployment. Let me know if I didn't understand your question!

naming/README.md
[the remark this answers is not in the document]
Lcchy: All right will do!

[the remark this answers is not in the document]
Lcchy: Yes we discussed this also with @ellttBen and couldn't find any solution to this.. For now it seems like the PDA finding will have to be done on chain..

naming/README.md
```diff
@@ -0,0 +1,93 @@
...
+    Active,
+}
+
+pub type ValueType = u16;
+
+#[derive(BorshDeserialize, BorshSerialize)]
+pub struct NameState {
+    status: u8,
+    values: Vec<(ValueType, String)>,
+    owner_pubkey: Option<Pubkey>,
```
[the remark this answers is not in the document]
Lcchy: Sounds good to me!

naming/README.md
Lcchy: Good question, I guess they would be broken until someone manually deletes them.
Either this or DeleteName can delete multiple accounts and walk down a sub-domain chain but this again would be limited by the transaction size.

Lcchy:
@mvines Thanks for the review!

> ```rust
> /// The data for a Name Registry account is always prefixed a `NameRecordHeader` structure.
> ///
> /// The layout of the remaining bytes in the account data are determined by the record `class`
> #[derive(Clone, Debug, BorshSerialize, BorshDeserialize, BorshSchema, PartialEq)]
> pub struct NameRecordHeader {
>     // Names are hierarchical.  `parent_name` contains the account address of the parent
>     // name, or `Pubkey::default()` if no parent exists.
>     pub parent_name: Pubkey,
> 
>     // The owner of this name
>     pub owner: Pubkey,
> 
>     // The class of data this account represents (DNS record, twitter handle, SPL Token name/symbol, etc)
>     //
>     // If `Pubkey::default()` the data is unspecified.
>     pub class: Pubkey,
> }
> ```

I will add the parent_name field to the NameRecord, sounds good. For the class field I think that this would already be partially given by ValueType. The right to change or not an entry is given to the MetaData authority (the owner in my first draft). For the twitter handle authority this can then be handled exactly like in your WalletNameMapping example.

> ```rust
>     /// The address of the name record (account #1) is a program-derived address with the following
>     /// seeds to ensure uniqueness:
>     ///     * SHA256(`Create::name`)
>     ///     * Account class (account #3)
>     ///     * Parent name record address, if present (account #4)
> ```

Will do, thanks!

Lcchy:
I just pushed the modifications to the README, let me know what you think!

antoineherzog:
Hi Guys!
I believe the readme is missing a list of use-cases. Which use-case are you trying to achieve? :
- Decentralized DNS like unstoppable domain
- a decentralized username for Solana Users to:
 --> receive crypto-payments only of Solana coin? or multiple coins from different blockchains
--> declare their twitter identity
- a decentralized username to claim some ownership around public addresses (specially usefull for NFTs creator)
- attach identity right and being able to condition asset transfer based on some specific conditions

I would be happy to help or give more feedbacks as I have been CEO of blockchain name service business with starname.

Each use-case is interesting and have some different requirements. At least have a clear picture of which use-cases do the Solana community want to develop is important I believe.

We could organize with the Solana community on this topics with wallet providers etc.. for example.

Lcchy:
Hi Antoine, thanks for your feedback! 

I believe that with the program draft you would be able to cover most of those use cases, but please let me know if I am missing something:
- Decentralized DNS: declare "mywebsite.sol" as a name and write "myrealwebsiteURL.com" as a value for it. When writing "mywebsite.sol" in the browser bar the browser extension will redirect you to the real URL
- Decentralized Username: Solana could open a TLD called "sol-user" or something and let people create their usernames name accounts in which they map to any values (solana (or other blockchain) pubkeys, urls, twitter handle etc..). As we plan on integrating the extension to a wallet, you could use it to transfer funds to a username.
- Verified username and ownership: As in the readme, use a twitter-scraping-bot which controls issuing of sub-domains of "verified-twitter" for example. The claimed pubkey then has the right to write values into this account and therefore can map to anything it wants.
-  I didn't get your last point, would this be a degree (strength) of identity verification?

I also thought of a possible integration with [keybase](https://keybase.io/) which could be nice, it's a sort of identity claiming network where you can share public keys.
But anything is possible, I would be happy to try and make other use cases work with this!

naming/README.md
t-nelson: Could keep the reservations in a separate account on chain.  That's another key to protect, but does offer more strict guarantees

naming/README.md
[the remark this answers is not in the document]
t-nelson: Maybe we can con @jackcmay into a syscall to do it in native code :wink:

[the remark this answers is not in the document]
mvines: `Pubkey::find_program_address()` is already a syscall, so we're good here.  Picking the first available bump seed is an approach we take elsewhere already (like ATA) so no issues with using that here as well IMO

mvines:
> I will add the parent_name field to the NameRecord, sounds good. For the class field I think that this would already be partially given by ValueType. 

The `ValueType: u16` approach feels too limiting to me.  For example we will run into conflicts when two independent groups decide that a value type of 42 means different things, and now we're back to needing a centralized registry of types.  By using the full Solana address space to define the `ValueType`, what I called a `class`, we avoid this as well as introducing the ability to easily connect a ValueType/class to an authority (whomever can sign with that Solana address, be it a system account or program)

antoineherzog:
Regarding decentralized username

- You need to make sure that the owner of the "sol-user" has no right on any username. So the owner of a username is the only one to be able to edit the metadata. Is this the case in your current program?

- for decentralized username, it would great to pick a specific format before the username as convention like **~antoine** so any wallet partner can recognize the format and resolve to a specific address. 

- transfer of ownership needs to be possible and ideally we should be able plug a contract to have a secondary market to enable FOMO on SOL username.

- Also, we can only map Solana address right now? It would great if you map any kind of asset (not only Solana Asset) to an address following either:
the CAIP-19 specification (https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md) CAIP are already used by wallet connect or the slip44 specification or the trust wallet universal asset id spec.

Another approach would be to use the username has a starting point to link to a DID. There are some did provider like https://www.ceramic.network/ or https://aleph.im/ This way, we can add much more type definition and certificate directly to ceramic network. The username is just a pointer to a much more complete data ressources. This approach make your program simpler and you rely an DID partners to store the information.

Regarding certificate, I think there is a really interesting use-case to solve this problem:
https://twitter.com/laufman/status/1370715624058544128?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E1370715624058544128%7Ctwgr%5E%7Ctwcon%5Es1_c10&ref_url=https%3A%2F%2Fwww.theverge.com%2F2021%2F3%2F20%2F22334527%2Fnft-scams-artists-opensea-rarible-marble-cards-fraud-art

Right now, it is very easy to sell a fake NFT and it is very difficult to prove you are the creator of the NFTs. By attaching certificates a signed message for each pubkey you owned to your username, anyone can verify the list of pubkey that a username controls.

My last point was more about regulated finance, most of the security token cannot move freely if the user doesn't have specific properties. So you could attach a KYC information off-chain or on-chain? to a username and you could move your asset only if the username has specific information attached.

naming/README.md
t-nelson: > The name variant is useful when changing an authority on one record with implicit propagation to other records which mention its name as their authority. Imagine having authority over a certain domain and seamlessly transferring it and all of its subdomains in one transaction. This allows for larger trustless transactions.

It seems like this can be done just as easily with by querying the program accounts by authority from the client side (similar to `getTokenAccountsByOwner` for SPL Token

mvines:
Great input @antoineherzog!  I view the naming service as the primate that enables higher-level constructs like what you describe.  

I view this service has merely buckets of name to value mappings that are divided into unique domains under some authority.  Authorities are expressed as a Solana address and could be backed by a person, program, or nothing.   
Name records are owned by somebody, and they can be transferred to somebody else (in some cases authority approval is required for the transfer).

All the Solana wallet addresses are under one authority for example. The authorities are well-known and documented. But if somebody doesn't like the existing Solana wallet address domain or authority, they could create their own

antoineherzog:
One more point,

I think it would be a mistake to replicate ENS in terms of design. For example, recursivity of DNS doesn't add lot of value for decentralized username system and add lots of complexity into the system.

To be honest, we should focus on a primary usecase we want to do well and focus on it.

I tend to think that it is better to have several program like decentralized DNS or decentralized username in two different program instead of doing a one size fit all maybe? 

I tend to think as well that decentralized DNS add very low value in terms of use-case as 99.999% are currently using traditionnal DNS to view webpage and works pretty well. Personnally I would focus only on decentralized username. Just my personal view on this topic.

Lcchy:
Thanks for the feedback! The applications you mention @antoineherzog seem very interesting, like @mvines I also think that these extensions could be built on the naming service as seperate programs in order to keep the modularity. 
For the username vs Dns question I hope that the design will permit both kind of usages, as you could associate a url and a pubkey to any domain name or a username. 
I am going to push a first version of the program to the PR tomorrow, let me know what you think!

Lcchy:
All right I just pushed the first version :tada: Here are some bullet points about the implementation:
- The name address is derived from the following: HASH256(HASH_PREFIX, name). The parent account address isn't always existing so I did not include it. Let me know if you that this is not sufficient.
- I used a `ValueType = Pubkey`, thanks for the suggestion @mvines. This should indeed prevent some common values collisions but it also means that different applications that build on top of this program will each use their own specifications which makes the cross-application usage more difficult. A centralized spec maintained by Solana would be pretty interesting I think as a standard.
- The use of an Ownerkind that referres to another name implies a bit more complexity in the account handling as I needed to handle the different possible user inputs.
- I wanted to store the values in a `HashMap<ValueType, String>` but the use of std::collections::HashMap gave an ELF missalignment error when using test-bpf. I ended up using a `Vec` but I will switch to the faster HashMap hopefully soon.

Let me know what you think, I am totally open to changes of course

antoineherzog:
I am not familiar yet enought with the rust coding in Solana however I read your code to know which name are valid and which name are not valid.

In case of DNS, the underscore can accepted as subdomain but not as domain.

So the URL _test.antoine.com is valid however

_antoine.com is not

What is the current regex for a valid name in your program?

Lcchy:
@antoineherzog I tried to stay as open as possible for the domain names. That means that the hierarchy is only given by the "." separator, any other character is part of the domain name.
In this case `URL _test.antoine.com` and `_antoine.com` are both valid names.
So really any string (e.g. `"my_name"`) is valid as a name because even if they contain no points, they will be treated as `my_name.` where this trailing point is the root domain name (which is `"."`), in this example `my_name` is a top-level domain name.

antoineherzog:
So basically a Solana username could be ______---1# ?

For safety in the username payment usecase, it might be super confusing.

From the DNS perspective, it is not that great as well as we are diverging from the current DNS spec.

Also are you uppercase sensitive currently?  For example, Twitter is uppercase insensitive

Why not start with only alphanumeric regex like "^[a-z0-9]*$"

Lcchy:
The use of the wider namespace is that future namings of any sort could be handled.
For the on-chain program the names and values don't mean much as it only keeps them in a registry (that is, apart from the dns-like sub-domain hierarchy ordered by the "." symbol).  
The namespace restriction can be done at any level via the issuing_authority that controls which sub-domains that can be issued.

For example, you would first create the top-level domain name `solana-user` (with the authorization of the root) and hold the keys of the corresponding issuing_authority on a server. The same server could host a UI where the user submits his request to create a username.  
You could then verify if the requested username fits a certain wanted pattern in order to sign and send the transaction (or make the client send it).
The username would then look like `username.sol-user` but I think this is not too bad as this could be stripped in any UI.
Another TLD could be `verified-twitter` for which the UI also asks for a link to a tweet containing a message signed with the user key with which the user will be able to change the metadata (or values) that the `<twitter-handle>.verified-twitter` name account will hold.

antoineherzog:
I understand your point and actually I am agree with the simplicity of your solution as we don't need any governance to update a regex of character allowed for the name service itself which is kind of great.

However let me challenge a little bit your design in two points:
- ideally we don't want to have to deal with off-chain signer to on a server owned by an entity (that could be hacked) to allow or not some registration when it comes to enforcing a specific policy. Right now, I think we will need to trust bonfida instead of Solana that you are enforcing some rules regarding regex etc.. You are running technically a program off-chain. It is definitely not ideal for a decentralized username. We should be able to have another sub program on-chain called an issuer for each potential new domain that could be in charge to register or not some sub-domain in its domain on behalf of user. Would it possible in your current solution?  I am not sure it is possible right now as you have having two actors, the issuing authority and the metadata autority. I would make sure that we can have an issuer as program on chain. It opens a lot of possibility even to charge additional fee as the issuer is calling the name service to create a name.

- regarding having a dedicated twitter domain, it don't think it is the right approach. Ideally we want a universal username on Solana where each user can freely decide to verify his twitter account by providing the proof of ownership by posting a dedicated twitter message with a link saying for example: I am ~antoine on Solana name service.

I hope it helps

naming/program/src/state.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#[derive(Debug, PartialEq, BorshDeserialize, BorshSerialize, Clone)]
+pub struct NameRegistryState {
+    pub status: NameStatus,
+    // This is the complete name, including all parents
+    pub name: String,
+    pub metadata_authority: OwnerKind,
+    pub issuing_authority: IssuerKind,
+    pub values: Vec<(ValueType, String)>, // TODO use hashmaps?
```
mvines, line 36: I think we want one "Value" per account, like my design sketch at https://github.com/solana-labs/solana-program-library/pull/1600#issuecomment-820849453

The problem with multiple values is that it makes searching for data over RPC very inefficient as the `getProgramAccounts` [memcmp](https://docs.solana.com/developing/clients/jsonrpc-api#filters) filter can no longer be used because position of the `ValueType` or value data of interest is no longer fixed in the account data.

The value data itself should just be bytes at this level as well.  Depending on the `ValueType`, the data may be a String but not necessarily.

naming/program/src/state.rs
```diff
@@ -0,0 +1,47 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use solana_program::{msg, program_error::ProgramError, pubkey::Pubkey};
+
+#[derive(Debug, PartialEq, BorshDeserialize, BorshSerialize, Clone, Copy)]
+pub enum NameStatus {
+    Deleted,
```
mvines, line 6: What's `NameStatus::Deleted` for?  If the name is deleted then I'd think the account wouldn't exist.  Is this really `NameStatus::Uninitialized`?

naming/program/src/state.rs
```diff
@@ -0,0 +1,47 @@
...
+    Account(Pubkey),
+    Name(String), // Usefull for easy ownership transfer of a whole sub-tree of domain names
+    Frozen,       // Metadata is locked, terminal state
+}
+
+#[derive(Debug, PartialEq, BorshDeserialize, BorshSerialize, Clone)]
+pub struct NameRegistryState {
+    pub status: NameStatus,
+    // This is the complete name, including all parents
+    pub name: String,
```
mvines, line 33: 1. Including the parents in the name seems redundant IDK
2. I don't think we need to include the name at all :).  The name is already known because it was included in the address derivation hash.  For name discovery, I was envisioning a name record would have its own account (one value per account as per my comment below) where the "value data" is the name itself.

naming/program/src/state.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub metadata_authority: OwnerKind,
+    pub issuing_authority: IssuerKind,
```
mvines, line 35: It would be nice to combine these two fields, and get to a single Owner Kind - "Pubkey".  Frozen then becomes `Pubkey::default()`.

mvines:
@Lcchy - hey I feel I'm still pretty far design wise from the latest iteration.  I left some comments but I'm happy to jump on a call sometime if you feel like a more interactive discussion would help

antoineherzog:
@Lcchy I would be happy to jump on a call as well (on my side, more to understand the design decisions at a higher level)

naming/program/src/state.rs
Lcchy: I agree that the values can be stored as bytes directly, I'm going to change it now. I would be happy to discuss the rest with you directly :) I think that this is a core decision we have to take from which the other questions can be at least partially answered.

naming/program/src/state.rs
Lcchy: Ah yes it's more a `Unitialiazed` indeed, I'm going to fix it now.

Lcchy:
I just pushed some javascript bindings for the program and fixed some small things that were suggested by @mvines, thanks.
This is still temporary while we find a final way of storing the values (one per account or all in one account) and handle the different authorities.

Lcchy:
Ok so I just pushed a new (and rewritten) version of the program that fits more closely with the description @mvines gave. This allows for greater flexibility for future usage as the constraints are pretty loose. 
More complex functionality can then be built off- and on-chain with UIs and cross program invocations.
Let me know what you think :)

antoineherzog:
Can a program own a name? can this program invoke the solana name service to create sub-names?  (this way we can have special on-chain logic and criteria of issuances of sub-names for each top level name)

naming/program/src/state.rs
```diff
@@ -0,0 +1,105 @@
...
+pub fn write_data(account: &AccountInfo, input: &[u8], offset: u32) {
+    let mut account_data = account.data.borrow_mut();
+    for i in 0..input.len() {
+        account_data[NameRecordHeader::LEN + offset as usize + i] = input[i];
+    }
+}
+
+////////////////////////////////////////////////////////////
+
+pub const HASH_PREFIX: &str = "solana name service";
```
mvines, line 71: nit: how about
```suggestion
pub const HASH_PREFIX: &str = "SPL Name Service";
```

naming/program/src/state.rs
```diff
@@ -0,0 +1,105 @@
...
+    name_class_opt: Option<&Pubkey>,
+    parent_name_address_opt: Option<&Pubkey>,
+) -> (Pubkey, Vec<u8>) {
+    // let hashed_name: Vec<u8> = hashv(&[(HASH_PREFIX.to_owned() + name).as_bytes()]).0.to_vec();
+    let mut seeds_vec: Vec<u8> = hashed_name;
+
+    let name_class = match name_class_opt {
+        Some(k) => *k,
+        None => Pubkey::default(),
+    };
```
mvines, line 87: I think you can do `let name_class = name_class_opt.cloned().unwrap_or_default()`

naming/program/src/processor.rs
```diff
@@ -0,0 +1,217 @@
...
+            return Err(ProgramError::InvalidArgument);
+        }
+        if *parent_name_account.key != Pubkey::default() && !parent_name_owner.unwrap().is_signer {
+            msg!("The given parent name account owner is not a signer.");
+            return Err(ProgramError::InvalidArgument);
+        }
+
+        if name_account.data.borrow().len() == 0 {
+            // Issue the name registry account
+            let create_account_instruction = create_account(
```
mvines, line 67: So this is annoying but instead of `system_instruction::create_account`, we should instead issue these three in instructions:
* system_instruction::transfer
* system_instruction::allocate
* system_instruction::assign

This is exactly what `create_account` does internally, *except* that `create_account` will fail if a `system_instruction::transfer` has previously been used to send lamports to the account.  This means that I can prevent you from grabbing a name simply by running `solana transfer`

Example at https://github.com/solana-labs/solana-program-library/blob/0f4f2b8de909f1bf356260bde8b401813c3dfdec/associated-token-account/program/src/processor.rs#L58-L98

naming/program/src/processor.rs
```diff
@@ -0,0 +1,217 @@
...
+            && name_account.data.borrow()[0] != (NameStatus::Uninitialized as u8)
+        {
+            msg!("The given name account already exists.");
+            return Err(ProgramError::InvalidArgument);
+        }
+        if *name_class.key != Pubkey::default() && !name_class.is_signer {
+            msg!("The given name class is not a signer.");
+            return Err(ProgramError::InvalidArgument);
+        }
+        if *parent_name_account.key != Pubkey::default() && !parent_name_owner.unwrap().is_signer {
```
mvines, line 60: We also need to deserialize `parent_name_account` and check that the `parent_name_owner` key is really the owner

naming/program/src/processor.rs
```diff
@@ -0,0 +1,217 @@
...
+            msg!("The given name owner is incorrect or not a signer.");
+            return Err(ProgramError::InvalidArgument);
+        }
+
+        // Close the account by transferring the rent sol
+        let source_amount: &mut u64 = &mut name_account.lamports.borrow_mut();
+        let dest_amount: &mut u64 = &mut refund_target.lamports.borrow_mut();
+        *dest_amount += *source_amount;
+        *source_amount = 0;
+
```
mvines, line 179: Let's zero out the full account data here as well and serialize it back into `name_account`.  It's possible that the Delete instruction is called from another program, and that program tries to do something with `name_account` after it's deleted here

mvines (commented):
> Can a program own a name?

Yes it can.

> can this program invoke the solana name service to create sub-names? (this way we can have special on-chain logic and criteria of issuances of sub-names for each top level name)

Yep, I believe it can.

@Lcchy - I only reviewed the Rust program for now, I had a couple comments but it looks good overall!

naming/program/src/state.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn is_initialized(&self) -> bool {
+        self.name_status == NameStatus::Initialized
+    }
+}
+
+pub fn write_data(account: &AccountInfo, input: &[u8], offset: u32) {
+    let mut account_data = account.data.borrow_mut();
+    for i in 0..input.len() {
+        account_data[NameRecordHeader::LEN + offset as usize + i] = input[i];
+    }
```
mvines, line 66: It may be more performant to use https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice here instead of re-implementing memcpy

naming/program/src/processor.rs
Lcchy: Oh I didn't know about this limitation of `create`, thanks!

naming/program/src/processor.rs
Lcchy: Ah yes of course!

naming/program/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "spl-naming"
```
mvines, line 2: WDYT about calling this the:
```suggestion
name = "spl-nam-service"
```
and then moving `naming/*` into `/name-service/*`?

naming/README.md
```diff
@@ -0,0 +1,25 @@
+# Solana Naming Service [WIP]
```
mvines, line 1: I'm game to move this README.md to a `name-service.md` file in https://github.com/solana-labs/solana-program-library/tree/master/docs/src, so it shows up on the docs at spl.solana.com

naming/program/src/error.rs
```diff
@@ -0,0 +1,23 @@
+use num_derive::FromPrimitive;
+use solana_program::{decode_error::DecodeError, program_error::ProgramError};
+use thiserror::Error;
+
+#[derive(Clone, Debug, Eq, Error, FromPrimitive, PartialEq)]
+pub enum NamingError {
```
mvines, line 6:
```suggestion
pub enum NameServiceError {
```

mvines (commented):
I had a couple naming (har har) nits, but beyond that I feel like this PR is ready to land. 

Early this week I'd like to try testing out the name service for some work I'm doing, so it'll be easier if this is merge and we can make incremental changes from that baseline.

naming/program/Cargo.toml
Lcchy: sounds good!

name-service/program/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+[dependencies]
+solana-program = "1.5.6"
+num-traits = "0.2"
+borsh = "0.8.1"
+num-derive = "0.3.3"
+thiserror = "1.0.24"
+
+[dev-dependencies]
+solana-program-test = "1.5.0"
+solana-sdk = "1.5.0"
```
mvines, line 25:
```suggestion
[dependencies]
solana-program = "1.6.6"
num-traits = "0.2"
borsh = "0.8.1"
num-derive = "0.3.3"
thiserror = "1.0.24"

[dev-dependencies]
solana-program-test = "1.6.6"
solana-sdk = "1.6.6"
```

mvines:
lol automerge is too eager!  CI didn't complete yet but I'll keep an eye on it

antoineherzog:
Excited about this PR!!

PR #1601, opened 2021-04-16
ownable and pausable libraries

skotep:
First attempt at Ownable and Pausable libraries for Solana programs.  These libraries provide traits and associated functions for ownership and operation control. 

* An Ownable program is one with a optional owner PublicKey.  The ownership can be initialized, transferred, or renounced.  There is also a function to validate if a program is owned by a given PubllicKey.
* A Pausable program is one with an owner and a paused boolean.  The owner can pause and resume the program.  It is up to the program implementation to respect the paused state.

https://github.com/solana-labs/solana-program-library/blob/master/token-swap/proposals/Pausable.md

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3798,6 +3798,39 @@ dependencies = [
 ]
 
 [[package]]
+name = "spl-ownable"
+version = "0.1.0"
+dependencies = [
+ "arrayref",
+ "byteorder",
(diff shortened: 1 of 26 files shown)


bartosz-lipinski:
@skotep can you add an example of how do you see those two libraries being used with some tests using: program-test or solana-test-validator.

PR #1605, opened 2021-04-17
Questions about the code

luciotato:


stake-pool/program/src/processor.rs
```diff
@@ -686,6 +686,13 @@ impl Processor {
                 }
                 validator_stake_record.last_update_epoch = clock.epoch;
                 validator_stake_record.stake_lamports = **validator_stake_account.lamports.borrow();
+                /*
+                Hi jon! question:
+                validator_stake_account.lamports = total lamports in the account,
+                but some of that is staked and some is rewards. Is that correct?
+                What are you doing with the accumulated rewards?
+                Are those automatically re-staked somehow?
```
joncinque, line 694: Hi Lucio!
The staking rewards are automatically redelegated as they're accrued, so the total lamports amount is all stake (minus the rent-exempt amount).  You can see some more info at https://solana.com/staking#whereare

luciotato: so this documentation is outdated? https://docs.solana.com/cluster/stake-delegation-and-rewards#stakeinstructionwithdrawu64

it reads:
> Lamports build up over time in a Stake account and any excess over activated stake can be withdrawn. The transaction must be signed by the stake's authorized_withdrawer.

joncinque: These are both true.  This documentation doesn't mention that only an inactive stake can be withdrawn from, so you'd have to first deactivate, wait an epoch, and then you can withdraw the lamports.  Or you would first split a certain amount from your active stake, then deactivate, then withdraw.

luciotato: great! thanks!

joncinque:
Closing -- feel free to reach out about any other questions!

PR #1616, opened 2021-04-21
stake-pool: Add depositor key on init, required on deposit

### LINKED ISSUE #1502: stake-pool: Add optional depositor key to pool
#### Problem

Not all stake pools want to be public

#### Solution

Add an optional depositor key, and only allow a deposit if that depositor key has signed the deposit

joncinque:
Some stake pools need to be private, and not allow outside depositors.

Enhance the existing deposit authority in the stake pool be configurable
on initialization, and then require its signature on deposit.

The existing deposit authority is a program address, making deposits
permissionless. This allows a pool creator to set their own deposit_authority on
initialization. In a great turn of events, almost everything else works
the same way!

Here's the current workflow for deposit, where the user calls
stake_program::authorize and stake_pool::deposit in the same
transaction:

* stake_program::authorize assigns staker and withdraw authority to the
  stake pool deposit authority
* stake_pool::deposit
    - uses the deposit authority to assign authority on the deposited
  stake account to the stake pool withdraw authority
    - uses the withdraw authority to merge the deposited stake into the validator stake

The deposit authority must "sign" the transaction in order to reassign
authority to the withdraw authority. Currently, as a program address, it
can just do that. With this change, if the deposit authority is set
during initialization, then that deposit authority must sign the
instruction.

There's also a little update for ease-of-use to always do the
stake_program::authorize in the same transaction as stake_pool::deposit.
This way, in case someone tries to deposit into a forbidden stake pool, the
whole transaction will bail and their stake will stay as theirs.

Fixes #1502

### DIFF
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -32,6 +32,7 @@ pub enum StakePoolInstruction {
     ///   7. `[]` Clock sysvar
     ///   8. `[]` Rent sysvar
     ///   9. `[]` Token program id
+    ///  10. `[s]` (Optional) Deposit authority that must sign all deposits
     Initialize {
         /// Fee assessed as percentage of perceived rewards
         #[allow(dead_code)] // but it's not
@@ -70,13 +71,13 @@ pub enum StakePoolInstruction {
     ///
     ///   0. `[w]` Stake pool
     ///   1. `[s]` Staker
-    ///   2. `[]` Stake pool deposit authority
-    ///   3. `[]` Stake pool withdraw authority
-    ///   4. `[w]` Validator stake list storage account
-    ///   5. `[w]` Stake account to add to the pool, its withdraw authority should be set to stake pool deposit
-    ///   6. `[]` Clock sysvar
-    ///   7. '[]' Sysvar stake history account
-    ///   8. `[]` Stake program
+    ///   2. `[]` Stake pool withdraw authority
+    ///   3. `[w]` Validator stake list storage account
+    ///   4. `[w]` Stake account to add to the pool, its withdraw authority must
+    ///      be set to the staker
+    ///   5. `[]` Clock sysvar
+    ///   6. '[]' Sysvar stake history account
+    ///   7. `[]` Stake program
     AddValidatorToPool,
 
     ///   (Staker only) Removes validator from the pool
@@ -195,7 +196,7 @@ pub enum StakePoolInstruction {
     ///   1. `[w]` Validator stake list storage account
     ///   2. `[]` Stake pool deposit authority
     ///   3. `[]` Stake pool withdraw authority
-    ///   4. `[w]` Stake account to join the pool (withdraw should be set to stake pool deposit)
+    ///   4. `[w]` Stake account to join the pool (withdraw authority should be set to the stake pool deposit authority)
     ///   5. `[w]` Validator stake account for the stake account to be merged with
     ///   6. `[w]` User account to receive pool tokens
     ///   8. `[w]` Pool token mint account
@@ -267,6 +268,7 @@ pub fn initialize(
     pool_mint: &Pubkey,
     manager_pool_account: &Pubkey,
     token_program_id: &Pubkey,
+    deposit_authority: Option<Pubkey>,
     fee: Fee,
     max_validators: u32,
 ) -> Result<Instruction, ProgramError> {
@@ -275,7 +277,7 @@ pub fn initialize(
         max_validators,
     };
     let data = init_data.try_to_vec()?;
-    let accounts = vec![
+    let mut accounts = vec![
         AccountMeta::new(*stake_pool, true),
         AccountMeta::new_readonly(*manager, true),
         AccountMeta::new_readonly(*staker, false),
@@ -287,6 +289,9 @@ pub fn initialize(
         AccountMeta::new_readonly(sysvar::rent::id(), false),
         AccountMeta::new_readonly(*token_program_id, false),
     ];
+    if let Some(deposit_authority) = deposit_authority {
+        accounts.push(AccountMeta::new_readonly(deposit_authority, true));
+    }
     Ok(Instruction {
         program_id: *program_id,
         accounts,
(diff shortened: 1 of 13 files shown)


stake-pool/program/src/instruction.rs
```diff
@@ -30,6 +30,7 @@ pub enum StakePoolInstruction {
     ///   7. `[]` Clock sysvar
     ///   8. `[]` Rent sysvar
     ///   9. `[]` Token program id
+    ///  10. `[s]` (Optional) Deposit authority that must sign all deposits
```
mvines, line 33: The deposit authority shouldn't need to sign `Initialize`.  This'll will save a transaction signature fee for the user

stake-pool/program/src/instruction.rs
```diff
@@ -181,7 +182,7 @@ pub enum StakePoolInstruction {
     ///   1. `[w]` Validator stake list storage account
     ///   2. `[]` Stake pool deposit authority
     ///   3. `[]` Stake pool withdraw authority
-    ///   4. `[w]` Stake account to join the pool (withdraw should be set to stake pool deposit)
+    ///   4. `[w]` Stake account to join the pool (withdraw authority should be set to the stake pool deposit authority)
```
mvines, line 185:
```suggestion
    ///   4. `[w]` Stake account to join the pool (withdraw authority for the stake account should be first set to the stake pool deposit authority)
```

stake-pool/program/src/instruction.rs
mvines: Can you also document here the behaviour when the deposit authority is not provided

stake-pool/program/src/instruction.rs
```diff
@@ -418,24 +421,27 @@ pub fn update_stake_pool_balance(
     })
 }
 
-/// Creates a 'Deposit' instruction.
+/// Creates instructions required to deposit into a stake pool, given a stake
+/// account owned by the user
 pub fn deposit(
```
mvines, line 426: wydt about  `deposit()` and `deposit_private()` functions, and removing `deposit_authority_must_sign: bool` from `deposit()`?  `deposit_private()` implies the deposit authority will sign, `deposit()` doesn't.  Just a thought, I don't feel strongly

stake-pool/program/src/instruction.rs
```diff
@@ -418,24 +421,27 @@ pub fn update_stake_pool_balance(
...
-    stake_pool_withdraw: &Pubkey,
+    stake_pool_deposit_authority: &Pubkey,
+    stake_pool_withdraw_authority: &Pubkey,
     stake_to_join: &Pubkey,
     validator_stake_accont: &Pubkey,
     pool_tokens_to: &Pubkey,
     pool_mint: &Pubkey,
     token_program_id: &Pubkey,
-) -> Result<Instruction, ProgramError> {
+    current_stake_authority: &Pubkey,
```
mvines, line 437: nit: let's place `current_stake_authority` next to `stake_to_join` in the argument list, perhaps:
```
...
deposit_stake_address: &Pubkey,
deposit_stake_withdraw_authority: &Pubkey,
...
```

stake-pool/program/src/processor.rs
```diff
@@ -132,28 +138,86 @@ impl Processor {
         )
     }
 
-    /// Issue a stake_set_manager instruction.
-    #[allow(clippy::too_many_arguments)]
+    /// Issue stake_program::authorize instructions to update both authorities
     fn stake_authorize<'a>(
+        stake_account: AccountInfo<'a>,
+        authority: AccountInfo<'a>,
+        new_staker: &Pubkey,
```
mvines, line 145:
```suggestion
        new_stake_authority: &Pubkey,
```

stake-pool/program/src/processor.rs
```diff
@@ -132,28 +138,86 @@ impl Processor {
         )
     }
 
-    /// Issue a stake_set_manager instruction.
-    #[allow(clippy::too_many_arguments)]
+    /// Issue stake_program::authorize instructions to update both authorities
     fn stake_authorize<'a>(
+        stake_account: AccountInfo<'a>,
+        authority: AccountInfo<'a>,
```
mvines, line 144:
```suggestion
        stake_authority: AccountInfo<'a>,
```

stake-pool/program/src/processor.rs
```diff
@@ -493,6 +567,11 @@ impl Processor {
             &vote_account_address,
         )?;
 
+        if meta.lockup != stake_program::Lockup::default() {
+            msg!("Invalid lockup exists on validator stake account");
```
mvines, line 571:
```suggestion
            msg!("Validator stake account has a lockup");
```

stake-pool/program/src/processor.rs
```diff
@@ -493,6 +567,11 @@ impl Processor {
...
+            return Err(StakePoolError::WrongStakeState.into());
```
mvines, line 572: We could have a more descriptive error here than "Wrong" I think

stake-pool/program/src/instruction.rs
mvines: Or how about this:
1. `deposit()` doesn't have `stake_pool_deposit_authority`, it runs `find_deposit_authority_program_address()` and adds it automatically
2. `deposit_with_authority()` has `stake_pool_deposit_authority`, and assumes `deposit_authority_must_sign = true` 

The cli can then select 1 or 2 depending on whether a `--deposit-authority` was provided to it by the user

mvines (dismissed):
Looks great @joncinque!

stake-pool/program/src/instruction.rs
joncinque: Will do, I wanted to keep things safe to be sure that the depositor was around on creation.

stake-pool/program/src/instruction.rs
joncinque: much clearer, thanks!

stake-pool/program/src/instruction.rs
joncinque: Oh I really like that idea!

stake-pool/program/src/processor.rs
joncinque: Yeah, there are still a few renames to undertake...

stake-pool/program/src/processor.rs
joncinque: Hehe, this message was a bit of future-proofing for when we have lockup-enabled stake pools, but I'll use this message

PR #1617, opened 2021-04-21
stake-pool: Add reserve stake and decrease / increase validator stake instruction

### LINKED ISSUE #1496: stake-pool: Add split into transient account instruction
#### Problem

The manager needs to rebalance the stake pool without taking custody.

#### Solution

Give the manager ability to split an active stake into its transient stake account and deactivate it one instruction.  This fails if there's already a transient stake account, or merges if the transient stake account exists and is deactivating.

### LINKED ISSUE #1497: stake-pool: Manager instruction to move deactivated transient stake
#### Problem

Towards manager rebalancing, the manager needs to move deactivated stake from one transient stake account to another.

#### Solution

Add the instruction

### LINKED ISSUE #1498: stake-pool: Manager instruction to delegate a transient stake account
#### Problem

Managers need to rebalance!

#### Solution

Add a manager instruction to delegate a transient stake account.

joncinque:
#### Problem

The stake pool staker needs to rebalance the pool without taking ownership of stake.

#### Solution

Add an additional reserve stake account, totally owned by the pool, which is used for rebalancing purposes.  Also, add a transient stake account associated with each validator stake account, that can be used to activate / deactivate portions of stake.

Fixes #1496 Fixes #1497 Fixes #1498


keemy:
hey I was doing some stuff and ran into an issue which after banging my head for a while seems to be because in solana version 1.7.x+
https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/src/processor.rs#L972
and  
https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/src/processor.rs#L1126

don't work anymore because they now validate even on 0 transfers
https://github.com/solana-labs/solana/blob/e9c234d89f8e30e1b011f08004f5ab14a07b67e7/runtime/src/system_instruction_processor.rs#L203

not sure the best way to go about fixing this is?

joncinque:
@keemy ah yes, that'll happen, we just have to split up the `create_account` into `allocate` + `assign`, just as it's being done in the new runtime code.  I'll put in a PR for that asap

PR #1621, opened 2021-04-22
SPL Program Metadata 

oJshua:
This program allows for metadata to be associated with programs that can help to ensure the authenticity of the program, provide an IDL (Interface Description Language) specification describing how the program is used, provide the method of data serialization employed by a program, and other important details related to a program (e.g. GitHub repository, website, assets etc.).


mvines:
Yes.   I think we could assume the presence of https://github.com/solana-labs/solana-program-library/pull/1600 as we build this out too, where program metadata is located via the naming service

mvines:
@oJshua - hey is this worth me looking at yet, or is it still a WIP?

oJshua:
@mvines Not yet -- very much WIP. I'm re-adapting it to work with the Name Service program over the next couple days. Stay tuned!

oJshua:
@mvines I'm working on the JavaScript / docs portion first half of the week, but the Rust side could use a look when you have some time.

jstarry:
Hey @oJshua can you clean up the files in this PR? There's a lot that should be gitignore'd

program-metadata/program/src/state.rs
```diff
@@ -0,0 +1,43 @@
...
+#[derive(Clone, BorshSerialize, BorshDeserialize, Debug)]
+pub struct MetadataEntry {
+    pub account_type: AccountType,
+    pub name: String,
+    pub value: String,
+}
+
+#[repr(C)]
+#[derive(Clone, BorshSerialize, BorshDeserialize, Debug)]
+pub struct VersionedIdl {
```
jstarry, line 37: Does the IDL include the program name?

program-metadata/program/src/state.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub account_type: AccountType,
+    pub effective_slot: u64,
```
jstarry, line 39: Will this be used to check if an IDL has been updated for the latest program upgrade?

program-metadata/program/src/state.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub idl_url: String,
+    pub idl_hash: [u8; 32],
+    pub source_url: String,
```
jstarry, line 42: No hash for this?

jstarry, line 42: Is the intent that explorers could use the source to verify the deployed code? We might want to include bpf sdk version here as well as expected bytecode hash?

program-metadata/program/src/processor.rs
```diff
@@ -0,0 +1,449 @@
...
+    let rent_sysvar_info = next_account_info(account_info_iter)?;
+    let name_service_info = next_account_info(account_info_iter)?;
+
+    if name.len() > MAX_NAME_LENGTH {
+        return Err(MetadataError::NameTooLong.into());
+    }
+
+    if value.len() > MAX_VALUE_LENGTH {
+        return Err(MetadataError::ValueTooLong.into());
+    }
```
jstarry, line 102: It would be more flexible to add limit the sum of these two lengths. That way devs could use longer values with shorter names

jstarry (commented):
I think it would be nice if this program could be used for solana native programs too. Have you considered what that would look like?

Also, I think the usage of the program upgrade authority as the metadata authority is going to cause issues. It's technically possible for the interface of a non-upgradeable program to change.

I think it would be more flexible if programs could invoke the metadata program with a special PDA. Any upgradeable programs can add support for this and non-upgradeable programs could add support before disabling upgrades.

We could also have an explicit allow list of authorities which can set metadata for solana native programs and popular pre-existing non-upgradeable bpf programs like serum, spl-token, etc.

program-metadata/js/yarn.lock
```diff
@@ -0,0 +1,3269 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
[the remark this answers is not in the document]
oJshua: Oops, forgot I started this with yarn. Thanks!

oJshua:
Thanks @jstarry! This is helpful feedback. I will catch up on how non-upgradeable programs work real quick.

Just to verify my understanding is correct, a program wanting to use the program metadata program would invoke the metadata program with its special PDA. i.e. the program might implement a SetProgramMetadata instruction that makes the call to program metadata, signing with the pda?

jstarry:
> Just to verify my understanding is correct, a program wanting to use the program metadata program would invoke the metadata program with its special PDA. i.e. the program might implement a SetProgramMetadata instruction that makes the call to program metadata, signing with the pda?

Yes, exactly. Maybe a "special" PDA isn't required, not too sure yet. We could also add a feature to the runtime which allows a program like the metadata program to know which program invoked it.

oJshua:
> We could also add a feature to the runtime which allows a program like the metadata program to know which program invoked it.

I like this idea. I'll finish cleaning this up a little bit and we can discuss further.

PR #1627, opened 2021-04-23
stake-pool: Add ability to withdraw from reserve if no stake available

### LINKED ISSUE #1500: stake-pool: Add ability to withdraw deactivated transient stake accounts
#### Problem

Since users can only withdraw activated stake, there's a possibility for a delinquent pool manager to deactivate all stake into transient stake accounts and leave it there.

#### Solution

If a transient stake account has been deactivated for more than one epoch, allow users to withdraw directly from it.

joncinque:
#### Problem

With the reserve stake account, stakers get flexibility to move stake between validators.  Since stake pool users can only withdraw from active validator stakes up to a certain amount, if a malicious and / or delinquent staker decreases the stake from all of the validators, there's no way for people to get their SOL back.

#### Solution

Allow users to withdraw from the reserve stake account, but *only* if there are no other options available, meaning that all of the validator stake accounts have no withdrawable lamports left.  The error message will specify a possible valid option if someone tries to withdraw from the reserve, but there's still some possible lamports.

Fixes #1500

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -1640,30 +1640,59 @@ impl Processor {
             return Err(StakePoolError::InvalidState.into());
         }
 
-        let (meta, stake) = get_stake_state(stake_split_from)?;
-        let vote_account_address = stake.delegation.voter_pubkey;
-        check_validator_stake_address(
-            program_id,
-            stake_pool_info.key,
-            stake_split_from.key,
-            &vote_account_address,
-        )?;
-
-        let validator_list_item = validator_list
-            .find_mut(&vote_account_address)
-            .ok_or(StakePoolError::ValidatorNotFound)?;
-
         let withdraw_lamports = stake_pool
             .calc_lamports_withdraw_amount(pool_tokens)
             .ok_or(StakePoolError::CalculationFailure)?;
 
-        let required_lamports = minimum_stake_lamports(&meta);
-        let current_lamports = **stake_split_from.lamports.borrow();
-        let remaining_lamports = current_lamports.saturating_sub(withdraw_lamports);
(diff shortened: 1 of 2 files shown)


stake-pool/program/src/processor.rs
```diff
@@ -1640,30 +1640,59 @@ impl Processor {
...
+                let (validator_stake_address, _) = crate::find_stake_program_address(
+                    &program_id,
+                    &withdrawable_entry.vote_account_address,
+                    stake_pool_info.key,
+                );
+                msg!("Error withdrawing from reserve: validator stake account {} has {} lamports available, please use that first.", validator_stake_address, withdrawable_entry.stake_lamports);
+                return Err(StakePoolError::StakeLamportsNotEqualToMinimum.into());
+            }
+
+            // check that reserve has enough (should never fail, but who knows?)
```
mvines, line 1663: I guess this could fail if all the validator stake accounts and the reserve is currently empty because all the pool stake is in the process of warming up or cooling down in transient accounts. 

This is technically still an attack that the pool staker could play, continually warming up/cooling down stake in the pool by issuing a transaction immediately after the new epoch.  As a token holder, you'd just need to race the pool staker and eventually you'll probably beat them!  Sucks if ever happened but it's recoverable eventually...

mvines (approved):
lgtm!

stake-pool/program/src/processor.rs
joncinque: I had this same thought... extending to cover that situation gets much trickier.  We'd need to split the `stake_lamports` being tracked in the validator list into `active` and `transient` stake to tell if we're in that situation without deserializing *all* of the transient stake accounts

PR #1635, opened 2021-04-26
Governance program

SebastianBor:
### Summary:
Governance is a program the chief purpose of which is to control the upgrade of other programs through democratic means.
It can also be used as an authority provider for mints and other forms of access control as well where we may want
a voting population to vote on disbursement of access or funds collectively.

Governance can handle arbitrary executions of code, but it's real power lies in the power to upgrade programs. It does this through executing commands to the bpf-upgradable-loader program.
Bpf-upgradable-loader allows any signer who has Upgrade authority over a Buffer account and the Program account itself
to upgrade it using its Upgrade command.

### Governance accounts:
governance-accounts

### Proposal workflow:
governance-workflow


governance/program/src/state.rs
```diff
@@ -0,0 +1,237 @@
...
+    Voting,
+
+    /// Votes complete, in execution phase
+    Executing,
+
+    /// Completed, can be rebooted
+    Completed,
+
+    /// Deleted
+    Deleted,
```
SebastianBor, line 186: Do we want to keep deleted proposals? If no then I think we don't need Deleted state.

governance/program/src/state.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub admin_mint: Pubkey,
+
+    /// Mint that creates voting tokens of this Proposal
+    pub vote_mint: Pubkey,
+
+    /// Mint that creates evidence of voting YES via token creation
+    pub yes_vote_mint: Pubkey,
+
+    /// Mint that creates evidence of voting NO via token creation
+    pub no_vote_mint: Pubkey,
```
SebastianBor, line 106: Do we need these vote/yes/no mints? 
As I understand users deposit the governance source tokens and get the vote tokens in exchange, but I think we could skip the deposit step all together and just let users vote. This way we wouldn't need these mints and DepositSourceTokens instruction.
What do you think?

governance/program/src/state.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub voter: Pubkey,
+
+    /// How many votes were unspent
+    pub undecided_count: u64,
+
+    /// How many votes were spent yes
+    pub yes_count: u64,
+
+    /// How many votes were spent no
+    pub no_count: u64,
```
SebastianBor, line 217: If we remove these 3 voting mints this can be just enum Option
`pub enum VoteResult {
 Yes(u8),
 No(u8)
}`
I think there is no need to support simultaneous Yes and No votes and no vote will be None

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///   0. `[writable]` Governance voting record account
+    ///                   Can be uninitialized or initialized(if already used once in this proposal)
+    ///                   Must have address with PDA having seed tuple [Governance acct key, proposal key, your voting account key]
+    ///   1. `[writable]` Proposal State account
+    ///   2. `[writable]` Your Vote account.
+    ///   3. `[writable]` Your Yes Vote account
+    ///   4. `[writable]` Your No Vote account
+    ///   5. `[writable]` Vote Mint account
+    ///   6. `[writable]` Yes Vote Mint account
+    ///   7. `[writable]` No Vote Mint account
```
joncinque, line 189: Getting rid of all of these in favor just a source token account would be great!

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///   7. `[]` Clock sysvar.
+    SignProposal,
+
+    /// [Requires Voting tokens]
+    /// Burns voting tokens, indicating you approve and/or disapprove of running this set of transactions. If you tip the consensus,
+    /// then the transactions can begin to be run at their time slots when people click execute. You are then given yes and/or no tokens
+    ///
+    ///   0. `[writable]` Governance voting record account
+    ///                   Can be uninitialized or initialized(if already used once in this proposal)
+    ///                   Must have address with PDA having seed tuple [Governance acct key, proposal key, your voting account key]
```
joncinque, line 182: would the PDA for this use your source token account address then? or would you use the owner of the token account?

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///   12. `[]` Transfer authority
+    ///   13. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_key]
+    ///   14. `[]` Token program account
+    ///   15. `[]` Clock sysvar
+    Vote {
+        /// Yes/No  with amount of votes
+        vote: Vote,
+    },
+
+    /// Executes a command in the Proposal
```
joncinque, line 202: So this would only be possible for a Proposal which is in the Queued state, and is past the `delay_slots`, correct?

Also, we may be missing some of the states outlined.  From what I understand, this instruction would move a proposal from Voting (or Queued) to Defeated, Queued, or Executed.

I don't think we'll need the "Succeeded" state in that case.

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///   1. `[writable]` Proposal State account
+    ///   2. `[]` Program being invoked account
+    ///   3. `[]` Proposal account
+    ///   4. `[]` Governance account
+    ///   5. `[]` Governance program account pub key
+    ///   6. `[]` Clock sysvar
+    ///   7+ Any extra accounts that are part of the instruction, in order
+    Execute,
+
+    /// Creates an empty governance vote record
```
joncinque, line 214: I don't think we'll need this instruction unless we run into a compute limit.  Otherwise, `Vote` can do this work

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///   3. `[writable]` Source Token Holding account for Proposal that will accept the tokens in escrow
+    ///   4. `[writable]` Vote Mint account
+    ///   5. `[]` Proposal account.
+    ///   6. `[]` Transfer authority
+    ///   7. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_key]
+    ///   8. `[]` Token program account
+    DepositSourceTokens {
+        /// How many voting tokens to deposit
+        voting_token_amount: u64,
+    },
```
joncinque, line 241: Seems simpler to not have this :smile:

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///   5. `[writable]` Source Token Holding account owned by the Governance that will has the actual tokens in escrow
+    ///   6. `[writable]` Vote Mint account
+    ///   7. `[writable]` Yes Vote Mint account
+    ///   8. `[writable]` No Vote Mint account
+    ///   9. `[]` Proposal State account
+    ///   10. `[]` Proposal account
+    ///   11. `[]` Transfer authority
+    ///   12. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_key]
+    ///   13. `[]` Token program account
+    WithdrawVotingTokens {
```
joncinque, line 260: So this would be `WithdrawSourceTokens` or `WithdrawGovernanceTokens` for people to retract their vote?

governance/program/src/state.rs
```diff
@@ -0,0 +1,247 @@
...
+    Uninitialized,
+
+    /// 1 - Governance account
+    Governance,
+
+    /// 2 - Proposal account for Governance account. A single Governance account can have multiple Proposal accounts
+    Proposal,
+
+    /// 3 - Proposal voting state account. Every Proposal account has exactly one ProposalState account
+    ProposalState,
```
joncinque, line 33: Let's hope that by removing all of those extra mints, you can actually fit the whole proposal in one struct, rather than splitting it up

governance/program/src/state.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub governance_mint: Pubkey,
+
+    /// Council mint
+    pub council_mint: Option<Pubkey>,
+
+    /// Program ID that is governed by this Governance
+    pub program: Pubkey,
+
+    /// Time limit in slots for proposal to be open to voting
+    pub time_limit: u64,
```
joncinque, line 80: What does this decide? If we don't sign fast enough, the proposal goes into `Expired` before ever going into `Voting`?

governance/program/src/state.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub no_vote_mint: Pubkey,
+
+    /// Used to validate signatory tokens in a round trip transfer
+    pub signatory_validation: Pubkey,
+
+    /// Used to validate admin tokens in a round trip transfer
+    pub admin_validation: Pubkey,
+
+    /// Used to validate voting tokens in a round trip transfer
+    pub vote_validation: Pubkey,
```
joncinque, line 125: We may also be able to get rid of these, since they increase the size of the struct by a lot.  Perhaps we just need to refactor some of the ownership / signature checking in `spl_token`

governance/program/src/state.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub account_type: GovernanceAccountType,
+
+    /// Proposal account
+    pub proposal: Pubkey,
+
+    /// The user who casted this vote
+    pub voter: Pubkey,
+
+    /// How many votes were unspent
+    pub undecided_count: u64,
```
joncinque, line 221: Will this still be needed?

governance/program/src/state.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    /// How many votes were spent yes
+    pub yes_count: u64,
+
+    /// How many votes were spent no
+    pub no_count: u64,
```
joncinque, line 227: Perhaps this can simply reuse the `Vote` type?

joncinque (commented):
This is starting to look really clear, still a few questions about those extra mints and how to handle some of the larger types

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///   0. `[writable]` Governance account. The account pubkey needs to be set to program-derived address (PDA) with the following seeds:
+    ///           1) 'governance' const prefix,
+    ///           2) Governed Program account key
+    ///   1. `[]` Account of the Program governed by this Governance account
+    ///   2. `[writable]` Program Data account of the Program governed by this Governance account
+    ///   3. `[signer]` Current Upgrade Authority account of the Program governed by this Governance account
+    ///   4. `[]` Governance mint that this Governance uses
+    ///   5. `[signer]` Payer
+    ///   6. `[]` System account
+    ///   7. `[]` Bpf_upgrade_loader account
```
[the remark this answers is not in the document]
SebastianBor: This instruction calls` Bpf_upgrade_loader` via CPI to change `upgrade_authority` and CPI `invoke()` needs the called program account

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///
+    ///   0. `[writable]` Proposal State account
+    ///   1. `[writable]` Proposal Transaction account
+    ///   2. `[writable]` Signatory account
+    ///   3. `[writable]` Signatory Validation account
+    ///   5. `[]` Proposal account
+    ///   6. `[]` Transfer Authority
+    ///   7. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_key]
+    ///   9. `[]` Token program account
+    RemoveTransaction,
```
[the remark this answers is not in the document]
SebastianBor: ok

governance/program/src/state.rs
SebastianBor: Ok, I kept it for now but I'll get back to it so see if we can merge the accounts.

governance/program/src/state.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+/// Governance Account
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct Governance {
+    /// Account type
+    pub account_type: GovernanceAccountType,
+
+    /// Voting threshold in % required to tip the vote
+    pub vote_threshold: u8,
```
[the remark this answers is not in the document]
SebastianBor: `vote_threshold` is the precent of the entire pool of governance tokens eligible for vote which must vote to tip the vote either way. 
So, if the threshold is set to 10% then out of 1 million governance tokens the proposal would have to get 100k Yes votes to pass or 100k No votes to be defeated.

governance/program/src/state.rs
```diff
@@ -0,0 +1,247 @@
...
+    pub program: Pubkey,
+
+    /// Time limit in slots for proposal to be open to voting
+    pub time_limit: u64,
+
+    /// Optional name
+    pub name: [u8; MAX_GOVERNANCE_NAME_LENGTH],
+
+    /// Running count of proposals
+    pub proposal_count: u32,
```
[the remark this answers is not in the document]
SebastianBor: Yeah, I was actually toying with using PDAs as a way to have predictable keys used for lookups but was not sure if that was a good idea. Thank you for clarification.

Cargo.lock
```diff
@@ -31,6 +31,17 @@ version = "0.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "739f4a8db6605981345c5654f3a85b056ce52f37a39d34da03f25bf2151ea16e"
 
+[[package]]
```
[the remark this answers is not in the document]
SebastianBor: ok

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// This transaction needs to contain authority to execute the program.
+    ///
+    ///   0. `[writable]` Uninitialized Proposal Transaction account
+    ///   1. `[writable]` Proposal state account
+    ///   2. `[writable]` Signatory account
+    ///   3. `[writable]` Signatory Validation account
+    ///   4. `[]` Proposal account.
+    ///   5. `[]` Governance account
+    ///   6. `[]` Transfer authority
+    ///   7. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_key]
```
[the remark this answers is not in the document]
SebastianBor: You are right, the PDA which signs program upgrades is derived from the Governed Program Id and it's also  the Governance account key.
The PDA in this transaction is the PDA derived from proposal and it owns the Proposal accounts including the Source Token Holding Account which is used to store voters governance tokens.

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///   9. `[]` Token program account
+    AddCustomSingleSignerTransaction {
+        /// Slot during which this will run
+        delay_slots: u64,
+        /// Instruction
+        instruction: [u8; MAX_PROPOSAL_INSTRUCTION_DATA_LENGTH],
+        /// Position in transaction array
+        position: u8,
+        /// Point in instruction array where 0 padding begins - inclusive, index should be where actual instruction ends, not where 0s begin
+        instruction_end_index: u16,
```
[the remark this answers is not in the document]
SebastianBor: Ok

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// Delete Proposal entirely.
+    ///
+    ///   0. `[writable]` Proposal state account pub key.
+    ///   1. `[writable]` Admin account
+    ///   2. `[writable]` Admin Validation account.
+    ///   3. `[]` Proposal account pub key.
+    ///   4. `[]` Transfer authority.
+    ///   5. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_key]
+    ///   6. `[]` Token program account.
+    DeleteProposal,
```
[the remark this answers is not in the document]
SebastianBor: I removed the Deleted state for now and added Cancelled instead. We might want to keep all the proposals for historical purposes but we also have to keep Proposals in case voters still have source tokens deposited in them.

governance/program/src/instruction.rs
SebastianBor: Done

governance/program/src/instruction.rs
SebastianBor: Voter (the owner of the source token account) + Proposal

governance/program/src/instruction.rs
SebastianBor: First Execute instruction would move Proposal into Executing state and the last Execute instruction would move the Proposal into Completed state. 
Proposal enters Succeeded state after the Yes vote which tips it. We don't explicitly queue it after that. It stays in the Succeeded state until somebody executes it. And yes, you can't execute it until the `hold_up_time` expires. 
I also think the Admin of the proposal still should be allowed to Cancel Proposal while it's in Succeeded but `max_voting_time` haven't expired yet. What do you think?

governance/program/src/instruction.rs
SebastianBor: Correct. It was leftover from the old workflow when source tokens where explicitly deposited. I removed it.

governance/program/src/instruction.rs
SebastianBor: Correct. WithdrawSourceTokens it is.

governance/program/src/state.rs
SebastianBor: It's the max time allocated for voting. I renamed it to `max_voting_time` to be self explanatory.

governance/program/src/state.rs
SebastianBor: Agree, I removed the validation accounts

governance/program/src/state.rs
SebastianBor: No, removed

governance/program/src/state.rs
SebastianBor: Yes, Vote it is now

Cargo.toml
```diff
@@ -22,6 +23,7 @@ members = [
   "token/program",
   "utils/cgen",
   "utils/test-client",
+
```
mvines, line 26:
```suggestion
```

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,214 @@
...
+
+/// Instructions supported by the Governance program
+#[derive(Clone)]
+#[allow(clippy::large_enum_variant)]
+pub enum GovernanceInstruction {
+    /// Creates Governance account
+    ///
+    ///   0. `[writable]` Governance account. The account pubkey needs to be set to program-derived address (PDA) with the following seeds:
+    ///           1) 'governance' const prefix,
+    ///           2) Governed Program account key
```
mvines, line 16: "account key" - do you mean an address?

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,214 @@
...
+    ///   7. `[writable]` Initialized Source Token Holding account
+    ///   8. `[]` Source mint account
+    ///   9. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_key]
+    ///   10. '[]` Token program account
+    ///   11. `[]` Rent sysvar
+    InitializeProposal {
+        /// Link to gist explaining proposal
+        description_link: [u8; MAX_PROPOSAL_DESCRIPTION_LINK_LENGTH],
+        /// Name of the proposal
+        name: [u8; MAX_PROPOSAL_NAME_LENGTH],
```
mvines, line 63: How about using the `String` type here?  Passing in these two large arrays means that the `InitializeProposal` instruction data is mostly full of unnecessary zeros.  Also please document the character format (UTF-8 I assume).

governance/program/src/state.rs
```diff
@@ -0,0 +1,225 @@
...
+pub const MAX_GOVERNANCE_NAME_LENGTH: usize = 32;
+
+/// Max length of a proposal description link
+pub const MAX_PROPOSAL_DESCRIPTION_LINK_LENGTH: usize = 200;
+
+/// Max length of a proposal name
+pub const MAX_PROPOSAL_NAME_LENGTH: usize = 32;
+
+/// Max number of transactions allowed for a proposal
+pub const MAX_TRANSACTIONS: usize = 5;
```
mvines, line 15: I think you mean Instructions instead of Transactions.   A program (this program) can issue instructions but not transactions

governance/program/src/state.rs
```diff
@@ -0,0 +1,225 @@
...
+    Governance,
+
+    /// 2 - Proposal account for Governance account. A single Governance account can have multiple Proposal accounts
+    Proposal,
+
+    /// 3 - Proposal voting state account. Every Proposal account has exactly one ProposalState account
+    ProposalState,
+
+    /// 4 - Vote record account for a given Proposal.  Proposal can have 0..n voting records
+    VoteRecord,
```
mvines, line 33: Why not vote with SPL Tokens, like https://spl.solana.com/feature-proposal does?  Feature Proposal Program only contains a Yes address, but here there could be a Yes and No address

governance/program/src/state.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub voting_began_at: u64,
+
+    /// when the Proposal entered draft state
+    pub created_at: u64,
+
+    /// when the Proposal entered completed state, also when execution ended naturally.
+    pub completed_at: u64,
+
+    /// when the Proposal entered deleted state
+    pub deleted_at: u64,
```
mvines, line 152:
```suggestion
    /// When the Proposal ended voting - this will also be when the set was defeated or began executing naturally.
    pub voting_ended_at: Option<Slot>,

    /// When the Proposal began voting
    pub voting_began_at: Option<Slot>,

    /// when the Proposal entered draft state
    pub created_at: Option<Slot>,

    /// when the Proposal entered completed state, also when execution ended naturally.
    pub completed_at: Option<Slot>,

    /// when the Proposal entered deleted state
    pub deleted_at: Option<Slot>,
```

governance/program/src/state.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    /// The number of the transactions already executed
+    pub number_of_executed_transactions: u8,
+
+    /// The number of transactions included in the proposal
+    pub number_of_transactions: u8,
+
+    /// Array of pubkeys pointing at Proposal Transactions, up to 5
+    pub transactions: [Pubkey; MAX_TRANSACTIONS],
```
mvines, line 161: "Instructions"

governance/program/src/state.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub created_at: u64,
+
+    /// when the Proposal entered completed state, also when execution ended naturally.
+    pub completed_at: u64,
+
+    /// when the Proposal entered deleted state
+    pub deleted_at: u64,
+
+    /// The number of the transactions already executed
+    pub number_of_executed_transactions: u8,
```
mvines, line 155: This counter is here because we don't think the all the instructions can be executed at once?

governance/program/src/state.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub token_threshold_to_create_proposal: u8,
+
+    /// Minimum waiting time in slots for a transaction to be executed after proposal is voted on
+    pub min_transaction_hold_up_time: u64,
+
+    /// Governance mint
+    pub governance_mint: Pubkey,
+
+    /// Council mint
+    pub council_mint: Option<Pubkey>,
```
mvines, line 79: I'm a little confused as to what these are?  They are SPL Token mints?  For what purpose?

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,214 @@
...
+    ///   3. `[]` Proposal account.
+    ///   4. `[]` Governance account
+    ///   5. `[]` Transfer authority
+    ///   6. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_key]
+    ///   7. `[]` Governance program account
+    AddCustomSingleSignerTransaction {
+        /// Slot waiting time between vote period ending and this being eligible for execution
+        hold_up_time: u64,
+        /// Instruction
+        instruction: Vec<u8>,
```
mvines, line 111: I'd expect this to be an `Instruction` type, not a `Vec<u8>`?

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,214 @@
...
+    ///   1. `[writable]` Proposal state account
+    ///   2. `[writable]` Signatory account
+    ///   3. `[]` Proposal account.
+    ///   4. `[]` Governance account
+    ///   5. `[]` Transfer authority
+    ///   6. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_key]
+    ///   7. `[]` Governance program account
+    AddCustomSingleSignerTransaction {
+        /// Slot waiting time between vote period ending and this being eligible for execution
+        hold_up_time: u64,
```
mvines, line 109: Why support a per-instruction `hold_up_time`?

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,214 @@
...
+        /// Instruction
+        instruction: Vec<u8>,
+        /// Position in transaction array
+        position: u8,
```
mvines, line 113: How about just supporting a single instruction?   If necessary, that instruction could be to a program that in turn executes N other instructions.

mvines (commented):
I made some initial comments but I'm still a little confused overall TBH and trying to orient myself.  It feels like we're trying to do shoot for something very generic but at the same time specializing parts of the interface for program upgrades, which I find unsettling design wise.

I don't fully grok the Council yet and why we need it.

governance/program/src/instruction.rs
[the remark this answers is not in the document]
joncinque: Ah I see, I was imagining that the current program upgrade authority would reassign the upgrade authority to the governance *outside* of the governance program.  I'm not a huge fan of reassigning the authority in the `CreateGovernance` instruction.  It's not clear that creating a governance should also change the upgrade authority on a program.

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,264 @@
...
+    CreateGovernance {
+        /// Vote threshold in % required to tip the vote
+        vote_threshold: u8,
+
+        /// Minimum slot time-distance from creation of proposal for an instruction to be placed
+        minimum_slot_waiting_period: u64,
+        /// Time limit in slots for proposal to be open to voting
+        time_limit: u64,
+        /// Governance name
+        name: [u8; MAX_GOVERNANCE_NAME_LENGTH],
```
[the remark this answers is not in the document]
joncinque: I meant the `name` field specifically.  Since the governance is created for one program, I don't know what additional useful information a "name" gives.

governance/program/src/state.rs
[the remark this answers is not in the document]
joncinque: Ok, I see what you mean.  I thought this was just the threshold for the proposal to be accepted.  So for your example, the check is:

* did at least 10% of the total pool vote?
* are there more Yes votes than No votes?

is that right?

governance/program/src/instruction.rs
SebastianBor: Renamed to address

governance/program/src/instruction.rs
SebastianBor: Types changed to String. 
Should I still add explicit comment for UTF-8  or we consider it to be implied by using the String type please?

governance/program/src/state.rs
SebastianBor: Changed to instruction.

governance/program/src/state.rs
SebastianBor: Renamed

governance/program/src/instruction.rs
SebastianBor: Type changed to Instruction

governance/program/src/instruction.rs
mvines: It wouldn't hurt to be explicit IMO.  The program implementation can use `from_utf8()`, like the memo program, to validate the UTF8 input string:
https://github.com/solana-labs/solana-program-library/blob/b99c9b375c6895a4a94dadd578c3a142b924a6fe/memo/program/src/processor.rs#L28

governance/program/src/instruction.rs
SebastianBor: The idea behind supporting multiple instructions is to allow a single proposal for multiple instructions which must be spaced in time. 
For example you can have a proposal for disbursement of founds in 3 yearly instalments. In order to achieve that you can create 3 transfer instructions with hold up time 1, 2 and 3 years accordingly.

mvines: Ah, I see. Do we actually have a list of requirements for what this program is trying to meet?

governance/program/src/instruction.rs
SebastianBor: Hold up time per instruction is used to create time locked instructions. For example a set of future cashflows. See more details at https://github.com/solana-labs/solana-program-library/pull/1635#discussion_r625321554

governance/program/src/state.rs
SebastianBor: Correct. The design assumes you can have time locked instructions within a single proposal. See more details at https://github.com/solana-labs/solana-program-library/pull/1635#discussion_r625321554

SebastianBor:
> I made some initial comments but I'm still a little confused overall TBH and trying to orient myself. It feels like we're trying to do shoot for something very generic but at the same time specializing parts of the interface for program upgrades, which I find unsettling design wise.

You are right we are aiming to create all purpose governance program but the prototype was focused on program upgrades. Given that I think we should specialise the parts of the api where it targets program upgrades explicitly. I'm thinking about the following structure:
`Governance` - Top level account aggregating governances for given governance tokens (for example MNGO or COMP)
`ProgramGovernance` - Governance account for program upgrades. `Governance` could have multiple of those for each program within the protocol and also other none program governances in future versions.

Please let me know what you think

SebastianBor:
> I don't fully grok the Council yet and why we need it.

Council is a way to define an optional voting population which is different from the community of  main token holders 


A practical application of this policy may be to have a very large population controlling
major version upgrades of Solana via normal SOL, for instance, but hot fixes be controlled via Council tokens,
of which there may be only 30, and which may be themselves minted and distributed via proposals by the governing population.

governance/program/src/state.rs
SebastianBor: Both `Governance` and `Council` mints are for SPL tokens and either one can be selected to vote on a `Proposal` (I explained `Council` in more details in https://github.com/solana-labs/solana-program-library/pull/1635#issuecomment-831504872)

When voters vote they deposit their governance tokens and the amount of deposited tokens is used as their vote weight. The tokens are stored in `Source Token Holding` account while a proposal is being voted on. 

We could use the feature-proposal like approach but that could make creating proposals very expensive (based on the voting token holders population) because for every `Proposal` we would have to create an account with vote tokens for every community member regardless if they would eventually vote or not.

Please note in the `Governance` program  prototype the source tokens are stored per `Proposal` but I'm actually thinking we should store them per top level `Governance`. It would mean you can deposit your source tokens once and then vote with that weight on any number of proposals.

governance/program/src/state.rs
SebastianBor: Yes, they are both, Governance and Council mint, externally provided SPL token mints for a given community or protocol. For example the Governance mint could be for MNGO

governance/program/src/instruction.rs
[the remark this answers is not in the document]
SebastianBor: Ok, thank you for clarification. 
The name field is used to provide a user friendly description for `Governance`. However given the discussion about specialisation https://github.com/solana-labs/solana-program-library/pull/1635#issuecomment-831502980 I think it should just be program_name unless there is a way to get program name given its address. Is the possible please?

governance/program/src/instruction.rs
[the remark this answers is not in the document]
SebastianBor: In order for `Governance` program to be able to upgrade the governed program we have to transfer its `upgrade_authority` to the `Governance` but the question is when is the best time to do so.

My thinking was that it would be quite elegant that by the fact of creating a Governance for a program you give up your `upgrade_authority` on it at the same time. It also proves that you are program owner and you must sign the transaction.

If you think that's too far fetched and rather happen at later time then at minimum we should check the current upgrade_authority signed the transaction to prevent creating governance for programs you don't own.

governance/program/src/state.rs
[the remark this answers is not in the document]
SebastianBor: I'm afraid my explanation and example was not correct, so let me try again.
`vote_threshold` is the the absolute threshold for the proposal to be accepted out of the entire population of source tokens. 
Let's say we set the threshold at 60% and we have 100k tokens. It means that 60k Yes votes are need to approve proposal and 40k+ would defeat it.
This is why it should always be 50%+ and my 10% example didn't make any sense.

governance/program/src/instruction.rs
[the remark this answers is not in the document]
mvines: > If you think that's too far fetched and rather happen at later time then at minimum we should check the current upgrade_authority signed the transaction to prevent creating governance for programs you don't own.

This way is my vote: I first create a program myself, iterate on it, then it gets popular and I want to subject future upgrades of it to governance.  I could do this by assigning the upgrade authority to the Governance program in the same transaction that initializes the Governance account

mvines:
> > I don't fully grok the Council yet and why we need it.
> 
> Council is a way to define an optional voting population which is different from the community of main token holders
> 
> A practical application of this policy may be to have a very large population controlling
> major version upgrades of Solana via normal SOL, for instance, but hot fixes be controlled via Council tokens,
> of which there may be only 30, and which may be themselves minted and distributed via proposals by the governing population.

Thanks!  The hard coding of a Governance mint and a Council mint doesn't sit well with me yet. For example what if a 3rd body is needed for some applications, or a Council isn't needed.  Instead a Governance could just have 1 or more mints associated with it, and the mint that is used as the Governance mint vs Council mint is determining when a Proposal is created for that Governance account.

governance/program/src/state.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Source Token Holding account
+    pub source_holding: Pubkey,
+
+    /// Source Mint - either governance or council mint from Governance
+    pub source_mint: Pubkey,
+}
+
+/// Proposal state
+#[derive(Clone)]
+pub struct ProposalState {
```
mvines, line 111: Is there a reason why `ProposalState` is not just a part of `Proposal`?

governance/program/src/state.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// when the Proposal entered deleted state
+    pub deleted_at: Option<Slot>,
+
+    /// The number of the instructions already executed
+    pub number_of_executed_instructions: u8,
+
+    /// The number of instructions included in the proposal
+    pub number_of_instructions: u8,
+
+    /// Array of pubkeys pointing at Proposal instructions, up to 5
```
mvines, line 151: These are `CustomSingleSignerInstruction` accounts right?  Maybe clarify the comment

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    /// [Requires Admin token]
+    /// Remove instruction from the Proposal
+    ///
+    ///   0. `[writable]` Proposal State account
+    ///   1. `[writable]` Proposal instruction account
+    ///   2. `[signer]` Admin account
+    ///   3. `[]` Proposal account
+    ///   4. `[]` Proposal Authority account. PDA with seeds: ['governance',proposal_address]
+    RemoveInstruction,
```
mvines, line 118: Do we need RemoveInstruction?  Seems like you'd know what you want to in advance, and would only ever add instructions you really want

governance/program/src/state.rs
```diff
@@ -0,0 +1,219 @@
...
+    Proposal,
+
+    /// 3 - Proposal voting state account. Every Proposal account has exactly one ProposalState account
+    ProposalState,
+
+    /// 4 - Vote record account for a given Proposal.  Proposal can have 0..n voting records
+    VoteRecord,
+
+    /// 5 - Custom Single Signer Instruction account which holds an instruction to execute for Proposal
+    CustomSingleSignerInstruction,
```
mvines, line 27: Wouldn't the instruction here require no signers?  If any signers are required, this means that nobody but the signer can call `Execute` to trigger the instructions once the proposal is accepted I think

governance/program/src/state.rs
mvines: Oh got it.  Thanks

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,203 @@
...
+    ///  If you tip the consensus then the instructions can begin to be run after their hold up time
+    ///
+    ///   0. `[writable]` Governance Vote Record account. Needs to be set with pubkey set to PDA with seeds of the
+    ///                   1) 'governance' const prefix,
+    ///                   2)  Voter account address
+    ///                   3)  Proposal account address     
+    ///   3. `[]` Proposal account
+    ///   1. `[writable]` Proposal State account
+    ///   2. `[]` Source Token Mint account
+    ///   1. `[writable]` User Source Token account to deposit tokens from
```
mvines, line 167: I'm not sure it's necessary to actually move the source tokens into escrow.  The `Vote` instruction could freeze the tokens in the source account directly: https://github.com/solana-labs/solana-program-library/blob/a539ef021f913a1ccc60ffdec41cd2eb2fc7f299/token/program/src/instruction.rs#L228

`WithdrawSourceTokens` then becomes `ReleaseAccount` or something, that can only be executed once the proposal completes or is cancelled.

Is there a use case for why one would want to vote with less than their total tokens (ie, the `u64` within `Vote::Yes` and `Vote::No`)?  I could also do this myself before voting but transferring the tokens I didn't want to use for voting into a separate token account before calling the `Vote` instruction, so it feels like we don't need this extra knob

governance/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Voting Dynamics
+
+Each Proposal that gets created creates a holding account for all source tokens that voters will deposit in order to vote.
+To vote, one needs to deposit the source token for a given Proposal. One may deposit let's say 1 Uniswap and get 1 vote.
+The source tokens must be provided to the Vote instruction and are put into the Source Token Holding account.
+
+Once the vote completes, the source tokens can be withdrawn from the holding account of the Proposal.
+It's also possible to change your mind and withdraw your tokes before voting is completed but in such a case you vote
+won't count towards the final outcome of the Proposal.
```
mschneider, line 76: What happens if two proposals are up for vote at the same time?

governance/program/src/instruction.rs
SebastianBor: - It might be simpler to freeze tokens but it would put an unnecessary restriction on the mints which could be used for Governance because once you set freezing authority to None you can't change it.
- Also I'm thinking that once you get more governance tokens you could deposit more to increase your vote weight and equally if you need to you could take them back at any time. If we froze the source account then any further transfers would be more complicated. The assumption being that if you increase/decrease your staked tokens your vote gets adjusted accordingly in all active proposals.
-  I can't think at the moment about any scenario when you wouldn't want to use all your voting weight to vote for a  proposal. The only possible scenario would be if you wanted to use the tokens for something else 

Given the above I would stay with depositing the governing tokens into Governance

governance/program/src/state.rs
SebastianBor: The instruction would be executed using Governance PDA as signer

governance/program/src/instruction.rs
SebastianBor: The idea is that while the `Proposal` is in `Draft` mode you can keep editing it until you are happy with it. This is why the Proposal editing instructions exist `AddCustomSingleSignerInstruction`, `RemoveInstruction`, `UpdateInstructionHoldUpTime`. Without that if you make a mistake you would have to delete `Proposal`

governance/program/src/state.rs
SebastianBor: In the current version yes, they are `CustomSingleSignerInstruction` but in general case it can be any instruction type we add support for in future versions.

governance/program/src/state.rs
SebastianBor: It's a bit legacy when Proposal was split into `Proposal` and `ProposalState` because the combined account was too big. We have simplified it now and if the size is no longer a problem I would merge it during implementation.

SebastianBor:
> > > I don't fully grok the Council yet and why we need it.
> > 
> > 
> > Council is a way to define an optional voting population which is different from the community of main token holders
> > A practical application of this policy may be to have a very large population controlling
> > major version upgrades of Solana via normal SOL, for instance, but hot fixes be controlled via Council tokens,
> > of which there may be only 30, and which may be themselves minted and distributed via proposals by the governing population.
> 
> Thanks! The hard coding of a Governance mint and a Council mint doesn't sit well with me yet. For example what if a 3rd body is needed for some applications, or a Council isn't needed. Instead a Governance could just have 1 or more mints associated with it, and the mint that is used as the Governance mint vs Council mint is determining when a Proposal is created for that Governance account.

It's in fact like that. Both `Governance` and `Council` mints are provided externally when `Governance` is created with the `Council` being optional. So it's simplified version where you can have 1 or 2 mints and then when you crate a `Proposal` you decide which mint votes. I'm not sure if there is any value in making it 1 or N at the moment.

governance/program/src/instruction.rs
SebastianBor: It was inspired by COMP Governance https://compound.finance/docs/governance

governance/program/src/instruction.rs
SebastianBor: Done

governance/program/src/instruction.rs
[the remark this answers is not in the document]
SebastianBor: Ok

governance/README.md
SebastianBor: We are changing the mechanism of voting based on the PR feedback and README hasn't been updated accordingly yet. In the final version we won't have this restriction.

governance/program/src/instruction.rs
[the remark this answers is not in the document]
joncinque: Hm, seems like we'll need another program just for program metadata in that case! there's no concept of a "name" for a program, except for hardcoded mappings in the explorer.  Considering we have `token-list`, perhaps we should also make a `program-list` or spinoff SPL Program Metadata from the Token Metadata program.  Either way, I would keep this field out of the governance.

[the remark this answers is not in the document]
SebastianBor: ok

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    },
+
+    /// Deposits governing tokens (Governance or Council) to Governance Realm and establishes your voter weight to be used for voting within the Realm.
+    /// Note: If subsequent (top up) deposit is made and there are active votes for the Voter then the weights won't be updated automatically.
+    /// It can be done by relinquishing votes on active Proposals and voting again with the new weight.
+    ///
+    ///  0. `[]` Governance Realm account.
+    ///  1. `[]` Governing Token mint.
+    ///  2. `[writable]` Governing Token Holding account.
+    ///  3. `[writable]` Governing Token Source account.
```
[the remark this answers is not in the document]
mvines: 🤔 or we use token freezing instead like I mentioned in DM?  I'm liking the idea of using token freezing the more I think about it

governance/program/src/state.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// The user who casted this vote
+    pub voter: Pubkey,
+
+    /// Voter's vote Yes/No and amount
+    pub vote: Option<Vote>,
+}
+
+/// Account for an instruction to be executed for Proposal
+#[derive(Clone)]
+pub struct CustomSingleSignerInstruction {
```
mvines, line 227: It's not clear what account signs this instruction.  This should probably be documented here

governance/program/src/state.rs
```diff
@@ -0,0 +1,239 @@
...
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct ProposalVoteRecord {
+    /// Governance account type
+    pub account_type: GovernanceAccountType,
+
+    /// Proposal account
+    pub proposal: Pubkey,
+
+    /// The user who casted this vote
+    pub voter: Pubkey,
```
mvines, line 219: What address is this exactly?  `voter` is really generic, is this the `VoterRecord` account address?  If so, call this field `voter_record`

governance/program/src/instruction.rs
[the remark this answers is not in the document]
mvines: edit: freezing won't work!  So using a Token Delegate account to transfer tokens in to avoid Governance needing to deal with Token Multisigs seems like what we need here then

governance/program/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+edition = "2018"
+
+
+[features]
+no-entrypoint = []
+test-bpf = []
+
+
+[dependencies]
+solana-program = "1.6.2"
```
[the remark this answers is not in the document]
SebastianBor: Sure

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// 1. `[writable]` Governance Realm account.
+    /// 2. `[]` Governance Token Mint.
+    /// 3. `[writable, signer]` Governances Token Holding account.
+    /// 4. `[signer]` Payer.
+    /// 5. `[]` System.
+    /// 6. `[]` SPL Token.
+    /// 7. `[]` Sysvar Rent.
+    /// 8. `[]` Council Token mint - optional.
+    /// 9. `[writable, signer]` Council Token Holding account - optional.
+    CreateGovernanceRealm {
```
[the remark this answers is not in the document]
SebastianBor: GovernanceRealm renamed to Realm

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+
+/// Instructions supported by the Governance program
+#[derive(Clone)]
+#[allow(clippy::large_enum_variant)]
+pub enum GovernanceInstruction {
+    /// Creates Governance Realm account which aggregates governances for given Governance Mint and optional Council Mint.
+    ///
+    /// 1. `[writable]` Governance Realm account.
+    /// 2. `[]` Governance Token Mint.
+    /// 3. `[writable, signer]` Governances Token Holding account.
```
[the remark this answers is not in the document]
SebastianBor: Done

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// 4. `[signer]` Payer.
+    /// 5. `[]` System.
+    /// 6. `[]` SPL Token.
+    /// 7. `[]` Sysvar Rent.
+    /// 8. `[]` Council Token mint - optional.
+    /// 9. `[writable, signer]` Council Token Holding account - optional.
```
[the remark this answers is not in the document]
SebastianBor: Done

governance/program/src/instruction.rs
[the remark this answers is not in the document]
SebastianBor: I added `vote_authority` to split governing token ownership and the right to do actual voting. 
Also added a new instruction SetVoteAuthority to make it possible to change it. 
by default `vote_authority` is set to governing token owner.

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    ///  0. `[]` Governance Realm account.
+    ///  1. `[]` Governing Token mint.
+    ///  2. `[writable]` Governing Token Holding account.
+    ///  3. `[writable]` Governing Token Source account.
+    ///  4. `[writable]` Voter Record account.
+    ///  6. `[]` SPL Token.
+    ///  7. `[]` Sysvar Rent.   
+    WithdrawGoverningTokens {
+        /// Amount of Governing tokens to withdraw. If None then all Voter tokens from Holding Account would be withdrawn.
+        amount: Option<u64>,
```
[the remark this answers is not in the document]
SebastianBor: I'm not sure if there is a case for that. In theory you could want to transfer some of your governing tokens but not all to somebody else.
Since I'm not sure it's needed I removed it for now

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    CreateProgramGovernance {
+        /// Voting threshold in % required to tip the vote.
+        /// It's the percentage of tokens out of the entire pool of governance tokens eligible to vote.
+        vote_threshold: u8,
+
+        /// Minimum waiting time in slots for an instruction to be executed after proposal is voted on.
+        min_instruction_hold_up_time: Slot,
+
+        /// Time limit in slots for proposal to be open for voting.
+        max_voting_time: Slot,
```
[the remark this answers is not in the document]
SebastianBor: Done

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    ///
+    ///   0. `[writable]` Uninitialized Proposal Instruction account.
+    ///   1. `[writable]` Proposal state account.
+    ///   2. `[signer]` Admin account.
+    ///   3. `[]` Proposal account.
+    ///   4. `[]` Governance account.
+    ///   5. `[]` Governance program account.
+    AddCustomSingleSignerInstruction {
+        /// Slot waiting time between vote period ending and this being eligible for execution.
+        hold_up_time: Slot,
```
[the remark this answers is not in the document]
SebastianBor: Done

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// [Requires Admin token]
+    /// Adds an instruction to the Proposal. Max of 5 of any  type. More than 5 will throw error.
+    ///
+    ///   0. `[writable]` Uninitialized Proposal Instruction account.
+    ///   1. `[writable]` Proposal state account.
+    ///   2. `[signer]` Admin account.
+    ///   3. `[]` Proposal account.
+    ///   4. `[]` Governance account.
+    ///   5. `[]` Governance program account.
+    AddCustomSingleSignerInstruction {
```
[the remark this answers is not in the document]
SebastianBor: Removed

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    ///   5. `[signer]` Transfer authority.
+    ///   6. '[]` Token program account.
+    RemoveSignatory,
+
+    /// [Requires Admin token]
+    /// Adds an instruction to the Proposal. Max of 5 of any  type. More than 5 will throw error.
+    ///
+    ///   0. `[writable]` Uninitialized Proposal Instruction account.
+    ///   1. `[writable]` Proposal state account.
+    ///   2. `[signer]` Admin account.
```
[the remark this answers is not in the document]
SebastianBor: `Admin` account defines  ownership of a `Proposal`. When `Proposal` is crated `Proposal Admin Mint` is also created at the same time for the `Proposal` and `Admin Token` minted for the user who created the `Proposal`.
Possession of the  `Admin Token` indicates ownership of the `Proposal` and it must be provided for all `Proposal` editing instructions.

governance/program/src/state.rs
SebastianBor: This instruction will be signed by `Governance` PDA. For example for `ProgramGovernance` the transaction to upgrade program will be signed by `ProgramGovernance` PDA. 
I updated documentation accordingly.

governance/program/src/state.rs
SebastianBor: This is the owner of the deposited governing token. I renamed it to `governing_token_owner` and updated comments to make it clearer.

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    ///
+    ///   0. `[writable]` Initialized Signatory account.
+    ///   1. `[writable]` Initialized Signatory Mint account.
+    ///   2. `[signer]` Admin account.
+    ///   3. `[writable]` Proposal State account.
+    ///   4. `[]` Proposal account.
+    ///   5. '[]` Token program account.
+    AddSignatory,
+
+    /// [Requires Admin token]
```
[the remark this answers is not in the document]
SebastianBor: Admin token is used as ownership of the Proposal. See more details in https://github.com/solana-labs/solana-program-library/pull/1635#discussion_r631717188

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,236 @@
...
+    ///   3. `[writable]` Proposal State account.
+    ///   4. `[]` Proposal account.
+    ///   5. `[signer]` Transfer authority.
+    ///   6. '[]` Token program account.
+    RemoveSignatory,
+
+    /// [Requires Admin token]
+    /// Adds an instruction to the Proposal. Max of 5 of any  type. More than 5 will throw error.
+    ///
+    ///   0. `[writable]` Uninitialized Proposal Instruction account.
```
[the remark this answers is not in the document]
SebastianBor: Yes, I updated to comment to make it straightforward

governance/program/src/state.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// Yes vote
+    Yes(u64),
+
+    /// No vote
+    No(u64),
+}
+
+/// Yes/No Vote
+#[derive(Clone, Debug, PartialEq)]
+pub enum Vote {
```
mvines, line 54: nit: I think this enum now lives in instruction.rs, it's only used by `GovernanceInstruction::Vote`

governance/program/src/state.rs
```diff
@@ -0,0 +1,281 @@
+//! Program accounts
+
```
mvines, line 2: nit: all the types in this file should be `#[repr(C)]` so the state layout is well known for non-Rust clients

governance/program/src/state.rs
```diff
@@ -0,0 +1,281 @@
...
+/// Account PDA seeds: ['governance', realm, token_mint, token_owner ]
+pub struct VoterRecord {
+    /// Governance account type
+    pub account_type: GovernanceAccountType,
+
+    /// The Realm the VoterRecord belongs to
+    pub realm: Pubkey,
+
+    /// The type of the Governing Token the VoteRecord is for
+    pub token_type: GoverningTokenType,
```
mvines, line 96: A third alternative for `GoverningTokenType` could be to expand `GovernanceAccountType::VoterRecord` into two:
* `GovernanceAccountType::GovernanceRecord`
* `GovernanceAccountType::CouncilRecord`

governance/program/src/state.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+/// Governance Realm Account
+/// Account PDA seeds" ['governance', name]
+pub struct Realm {
+    /// Governance account type
+    pub account_type: GovernanceAccountType,
+
+    /// Governance mint
+    pub governance_mint: Pubkey,
```
mvines, line 68: nit: I feel the Governance mint (and all associated naming around it) is slightly confusing, since the entire program is also called Governance.

The Council is a smaller group, so perhaps we can come up with a different term than "Governance" for the entire population of users, some ideas:
* User mint
* Everybody mint
* Population mint
* Masses mint
Not liking any of these so much.  But what do you think?

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,247 @@
...
+
+/// Instructions supported by the Governance program
+#[derive(Clone)]
+#[allow(clippy::large_enum_variant)]
+pub enum GovernanceInstruction {
+    /// Creates Governance Realm account which aggregates governances for given Governance Mint and optional Council Mint
+    ///
+    /// 0. `[writable]` Governance Realm account. PDA seeds:['governance',name]
+    /// 1. `[]` Governance Token Mint
+    /// 2. `[writable]` Governances Token Holding account
```
mvines, line 15: Please document that account 0 must be the set as the SPL Token owner of this holding account, same with the council token mint

governance/program/src/instruction.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// The last Signatory token to be burned moves the state to Voting
+    ///
+    ///   0. `[writable]` Proposal State account
+    ///   1. `[writable]` Signatory account
+    ///   2. `[writable]` Signatory Mint account
+    ///   3. `[]` Proposal account
+    ///   4. `[signer]` Transfer authority
+    ///   5. `[]` Token program account
+    ///   6. `[]` Clock sysvar
+    SignProposal,
```
mvines, line 204: The usage of "Sign" here confused me for a moment.  "Sign" brings me to "signing a transaction" very quickly, in this context, but that's not what's happening here at all.   How about `"ApproveProposal"`?

governance/program/src/state.rs
mvines: Can we just merge the two as a part of this PR so we don't forget?

mvines (commented):
I like the changes since last time!  Beyond the comments I have here, some of which are final nits, looks good to me.

governance/program/src/state.rs
mvines: "User mint" actually doesn't seem so bad.   eg, the users can vote on a proposal or the council can.

governance/program/src/state.rs
SebastianBor: Done

governance/program/src/state.rs
SebastianBor: Moved

governance/program/src/state.rs
SebastianBor: Done

governance/program/src/state.rs
SebastianBor: I stayed with `GoverningTokenType` because it's also used on `Proposal` to indicate which token is used for voting

governance/program/src/state.rs
SebastianBor: Yeah, I also thought it was a bit confusing initially but couldn't find a good alternative. 
'Users mint' seems a bit to vague to me.
But I think I have a good one now. I renamed it to Community Token / Mint. This way we can have either the general large community voting or a smaller council.

governance/program/src/instruction.rs
SebastianBor: Done

governance/program/src/instruction.rs
SebastianBor: I renamed it to SignOff. I think SignOff has stronger implications than just an approval and better describes this action

mvines (dismissed):

christian-2017:
Hey, love the work on governance. I have a few thoughts that I was wondering if you would consider.

For every proposal. Users vote both with their wallet(1 vote) and with their sol(1 vote per sol). 66% of sol in the voting wallets have to vote yes for a proposal. 66% of the number of wallets voting have to vote yes for a proposal go through. **THIS HELPS WITH PEOPLE WHO HAVE STAKE GETTING MORE OF A SAY, WHILE MAINTAINING PURE DEMOCRACY AS WELL**

Additionally, for a proposal to go through, there must be a certain percentage of total SOL/wallets that vote for an improvement proposal in order for it to go through.

Additionally, I think voting should be able to happen without taking your SOL out of staking, so possibly a vote token could be minted when you stake.

I also think an election every few years for a handful of devs to a council would be a good way to establish some checks and balances! Then this council also has to vote 66% for a proposal for it to go through.

In my opinion, with governance, we should try to mimic the democratized governments of the world with the added twist that SOL and wallet number should have value in the voting system. Checks and balances are key!

Let me know what you think! I have never commented on github before, but believe I have some good ideas about a democratized and sustainable governance system.

PR #1643, opened 2021-04-28
token-swap typescript: Move things around, fix typing problems

Arrowana:


token-swap/js/scripts/typegen.sh
```diff
@@ -0,0 +1,19 @@
+set -e
+
+# Generate typescript declarations
+npx tsc -p tsconfig.d.json -d
+
+# Flatten typescript declarations
+npx rollup -c rollup.config.types.js
+
+# Replace export with closing brace for module declaration
+sed -i '$s/export {.*};/}/' lib/index.d.ts
```
Arrowana, line 10: Really we need some scripting to end up with functional module? This comes from web3.js.
I am no js expert so I followed

token-swap/js/client/index.ts
```diff
@@ -96,111 +89,6 @@ export const CurveType = Object.freeze({
  * A program to exchange tokens against a pool of liquidity
  */
 export class TokenSwap {
-  /**
```
joncinque, line 99: Could we keep all of these declarations and not do the `public ....` in the constructor? There's so many variables that it's useful to have these explanations

token-swap/js/scripts/typegen.sh
joncinque: I'm not too familiar with this either, @jstarry is this an ok model to follow for other JS bindings?

token-swap/js/package.json
```diff
@@ -15,21 +15,31 @@
...
   "types": "lib/index.d.ts",
+  "browserslist": [
+    "defaults",
+    "not IE 11",
+    "maintained node versions"
+  ],
   "files": [
     "/lib",
-    "/module.flow.js"
+    "/module.flow.js",
```
joncinque, line 32: Can you add `module.flow.js` to the local .gitignore?

joncinque (commented):
Looks really good for the most part! Just a couple of comments.  I've also kicked off the CI

token-swap/js/client/index.ts
Arrowana: Most are pointless to be honest, since they repeat the variable name.
Look right above the ctor, the less obvious params are already explained there in a comment.

joncinque: Oh right, in that case, could you document all of the parameters above the constructor?

token-swap/js/scripts/typegen.sh
Arrowana: Done

token-swap/js/client/index.ts
Arrowana: Done

token-swap/js/package.json
Arrowana: Done

token-swap/js/scripts/typegen.sh
jstarry: Probably not needed. We needed to do this for `@solana/web3.js` because we needed to make sure that transitive dependency on `@solana/web3.js` with an old `module.d.ts` wouldn't override the typings included in the new version. Because of TypeScript's module resolution algorithm, any types declared in a module will take precedence. I think it's unlikely that `@solana/spl-token-swap` will be used as both a high-level dependency as well as a transitive dependency so I think it's safe to skip this step

Arrowana: I removed the sed magic, however, not sure which part exactly isn't needed.

Type generation is now reduced to:
https://github.com/solana-labs/solana-program-library/blob/9a61839f63786c7c7bb84b43cf7b6300dae7dfd5/token-swap/js/package.json#L41

joncinque:
Thanks so much for the work @Arrowana !

Arrowana:
Thanks for finalizing it! Alright so we can get away with tsc only, that's great

PR #1645, opened 2021-04-29
Add heap storage program to SPL

n00m4d:
## Heap Storage
Heap is a tree-like data structure for storing large ordered sets of data. You can think of a heap as a binary tree with comparable nodes and each node having up to two child nodes. For the ordered heap each parent node should be either larger (max-heap) or smaller (min-heap) than the child node.

This program is an abstract heap implementation, it stores arbitrary 32-byte values which can be interpreted by the entity using this contract. New nodes can be added only to the end of the heap and removed only from the top of the heap. It is also possible to swap adjacent parent-child nodes.

The typical usage of this data structure is to add elements to the heap (like lending obligations for example) and if the node needs to be removed (obligation to be liquidated) it has to be swapped to the top first (too see if there are less healthy obligations to be liquidated before).

### Accounts
This program manages two account types: Heap and Node. Heap is a single account storing data about the heap like its size and authority which can modify the heap. And Node account stores each particular node of the heap.

diagram


bartosz-lipinski:
For context, this data structure could be used to order liquidations on the lending protocol.

ysavchenko:
Talked with @bartosz-lipinski and added contract example using this storage heap.

It has one simple storage account (with a simple integer stored) and 3 instructions (+1 init one): Add, Remove and Sort.

- Add adds account as a heap node
- Remove removes the node, but only if it is in the root
- Sort accepts two accounts (parent/child in the heap) and issues Swap instructions if they are not in incorrect order (comparing integers stored in the example account)

Basically what it only allows removing only account with smallest value (not globally in the list of course, but locally in one heap branch). And this is how it is going to be used in the `token-lending` contract to ensure ordered obligation liquidations.

PR #1652, opened 2021-04-30
stake-pool-cli: Add CI

joncinque:
I'm not sure if this will be useful, but while fixing up the CLI for stake pools, I needed an easy way to get a stake pool setup with lots of validators, so I wrote a script that uses all of the main CLI commands (minus the simple setters).

I plan to use a similar script when testing the stake-pool-o-matic.

We don't have scripts to test any of the CLIs in CI, but this is similar in concept to a lot of the JS integration tests, just done through CLI instead of JS bindings, so I thought we may want to include it.

Builds on #1650 , so you can ignore everything in `stake-pool/cli/src/main.rs`


mvines:
> I plan to use a similar script when testing the stake-pool-o-matic.

Check out this test:
https://github.com/solana-labs/stake-o-matic/blob/7049baae457fdb39939a30a5f02b4808b82139b4/src/stake_pool_v0.rs#L811

I feel like although this kind of integration test is slow, I feel it's better than a manual integration test cobbled together with bash

mvines:
Can you break out the stake-pool/cli/src/main.rs changes?  These look great to go in on their own

jon-chuang:
> > I plan to use a similar script when testing the stake-pool-o-matic.
> 
> Check out this test:
> https://github.com/solana-labs/stake-o-matic/blob/7049baae457fdb39939a30a5f02b4808b82139b4/src/stake_pool_v0.rs#L811
> 
> I feel like although this kind of integration test is slow, it's way better than a manual integration test cobbled together with bash

In what sense is this slow/slower than a bash-based test?

mvines:
> In what sense is this slow/slower than a bash-based test?

I don't think it is, in fact it'll be quicker

joncinque:
> Check out this test:
https://github.com/solana-labs/stake-o-matic/blob/7049baae457fdb39939a30a5f02b4808b82139b4/src/stake_pool_v0.rs#L811

I'll put this to draft and port it over in that case.  Considering the JS tests take a few minutes, and this took 11 minutes on a successful run, I don't want to slow down the CI any more than needed or add any worst practices.

Is there any use for keeping this script around as a learning tool or for stake bot writers to reuse?  Or do you imagine those also using a `TestValidator` based test?

jon-chuang:
What's the test bottleneck? Is it waiting for epochs? It doesn't seem like the tests should take so long.

mvines:
> What's the test bottleneck? Is it waiting for epochs? It doesn't seem like the tests should take so long.

yep, 32 slots per epoch

jon-chuang:
32 slots is about 12.8s. That's not a lot. You'd need 50 epochs to get up to 11 mins.

Could the number of slots be reduced further?

PR #1656, opened 2021-05-03
stake-pool: Stake pool deserializer

lieuzhenghong:
This PR allows successful deserialisation of `StakePool` and `ValidatorList` account data. I wrote Borsh.js schema and Typescript bindings in `schema.ts`, and some helper functions in `index.ts` to try out the bindings. 

For consistency, I've used the same Prettier settings as the one used in `token/`.

**NOTE:** We want to get `borsh-js` to release a new version ASAP (the one that includes the u128/u256/u512 methods by @joncinque), because currently the import in `package.json` relies on having borsh on your local machine, which is really not ideal:

```package.json
"borsh": "file:../../../solana-repos/borsh-js",
```


stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+        pubkey: a.pubkey,
+        account: {
+          data: null,
+          executable: a.account.executable,
+          lamports: a.account.lamports,
+          owner: a.account.owner,
+        },
+      };
+
+      if (a.account.data.length === STAKE_POOL_ACCT_LENGTH) {
```
joncinque, line 55: We shouldn't have to do this, it defeats the whole purpose of a deserializing library like borsh!  I see two possible situations:
* we know the type being deserialized, so we just use it's decode right away
* we really don't know, in which case `AccountType` can actually have a value of `StakePool` / `ValidatorList`. You would just need to omit the account type enum at the start.  We shouldn't really run into that situation though.  If we need to do `getProgramAccounts`, we can use a filter on the first byte, either it equals 1 (StakePool), or 2 (ValidatorList)

stake-pool/js/src/schema.js
```diff
@@ -0,0 +1,126 @@
+import borsh from 'borsh';
```
joncinque, line 1: Same here, why js and ts?

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export class AccountType extends Enum {}
+export class AccountTypeEnum extends Assignable {}
+
+export class StakePoolAccount extends Assignable {
+  accountType: AccountType;
+  manager: PublicKey;
+  staker: PublicKey;
+  depositAuthority: PublicKey;
+  withdrawBumpSeed: number; // what is this? u8 in Rust
```
joncinque, line 50: number's good here

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,167 @@
...
+  validatorList: PublicKey;
+  reserveStake: PublicKey;
+  poolMint: PublicKey;
+  managerFeeAccount: PublicKey;
+  totalStakeLamports: number;
+  poolTokenSupply: number;
+  lastUpdateEpoch: number;
```
joncinque, line 57: These should be BN, since node doesn't support 64-bit ints

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,167 @@
...
+export class ValidatorListAccount extends Assignable {
+  accountType: AccountType;
+  maxValidators: number;
+  validators: [ValidatorStakeInfo];
+}
+export class ValidatorStakeInfo extends Assignable {
+  status: StakeStatus;
+  voteAccountAddress: PublicKey;
+  stakeLamports: number;
+  lastUpdateEpoch: number;
```
joncinque, line 69: Same with these

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,167 @@
...
+    this.enum = '';
+    Object.keys(properties).forEach(key => {
+      this.enum = key;
+    });
+  }
+}
+
+export class Fee extends Assignable {
+  denominator: number;
+  numerator: number;
```
joncinque, line 39: These should be ok as number

joncinque (commented):
The structs themselves look good!  Mainly some comments on the client side

stake-pool/js/src/index.js
```diff
@@ -0,0 +1,64 @@
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
```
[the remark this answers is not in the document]
lieuzhenghong: Yes, the `.js` files aren't needed. These are built by `npx tsc` -- will remove them.

stake-pool/js/src/schema.ts
lieuzhenghong: why is this OK as number?

stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+import solanaWeb3 from '@solana/web3.js';
+
+export class StakePool {
+  /**
+   * Wrapper class for a stake pool.
+   * Each stake pool has a stake pool account and a validator list account.
+   * (Optionally) a stake pool can also have a name and ticker.
+   */
+  name: string;
+  ticker: string;
```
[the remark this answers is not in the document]
lieuzhenghong: These aren't stake pools per se, these are "combined" stake pools that have both stake pool and validator list accounts. Dom wants each stake pool to have a human-readable name and ticker. I can remove them if you'd like?

stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+function decodeSerializedStakePool(
+  serializedStakePool: Buffer,
+  accountType,
+): schema.StakePoolAccount | schema.ValidatorListAccount {
+  return accountType.decode(serializedStakePool);
+}
+
+async function getStakePoolAccounts(
```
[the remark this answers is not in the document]
lieuzhenghong: Yep, I agree. I have written two such functions: `getStakePoolAccount` and `getValidatorListAccount`.

lieuzhenghong:
I've incorporated all of the requested changes from Jon:

- Removed extraneous `.js` files
- Changed schema 
- Stopped using `a.account.data.length` and instead filtered on the first byte to distinguish between `StakePool` and `ValidatorList`
- Wrote individual functions to get a single `StakePool` or `ValidatorList` account

And a few of my own:

- Added two helper interfaces, `decodedStakePool` and `decodedValidatorList`.
- Wrote helper functions to pretty print public keys and decoded accounts
- Cleaned up the code

stake-pool/js/src/index.ts
lieuzhenghong: Fair enough --- love the idea to filter on the first byte. I've done that now.

stake-pool/js/src/schema.ts
joncinque: Although they're both `u64`s underneath, people *most likely* wouldn't use something so big.  To be totally safe, they should be `BN`s too, but it probably isn't important

stake-pool/js/src/index.ts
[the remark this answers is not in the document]
joncinque: Hm, seems like this part should be provided by another service.  Where are the human-readable name / ticker supposed to come from?  If we have a name anywhere, I thought it would be on the mint for the stake pool token using one of the token metadatas.

stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,168 @@
...
+   * @param connection: An active web3js connection.
+   * @param stakePoolProgramAddress: The public key (address) of the StakePool program.
+   */
+  try {
+    let response = await connection.getProgramAccounts(stakePoolProgramAddress);
+
+    const stakePoolAccounts = response.map(a => {
+      let decodedData;
+
+      if (a.account.data.readUInt8() === 1) {
```
joncinque, line 94: For a next phase (v2?), we can do this even better :smile: we should probably have all of our state structs given as:
```rs
pub struct State {
    Uninitialized,
    StakePool(StakePool),
    ValidatorList(ValidatorList),
}
```
And then we can just call deserialize and we're good

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,166 @@
...
+  staker: PublicKey;
+  depositAuthority: PublicKey;
+  withdrawBumpSeed: number;
+  validatorList: PublicKey;
+  reserveStake: PublicKey;
+  poolMint: PublicKey;
+  managerFeeAccount: PublicKey;
+  totalStakeLamports: BN;
+  poolTokenSupply: BN;
+  lastUpdateEpoch: BN;
```
joncinque, line 57: This is missing `Fee`

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,166 @@
...
+    });
+  }
+}
+
+export class Fee extends Assignable {
+  denominator: number;
+  numerator: number;
+}
+
+export class AccountType extends Enum {}
```
joncinque, line 42: This is currently not usable, correct? Can you add the enum variants?

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,166 @@
...
+  maxValidators: number;
+  validators: [ValidatorStakeInfo];
+}
+export class ValidatorStakeInfo extends Assignable {
+  status: StakeStatus;
+  voteAccountAddress: PublicKey;
+  stakeLamports: BN;
+  lastUpdateEpoch: BN;
+}
+export class StakeStatus extends Enum {}
```
joncinque, line 71: Same here, this needs the enum variants

joncinque:
Looking close, just a few more comments! Then we'll need to wait for the borsh stuff to be integrated, let me ping on that side

stake-pool/js/src/schema.ts
lieuzhenghong: Resolved with TypeScript enums, as discussed on Discord: https://discord.com/channels/428295358100013066/766355323580579901/839204220124659783

stake-pool/js/src/schema.ts
lieuzhenghong: Added

stake-pool/js/src/index.ts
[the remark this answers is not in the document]
lieuzhenghong: I think the idea is for stake pool managers to give us their desired stake pool name and ticker. Using the token metadata is a good idea.

lieuzhenghong:
1. Fix `borshjs` import (**blocking**: if the `near` folks don't release a new version, we may have to publish our own package)
2. Write unit and integration tests 
2. Fill in fields in `package.json`

lieuzhenghong:
@joncinque , it looks like near still hasn't released a new version -- what do we have to do?

joncinque:
They actually did release their new version!  And it was integrated with the newest web3.js in https://github.com/solana-labs/solana/pull/17239, so you can go ahead and use all of the newest stuff

lieuzhenghong:
I've upgraded the `borsh` import, wrote tests to verify correctness of `schema.decode`, and filled in fields in `package.json`. I think I'm ready to merge.

stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+export class StakePool {
+  /**
+   * Wrapper class for a stake pool.
+   * Each stake pool has a stake pool account and a validator list account.
+   */
+  stakePool: decodedStakePool;
+  validatorList: decodedValidatorList;
+}
+
+export interface decodedStakePool {
```
joncinque, line 18: I think the type names can be reversed:
```suggestion
export interface StakePoolAccount {
```

stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+  pubkey: solanaWeb3.PublicKey;
+  account: solanaWeb3.AccountInfo<schema.StakePoolAccount>;
+}
+
+export interface decodedValidatorList {
```
joncinque, line 23:
```suggestion
export interface ValidatorListAccount {
```

stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,172 @@
+import * as schema from './schema.js';
+import solanaWeb3 from '@solana/web3.js';
+
+/**
+ * Sample code to demonstrate how to use the JS bindings
+ * Also contains several useful helper functions
+ */
+
+export class StakePool {
```
joncinque, line 9: This can be `StakePoolAccounts`?

stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,172 @@
...
+    } else {
+      console.log(val, sp[val]);
+    }
+  }
+  console.log('Executable?:', account.account.executable);
+  console.log('Lamports:', account.account.lamports);
+  console.log('Owner PubKey:', account.account.owner.toString());
+}
+
+export default function testOnDevnet() {
```
joncinque, line 151: Should this be in the test dir?

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,181 @@
+import {Schema, serialize, deserializeUnchecked} from 'borsh';
+import BN from 'bn.js';
+
+// Class wrapping a plain object
+export abstract class Assignable {
```
joncinque, line 5: You can remove this and use `Struct` https://github.com/solana-labs/solana/blob/ec1a307a7c2618b9bdea7b82a39a36ba194ae0d0/web3.js/src/util/borsh-schema.ts#L5

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  static decode<T extends Assignable>(data: Buffer): T {
+    return deserializeUnchecked(SCHEMA, this, data);
+  }
+}
+
+// Class representing a Rust-compatible enum, since enums are only strings or
+// numbers in pure JS
+export abstract class Enum extends Assignable {
```
joncinque, line 23: You can remove this and use `Enum` from web3.js https://github.com/solana-labs/solana/blob/ec1a307a7c2618b9bdea7b82a39a36ba194ae0d0/web3.js/src/util/borsh-schema.ts#L21

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,181 @@
...
+
+export class AccountTypeEnum extends Assignable {}
+
+export enum AccountTypeKind {
+  Uninitialized = 'Uninitialized',
+  StakePool = 'StakePool',
+  ValidatorList = 'ValidatorList',
+}
+
+export class StakePoolAccount extends Assignable {
```
joncinque, line 52: And this can be renamed to `StakePool`

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,181 @@
...
+  reserveStake: PublicKey;
+  poolMint: PublicKey;
+  managerFeeAccount: PublicKey;
+  totalStakeLamports: BN;
+  poolTokenSupply: BN;
+  lastUpdateEpoch: BN;
+  fee: Fee;
+}
+
+export class ValidatorListAccount extends Assignable {
```
joncinque, line 68: This renamed to `ValidatorList`

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,181 @@
...
+
+export enum StakeStatusKind {
+  Active = 'Active',
+  DeactivatingTransient = 'DeactivatingTransient',
+  ReadyForRemoval = 'ReadyForRemoval',
+}
+
+export class PublicKey extends Assignable {
+  value: BN;
+}
```
joncinque, line 91: This can be removed in favor of the web3.js PublicKey

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,181 @@
...
+
+export const SCHEMA: Schema = constructStakePoolSchema();
```
joncinque, line 93: You can add to `SOLANA_SCHEMA` defined at https://github.com/solana-labs/solana/blob/ec1a307a7c2618b9bdea7b82a39a36ba194ae0d0/web3.js/src/util/borsh-schema.ts#L34

stake-pool/js/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+    "compilerOptions": {
+        "target": "ES6",
+        "module": "ES6",
+        "esModuleInterop": true,
+        "moduleResolution": "Node",
+        "outDir": "dist"
+    },
+    "include": [
+        "src/*.ts"
```
joncinque, line 10:
```suggestion
        "src/**/*.ts"
```

joncinque (commented):
Looking really close, just a few more bits to take care of!

stake-pool/js/src/schema.ts
lieuzhenghong: I need to use `deserializeUnchecked` in the `decode` method or validatorlist won't decode properly. Is it possible to change this in `borsh-schema.ts`?

joncinque: Ah right very good point.  Could you add another function `decodeUnchecked` in `borsh-schema.ts` which calls `deserializeUnchecked` under the hood?  Tag me on the PR and I'll review it as quickly as possible.

lieuzhenghong:
- Used new version of borsh.js in PR https://github.com/solana-labs/solana/pull/17620 (thanks @joncinque)
- Refactored type names
- Add new `testOnDevnet` test
- Upgrade web3js version

```
  schema.decode
    StakePoolAccount
      ✓ should successfully decode StakePoolAccount account data
    ValidatorListAccount
      ✓ should successfully decode ValidatorListAccount account data
      ✓ should successfully decode ValidatorListAccount with nonempty ValidatorInfo

  index.ts/PrettyPrintPubkey
    ✓ should successfully pretty print a pubkey

  testOnDevnet
    ✓ should successfully decode all validators from devnet (476ms)


  5 passing (862ms)
  ```

File: stake-pool/js/package.json [hunk 25]
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "stake-pool-js",
+  "version": "0.0.1",
```
joncinque, line 3: We can probably start on version 0.1.0
```suggestion
  "version": "0.1.0",
```

stake-pool/js/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "description": "JS bindings for the stake pool program",
+  "private": true,
+  "scripts": {
+    "lint": "npx prettier src/*.ts -w",
+    "compile": "tsc",
+    "test": "./node_modules/mocha/bin/mocha -p ./dist"
```
joncinque, line 9: We should be able to just call `mocha`, but this probably requires a `build` step beforehand
```suggestion
    "test": "mocha -p ./dist"
```

stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export async function getStakePoolAccount(
+  connection: solanaWeb3.Connection,
+  stakePoolPubKey: solanaWeb3.PublicKey,
+): Promise<StakePoolAccount> {
+  /**
+   * Retrieves and deserializes a StakePool account using a web3js connection and the stake pool address.
+   * @param connection: An active web3js connection.
+   * @param stakePoolPubKey: The public key (address) of the stake pool account.
+   */
```
joncinque, line 37: I always thought these comments had to be before the code to apply correctly, e.g. 
```
/**
 * my comment
 */
export async function myFunction()
```
Can it also be in this position?

stake-pool/js/src/index.ts
```diff
@@ -0,0 +1,167 @@
...
+  try {
```
joncinque, line 38: we probably don't need to wrap the whole function in a try-catch block

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,147 @@
...
+
+export class StakeStatusEnum extends Struct {}
+
+export enum StakeStatusKind {
+  Active = 'Active',
+  DeactivatingTransient = 'DeactivatingTransient',
+  ReadyForRemoval = 'ReadyForRemoval',
+}
+
+export class PublicKey extends Struct {
```
joncinque, line 57: PublicKey is borsh-compatible, so feel free to remove this and use the web3 version everywhere!

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,147 @@
...
+  value: BN;
+}
+
+export const SCHEMA: Schema = constructStakePoolSchema(SOLANA_SCHEMA);
+
+export function constructStakePoolSchema(SCHEMA: Schema): Schema {
```
joncinque, line 63: nit, all caps variable names are typically globals
```suggestion
export function constructStakePoolSchema(schema: Schema): Schema {
```

stake-pool/js/src/schema.ts
```diff
@@ -0,0 +1,147 @@
...
+  Active = 'Active',
+  DeactivatingTransient = 'DeactivatingTransient',
+  ReadyForRemoval = 'ReadyForRemoval',
+}
+
+export class PublicKey extends Struct {
+  value: BN;
+}
+
+export const SCHEMA: Schema = constructStakePoolSchema(SOLANA_SCHEMA);
```
joncinque, line 61: nit: This is a bit of a strange function, since it's modifying `SOLANA_SCHEMA` in place and then also returning it.  I would say the function should either clone the map, then modify and return, OR just modify it in place.

Take it or leave it, but to me the best bet would be to rename the function `addStakePoolSchema` and modify SOLANA_SCHEMA in place.

stake-pool/js/src/test.ts
```diff
@@ -0,0 +1,194 @@
...
+import BN from 'bn.js';
+import assert from 'assert';
+import {PublicKey, Connection} from '@solana/web3.js';
+
+describe('schema.decode', () => {
+  describe('StakePoolAccount', () => {
+    it('should successfully decode StakePoolAccount account data', () => {
+      const decodedData = schema.StakePool.decode(
+        new Buffer(
+          '014f962536a1b5b3a1f1036599ca7c164d6ad0677fed896f12ecdd09ada2cd23dc4f962536a1b5b3a1f1036599ca7c164d6ad0677fed896f12ecdd09ada2cd23dcbb09705cd25f4ad3b9e2c78e3f62ba3e290f798150499edc4f851a1a45e71159ff1086831390e666d64d18cac8f0c86eb91d799eda647319634280b89548ba0371ee618745ef3ebb93164e66ac3560e934b04e49b89ad38baa3b104284abb1a574003e566772f55676cf243c647fab463afa138cb156d2080062d2957b10bf22870fdbe150fc449db6268eb9214a848a90162d9bd4bb81e722cacb41d3dcb483b706ddf6e1d765a193d9cbe146ceeb79ac1cb485ed5f5b37913a8cf5857eff00a9000000000000000000000000000000007c00000000000000e8030000000000003800000000000000',
```
joncinque, line 12: this test may be slightly brittle... I've since added some fields, which would break these tests, and modifying them is pretty annoying.  Maybe you can test the round-trip logic of creating the type, serializing, and deserializing?

stake-pool/js/src/test.ts
```diff
@@ -0,0 +1,194 @@
...
+    it('should successfully decode ValidatorListAccount account data', () => {
+      const decodedData = schema.ValidatorList.decodeUnchecked(
+        new Buffer(
+          '020a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
```
joncinque, line 105: We should include some data in this to make it test the full deserialization behavior

stake-pool/js/src/test.ts
```diff
@@ -0,0 +1,194 @@
...
+            '99572085579321386496717000324290408927851378839748241098946587626478579848783',
+          ),
+        }),
+      ),
+      '6MfzrQUzB2mozveRWU9a77zMoQzSrYa4Gq46KswjupQB',
+    );
+  });
+});
+
+describe('Integration test', () => {
```
joncinque, line 143: We can keep this for now, but when this package is integrated in the CI, we should move these tests into a different place as a separate suite of tests.  This way, we'll be sure to never break backwards compatibility :smile:

stake-pool/js/webpack.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
joncinque, line 1: Is this used?

joncinque (commented):
Looks really close! Just a few small things to work out, then this should be good to integrate.  Once it's in, then we can add support for the instructions and maybe some end-to-end tests setting up a pool and all that!

stake-pool/js/src/test.ts
lieuzhenghong: Not sure what you mean here: wouldn't adding fields break the entire deserialiser, not just the tests?

stake-pool/js/src/schema.ts
lieuzhenghong: Makes sense: have made the changes in https://github.com/solana-labs/solana-program-library/pull/1656/commits/3e8073bf1f2a560d5a746175cef3c4c1d21510a9

stake-pool/js/webpack.config.js
lieuzhenghong: removed in https://github.com/solana-labs/solana-program-library/pull/1656/commits/0922acc76c05eaea559901e141034d2609db0a56

stake-pool/js/src/test.ts
lieuzhenghong: Done in https://github.com/solana-labs/solana-program-library/pull/1656/commits/01f557a157dcf11837dc097f21df1ce4bd42c23c

stake-pool/js/src/schema.ts
lieuzhenghong: done in https://github.com/solana-labs/solana-program-library/pull/1656/commits/1970f0d7f88969c718bca6ec07829191eec9de6a

stake-pool/js/src/index.ts
lieuzhenghong: done in https://github.com/solana-labs/solana-program-library/pull/1656/commits/87522abe8ad87806d24fd071171b2548c4483e25

stake-pool/js/src/test.ts
lieuzhenghong: done in https://github.com/solana-labs/solana-program-library/pull/1656/commits/02df5acc687bf04bea38c4033cf81513a20c7dcd

stake-pool/js/src/test.ts
lieuzhenghong: solved in https://github.com/solana-labs/solana-program-library/pull/1656/commits/e122e32c348ef0ec449b11e0715e32a17092e9bc

joncinque:
All right, let's get this in!  We can do more tweaks for the updated fields and integrate CI in later work.  Thanks for your contribution!

PR #1676, opened 2021-05-07
lending: Command Line Interface

jordansexton:
# `spl-token-lending` CLI

We recently removed the outdated client setup scripts for creating a lending market and adding the initial reserves. Lending program deployers need a way to do these things. ~We may wish to expand the scope to include depositing, borrowing, and other user instructions.~ Other instructions saved for a follow up PR.

#### Todo

- [x] Create a lending market
- [x] Add a reserve
- [x] Add docs + examples


jordansexton:
@jstarry This is based on the current Pyth branch, so review could wait until I address your feedback and that's ready to merge.

jordansexton:
@jstarry I'll perform a final round of testing of the README instructions tomorrow, but the code is ready for review.

token-lending/cli/README.md
```diff
@@ -0,0 +1,129 @@
...
+   ```
+   This pubkey will be your Program ID.
+
+1. Open `./token-lending/program/src/lib.rs` in your editor. In the line
+   ```rust
+   solana_program::declare_id!("TokenLending11111111111111111111111111111111");
+   ```
+   replace `TokenLending11111111111111111111111111111111` with your Program ID, e.g.:
+   ```rust
+   solana_program::declare_id!("25MruiXVk27KoQPNKpMr44UDPHPCcSpRBfhXcgMDTwqQ");
```
jstarry, line 54: How about deploying a new version of the lending program to devnet and testnet and updating the declared id in source? Then, devs can optionally override with their own deployed lending program with `--lending-program-id` in the cli

token-lending/cli/src/main.rs
```diff
@@ -0,0 +1,700 @@
...
+        .get_minimum_balance_for_rent_exemption(Reserve::LEN)?;
+    let collateral_mint_balance = config
+        .rpc_client
+        .get_minimum_balance_for_rent_exemption(Mint::LEN)?;
+    let collateral_supply_balance = config
+        .rpc_client
+        .get_minimum_balance_for_rent_exemption(Token::LEN)?;
+    let user_collateral_balance = config
+        .rpc_client
+        .get_minimum_balance_for_rent_exemption(Token::LEN)?;
```
jstarry, line 194: no need to refetch here, the value is the same as above

token-lending/cli/src/main.rs
```diff
@@ -0,0 +1,700 @@
...
+                Token::LEN as u64,
+                &spl_token::id(),
+            ),
+        ],
+        Some(&config.fee_payer.pubkey()),
+    );
+
+    let mut init_reserve_transaction = Transaction::new_with_payer(
+        &[
+            approve(
```
jstarry, line 276: Every approve instruction should have a corresponding revoke instruction at the end of the transaction in case some of the delegated amount wasn't transferred

token-lending/cli/src/main.rs
```diff
@@ -0,0 +1,700 @@
...
+                Token::LEN as u64,
+                &spl_token::id(),
+            ),
+            create_account(
+                &config.fee_payer.pubkey(),
+                &liquidity_host_keypair.pubkey(),
+                liquidity_host_balance,
+                Token::LEN as u64,
+                &spl_token::id(),
+            ),
```
jstarry, line 269: Is it necessary to create the host fee account?

token-lending/cli/src/main.rs
```diff
@@ -0,0 +1,700 @@
...
+
+    let total_balance = reserve_balance
+        + collateral_mint_balance
+        + collateral_supply_balance
+        + user_collateral_balance
+        + liquidity_supply_balance
+        + liquidity_fee_receiver_balance
+        + liquidity_host_balance;
+
+    let mut create_accounts_transaction_1 = Transaction::new_with_payer(
```
jstarry, line 213: The necessity of splitting up the transactions like this make me feel that new reserves should create accounts themselves using program derived addresses to avoid needing to stuff so many signatures into a transaction. The reason is that if account creations are in separate transactions from account initialization, anyone can front run the initialization with their own transaction to take ownership of these accounts

token-lending/cli/src/main.rs
```diff
@@ -0,0 +1,700 @@
...
+            ),
+        ],
+        Some(&config.fee_payer.pubkey()),
+    );
+
+    let mut init_reserve_transaction = Transaction::new_with_payer(
+        &[
+            approve(
+                &spl_token::id(),
+                &source_liquidity_pubkey,
```
jstarry, line 278: This requires the owner of `source_liquidity` to sign but I don't see that listed as a signer

token-lending/cli/src/main.rs
```diff
@@ -0,0 +1,700 @@
...
+                    Arg::with_name("lending_market_owner")
+                        .long("owner")
+                        .validator(is_pubkey)
+                        .value_name("PUBKEY")
+                        .takes_value(true)
+                        .required(true)
+                        .help("Owner required to sign when adding reserves to the lending market"),
+                )
+                .arg(
+                    Arg::with_name("oracle_program_id")
```
jstarry, line 410: It would be nice to have a default value for this

token-lending/cli/src/main.rs
```diff
@@ -0,0 +1,700 @@
...
+                        .help("Initial amount of liquidity to deposit into the new reserve"),
+                )
+                .arg(
+                    Arg::with_name("pyth_product")
+                        .long("pyth-product")
+                        .validator(is_pubkey)
+                        .value_name("PUBKEY")
+                        .takes_value(true)
+                        .required(true)
+                        .help("Pyth product account"),
```
jstarry, line 474: Can you add instructions on where a user would be look up which account address to use?

token-lending/cli/src/main.rs
```diff
@@ -0,0 +1,700 @@
...
+                        .long("max-borrow-rate")
+                        .validator(is_parsable::<u8>)
+                        .value_name("PERCENT")
+                        .takes_value(true)
+                        .required(true)
+                        .default_value("30")
+                        .help("Max borrow APY: min <= optimal <= max"),
+                )
+                .arg(
+                    Arg::with_name("borrow_fee_wad")
```
jstarry, line 556: This isn't very user friendly. Can this be passed as a percent instead?

token-lending/cli/src/main.rs
jordansexton: I think this works given the provided instructions because the owner of `source_liquidity` happens to be the same as the fee payer that makes the transfer, and sometimes also the lending market owner. Should the command itself require another signer then?

token-lending/cli/src/main.rs
jordansexton: I agree, it's pretty unwieldy. It also sucks when running these commands if you create them and then your `InitReserve` fails, you just have a bunch of dangling accounts. Good point about the security issue too, I wasn't aware someone else could just initialize accounts you create.

token-lending/cli/src/main.rs
jstarry: Yeah there ends up being lots of overlap. You may want to look at the spl-token-cli to see how it handles signers (it's complex)

token-lending/cli/src/main.rs
jordansexton: Done.

token-lending/cli/src/main.rs
jordansexton: Done.

token-lending/cli/src/main.rs
jordansexton: Removed.

token-lending/cli/src/main.rs
jordansexton: Done.

token-lending/cli/README.md
```diff
@@ -0,0 +1,129 @@
+# SPL Token Lending program command line interface
+
+A basic CLI for initializing lending markets and reserves for SPL Token Lending.
+See https://spl.solana.com/token-lending for more details
+
+## Deploy a lending program on Solana
```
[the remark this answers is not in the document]
jordansexton: Done.

token-lending/cli/README.md
```diff
@@ -0,0 +1,129 @@
...
+   If the deployment doesn't succeed, follow [this guide](https://docs.solana.com/cli/deploy-a-program#resuming-a-failed-deploy) to resume it.
+
+1. Get your pubkey:
+   ```shell
+   solana address -k owner.json
+
+   # F5242QU7NC4jM8yjqqiv8pZRjL9K8229EBTXHBYtSAo5
+   ```
+
+1. Create a lending market, using the pubkey from the previous step as the `owner`:
```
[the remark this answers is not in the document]
jordansexton: Done & done.

token-lending/cli/src/main.rs
jordansexton: Done.

token-lending/cli/README.md
jordansexton: Done.

token-lending/cli/src/main.rs
jordansexton: Is it acceptable to make this change in a followup PR? It affects the program API, CLI, and tests. I'll create an issue if so.

token-lending/cli/src/main.rs
jordansexton: Done.

token-lending/cli/src/main.rs
jordansexton: Done. I did this somewhat simply by just providing additional keypairs rather than worrying about signers that could be external. I'm thinking the latter could be added later.

token-lending/program/src/processor.rs
```diff
@@ -1291,6 +1292,7 @@ fn process_repay_obligation_liquidity(
         msg!("Obligation lending market does not match the lending market provided");
         return Err(LendingError::InvalidAccountInput.into());
     }
+    // @FIXME: why is it necessary to refresh before repay?
```
jordansexton, line 1295: I spotted these during review. I'll follow up in another PR.

token-lending/README.md
```diff
@@ -1,15 +1,114 @@
...
+### On-chain programs
 
 | Cluster | Program Address |
 | --- | --- |
 | Mainnet Beta | [`LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi`](https://explorer.solana.com/address/LendZqTs7gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi) |
-| Testnet | [`LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi`](https://explorer.solana.com/address/LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi?cluster=testnet) |
-| Devnet | [`LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi`](https://explorer.solana.com/address/LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi?cluster=devnet) |
\ No newline at end of file
+| Testnet | [`6TvznH3B2e3p2mbhufNBpgSrLx6UkgvxtVQvopEZ2kuH`](https://explorer.solana.com/address/LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi?cluster=testnet) |
+| Devnet | [`6TvznH3B2e3p2mbhufNBpgSrLx6UkgvxtVQvopEZ2kuH`](https://explorer.solana.com/address/LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi?cluster=devnet) |
```
jordansexton, line 15: TODO: Deploy to testnet, fix the explorer links.

token-lending/README.md
```diff
@@ -1,15 +1,114 @@
...
+   ```
+   This pubkey will be your Program ID.
+
+1. Open `./token-lending/program/src/lib.rs` in your editor. In the line
+   ```rust
+   solana_program::declare_id!("LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi");
+   ```
+   replace the Program ID with yours, e.g.:
+   ```rust
+   solana_program::declare_id!("6TvznH3B2e3p2mbhufNBpgSrLx6UkgvxtVQvopEZ2kuH");
```
jordansexton, line 75: TODO: change this pubkey.

token-lending/README.md
```diff
@@ -1,15 +1,114 @@
...
+   ```
+   If the deployment doesn't succeed, follow [this guide](https://docs.solana.com/cli/deploy-a-program#resuming-a-failed-deploy) to resume it.
+
+1. Wrap some of your SOL as an SPL Token:
+   ```shell
+   spl-token wrap 10 -- owner.json
+
+   # Wrapping 10 SOL into AJ2sgpgj6ZeQazPPiDyTYqN9vbj58QMaZQykB9Sr6XY
+   ```
+   You'll use this for initial reserve liquidity. Note the SPL Token account pubkey (e.g. `CsbAUDhZfPpkv8jCcV9PPQqfBkUVd5kntubhBLLgMLVF`).
```
jordansexton, line 112: TODO: fix this pubkey.

token-lending/cli/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[dependencies]
+clap = "2.33.3"
+solana-clap-utils = "1.6.6"
+solana-cli-config = "1.6.6"
+solana-client = "1.6.6"
+solana-logger = "1.6.6"
+solana-sdk = "1.6.6"
+solana-program = "1.6.6"
+spl-token-lending = { path="../program", features = [ "no-entrypoint" ] }
+spl-token = { path="../../token/program", features = [ "no-entrypoint" ]  }
```
jordansexton, line 20: TODO: update these deps.

token-lending/cli/README.md
```diff
@@ -0,0 +1,101 @@
...
+  --pyth-product 8yrQMUyJRnCJ72NWwMiPV9dNGw465Z8bKUvnUC8P5L6F \
+  --pyth-price   BdgHsXrH1mXqhdosXavYxZgX6bGqTdj5mh2sxDhF8bJy
+
+# Adding reserve 69BwFhpQBzZfcp9MCj9V8TLvdv9zGfQQPQbb8dUHsaEa
+# Signature: 2yKHnmBSqBpbGdsxW75nnmZMys1bZMbHiczdZitMeQHYdpis4eVhuMWGE29hhgtHpNDjdPj5YVbqkWoAEBw1WaU
+# Signature: 33x8gbn2RkiA5844eCZq151DuVrYTvUoF1bQ5xA3mqkibJZaJja2hj8RoyjKZpZqg2ckcSKMAeqWbMeWC6vAySQS
+# Signature: 3dk79hSgzFhxPrmctYnS5dxRhojfKkDwwLxEda9bTXqVELHSL4ux8au4jwvL8xuraVhaZAmugCn4TA1YCfLM4sVL
+```
+
+Note the reserve pubkey (e.g. `BEnpQRzjn1CN9GJBXyePEga91m8FBqiHoceJ2HoKQW1t`). You'll use this to deposit liquidity, redeem collateral, borrow, repay, and liquidate.
```
jordansexton, line 101: TODO: fix this pubkey.

token-lending/cli/src/main.rs
```diff
@@ -0,0 +1,731 @@
...
+            let liquidation_threshold = value_of(arg_matches, "liquidation_threshold").unwrap();
+            let min_borrow_rate = value_of(arg_matches, "min_borrow_rate").unwrap();
+            let optimal_borrow_rate = value_of(arg_matches, "optimal_borrow_rate").unwrap();
+            let max_borrow_rate = value_of(arg_matches, "max_borrow_rate").unwrap();
+            let borrow_fee = value_of::<f64>(arg_matches, "borrow_fee").unwrap();
+            let flash_loan_fee = value_of::<f64>(arg_matches, "flash_loan_fee").unwrap();
+            let host_fee_percentage = value_of(arg_matches, "host_fee_percentage").unwrap();
+
+            let borrow_fee_wad = (borrow_fee * WAD as f64) as u64;
+            let flash_loan_fee_wad = (flash_loan_fee * WAD as f64) as u64;
```
jordansexton, line 384: @jstarry does this look safe?

token-lending/program/src/processor.rs
```diff
@@ -1859,6 +1861,20 @@ fn spl_token_init_mint(params: TokenInitializeMintParams<'_, '_>) -> ProgramResu
...
+    account_infos: &[AccountInfo],
+    authority_signer_seeds: &[&[u8]],
+) -> ProgramResult {
+    if authority_signer_seeds.is_empty() {
+        invoke(instruction, account_infos)
+    } else {
+        invoke_signed(instruction, account_infos, &[authority_signer_seeds])
+    }
+}
+
```
jordansexton, line 1877: @jstarry this fixes the issue with the PDA.

token-lending/cli/README.md
jordansexton: The lending program has been deployed to devnet and testnet, and a lending market has been added to both. A reserve has been added to devnet. However, the testnet program cannot have reserves added, as the Pyth program is not available outside of devnet yet.

jordansexton:
@jstarry this is ready for review again. Thanks in advance!

token-lending/program/src/processor.rs
nope-finance: one minor thing for this is I was saying you can actually call invoke_signed still but you have to pass in &[] instead of &[authority_signer_seeds] so it might be easier to just change the way you were calling the functions which need to sign to pass in  &[&[&[u8]]] instead of  &[&[u8]] to match the signing function signature anyway. but maybe that has some other drawbacks im not seeing?

token-lending/cli/src/main.rs
jstarry: Yeah, follow up sounds good

token-lending/cli/README.md
jstarry: No problem, devnet is good enough for now, thanks!

token-lending/README.md
```diff
@@ -1,15 +1,118 @@
...
-| Devnet | [`LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi`](https://explorer.solana.com/address/LendZqTs8gn5CTSJU1jWKhKuVpjJGom45nnwPb2AMTi?cluster=devnet) |
\ No newline at end of file
+| Testnet | [`6TvznH3B2e3p2mbhufNBpgSrLx6UkgvxtVQvopEZ2kuH`](https://explorer.solana.com/address/6TvznH3B2e3p2mbhufNBpgSrLx6UkgvxtVQvopEZ2kuH?cluster=testnet) |
+| Devnet | [`6TvznH3B2e3p2mbhufNBpgSrLx6UkgvxtVQvopEZ2kuH`](https://explorer.solana.com/address/6TvznH3B2e3p2mbhufNBpgSrLx6UkgvxtVQvopEZ2kuH?cluster=devnet) |
+
+### Deploy a lending program (optional)
+
+This is optional! You can simply add your own market and reserves to the existing on-chain programs.
+
+You can skip these steps and use the [Token Lending CLI](./cli/README.md) to create a lending market and add reserves to it.
```
jstarry, line 23: nit: what does "it" mean here? the already existing on-chain programs?

token-lending/program/src/processor.rs
jordansexton: Looks like the implementation of `invoke` just calls `invoke_signed` with an empty array anyway:
```rust
pub fn invoke(instruction: &Instruction, account_infos: &[AccountInfo]) -> ProgramResult {
    invoke_signed(instruction, account_infos, &[])
}
```

token-lending/README.md
jordansexton: Ah, `it` here is `a lending market`. But I can change this to make it more clear.

token-lending/cli/src/main.rs
jstarry: Yup, this looks fine to me. Search for "saturating cast" here: https://doc.rust-lang.org/nomicon/casts.html

token-lending/README.md
jordansexton: Done.

token-lending/program/src/processor.rs
jstarry: The borrow balance includes interest so when you repay you need to ensure that interest has been accrued first

jordansexton: Yeah, you're right. I was thinking about the fact that the withdraw collateral instruction is separate now, and that requires refreshing all the reserves and obligation, so it might not be needed on repay. However, if you were able to repay without accruing interest, refresh, and then withdraw, the interest wouldn't accrue correctly in the refresh because your balance had been changed.

jordansexton: Removed these comments.

jstarry: Great, another thing is that I remember in the initial multi-collateral PR of yours we discussed how it would be ideal you could do multiple obligation operations in the same transaction, not sure if the stale flag on obligations would prevent any desired lending flows at this point.

PR #1679, opened 2021-05-08
Fix cross-program-invocation C example

machacekch:
Correctly check if deriving the PDA with the given seed was successful.
Don't pass the system program pubkey to the allocate instruction: the first account passed must be the one to be allocated.


mvines:
Thanks again :)

PR #1680, opened 2021-05-08
"Account" typically only has 2 Cs

CriesofCarrots:
Fix spelling


mvines:
Mabad, that's the Canadian spelling.  The third c is silent.

PR #1682, opened 2021-05-08
Ensure only docs files are modified when triggering all-checks success

CriesofCarrots:
If a PR has docs changes in additional to other code changes, Mergify may merge it before the full CI suite completes.


joncinque:
Nice catch!

PR #1744, opened 2021-05-15
Comment says javascript but code is typescript

Henry-E:


docs/src/associated-token-account.md
```diff
@@ -55,7 +55,7 @@ The [get_associated_token_address](https://docs.rs/spl-associated-token-account/
 Rust function may be used by clients to derive the wallet's associated token address.
 
 
-The associated account address can be derived in Javascript with:
+The associated account address can be derived in Typescript with:
```
joncinque, line 58: In that case, we may as well go all the way and also fix up the case :smile: 
```suggestion
The associated account address can be derived in TypeScript with:
```

joncinque (commented):
Thanks for the PR! We typically use the term "JavaScript" to mean all JS-type languages, and you're right that we can be clearer in this case.

PR #1757, opened 2021-05-18
lending: use Pyth as oracle

jordansexton:
Removes flux-aggregator and adds Pyth. Adds mocking for Pyth prices in tests. All tests are passing.


jordansexton:
@leonardgee There's a bit to do around decimal normalization but this is functional.

token-lending/program/src/state/lending_market.rs
```diff
@@ -46,8 +47,8 @@ pub struct InitLendingMarketParams {
     pub bump_seed: u8,
     /// Owner authority which can add new reserves
     pub owner: Pubkey,
-    /// Quote currency token mint
-    pub quote_token_mint: Pubkey,
+    /// Currency market prices are quoted in (e.g. "USD", or a SPL token mint pubkey)
+    pub quote_currency: [u8; 32],
```
jordansexton, line 51: Inconsistent with above doc comment.

token-lending/program/src/state/lending_market.rs
```diff
@@ -66,22 +67,22 @@ impl Pack for LendingMarket {
     fn pack_into_slice(&self, output: &mut [u8]) {
         let output = array_mut_ref![output, 0, LENDING_MARKET_LEN];
         #[allow(clippy::ptr_offset_with_cast)]
-        let (version, bump_seed, owner, quote_token_mint, token_program_id, _padding) =
-            mut_array_refs![output, 1, 1, PUBKEY_BYTES, PUBKEY_BYTES, PUBKEY_BYTES, 128];
+        let (version, bump_seed, owner, quote_currency, token_program_id, _padding) =
+            mut_array_refs![output, 1, 1, PUBKEY_BYTES, 32, PUBKEY_BYTES, 128];
```
jordansexton, line 71: This is to allow the implementation to be expanded later with backwards compatibility.

token-lending/program/src/state/reserve.rs
```diff
@@ -356,14 +355,15 @@ pub struct ReserveLiquidity {
...
-    pub oracle_pubkey: COption<Pubkey>,
+    /// Reserve liquidity oracle account
+    pub oracle_pubkey: Pubkey,
     /// Reserve liquidity available
     pub available_amount: u64,
     /// Reserve liquidity borrowed
     pub borrowed_amount_wads: Decimal,
     /// Reserve liquidity cumulative borrow rate
     pub cumulative_borrow_rate_wads: Decimal,
+    // @TODO: make Decimal
```
jordansexton, line 366: I'm considering making the market price here a Decimal, as it's used on all the obligation refreshes and would require fewer math operations.

token-lending/program/src/state/reserve.rs
```diff
@@ -712,11 +712,11 @@ impl IsInitialized for Reserve {
     }
 }
 
-const RESERVE_LEN: usize = 567; // 1 + 8 + 1 + 32 + 32 + 1 + 32 + 32 + (4 + 32) + 16 + 8 + 8 + 16 + 32 + 32 + 8 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 8 + 8 + 1 + 248
-                                // @TODO: break this up by reserve / liquidity / collateral / config https://git.io/JOCca
+const RESERVE_LEN: usize = 563; // 1 + 8 + 1 + 32 + 32 + 1 + 32 + 32 + 32 + 16 + 8 + 8 + 16 + 32 + 32 + 8 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 8 + 8 + 1 + 248
```
jordansexton, line 715: We lose 4 bytes without the COption prefix.

token-lending/program/tests/borrow_obligation_liquidity.rs
```diff
@@ -20,7 +20,7 @@ use spl_token_lending::{
 use std::u64;
 
 #[tokio::test]
-async fn test_borrow_quote_currency() {
+async fn test_borrow_usdc_fixed_amount() {
```
jordansexton, line 23: There is no SPL Lending Market quote currency, so all tests that use USDC have been changed accordingly.

token-lending/program/tests/fixtures/README.md
```diff
@@ -1,11 +1,16 @@
...
-solana account ELz3y7PCa6gp8GjD8TrXnggmS2bxsDrznrxAMmSz9Thq --output-file eth_usd_aggregator.bin
+
+# Pyth product: SOL/USD
+solana account 8yrQMUyJRnCJ72NWwMiPV9dNGw465Z8bKUvnUC8P5L6F --output-file 8yrQMUyJRnCJ72NWwMiPV9dNGw465Z8bKUvnUC8P5L6F.bin
+# Pyth price: SOL/USD
+solana account BdgHsXrH1mXqhdosXavYxZgX6bGqTdj5mh2sxDhF8bJy --output-file BdgHsXrH1mXqhdosXavYxZgX6bGqTdj5mh2sxDhF8bJy.bin
+# Pyth product: SRM/USD
+solana account 5agdsn3jogTt8F537GW3g8BuLaBGrg9Q2gPKUNqBV6Dh --output-file 5agdsn3jogTt8F537GW3g8BuLaBGrg9Q2gPKUNqBV6Dh.bin
+# Pyth price: SRM/USD
+solana account 2Mt2wcRXpCAbTRp2VjFqGa8SbJVzjJvyK4Tx7aqbRtBJ --output-file 2Mt2wcRXpCAbTRp2VjFqGa8SbJVzjJvyK4Tx7aqbRtBJ.bin
```
jordansexton, line 15: Dumps of Pyth accounts for mocking values in the tests. Pyth doesn't currently provide a USDC/USD feed, so we use the SRM account and set the price to $1 later.

token-lending/program/tests/helpers/mod.rs
```diff
@@ -300,6 +250,13 @@ pub fn add_reserve(
 
     let current_slot = slots_elapsed + 1;
 
+    let market_price = oracle
+        .price
+        .checked_div(oracle.decimals)
+        .unwrap()
+        .checked_mul(10u64.pow(6))
+        .unwrap();
```
jordansexton, line 258: Need to update this if price decimal normalization changes.

token-lending/program/tests/helpers/mod.rs
```diff
@@ -782,13 +724,12 @@ impl TestReserve {
...
-            (None, 1 * FRACTIONAL_TO_USDC)
-        } else {
-            panic!("aggregator is required");
-        };
+        let market_price = oracle
+            .price
+            .checked_div(oracle.decimals)
+            .unwrap()
+            .checked_mul(10u64.pow(6))
+            .unwrap();
```
jordansexton, line 732: Ditto.

token-lending/program/tests/helpers/mod.rs
```diff
@@ -1181,99 +1056,90 @@ pub fn add_usdc_mint(test: &mut ProgramTest) -> TestQuoteMint {
...
-    pub name: String,
-    pub pubkey: Pubkey,
-    pub price: u64,
+pub fn add_sol_oracle(test: &mut ProgramTest) -> TestOracle {
+    add_oracle(
+        test,
+        Pubkey::from_str(SOL_PYTH_PRODUCT).unwrap(),
+        Pubkey::from_str(SOL_PYTH_PRICE).unwrap(),
+        // Set SOL price to $20
+        20 * LAMPORTS_TO_SOL,
```
jordansexton, line 1079: $20 SOL feels like so long ago...

But this is the value the tests all use.

token-lending/program/tests/helpers/mod.rs
```diff
@@ -1181,99 +1056,90 @@ pub fn add_usdc_mint(test: &mut ProgramTest) -> TestQuoteMint {
...
-        TestAggregatorPair::SRM_USDC => ("SRM:USDC", 6, 5 * FRACTIONAL_TO_USDC),
-    };
+pub fn add_usdc_oracle(test: &mut ProgramTest) -> TestOracle {
+    add_oracle(
+        test,
+        // Mock with SRM since Pyth doesn't have USDC yet
+        Pubkey::from_str(SRM_PYTH_PRODUCT).unwrap(),
+        Pubkey::from_str(SRM_PYTH_PRICE).unwrap(),
+        // Set USDC price to $1
+        1 * LAMPORTS_TO_SOL,
```
jordansexton, line 1090: hack++

token-lending/program/tests/helpers/mod.rs
```diff
@@ -1181,99 +1056,90 @@ pub fn add_usdc_mint(test: &mut ProgramTest) -> TestQuoteMint {
...
 
-    let mut account = Account::new(
-        u32::MAX as u64,
-        borsh_utils::get_packed_len::<Aggregator>(),
-        &spl_token_lending::id(),
+    let decimals = 10u64
+        .checked_pow(pyth_price.expo.checked_abs().unwrap().try_into().unwrap())
+        .unwrap();
+
+    pyth_price.agg.price = price.try_into().unwrap();
```
jordansexton, line 1122: Mutate the Pyth fixture to mock the desired price.

token-lending/program/tests/helpers/mod.rs
```diff
@@ -1384,3 +1250,8 @@ pub async fn get_token_balance(banks_client: &mut BanksClient, pubkey: Pubkey) -
         .unwrap()
         .amount
 }
+
+fn cast_mut<T>(d: &mut [u8]) -> &mut T {
+    let (_, pxa, _) = unsafe { d.align_to_mut::<T>() };
+    &mut pxa[0]
+}
```
jordansexton, line 1257: Copied with modification from Pyth rust client. There's no safe unpack mechanism defined.

jordansexton (commented):
@bartosz-lipinski @jstarry Added a review with comments clarifying most of the design changes.

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1720,68 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+        pyth_client::PriceType::Price => {}
+        _ => {
+            msg!("Oracle price type is invalid");
+            return Err(LendingError::InvalidOracleConfig.into());
+        }
+    }
+    if pyth_price.valid_slot < clock.slot {
+        msg!("Oracle price is stale");
+        return Err(LendingError::InvalidOracleConfig.into());
+    }
```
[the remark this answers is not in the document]
jordansexton: This appears to be a good use for this, but this is too strict and will almost always be seen as invalid. 2803208 adjusts the valid window to 60 seconds. I'm not sure what timeframe is reasonable, and it may vary by currency. We could make it configurable per reserve if it matters.

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1724,69 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+
+    match pyth_price.ptype {
+        pyth_client::PriceType::Price => {}
+        _ => {
+            msg!("Oracle price type is invalid");
+            return Err(LendingError::InvalidOracleConfig.into());
+        }
+    }
+
+    if pyth_price.valid_slot < clock.slot - SLOTS_PER_MINUTE {
```
jstarry, line 1755: a minute is a long time, I think we need this to be shorter. Do we have any idea yet how often pyth feeds will be updated?

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1724,69 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+        msg!("Oracle price is stale");
+        return Err(LendingError::InvalidOracleConfig.into());
+    }
+
+    let price: u64 = pyth_price
+        .agg
+        .price
+        .checked_abs()
```
jstarry, line 1763: wow, prices can be negative?? that seems bad, can we get pyth to change that

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1720,68 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+        .ok_or(LendingError::MathOverflow)?
+        .try_into()
+        .map_err(|_| LendingError::MathOverflow)?;
+
+    let pyth_decimals = 10u64
+        .checked_pow(pyth_exponent)
+        .ok_or(LendingError::MathOverflow)?;
+
+    // @FIXME: magic number
+    let quote_decimals = 10u64.pow(6);
```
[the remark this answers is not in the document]
jstarry: This might be too small, imagine a token with a really high circulating supply which is worth a really small amount. Maybe we can store both the price and the exponent and then determine market value directly from those?

token-lending/program/tests/helpers/mod.rs
jstarry: Constant would be nice

token-lending/program/src/processor.rs
jordansexton: They have varying `PriceType` for price objects. I thought maybe they are used differently, and a negative value for one might indicate a change since last or something. I'll definitely try to ask.

token-lending/program/src/processor.rs
```diff
@@ -262,38 +254,57 @@ fn process_init_reserve(
         msg!("Lending market owner provided must be a signer");
         return Err(LendingError::InvalidSigner.into());
     }
-    if &lending_market.quote_token_mint != quote_token_mint_info.key {
-        msg!("Lending market quote token mint does not match the quote token mint provided");
-        return Err(LendingError::InvalidAccountInput.into());
+
+    let pyth_product_data = &pyth_product_info.try_borrow_data()?;
+    let pyth_product = pyth_client::cast::<pyth_client::Product>(pyth_product_data);
+    if pyth_product.magic != pyth_client::MAGIC {
```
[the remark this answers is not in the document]
jordansexton: Freeing up extra bytes for this must be what all the terse variable names are for.

token-lending/program/src/processor.rs
jordansexton: > The pyth program accumulates prices with respect to whatever the current slot is inside the solana node. This is called the "valid_slot" above. As soon as the slot ticks forward by one, the pyth program computes the aggregate price and publishes it with respect to the new "publish_slot" and starts repeating the process with a new "valid_slot".

Anecdotally, they seem to be updated very rapidly. I will make this quite short and see how it behaves.

token-lending/program/src/processor.rs
jordansexton: Clarified that's what this is for:

> Price used is i64 because prices can be sometimes negative 
> By price, it can range across multiple price types ; price of a future or yield can be negative
> Also a publisher might publish -1 in price as a null price so it also covers this situation

token-lending/program/tests/helpers/mod.rs
jordansexton: This has been removed in favor of using `Decimal` for market price and using the price + exponent directly from the oracle.

token-lending/program/src/state/reserve.rs
jordansexton: Done.

token-lending/program/src/processor.rs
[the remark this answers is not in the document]
jordansexton: Fixed.

token-lending/program/src/processor.rs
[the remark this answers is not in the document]
jordansexton: Changed to 5 slots.

token-lending/program/src/processor.rs
```diff
@@ -262,38 +254,57 @@ fn process_init_reserve(
         msg!("Lending market owner provided must be a signer");
         return Err(LendingError::InvalidSigner.into());
     }
-    if &lending_market.quote_token_mint != quote_token_mint_info.key {
-        msg!("Lending market quote token mint does not match the quote token mint provided");
-        return Err(LendingError::InvalidAccountInput.into());
+
+    let pyth_product_data = &pyth_product_info.try_borrow_data()?;
```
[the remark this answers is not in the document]
jordansexton: Done.

token-lending/program/src/processor.rs
```diff
@@ -262,38 +254,57 @@ fn process_init_reserve(
...
-        if account_info_iter.peek().is_some() {
-            msg!("Reserve liquidity oracle cannot be provided when reserve liquidity is the quote currency");
-            return Err(LendingError::InvalidAccountInput.into());
-        }
-        // 1 because quote token price is equal to itself
-        (COption::None, 1)
-    } else {
-        let reserve_liquidity_oracle_info = next_account_info(account_info_iter)?;
-        assert_rent_exempt(rent, reserve_liquidity_oracle_info)?;
+    let pyth_price_pubkey = Pubkey::new(&pyth_product.px_acc.val);
```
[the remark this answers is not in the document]
jordansexton: Done.

token-lending/program/src/processor.rs
jordansexton: https://explorer.solana.com/address/5mkqGkkWSaSk2NL9p4XptwEQu4d5jFTJiurbbzdqYexF?cluster=devnet (Pyth program on devnet) shows the updates happening very frequently. I've shortened this to 5 slots for now.

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1720,68 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+    T: Iterator<Item = &'a u8>,
+{
+    let mut len = *ite.next().unwrap() as usize;
+    let mut val = String::with_capacity(len);
+    while len > 0 {
+        val.push(*ite.next().unwrap() as char);
+        len -= 1;
+    }
+    val
+}
```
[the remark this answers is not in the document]
jordansexton: This is done, open to further revision though.

token-lending/program/src/processor.rs
```diff
@@ -722,8 +726,8 @@ fn process_refresh_obligation(program_id: &Pubkey, accounts: &[AccountInfo]) ->
         let market_value = deposit_reserve
             .collateral_exchange_rate()?
             .decimal_collateral_to_liquidity(collateral.deposited_amount.into())?
-            .try_mul(deposit_reserve.liquidity.market_price)?
-            .try_div(decimals)?;
+            .try_div(decimals)?
+            .try_mul(deposit_reserve.liquidity.market_price)?;
```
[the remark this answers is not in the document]
jordansexton: Done.

token-lending/program/src/state/lending_market.rs
jordansexton: Fixed.

token-lending/program/src/processor.rs
```diff
@@ -262,38 +254,57 @@ fn process_init_reserve(
         msg!("Lending market owner provided must be a signer");
         return Err(LendingError::InvalidSigner.into());
     }
-    if &lending_market.quote_token_mint != quote_token_mint_info.key {
-        msg!("Lending market quote token mint does not match the quote token mint provided");
-        return Err(LendingError::InvalidAccountInput.into());
+
+    let pyth_product_data = &pyth_product_info.try_borrow_data()?;
+    let pyth_product = pyth_client::cast::<pyth_client::Product>(pyth_product_data);
```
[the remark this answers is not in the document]
jordansexton: Done.

jordansexton:
@jstarry This is ready for review again. Making reserve market prices Decimal saves a bunch of calculation. I also added safe unpacking for the Pyth accounts. My overall plan is to get mergeable and then refactor the oracle implementation into a separate program.

token-lending/program/src/processor.rs
jstarry: Makes sense, thanks for checking

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1719,85 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
     Mint::unpack(data).map_err(|_| LendingError::InvalidTokenMint)
 }
 
+fn get_pyth_product_quote_currency(pyth_product: &pyth::Product) -> [u8; 32] {
+    let mut quote_currency = [0u8; 32];
+
+    let mut attribute_iter = pyth_product.attr[..].iter();
```
jstarry, line 1725: Iterating over each byte is probably too inefficient. It would be better to use indexes

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1719,85 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+
+    let price: u64 = pyth_price
+        .agg
+        .price
+        .checked_abs()
+        .ok_or(LendingError::MathOverflow)?
+        .try_into()
+        .map_err(|_| LendingError::MathOverflow)?;
+
+    // @TODO: handle the case that the exponent is positive?
```
jstarry, line 1784: is this comment still needed?

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1719,85 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+}
+
+fn get_pyth_price(pyth_price_info: &AccountInfo, clock: &Clock) -> Result<Decimal, ProgramError> {
+    const STALE_AFTER_SLOTS_ELAPSED: u64 = 5;
+
+    let pyth_price_data = pyth_price_info.try_borrow_data()?;
+    let pyth_price = pyth::load::<pyth::Price>(&pyth_price_data)
+        .map_err(|_| ProgramError::InvalidAccountData)?;
+
+    if pyth_price.ptype as u32 != pyth::PriceType::Price as u32 {
```
jstarry, line 1762: nit: you can derive `PartialEq` for `PriceType` and then check for equality without casting

token-lending/program/src/pyth.rs
```diff
@@ -0,0 +1,148 @@
+#![allow(missing_docs)]
```
jstarry, line 1: Is this file copied from somewhere? Could you add a comment about that if so?

token-lending/program/src/pyth.rs
```diff
@@ -0,0 +1,148 @@
...
+    Volatility,
+}
+
+impl Default for PriceType {
+    fn default() -> Self {
+        PriceType::Price
+    }
+}
+
+#[derive(Default, Copy, Clone)]
```
jstarry, line 93: nit: I don't think the `Default` auto derive here is helpful

token-lending/program/tests/helpers/mod.rs
```diff
@@ -1384,3 +1253,8 @@ pub async fn get_token_balance(banks_client: &mut BanksClient, pubkey: Pubkey) -
         .unwrap()
         .amount
 }
+
+fn cast_mut<T>(d: &mut [u8]) -> &mut T {
```
jstarry, line 1257: looks like this is unused

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1719,85 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+        .ok_or(LendingError::MathOverflow)?;
+    if slots_elapsed >= STALE_AFTER_SLOTS_ELAPSED {
+        msg!("Oracle price is stale");
+        return Err(LendingError::InvalidOracleConfig.into());
+    }
+
+    let price: u64 = pyth_price
+        .agg
+        .price
+        .checked_abs()
```
jstarry, line 1779: Why take absolute value here? I think it would be better to directly use `try_into` and return an error if the value is negative.

token-lending/program/src/pyth.rs
jordansexton: True, this was copied from an example of an Anchor program using Pyth. I'll clean these up.

token-lending/program/tests/helpers/mod.rs
jordansexton: Removed.

token-lending/program/src/pyth.rs
jordansexton: Done.

token-lending/program/src/processor.rs
jordansexton: I think so. I guess the exponent could be positive, reflecting some number of zeros to add to the price? I've handled both cases now.

token-lending/program/src/processor.rs
```diff
@@ -262,38 +257,53 @@ fn process_init_reserve(
...
+    }
 
-        let aggregator = Aggregator::load_initialized(reserve_liquidity_oracle_info)?;
-        if aggregator.config.decimals != quote_token_mint.decimals {
-            msg!(
-                "Quote token mint decimals does not match the aggregator config decimals provided"
-            );
-            return Err(LendingError::InvalidOracleConfig.into());
-        }
+    let pyth_product_pubkey_bytes: &[u8; 32] = pyth_price_info
```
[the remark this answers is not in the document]
jordansexton: Good catch, thanks.

token-lending/program/src/processor.rs
```diff
@@ -262,38 +257,53 @@ fn process_init_reserve(
...
+        return Err(LendingError::InvalidOracleConfig.into());
+    }
 
-        (
-            COption::Some(*reserve_liquidity_oracle_info.key),
-            read_median(reserve_liquidity_oracle_info)?.median,
-        )
-    };
+    let quote_currency = get_pyth_product_quote_currency(pyth_product);
+    if quote_currency == [0u8; 32] {
```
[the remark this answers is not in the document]
jordansexton: Done. This is also nice because it can just return early.

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1719,85 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
 
+fn get_pyth_product_quote_currency(pyth_product: &pyth::Product) -> [u8; 32] {
+    let mut quote_currency = [0u8; 32];
+
+    let mut attribute_iter = pyth_product.attr[..].iter();
+    let mut attributes_size = pyth::PROD_ATTR_SIZE;
+
+    while attributes_size > 0 {
+        let mut size = *attribute_iter.next().unwrap() as usize;
+        let mut key = Vec::with_capacity(size);
```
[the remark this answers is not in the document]
jordansexton: Makes sense, I'm reworking the whole iteration. I just used Pyth's originally, but their use case was just to take a bunch of products and print all their attributes as strings.

token-lending/program/src/processor.rs
jordansexton: Done.

token-lending/program/src/processor.rs
jordansexton: Done.

token-lending/program/src/processor.rs
jstarry: Isn't the exponent always positive?

jordansexton: It's an `i32`, I suspect not.

jordansexton: It's usually negative, e.g. `-9` for SOL.

token-lending/program/src/pyth.rs
jordansexton: Done.

token-lending/program/src/processor.rs
jordansexton: Done.

token-lending/program/src/pyth.rs
jordansexton: Removed, none of the defaults were needed.

token-lending/program/tests/helpers/mod.rs
jordansexton: Removed. Thanks!

token-lending/program/src/processor.rs
jordansexton: Done.

token-lending/program/src/processor.rs
jordansexton: Implemented.

token-lending/program/src/processor.rs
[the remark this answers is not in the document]
jordansexton: Done, and added an optimization where the key bytes are only matched if the length matches.

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1719,85 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+        for _ in 0..size {
+            key.push(*attribute_iter.next().unwrap());
+        }
+
+        if &key[..] == b"quote_currency" {
+            size = *attribute_iter.next().unwrap() as usize;
+            #[allow(clippy::needless_range_loop)]
+            for i in 0..size {
+                quote_currency[i] = *attribute_iter.next().unwrap();
```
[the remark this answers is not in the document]
jordansexton: Done.

token-lending/program/src/processor.rs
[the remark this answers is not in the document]
jordansexton: Did this, and then ended up inlining the `Result<[0u8; 32], ProgramError>` that would've happened if `None` anyway.

token-lending/program/src/processor.rs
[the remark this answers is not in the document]
jordansexton: Fixed.

jordansexton:
@jstarry Address all feedback I think. Let me know if the new attribute iteration seems reasonable.

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1715,87 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+    let mut offset = 0;
+    while offset < pyth::PROD_ATTR_SIZE {
+        let mut len = pyth_product.attr[offset] as usize;
+        offset += 1;
+
+        if len == LEN {
+            let key = &pyth_product.attr[offset..(offset + len)];
+            if key == KEY {
+                offset += len;
+                len = pyth_product.attr[offset] as usize;
```
jstarry, line 1731: Can you add a check that `len` is less than or equal to 32?

token-lending/program/src/processor.rs
```diff
@@ -1714,6 +1715,87 @@ fn unpack_mint(data: &[u8]) -> Result<Mint, LendingError> {
...
+    const LEN: usize = 14;
+    const KEY: &[u8; LEN] = b"quote_currency";
+
+    let mut offset = 0;
+    while offset < pyth::PROD_ATTR_SIZE {
+        let mut len = pyth_product.attr[offset] as usize;
+        offset += 1;
+
+        if len == LEN {
+            let key = &pyth_product.attr[offset..(offset + len)];
```
jstarry, line 1728: For this and the value copy below, we should check that `pyth_produce.attr` is long enough

jstarry (approved):

token-lending/program/src/processor.rs
```diff
@@ -1,16 +1,17 @@
 //! Program state processor
+
```
leonardgee, line 2: nit: why do we add new line here.

token-lending/program/src/processor.rs
jordansexton: Done.

token-lending/program/src/processor.rs
jordansexton: Done.

PR #1763, opened 2021-05-19
Governance: Create governance accounts

SebastianBor:
#### Implemented instructions:
- `CreateAccountGovernance`
- `CreateProgramGovernance`

#### Design changes:
During implementation I realised we over-specified the Governance account to be just `ProgramGovernance`.
In this PR I generalised the concept to be a governance over an arbitrary account and expressed it using a generic `Governance` struct. It can be a `Program` account but also a `Mint` or a `Token` account for example. 
The difference between these different account types can be captured in the relevant `CreateGovernance` instruction which has to take over the relevant account authority. In the case of an upgradable program account this is the program's `upgrade_authority`.
In this version I only implemented `CreateProgramGovernance` and generic `CreateAccountGovernance` but we can expand on this concept  and easily add more specific `CreateGovernance` instructions in future versions.


governance/program/tests/program_test/mod.rs
```diff
@@ -448,26 +456,217 @@ impl GovernanceProgramTest {
...
-        self.get_account::<Realm>(root_governance_address).await
+        self.get_borsh_account::<Realm>(root_governance_address)
+            .await
+    }
+
+    #[allow(dead_code)]
+    pub async fn get_account_governance_account(
+        &mut self,
+        program_governance_address: &Pubkey,
+    ) -> AccountGovernance {
```
joncinque, line 629: Nit for these functions, since there's no need to await for something within the function, you can make them non-`async`, have them return `impl Future<Output = AccountGovernance>`, and not await inside the body of the function

governance/program/tests/program_test/mod.rs
```diff
@@ -1,6 +1,9 @@
+use std::borrow::Borrow;
```
joncinque, line 1: Is this needed?

governance/program/src/state/account_governance.rs
```diff
@@ -0,0 +1,102 @@
...
+pub struct GovernanceConfig {
+    /// Governance Realm
+    pub realm: Pubkey,
+
+    /// Account governed by this Governance. It can be for example Program account, Mint account or Token Account
+    pub governed_account: Pubkey,
+
+    /// Voting threshold in % required to tip the vote
+    /// It's the percentage of tokens out of the entire pool of governance tokens eligible to vote
+    pub vote_threshold: u8,
```
joncinque, line 19: nit: Since it's a percentage, we can make it clear by calling it `vote_threshold_percentage`

governance/program/src/state/account_governance.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// Minimum % of tokens for a governance token owner to be able to create a proposal
+    /// It's the percentage of tokens out of the entire pool of governance tokens eligible to vote
+    pub token_threshold_to_create_proposal: u8,
```
joncinque, line 23: nit: same with this

governance/program/src/state/enums.rs
```diff
@@ -15,8 +15,8 @@ pub enum GovernanceAccountType {
     /// Voter record for each voter and given governing token type within a Realm
     VoterRecord,
 
-    /// Program Governance account
-    ProgramGovernance,
+    /// Account Governance account
+    AccountGovernance,
```
joncinque, line 19: Since program governances and account governances are not the same thing, mainly because they're derived using different seeds, we should have at least two different enums.  I think that'll make frontend integration easier.  When querying, we'll know explicitly if it's a program or account governance.  We can probably reuse the same type.  I'm not fussed either way

governance/program/src/processor/process_create_account_governance.rs
```diff
@@ -0,0 +1,52 @@
...
+    let account_governance_info = next_account_info(account_info_iter)?; // 0
+    let payer_info = next_account_info(account_info_iter)?; // 1
+    let system_info = next_account_info(account_info_iter)?; // 2
+
+    let rent_sysvar_info = next_account_info(account_info_iter)?; // 3
+    let rent = &Rent::from_account_info(rent_sysvar_info)?;
+
+    let account_governance_data = AccountGovernance {
+        account_type: GovernanceAccountType::AccountGovernance,
+        config: config.clone(),
```
joncinque, line 37: This seems prone to be accidentally misused, since it just accepts whatever realm is specified in the config.  We should probably also provide the realm account, and verify it's actually a realm.

Also, I may be missing it, but are there any checks on the other fields of the config? Mainly, are we sure that all of the `u8`s are real percentages?  And probably that the `voting threshold >= proposal creation threshold`.  Any others?

I'm not sure if we also want to validate the waiting times are within a set of bounds.

And with that, you'll want to add some tests to fail correctly if the realm doesn't exist.

governance/program/src/processor/process_create_program_governance.rs
```diff
@@ -0,0 +1,69 @@
...
+    let rent_sysvar_info = next_account_info(account_info_iter)?; // 6
+    let rent = &Rent::from_account_info(rent_sysvar_info)?;
+
+    set_program_upgrade_authority(
+        &config.governed_account,
+        governed_program_data_info,
+        governed_program_upgrade_authority_info,
+        program_governance_info,
+        bpf_upgrade_loader_info,
+    )?;
```
joncinque, line 50: Just because I'm paranoid, I'd have this done at the end of the instruction, just in case the runtime ever has a bug introduced allowing instructions to be in a half-finished state

governance/program/src/processor/process_create_program_governance.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    let program_governance_data = AccountGovernance {
+        account_type: GovernanceAccountType::AccountGovernance,
+        config: config.clone(),
```
joncinque, line 54: You'll want to do some of the realm-related checks mentioned in my previous comment

governance/program/src/tools/token.rs
```diff
@@ -3,17 +3,45 @@
...
+
+impl PrintProgramError for SplTokenToolsError {
+    fn print<E>(&self) {
+        msg!("SPL-TOKEN-TOOLS-ERROR: {}", &self.to_string());
+    }
+}
+
+impl From<SplTokenToolsError> for ProgramError {
+    fn from(e: SplTokenToolsError) -> Self {
+        ProgramError::Custom(e as u32)
```
joncinque, line 36: When you have multiple error types, because of this line, they'll get all mixed up in the logs, so you'll be better off keeping them all in one enum, and printing a specific message before returning them.

This is already an issue when programs propagate error codes from the token program, and get confused because they're not sure if it's an error from their program or the token program.

joncinque (commented):
Lots of great stuff here again!  A few questions / comments, mainly because you asked for lots of feedback :smile:

governance/program/src/instruction.rs
```diff
@@ -418,3 +421,61 @@ pub fn set_vote_authority(
...
+
+    Instruction {
+        program_id: id(),
+        accounts,
+        data: instruction.try_to_vec().unwrap(),
+    }
+}
+
+/// Creates CreateProgramGovernance instruction
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
SebastianBor: Removed

governance/program/src/processor/process_create_account_governance.rs
```diff
@@ -0,0 +1,52 @@
...
+use solana_program::{
+    account_info::{next_account_info, AccountInfo},
+    entrypoint::ProgramResult,
+    pubkey::Pubkey,
+    rent::Rent,
+    sysvar::Sysvar,
+};
+
+/// Processes CreateAccountGovernance instruction
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
SebastianBor: Removed

governance/program/tests/program_test/tools.rs
```diff
@@ -1,12 +1,55 @@
...
+pub fn read_test_program_elf(name: &str) -> Vec<u8> {
+    let path = get_test_program_path(name);
+    let mut file = File::open(&path).unwrap_or_else(|err| {
+        panic!("Failed to open {}: {}", path.display(), err);
+    });
+    let mut elf = Vec::new();
+    file.read_to_end(&mut elf).unwrap();
+
+    elf
+}
```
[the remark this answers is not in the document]
SebastianBor: Done

governance/program/tests/program_test/tools.rs
```diff
@@ -1,12 +1,55 @@
+use std::{convert::TryFrom, fs::File, io::Read, path::PathBuf};
+
 use solana_program::{instruction::InstructionError, program_error::ProgramError};
 use solana_sdk::{transaction::TransactionError, transport::TransportError};
 
+/// Instruction errors not mapped in the sdk
```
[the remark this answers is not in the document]
SebastianBor: Agree, I added 'TODO: Add to SDK' comment and will do that later

governance/program/tests/program_test/mod.rs
```diff
@@ -448,26 +456,217 @@ impl GovernanceProgramTest {
...
+    #[allow(dead_code)]
+    pub async fn get_upgradable_loader_account(
+        &mut self,
+        address: &Pubkey,
+    ) -> UpgradeableLoaderState {
+        self.get_bincode_account(address).await
+    }
 
-        T::unpack(&raw_account.data).unwrap()
+    pub async fn get_borsh_account<T: BorshDeserialize>(&mut self, address: &Pubkey) -> T {
```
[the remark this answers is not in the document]
SebastianBor: Yes, that was my first take too. Flagged with  'TODO: Add to SDK' comment and will do that later

governance/program/tests/program_test/mod.rs
SebastianBor: Since you also have to add lifetimes I prefer using async/await which feels more concise to me than explicit Feature return

```
    pub fn get_account_governance_account<'a>(
        &'a mut self,
        program_governance_address: &'a Pubkey,
    ) -> impl Future<Output = AccountGovernance> + 'a {

```

governance/program/tests/program_test/mod.rs
SebastianBor: Yes, to deserialise bitcode based UpgradeableLoaderState account

governance/program/src/state/account_governance.rs
SebastianBor: Done

governance/program/src/state/account_governance.rs
SebastianBor: Done

governance/program/src/state/enums.rs
SebastianBor: I've been thinking about this because I didn't like  `AccountGovernance` name which felt too specific to me. 
I renamed it to be more generic `Governance` and added `AccountGovernance` and `ProgramGovernance` as account types. This way we can have multiple types using the same struct.
With a single struct covering different types we can write generic code in other places where the type is not important, for example when creating `Proposals`.

governance/program/src/processor/process_create_program_governance.rs
SebastianBor: Done

governance/program/src/tools/token.rs
SebastianBor: This is unfortunate, I wanted to keep it generic. 
I moved the errors back to GovernanceError

governance/program/src/processor/process_create_program_governance.rs
SebastianBor: Added assert for valid realm

governance/program/src/processor/process_create_account_governance.rs
SebastianBor: Added assert_is_valid_realm and relevant tests and now working on the other checks

SebastianBor: Config validation added 
- `Realm` is checked to be a valid account (initialised, correct type and owner)
- `vote_threshold_percentage` checked to be in <50,100> range
- I don't think if there are any bounds for the time values. In theory it could be anything, so no validation
- I replaced `token_threshold_percentage_to_create_proposal`with `min_tokens_to_create_proposal`. Percentage was not a good idea because if there are more tokens minted we might need bps or even smaller units to accept users with a single token for example. 
- It also made me thinking about transferring upgrade_authority at `Governance` creation time. Since it's a big commitment and any mistakes might be irreversible I made it optional. If the authority is not transferred then program checks the current `upgrade_authority` signed the transaction to prevent others from creating governances for programs they don't own. The authority then can be transfer to Governance at alter time in this case.

governance/program/src/tools/bpf_loader_upgradeable.rs
```diff
@@ -52,3 +56,43 @@ pub fn set_program_upgrade_authority<'a>(
...
+        .map_err(|_| GovernanceError::InvalidProgramDataAccountData)?
+    {
+        UpgradeableLoaderState::ProgramData {
+            slot: _,
+            upgrade_authority_address,
+        } => upgrade_authority_address,
+        _ => None,
+    };
+
+    match upgrade_authority {
```
joncinque, line 85: nit: using something like `ok_or` can make logic like this really slick, especially if it returns a result.  You can write
```
let upgrade_authority = upgrade_authority.ok_or(Err(GovernanceError::ProgramNotUpgradable.into()))?;
```
and then do `if upgrade_authority != ....`

governance/program/src/tools/bpf_loader_upgradeable.rs
```diff
@@ -52,3 +56,43 @@ pub fn set_program_upgrade_authority<'a>(
...
+    if program_data_info.owner != &bpf_loader_upgradeable::id() {
+        return Err(ProgramError::IncorrectProgramId);
+    }
+    let program_data_address = get_program_data_address(program_address);
+
+    if program_data_address != *program_data_info.key {
+        return Err(GovernanceError::InvalidProgramDataAccountAddress.into());
+    }
+
+    let upgrade_authority = match deserialize(&program_data_info.data.borrow())
```
joncinque, line 75: nit: since only one condition is matched, you can use `let upgrade_authority = if let UpgradeableLoaderState::ProgramData {} = deserialize() { } else { None }`

joncinque (commented):
Just a few extra comments that can be addressed in a separate PR

governance/program/src/state/governance.rs
```diff
@@ -1,8 +1,15 @@
...
+    error::GovernanceError, id, state::enums::GovernanceAccountType, tools::account::AccountMaxSize,
+};
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
-use solana_program::{program_pack::IsInitialized, pubkey::Pubkey};
+use solana_program::{
+    account_info::AccountInfo, program_error::ProgramError, program_pack::IsInitialized,
+    pubkey::Pubkey,
+};
+
+use super::realm::assert_is_valid_realm;
```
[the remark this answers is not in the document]
SebastianBor: Fixed in 6492e9eb

governance/program/src/state/governance.rs
```diff
@@ -100,3 +107,23 @@ pub fn get_account_governance_address<'a>(
...
+    governance_config: &GovernanceConfig,
+    realm_info: &AccountInfo,
+) -> Result<(), ProgramError> {
+    if realm_info.key != &governance_config.realm {
+        return Err(GovernanceError::InvalidGovernanceConfig.into());
+    }
+
+    assert_is_valid_realm(realm_info)?;
+
+    if governance_config.vote_threshold_percentage < 50
```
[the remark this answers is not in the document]
SebastianBor: Yeah, I'm addressing this issue in the following PR for voting workflow.
With threshold < 50 we have a race condition where you can vote on `Proposal` and reach the threshold to tip the vote before others can have a chance to vote Nay. 
In the voting PR I'm adding  logic to only tip votes if it's not possible to vote the other way any longer, and in case vote can't be tipped automatically  by the the end of voting period I'm adding a new instruction `ComleteVote` to manually transition `Proposal` `Voting` state to either `Succeeded` or `Defeated` state

governance/program/src/state/governance.rs
```diff
@@ -16,11 +23,10 @@ pub struct GovernanceConfig {
...
     /// Voting threshold in % required to tip the vote
     /// It's the percentage of tokens out of the entire pool of governance tokens eligible to vote
-    pub vote_threshold: u8,
+    pub vote_threshold_percentage: u8,
 
-    /// Minimum % of tokens for a governance token owner to be able to create a proposal
-    /// It's the percentage of tokens out of the entire pool of governance tokens eligible to vote
-    pub token_threshold_to_create_proposal: u8,
+    /// Minimum number of tokens a governance token owner must possess to be able to create a proposal
+    pub min_tokens_to_create_proposal: u16,
```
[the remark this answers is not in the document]
SebastianBor: I think percentage is not the correct solution to the problem.

Ideally we would need some histogram and then say the level of tokens covering let's say 90% token holders is the level required to create a `Proposal`. This way we would be able to say the bottom 10% can't create Proposals, however you would need all token accounts to calculate that level which is not practical. 

The problem with percentage is that you can accidentally deny rights to create Proposals for everybody if you mint more governance tokens.
Let's say we set the threshold to 1% for population of 100 users with 1 token each. It means anybody can create a Proposal because 1 token is required.
However if you mint 100 more tokens to 100 more users you would need 2 tokens to create Proposal which means nobody could create `Proposals` at that point. 

This is why I think the absolute number of tokens to create a Proposal is better solution for now and each community has to decide on that level. It's not a gatekeeper anyway. It's just a way to prevent spamming governance with Proposals.
In future versions we should allow the Governance program to govern its Config and hence the thresholds could be updated via Proposals

governance/program/src/tools/bpf_loader_upgradeable.rs
SebastianBor: Done in d2c761d0

governance/program/src/tools/bpf_loader_upgradeable.rs
SebastianBor: Done in d2c761d0

PR #1764, opened 2021-05-19
stake-pool: Fix fee calculation

joncinque:
#### Problem

Per discussion with @ruuda, the current calculation method gives the pool owner too many tokens.  Since the new lamports are already part of the pool, and we're minting a part of it, it's not the same as a normal deposit / withdrawal calculation where all of the tokens are new.

The old calculation simply did:

```
fee_in_pool_tokens = fee_in_sol * pre_pool_token_supply / pre_epoch_lamports
```

But those pool tokens are worth more than `(post_epoch_lamports - pre_epoch_lamports) * fee_numerator / fee_denominator`.  The correct equation should be:

```
fee_in_pool_tokens / (fee_in_pool_tokens + pre_pool_token_supply) = fee_in_sol / post_epoch_lamports
```

which simplifies to:

```
fee_in_pool_tokens = fee_in_sol * pool_token_supply / (post_epoch_lamports - fee_in_sol)
```

#### Solution

Fix the calculation, and add a couple of tests for it, including a proptest

### DIFF
--- a/stake-pool/program/src/state.rs
+++ b/stake-pool/program/src/state.rs
@@ -113,16 +113,26 @@ impl StakePool {
         )
         .ok()
     }
-    /// calculate the fee in pool tokens that goes to the manager
+
+    /// Calculate the fee in pool tokens that goes to the manager
+    ///
+    /// This function assumes that `reward_lamports` has not already been added
+    /// to the stake pool's `total_stake_lamports`
     pub fn calc_fee_amount(&self, reward_lamports: u64) -> Option<u64> {
         if self.fee.denominator == 0 {
             return Some(0);
         }
-        let pool_amount = self.calc_pool_tokens_for_deposit(reward_lamports)?;
+        let total_stake_lamports = self.total_stake_lamports.checked_add(reward_lamports)? as u128;
+        let fee_lamports = (reward_lamports as u128)
+            .checked_mul(self.fee.numerator as u128)
+            .unwrap()
+            .checked_div(self.fee.denominator as u128)
+            .unwrap();
         u64::try_from(
-            (pool_amount as u128)
-                .checked_mul(self.fee.numerator as u128)?
-                .checked_div(self.fee.denominator as u128)?,
+            (self.pool_token_supply as u128)
+                .checked_mul(fee_lamports)
+                .unwrap()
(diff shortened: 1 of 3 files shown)


stake-pool/program/src/state.rs
```diff
@@ -113,16 +113,26 @@ impl StakePool {
...
+    /// Calculate the fee in pool tokens that goes to the manager
+    ///
+    /// This function assumes that `reward_lamports` has not already been added
+    /// to the stake pool's `total_stake_lamports`
     pub fn calc_fee_amount(&self, reward_lamports: u64) -> Option<u64> {
         if self.fee.denominator == 0 {
             return Some(0);
         }
-        let pool_amount = self.calc_pool_tokens_for_deposit(reward_lamports)?;
+        let total_stake_lamports = self.total_stake_lamports.checked_add(reward_lamports)? as u128;
```
ruuda, line 125: If we widen to `u128` anyway, we can widen `self.total_stake_lamports` and `reward_lamports` before the add, and then the add is guaranteed not to overflow.

stake-pool/program/src/state.rs
```diff
@@ -113,16 +113,26 @@ impl StakePool {
...
+        let fee_lamports = (reward_lamports as u128)
+            .checked_mul(self.fee.numerator as u128)
+            .unwrap()
```
ruuda, line 128: This multiplication is guaranteed not to overflow, because `rewards_lamports` and `fee.numerator` are both `u64`s that get widened to `u128`, and `u64::MAX * u64::MAX < u128::MAX`.

stake-pool/program/src/state.rs
```diff
@@ -113,16 +113,26 @@ impl StakePool {
...
+            .checked_div(self.fee.denominator as u128)
+            .unwrap();
```
ruuda, line 130: This division is guaranteed not to divide by zero, because there is already a check for `fee.denominator == 0` at the start of the function.

ruuda (approved):
The new formula looks correct to me, and the new code implements the new formula.

stake-pool/program/src/state.rs
joncinque: The total lamports in the entire Solana system fits in a u64, so they're guaranteed by the runtime to not overflow.  You're right though, since we're widening either way, we can do it beforehand.

PR #1794, opened 2021-05-25
token-swap: Make single token withdrawal fair with better calcs

joncinque:
#### Problem

The withdraw one side calculation is too conservative, since it boils down to:

```
pool_tokens = calculate_one_sided_deposit(withdraw_amount, token_a_amount - withdraw_amount, token_b_amount, pool_token_supply)
```

It reuses the deposit calc, and tries to simulate the same situation by reducing one of the sides.  This isn't correct though, and actually requires a totally different calculation.  This was fine for awhile since it devalues withdrawals, keeping the pool healthy.

#### Solution

Split the curve API for deposits and withdrawals, implement the new withdrawal function for all curves, and proptests for each.


token-swap/program/src/curve/constant_price.rs
```diff
@@ -15,6 +15,46 @@ use {
     spl_math::{checked_ceil_div::CheckedCeilDiv, precise_number::PreciseNumber, uint::U256},
 };
 
+/// Get the amount of pool tokens for the given amount of token A or B.
+///
+/// The constant product implementation uses the Balancer formulas found at
+/// https://balancer.finance/whitepaper/#single-asset-deposit, specifically
+/// in the case for 2 tokens, each weighted at 1/2.
+pub fn trading_tokens_to_pool_tokens(
```
rawfalafel, line 23: Looks like you can delete this function

rawfalafel: Oh I see. This function is only used inside this specific curve. Nevermind!

PR #1803, opened 2021-05-26
lending: Flash Loan fixes

jordansexton:
This PR

- Makes the reserve account writable and persists the borrow and repay so the flash loan receiver program has a consistent view of the state.
- Reorganizes the accounts expected by the FlashLoan instruction so that the writable accounts come first, the reserve accounts come in order, and the flash loan receiver program ID and variadic accounts come last.
- Renames some things for consistency with the other instructions.
- Improves related tests.

### DIFF
--- a/token-lending/program/src/processor.rs
+++ b/token-lending/program/src/processor.rs
@@ -1524,10 +1524,6 @@ fn process_liquidate_obligation(
     Ok(())
 }
 
-const RECEIVE_FLASH_LOAN_INSTRUCTION_DATA_SIZE: usize = 9;
-
-const RECEIVE_FLASH_LOAN_INSTRUCTION_TAG: u8 = 0u8;
-
 #[inline(never)] // avoid stack frame limit
 fn process_flash_loan(
     program_id: &Pubkey,
@@ -1589,11 +1585,6 @@ fn process_flash_loan(
         liquidity_amount
     };
 
-    if reserve.liquidity.available_amount < flash_loan_amount {
-        msg!("Flash loan amount cannot exceed available amount");
-        return Err(LendingError::InsufficientLiquidity.into());
-    }
-
     let flash_loan_amount_decimal = Decimal::from(flash_loan_amount);
     let (origination_fee, host_fee) = reserve
         .config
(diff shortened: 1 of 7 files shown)


token-lending/program/src/processor.rs
```diff
@@ -1571,47 +1573,34 @@ fn process_flash_loan(program_id: &Pubkey, amount: u64, accounts: &[AccountInfo]
...
-
-    if &reserve.liquidity.fee_receiver != flash_loan_fee_receiver_account_info.key {
-        msg!("Reserve liquidity fee receiver does not match the flash loan fee receiver provided");
+    if &reserve.liquidity.fee_receiver != reserve_liquidity_fee_receiver_info.key {
+        msg!("Reserve liquidity fee receiver does not match the reserve liquidity fee receiver provided");
         return Err(LendingError::InvalidAccountInput.into());
     }
 
-    let flash_loan_amount = if amount == u64::MAX {
+    // @FIXME: if u64::MAX is flash loaned, fees should be inclusive as with ordinary borrows
```
jordansexton, line 1581: I plan to do this in a followup PR.

jordansexton:
@jstarry This is ready for review. Thanks in advance!

token-lending/program/src/processor.rs
```diff
@@ -1571,94 +1577,92 @@ fn process_flash_loan(program_id: &Pubkey, amount: u64, accounts: &[AccountInfo]
...
+        source: source_liquidity_info.clone(),
+        destination: destination_liquidity_info.clone(),
+        amount: flash_loan_amount,
+        authority: lending_market_authority_info.clone(),
+        authority_signer_seeds,
+        token_program: token_program_id.clone(),
+    })?;
+
+    const RECEIVE_FLASH_LOAN_INSTRUCTION_DATA_SIZE: usize = 9;
+    const RECEIVE_FLASH_LOAN_INSTRUCTION_TAG: u8 = 0u8;
```
jstarry, line 1639: Can you add a fixme for this too? Using a `0u8` to indicate a special flash loan instruction is likely to cause conflicts in receiver programs and will make it harder for applications to be updated for flash loan support

token-lending/program/src/processor.rs
```diff
@@ -1571,94 +1577,92 @@ fn process_flash_loan(program_id: &Pubkey, amount: u64, accounts: &[AccountInfo]
...
+
+    let mut data = Vec::with_capacity(RECEIVE_FLASH_LOAN_INSTRUCTION_DATA_SIZE);
+    data.push(RECEIVE_FLASH_LOAN_INSTRUCTION_TAG);
+    data.extend_from_slice(&returned_amount_required.to_le_bytes());
+
     invoke(
-        &flash_loan_receiver_instruction,
+        &Instruction {
+            program_id: *flash_loan_receiver_program_id.key,
```
jstarry, line 1647: To be safe, let's make sure that this program id is not the same as the lending program id to avoid reentrancy attacks

token-lending/program/src/processor.rs
jordansexton: That makes sense, what do you suggest as an alternative?

token-lending/program/src/processor.rs
jordansexton: Done.

token-lending/program/src/processor.rs
jstarry: Ideally we hash the name of the instruction and take the first 8 bytes like solidity does. I think we can get by for now with a random 8 byte sequence until we have a standard abi

jordansexton: Cool, will do.

jstarry (approved):

PR #1814, opened 2021-05-27
stake-pool: Add program_id to all instruction ctors

joncinque:
#### Problem

We're going to have more than one stake pool program, and some of the instruction constructors don't allow you to set a program id.

#### Solution

Add `program_id` as a parameter on all constructors.

cc @ruuda let me know how this looks

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -287,6 +287,7 @@ fn command_vsa_create(
         &[
             // Create new validator stake account address
             spl_stake_pool::instruction::create_validator_stake_account_with_vote(
+                &spl_stake_pool::id(),
                 &stake_pool_address,
                 &config.staker.pubkey(),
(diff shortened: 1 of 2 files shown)


ruuda (approved):
Great, thanks!

jon-chuang:
Hey, @joncinque , would you be able to give an explanation of why we'd like to have multiple stake pool programs?

joncinque:
@jon-chuang Yep, that's exactly the point, they will have their own program, which they can maintain and upgrade in lockstep with the Lido program.  This way, the foundation / community stake pools and program can operate totally independently.  Since Lido users are trusting the Lido program already with their deposits, in their case it makes sense to have a trusted stake pool program to go with it, that can't get changed in any undesired way.  They can also control when to update to new code as needed, after it passes their regression testing suite and any other checks needed.

PR #1831, opened 2021-05-28
stake-pool: Add preferred deposit / withdraw validator accounts

joncinque:
#### Problem

Stake pools can be arbitraged by depositing to one validator and withdrawing from another.

#### Solution

Cover a lot of potential attack vectors, and allow the stake pool staker to restrict deposits and withdrawals to specific validators.  If the preferred deposit validator is set, all deposits must go through it.  If a preferred withdrawer is set, all withdrawals must go through it if it has *any* stake at all.  If it's empty, then normal rules apply.

cc @aeyakovenko

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -32,6 +32,7 @@ use {
         self,
         borsh::get_instance_packed_len,
         find_stake_program_address, find_withdraw_authority_program_address,
+        instruction::PreferredValidatorType,
         stake_program::{self, StakeState},
         state::{Fee, StakePool, ValidatorList},
     },
@@ -462,6 +463,34 @@ fn command_decrease_validator_stake(
         &[config.fee_payer.as_ref(), config.staker.as_ref()],
         recent_blockhash,
     );
+    send_transaction(&config, transaction)?;
+    Ok(())
+}
+
+fn command_set_preferred_validator(
+    config: &Config,
+    stake_pool_address: &Pubkey,
+    preferred_type: PreferredValidatorType,
+    vote_address: Option<Pubkey>,
+) -> CommandResult {
+    let stake_pool = get_stake_pool(&config.rpc_client, stake_pool_address)?;
+    let mut transaction = Transaction::new_with_payer(
+        &[spl_stake_pool::instruction::set_preferred_validator(
+            &spl_stake_pool::id(),
+            &stake_pool_address,
+            &config.staker.pubkey(),
+            &stake_pool.validator_list,
+            preferred_type,
+            vote_address,
+        )],
+        Some(&config.fee_payer.pubkey()),
+    );
+
+    let (recent_blockhash, fee_calculator) = config.rpc_client.get_recent_blockhash()?;
+    check_fee_payer_balance(config, fee_calculator.calculate_fee(&transaction.message()))?;
+    let mut signers = vec![config.fee_payer.as_ref(), config.staker.as_ref()];
+    unique_signers!(signers);
+    transaction.sign(&signers, recent_blockhash);
     send_transaction(&config, transaction)?;
     Ok(())
 }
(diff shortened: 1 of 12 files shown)


jon-chuang:
I don't think this is a good idea as the pool's intention may not be to allow users to arbitrarily "bankrupt" a single validator. It's also fairly complex from the perspective of users.

From the perspective of the stake pool manager, needing to respond whenever a particular validator's funds have been "bankrupted" and defaulting to the previous behaviour doesn't seem to solve any issue whatsover.

Further, once warp stake is implemented in stake program V2, this sort of workaround is pointless since in that case, a user can ask for withdrawals of stake directly to any stake account on an arbitrary validator of preference at the turn of the epoch. 

Thus, I don't think one can solve the problem adequately, unless one resorts to the reserve model.

jon-chuang:
Here is the alternative extension of the current design which I think would solve these issues:
https://hackmd.io/@gPH8I-Z5RcSMH2KTXLDeDg/HyQ5TRJ5u

joncinque:
@jon-chuang Absolutely, the concept isn't to make the free-warp attack impossible, but rather to make it much more expensive to carry it out.  If the pool doesn't want to "bankrupt" a validator account, then they likely would not want to use this feature.

> It's also fairly complex from the perspective of users.

I'm not sure I understand what you mean about this.

> From the perspective of the stake pool manager, needing to respond whenever a particular validator's funds have been "bankrupted", and failing that, defaulting to the previous behaviour, with attendant security issues, doesn't seem to solve anything.

The pool manager should be rebalancing every epoch, so even if a validator is bankrupt one epoch, it should get refilled soon after.  And if the preferred deposit accout is the same as the withdraw, then the account will get refilled at the start of the attack, rendering it impossible.  Remember, to carry out the attack, you need to deposit in one place and withdraw from another.  If those places are the same, then it's impossible.  The only possibility is if you already have stake in the pool and have enough to deplete the preferred withdraw account.

> Further, once warp stake is implemented in stake program V2, this sort of workaround is pointless since in that case, a user can ask for withdrawals of stake directly to any stake account on an arbitrary validator of preference at the turn of the epoch.

The warp stake instruction will work differently, since it will take the previous validator into account when figuring out your rewards.  With the stake pool attack we've been discussing, you would instantly get normal active stake on another validator, bypassing any potential slashing events.

> A program that respects epoch boundaries is what's needed here.

Heck yeah, absolutely :smile:

jon-chuang:
Hey Jon thanks again for the detailed replies.

Some thoughts:
If this direction is to be pursued, one idea could be to have a pool parameter `single_validator_withdrawal_limit`, which sets the percentage withdrawal/deposit limit from a single Validator before cycling to the next Validator, and thereafter proceeding in round-robin fashion. If funds are added, it goes the opposite direction.

For instance, if we set the value to be 10, then if Validator A has has had 999 out of initial 10,000 SOL withdrawn, a new withdrawal of 10 SOL would withdraw 1 SOL from A, and 9 SOL from B. 

At the very least, this keeps the allocation fairly balanced, and limits the degree a single delegator can "skew" stake.

So in addition to the idea of this PR, one also has an automatic "round robin" behaviour and a deposit/withdrawal limit to each Validator.

I could probably live with that, except for the "manual" part of the picture where the user has to do this staking, activating, and depositing themselves.

The other downside is that a user would hardly have stake that's already happened to have activated in the right Validator at the right time...

To facilitate manager reshuffling, a pool can also designate an allocator authority, which can circumvent the above round robin and limits.

joncinque:
@jon-chuang Hm, that's an interesting idea, and would solve some of the balancing issues.  This all assumes that a stake pool wants even distribution between validators, of course.

Let's keep in mind that the usability for the deposit side of this would extremely tricky -- remember that people need to warm up stake to a validator in order to deposit it.  With a model like this, to safely deposit, they would need to create `total_deposited_stake / single_validator_churn_limit` stake accounts, respecting the order of validators.  And if there are a lot of validators, they just need to hope that deposits are constantly happening so that "their turn" can come up to deposit, and that no one else is trying to deposit.

From the withdraw side, it'll be much more workable.  If multiple people are trying to withdraw at the same time, they may just see a bunch of failures, ie. you and me both try to withdraw from the "current validator", one of us fails and has to try again on the next one.

jon-chuang:
Yeah. I was thinking it could be good just for the withdrawal side. Maybe for the deposit side, you could set a limit for how much can be deposited to one Validator that deviates from the desired distribution. And you could have the same thing for withdrawal.

> This all assumes that a stake pool wants even distribution between validators, of course.

This is not quite correct. If you read carefully, I suggest that the limit is in terms of the percentage of the current amount staked in that Validator. So it proportionally increments or decrements the values in a round robin fashion, based on the initial distribution (say at the epoch boundary, when the manager redistributes).

joncinque:
Ah fair enough, proof that I did not read carefully :smile:

jon-chuang:
Haha no worries, fastest way to get the right answer is to tentatively state the wrong one 😇

PR #1855, opened 2021-06-03
token-docs: add token registry

joncinque:
#### Problem

Someone was asking about a token registry, and the docs don't mention the token-list.

#### Solution

Mention it!

### DIFF
--- a/docs/src/token.md
+++ b/docs/src/token.md
@@ -838,7 +838,7 @@ At the moment there exist two solutions for Token Mint registries:
 
 * hard coded addresses in the wallet or dapp
 * [spl-token-registry](https://www.npmjs.com/package/@solana/spl-token-registry)
-package, maintained at [solana-labs/token-list](https://github.com/solana-labs/token-list)
+package, maintained at [https://github.com/solana-labs/token-list](https://github.com/solana-labs/token-list)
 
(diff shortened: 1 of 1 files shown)


docs/src/token.md
```diff
@@ -834,7 +834,13 @@ The sender's wallet must not require that the recipient's main wallet address
 hold a balance before allowing the transfer.
 
 ### Registry for token details
-At the moment Token Mint addresses need to be hard coded by each wallet.  **Improving this situation is a work in progress.**
+At the moment there exist two solutions for Token Mint registries:
+
+* hard coded addresses in the wallet or dapp
+* [spl-token-registry](https://www.npmjs.com/package/@solana/spl-token-registry)
+package, maintained at [solana-labs/token-list](https://github.com/solana-labs/token-list)
```
CriesofCarrots, line 841:
```suggestion
package, maintained at [https://github.com/solana-labs/token-list](https://github.com/solana-labs/token-list)
```
I slightly prefer the explicit url here, but don't feel too strongly about it, so take it or leave it.

CriesofCarrots (approved):

docs/src/token.md
joncinque: Makes sense to me, thanks!

PR #1888, opened 2021-06-08
Update some api urls

CriesofCarrots:


CriesofCarrots:
CI failure seems maybe related to build caches. But anyway, this PR doesn't touch rust code.

PR #1893, opened 2021-06-09
ci: Split out test-bpf into separate workflows for each program

joncinque:
We're running out of space on the CI runners, so it's time to finally split out each program into its own workflow.

Currently, this only moves the `cargo test-bpf` and JS tests (where applicable), so that we can still run `cargo test` and `cargo build` on the full workspace.  If we still run out of space, then we'll need to do the build + test in separate dirs.

Programs to do:
 
- [x] token
- [x] swap
- [x] lending
- [x] binary-oracle-pair
- [x] examples (rust and c)
- [x] feature-proposal
- [x] governance
- [x] libraries
- [x] memo
- [x] name-service
- [x] record
- [x] shared-memory
- [x] stake-pool


joncinque:
There's a lot of copy-paste here on the yml files, which we can reduce once github composite actions support "uses" clauses.  I had it earlier with all of the simple `cargo-test-bpf` stages as a matrix, but I liked the idea of only running required `cargo test-bpf` stages, where code had been changed, so I split them all off into separate files.

To trigger each build, I also made a tiny change of updating a dev dependency or readme file.  There's a slightly bigger change to the token-lending test, which wasn't passing in its current form.

joncinque:
Now we finish in 13 minutes instead of 50 :smile:

CriesofCarrots (approved):
Thank you so much for working on this! The CI speed-up is going to be wonderful, and hopefully we can slim down the workflows code as github actions improve their api.

PR #1895, opened 2021-06-10
Minor fixes in name-service

Lcchy:
Fix a typo in the name-service program and account for record state space in registry account allocation (non-breaking).

Update the reverse registry lookup for the twitter handle name-service lookup.


name-service/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@solana/spl-name-service",
-  "version": "0.1.2",
+  "name": "@bonfida/spl-name-service",
```
bartosz-lipinski, line 2: can we keep this as @solana

bartosz-lipinski (dismissed):
Looks awesome. Just undo the change to the package name.

name-service/js/package.json
Lcchy: Yes for sure!

PR #1904, opened 2021-06-12
Also clone program data for stake pool test-validator

jon-chuang:
This is needed for the program to work


joncinque:
Thanks for the contribution, we'll eventually have to fix up the test validator to properly copy all accounts for upgradeable programs.

PR #1947, opened 2021-06-22
Governance: use runtime program_id

SebastianBor:
### Summary 

Remove static program_id to support runtime clones of the program


joncinque (approved):
Looks good!

PR #1948, opened 2021-06-22
stake-pool: Add lockup support

joncinque:
#### Problem

Stake pools don't support lockups.

#### Solution

Add lockup support!  The main idea is to add the lockup inside of the stake pool itself and then add the custodian as an optional signer for the required instructions.  Since the custodian needs to sign any changes of the stake account withdraw authority, that means we need to add the custodian to:
* add validator to pool
* remove validator from pool
* deposit
* withdraw

This also adds CLI support, mimicking how the main solana CLI does lockups

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -10,9 +10,10 @@ use {
         Arg, ArgGroup, SubCommand,
     },
     solana_clap_utils::{
-        input_parsers::{keypair_of, pubkey_of},
+        input_parsers::{keypair_of, pubkey_of, unix_timestamp_from_rfc3339_datetime, value_of},
         input_validators::{
-            is_amount, is_keypair, is_keypair_or_ask_keyword, is_parsable, is_pubkey, is_url,
+            is_amount, is_keypair, is_keypair_or_ask_keyword, is_parsable, is_pubkey,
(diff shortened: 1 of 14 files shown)


stake-pool/cli/src/main.rs
```diff
@@ -98,9 +101,12 @@ fn send_transaction(
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
 fn command_create_pool(
```
mvines, line 105: For pool creation, I feel that a better way could be to provide a prototypical stake account.  The program then examines that stake account on-chain and copies over the Lockup.  This way there's 0% chance of user error when specifying the lockup parameters.

mvines:
Code looks great overall, but I'll pause until we resolve the design discussion from Discord regarding whether stake accounts remain locked while held by the pool

joncinque:
Seems to me that the best option is to remove the lockup on deposit and reapply on withdraw.  Should we loop anyone else into this discussion @mvines ?

mvines:
> Seems to me that the best option is to remove the lockup on deposit and reapply on withdraw. Should we loop anyone else into this discussion @mvines ?

@aeyakovenko - do you want to chime in on this during your vacation?

PR #1950, opened 2021-06-22
stake-pool: Split transient stake account creation

joncinque:
#### Problem

The runtime was updated in 1.7 to not allow 0-lamport account creation, creating an issue reported at https://github.com/solana-labs/solana-program-library/pull/1617#issuecomment-864482976

Stake pools still leans on 0-lamport account creation for transient stake accounts.

#### Solution

Don't do `create_account` with 0 lamports when working with transient stake accounts, and instead separate out allocate + assign.

cc @keemy

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -132,6 +132,30 @@ fn check_account_owner(
     }
 }
 
+/// Create a transient stake account without transferring lamports
+fn create_transient_stake_account<'a>(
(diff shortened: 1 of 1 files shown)


keemy:
sick thanks! yeah sorry didn't know the system program well enough to do this myself.

joncinque:
@keemy np! You can start using this immediately as is, and I'll be releasing the crates within the next couple of days

PR #1958, opened 2021-06-24
Update all solana dependencies to 1.7.3, fix issues

joncinque:
Little issues cropped up on the update:

* stake-pool: new error `AccountNotRentExempt` needs to be checked
* token-lending: need to use `ProgramTest.prefer_bpf(false)` for flash loan tests
* token(js): flow does not appreciate passing an instance method to a function, so use an anonymous function
* update-solana-dependencies.sh: also update the version used in ci/solana-version.sh for consistency


CriesofCarrots:
@joncinque , do you mind testing solana v1.6 `downstream-projects` build with this locally?

t-nelson (dismissed):
lgtm. thanks!

PR #1965, opened 2021-06-25
Test automerge with `all_github_action_checks` twice

joncinque:
#### Problem

Automerge may be broken due to the split of GitHub Actions workflows -- once `all_github_action_checks` passes on the main `pull-request.yml` workflow, it may merge before finishing the program specific workflow.

#### Solution

This may not work, so the test is to add `all_github_action_checks` to the token workflow and just change some whitespace to trigger it.  If automerge waits for both, then we're all good!  If not, back to the drawing board.

### DIFF
--- a/.github/workflows/pull-request-token.yml
+++ b/.github/workflows/pull-request-token.yml
@@ -12,6 +12,14 @@ on:
     - 'token/**'
 
 jobs:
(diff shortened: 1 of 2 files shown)


joncinque:
nevermind, this won't work

PR #1984, opened 2021-06-28
Governance: Change timestamps and time periods to UnixTimestamp

SebastianBor:
#### Summary 
Change all `Proposal` lifetime timestamps and configurable time periods (`voting_time`, `instruction_hold_up_time`) from slots to UnixTimestamp to make them human friendly

### DIFF
--- a/governance/program/tests/program_test/mod.rs
+++ b/governance/program/tests/program_test/mod.rs
@@ -1657,7 +1657,7 @@ impl GovernanceProgramTest {
         let mut n = 1;
 
         while clock.unix_timestamp <= unix_timestamp {
-            // Since the exact time is not deterministic wrap by arbitrary 400 slots until we pass the timestamp
+            // Since the exact time is not deterministic keep wrapping by arbitrary 400 slots until we pass the requested timestamp
             self.context.warp_to_slot(clock.slot + n * 400).unwrap();
 
             n = n + 1;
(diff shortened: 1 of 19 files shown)


governance/program/tests/program_test/mod.rs
```diff
@@ -1505,6 +1508,39 @@ impl GovernanceProgramTest {
...
+            .await
+    }
+
+    #[allow(dead_code)]
+    pub async fn advance_clock_past_timestamp(&mut self, unix_timestamp: UnixTimestamp) {
+        let mut clock = self.get_clock().await;
+        let mut n = 1;
+
+        while clock.unix_timestamp <= unix_timestamp {
+            // Since the exact time is not deterministic keep wrapping by arbitrary 400 slots until we pass the requested timestamp
```
joncinque, line 1523: Does this actually work to advance the timestamp the amount expected?  When we warp, the new bank is supposed to calculate the new timestamp based on voting activity, but without voting activity, I'm not sure how it calculates.

joncinque (approved):
Looks good, just the one question about advancing the clock, since I didn't think that was possible!

governance/program/tests/program_test/mod.rs
SebastianBor: Yes, it does work. I've looked into the code and I don't fully understand it but the updated timestamp is calculated by [get_timestamp_estimate](https://github.com/solana-labs/solana/blob/c1b9d40a64e57375e4298d8d96345a4bb17b5de3/runtime/src/bank.rs#L2083) and I empirically confirmed that for the given starting conditions (slot, epoch, warp_to_slot) the result is deterministic. 
vote_accounts() is not empty in that call however the timestamp is moved by the drift correction logic.

joncinque: Ah, very cool -- nice job!

PR #1996, opened 2021-06-29
lending: JavaScript / TypeScript client library

jordansexton:
Old client JS has been removed since the functions it had can be performed easily with the CLI.

TypeScript bindings for de/serialization and instructions have been added and loosely tested during the redevelopment of Oyster Lending.

All instructions are up to date with the program, except for `FlashLoan` which hasn't been implemented in JS yet.

Generated TS docs on Github Pages @ https://solana-labs.github.io/solana-program-library/token-lending/

NPM package published @ https://www.npmjs.com/package/@solana/spl-token-lending


token-lending/js/package.json
```diff
@@ -1,72 +1,55 @@
...
+        "lib"
+    ],
+    "scripts": {
+        "build": "rm -rf lib/* && tsc",
+        "deploy": "gh-pages -d docs -e token-lending",
+        "docs": "rm -rf docs/* && typedoc",
+        "lint": "eslint . --ext .ts && prettier --check '**/*.{ts,js,json}'",
+        "lint:fix": "eslint . --ext .ts --fix && prettier --write '**/*.{ts,js,json}'"
+    },
+    "dependencies": {
```
jstarry, line 30: Can you add a dependency on `buffer-layout` here?

token-lending/js/src/instructions/borrowObligationLiquidity.ts
```diff
@@ -0,0 +1,58 @@
...
+    borrowReserveLiquidityFeeReceiver: PublicKey,
+    obligation: PublicKey,
+    lendingMarket: PublicKey,
+    lendingMarketAuthority: PublicKey,
+    obligationOwner: PublicKey,
+    hostFeeReceiver?: PublicKey
+): TransactionInstruction => {
+    const dataLayout = struct([u8('instruction'), u64('liquidityAmount')]);
+
+    const data = Buffer.alloc(dataLayout.span);
```
jstarry, line 23: This requires the client to either be a node app or have a bundler which can polyfill this global. Have you had any issues with using it in oyster-lending?

token-lending/js/src/state/obligation.ts
```diff
@@ -0,0 +1,132 @@
...
+        lastUpdate,
+        lendingMarket,
+        owner,
+        depositedValue,
+        borrowedValue,
+        allowedBorrowValue,
+        unhealthyBorrowValue,
+        deposits,
+        borrows,
+    } as Obligation;
```
jstarry, line 125: nit: this could break easily in the future, is it necessary? Could you do `const obligation: Obligation = ...`

jstarry (dismissed):
Looks very clean! I didn't comb over every instruction/state but I assume any bugs there will pop up while integrating into the UI

token-lending/js/package.json
jordansexton: Good catch, done.

token-lending/js/src/state/obligation.ts
jordansexton: Changed to `const obligation: Obligation`, though it probably isn't necessary anyway as the function declares the return type.

token-lending/js/src/instructions/borrowObligationLiquidity.ts
jordansexton: Oyster uses create-react-app which handles the bundling (I think using Webpack) so that takes care of the polyfill. We could us Rollup for this library and build Node and browser versions, but I don't mind keeping this library simple and leaving the choice to the app that depends on it. I'll make a note in the readme for now, and add a bundler to the package if we get feedback.

jordansexton: I looked into this and Webpack 5 recently removed support for automatically polyfilling this, which makes it more annoying if you're using this as expected (usually for the web). I didn't want to add a bundler, but I'd rather do that than make users deal with it.

jordansexton: Added rollup + config. Will test using Oyster Lending and follow up in a new PR.

PR #2002, opened 2021-06-30
Governance: future proofing for v2

SebastianBor:
#### Summary 
The PR contains a set of changes which aim to future proof `Governance` program accounts for future versions

1. `VoteThresholdPercentageType` - vote threshold percentage type to support quorum based votes in future versions.

2. `VoteWeightSource` - vote weight source to support account snapshot based voter weights in future versions. With snapshots it'll be possible to vote without depositing governing tokens. 
3. `InstructionExecutionStatus` - status to support failed instructions in future versions and prevent Proposals permanently staying in `Succeeded` or `Executing` state.
4. `InstructionExecutionFlags` - execution flags to support ordered and transactional instructions in future versions.
5. `proposal_cool_off_time`- time period to allow proposal cancellation after voting. 
3. Move all fields with dynamic size to the end of an account layout to support search on all fixed fields
4. Add reserved padding space to permanent accounts. The permanent accounts like `Realm` or `Governance` once created would stay as is forever and hence will be the most difficult to change in the future. 
5. Add missing tests for withdrawals and serialization

### DIFF
--- a/governance/program/src/error.rs
+++ b/governance/program/src/error.rs
@@ -13,7 +13,7 @@ use thiserror::Error;
 pub enum GovernanceError {
     /// Invalid instruction passed to program
     #[error("Invalid instruction passed to program")]
-    InvalidInstruction = 500,
+    InvalidInstruction = 500, // Start Governance custom errors from 500 to avoid conflicts with programs invoked via CPI
 
     /// Realm with the given name and governing mints already exists
     #[error("Realm with the given name and governing mints already exists")]
--- a/governance/program/src/error.rs
+++ b/governance/program/src/error.rs
@@ -13,7 +13,7 @@ use thiserror::Error;
 pub enum GovernanceError {
     /// Invalid instruction passed to program
     #[error("Invalid instruction passed to program")]
-    InvalidInstruction,
+    InvalidInstruction = 500,
 
     /// Realm with the given name and governing mints already exists
     #[error("Realm with the given name and governing mints already exists")]
--- a/governance/program/src/error.rs
+++ b/governance/program/src/error.rs
@@ -267,6 +267,10 @@ pub enum GovernanceError {
     /// Given VoteWeightSource is not supported
     #[error("Given VoteWeightSource is not supported")]
     VoteWeightSourceNotSupported,
+
+    /// Proposal cool off time is not supported
+    #[error("Proposal cool off time is not supported")]
+    ProposalCoolOffTimeNotSupported,
 }
 
 impl PrintProgramError for GovernanceError {
--- a/governance/program/src/error.rs
+++ b/governance/program/src/error.rs
@@ -259,6 +259,10 @@ pub enum GovernanceError {
     /// Current token owner must sign transaction
     #[error("Current token owner must sign transaction")]
     TokenOwnerMustSign,
+
+    /// Given VoteThresholdPercentageType is not supported
+    #[error("Given VoteThresholdPercentageType is not supported")]
+    VoteThresholdPercentageTypeNotSupported,
 }
 
 impl PrintProgramError for GovernanceError {
--- a/governance/program/src/error.rs
+++ b/governance/program/src/error.rs
@@ -263,6 +263,10 @@ pub enum GovernanceError {
     /// Given VoteThresholdPercentageType is not supported
     #[error("Given VoteThresholdPercentageType is not supported")]
     VoteThresholdPercentageTypeNotSupported,
+
+    /// Given VoteWeightSource is not supported
+    #[error("Given VoteWeightSource is not supported")]
+    VoteWeightSourceNotSupported,
 }
 
 impl PrintProgramError for GovernanceError {
--- a/governance/program/src/state/enums.rs
+++ b/governance/program/src/state/enums.rs
@@ -92,3 +92,20 @@ impl Default for ProposalState {
         ProposalState::Draft
     }
 }
+
+/// The type of the vote threshold percentage used to resolve a vote on a Proposal
+#[repr(C)]
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub enum VoteThresholdPercentageType {
+    /// Voting threshold of Yes votes in % required to tip the vote
+    /// It's the percentage of tokens out of the entire pool of governance tokens eligible to vote
+    /// Note: If the threshold is below or equal to 50% then an even split of votes ex: 50:50 or 40:40 is always resolved as Defeated
+    /// In other words +1 vote tie breaker is always required to have a successful vote
+    YesVote,
+
+    /// The minimum number of votes in % out of the entire pool of governance tokens eligible to vote
+    /// which must be cast for the vote to be valid
+    /// Once the quorum is achieved a simple majority (50%+1) of Yes votes is required for the vote to succeed
+    /// Note: Quorum is not implemented in V1
+    Quorum,
+}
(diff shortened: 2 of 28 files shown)


governance/program/src/error.rs
```diff
@@ -13,7 +13,7 @@ use thiserror::Error;
 pub enum GovernanceError {
     /// Invalid instruction passed to program
     #[error("Invalid instruction passed to program")]
-    InvalidInstruction,
+    InvalidInstruction = 500, // Start Governance custom errors from 500 to avoid conflicts with programs invoked via CPI
```
joncinque, line 16: We should really come up with a better custom error type that automatically adds the program id in some way

governance/program/src/state/enums.rs
```diff
@@ -92,3 +92,60 @@ impl Default for ProposalState {
...
+/// The source of voter weights used to vote on proposals
+#[repr(C)]
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub enum VoteWeightSource {
+    /// Governing token deposits into the Realm are used as voter weights
+    Deposit,
+    /// Governing token account snapshots as of the time a proposal entered voting state are used as voter weights
+    /// Note: Snapshot source is not supported in the current version
+    /// Support for account snapshots are required in solana and/or arweave as a prerequisite
+    Snapshot,
```
joncinque, line 122: How would the design need to change to properly support this? Would you need to change any other structs to properly snapshot the accounts at the time the proposal was created?

governance/program/src/state/enums.rs
```diff
@@ -92,3 +92,60 @@ impl Default for ProposalState {
...
+    /// Instruction execution failed
+    /// Note: Error status is not supported yet because when CPI call fails it always terminates parent instruction
+    /// We either have to make it possible to change that behavior or add an instruction to manually set the status
+    Error,
+}
+
+/// Instruction execution flags defining how instructions are executed for a Proposal
+#[repr(C)]
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub enum InstructionExecutionFlags {
```
joncinque, line 141: Should you have an additional flag for the current behavior of `Unordered`?  Then this doesn't need to be an `Option` in the Proposal.

governance/program/src/state/governance.rs
```diff
@@ -22,20 +24,30 @@ pub struct GovernanceConfig {
...
-    /// It's the percentage of tokens out of the entire pool of governance tokens eligible to vote
-    // Note: If the threshold is below or equal to 50% then an even split of votes ex: 50:50 or 40:40 is always resolved as Defeated
-    // In other words +1 vote tie breaker is required to have successful vote
-    pub yes_vote_threshold_percentage: u8,
+    /// The type of the vote threshold used for voting
+    /// Note: In the current version only YesVote threshold is supported
+    pub vote_threshold_percentage_type: VoteThresholdPercentageType,
+
+    /// The vote threshold value of the type defined by vote_threshold_percentage_type
+    pub vote_threshold_percentage: u8,
```
joncinque, line 32: tiny little nit-idea: you could have `VoteThresholdType` be an enum-with-value, like `VoteThresholdType::YesVote(u8)`, since these two fields seem strongly intertwined.

governance/program/src/state/proposal_instruction.rs
```diff
@@ -96,11 +99,15 @@ pub struct ProposalInstruction {
 
     /// Executed at flag
     pub executed_at: Option<UnixTimestamp>,
+
+    /// Instruction execution status
+    /// Note: The field is not used in the current version
+    pub execution_status: Option<InstructionExecutionStatus>,
```
joncinque, line 105: What will this be for?  Maybe you could have `InstructionExecutionStatus` as `NotRun`, `Success`, and `Error`, and then this doesn't need to be an `Option`

governance/program/src/state/token_owner_record.rs
```diff
@@ -36,22 +36,25 @@ pub struct TokenOwnerRecord {
...
     /// The number of votes cast by TokenOwner but not relinquished yet
     /// Every time a vote is cast this number is increased and it's always decreased when relinquishing a vote regardless of the vote state
     pub unrelinquished_votes_count: u32,
 
     /// The total number of votes cast by the TokenOwner
     /// If TokenOwner withdraws vote while voting is still in progress total_votes_count is decreased  and the vote doesn't count towards the total
     pub total_votes_count: u32,
+
+    /// Reserved space for future versions
+    pub reserved: u64,
```
joncinque, line 48: nit: if you want to reserve space as clearly as possible, you could declare it as `[u8; 8]`, making it totally clear that it's just buffer space.  This way, if you add in a `u16` field, it's easy to then change the reserved space to `[u8; 6]`

joncinque (dismissed):
Looks great, just a few questions, none of them blocking

governance/program/src/state/token_owner_record.rs
SebastianBor: good idea, done

governance/program/src/state/proposal_instruction.rs
SebastianBor: I was debating with myself whether we should use `Option` to indicate `None` or a native enum value and it was a close call which you tipped in favour of enum value :).

I added the status to indicate whether the instruction was executed successfully or it returned an error. It means the default `None` status indicates it hasn't been attempted to run yet.
It is required to prevent a `Proposal` being permanently displayed as being in `Succeeded` or `Executing` state when one of its instruction is permanently broken and can't be executed successfully.
I left it as unused in this version because currently there is no clear way to determine whether instruction execution was successful because CPI call error always terminates parent transaction. 
I'm thinking the best option would be to make it optional in the runtime and make it possible to opt out of that behaviour and allow the parent to get CPI call errors.
However if that turns out to be impossible, for whatever reason, then the only option is to add a program instruction to manually flag instruction as broken.

governance/program/src/state/governance.rs
SebastianBor: Cool idea, thx. I keep forgetting about this cool enum feature in Rust.

governance/program/src/state/enums.rs
SebastianBor: Option changed to native None value in enum

governance/program/src/state/enums.rs
SebastianBor: This is a difficult one because we are talking about a feature which doesn't exist yet. However my take on it is that it should be externalised and `Governance` program would make a CPI call to get voter's vote weight.
For example: 
1) We can create `Snapshot` program which would create and store account snapshot natively on Solana. 
2) We store snapshots on arweave and make calls to a program that would allow us to read the snapshots from there.

In the cases above all we need is `voting_at_slot` field so we know the `as_of_slot` we should query the snapshots for and that field was added in this PR. Other than that I think we don't need any struct changes.

Saying all that, I'm not sure about this snapshot strategy. This is how it's done on `ETH` but it allows you to play the system and create a proposal at a time convenient for you. 
Some other ideas from users are for example to use the time you have staked your tokens in a pool as a weight on top of the amount of tokens. But in this case I also think it'll be just a CPI call to whatever program can determine the weight.

governance/program/src/error.rs
SebastianBor: Can't agree more. The fact that we can't surface correct errors from programs to users makes a very terrible UX. 

I looked into [ProgramError](https://github.com/solana-labs/solana/blob/master/sdk/program/src/program_error.rs) and in theory it should be just another error type `ProgramError:ProgramError(pubkey,u32`)  in addition to `ProgramError:Custom(u32)` but maybe I'm oversimplifying it without full understanding of the stack

PR #2003, opened 2021-07-01
we added a secondary oracle and functionality to deposit asset + ctoken in same instruction 

nope-finance:
cc. @jordansexton


token-lending/cli/src/main.rs
```diff
@@ -311,7 +321,7 @@ fn main() {
         .get_matches();
 
     let mut wallet_manager = None;
-    let config = {
+    let mut config = {
```
jordansexton, line 324: It doesn't look like this gets mutated, does it?

token-lending/program/src/processor.rs
```diff
@@ -1769,6 +1945,18 @@ fn get_pyth_product_quote_currency(pyth_product: &pyth::Product) -> Result<[u8;
     Err(LendingError::InvalidOracleConfig.into())
 }
 
+fn get_price(
+    switchboard_feed_info: &AccountInfo,
+    pyth_price_account_info: &AccountInfo,
+    clock: &Clock,
+) -> Result<Decimal, ProgramError> {
+    let pyth_price = get_pyth_price(pyth_price_account_info, clock).unwrap_or_default();
+    if pyth_price != Decimal::zero() {
```
jordansexton, line 1954: What primarily causes this to occur such that it fails over to Switchboard for the price? Is it staleness of the Pyth price?

jordansexton (commented):
I took a look at this. Interesting set of changes, especially the ability to deposit liquidity directly to an obligation.

As for the Switchboard functionality, I'm curious what/when you want it for. Has Pyth not been highly reliable?

token-lending/cli/src/main.rs
nope-finance: yeh but i found that for testing I was needing to mutate it a lot (dryrun doesn't work for the init reserve because there are multiple txns so the later just always fail because the accounts aren't created so i had to have it run the first 2 for real then switch to dryrun)

token-lending/program/src/processor.rs
nope-finance: usually staleness yes, we have had several instances of this happening on devnet, a bunch of feeds were down for 12 hours once and one feed was down for ~3 days

nope-finance:
I think the idea with switchboard is for now just to have a backup because we did run in to some issues, but in the future we might want to have the ability for each reserve to specify an oracle if we plan on having some exotic assets, switchboard seemed really onboard with adding basically whatever we wanted so that flexibility seemed nice.

PR #2015, opened 2021-07-02
Bump solana and borsh crates

CriesofCarrots:
Borsh bump to v0.9 needs to stay in sync with the borsh version in solana crates.


joncinque (approved):
Looks good! Thanks for taking care of this

PR #2039, opened 2021-07-07
Governance: SetGovernanceConfig instruction

SebastianBor:
#### Implemented instructions:

- `SetGovernanceConfig` - Instruction allows to update `GovernanceConfig`. The instruction must be signed by `Governance` PDA and hence can only be executed as an instruction of a `Proposal` for the `Governance` the `Proposal` belongs to. It allows self governance and tweaking of the `GovernanceConfig` parameters for a given community. For example it can be used to adjust vote threshold which makes sense for the given community engagement.


joncinque (approved):
Looks great!

PR #2041, opened 2021-07-08
stake-pool: Optimizations for 4k validators

joncinque:
#### Problem

Stake pools can't support more than 200 validators at the moment because of memory and compute budget limitations during serialization and deserialization.

#### Solution

Update the on-chain serde to support up to 4k validators!  Since the instructions that involve iterating through the whole pool can't do too much else, this also introduces a new `CleanupRemovedValidatorEntries` instruction, which just cleans up any validator that's marked `ReadyForRemoval`.  There's no requirement to perform this except when validator entries are ready to be removed, and it's permissionless.  Likely, the stake pool staker will call this whenever needed to free up space in their validator list.

Additionally, there's some unsafe Rust used for very specific reasons:

* casting [u8] slices to ValidatorStakeInfo entries (mut and const)
* memmove during the "cleanup" instruction
* memcmp to find entries in the list (not unsafe but should be flagged)

This is mostly done through a `BigVec` type, which was written to be generic and reusable with any `Pack` types.  It doesn't have the full `Vec` interface, but I think a lot of good bits are there to be reused!  It was certainly a nice refresher on lifetimes.

Unfortunately, these changes require non-backwards compatible changes to the structs.  It's really tricky to share an account with variable space with one of these big vecs.  For example, if the struct has an `Option<u8>` somewhere, that can be serialized either to `[0]` for `None` or `[1, 255]` for `Some(255)`, which breaks the in-place deserialization.

### DIFF
--- a/stake-pool/program/src/big_vec.rs
+++ b/stake-pool/program/src/big_vec.rs
@@ -1,9 +1,11 @@
-//! Big vector type, used with Borsh vectors that can't be serde'd
+//! Big vector type, used with vectors that can't be serde'd
 
 use {
     arrayref::array_ref,
     borsh::{BorshDeserialize, BorshSerialize},
-    solana_program::{program_error::ProgramError, program_pack::Pack},
+    solana_program::{
+        program_error::ProgramError, program_memory::sol_memmove, program_pack::Pack,
+    },
     std::marker::PhantomData,
 };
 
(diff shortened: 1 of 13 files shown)


stake-pool/program/src/big_vec.rs
```diff
@@ -0,0 +1,245 @@
...
+            let end_index = self.current_index + T::LEN;
+            let value = Some(unsafe {
+                &mut *(self.inner.data[self.current_index..end_index].as_ptr() as *mut T)
+            });
+            self.current += 1;
+            self.current_index = end_index;
+            value
+        }
+    }
+}
```
mvines, line 245: Mildly terrifying that there's no tests for the partial vec reimplementation.  Can we lift some from the Rust `vec` test suite so we don't need to reinvent them?

mvines:
I looked over most of the code (skimmed the tests), makes sense to me and also am sad that it's come to this. Good thing we just finished that security audit 🙃 

I feel that we maybe could have done this slightly differently but having multiple validator list accounts (with ~200 validators in each) and avoid the custom serde and big_vec, but that approach would have other issues, so this design direction wfm all things considered.

stake-pool/program/src/big_vec.rs
joncinque: Yep, on it, looks like we have 2.7k lines of tests to work with :grin:

joncinque:
Trust me, I didn't want to have to resort to this either.  This went through a few iterations, starting with just deserializing bits using Borsh and then reserializing those back in, but most times, to do anything, we have to iterate through the whole thing somehow.  For example, just finding the last element in the array by casting to `Pubkey` and doing `==` blew out the limit, which is why `memcmp` was necessary.

We could have definitely done something using rollup accounts, but that would fundamentally change the design and involve a lot more steps / instructions to do right.  Since we really want stake pools to increase decentralization, I wanted to maximize the supported number of validators for the future of the network :grin: 

This experience has made me wonder what we should do about in-place serde for other programs that need big accounts.  Serum uses bytemuck and other low-level libraries to manage their big buffers, which certainly works, and other programs use different libraries like Cap'n Proto, Protobuffers, or FlatBuffers.  Or maybe we add the functionality for in-place serde in Borsh.

joncinque:
Tests added in db1572e

mvines (approved):

PR #2045, opened 2021-07-08
spl-token docs: Create NFT token with 0 decimal place

### LINKED ISSUE #2040: Example of creating non-fungible token seems wrong?
https://github.com/solana-labs/solana-program-library/blob/master/docs/src/token.md#example-create-a-non-fungible-token

The example gives the created token 9 decimal places
```
$ spl-token create-token --decimals 9
```
Then mint "1" token

Isn't that creates 1000000000 fungible tokens since you can transfer 0.000000001 to others?
And none of those tokens has any uniqueness since none of them has an unique id,

I suppose to create an actual non-fungible token you should create token with 0 decimal places:
```
$ spl-token create-token --decimals 0
```
then mint one and only token and disable minting.

HaoCherHong:
Close #2040


mvines:
Thanks!

PR #2046, opened 2021-07-08
Make twitter constants configurable for easier testing in name service js

ChewingGlass:
The diff here ended up uglier than I expected. All I've done is wrap twitter.ts in a class that takes the constants `TWITTER_ROOT_PARENT_REGISTRY_KEY` and `TWITTER_VERIFICATION_AUTHORITY` in its constructor, defaulting to the pre-configured ones.

This makes it easier to test and experiment with, as you can create different registries that act like the main twitter registry. 

This is part of another PR to update the name service to allow oauth claiming of twitter handles (i.e. you just click login with twitter instead of tweeting). 

I've done a major version bump, as this is going to change the API for any name service consumers.

### DIFF
--- a/name-service/js/package.json
+++ b/name-service/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@solana/spl-name-service",
-  "version": "0.1.2",
+  "version": "1.0.0",
   "description": "SPL Name Service JavaScript API",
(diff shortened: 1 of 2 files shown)


ChewingGlass:
name service here is out of date, making a PR upstream.

PR #2058, opened 2021-07-09
Eliminate doc warnings

brson:
This patch series allows `cargo doc` to run without emitting any warnings. It changes bare URLs to "autolinks", surrounded by angle brackets, and it changes some "account references" lists to put their bracketed requirements (sign/write) in code spans (as with other such lists in this repo) so they won't be interpreted as markdown links.


jstarry (approved):
looks great, thanks!

PR #2060, opened 2021-07-10
Governance: Move realm and goverend_account from config to direct field

SebastianBor:
#### Change Summary

This is a maintenance change to improve program quality without any feature changes.

While implementing `SetGovernanceConfig` instruction I realised it was too difficult. I had to add code and tests for attack vectors which shouldn't have even existed. The reason for that was `GovernanceConfig` struct which was being changed but at the same time had immutable data on it: `realm` and `governed_account`. This change moves these fields from `GovernanceConfig` to direct `Governance` account fields and hence protects their immutability without any extra code.

### DIFF
--- a/governance/program/src/error.rs
+++ b/governance/program/src/error.rs
@@ -276,10 +276,6 @@ pub enum GovernanceError {
     #[error("Governance PDA must sign")]
     GovernancePdaMustSign,
 
-    /// Invalid config realm for Governance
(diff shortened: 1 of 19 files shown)


joncinque (approved):
looks good, this does make a lot more sense!

SebastianBor:
> looks good, this does make a lot more sense!

thx

PR #2084, opened 2021-07-14
Breakup stake-pool update test transactions

jackcmay:
Stake-pool's `huge_test` includes the test `update` that attempts to process a very large transaction that fails the new compute budget tx-wide limit.  The test's transaction is not a practical use case and was done because it was easy to throw all the instructions into one transaction.  Break the test's transaction into multiple.


joncinque (approved):
Looks good, thanks Jack!

PR #2092, opened 2021-07-15
stake-pool: Add new fee fields for liquid staking and arbitrage prevention

joncinque:
In order to launch their stake pool, @lieuzhenghong and co. would like to be able to assess a fee for liquid deposits (for a fair price), and on withdrawals (to prevent arbitrage).

They will do the work to integrate the fees, but this at least adds the new fields to avoid backwards compatibility issues whenever their work is ready.

Let me know how it looks!

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -613,6 +613,9 @@ impl Processor {
         stake_pool.next_epoch_fee = None;
         stake_pool.preferred_deposit_validator_vote_address = None;
         stake_pool.preferred_withdraw_validator_vote_address = None;
+        stake_pool.deposit_fee = Fee::default();
+        stake_pool.withdrawal_fee = Fee::default();
+        stake_pool.next_withdrawal_fee = None;
 
         stake_pool
             .serialize(&mut *stake_pool_info.data.borrow_mut())
--- a/stake-pool/program/src/state.rs
+++ b/stake-pool/program/src/state.rs
@@ -8,6 +8,7 @@ use {
     solana_program::{
         account_info::AccountInfo,
         borsh::get_instance_packed_len,
+        clock::Epoch,
         msg,
         program_error::ProgramError,
         program_memory::sol_memcmp,
@@ -103,6 +104,15 @@ pub struct StakePool {
 
     /// Preferred withdraw validator vote account pubkey
     pub preferred_withdraw_validator_vote_address: Option<Pubkey>,
+
+    /// Fee assessed on deposits
+    pub deposit_fee: Fee,
+
+    /// Fee assessed on withdrawals
+    pub withdrawal_fee: Fee,
+
+    /// Future withdrawal fee, to be set at the given epoch
+    pub next_withdrawal_fee: Option<(Epoch, Fee)>,
 }
 impl StakePool {
     /// calculate the pool tokens that should be minted for a deposit of `stake_lamports`
--- a/stake-pool/program/tests/huge_pool.rs
+++ b/stake-pool/program/tests/huge_pool.rs
@@ -26,7 +26,7 @@ use {
         find_withdraw_authority_program_address, id,
         instruction::{self, PreferredValidatorType},
         stake_program,
-        state::{AccountType, StakePool, StakeStatus, ValidatorList, ValidatorStakeInfo},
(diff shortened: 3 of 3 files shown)


jon-chuang:
Hey @joncinque , thanks!

For increased safety, I would think there ought to be a maximum withdrawal fee, since if the fee is set at 100%, or anything significant, withdrawals are essentially unsafe, and we can't rely on users being attentive enough to notice when the fee massively spikes, potentially irreversibly.

Do you think a hard cap at 3% fee makes sense here? I think this is good enough to cap the maximum damage to a non-attentive staker.

On the other hand, I also think requiring withdrawal fee changes to take about 3 epochs (6-9 days) to activate would also give an extra buffer for people to become aware of potential bad-actor behaviour.

I'm quite uncertain whether you think a cap/delay hard-coded in the software, or a configurable cap/delay that is permanent upon initiation of the stakepool makes more sense?

joncinque:
> Do you think a hard cap at 2 or 3% fee makes sense here? I think this is good enough to cap the maximum damage to a non-attentive staker. On testnet, it seems the per-epoch rewards is about 1.xx%.

We shouldn't enforce a maximum fee at the protocol level.  There are use cases for a 100% withdrawal fee stake pool, just like there are usecases for a 100% commission validator.  You can fake a lockup in your pool with a 100% withdrawal fee, which could certainly have uses.  For example, I can create a pool with a 100% withdrawal fee, promising higher returns than normal, and set the future withdrawal fee to 0% in 50 epochs.

We could, however, enforce a gradual change in the withdrawal fee whenever it's increased.  For example, you can only increase it by maximum 2% every epoch.

jon-chuang:
I think I'll go with a maximum multiplier per epoch. This means, if it starts small, it'll stay small for a few epochs. But once the fee is big, it can increase quite quickly to 100%.

I'll go with the multiplier 2x.

So if it is 0.5%, it can increase to maximum 1% in the next epoch, 2% in the following epoch, 4%, 8% etc.

It kills two birds with one stone in also protecting inattentive stakers.

Let me know if this sounds good to you @joncinque

joncinque:
@jon-chuang yep that's a great idea.  Do these structure changes look good to you then? Let me know and I'll merge / release this

jon-chuang:
@joncinque, I made a modification so that the next_withdrawal_fee is Option<Fee> instead, since the maximum multiplier already serves the purpose of giving delegators time to catch on to fee increases.

Do you think that's ok or do we still need an Epoch to accompany it?

jon-chuang:
We'd also like to be able to add referral fees, which are a portion of the deposit/liquid staking fees which go to the referrer, similar to the referral fees offered by Serum for trades.

The invariant is `referral_fee <= deposit_fee`

joncinque:
@jon-chuang 

> I made a modification so that the next_withdrawal_fee is Option<Fee> instead, since the maximum multiplier already serves the purpose of giving delegators time to catch on to fee increases.

Yep, that makes sense.

> We'd also like to be able to add referral fees, which are a portion of the deposit/liquid staking fees which go to the referrer, similar to the referral fees offered by Serum for trades.  The invariant is referral_fee <= deposit_fee

Would it be `deposit_fee = referral_fee + manager_fee` or `total_fee = deposit_fee + referral_fee`.  Meaning, do you calculate them separately and add them up? Or do you take the referral fee as a proportion of the total deposit fee?  People always got confused about this.  It may be best to just express it as a `u8` percentage of the total fee to keep things as simple as possible

jon-chuang:
Ok, that's fair, we'll make it a u8 then.

jon-chuang:
Hey @joncinque for the `DepositSol` instruction, do we want a deposit authority?

For our project, we probably don't need it. Since deposits of SOL will not carry any slashing risk.

Still, it could be good to have the feature around. 
In that case I propose: the deposit authority for `DepositStake` doubles as that for `DepositSol`, and we add an additional `bool`, `require_deposit_sol_authority`.
This can be toggled via `solana-stake-pool require-deposit-sol-authority [true/false]`

What do you think?

joncinque:
@jon-chuang how about having a separate `liquid_deposit_authority` or `sol_deposit_authority` as an `Option<Pubkey>` ?

joncinque:
Closing since the work was done in #2107

PR #2093, opened 2021-07-15
chore: examples directory with mint transfer example

chaseeb:
created examples directory and example of how to mint and transfer spl tokens

### DIFF
--- a/token/js/examples/create_mint_and_transfer_tokens.js
+++ b/token/js/examples/create_mint_and_transfer_tokens.js
@@ -1,5 +1,5 @@
-import * as web3 from '@solana/web3.js';
-import * as splToken from '@solana/spl-token';
+var web3 = require('@solana/web3.js');
+var splToken = require('@solana/spl-token');
 
 (async () => {
   // Connect to cluster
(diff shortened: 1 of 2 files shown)


token/js/examples/create_mint_and_transfer_tokens.js
```diff
@@ -0,0 +1,77 @@
...
+  await connection.confirmTransaction(fromWalletAirdropSignature);
+
+  // Generate a new random public key and airdrop 1 SOL
+  var toWallet = web3.Keypair.generate();
+  var toWalletAirdropSignature = await connection.requestAirdrop(
+    toWallet.publicKey,
+    web3.LAMPORTS_PER_SOL,
+  );
+  //wait for airdrop confirmation
+  await connection.confirmTransaction(toWalletAirdropSignature);
```
jstarry, line 27: This isn't necessary because this example never uses the toWallet to pay for a transaction.

chaseeb:
Doesn’t that account need to be created and have sol in it in order to
receive a token?

jstarry:
> Doesn’t that account need to be created and have sol in it in order to receive a token?

Nope

chaseeb:
✅

PR #2094, opened 2021-07-15
Bump token-lending test caps

jackcmay:
The Solana compute cap is being applied transaction-wide instead of per instruction.  Bump the test's compute caps to compensate


jstarry:
lgtm as long as tests pass

PR #2101, opened 2021-07-17
Fix token/perf-monitor

fanatid:
`token/perf-monitor` is outdated now, I updated the test code to working state.


joncinque:
This was fixed up using program-test in https://github.com/solana-labs/solana-program-library/pull/2586, closing

fanatid:
Thanks, while getting instructions count with `solana-program-test` looks easier, `perf-monitor` was a good example of VM usage.

joncinque:
That's a good point. We should consider getting another example for people to use the VM directly :thinking:

PR #2102, opened 2021-07-18
Add token/rust: spl-token-api

fanatid:
Right now `spl-token` have `cli` and JS API. This proposal is about add a nice Rust API for tokens management which can be done through [solana_client::rpc_client::RpcClient](https://docs.rs/solana-client/1.7.6/solana_client/rpc_client/struct.RpcClient.html) and [solana_program_test::BanksClient](https://docs.rs/solana-program-test/1.7.6/solana_program_test/struct.BanksClient.html). This API can especially be useful for writing tests with `solana_program_test`.

This is only a small prototype, if a package will be approved I'll finish PR.


mvines:
Yes please, this would be great! Here's a direction I think would work nicely:
1. This becomes the `spl-token-client` crate (analogous to the `solana-client` crate)
2. Longer-term `spl-token-cli` gets ported to use `spl-token-client`

CriesofCarrots:
Hi @fanatid , I'd love to see this land. Are you available to help get it wrapped up?
I can also take it over, if you've moved on to other things.

fanatid:
@CriesofCarrots would happy to help with this PR :slightly_smiling_face:

token/rust/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+license = "Apache-2.0"
+name = "spl-token-client"
+repository = "https://github.com/solana-labs/solana-program-library"
+version = "0.0.1"
+
+[dependencies]
+async-trait = "0.1"
+solana-client = "1.7.11"
+solana-program-test = "1.7.11"
+solana-sdk = "1.7.11"
```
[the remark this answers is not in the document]
CriesofCarrots: We'll be updating to v1.9.1 soon, but if we get this in quickly, we don't have to deal with those api changes here.

token/rust/src/client.rs
```diff
@@ -0,0 +1,235 @@
...
+use solana_program_test::{tokio::sync::Mutex, BanksClient, ProgramTestContext};
+use solana_sdk::{
+    account::Account, hash::Hash, pubkey::Pubkey, signature::Signature, transaction::Transaction,
+};
+use std::{fmt, future::Future, pin::Pin, sync::Arc};
+
+type BoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
+
+/// Basic trait for sending transactions to validator.
+pub trait SendTransaction {
```
CriesofCarrots, line 12: Hmm, it sort of seems like almost everything in this module is generic enough to be upstreamed somewhere; maybe a new spl-client crate or even solana-client? I'm not sure what is the best fit yet.
Since I don't want to hold this up on that decision or a potential new solana-client release, how about we rename everything here to remove Token references, and then it will be easy to drop-in replace after I upstream?

token/rust/tests/program-test.rs
```diff
@@ -0,0 +1,267 @@
...
+use spl_token_client::{
+    client::{TokenBanksClient, TokenBanksClientProcessTransaction, TokenClient},
+    token::Token,
+};
+use std::sync::Arc;
+
+struct TestContext {
+    // pub ctx: Arc<Mutex<ProgramTestContext>>,
+    // pub client: Arc<dyn TokenClient<TokenBanksClientProcessTransaction>>,
+    // pub payer: Keypair,
```
CriesofCarrots, line 19: Remove?

token/rust/tests/program-test.rs
```diff
@@ -0,0 +1,267 @@
...
+            None,
+            decimals,
+        )
+        .await
+        .expect("failed to create mint");
+
+        Self {
+            // ctx,
+            // client,
+            // payer,
```
CriesofCarrots, line 62: Remove?

token/rust/src/token.rs
```diff
@@ -0,0 +1,260 @@
...
+    AccountNotFound,
+    #[error("invalid account owner")]
+    AccountInvalidOwner,
+    #[error("invalid account mint")]
+    AccountInvalidMint,
+}
+
+pub type TokenResult<T> = Result<T, TokenError>;
+
+pub struct Token<ST, TS> {
```
CriesofCarrots, line 32: Heh, `<ST, TS>` is a bit much. How about just <T, S>?
T: type / Transaction (sender)
S: Signer

token/rust/src/token.rs
```diff
@@ -0,0 +1,260 @@
...
+impl<ST, TS> Token<ST, TS>
+where
+    ST: SendTransaction,
+    TS: Signer,
+{
+    async fn process_ixs<T: Signers>(
+        client: &Arc<dyn TokenClient<ST>>,
+        payer: &TS,
+        instructions: &[Instruction],
+        signing_keypairs: &T,
```
CriesofCarrots, line 59: I think you can just roll `payer` into the `signing_keypairs` slice, can't you?

token/rust/src/token.rs
```diff
@@ -0,0 +1,260 @@
...
+                    decimals,
+                )?,
+            ],
+            &[mint_account],
+        )
+        .await?;
+
+        Ok(Self::new(client, mint_account.pubkey(), payer))
+    }
+
```
CriesofCarrots, line 128: Does it make sense to include a `payer: Option<S>` input to the methods below to allow overriding the Token default fee-payer?

token/rust/src/client.rs
```diff
@@ -0,0 +1,235 @@
...
+
+impl SendTransactionBanksClient for TokenBanksClientProcessTransaction {
+    fn send<'a>(
+        &self,
+        client: &'a mut BanksClient,
+        transaction: Transaction,
+    ) -> BoxFuture<'a, TokenClientResult<Self::Output>> {
+        Box::pin(async move {
+            client
+                .process_transaction(transaction)
```
CriesofCarrots, line 42: It's a pity that the BanksClient method(s) don't take &Transaction.
Can you pass by reference until here, though?

CriesofCarrots (commented):
Thanks for letting me pull you back in here :)
Looks solid. A few suggestions from me for starters.

token/rust/src/token.rs
fanatid: I added functions looking on the JS API and there is no optional payer. We can add function:
```rust
fn with_payer<S2: Signer>(&self, payer: S2) -> Token<T, S2> {
  Token { client: Arc::clone(&self.client), pubkey: self.pubkey, payer }
}
```
so it will be possible to do something like `token.with_payer(new_payer_here).create_associated_token_account(...)`.

token/rust/src/client.rs
fanatid: Pass by reference and create `Transaction` with clone for BanksClient method?

token/rust/src/token.rs
fanatid: I just changed `process_ixs` method, so no `client` and `payer` in arguments now.

token/rust/Cargo.toml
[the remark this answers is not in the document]
fanatid: Changed to `=1.8.1`.

fanatid:
Looks like something is wrong in `solana-program-test:1.8.1`.
I added `println!` locally and
1. request mint account info, supply is 1
2. mint amount 1 and client returned ok
3. request mint account info again, supply is 1

token/rust/src/client.rs
CriesofCarrots: Yes, that's what I was thinking. I don't feel strongly about it, though

CriesofCarrots:
> Looks like something is wrong in `solana-program-test:1.8.1`. I added `println!` locally and
> 
>     1. request mint account info, supply is 1
> 
>     2. mint amount 1 and client returned ok
> 
>     3. request mint account info again, supply is 1

Hm, I'll take a look tomorrow

fanatid:
Also, it's very strange but I receive error only with an empty `target` directory, if I run already compiled test binary everything is ok.

CriesofCarrots:
Hey, @fanatid , so the `set_authority` test has a race in it that is not obvious.
Since the set_authority test attempts the same mint operation [here](https://github.com/solana-labs/solana-program-library/pull/2102/files#diff-2e43f3ebfb4b69c4efa0ffb629fe776947954db1b2f769ac9e4d8d24d50df416R140) and [here](https://github.com/solana-labs/solana-program-library/pull/2102/files#diff-2e43f3ebfb4b69c4efa0ffb629fe776947954db1b2f769ac9e4d8d24d50df416R163), if those transactions get the same blockhash, they end up identical. It turns out that the BanksClient/Server doesn't return an error when asked to process the same transaction a 2nd time; it hands the duplicate transaction off to a Bank, but then finds the original successful status regardless of the fact that the duplicate was not committed.

The race is that if the Token client has happened to get a new blockhash for the second mint operation, the transactions are not identical, and all is well. If not (because BanksClient is still returning the original blockhash), the transactions are identical, and the test's `assert!(is_err)` will fail. ProgramTest seems to move somewhat slower on the CI machine or when running the tests in parallel, so this failure occurs more often.

Anyway, that's probably more detail than you wanted.
The short-term fix is to twiddle one of the mint operations so that they aren't the same. I'll make a suggest above.

This issue describes the more comprehensive fix to BanksServer that we are looking at: https://github.com/solana-labs/solana/issues/22029

token/rust/tests/program-test.rs
```diff
@@ -0,0 +1,267 @@
...
+        ..
+    } = TestContext::new().await;
+
+    let alice_vault = token
+        .create_associated_token_account(&alice.pubkey())
+        .await
+        .expect("failed to create associated token account");
+
+    token
+        .mint_to(&alice_vault, &mint_authority, 1)
```
CriesofCarrots, line 141:
```suggestion
        .mint_to(&alice_vault, &mint_authority, 2)
```

token/rust/src/client.rs
CriesofCarrots: Also, bump on this

token/rust/src/token.rs
CriesofCarrots: I don't think we need to be beholden to the JS api, do we?
But anyway `with_payer()` should work fine. Thanks!

fanatid:
Thank you for the explanation @CriesofCarrots! I thought about race conditions but did not understand why the test fails only with an empty `target` directory and is always ok with compiled binary.

token/rust/src/client.rs
fanatid: Not sure that I understand what should be renamed :disappointed:

CriesofCarrots: > Not sure that I understand what should be renamed

Everything in this client.rs that has the string `Token` in it, since none of it is specific to Token. I'll make a couple suggestions below as examples.
I'm not totally set on the prefix `Program`, so if you think of a better generic let me know.

token/rust/src/client.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+}
+
+//
+pub type TokenClientError = Box<dyn std::error::Error + Send + Sync>;
+pub type TokenClientResult<T> = Result<T, TokenClientError>;
+
+/// Token client interface.
+#[async_trait]
+pub trait TokenClient<ST>
```
CriesofCarrots, line 82:
```suggestion
/// Generic client interface for programs.
#[async_trait]
pub trait ProgramClient<ST>
```

token/rust/src/client.rs
```diff
@@ -0,0 +1,235 @@
...
+        client: &'a RpcClient,
+        transaction: Transaction,
+    ) -> BoxFuture<'a, TokenClientResult<Self::Output>> {
+        Box::pin(async move { client.send_transaction(&transaction).map_err(Into::into) })
+    }
+}
+
+//
+pub type TokenClientError = Box<dyn std::error::Error + Send + Sync>;
+pub type TokenClientResult<T> = Result<T, TokenClientError>;
```
CriesofCarrots, line 78:
```suggestion
pub type ProgramClientError = Box<dyn std::error::Error + Send + Sync>;
pub type ProgramClientResult<T> = Result<T, ProgramClientError>;
```

token/rust/src/client.rs
fanatid: Changed to reference.

CriesofCarrots (approved):
Awesome, looks great. Thank you for all the polish!

PR #2107, opened 2021-07-19
Stake pool/withdrawal fee

jon-chuang:
This PR introduces a withdrawal fee for `spl-stake-pool`. 

The withdrawal fee prevents nuisance delegators from performing validator arbitrage.

The following has been achieved:
1. Initialize stake pool with withdrawal fee
2. `SetWithdrawalFee` instruction. 
  a. Checks that fee increases do not exceed a factor of 1.5x. 
  b. If fee starts at 0, fee can only increase up to 0%->0.15% in the following epoch.
3. Tests for new functionality
4. cli functionality for initialize with optional withdrawal_fee (defaults to (0, 0) if not provided)
5. Add `set-withdrawal-fee` functionality to the cli

Sample output from cli:
```
spl-stake-pool create-pool --fee-denominator 1000 --fee-numerator 2 --withdrawal-fee-denominator 1000 --withdrawal-fee-numerator 3 --max-validators 1000`
Creating reserve stake 98pgpFtaxTRGTD5YaQPgdggMQ5g8ujFQF79bKser66fz
Creating mint JGhKEDuQS3tDNLKFTENnX4mwFTTYqQpgM8dX7xdRHWQ
Creating pool fee collection account DHVTmhCb1qGSwk2gLqJ1BMHJwKRFJ9Xiy9UJF18NT9HE
Signature: 3aZYistTpaX8dWfS4i4SM6oLVT24RSEXwjx5Y3rgKHC13CWarC1at8W9e7iFu84f8qvtumWbNRidRheiVpCd9XS9
Creating stake pool Vg1uDgymfa3tkDweYiiHzK5jnWZvprGEp5PQUgvjxek
Signature: 4hcDiXcS2jq3bayTKjAxPGgosWBJ9XABRuxSjdsUCxauSc1NsQtgUGZhcMMET4gJZGshkYZ9NU9BSaj8QC1a7VHr
```

```
spl-stake-pool set-withdrawal-fee 9WHSSh7MSNVDm7H1j51RvsYGRe4j4SSKLrdNTkMtuM1n 15 10000
Signature: 2Ly92Ai1mpd2sLLUefH55ddEr6eU6qwVbydNKtMvg6D92uhPYAfQkzD2XsT4qZGTZCwFBLftUd2W1wYJ87JY9S41
```

@billythedummy


stake-pool/program/src/processor.rs
```diff
@@ -594,6 +617,16 @@ impl Processor {
...
+        // Numerator should be smaller than or equal to denominator (fee <= 1)
+        if withdrawal_fee.numerator > withdrawal_fee.denominator {
+            return Err(StakePoolError::FeeTooHigh.into());
+        }
+        // To keep u64 arithmetic sane, we don't allow denominators (and consequently numerators)
+        // that are too large
+        if withdrawal_fee.denominator > 1_000_000 {
+            msg!("Withdrawal fee denominator too large");
+            return Err(StakePoolError::InvalidFeeDenominator.into());
+        }
```
joncinque, line 629: What's the reason for this?  If we want to set a max fee denominator of 1 million, we may as well change the whole thing to a `u32`, have it as a `per million` and not use the `Fee` type at all.  For consistency, I'd recommend sticking with the `Fee` type

stake-pool/program/src/processor.rs
```diff
@@ -1890,6 +1930,7 @@ impl Processor {
         let clock = &Clock::from_account_info(clock_info)?;
         let token_program_info = next_account_info(account_info_iter)?;
         let stake_program_info = next_account_info(account_info_iter)?;
+        let manager_fee_info = next_account_info(account_info_iter)?;
```
joncinque, line 1933: Don't forget to check that this is the right manager fee account! Otherwise, I can just put my own account in twice

stake-pool/program/src/processor.rs
```diff
@@ -2162,16 +2220,88 @@ impl Processor {
...
+            (
+                stake_pool.withdrawal_fee.numerator,
+                stake_pool.withdrawal_fee.denominator,
+            )
+        };
+
+        // This is always safe since numerator <= denominator <= 1_000_000
+        // and |MAX_WITHDRAWAL_FEE_INCREASE| <= 10
+        if old_num * fee.denominator * MAX_WITHDRAWAL_FEE_INCREASE.numerator
+            < fee.numerator * old_denom * MAX_WITHDRAWAL_FEE_INCREASE.denominator
```
joncinque, line 2279: Would you mind adding a comment for this math?  Even something simple like:
```
new_fee / old_fee <= MAX_WITHDRAWAL_FEE_INCREASE
```

stake-pool/program/src/processor.rs
```diff
@@ -2245,6 +2375,10 @@ impl Processor {
                 msg!("Instruction: SetStaker");
                 Self::process_set_staker(program_id, accounts)
             }
+            StakePoolInstruction::SetWithdrawalFee { fee } => {
+                msg!("Instruction: SetFee");
```
joncinque, line 2379:
```suggestion
                msg!("Instruction: SetWithdrawalFee");
```

stake-pool/program/src/state.rs
```diff
@@ -128,19 +137,28 @@ impl StakePool {
...
-        if self.fee.denominator == 0 || reward_lamports == 0 {
+        if reward_lamports == 0 {
             return Some(0);
         }
+        // If by chance the reserve account has lamports > 0 even though
+        // total_stake_lamports == 0, we credit it entirely to the manager,
+        // regardless of the fee.
+        if self.total_stake_lamports == 0 || self.pool_token_supply == 0 {
+            return Some(reward_lamports);
+        }
```
joncinque, line 158: Looks like this is from your other PR -- could you remove it?

stake-pool/program/src/state.rs
```diff
@@ -534,6 +552,20 @@ pub struct Fee {
     pub numerator: u64,
 }
 
+impl Fee {
+    /// Applies the Fee's rates to a given amount, `amt`
+    /// returning the amount to be subtracted from it as fees
+    /// (0 if denominator is 0 or amt is 0),
+    /// or None if overflow occurs
+    pub fn apply(&self, amt: u128) -> Option<u128> {
```
joncinque, line 560: I like this!

nit: this should use `u64` as the interface. since the fee is <= 1, and amounts are only ever u64, it will fit.  If not, it'll correctly blow up

stake-pool/program/src/state.rs
```diff
@@ -534,6 +552,20 @@ pub struct Fee {
...
+        if self.denominator == 0 {
+            return Some(0);
+        }
+        amt.checked_mul(self.numerator as u128)?
+            .checked_div(self.denominator as u128)
```
joncinque, line 565: Actually, this brings up a potential bug! If someone sets their fee quite low, there's a chance that someone can withdraw pico amounts to get this calculation to always be 0.  As such, this should return a minimum of 1 if numerator is > 0

stake-pool/program/tests/huge_pool.rs
```diff
@@ -26,13 +26,13 @@ use {
...
         stake_program,
-        state::{AccountType, StakePool, StakeStatus, ValidatorList, ValidatorStakeInfo},
+        state::{AccountType, Fee, StakePool, StakeStatus, ValidatorList, ValidatorStakeInfo},
         MAX_VALIDATORS_TO_UPDATE, MINIMUM_ACTIVE_STAKE,
     },
     spl_token::state::{Account as SplAccount, AccountState as SplAccountState, Mint},
 };
 
-const HUGE_POOL_SIZE: u32 = 4_000;
+const HUGE_POOL_SIZE: u32 = 3950;
```
joncinque, line 35: Ah man, what a shame, this makes me sad.  With no withdrawal fee, we can't get to 4k? That seems weird, unless deserialization of the new fields is that heavy.  Which tests fail at 4k?

stake-pool/program/tests/set_fee.rs
```diff
@@ -167,7 +167,7 @@ async fn fail_bad_fee() {
             let program_error = error::StakePoolError::FeeTooHigh as u32;
             assert_eq!(error_index, program_error);
         }
-        _ => panic!("Wrong error occurs while malicious try to set manager"),
+        _ => panic!("Wrong error occurs when setting fee too high"),
```
joncinque, line 170: Nice catch :smile:

stake-pool/program/tests/withdraw.rs
```diff
@@ -62,9 +62,9 @@ async fn setup() -> (
     .await
     .unwrap();
 
-    let tokens_to_burn = deposit_info.pool_tokens / 4;
+    let tokens_to_withdraw = deposit_info.pool_tokens / 4;
```
joncinque, line 65: much better, I never liked `tokens_to_burn` for some reason

joncinque (commented):
Lots of great stuff here!  Very excited to see that it's moving forward so quickly

stake-pool/program/src/processor.rs
jon-chuang: It was to simplify the arithmetic so as not to require checked muls. We need to multiply 3 numbers together to check the fee increase limit, and knowing the bounds means we can be safer in terms of error handling.

stake-pool/program/src/state.rs
jon-chuang: We were thinking that such errors were inconsequential as transaction fees would dominate at that point.

stake-pool/program/tests/huge_pool.rs
jon-chuang: The test which fails is upgrade for big_pools. Specifically, it is update pool values which fails. I tested that master fails with pool size 4100, so we didn't regress **that** much... 🙂

stake-pool/program/src/instruction.rs
```diff
@@ -942,6 +959,7 @@ pub fn withdraw(
         AccountMeta::new_readonly(sysvar::clock::id(), false),
         AccountMeta::new_readonly(*token_program_id, false),
         AccountMeta::new_readonly(stake_program::id(), false),
+        AccountMeta::new(*manager_fee_account, false),
```
jon-chuang, line 962: I guess I'll make it number 8 here as well

stake-pool/program/src/processor.rs
jon-chuang: hmm yeah, it also reminds me, we need to include the manager field in the withdraw command in the cli

stake-pool/program/tests/withdraw.rs
jon-chuang: credit to @billythedummy

stake-pool/program/src/processor.rs
```diff
@@ -2162,16 +2220,88 @@ impl Processor {
...
+        let mut stake_pool = try_from_slice_unchecked::<StakePool>(&stake_pool_info.data.borrow())?;
+        if !stake_pool.is_valid() {
+            return Err(StakePoolError::InvalidState.into());
+        }
+
+        stake_pool.check_manager(manager_info)?;
+
+        if stake_pool.last_update_epoch < clock.epoch {
+            return Err(StakePoolError::StakeListAndPoolOutOfDate.into());
+        }
```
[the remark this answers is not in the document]
jon-chuang: Actually, we kind of want this. Just like for `set_fee`, if the epoch has changed, we want the manager to only be able to `set_withdrawal_fee` for the following epoch, not the current one.

stake-pool/program/tests/set_fee.rs
jon-chuang: caught a few more :-)

jon-chuang:
Thanks for the quick review @joncinque , let me know if there are any other issues you spot. We'd be working on the liquid staking instruction next.

@billythedummy , would you be able to review as well when you can spare the time? Much appreciated.

stake-pool/program/src/processor.rs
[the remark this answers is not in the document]
billythedummy: right, agreed with @jon-chuang. I didn't realize before but without this check it's possible for a manager to set the withdrawal fee for the current epoch instead of the next one by calling `set_withdrawal_fee` before anyone has called `update_stake_pool_balance`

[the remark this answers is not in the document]
billythedummy: we're working on a new branch with the `Fee` struct reworked so that we don't need checked arithmetic operations. In brief here's the proposal:

- `numerator` and `denominator` are private `u64` fields
- following restrictions checked on `Fee` initialization: `denominator` >= 1, `numerator` <= `denominator`, `denominator` < some arbitrary precision such that multiplication and division will never overflow `u64`. We're thinking `1_000_000_000`
- impl `PartialOrd` and overload `std::ops::Mul`

so, basically changing `Fee` into a constrained fraction

stake-pool/program/src/processor.rs
billythedummy: we're working on a new branch with the `Fee` struct reworked so that we don't need checked arithmetic operations. In brief here's the proposal:

- `numerator` and `denominator` are private `u64` fields
- following restrictions checked on `Fee` initialization: `denominator` >= 1, `numerator` <= `denominator`, `denominator` < some arbitrary precision such that multiplication and division will never overflow `u64`. We're thinking `1_000_000_000`
- impl `PartialOrd` and overload `std::ops::Mul`

so, basically changing `Fee` into a constrained fraction

jon-chuang: following some discussion with @billythedummy the approach pursued was a bit over the top. 

Decided to just go with u32 fees as I don't think anyone will ever need more than 9 dps of precision.

stake-pool/program/tests/huge_pool.rs
jon-chuang: hehe clever me noticed that `apply_fee` wasn't inlined, we're back at `4_000` now  :muscle:

stake-pool/program/src/processor.rs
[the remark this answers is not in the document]
joncinque: Ah right, sounds good then!

stake-pool/program/src/state.rs
joncinque: Makes sense, all good then

stake-pool/program/src/processor.rs
```diff
@@ -2166,16 +2221,91 @@ impl Processor {
...
+            )
+        } else {
+            (
+                stake_pool.withdrawal_fee.numerator,
+                stake_pool.withdrawal_fee.denominator,
+            )
+        };
+
+        // Check that new_fee / old_fee <= MAX_WITHDRAWAL_FEE_INCREASE
+        // No checks are required as all the variables are u32
```
joncinque, line 2275: This comment is out of date.  Should we say that we allow this to blow up if a ridiculous fee is put in?

stake-pool/program/src/state.rs
```diff
@@ -103,6 +103,15 @@ pub struct StakePool {
 
     /// Preferred withdraw validator vote account pubkey
     pub preferred_withdraw_validator_vote_address: Option<Pubkey>,
+
+    /// Fee assessed on deposits
+    pub deposit_fee: Fee,
```
joncinque, line 108: Could you add the `u8` deposit referral percentage while we're at it?

stake-pool/program/tests/huge_pool.rs
```diff
@@ -26,13 +26,13 @@ use {
...
         stake_program,
-        state::{AccountType, StakePool, StakeStatus, ValidatorList, ValidatorStakeInfo},
+        state::{AccountType, Fee, StakePool, StakeStatus, ValidatorList, ValidatorStakeInfo},
         MAX_VALIDATORS_TO_UPDATE, MINIMUM_ACTIVE_STAKE,
     },
     spl_token::state::{Account as SplAccount, AccountState as SplAccountState, Mint},
 };
 
-const HUGE_POOL_SIZE: u32 = 4_000;
+const HUGE_POOL_SIZE: u32 = 4000;
```
joncinque, line 35: nit for consistency
```suggestion
const HUGE_POOL_SIZE: u32 = 4_000;
```

stake-pool/program/tests/set_fee.rs
```diff
@@ -216,6 +216,6 @@ async fn fail_not_updated() {
             let program_error = error::StakePoolError::StakeListAndPoolOutOfDate as u32;
             assert_eq!(error_index, program_error);
         }
-        _ => panic!("Wrong error occurs while malicious try to set manager"),
+        _ => panic!("Wrong error occurs when stake pool out of date"),
```
joncinque, line 219: double oops, nice catch!

stake-pool/program/tests/set_withdrawal_fee.rs
```diff
@@ -0,0 +1,407 @@
...
+        .err()
+        .unwrap()
+        .unwrap();
+
+    match error {
+        TransactionError::InstructionError(_, InstructionError::Custom(error_index)) => {
+            let program_error = error::StakePoolError::WrongManager as u32;
+            assert_eq!(error_index, program_error);
+        }
+        _ => panic!("Wrong error occurs while malicious try to set manager"),
```
joncinque, line 210:
```suggestion
        _ => panic!("Wrong error occurs while signing with the wrong manager"),
```

joncinque (commented):
Looking really close! Just some minor things, then this should be good to go

joncinque (approved):
Looks good to me! Let me know when it's ready from your side and we'll integrate

jon-chuang:
Yep LGTM too after final once-over @joncinque . I'll let you to the honours.

PR #2110, opened 2021-07-20
Chore: bump js version

bartosz-lipinski:
Increment version for:
* @solana/spl-name-service
* @solana/spl-token


joncinque:
Seems fine, but what's the rationale for downgrading token to v1 package-lock?

PR #2114, opened 2021-07-20
rework fees to be u32

jon-chuang:
The restriction makes arithmetic easier to handle

Fees don't really need more than 6, let alone 9 digits of precision

Discussion: https://github.com/solana-labs/solana-program-library/pull/2107#discussion_r672628739

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -1813,8 +1813,8 @@ fn main() {
     let _ = match matches.subcommand() {
         ("create-pool", Some(arg_matches)) => {
             let deposit_authority = pubkey_of(arg_matches, "deposit_authority");
-            let numerator = value_t_or_exit!(arg_matches, "fee_numerator", u64);
-            let denominator = value_t_or_exit!(arg_matches, "fee_denominator", u64);
+            let numerator = value_t_or_exit!(arg_matches, "fee_numerator", u32);
+            let denominator = value_t_or_exit!(arg_matches, "fee_denominator", u32);
             let max_validators = value_t_or_exit!(arg_matches, "max_validators", u32);
             let pool_keypair = keypair_of(arg_matches, "pool_keypair");
             let mint_keypair = keypair_of(arg_matches, "mint_keypair");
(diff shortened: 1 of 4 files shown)


joncinque:
I had originally meant that you can express the fee as one `u32`, and that number would represent `parts per million`, so that you would calculate as `fee / 1_000_000`.  Do you prefer to go that route?

Note that this change here would break the existing foundation stake pool on testnet unless we add an upgrade instruction to take an old stake pool, deserialize it, and write it back as a new one.  If you can add something to do that, then we can release this version of the program, do the upgrade, then get rid of the instruction.

jon-chuang:
Hmm @joncinque , wouldn't this upgrade instruction also need to be made thanks to the new struct field changes with deposit withdrawal and referral fees etc?

joncinque:
@jon-chuang Potentially, I think those will be fine since they'll just default to 0, which should be valid, correct?

jon-chuang:
Hmm, I'm not quite sure what you mean by default to zero. 

A new stake pool program will attempt to deserialize the struct, and if there are struct changes, it is going to break, period.

joncinque:
@jon-chuang in most cases, we'll actually be OK. borsh packs things in as much as possible, so we usually have an overallocated buffer.  adding a few new fields at the end is ok in that case.  for example, `Option<Pubkey>` is just encoded as `[0]` for `None`, and `[1, byte 1, byte 2 .... byte 32]` for `Some`. if we have that option set to None, we have an extra 32 bytes unused that the new fields can spill into.  it's not great, but since the foundation stake pool will not set preferred deposit / withdraw validators, it has a lot of extra bytes to work with, so new fields will actually be ok.  Changing how the fee looks will absolutely break it though

jon-chuang:
Hmm. Well tbh, let's just use u64 fields then, I'll just use checked u128 muls. Worst thing that can happen, a user needs to set fee to something lower precision, if they somehow manage to bust through 64 bits of precision on each fee...

joncinque:
That would be my choice.  I know the two `u64`s are kind of silly, but it's a bit of a hassle to update, unless you want to do the work, of course :wink:

PR #2124, opened 2021-07-21
Governance: Use separate thresholds to create proposals for council and community tokens

SebastianBor:
#### Summary 

- This change introduces `min_council_tokens_to_create_proposal` and `min_community_tokens_to_create_proposal` instead of a single `min_tokens_to_create_proposal` threshold to make it possible to use different thresholds for council and community tokens. A single threshold `min_tokens_to_create_proposal` is not sufficient for scenarios when the `Council` and `Community` mints supplies are significantly different (the most common case) and have different decimal places.

- This change also makes it possible for community token holders to create proposals voted on by the council and council token holders can now create proposals voted on by the community. 
  This restriction existed in previous versions but it was a side effect of the implementation and not by design.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3730,7 +3730,7 @@ dependencies = [
 
 [[package]]
 name = "spl-governance"
-version = "1.0.3"
+version = "1.0.4"
 dependencies = [
  "arrayref",
  "assert_matches",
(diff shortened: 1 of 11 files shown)


joncinque (approved):
Sorry for the slowness on this, but it looks great! The test cases covered all the best conditions I could think of

SebastianBor:
> Sorry for the slowness on this, but it looks great! The test cases covered all the best conditions I could think of

thank you

PR #2125, opened 2021-07-21
stake-pool: Update docs with new instructions and output

joncinque:
The stake pool docs are pretty out of date, and caused some confusion to users, so let's clarify them!

* add set-preferred-validator instruction
* update all output to how it looks
* remove all mentions of the word "derivative"
* explain practically how to add a validator to the stake pool
* explain why funds are safe in the system

@c3pko could you take a look at this? All comments are very welcome

### DIFF
--- a/docs/src/stake-pool.md
+++ b/docs/src/stake-pool.md
@@ -43,7 +43,7 @@ available at:
 - https://docs.solana.com/cli/manage-stake-accounts
 
 To reach a wider audience of users, stake pool managers are encouraged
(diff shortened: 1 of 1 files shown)


c3pko (commented):
this looks good for the most part! quick question--in (new) lines 692-700 you describe a user withdrawing their pool tokens. Is that right? Does one withdraw their pool tokens or their staked SOL?

joncinque:
That's a much better way of putting it -- yes, they withdraw the staked SOL in exchange for pool tokens.  thanks for the feedback!

PR #2129, opened 2021-07-22
Governance: Capture realm authority and proposal mint supply

SebastianBor:
#### Summary

The changes in this PR capture 1) realm authority in `Realm` account and 2) governing token mint supply in `Proposal` account but they are not used yet. The captured fields will be used in subsequent PRs and this PR is created only as a spearhead development to avoid breaking changes. 

- `Realm authority` - Realm authority will be used to control permissions to create `Governances` within realms. Since a realm is de facto a DAO its structure  should be controlled through the authority. After a realm is created the authority can be transferred to `Realm Governance` and hence any changes to the realm will be subject to votes on proposals.
  
- `governing token mint vote supply`  - ~~The supply will be used as the reference total supply to calculate voting results. It's captured on the proposal when it enters voting state to prevent moving goal posts when the proposal is being voted on and~~ . It's captured at the time a vote is decided  to show correct vote results for historical proposals in cases where the mint supply changes after vote completes.

### DIFF
--- a/governance/program/src/state/proposal.rs
+++ b/governance/program/src/state/proposal.rs
@@ -411,21 +411,6 @@ pub fn get_proposal_data_for_governance_and_governing_mint(
     Ok(proposal_data)
 }
 
-/// Deserializes Proposal and validates it belongs to the given Governing Mint
-pub fn get_proposal_data_for_governing_mint(
-    program_id: &Pubkey,
-    proposal_info: &AccountInfo,
-    governing_token_mint: &Pubkey,
-) -> Result<Proposal, ProgramError> {
-    let proposal_data = get_proposal_data(program_id, proposal_info)?;
-
-    if proposal_data.governing_token_mint != *governing_token_mint {
-        return Err(GovernanceError::InvalidGoverningMintForProposal.into());
-    }
-
-    Ok(proposal_data)
-}
-
 /// Deserializes Proposal and validates it belongs to the given Governance
 pub fn get_proposal_data_for_governance(
     program_id: &Pubkey,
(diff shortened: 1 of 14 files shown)


governance/program/src/state/proposal.rs
```diff
@@ -88,6 +88,12 @@ pub struct Proposal {
     /// Note: This field is not used in the current version
     pub execution_flags: InstructionExecutionFlags,
 
+    /// The supply of the Governing Token mint at the time Proposal entered voting stage
+    /// It's used as the reference total supply to calculate voting results
+    /// It's captured on the Proposal to prevent moving goal posts when the Proposal is being voted on
+    /// and to show correct vote results for historical proposals once the mint supply changes
+    pub governing_token_mint_supply: Option<u64>,
```
joncinque, line 95: Will you also need to capture each token owner's amount as well at the start of voting?  Or is that meant for a phase 2 item?  Otherwise, it seems like this makes it easier to fudge the vote, for example, if you mint a bunch of tokens after this is recorded.

joncinque (dismissed):
Looks great!  Just one question that shouldn't stop things

governance/program/src/state/proposal.rs
SebastianBor: Thank you, you are completely right. If we could get the snapshot of the token owner's accounts we could also get the snapshot of the mint at the same time and this is what `voting_at_slot` is for (once we can use snapshots).
I reverted the changes and renamed the field to `governing_token_mint_vote_supply` to capture the mint supply at the time vote was decided. It's used to show correct vote results for decided votes. The common use case when it was causing issues was for votes on proposals to mint governing tokens. After executing it the calculations on the UI were instantaneously wrong.

PR #2131, opened 2021-07-23
Bump spl-token-cli

CriesofCarrots:


CriesofCarrots:
Release of this version is blocked on v1.7.7 solana crates, with spl-token v3.2:
```
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for `spl-token`.
      ... required by package `solana-account-decoder v1.7.4`
      ... which is depended on by `spl-token-cli v2.0.14 (/solana-program-library/token/cli)`
  versions that meet the requirements `=3.1.1` are: 3.1.1

  all possible versions conflict with previously selected packages.

    previously selected package `spl-token v3.2.0`
      ... which is depended on by `spl-token-cli v2.0.14 (/solana-program-library/token/cli)`
      ```

PR #2137, opened 2021-07-23
Governance: SetRealmAuthority instruction

SebastianBor:
#### Summary 

- `SetRealmAuthority` - instruction allows to set new realm authority or set it to None. The instruction must be signed by the current authority and once set to None can't be changed any longer.
 Note: The authority is not used in this version yet. 

- `test_sign_off_proposal_with_signatory_must_sign_error` - added missing test checking for signatory signature


governance/program/src/processor/process_set_realm_authority.rs
```diff
@@ -0,0 +1,35 @@
...
+    pubkey::Pubkey,
+};
+
+use crate::{error::GovernanceError, state::realm::get_realm_data_for_authority};
+
+/// Processes SetRealmAuthority instruction
+pub fn process_set_realm_authority(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    new_realm_authority: &Option<Pubkey>,
```
joncinque, line 16: nit: you can pass this by value `Option<Pubkey>`

joncinque (approved):
Looks good! Just the one nit

governance/program/src/processor/process_set_realm_authority.rs
SebastianBor: done, thx

PR #2140, opened 2021-07-24
Stake pool/liquid deposits

jon-chuang:
@billythedummy

This PR adds liquid deposits as well as deposit fees.

The PR builds directly on top of https://github.com/solana-labs/solana-program-library/pull/2107

With these changes, we believe stake pools will be much safer and easier to use.


jon-chuang:
restart PR to be on trunk

PR #2141, opened 2021-07-26
Stake pool/liquid deposits (rebased on new master)

jon-chuang:
@billythedummy

This PR adds liquid deposits as well as deposit fees.

The PR builds on the merged PR https://github.com/solana-labs/solana-program-library/pull/2107

With these changes, we believe stake pools will be much safer and easier to use.


stake-pool/program/tests/deposit_sol.rs
```diff
@@ -0,0 +1,467 @@
...
+                error_index,
+                error::StakePoolError::InvalidSolDepositAuthority as u32
+            );
+        }
+        _ => panic!("Wrong error occurs while try to make a deposit with wrong stake program ID"),
+    }
+}
+
+#[tokio::test]
+async fn success_with_referral_fee() {
```
joncinque, line 378: Great tests!

stake-pool/cli/src/main.rs
```diff
@@ -311,8 +317,15 @@ fn command_create_pool(
...
-    initialize_transaction.sign(&initialize_signers, recent_blockhash);
+    if let Some(stake_deposit_authority) = stake_deposit_authority {
+        let mut initialize_signers = initialize_signers.clone();
+        initialize_signers.push(&stake_deposit_authority);
+        unique_signers!(initialize_signers);
+        initialize_transaction.sign(&initialize_signers, recent_blockhash);
+    } else {
+        unique_signers!(initialize_signers);
+        initialize_transaction.sign(&initialize_signers, recent_blockhash);
+    }
```
joncinque, line 328: Oops, thanks for fixing this!

stake-pool/cli/src/main.rs
```diff
@@ -347,6 +360,15 @@ fn command_vsa_create(
...
 
+fn command_show(config: &Config, stake_pool_address: &Pubkey) -> CommandResult {
+    let stake_pool = get_stake_pool(&config.rpc_client, stake_pool_address)?;
+    println!(
+        "Stake pool at address {}:\n{:#?}",
+        stake_pool_address, stake_pool
+    );
+    Ok(())
+}
+
```
joncinque, line 371: Can we remove this and add any additional required info to `list`?

stake-pool/cli/src/main.rs
```diff
@@ -567,11 +589,12 @@ fn add_associated_token_account(
     account
 }
 
-fn command_deposit(
+fn command_deposit_stake(
     config: &Config,
     stake_pool_address: &Pubkey,
     stake: &Pubkey,
     token_receiver: &Option<Pubkey>,
+    referrer: &Option<Pubkey>,
```
joncinque, line 597: nit: just to disambiguate, can we call this `referrer_token_account`?  Just this interface makes it seem like it will derive an ATA from this pubkey

stake-pool/cli/src/main.rs
```diff
@@ -660,8 +688,125 @@ fn command_deposit(
...
+    from: &Option<Keypair>,
+    token_receiver: &Option<Pubkey>,
+    referrer: &Option<Pubkey>,
+    amount: f64,
+) -> CommandResult {
+    if !config.no_update {
+        command_update(config, stake_pool_address, false, false)?;
+    }
+
+    let amount = spl_token::ui_amount_to_amount(amount, 9);
```
joncinque, line 726: Since you're using sol here, best to use `native_token::sol_to_lamports`

stake-pool/cli/src/main.rs
```diff
@@ -660,8 +688,125 @@ fn command_deposit(
...
+    let from_pubkey = from.as_ref().map_or_else(
+        || config.fee_payer.try_pubkey().unwrap(),
+        |keypair| keypair.try_pubkey().unwrap(),
+    );
+    let from_balance = config.rpc_client.get_balance(&from_pubkey)?;
+    if from_balance < amount {
+        return Err(format!(
+            "Not enough SOL to deposit into pool: {}.\nMaximum deposit amount is {} SOL.",
+            spl_token::amount_to_ui_amount(amount, 9),
+            spl_token::amount_to_ui_amount(from_balance, 9)
```
joncinque, line 738: Here you can use `Sol(lamports)` to get the pretty sign, and not have to write `SOL` in the message

stake-pool/cli/src/main.rs
```diff
@@ -660,8 +688,125 @@ fn command_deposit(
...
+    send_transaction(config, transaction)?;
+    Ok(())
+}
+
+fn command_deposit_sol(
+    config: &Config,
+    stake_pool_address: &Pubkey,
+    from: &Option<Keypair>,
+    token_receiver: &Option<Pubkey>,
+    referrer: &Option<Pubkey>,
```
joncinque, line 719: nit: can we call this `referrer_token_account`?

stake-pool/cli/src/main.rs
```diff
@@ -660,8 +688,125 @@ fn command_deposit(
...
+    let referrer = referrer.unwrap_or(stake_pool.manager_fee_account);
+
+    let pool_withdraw_authority =
+        find_withdraw_authority_program_address(&spl_stake_pool::id(), stake_pool_address).0;
+
+    let mut deposit_instructions = if let Some(sol_deposit_authority) =
+        config.sol_depositor.as_ref()
+    {
+        let expected_sol_deposit_authority = stake_pool.sol_deposit_authority.ok_or_else(|| {
+            "SOL deposit authority specified in config but stake Pool has none".to_string()
```
joncinque, line 771:
```suggestion
            "SOL deposit authority specified in arguments but stake pool has none".to_string()
```

stake-pool/cli/src/main.rs
```diff
@@ -696,7 +841,7 @@ fn command_list(config: &Config, stake_pool_address: &Pubkey) -> CommandResult {
         println!("Validator List: {}", stake_pool.validator_list);
         println!("Manager: {}", stake_pool.manager);
         println!("Staker: {}", stake_pool.staker);
-        println!("Depositor: {}", stake_pool.deposit_authority);
+        println!("Depositor: {}", stake_pool.stake_deposit_authority);
```
joncinque, line 844: Can we print the sol deposit authority if it exists? And also the deposit, withdrawal, and referral fees?

stake-pool/cli/src/main.rs
```diff
@@ -660,8 +688,125 @@ fn command_deposit(
...
+            "Not enough SOL to deposit into pool: {}.\nMaximum deposit amount is {} SOL.",
+            spl_token::amount_to_ui_amount(amount, 9),
+            spl_token::amount_to_ui_amount(from_balance, 9)
+        )
+        .into());
+    }
+
+    let stake_pool = get_stake_pool(&config.rpc_client, stake_pool_address)?;
+
+    let mut instructions: Vec<Instruction> = vec![];
```
joncinque, line 745: For added safety, let's craft this transaction as:

* create ephemeral keypair
* transfer `amount` from `from` to the ephemeral keypair
* pass the ephemeral keypair to `sol_deposit` and sign with it

this way, people can feel better and know that the maximum amount at risk is however much is transferred to the ephemeral keypair, and *not* everything in their account

stake-pool/cli/src/main.rs
```diff
@@ -1581,8 +1826,8 @@ fn main() {
                 .required(true)
             )
         )
-        .subcommand(SubCommand::with_name("deposit")
-            .about("Add stake account to the stake pool")
+        .subcommand(SubCommand::with_name("deposit-stake")
+            .about("Deposit fully-activated stake account into the stake pool in exchange for pool tokens")
```
joncinque, line 1830: This is a nice clarification. We've had people confused about using the term `fully-activated` in the past, so just a slight simplification:
```suggestion
            .about("Deposit active stake account into the stake pool in exchange for pool tokens")
```

stake-pool/cli/src/main.rs
```diff
@@ -1611,6 +1856,61 @@ fn main() {
                           Defaults to the token-owner's associated pool token account. \
                           Creates the account if it does not exist."),
             )
+            .arg(
+                Arg::with_name("referrer")
+                    .validator(is_pubkey)
+                    .value_name("ADDRESS")
+                    .takes_value(true)
+                    .help("Account to receive the referral fees for deposits. \
+                          Defaults to the pool manager."),
```
joncinque, line 1865:
```suggestion
                          Defaults to the token receiver."),
```

stake-pool/cli/src/main.rs
```diff
@@ -1611,6 +1856,61 @@ fn main() {
                           Defaults to the token-owner's associated pool token account. \
                           Creates the account if it does not exist."),
             )
+            .arg(
+                Arg::with_name("referrer")
+                    .validator(is_pubkey)
+                    .value_name("ADDRESS")
+                    .takes_value(true)
+                    .help("Account to receive the referral fees for deposits. \
```
joncinque, line 1864:
```suggestion
                    .help("Pool token account to receive the referral fees for deposits. \
```

stake-pool/cli/src/main.rs
```diff
@@ -1611,6 +1856,61 @@ fn main() {
...
+                    .takes_value(true)
+                    .help("Amount in SOL to deposit into the stake pool reserve account."),
+            )
+            .arg(
+                Arg::with_name("from")
+                    .long("from")
+                    .validator(is_keypair)
+                    .value_name("KEYPAIR")
+                    .takes_value(true)
+                    .help("Account to deposit SOL from. \
```
joncinque, line 1892: nit: we use this in other parts of the CLI, like `create-stake-account`
```suggestion
                    .help("Source account of funds. \
```

stake-pool/cli/src/main.rs
```diff
@@ -1611,6 +1856,61 @@ fn main() {
...
+                    .index(2)
+                    .validator(is_amount)
+                    .value_name("AMOUNT")
+                    .takes_value(true)
+                    .help("Amount in SOL to deposit into the stake pool reserve account."),
+            )
+            .arg(
+                Arg::with_name("from")
+                    .long("from")
+                    .validator(is_keypair)
```
joncinque, line 1889: If people want to use this with ledger, we'll need to support `is_keypair_or_ask_keyword` -- what do you think?

stake-pool/cli/src/main.rs
```diff
@@ -1611,6 +1856,61 @@ fn main() {
...
+                          Creates the account if it does not exist."),
+            )
+            .arg(
+                Arg::with_name("referrer")
+                    .long("referrer")
+                    .validator(is_pubkey)
+                    .value_name("REFERRER_TOKEN_ADDRESS")
+                    .takes_value(true)
+                    .help("Account to receive the referral fees for deposits. \
+                          Defaults to the pool manager."),
```
joncinque, line 1912: Just thinking, if someone is using this from the CLI, how about they get the referral instead?
```suggestion
                    .help("Pool token account to receive the referral fees for deposits. \
                          Defaults to the token receiver."),
```

stake-pool/cli/src/main.rs
```diff
@@ -1648,8 +1948,8 @@ fn main() {
                     .help("Do not automatically merge transient stakes. Useful if the stake pool is in an expected state, but the balances still need to be updated."),
             )
         )
-        .subcommand(SubCommand::with_name("withdraw")
-            .about("Withdraw amount from the stake pool")
+        .subcommand(SubCommand::with_name("withdraw-stake")
+            .about("Withdraw amount from the stake pool into designated stake account")
```
joncinque, line 1952: The destination stake account isn't required, so this addition could confuse people

stake-pool/cli/src/main.rs
```diff
@@ -660,8 +688,125 @@ fn command_deposit(
...
+    // Create token account if not specified
+    let token_receiver = token_receiver.unwrap_or(add_associated_token_account(
+        config,
+        &stake_pool.pool_mint,
+        &config.token_owner.pubkey(),
+        &mut instructions,
+        &mut total_rent_free_balances,
+    ));
+
+    let referrer = referrer.unwrap_or(stake_pool.manager_fee_account);
```
joncinque, line 762: Kindness to users?
```suggestion
    let referrer = referrer.unwrap_or(token_receiver);
```

stake-pool/cli/src/main.rs
```diff
@@ -621,36 +644,41 @@ fn command_deposit(
         &mut total_rent_free_balances,
     ));
 
+    let referrer = referrer.unwrap_or(stake_pool.manager_fee_account);
```
joncinque, line 647: We :heart: CLIs?
```suggestion
    let referrer = referrer.unwrap_or(token_receiver);
```

stake-pool/cli/src/main.rs
```diff
@@ -1432,6 +1665,18 @@ fn main() {
...
+            .arg(
+                Arg::with_name("pool")
+                    .index(1)
+                    .validator(is_pubkey)
+                    .value_name("POOL_ADDRESS")
+                    .takes_value(true)
+                    .required(true)
+                    .help("Stake pool address"),
+            )
+        )
```
joncinque, line 1679: Let's remove this too

stake-pool/cli/src/main.rs
```diff
@@ -1758,8 +2058,36 @@ fn main() {
...
+                        let default = if x == String::from("none") {
+                            Ok(())
+                        } else {
+                            Err(String::from("Not a pubkey or 'none'"))
+                        };
+                        is_pubkey(x).or(default)
+                    })
+                    .value_name("ADDRESS_OR_NONE")
+                    .takes_value(true)
+                    .help("'none', or a public key for the new stake pool sol deposit authority."),
```
joncinque, line 2086: How about following the convention from the preferred validator CLI?  There, you specify `--unset` if you want to unset or `--vote-account <PUBKEY_HERE>` to specify the preferred validator vote account.  I'm not sure which is better, but we should at least be consistent

stake-pool/cli/src/main.rs
```diff
@@ -1769,9 +2097,17 @@ fn main() {
                     .required(true)
                     .help("Stake pool address."),
             )
+            .arg(Arg::with_name("fee_type")
+                .index(2)
+                .value_name("FEE_TYPE")
+                .possible_values(&["pool","deposit", "withdrawal"]) // PreferredValidatorType enum
```
joncinque, line 2103: There's three different names for the original fee at this point: management, pool, and epoch.  `pool` is probably too vague, so I'm happy with either `management` or `epoch`, with a slight preference toward `epoch`.  Either way, let's keep it consistent

stake-pool/cli/src/main.rs
```diff
@@ -1884,18 +2222,21 @@ fn main() {
 
     let _ = match matches.subcommand() {
         ("create-pool", Some(arg_matches)) => {
-            let deposit_authority = pubkey_of(arg_matches, "deposit_authority");
+            let stake_deposit_authority = keypair_of(arg_matches, "stake_deposit_authority");
             let numerator = value_t_or_exit!(arg_matches, "fee_numerator", u64);
             let denominator = value_t_or_exit!(arg_matches, "fee_denominator", u64);
             let w_numerator = value_t!(arg_matches, "withdrawal_fee_numerator", u64);
             let w_denominator = value_t!(arg_matches, "withdrawal_fee_denominator", u64);
+            let d_numerator = value_t!(arg_matches, "withdrawal_fee_numerator", u64);
```
joncinque, line 2230:
```suggestion
            let d_numerator = value_t!(arg_matches, "deposit_fee_numerator", u64);
```

stake-pool/cli/src/main.rs
```diff
@@ -1884,18 +2222,21 @@ fn main() {
...
+            let d_denominator = value_t!(arg_matches, "withdrawal_fee_denominator", u64);
```
joncinque, line 2231:
```suggestion
            let d_denominator = value_t!(arg_matches, "deposit_fee_denominator", u64);
```

stake-pool/cli/src/main.rs
```diff
@@ -1904,12 +2245,21 @@ fn main() {
...
                 max_validators,
                 pool_keypair,
                 mint_keypair,
                 reserve_keypair,
             )
         }
+        ("show", Some(arg_matches)) => {
+            let stake_pool_address = pubkey_of(arg_matches, "pool").unwrap();
+            command_show(&config, &stake_pool_address)
+        }
```
joncinque, line 2262: Let's remove this too

joncinque (commented):
Really great work here! It was a lot to add all of these, so I really appreciate the time that you spent on this.  A lot of the comments are pretty minor, but the biggest one will be the separation of sol deposit fees and stake deposit fees.

stake-pool/program/src/processor.rs
```diff
@@ -813,20 +830,13 @@ impl Processor {
         // Check amount of lamports
         let stake_lamports = **stake_account_info.lamports.borrow();
         let minimum_lamport_amount = minimum_stake_lamports(&meta);
-        if stake_lamports != minimum_lamport_amount {
+        if stake_lamports != minimum_lamport_amount
+            || stake.delegation.stake != MINIMUM_ACTIVE_STAKE
+        {
             msg!(
-                "Error: attempting to add stake with {} lamports, must have {} lamports",
+                "Error: attempting to add (stake: {}, delegation: {}), below minimum",
```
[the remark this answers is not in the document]
jon-chuang: unfortunately, message can only take up to 4 arguments.

The reason for combining the error is that I encounter a stackoverflow for the instruction unless combining this particular if/else.

When trying to format the minimum into a single tuple, I also encountered the stackoverflow error.

I'll try again with a different build config.

stake-pool/program/src/processor.rs
```diff
@@ -1762,6 +1774,14 @@ impl Processor {
             return Err(ProgramError::IncorrectProgramId);
         }
 
+        if stake_pool.manager_fee_account != *manager_fee_info.key {
+            return Err(StakePoolError::InvalidFeeAccount.into());
+        }
+
+        if stake_pool.token_program_id != *referrer_fee_info.owner {
+            return Err(StakePoolError::InvalidFeeAccount.into());
+        }
```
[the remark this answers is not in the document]
jon-chuang: one less if-else condition, I like it

stake-pool/program/src/state.rs
```diff
@@ -115,12 +115,17 @@ pub struct StakePool {
...
     /// Fees paid out to referrers on referred deposits.
     /// Expressed as a percentage (0 - 100) of deposit fees.
-    /// i.e. `deposit_fee`% is collected as deposit fees for every deposit
+    /// i.e. `deposit_fee`% of SOL deposited is collected as deposit fees for every deposit
     /// and `referral_fee`% of the collected deposit fees is paid out to the referrer
     pub referral_fee: u8,
+
+    /// Toggles whether the `DepositSol` instruction requires a signature from
+    /// the `deposit_authority`
+    pub sol_deposit_authority: Option<Pubkey>,
```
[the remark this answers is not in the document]
jon-chuang: Yeah, I was thinking the same. More flexibility would definitely help.

stake-pool/program/tests/huge_pool.rs
```diff
@@ -32,7 +32,7 @@ use {
     spl_token::state::{Account as SplAccount, AccountState as SplAccountState, Mint},
 };
 
-const HUGE_POOL_SIZE: u32 = 4_000;
+const HUGE_POOL_SIZE: u32 = 3_950;
```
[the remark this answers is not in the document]
jon-chuang: I'm not sure this time... 4000 is definitely the nicer number, but, hopefully that won't matter in due time with message processor/VM improvements. 

Still, I'll investigate a bit.

stake-pool/program/tests/set_withdrawal_fee.rs
```diff
@@ -321,42 +321,6 @@ async fn fail_high_withdrawal_fee_increase_from_0() {
     }
 }
 
-#[tokio::test]
```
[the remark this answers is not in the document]
jon-chuang: It's actually a duplicate of fail_high_withdrawal_fee

jon-chuang:
@joncinque how do you feel about refactoring the deposit authority so that it always is controlled by the program, and then make it possible to toggle the deposit authority on/off? The struct would change so that deposit authority is set as the program's permanently, and there is a new field `manager_stake_deposit_authority: Option<Pubkey>`, and the old one is renamed `stake_account_deposit_authority`.

It's very inflexible the way it is currently, as it's really all-or-nothing at the create pool stage.

We'd like the authority to gate deposits into the pool. 

Not sure if it will break compatibiltility with deployed testnet pools.

To retain backwards compatibility, we can not enforce that the `stake_account_deposit_authority` be program derived. But we recommend that the gating stake deposits be done via `manager_stake_deposit_authority` going forward.

stake-pool/cli/src/main.rs
jon-chuang: Ok, user deserves credit for using the cli :P

stake-pool/cli/src/main.rs
jon-chuang: good point. In general, referrer is a fixed entity, else the token receiver. Perhaps `token_receiver` should also be renamed `pool_token_receiver_account`.

stake-pool/cli/src/main.rs
jon-chuang: yep, didn't realise this functionality existed!

stake-pool/cli/src/main.rs
jon-chuang: That's true. It would be nice to have the destination be an existing mergeable user stake account, however. 

The ephemeral stake account would be deleted once merged with the desination, in a single atomic transaction.

stake-pool/cli/src/main.rs
jon-chuang: Ok, I prefer `--set-to-none`, and have the two options `conflicts_with` one another

stake-pool/cli/src/main.rs
jon-chuang: ok, epoch it is.

stake-pool/cli/src/main.rs
jon-chuang: oops, thanks!

stake-pool/cli/src/main.rs
joncinque: Sure! Check out where `--unset` is being used, and rename all of those too then.  It works exactly in that way

stake-pool/program/src/processor.rs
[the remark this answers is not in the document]
joncinque: ah crap, that really stinks.  Since they're hard-coded values, it isn't a huge deal, you can let this go

stake-pool/program/tests/huge_pool.rs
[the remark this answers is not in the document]
joncinque: It's not a huge deal considering we're adding quite a bit of new logic to do the extra mints on deposit, don't spend too much time on it, I'd say.

stake-pool/program/tests/set_withdrawal_fee.rs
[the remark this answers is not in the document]
joncinque: Ah ok, disregard then, thanks!

stake-pool/cli/src/main.rs
joncinque: Yes please, great idea!

stake-pool/cli/src/main.rs
joncinque: Yeah that's been on my mental list for a bit... it would be a great feature

joncinque:
We can go that way.  Otherwise what if we control the ability to deposit through the fee?  If it's set to 100%, that's like saying "pool is private".  Then we don't need to change the deposit authority.

jon-chuang:
@joncinque , I think that approach only works for private pools, but not if you want to gate deposits on a per-validator basis. 

In which case, we're better off with a separate authority, so if that makes sense to you, we'll go ahead and implement that.

Is there still a use case for initializing with a non PDA deposit authority then? If not, should we remove that from the initialization?

joncinque:
@jon-chuang let me make sure i understand what you're suggesting

* `manager_deposit_stake_authority`: new field that can be toggled by the manager, forces deposits to be signed by this account
* `stake_account_deposit_authority`: always have this field as a PDA

Then on deposit, we just add a check for the signature if `manager_deposit_stake_authority` is set.  Before deposit, we always set the withdraw / staking authority to `stake_account_deposit_authority`.  Is that right?

jon-chuang:
> @jon-chuang let me make sure i understand what you're suggesting
> 
> * `manager_deposit_stake_authority`: new field that can be toggled by the manager, forces deposits to be signed by this account
> * `stake_account_deposit_authority`: always have this field as a PDA
> 
> Then on deposit, we just add a check for the signature if `manager_deposit_stake_authority` is set. Before deposit, we always set the withdraw / staking authority to `stake_account_deposit_authority`. Is that right?

Yes, that's precisely right.

joncinque:
All right, sounds good to me! I see what you mean about existing stake pools... let's not worry too much about it -- there's mainly the foundation stake pool, and i can always hack together an instruction to upgrade it once we're ready

stake-pool/program/src/instruction.rs
```diff
@@ -960,11 +1104,14 @@ pub fn withdraw(
         AccountMeta::new_readonly(sysvar::clock::id(), false),
         AccountMeta::new_readonly(*token_program_id, false),
         AccountMeta::new_readonly(stake_program::id(), false),
+        AccountMeta::new_readonly(system_program::id(), false),
```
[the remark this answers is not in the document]
billythedummy: yeah, seems like it. We'll remove this

stake-pool/program/src/instruction.rs
```diff
@@ -925,11 +989,91 @@ pub fn deposit_with_authority(
...
+    reserve_stake_account: &Pubkey,
+    lamports_from: &Pubkey,
+    pool_tokens_to: &Pubkey,
+    manager_fee_account: &Pubkey,
+    referrer_pool_tokens_account: &Pubkey,
+    pool_mint: &Pubkey,
+    token_program_id: &Pubkey,
+    amount: u64,
+) -> Vec<Instruction> {
+    let dummy_sol_deposit_authority = Pubkey::default();
```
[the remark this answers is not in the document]
billythedummy: i think we want to keep those fields separated for the use case where the `deposit_sol_authority` is an off-chain server or an on-chain PDA that users must go through for deposits.

jon-chuang:
Actually, @joncinque , realistically, the why shouldn't the program simply use the `withdraw_authority` to sign off on everything instead of requiring the `deposit_authority`? We could get rid of the `stake_account_deposit_authority` entirely

stake-pool/cli/src/main.rs
billythedummy: I believe the referrer is supposed to default to the pool manager if none is provided

jon-chuang:
CI is misbehaving, I decided to rebase again.

jon-chuang:
@joncinque since we now have `deposit_stake_fee` and `deposit_sol_fee`, does this mean that we also need a `stake_referral_fee` and `sol_referral_fee`? It seems like way too much to have to manage. Could we unify the referral fees in this case?

stake-pool/cli/src/main.rs
jon-chuang: This doesn't really grant support for ledger, does it?

That really requires remote wallet

jon-chuang: Anw, I've added the ask keyword in case ppl want it

joncinque: We're using remote wallet already, right?

jon-chuang: Then isn't what you want, is_signer?

jon-chuang: Oh, I see, you always ask the config for this.

joncinque:
ah, that seems to be the best solution!  and we'll still keep the "default" PDA as an option, correct?

stake-pool/cli/src/main.rs
```diff
@@ -696,10 +852,20 @@ fn command_list(config: &Config, stake_pool_address: &Pubkey) -> CommandResult {
         println!("Validator List: {}", stake_pool.validator_list);
         println!("Manager: {}", stake_pool.manager);
         println!("Staker: {}", stake_pool.staker);
-        println!("Depositor: {}", stake_pool.deposit_authority);
+        println!("Depositor: {}", stake_pool.stake_deposit_authority);
+        println!("SOL Deposit Authority: {}", sol_deposit_authority);
         println!("Withdraw Authority: {}", pool_withdraw_authority);
         println!("Pool Token Mint: {}", stake_pool.pool_mint);
         println!("Fee Account: {}", stake_pool.manager_fee_account);
+        println!("Epoch Fee: {}", stake_pool.fee);
```
joncinque, line 860: This is now being printed twice -- could you remove the other print?

stake-pool/cli/src/main.rs
```diff
@@ -659,9 +679,142 @@ fn command_deposit(
...
+        &user_sol_transfer,
+    ];
+    if let Some(keypair) = from.as_ref() {
+        signers.push(keypair)
+    }
+
+    let mut total_rent_free_balances: u64 = 0;
+
+    // Create the ephemeral SOL account
+    instructions.push(system_instruction::create_account(
```
joncinque, line 752: nit: you can just do a `transfer`, since the `allocate` + `assign` part of `create_account` is unnecessary

stake-pool/program/src/lib.rs
```diff
@@ -115,4 +117,4 @@ pub fn find_transient_stake_program_address(
     )
 }
 
-solana_program::declare_id!("SPoo1xuN9wGpxNjGnPNbRPtpQ7mHgKM8d9BeFC549Jy");
+solana_program::declare_id!("5ocnV1qiCgaQR8Jb8xWnVbApfaygJ8tNoZfgPwsgx9kx");
```
joncinque, line 120: :eyes:

stake-pool/program/src/processor.rs
```diff
@@ -2224,6 +2410,36 @@ impl Processor {
...
+
+        check_account_owner(stake_pool_info, program_id)?;
+        let mut stake_pool = try_from_slice_unchecked::<StakePool>(&stake_pool_info.data.borrow())?;
+        if !stake_pool.is_valid() {
+            return Err(StakePoolError::InvalidState.into());
+        }
+        stake_pool.check_manager(manager_info)?;
+        if for_stake_deposit {
+            stake_pool.stake_deposit_authority = new_sol_deposit_authority.unwrap_or(
+                find_deposit_authority_program_address(&crate::id(), stake_pool_info.key).0,
```
joncinque, line 2435: Use the passed `program_id` here instead of `crate::id()`, and you won't have to change the crate id

stake-pool/program/src/instruction.rs
```diff
@@ -315,10 +323,81 @@ pub enum StakePoolInstruction {
...
+    ///
+    ///  0. `[w]` StakePool
+    ///  1. `[s]` Manager
+    SetSolDepositFee {
+        /// Fee assessed as percentage of deposited SOL
         #[allow(dead_code)] // but it's not
         fee: Fee,
     },
+
+    ///  (Manager only) Update DepositStake referral fee
```
joncinque, line 392:
```suggestion
    ///  (Manager only) Update DepositSol referral fee
```

stake-pool/cli/src/main.rs
```diff
@@ -2018,17 +2472,33 @@ fn main() {
                 denominator,
                 numerator,
             };
-            command_set_fee(&config, &stake_pool_address, new_fee)
+            match arg_matches.value_of("fee_type").unwrap() {
```
joncinque, line 2475: This change-set is already getting pretty massive, but we could reduce the number of instructions and simplify the code quite a bit if we actually add this as an enum, `FeeType`.  That can be passed into generic `SetFee` and `SetReferralFee`, which would reduce the number of instructions from 5 to 2.  It would also get rid of that `for_stake_deposit` bool, which isn't very clear unfortunately.  What do you think?

billythedummy: What do you think of combining everything into 1 enum?

```
enum FeeType {
  ReferreralSol(u8),
  ReferralStake(u8),
  Epoch(Fee),
  Withdrawal(Fee),
  Deposit(Fee),
}
```

stake-pool/cli/src/main.rs
billythedummy: oh i didn't realize that block was there already, i think we can standardize that if/else block for all the fee types. It provides more info for the user

stake-pool/program/src/lib.rs
billythedummy: oops haha

jon-chuang:
> ah, that seems to be the best solution! and we'll still keep the "default" PDA as an option, correct?

Yes, that's correct. Not initializing or unsetting the stake_deposit_authority turns it back to the PDA via the priorly defined seeds.

stake-pool/cli/src/main.rs
joncinque: Yeah sure, that works!

stake-pool/cli/src/main.rs
joncinque: Either way works for me :smile:

joncinque:
Is this good to go from your side? I pushed some changes which will require a rebase unfortunately.  Don't worry about adapting to the changes too much, I mainly added a check on deposit / withdraw to make sure that a person is actually getting back more than 0 tokens, and then changed the rounding for the deposit / withdraw calcs.  Once you rebase and merge, then I can add the calcs in

jon-chuang:
To not let matters get out of hand, given that this is a 2.9K LOC PR, here is a summary of changes:


### Instructions:
Modified instructions
1. Renaming: `Deposit` -> `DepositStake`, `Withdraw`->`WithdrawStake`. 
2. Unified update fee instruction. `SetFee { Fee }` -> `SetFee { FeeType }` where `FeeType = enum { ${type}Fee(Fee) }`
3. Initialize also initializes with `stake_deposit_fee` and `stake_referral_fee`. Not sure if we want to get rid of that, we can initialize then update immediately through the cli.
4. `DepositStake` also accepts manager and referrer fee accounts to credit `deposit_stake_fee` and corresponding `stake_referral_fee`.

New instructions
1. `DepositSol`
2. `SetSolDepositAuthority`
3. `SetStakeDepositAuthority`. Leverages previously defined PDA.

Removed instructions:
1. `SetWithdrawalFee` (unified into `SetFee`)

TODO:
1. Could probably unify `deposit_sol_with_authority` by passing an option instead of having two instruction constructors.
2. Not sure if we want to maintain separate instruction constructors for each `FeeType`, or to simply expose the generic versions (right now, we expose both).
3. Not sure if we want to unify `SetStakeDepositAuthority` and `SetSolDepositAuthority` into single instruction, also using enums. Currently, the two instructions call the same `process_set_deposit_authority`.

### Processor changes (most crucial security-wise)
1. Apply fees to `DepositStake`. Additional checks: manager fee account matches `StakePool` struct's.
2. Add `DepositSol` instruction, basically same as `DepositStake` except for source of lamports (activated stake accounts' staked SOL + rent-free reserve v.s. signing source account). The major difference is `DepositSol` checks the system program (for carrying out SOL transfer), does not ingest or check the validator list account.
3. Unified `process_set_deposit_authority`.

Potential Security issues:
1. whether or not additional fee calculations introduce weird integer error bugs. (to a certain degree, no, as the errors are < 1 lamport per fee application).
2. Whether or not any mistakes were made in calculating input SOL v.s. output pool tokens (probably not, due to common use of previously defined `stake_pool.calc_pool_tokens_for_deposit`.

TODO:
1. Maybe we should add asserts to ensure that `pool_tokens_user + pool_tokens_manager_deposit_fee + pool_tokens_referral_fee == new_pool_tokens`.

### Struct changes:
1. `pub sol_deposit_authority: Option<Pubkey>`
2. `pub sol_deposit_fee: Fee`
3. `pub sol_referral_fee: u8`

### Tests:
We've added tests for all of the changes, including:
- set deposit authority
- set deposit fee 
- set referral fee
- deposit sol
- new test for deposit stake (with referral fee/invalid referrer)

###  Cli changes: 
1. create pool uses `stake_deposit_authority` keypair to initialize. 
2. new commands: 
a. `deposit_sol`, 
b. `set_deposit_authority` (for both deposit sol and deposit stake, via `for_stake_deposit` boolean).
c. monomorphized as `set-${type}-deposit-authority`.
d. Unified `set-stake-referral-fee` for both stake and sol deposits.
3. modified commands: `deposit-stake` (+ referrer account, + manager fee). command `list` (display new fields). Unified `set_fee` across 4 `Fee` kinds to reduce code bloat. 

TODO:
1. `--from` can be a hardware wallet (nice to have, future PR).

jon-chuang:
@joncinque I think I'm happy with this PR.

In terms of future changes we think would be good (nice to have, but not critical):
1. Multiple transient stake accounts (up to about 8). Iterate over a range in `[0, MAX_TRANSIENT_ACCS)` which we add to the transient account seed. I think we all agree that a single transient account is neither necessary nor ideal. To 

I saw you mentioning that the stake pool code is undergoing another round of audits, will these changes make it in time? We'd like to wrap things up on the stake pool program front for our first deployment to mainnet, hopefully in about 1 week's time. Perhaps if things can move fast enough we may depend on the solana-deployed version of the program after all, and it can be made to eventually be immutable, just like the Token program is. 

In terms of immutability, I'm just a bit worried about slashing/changes to the stake program.

joncinque:
Great work!

PR #2152, opened 2021-07-27
Governance: Capture vote_threshold_percentage on voted on proposal

SebastianBor:
#### Summary

Capture  `vote_threshold_percentage` on voted on proposal to correctly display historical vote results.
 `vote_threshold_percentage` can be changed over time through governance and hence we have to store the value used to resolve proposal vote.


joncinque (commented):
Looks good!

PR #2166, opened 2021-07-29
Governance: Realm config

SebastianBor:
#### Summary

Introduce `RealmConfig` to capture realm level configuration and make it possible to change it through governance in future versions. The `RealmConfig` contains:
-  `council_mint` - council mint was moved to the config to make it possible to change it
-   `community_mint_max_vote_weight_source` - defines how the max vote weight is calculated for community tokens
- `custodian` - none security critical authority with realm maintenance powers 

A following PR would implement `SetRealmConfig` instruction which would be used to change `RealmConfig`. It would have to be signed by realm authority and could be voted on through proposals

### DIFF
--- a/governance/program/src/state/realm.rs
+++ b/governance/program/src/state/realm.rs
@@ -14,6 +14,17 @@ use crate::{
 
 use crate::state::enums::GovernanceAccountType;
 
+/// Realm Config
+#[repr(C)]
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub struct RealmConfig {
+    /// Optional council mint
+    pub council_mint: Option<Pubkey>,
+
+    /// Reserved space for future versions
+    pub reserved: [u8; 8],
+}
+
 /// Governance Realm Account
 /// Account PDA seeds" ['governance', name]
 #[repr(C)]
@@ -25,11 +36,11 @@ pub struct Realm {
     /// Community mint
     pub community_mint: Pubkey,
 
+    /// Configuration of the Realm
+    pub config: RealmConfig,
+
     /// Reserved space for future versions
     pub reserved: [u8; 8],
-
-    /// Council mint
-    pub council_mint: Option<Pubkey>,
 
     /// Realm authority. The authority must sign transactions which update the realm (ex. adding governance, setting council)
     /// The authority can be transferer to Realm Governance and hence make the Realm self governed through proposals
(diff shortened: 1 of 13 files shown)


governance/program/src/state/realm.rs
```diff
@@ -8,11 +8,32 @@ use solana_program::{
...
+
+    /// An authority tasked with none critical and maintenance Realm operations
+    /// For example custodian authority is required to add governances to the Realm
+    /// There is no security risk with adding governances to the Realm but it should not be open for everybody
+    /// to prevent unrelated entries and noise
+    /// Note: This field is not used yet. It's reserved for future versions
+    pub custodian: Option<Pubkey>,
+
+    /// Reserved space for future versions
+    pub reserved: [u8; 8],
```
joncinque, line 35: Since you have the reserved bytes before the `Option` and `String` earlier, maybe it makes sense to do the same thing here, and put `council_mint` later in the struct as well

governance/program/src/state/realm.rs
```diff
@@ -8,11 +8,32 @@ use solana_program::{
...
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub struct RealmConfig {
+    /// Optional council mint
+    pub council_mint: Option<Pubkey>,
+
+    /// The source used for community mint max vote weight source
+    /// Note: This field is not used yet. It's reserved for future versions
+    pub community_mint_max_vote_weight_source: MintMaxVoteWeightSource,
+
+    /// An authority tasked with none critical and maintenance Realm operations
```
joncinque, line 27: `none critical` -> `non-critical`?

governance/program/tests/program_test/mod.rs
```diff
@@ -185,10 +185,16 @@ impl GovernanceProgramTest {
...
             community_mint: community_token_mint_keypair.pubkey(),
-            council_mint: Some(council_token_mint_keypair.pubkey()),
+
             name,
             reserved: [0; 8],
             authority: Some(realm_authority.pubkey()),
+            config: RealmConfig {
+                council_mint: Some(council_token_mint_keypair.pubkey()),
+                reserved: [0; 8],
+                custodian: Some(realm_authority.pubkey()),
```
joncinque, line 195: Once you're actually testing this, it could be better to use a different authority to avoid false positives

joncinque (commented):
Looks good! Just a few comments

governance/program/src/state/realm.rs
SebastianBor: Done

governance/program/src/state/realm.rs
SebastianBor: done

governance/program/tests/program_test/mod.rs
SebastianBor: yeap, done in next PR

PR #2180, opened 2021-07-29
stake-pool: Don't assess withdrawal fee from fee account

joncinque:
#### Problem

If a stake pool has a withdraw fee, even the fee account will be assessed a fee, making it very annoying for the manager to get SOL out of the pool.

#### Solution

Don't assess the withdrawal fee for the fee account.

cc @jon-chuang what do you think?


jon-chuang:
We had thought of that problem, Zeno's paradox!

Excellent fix, 100% with this.

PR #2188, opened 2021-07-30
Governance: Realm config

SebastianBor:
#### Summary 

- `SetRealmConfig` - instruction sets realm config 
- Pass all realm config parameters to `CreateRealm` instruction
- Change mint max vote weight percentage to fraction with 10^10 precision 
- Apply max vote weight fraction when calculating vote results 

Note: In this version it's not possible to change realm council. It's only possible to remove it which is the most common scenario when after initial realm setup the supervisory role of the council is removed.

### DIFF
--- a/governance/program/src/processor/process_create_realm.rs
+++ b/governance/program/src/processor/process_create_realm.rs
@@ -32,13 +32,14 @@ pub fn process_create_realm(
     let account_info_iter = &mut accounts.iter();
 
     let realm_info = next_account_info(account_info_iter)?; // 0
-    let governance_token_mint_info = next_account_info(account_info_iter)?; // 1
-    let governance_token_holding_info = next_account_info(account_info_iter)?; // 2
-    let payer_info = next_account_info(account_info_iter)?; // 3
-    let system_info = next_account_info(account_info_iter)?; // 4
-    let spl_token_info = next_account_info(account_info_iter)?; // 5
+    let realm_authority_info = next_account_info(account_info_iter)?; // 1
+    let governance_token_mint_info = next_account_info(account_info_iter)?; // 2
+    let governance_token_holding_info = next_account_info(account_info_iter)?; // 3
+    let payer_info = next_account_info(account_info_iter)?; // 4
+    let system_info = next_account_info(account_info_iter)?; // 5
+    let spl_token_info = next_account_info(account_info_iter)?; // 6
 
-    let rent_sysvar_info = next_account_info(account_info_iter)?; // 6
+    let rent_sysvar_info = next_account_info(account_info_iter)?; // 7
     let rent = &Rent::from_account_info(rent_sysvar_info)?;
 
     if !realm_info.data_is_empty() {
@@ -80,13 +81,6 @@ pub fn process_create_realm(
         None
     };
 
-    let realm_authority = if config_args.use_authority {
-        let realm_authority_info = next_account_info(account_info_iter)?;
-        Some(*realm_authority_info.key)
-    } else {
-        None
-    };
-
     let realm_custodian = if config_args.use_custodian {
         let realm_custodian_info = next_account_info(account_info_iter)?;
         Some(*realm_custodian_info.key)
@@ -100,7 +94,7 @@ pub fn process_create_realm(
 
         name: name.clone(),
         reserved: [0; 8],
-        authority: realm_authority,
+        authority: Some(*realm_authority_info.key),
         config: RealmConfig {
             council_mint: council_token_mint_address,
             reserved: [0; 8],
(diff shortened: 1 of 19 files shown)


governance/program/src/processor/process_create_realm.rs
```diff
@@ -57,11 +61,16 @@ pub fn process_create_realm(
         rent,
     )?;
 
-    let council_token_mint_address = if let Ok(council_token_mint_info) =
-        next_account_info(account_info_iter)
-    // 7
-    {
-        let council_token_holding_info = next_account_info(account_info_iter)?; //8
+    let realm_custodian = if config_args.use_custodian {
```
joncinque, line 64: How does this work if `council_token_mint = Some(...)` and `realm_custodian = None`? Is that the situation that's meant to fail?

governance/program/src/processor/process_set_realm_config.rs
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+        None
+    };
+
+    if config_args.use_council_mint {
+        let council_token_mint_info = next_account_info(account_info_iter)?;
+
+        // Council mint can only be at present set to none (removed) and changing it to other mint is not supported
+        // It might be implemented in future versions but it needs careful planning
+        // It can potentially open a can of warms like what happens with existing deposits or pending proposals
```
joncinque, line 47: Seems like a bit of a footgun -- if you have a council mint, and want to set the realm config, you have to keep sending in the council mint, and if you forget, it's unset forever.  Is that normal?

governance/program/src/state/proposal.rs
```diff
@@ -243,51 +248,90 @@ impl Proposal {
...
+                // When the fraction is used it's possible we can go over the calculated max_vote_weight
+                // and we have to adjust it in case more votes have been cast
+                let total_vote_count = self
+                    .yes_votes_count
+                    .checked_add(self.no_votes_count)
+                    .unwrap();
+
+                Ok(max_vote_weight.max(total_vote_count))
+            }
+            _ => Err(GovernanceError::VoteWeightSourceNotSupported.into()),
```
joncinque, line 283: nit: can this be `MintMaxVoteWeightSource::Absolute`?

governance/program/src/state/realm.rs
```diff
@@ -194,6 +208,20 @@ pub fn get_governing_token_holding_address(
     .0
 }
 
+/// Asserts given realm config args are correct
+pub fn assert_valid_realm_config_args(config_args: &RealmConfigArgs) -> Result<(), ProgramError> {
+    match config_args.community_mint_max_vote_weight_source {
+        MintMaxVoteWeightSource::SupplyFraction(fraction) => {
+            if !(1..=MintMaxVoteWeightSource::SUPPLY_FRACTION_BASE).contains(&fraction) {
```
joncinque, line 215: nit: does this perform well for `SUPPLY_FRACTION_BASE` in terms of compute units? I would have just done `1 <= fraction && fraction <= SUPPLY_FRACTION_BASE` to avoid generating a range or iterating through so many values.  I guess the compiler knows how to deal with this

governance/program/src/state/realm.rs
```diff
@@ -194,6 +208,20 @@ pub fn get_governing_token_holding_address(
...
+                return Err(GovernanceError::InvalidMaxVoteWeightSupplyFraction.into());
+            }
+        }
+        _ => return Err(GovernanceError::MintMaxVoteWeightSourceNotSupported.into()),
```
joncinque, line 219: nit: use `MintMaxVoteWeightSource::Absolute` here

joncinque (commented):
Just a few comments, looks good!

governance/program/src/state/proposal.rs
SebastianBor: done

governance/program/src/state/realm.rs
SebastianBor: done

governance/program/src/state/realm.rs
SebastianBor: Yes, it creates a range with start and end and checks only them. 
This change was suggested by clippy [manual_range_contains](https://rust-lang.github.io/rust-clippy/master/#manual_range_contains)

governance/program/src/processor/process_create_realm.rs
SebastianBor: No, it was not meant to fail, but  since the custodian was removed in the next PR it's irrelevant now.

governance/program/src/processor/process_set_realm_config.rs
SebastianBor: Yeah, I agree it's a bit of footgun but I think there are some mitigation factors.
1)  My thinking was that you always set the full realm config and not it's individual parts to make it a single and simpler instruction. 
2) In future versions I would like to support changing the council (and not just removing), but just wanted to avoid the complexity for now since I don't think it's going to be a common case anyway. Also the assumption is that any changes to the realm would go through a proposal vote so the chances of removing the council by mistake are rather slim.

PR #2210, opened 2021-08-03
lending: fix slots per year calculation

jordansexton:
The current calculation of slots per year for interest loses precision by performing division before multiplication.

### DIFF
--- a/token-lending/program/src/state/mod.rs
+++ b/token-lending/program/src/state/mod.rs
@@ -31,7 +31,7 @@ pub const UNINITIALIZED_VERSION: u8 = 0;
 
 /// Number of slots per year
 pub const SLOTS_PER_YEAR: u64 =
-    DEFAULT_TICKS_PER_SECOND / DEFAULT_TICKS_PER_SLOT * SECONDS_PER_DAY * 365;
+    DEFAULT_TICKS_PER_SECOND * SECONDS_PER_DAY * 365 / DEFAULT_TICKS_PER_SLOT;
 
 // Helpers
 fn pack_decimal(decimal: Decimal, dst: &mut [u8; 16]) {


jstarry:
This was definitely an unintentional loss of precision but it looks like the value was closer to reality before the fix than after. Right now we see slots taking about 550ms on mainnet-beta instead of the target time of 400ms. The loss of precision effectively means that we calculate slot time as if there are 2 slots per second rather than 2.5 slots per second.

I think a better fix would be to use unix timestamps and move away from using slot time estimation entirely

nope-finance:
@jstarry i feel like using the unix timestamps may pose other issues. I dont have the deepest understanding of how the runtime works but a few concerns listed below lmk what you think?

1 precision wise it might be path dependent (2 slot taking 400 then 600 ms might yield different end result compared to 2 slots taking 500 and 500 ms depending on implementation)

2 My rough understanding of the timestamps is that the runtime manipulates them to be per block bounded (e.g. look on testnet and see the timestamps are off my 5~6 days) which seems kind of concerning 

3. again rough understanding and more hypothetical/handwavy, the lead node gets to select the timestamp (and it just has to seem reasonable for the other nodes to agree) but i could imagine a scenario where validators might want to manipulate the timestamps closer together or further apart for somewhat questionable stuff or some form of MEV (VEV?). for example I want to take a loan on slot X and repay on slot X+1 (and suppose those are in the same block), maybe I could pay the lead validator for the block to put X and X+1 timestamps closer together such that i pay less interest?

jstarry:
> @jstarry i feel like using the unix timestamps may pose other issues

> 1. precision wise it might be path dependent (2 slot taking 400 then 600 ms might yield different end result compared to 2 slots taking 500 and 500 ms depending on implementation)

Precision is definitely a concern but this is no different from interest accrued for 400 slots then 600 slots vs 500 slots then 500 slots.

> 2. My rough understanding of the timestamps is that the runtime manipulates them to be per block bounded (e.g. look on testnet and see the timestamps are off my 5~6 days) which seems kind of concerning

This is a valid concern. Timestamps can drift in the case of cluster stalls and can only be corrected at a certain catchup speed. However, this scenario applies to using slots elapsed as well. Something like a decentralized clock oracle could be an alternative solution, but I'm not sure it's necessary.

> 3. again rough understanding and more hypothetical/handwavy, the lead node gets to select the timestamp (and it just has to seem reasonable for the other nodes to agree) but i could imagine a scenario where validators might want to manipulate the timestamps closer together or further apart for somewhat questionable stuff or some form of MEV (VEV?). for example I want to take a loan on slot X and repay on slot X+1 (and suppose those are in the same block), maybe I could pay the lead validator for the block to put X and X+1 timestamps closer together such that i pay less interest?

On Solana, the lead node can't select the timestamp, it's deterministically calculated using a stake weighted calculation from the timestamps of all recent votes in consensus. The leader could censor votes during its slots to slow down time, though. But that would come at the cost of potentially censoring votes that help its produced block get confirmed.

Paying less interest in this case seems pretty trivial since it would effectively be at most like 1s of accrued interest savings.

nope-finance:
> Precision is definitely a concern but this is no different from interest accrued for 400 slots then 600 slots vs 500 slots then 500 slots.
err im not worries about slot I was more worried about 400 miliseconds then 600 miliseconds in succesive blocks compares to 500 and 500. the reason being that unix timestamps are ~500x more granular than slot numbers so you might imagine that doing with an interest rate of X each ms
`val_0=C -> val_1 = val_0 * 1**(1+X)**403 ->  val_2 = val_1 * 1**(1+X)**597`
might have a different outcome (due to precision) than
`val_0=C -> val_1 = val_0 * 1**(1+X)**501 ->  val_2 = val_1 * 1**(1+X)**499`
if that makes sense.

> This is a valid concern. Timestamps can drift in the case of cluster stalls and can only be corrected at a certain catchup speed. However, this scenario applies to using slots elapsed as well. Something like a decentralized clock oracle could be an alternative solution, but I'm not sure it's necessary.

I was thinking about timestamp oracles (or alternatively twap slot_time oracles because timestamp ones would have to be updated and propagated every block which is not really feasible right now) but yeh maybe it is overkill

> On Solana, the lead node can't select the timestamp, it's deterministically calculated using a stake weighted calculation from the timestamps of all recent votes in consensus. The leader could censor votes during its slots to slow down time, though. But that would come at the cost of potentially censoring votes that help its produced block get confirmed.
>Paying less interest in this case seems pretty trivial since it would effectively be at most like 1s of accrued interest savings.

this is fair just wanted to make sure there wasn't some edge case not considered

PR #2252, opened 2021-08-11
stake-pool: Address a few points post-liquid staking

joncinque:
A few cleanup bits from the great work done in #2141

Program:
* fix checking sol deposit auth <-- this one's important
* add 0 pool token check for user, so that they don't deposit into the pool and get nothing back!
* avoid using `+` in favor of `saturating_add` for one check in particular (which may even be optimized away by the compiler)
* remove unused instruction creators

CLI:
* unify `set-sol-deposit-authority` and `set-stake-deposit-authority` into `set-deposit-authority`
* unify `set-sol-referral-fee`and `set-stake-referral-fee` into `set-referral-fee`

cc @jon-chuang @billythedummy

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -1886,13 +1886,15 @@ impl Processor {
             .checked_sub(pool_tokens_referral_fee)
             .ok_or(StakePoolError::CalculationFailure)?;
 
-        if pool_tokens_user + pool_tokens_manager_deposit_fee + pool_tokens_referral_fee
+        if pool_tokens_user
+            .saturating_add(pool_tokens_manager_deposit_fee)
+            .saturating_add(pool_tokens_referral_fee)
             != new_pool_tokens
         {
             return Err(StakePoolError::CalculationFailure.into());
         }
 
-        if new_pool_tokens == 0 {
+        if pool_tokens_user == 0 {
             return Err(StakePoolError::DepositTooSmall.into());
         }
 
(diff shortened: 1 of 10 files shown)


stake-pool/program/src/processor.rs
```diff
@@ -2005,9 +2007,7 @@ impl Processor {
             program_id,
             stake_pool_info.key,
         )?;
-        if let Some(sol_deposit_authority) = sol_deposit_authority_info {
-            stake_pool.check_sol_deposit_authority(sol_deposit_authority)?;
-        }
```
joncinque, line 2010: This was only doing the check if the last account was passed, which means that it was possible to bypass it totally if you just didn't pass the last account.

stake-pool/program/src/processor.rs
```diff
@@ -2005,9 +2007,7 @@ impl Processor {
...
+        stake_pool.check_sol_deposit_authority(sol_deposit_authority_info)?;
```
jon-chuang, line 2010: Hmm, I'm not sure this is the behaviour we want, shouldn't the sol deposit authority be optional?

jon-chuang: Oh wait, I missed that the ? operator only triggers the error if there is indeed Some(deposit_sol_auth)

stake-pool/program/src/processor.rs
jon-chuang: Oh man, that's a great catch, thanks

jon-chuang:
Thanks for cleaning up @joncinque , LGTM

Issue #2234, opened 2021-08-06
Token-swap js bindings don't allow passing curve parameters


ilmoi:
With current js bindings it's not possible to instantiate a Token Swap with curves that require parameters during initilization.

An example is the fixed price curve, which requires a single parameter:
```rust
pub struct ConstantPriceCurve {
    /// Amount of token A required to get 1 token B
    pub token_b_price: u64,
}
``` 

The layout is correctly planned out to include [32 bytes for curve parameters](https://github.com/solana-labs/solana-program-library/blob/master/token-swap/js/src/index.ts#L216), but the actual `commandDataLayout.encode()` call stops just 1 argument short and doesn't include `curveParamters`. As a result the program always reads them as 0s, which leads to the following error:

```
'Program log: Error: Given pool token amount results in zero trading tokens'
```

Which makes sense since the exchange rate is basically set to 0.

The solution would be to simply add a buffered uint 8 array along the lines of:
```js
const encodeLength = commandDataLayout.encode(
        let params = new Uint8Array(32); //<-- 32 byte long array
        //somehow modify it based on user's arguments
        {
          instruction: 0, // InitializeSwap instruction
          nonce,
          tradeFeeNumerator,
          tradeFeeDenominator,
          ownerTradeFeeNumerator,
          ownerTradeFeeDenominator,
          ownerWithdrawFeeNumerator,
          ownerWithdrawFeeDenominator,
          hostFeeNumerator,
          hostFeeDenominator,
          curveType,
          curveParameters: Buffer.from(params), //<-- then add it as the last argument here
        },
        data,
      );
      data = data.slice(0, encodeLength);
```
And to then also add a new argument for the constructor / init method that asks the user what, if any, curve params they want to pass. I've done a rough test locally and it works.

I'm happy to work on this and submit a PR. I'm just new to solana community / open source and wasn't sure whether it's ok to directly do that or whether I need open an issue first. If someone from the team could comment that'd be great.

joncinque:
You've got this exactly right, the work was just never done to add this to the JS bindings.  Please feel free to add a PR to integrate the curve params!  You can tag me in the PR to review whenever youget it done and use this issue to track the work if you like

PR #2258, opened 2021-08-11
added curve parameters in token swap

ilmoi:
Given that all curves currently take either no parameters at all (Constant Product) or a single u64 (Constant Price, Stable, Offset) - I decided to accept a single, optional u64 as a parameter fom the user. 

I thought that's going to make the user's life a lot easier than asking them to encode a raw 32 byte buffer. In the future if new curves are added that take different params, this can be changed.

I left a comment to point that out.

@joncinque - you might want to link the issue (https://github.com/solana-labs/solana-program-library/issues/2234) - I don't have priviliges, I don't think. Also let me know if you want package-lock.json removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ hfuzz_workspace
 **/*.so
 **/.DS_Store
 test-ledger
+.idea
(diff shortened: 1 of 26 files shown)


joncinque:
This looks great, thanks for the contribution! Could you just add a test like `createTokenSwap` that simply instantiates a pool with some other curve that requires the parameter?

ilmoi:
Looks like I messed up when rebasing. Let me open a fresh PR.

Issue #2260, opened 2021-08-11
stake-pool: instant stake activation


joncinque:
#### Problem

Any user can deposit a stake account with extra, non-staked lamports into the pool receiving pool tokens for all lamports -> pool tokens can be redeemed for a fully staked stake account.

#### Solution

A user depositing stake should only receive pool tokens equal to normal stake account rent-exemption (2.2 million lamports) + stake's delegation.  Note that we shouldn't use the provided stake account's rent exempt amount, because a user could do another version of this attack if they allocate a huge amount of data (1 MB for example), then provide that account with a small stake.  The rest of the stake is just taken by the pool, and that's fine.

jon-chuang:
@joncinque , for consistency, I consider everything apart from the `delegation.stake`, including the rent exempt amount, as being sol deposits.

I'm not sure if this is an undesirable property.

For instance, a pool manager may want to forbid direct SOL deposits, but this mechanism would allow such deposits.

Perhaps we should stick to the choice that a user providing extra lamports beyond `delegation.stake + RENT_EXEMPT` will not be reimbursed with additional pool tokens?

PR #2270, opened 2021-08-12
stake-pool: `DepositStake` only issues tokens for `delegation.stake` + `min(undelegated_stake_rent, additional_lamports)`

jon-chuang:
Fixes https://github.com/solana-labs/solana-program-library/issues/2260


stake-pool/program/src/processor.rs
```diff
@@ -1869,19 +1873,21 @@ impl Processor {
             .ok_or(StakePoolError::CalculationFailure)?;
 
         let new_pool_tokens = stake_pool
-            .calc_pool_tokens_for_deposit(all_deposit_lamports)
+            .calc_pool_tokens_for_deposit(stake_deposit_lamports + stake_rent)
```
joncinque, line 1876: not technically needed, but just for paranoia
```suggestion
            .calc_pool_tokens_for_deposit(stake_deposit_lamports.saturating_add(stake_rent))
```

joncinque (commented):
Great work! just one little suggestion then we can integrate this

jon-chuang:
@joncinque some notes on the choices made:
- Additional lamports can be 0 or `stake_lamports`, and correspondingly, `stake_deposit_lamports` is `delegation.stake + rent` or `delegation.stake`, depending on the situation. I believe it has something to do with differences in how merging is performed depending on whether a stake account is fully activated (rent becomes inactive lamports) or is activating (rent becomes delegation.stake lamports). 
- The way I got it to work consistently is to consider the amount "credited" towards the `total_stake_lamports` and towards pool tokens to be the newly added `delegation.stake` + `additional_lamports` capped at the limit of the rent amount. 
- When the stake is fully active, all rent will fully use up the quota of `additional_lamports`, hence result in an adequate solution.

stake-pool/program/src/processor.rs
```diff
@@ -1958,9 +1975,11 @@ impl Processor {
             .pool_token_supply
             .checked_add(new_pool_tokens)
             .ok_or(StakePoolError::CalculationFailure)?;
+        // We treat the extra lamports as though they were
+        // transferred directly to the reserve stake account.
         stake_pool.total_stake_lamports = stake_pool
             .total_stake_lamports
-            .checked_add(all_deposit_lamports)
+            .checked_add(credited_deposit_lamports)
```
joncinque, line 1982: Since we're withdrawing `additional_lamports` into the reserve, and the reserve lamports are part of `total_stake_lamports`, this should be `all_deposit_lamports` still

joncinque (commented):
just one last comment, then this is good to go!

stake-pool/program/src/processor.rs
jon-chuang: technically yes, but I wanted it to be lazily updated by `update_pool_balance` since I wanted it to be equivalent to directly transferring lamports into the reserve account - no one knows about that until the next update.

joncinque: If someone deposits too much into the stake pool, it gets more valuable right away, and not at the start of the next epoch, right?  If someone withdraws right after this, they should get the right amount, so I say go with `all_deposit_lamports`

jon-chuang: Ok, fair enough

joncinque:
Looks great, thanks for the work!

PR #2271, opened 2021-08-12
added support for curve params

### LINKED ISSUE #2234: Token-swap js bindings don't allow passing curve parameters

ilmoi:
Given that all curves currently take either no parameters at all (Constant Product) or a single u64 (Constant Price, Stable, Offset) - I decided to accept a single, optional u64 as a parameter fom the user.

I thought that's going to make the user's life a lot easier than asking them to encode a raw 32 byte buffer. In the future if new curves are added that take different params, this can be changed.

I left a comment to point that out.

Added an extra test as well.

@joncinque - you might want to link the issue (#2234) - I don't have priviliges, I don't think. Also let me know if you want package-lock.json removed.

Fixes #2234


joncinque:
Very slick! Once this passes CI, I'll merge it in.

joncinque:
Just ran `npm run pretty:fix` to get around the CI, should be good to go

joncinque:
Thanks for the contribution!

PR #2272, opened 2021-08-12
fix stake-pool: preferred validator can be set to validator not part of the pool

### LINKED ISSUE #2262: stake-pool: preferred validator can be set to validator not part of the pool
#### Problem

The preferred withdraw / deposit vote address can be set to a validator that is not part of the pool, by 1) removing the validator from the pool (state of the entry is now ReadyForRemoval) 2) changing the preferred address to that validator 3) calling CleanupRemovedValidatorEntries only after 2.

#### Solution

Avoid setting the preferred validator if it's marked as `ReadyForRemoval`.

billythedummy:
closes #2262

- denies setting validator to validators with status `ReadyForRemoval`
- add test for this in `set_preferred`

@jon-chuang

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -1315,9 +1315,17 @@ impl Processor {
                 vote_account_address.as_ref(),
                 ValidatorStakeInfo::memcmp_pubkey,
             );
-            if maybe_validator_stake_info.is_none() {
-                msg!("Validator for {} not present in the stake pool, cannot set as preferred deposit account");
-                return Err(StakePoolError::ValidatorNotFound.into());
+            match maybe_validator_stake_info {
+                Some(vsi) => {
+                    if vsi.status == StakeStatus::ReadyForRemoval {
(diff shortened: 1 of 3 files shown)


stake-pool/program/src/processor.rs
```diff
@@ -1315,9 +1315,17 @@ impl Processor {
                 vote_account_address.as_ref(),
                 ValidatorStakeInfo::memcmp_pubkey,
             );
-            if maybe_validator_stake_info.is_none() {
-                msg!("Validator for {} not present in the stake pool, cannot set as preferred deposit account");
-                return Err(StakePoolError::ValidatorNotFound.into());
+            match maybe_validator_stake_info {
+                Some(vsi) => {
+                    if vsi.status == StakeStatus::ReadyForRemoval {
```
joncinque, line 1320: I think we should also avoid a validator stake in the `DeactivatingTransient` state, so how about this becomes:
```suggestion
                    if vsi.status != StakeStatus::Active {
```

joncinque (commented):
Just one comment

PR #2273, opened 2021-08-12
fix: stake-pool require signature in SetManager

### LINKED ISSUE #2265: stake-pool: Require signature in `SetManager`
#### Problem

SetManager does not require signature from new manager, while Initialize does

#### Solution

You probably want to require the signature of the new manager in SetManager as well, to make sure that the new manager is a key with a valid private key

billythedummy:
Closes #2265 

~~Pending confirmation on whether to do the same for `SetStaker`~~

@jon-chuang


joncinque:
Looks great, nice job!

Issue #2261, opened 2021-08-11
stake-pool: Handle closed manager fee account


joncinque:
#### Problem

The manager fee account may be closed -> no more fees can be paid, so no withdraw possible.

#### Solution

Properly handle the situation where the manager fee account may be closed

Issue #2264, opened 2021-08-11
stake-pool: Add check for fee account in SetManager


joncinque:
#### Problem

`SetManager` allows setting a `manager_fee_account` that is not a valid SPL token account

#### Solution

Add a check

joncinque:
This has been fixed for a bit

PR #2277, opened 2021-08-12
stake-pool: skip manager fee if invalid account, check new manager fee owned by `spl_token`

jon-chuang:
Fixes: https://github.com/solana-labs/solana-program-library/issues/2261 https://github.com/solana-labs/solana-program-library/issues/2264

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -2180,7 +2180,7 @@ impl Processor {
         // To prevent a faulty manager fee account from preventing withdrawals
         // if the token program does not own the account, or if the account is not initialized
         let pool_tokens_fee = if stake_pool.manager_fee_account == *burn_from_pool_info.key
-            || !stake_pool.check_manager_fee_info(manager_fee_info).is_ok()
+            || stake_pool.check_manager_fee_info(manager_fee_info).is_err()
         {
             0
         } else {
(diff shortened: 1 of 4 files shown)


stake-pool/program/src/processor.rs
```diff
@@ -2171,7 +2171,12 @@ impl Processor {
             return Err(StakePoolError::InvalidState.into());
         }
 
-        let pool_tokens_fee = if stake_pool.manager_fee_account == *burn_from_pool_info.key {
+        // To prevent a faulty manager fee account from preventing withdrawals
+        // if the token program does not own the account, or if the account is not initialized
+        let pool_tokens_fee = if stake_pool.manager_fee_account == *burn_from_pool_info.key
+            || check_account_owner(manager_fee_info, &stake_pool.token_program_id).is_err()
+            || Account::unpack(&manager_fee_info.data.borrow())?.state != AccountState::Initialized
```
joncinque, line 2178: We should bundle these checks to be reused in other places, like deposit, sol deposit, and update stake pool balance.  The most important one is update stake pool balance, because failing that would also brick the pool for everyone.

`Account::unpack()` can fail for a lot of reasons, so we could still run into this issue with this change.  To make sure that `manager_fee_info` is valid, we need:

* owned by token_program_id
* unpacks to `Account`
* state is `Initialized`
* mint is correct

if any of those fails, give back `0`

joncinque (commented):
Looks great! We just need to add some more checks, and include them in a few more places

stake-pool/program/src/processor.rs
jon-chuang: Ok, done

joncinque:
Looks good, thanks for the work!

PR #2285, opened 2021-08-15
chore(token): expose transfer checked instruction

### LINKED ISSUE #2284: Token JS Client: Expose createTransferCheckedInstruction static method from Token class
# Overview

There's a static method in the Token class named `createTransferCheckedInstruction`, which is pretty convenient if you want to easily generate a TransferChecked TransactionInstruction. The mentioned method is not exposed in the [declarations file](https://github.com/danmt/solana-program-library/blob/master/token/js/module.d.ts). 

## Solution

While I'm not entirely sure how the entire build/release process goes, I fixed it locally by adding the below snippet into the [Token class declaration](https://github.com/danmt/solana-program-library/blob/master/token/js/module.d.ts#L67) in my node_modules.

```typescript
static createTransferCheckedInstruction(
      programId: PublicKey,
      source: PublicKey,
      mint: PublicKey,
      destination: PublicKey,
      owner: PublicKey,
      multiSigners: Array<Signer>,
      amount: number | u64,
      decimals: number,
    ): TransactionInstruction;
```

I would expect this to be a simple change and I can gladly put the PR together.

danmt:
closes #2284

### DIFF
--- a/token/js/module.d.ts
+++ b/token/js/module.d.ts
@@ -247,6 +247,16 @@ declare module '@solana/spl-token' {
       authority: PublicKey,
       multiSigners: Array<Signer>,
     ): TransactionInstruction;
+    static createTransferCheckedInstruction(
+      programId: PublicKey,
+      source: PublicKey,
+      mint: PublicKey,
+      destination: PublicKey,
+      owner: PublicKey,
+      multiSigners: Array<Signer>,
+      amount: number | u64,
+      decimals: number,
+    ): TransactionInstruction;
     static createAssociatedTokenAccountInstruction(
       associatedProgramId: PublicKey,
       programId: PublicKey,


danmt:
@mvines I'm tagging you just for follow up. If there's anything else I'll need to do, I'll gladly update the PR. Thanks for the guidance.

mvines:
Thanks for the PR!

danmt:
Any time!

mvines:
Would you like a new npm package with this change, or no rush?

danmt:
Would love it but there aren't lives on the line. No rush.

*Keeps checking every 5 minutes*

mvines:
0.1.7 is published

danmt:
I think something broke during the build.

Here's a screenshot of the package in 0.1.6:

Screenshot from 2021-08-16 20-48-08

Here's one with 0.1.7:

Screenshot from 2021-08-16 20-52-03

It looks like it chopped out the lib directory. I'm not sure what happened there.

danmt:
I'm not entirely sure how the types declarations are working but I just noticed there's a [module.d.ts](https://github.com/solana-labs/solana-program-library/blob/master/token/js/module.d.ts) which I updated and a [module.flow.js](https://github.com/solana-labs/solana-program-library/blob/master/token/js/module.flow.js). 

From where I'm standing, it looks like I had to update the module.flow.js **only** and the pipeline generates the .d.ts version. I might be wrong here though. I'm available to patch right away.

mvines:
oops, I forgot to run `npm run build` before publishing, hopefully this is now fixed in 0.1.8

danmt:
I just thought I broke the token library in my first contribution, that would be starting off on the wrong foot real bad.

BTW - It seems like the package is using flow-typed for the type declarations, I've never used it before but I just noticed there's this other `module.flow.js` file, separate from the `module.d.ts`. I just updated the `.d.ts`, do you think I should go ahead and update `module.flow.js`?

mvines:
I'm not sure if anybody is actually using the flow types anymore actually.  It wouldn't hurt but not sure anybody would care either way :)

danmt:
I just tested 0.1.8, we're good to go!

danmt:
The PR for the flow-typed thing is up too. Because, why not?

Issue #2266, opened 2021-08-11
stake-pool: Consider hard-coding `spl_token::id` instead of passing in token program


joncinque:
#### Problem

There are no guarantees that a custom token program will behave exactly the same way as SPL token, so it could have sketchy logic which obviates all of the initialize checks.

#### Solution

Consider hard-coding `spl_token::id()` everywhere

joncinque:
haha not quite, the idea is to avoid having `token_program_id` set on the stake pool, and just always use `spl_token::id()`. this is just a consideration though.  the only part that actually needs to be changed is the one references to `spl_token::id()` in the instruction code

jon-chuang:
Hmm yeah, but then changing would require a redeployment of the program. 

What's this `spl_token_v2` though? Is my question. Isn't it an upgradeable version of the token program?

billythedummy:
should we get rid of the `token_program_id` field on the stake pool state struct then? Though this change will not be backward compatible. 

> haha not quite, the idea is to avoid having `token_program_id` set on the stake pool, and just always use `spl_token::id()`. this is just a consideration though. the only part that actually needs to be changed is the one references to `spl_token::id()` in the instruction code

Also, sorry but I don't quite get "the only part that actually needs to be changed is the one references to `spl_token::id()` in the instruction code". What does "instruction code" refer to, the process_*instruction* functions?

joncinque:
I literally just meant this line: https://github.com/solana-labs/solana-program-library/blob/7d748fc507162cddb158ac18c72d4484477b0357/stake-pool/program/src/instruction.rs#L780 -- it's the only line that requires `spl_token::id()`, every other part of the program (I believe) would work with any token program, not just the official SPL token program

PR #2286, opened 2021-08-16
Use stake pool's declared token ID rather than hard-coded

jon-chuang:
If a stake pool wants to set up a non `spl_token_program::id` pool, they can do so, but their users will not be minting SPL tokens, (unless via a complex scheme where the alternative token program mints spl tokens in exchange for its tokens):

```rust
// on initialize:
if pool_mint_info.owner != token_program_info.key {
}
```

In my mind, there is nothing strictly wrong with exposing the extra flexibility of a parametrised token program. It is a little redundant though.

If still wanting to make it hard coded, I don't think this is an urgent issue per se. The current logic works, and is probably correct, and tearing it out could introduce errors, for little benefit.

The purpose of this PR is to use a fully non hard-coded token program id

Fixes:  #2266

### DIFF
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -768,6 +768,7 @@ pub fn update_stake_pool_balance(
     reserve_stake: &Pubkey,
     manager_fee_account: &Pubkey,
     stake_pool_mint: &Pubkey,
+    token_program_id: &Pubkey,
 ) -> Instruction {
     let accounts = vec![
         AccountMeta::new(*stake_pool, false),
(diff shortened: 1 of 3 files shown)


joncinque:
Looks good!

Note that it won't be possible to actually use a custom token program, mainly because the instruction creator for spl_token checks the program id: https://github.com/solana-labs/solana-program-library/blob/e168a5db92f4048879755fae2bc283bb7a3ec9bc/token/program/src/instruction.rs#L1019

Eventually, when we support more than one token id, we'll have to update the instruction creators separately, requiring a recompile / redeploy of the stake pool program, but we can take care of that once it's needed.  With this, we have the flexibility in the future

PR #2288, opened 2021-08-16
stake-pool: `UpdateValidatorListBalance` must be the sole instruction in the transaction

jon-chuang:
To ensure that merged transient accounts cannot be hijacked by secondary instructions, we enforce that the `UpdateValidatorListBalance` is the sole instruction in the transaction. The following suffices:
1. checking `load_instruction_at(1, instructions_info.data).is_err()` and that 
2. the instruction is called as the top-level invocation (since no self-reentrancy exists in the program, checking that `load_instruction_at(0, instructions_info.data)?.program_id == crate::id()` suffices).


joncinque:
After an internal discussion, we think the best way to fix this is through adding the seed in the validator list.  If you want to deploy a version of the program with this change, feel free, but I can do the work to add the seed

jon-chuang:
There's another reason (apart from complexity and backwards complexity) why I think that could be a bad idea - as described earlier, we'd like to have multiple transient stake accounts. If a single transient stake account is hijacked, then all the rest need to be  to effect a seed change. Thus, an account hijack still can cause massive disruption, especially if it can occur on a regular basis.

Furthermore, having the seed in the validator list doesn't solve the actual problem - that of needing accounts to be closed to prevent stealing stake. Which is arguably the most pressing concern.

I don't see why this PR does not solve the issue satisfactorily. Would you mind sharing?

joncinque:
If you're thinking of having multiple transient stake accounts, your best bet is to have a start and end seed of a `u64` each, and never reuse a seed ever.  That way, no problems if a transient account is ever hijacked, since the address won't ever be reused.

An approach like this limits composability and reusability with other programs.

jon-chuang:
Hmm, yes, but this particular instruction is a crank - meaning that there is no reason to call it in conjunction with other instructions or as a CPI.

Furthermore, I believe the intention is to remove this check once the runtime permits cleanly closing accounts (by resizing to 0 and handing over to system program) and the stake program is updated so that it closes accounts upon merge.

So are you saying that the new validator list seed fields should be `start_u64` and `end_u64`?

I mean, the stake stealing attack can be easily mitigated by checking that the withdrawal and staking authority of the transient accounts are indeed the pool's PDA `stake_withdrawal` pubkey.

joncinque:
You're correct on all of that.  We don't know how this will be used later, so we may as well fix it correctly in the stake pool program first.

> Furthermore, I believe the intention is to remove this check once the runtime permits cleanly closing accounts (by resizing to 0 and handing over to system program) and the stake program is updated so that it closes accounts upon merge.

Absolutely, this fix will probably take some time to spec and roll out since there's already been a lot of disagreement about the best way to do this.

> So are you saying that the new validator list seed fields should be start: u64 and end: u64? I actually have a problem with that - if you stretch the start and end long enough, i.e. by creating arbitrarily many transient accounts, then you can force running into compute budget limits for update_validator_list_balance

If you create too many transient accounts, you should be able to call `update_validator_list_balance` multiple times for the same validator, but with different sets of transient accounts, to work within the compute limit.

>  I mean, the stake stealing attack can be easily mitigated by checking that the withdrawal and staking authority of the transient accounts are indeed the pool's PDA stake_withdrawal pubkey.

Absolutely, this will be part of any multi-transient account solution.

> Anyway, let me know what you think regarding this being a temporary fix, if you can think of a valid reason to call update_validator_list_balance in conjunction with other instructions or as a CPI, I'll gladly relent.

Going with the above point, we know that the pool needs to check the withdraw key of the transient account, but what do we do if the withdraw key on that account is wrong?  We have to be able to create a transient stake account with a different key.  By setting the seed, we move towards something desired anyway, and mitigate the attack

jon-chuang:
Hmm, how do you propose we maintain backwards compatibility of the validator list, then? Will we go with having `start_seed: u64, end_seed: u64`?

> If you create too many transient accounts, you should be able to call update_validator_list_balance multiple times for the same validator, but with different sets of transient accounts, to work within the compute limit.

I believe that this either requires storing an additional field - the `updated_transients_pointer`. I'm a bit worried if you think the direction is to maintain arbitrarily many transient stake accounts. Especially due to the current lingering observed credits merge issue.

joncinque:
> Will we go with adding start_seed: u64, end_seed: u64?

Yes, we'll have to break backwards compatibility for this unfortunately. I'll figure something out for the testnet stake pools.  Thankfully, we're only at X

> I'm a bit worried if you think the direction is to maintain arbitrarily many transient stake accounts.

We can set a limit to start if we're worried, but easiest to me would be to just have the merges start with the `start_seed`, do as many as possible within compute limit, and then update `start_seed`. Then we could potentially do arbitrarily many.  There shouldn't be a need for that, but who knows.  It might get closer to permissionless rebalancing :smile: 

> Especially due to the current lingering observed credits merge issue.

Do you mean the issue fixed in https://github.com/solana-labs/solana/pull/18985 or something else?

jon-chuang:
> We can set a limit to start if we're worried, but easiest to me would be to just have the merges start with the start_seed, do as many as possible within compute limit, and then update start_seed. Then we could potentially do arbitrarily many. There shouldn't be a need for that, but who knows. It might get us closer to permissionless rebalancing smile

Hmm, I think this means that updating cannot happen unless merging can happen, doesn't it? And what happens if activation time exceeds a single epoch? I'm not quite in favour of merging being a prerequisite to updating.

> Do you mean the issue fixed in [solana-labs/solana#18985](https://github.com/solana-labs/solana/pull/18985) or something else?
ooh. Could we remove the 1 lamport requirement or significantly lower it then, dya think?

joncinque:
> I'm not quite in favour of merging being a prerequisite to updating.

I'm with you, and that's why the `no_merge` flag exists on the update calls.  That, and it made some testing easier.

> Could we remove the 1 lamport requirement or significantly lower it then, dya think?

Yep! I think we can go with any value once the feature is live on mainnet.  I haven't run any tests with stake pools just yet, but feel free to give it a shot and let me know what works :grin:

jon-chuang:
Hmm, but then, even without merging, the update validator list needs to walk all the transients in use to know what their current lamport values are, is that not correct? 

Either this walk is done in a single transaction, or it is done in multiple transactions. With the new transaction format, I suppose one could walk in excess of 200 transients in a single tx.

If you want more, then you have to keep track in a struct variable which transients have already been accounted for in that particular walk. I'm much more partial to this being stateless since the stateful update probably requires locking some part of the pool's functions (maybe just for that validator's transients) to be accurate.

I suppose we might need to issue a new instruction to remove excess lamports once the feature becomes active, though. For now I'll set it to 0.1 lamports on our deployment s

joncinque:
> Hmm, but then, even without merging, the update validator list needs to walk all the transients in use to know what their current lamport values are, is that not correct?

Ah yes, you're right, we'll definitely need to have a limit for that.

> I suppose we might need to issue a new instruction to remove excess lamports once the feature becomes active, though. For now I'll set it to 0.1 lamports on our deployment

That should be totally reasonable

joncinque:
Closing this since we went with a different approach in https://github.com/solana-labs/solana-program-library/pull/2304

PR #2289, opened 2021-08-16
chore(token): add createTransferCheckedInstruction to flow-typed

danmt:
This PR intention is to add the `createTransferCheckedInstruction` static method from the Token class in the flow-typed definitions for users still using it.


mvines:
thanks again!

danmt:
Any time!

PR #2304, opened 2021-08-20
stake-pool: Add seed for transient stake account

joncinque:
#### Problem

Stake pools don't use a configurable seed for transient stake accounts.

#### Solution

Use a configurable seed!

cc @jon-chuang

### DIFF
--- a/stake-pool/program/src/lib.rs
+++ b/stake-pool/program/src/lib.rs
@@ -25,6 +25,9 @@ const AUTHORITY_DEPOSIT: &[u8] = b"deposit";
 /// Seed for withdraw authority seed
 const AUTHORITY_WITHDRAW: &[u8] = b"withdraw";
 
+/// Seed for transient stake account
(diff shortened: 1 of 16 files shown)


stake-pool/program/src/lib.rs
```diff
@@ -25,9 +25,6 @@ const AUTHORITY_DEPOSIT: &[u8] = b"deposit";
 /// Seed for withdraw authority seed
 const AUTHORITY_WITHDRAW: &[u8] = b"withdraw";
 
-/// Seed for transient stake account
-const TRANSIENT_STAKE_SEED: &[u8] = b"transient";
```
jon-chuang, line 29: I'm not sure if removing this seed and relying solely on the u64 seed is an altogether good idea.

In general, I think it's a good practice to keep namespaces separate and identifiable, as I'd hope we'd do for the transient PDAs.

joncinque: That's fair, I'll add it back in

PR #2319, opened 2021-08-23
Governance: Chat program

SebastianBor:
#### Summary

- `GovernanceChat` - program allows to add comments (text and reactions) to governance proposals. All the chat messages are public, immutable and are stored on-chain. Only token owners who deposited the relevant governing tokens (community or council) into the realm are allowed to make comments. 
 The chat program is independent from the governance program and can be used as an optional add-on.

- `spl-governance-test-sdk` - A crate with extracted shared governance test utilities which can be used by other programs depending on the governance program.


governance-chat/program/src/processor.rs
```diff
@@ -0,0 +1,114 @@
...
+
+/// Processes PostMessage instruction
+pub fn process_post_message(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    body: MessageBody,
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+
+    let governance_program_info = next_account_info(account_info_iter)?; // 0
```
joncinque, line 50: I wonder if the associated governance program id should be hard-coded somewhere in the program.  Otherwise, people from governance A can post messages in the chat program for governance B, right? Unless this is meant to be the one chat program for all governances

governance-chat/program/src/processor.rs
```diff
@@ -0,0 +1,114 @@
...
+    let governance_info = next_account_info(account_info_iter)?; // 1
+    let proposal_info = next_account_info(account_info_iter)?; // 2
+    let token_owner_record_info = next_account_info(account_info_iter)?; // 3
+    let governance_authority_info = next_account_info(account_info_iter)?; // 4
+
+    let chat_message_info = next_account_info(account_info_iter)?; // 5
```
joncinque, line 56: nit: this isn't a big deal because you're using `create_account` later, but you may want to explicitly check here that this account is actually `Uninitialized`

governance-chat/program/src/state.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/// Chat message body
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub enum MessageBody {
+    /// Text message encoded as utf-8 string
+    Text(String),
+
+    /// Emoticon encoded using utf-8 characters
+    Reaction(String),
```
joncinque, line 27: nit: Is this meant to explain how to display this on the frontend as not a message, but just a reaction under the message? If so, can you explain that in the comment above?

joncinque (commented):
Looks great for the most part! Just a few little comments.  And as discussed earlier, could you move this under the `governance` directory?

governance-chat/program/src/processor.rs
SebastianBor: Yes you are right, in the current version the chat program is meant to be a singleton (one for all governance clones) because there is no security risk. 
However this might change in the future if we for example start using some reputation or community engagement score for governance rights and then each governance clone would need a corresponding chat clone. In that case I was thinking about putting the chat clone id on realm config which could then be validated against.

governance-chat/program/src/state.rs
SebastianBor: Done

PR #2362, opened 2021-08-31
Fix example codestyle

Akiyamka:
Use const instead of let / var without any readon


chaseeb (approved):

joncinque:
Thanks for the contribution!

PR #2364, opened 2021-08-31
spl-token-cli: Add memo support to transfer and burn commands

mvines:
Also rename `--memo` to `--with-memo` for consistency with the main solana cli


token/cli/src/main.rs
```diff
@@ -171,6 +171,13 @@ pub fn multisig_signer_arg<'a, 'b>() -> Arg<'a, 'b> {
         .help(MULTISIG_SIGNER_ARG.help)
 }
 
+pub fn memo_arg<'a, 'b>() -> Arg<'a, 'b> {
+    Arg::with_name("memo")
+        .long("with-memo")
+        .takes_value(true)
+        .help("Specify a memo string to include in the transaction.")
+}
```
CriesofCarrots, line 179: Import this from solana-clap-utils::memo?

mvines: ooo, thanks

PR #2374, opened 2021-09-05
spl-token-cli: Add bench subcommand

mvines:
Add a `spl-token bench` subcommand with some basic token benchmarking facilities.

```
SUBCOMMANDS:
    close-accounts     Close multiple token accounts used for benchmarking
    create-accounts    Create multiple token accounts for benchmarking
    deposit-into       Deposit tokens into multiple accounts
    withdraw-from      Withdraw tokens from multiple accounts
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3792,6 +3792,8 @@ dependencies = [
 name = "spl-token-cli"
 version = "2.0.14"
 dependencies = [
+ "chrono",
+ "chrono-humanize",
  "clap",
  "console 0.14.1",
  "serde",
@@ -3805,6 +3807,7 @@ dependencies = [
  "solana-logger",
  "solana-remote-wallet",
  "solana-sdk",
+ "solana-transaction-status",
  "spl-associated-token-account 1.0.3",
  "spl-memo 3.0.1",
  "spl-token 3.2.0",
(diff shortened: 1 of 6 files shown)


jstarry:
fyi there are some downstream clippy issues here: https://buildkite.com/solana-labs/solana/builds/54415#fa12ecff-eebb-41cc-9246-2639ae1762f9

I guess we'll need to setup a similar clippy check in this repo soon

mvines:
Clippy is present already.  The problems are:
1. Solana 1.8's RpcClient deprecates the fee calculator and friends in v1.7 too aggressively.   SPL is still on v1.7 but it has no way to migrated off the deprecated functionally before upgrading to v1.8.
2. The `progress_bar.set_message()` signature in 1.8 changes from 1.7, so similar to (1)

Will suppress warnings as needed via https://github.com/solana-labs/solana-program-library/pull/2376

jstarry:
Ah thanks, I was just digging into this and was getting confused what was happening here

mvines:
I've almost got it fixed here

mvines:
https://github.com/solana-labs/solana-program-library/pull/2376/commits/ff5591745fb37fd486427f9225e2d5e7bbb78231 should hopefully fix the downstream build once it lands

PR #2404, opened 2021-09-10
stake-pool: cli fix remove-vsa bug

billythedummy:
for the `remove-validator` subcommand, `instructions` vec was being overwritten to a single element vec containing the stake pool instruction. If a new stake account needs to be created, it will throw:

```
thread 'main' panicked at 'Transaction::sign failed with error KeypairPubkeyMismatch', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/solana-sdk-1.7.7/src/transaction.rs:258:13
```

since the new stake account's keypair is in signers but the create stake account instruction is not in instructions

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -455,6 +455,7 @@ fn command_vsa_remove(
             stake_pool_address,
             vote_account,
             new_authority,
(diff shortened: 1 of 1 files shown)


joncinque:
Nice catch, thanks!

PR #2424, opened 2021-09-16
(docs) Fix broken links

Konosh93:
Outdated docs still referred to JS files that have been removed or replaced already.


joncinque:
Great catch, thanks for the contribution!

PR #823, opened 2020-11-11
Add c examples

jackcmay:
Add C examples

Testing is limited to none due to lack of any meaningful full framework

PR #2427, opened 2021-09-17
fix logging example

jackcmay:
Accounts are serialized into random memory

Fixes issue brought up here: https://github.com/solana-labs/solana-program-library/pull/823


jackcmay:
@OluAgunloye C logging example issue you find fixed in the pr

PR #2435, opened 2021-09-21
stake-pool: Update docs and help messages

joncinque:
#### Problem

Stake pool docs are out of date.

#### Solution

Update them!  This mainly adds sections about:

* fees
* liquid staking
* 0.001 SOL minimum for validator accounts
* Rust docs
* easy-add validator

### DIFF
--- a/docs/src/stake-pool.md
+++ b/docs/src/stake-pool.md
@@ -52,14 +52,23 @@ providers for direct SOL deposits.
 ## Operation
 
 A stake pool manager creates a stake pool, and the staker includes validators that will
-receive delegations from the pool by adding "validator stake accounts" to the pool.
+receive delegations from the pool by adding "validator stake accounts" to the pool
+using the `add-validator` instruction. In this command, the stake pool creates
+a new stake account and delegates it to the desired validator.
 
 At this point, users can participate with deposits. They can directly deposit
-SOL into the stake pool's reserve account, to be redistributed by the staker.
+SOL into the stake pool using the `deposit-sol` instruction. Within this instruction,
+the stake pool will move SOL into the pool's reserve account, to be redistributed
+by the staker.
 
 Alternatively, users can deposit a stake account into the pool.  To do this,
 they must delegate a stake account to the one of the validators in the stake pool.
-Once it's active, the user can deposit their stake into the pool.
+If the stake pool has a preferred deposit validator, the user must delegate their
+stake to that validator's vote account.
+
+Once the stake becomes active, which happens at the following epoch boundary
+(maximum 2 days), the user can deposit their stake into the pool using the
+`deposit-stake` instruction.
 
 In exchange for their deposit (SOL or stake), the user receives SPL tokens
 representing their fractional ownership in pool. A percentage of the rewards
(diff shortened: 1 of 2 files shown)


c3pko:
Thanks for updating these docs! The new description of fees in particular is super helpful. A few suggestions/questions for you:

Line 55-60:
1. When a validator deposits SOL, does the stake pool manager create a stake account for them? And then deposit that into the stake account? (And that’s just seen as the additional risk with stake pools that take SOL deposits vs only taking fully activated stake accounts?)
2. When a user deposits SOL into a stake pool will it go directly into the reserve account?

Lines 60-62:
“Alternatively, users can deposit a stake account into the pool.  To do this, they must delegate a stake account to the one of the validators in the stake pool. Once it's active, the user can deposit their stake into the pool.”
-   Is this equivalent to delegating a stake account to a stake pool’s “preferred validator”?
- How long does a user need to wait after delegating a stake account before it’s active (and they can deposit their stake into the pool)?

Line 92: can we add a sentence here that reads something like “The user will get back their SOL stake account immediately but note it will take a 1-2 epochs for the stake account to deactivate and for the user to be able to withdraw their SOL”. Feel free to adjust with the correct language.

Line 124-125 this “Maintenance mode. If the pool needs more time to reset fees or otherwise, the manager can temporarily make the pool private” may be misinterpreted as the stake pool manager temporarily taking control of user deposits. Is there another way to describe “making the pool private”? Is it more like a temporary restriction on new deposits?

For lines 280 to 284, this is true in the current reference implementation right? This isn’t a new feature that existing stake pool (running this reference implementation) would need to upgrade to?


Line 428-429: when the validator stake account is removed, the withdraw authority doesn’t change (it still belongs to the delegator), so what specific authority in the stake account belongs to the administrator? Want to update this line “Unlike a normal withdrawal, the validator stake account is totally moved from the stake pool and into a new account belonging to the administrator” to reflect that stake pool managers don’t suddenly just get access to all delegator’s stake accounts.

joncinque:
Thanks for the review!  I've updated the docs to reflect your questions / feedback.  To answer the specific points:

> When a validator deposits SOL, does the stake pool manager create a stake account for them?

When adding a validator, the stake pool will do all the work of creating the stake account and adding it

> When a user deposits SOL into a stake pool will it go directly into the reserve account?

Yep!

> Is this equivalent to delegating a stake account to a stake pool’s “preferred validator”?

Yes, if the stake pool has a preferred validator.  I added something to clarify for that point.

> How long does a user need to wait after delegating a stake account before it’s active (and they can deposit their stake into the pool)?

Added language to clarify that, it'll soon be max one epoch once some fixes all roll out.

> Line 92: can we add a sentence here that reads something like “The user will get back their SOL stake account immediately but note it will take a 1-2 epochs for the stake account to deactivate and for the user to be able to withdraw their SOL”. Feel free to adjust with the correct language.

Done

> Line 124-125 this “Maintenance mode. If the pool needs more time to reset fees or otherwise, the manager can temporarily make the pool private” may be misinterpreted as the stake pool manager temporarily taking control of user deposits. Is there another way to describe “making the pool private”? Is it more like a temporary restriction on new deposits?

Good point, clarified that

> For lines 280 to 284, this is true in the current reference implementation right?

Yep, clarified that

> Line 428-429: when the validator stake account is removed, the withdraw authority doesn’t change (it still belongs to the delegator), so what specific authority in the stake account belongs to the administrator?

Clarified this too, and the validator stake account effectively gets moved to the staker when it's removed. The old account is no more, and the new one belongs entirely to the staker (both authorities).

PR #2450, opened 2021-09-24
Governance: voters weights add-in

SebastianBor:
#### Summary

The purpose of this change is to support voter weights provided by an external program (`addin`). This would allow to move away from a simple token deposit based weights and make it possible to provide custom strategies without the need to modify the `governance` program (open/close principle). 
The following voter weight addins are planned (not part of this PR):
- Time locked token deposits 
- Time locked token deposits with multiple locks (ex. 1mm locked for 4 years and 1k locked for 1year) 
- Token freezing instead of deposits
- Multiple tokens deposits 
- NFTs deposits 
- Reputation score 

#### Solution

- `VoterWeight`  `addin` interface was defined using `VoterWeightRecord` account. `Governance` program reads the account to get voter's weight instead of using the internal deposit pool.
- Each `addin` program manages their own `VoterWeightRecord` accounts
- The `Governance` program stores the chosen addin program id in `RealmConfig` extension account and flags it in realm config `use_community_voter_weight_addin`
- Sample `voter-weight-addin` program is provided as a template for creating custom addins 

#### Time sensitive voter weights
`VoterWeightRecord` has expiry flag for time sensitive voter weights. When the up to date voter weight needs to be calculated at the time of voting the client (UI) would call `Revise` instruction for the `addin` with all the needed accounts within the same transaction as the `governance` instruction and the `addin` would set the expiry to the current slot. Then the `governance` instruction would validate the expiry to ensure it uses up to date weight.
This way any custom logic can be supported and the `addins` might use any arbitrary set of accounts and current time to calculate the up to date voter weight. 

Alternative solution I was considering was to make a `CPI` call from the `governance` program to the `addin` program to get voter's weight, but that would need a custom set of accounts to be provided in all but trivial scenarios, and I decided the `Revise` instruction with `vote_weight_expiry` was a simpler solution and would cover any arbitrary scenario.


governance/program/src/addins/voter_weight.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// Governing Token Mint the VoterWeightRecord is associated with
+    /// Note: The addin can take deposits of any tokens and is not restricted to the community or council tokens only
+    // The mint here is to link the record to either community or council mint of the realm
+    pub governing_token_mint: Pubkey,
+
+    /// The owner of the governing token and voter
+    pub governing_token_owner: Pubkey,
+
+    /// Voter's weight
+    pub voter_weight: u64,
```
armaniferrante, line 47: Without the addin, is this number normally the same as number of "community tokens" one votes with?

SebastianBor: Yes. It’s the number of deposited tokens

governance/program/tests/process_create_mint_governance.rs
```diff
@@ -142,7 +142,7 @@ async fn test_create_mint_governance_without_transferring_mint_authority_with_au
             |i| {
```
joncinque, line 142: GH doesn't let me comment on `governance/program/tests/fixtures/spl_governance_voter_weight_addin.so`, so I'll do it here.  Right now, if you make a change to the addin, you need to build it and copy the so into here each time, right? Since the governance program and add-in are both in the same `BPF_OUT_DIR`, program-test should find them. You can have a nice error message saying `please run cargo build-bpf in governance/voter-weight-addin/program` if it isn't found

governance/program/src/instruction.rs
```diff
@@ -393,11 +414,26 @@ pub enum GovernanceInstruction {
...
+    ///   6. `[writable]` RealmConfig account. PDA seeds: ['realm-config', realm]
+    ///   7. `[]` Optional Community Voter Weight Addin Program Id    
     SetRealmConfig {
         #[allow(dead_code)]
         /// Realm config args
         config_args: RealmConfigArgs,
     },
+
+    /// Creates TokenOwnerRecord with 0 deposit amount
+    /// It's used to register TokenOwner when voter weight addin is used and the Governance program doesn't take deposits
```
joncinque, line 428: It seems like the "addin" could be better seen as a vault!  In the end, it will hold all of the tokens, correct?

governance/program/src/state/realm_config.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::{
+    error::GovernanceError,
+    state::enums::GovernanceAccountType,
+    tools::account::{get_account_data, AccountMaxSize},
+};
+
+/// RealmConfig account
+/// The account is an optional extension to RealmConfig stored on Realm account
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub struct RealmConfigAccount {
```
joncinque, line 19: nit: it's not totally clear why this needs to be in a separate account.

governance/program/src/state/realm_config.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub realm: Pubkey,
+
+    /// Addin providing voter weights for community token
+    pub community_voter_weight: Option<Pubkey>,
+
+    /// Reserved for community max vote weight addin
+    pub reserved_1: Option<Pubkey>,
+
+    /// Reserved for council voter weight addin
+    pub reserved_2: Option<Pubkey>,
```
joncinque, line 33: should this be called `council_voter_weight_addin` then?

governance/program/src/state/realm_config.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub struct RealmConfigAccount {
+    /// Governance account type
+    pub account_type: GovernanceAccountType,
+
+    /// The realm the config belong to
+    pub realm: Pubkey,
+
+    /// Addin providing voter weights for community token
+    pub community_voter_weight: Option<Pubkey>,
```
joncinque, line 27: nit: can we call this `community_voter_weight_addin`?

governance/program/src/state/realm_config.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Reserved for community max vote weight addin
+    pub reserved_1: Option<Pubkey>,
```
joncinque, line 30: What is this reserved for?

governance/voter-weight-addin/program/src/processor.rs
```diff
@@ -0,0 +1,84 @@
...
+        .map_err(|_| ProgramError::InvalidInstructionData)?;
+
+    msg!("GOVERNANCE-VOTER-WEIGHT-INSTRUCTION: {:?}", instruction);
+
+    match instruction {
+        VoterWeightAddinInstruction::Revise {} => Ok(()),
+        VoterWeightAddinInstruction::Deposit { amount } => {
+            process_deposit(program_id, accounts, amount)
+        }
+        VoterWeightAddinInstruction::Withdraw {} => Ok(()),
```
joncinque, line 37: I'm guessing this is also WIP since there's no token movements or checks?

joncinque (commented):
It took me awhile to understand it all, but I think the approach is sound.  My main question is about the provided skeleton program, and whether it'll actually move tokens.  A lot of the logic in `DepositGoverningTokens` relies on governance info, so you'll have to be careful duplicating those parts.  There's also the problem with topups / relinquishes, and making sure that it's reflected in vote records.

I wonder if there's some nice refactoring that can be done for the deposit / withdraw logic, so that the governance program reuses bits of the basic addin.

SebastianBor:
> It took me awhile to understand it all, but I think the approach is sound. My main question is about the provided skeleton program, and whether it'll actually move tokens. A lot of the logic in `DepositGoverningTokens` relies on governance info, so you'll have to be careful duplicating those parts. There's also the problem with topups / relinquishes, and making sure that it's reflected in vote records.
> 
> I wonder if there's some nice refactoring that can be done for the deposit / withdraw logic, so that the governance program reuses bits of the basic addin.

Your observations are correct, the provided add-in skeleton is not a full working example yet. For this PR I implemented only the bare minimum to make it possible to use it in governance tests. I'm planning to do the full end to end implementation in the next PR.

And yes, it's a good idea to make the common code to create add-ins reusable. Since the add-in has dependency on the governance program (and knowledge about it) I think it could just reference the `spl-governance` crate, but we could also create a crate with just the reusable utilities. I'm not sure which one would be better yet. 

Note, when the `add-in` is used it doesn't eliminate need for `TokenOwnerRecord`. It's still used to track realm members, their active proposals, active votes and allows to use governance delegates. And it's still the `governance` program responsibility to manage all of that. This is why `CreateTokenOwnerRecord` instruction was created as part of this PR to make it possible to create the account without depositing any tokens.

governance/program/src/processor/process_create_mint_governance.rs
```diff
@@ -48,13 +48,24 @@ pub fn process_create_mint_governance(
...
+    let governance_authority_info = next_account_info(account_info_iter)?; // 9
+
     assert_valid_create_governance_args(program_id, &config, realm_info)?;
 
     let realm_data = get_realm_data(program_id, realm_info)?;
     let token_owner_record_data =
         get_token_owner_record_data_for_realm(program_id, token_owner_record_info, realm_info.key)?;
 
-    token_owner_record_data.assert_can_create_governance(&realm_data)?;
+    token_owner_record_data.assert_token_owner_or_delegate_is_signer(governance_authority_info)?;
```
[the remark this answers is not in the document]
SebastianBor: Yes, I realised these checks were missing while doing the addin changes

governance/program/src/instruction.rs
```diff
@@ -393,11 +414,26 @@ pub enum GovernanceInstruction {
     ///       If that's required then it must be done before executing this instruction
     ///   3. `[writable]` Council Token Holding account - optional unless council is used. PDA seeds: ['governance',realm,council_mint]
     ///       The account will be created with the Realm PDA as its owner
+    ///   4. `[signer]` Payer
+    ///   5. `[]` System
+    ///   6. `[writable]` RealmConfig account. PDA seeds: ['realm-config', realm]
```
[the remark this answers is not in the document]
SebastianBor: No, it's not required. The new `RealmConfigAccount` is only needed when the `addin` is used. If an existing ream decides to use the `addin` then they would run `SetRealmConfig` which would in turn create the account.
Even after this change `RealmConfigAccount` would be optional and only crated for new realms if required.

governance/voter-weight-addin/program/src/instruction.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    /// Deposits governing token
+    /// 0. `[]` Governance Program Id
+    /// 1. `[]` Realm account
+    /// 2. `[]` Governing Token mint
+    /// 3. `[]` TokenOwnerRecord
+    /// 4. `[writable]` VoterWeightRecord
+    /// 5. `[signer]` Payer
+    /// 6. `[]` System
+    Deposit {
```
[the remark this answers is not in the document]
SebastianBor: Yes, it should handle governance token movements. The `addin` in this PR is just a mock used for governance testing and I would provide full implementation in the follow up PR.

governance/voter-weight-addin/program/src/processor.rs
```diff
@@ -0,0 +1,84 @@
...
+    let system_info = next_account_info(account_info_iter)?; // 6
+
+    let token_owner_record_data = get_token_owner_record_data_for_realm_and_governing_mint(
+        governance_program_info.key,
+        token_owner_record_info,
+        realm_info.key,
+        governing_token_mint_info.key,
+    )?;
+
+    // TODO: Custom deposit logic and validation goes here
```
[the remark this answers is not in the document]
SebastianBor: Yes, it's still WIP.  Users asked for an option to specify the deposit amount and I added it for the governance program deposits and would also support it in the `addin`.  
Note: The amount is not taken from TokenOwnerRecord, it'll be deposited into the `addin`

governance/program/src/state/realm_config.rs
SebastianBor: It only exist because I wanted to make it backward compatible and couldn't resize existing `Realm` accounts. 
Ideally we shouldn't need this and it could be added to `RealmConfig` stored on the Realm account but we need support to resize accounts on mainnet runtime first. 
There is also some small benefit for keeping it separate. If you don't use the advanced realm features you don't have to pay the extra rent, but I'm not sure if that alone would justify it.

governance/program/src/state/realm_config.rs
SebastianBor: Done, much better this way. 
The filed name was a leftover from the time when the account was called RealmAddins and the suffix was redundant :)

governance/program/src/state/realm_config.rs
SebastianBor: The `add-in` would allow to provide `max voter weight `externally. Currently the `governance` program uses the governing token mint supply as the max voter weight but there are scenarios when this is not sufficient:

- When `voter weight add-in `is used it's common that the max voter weight is also affected. For example the `voter weight add-in` might add scaling factors based on the token lock period, which it turns changes the max possible weight. 
Note: In this scenario the `voter weight addin` and `max voter weight addin` could be both implemented by the same program.
- When the governance token is fully minted but the full supply is not in circulation yet. In this case the `max voter weight addin` would provide the actual circulating supply.

governance/program/src/state/realm_config.rs
SebastianBor: Ok, I gave the `addins` concrete names. 
Yes, the common case for council add-in is to have a council without using tokens and with expiry date. This way council members can be removed by the community and they can also be selected periodically through an election process. 
With tokens based votes the members can only voluntarily burn their tokens

governance/program/src/state/realm_config.rs
SebastianBor: It can be the same program but it's not necessary. You can also have just one of the `addins`.
For example the common use case for just `max vote weight addin` is a governance token which is fully minted but not circulating and the locked none distributed supply is stored in a treasury account. In this case you can create a very simple reusable addin which can provide:  `max_vote_weight` = `mint_supply` - `treasury_amount`

SebastianBor:
> Sorry for the late re-review -- most of the points are cosmetic, and the overall concept behind this is sound. Do add the build.rs file to remove that *.so in the repo please!

Thank you, I've addressed the issues in a follow up PR https://github.com/solana-labs/solana-program-library/pull/2512

PR #2453, opened 2021-09-26
stake-pool: Use `matches!` to allow for fix in underlying error

joncinque:
#### Problem

`Unkown` is not how you spell `Unknown`, but a stake pool test relies on that misspelling, making it impossible to update the spelling in the monorepo.

#### Solution

use a `matches!` check in the test instead


mvines (approved):
Thaks // sic

PR #2463, opened 2021-09-28
math: Add f64 powf and powi

joncinque:
#### Problem

`powf` finally works in bpf!

#### Solution

To make sure it keeps working, here's a test calling `powi` and `powf` on floats

### DIFF
--- a/libraries/math/src/processor.rs
+++ b/libraries/math/src/processor.rs
@@ -148,7 +148,10 @@ pub fn process_instruction(
         MathInstruction::F64Pow { base, exponent } => {
             msg!("Calculating f64 Pow");
             sol_log_compute_units();
-            let _result = base.powi(5);
+            let result = base.powi(exponent as i32);
+            sol_log_compute_units();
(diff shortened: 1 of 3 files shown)


joncinque:
It might be worth adding this test in if `powf` is now supported

libraries/math/src/processor.rs
```diff
@@ -145,6 +145,15 @@ pub fn process_instruction(
             msg!("{}", result as u64);
             Ok(())
         }
+        MathInstruction::F64Pow { base, exponent } => {
+            msg!("Calculating f64 Pow");
+            sol_log_compute_units();
+            let _result = base.powi(5);
+            let result = base.powf(exponent);
```
buffalojoec, line 152: Is this processor just for testing purposes? I'm curious why you're just sticking 5 in there.

From what I can see in other uses of `spl-math` it's not actually a deployed program?

buffalojoec (dismissed):

buffalojoec (approved):

PR #2473, opened 2021-09-30
Stateless offer

aeyakovenko:
# Problem

There is no simple way to create one-off offers for token transfers

# Solution

Maker uses PDA as a hash of the offer constraints and approves the PDA as a token delegation.  The taker then needs to satisfy the constraints by delegating what the maker is asking for to the PDA, which then does the swap.

### DIFF
--- a/stateless-asks/program/src/instruction.rs
+++ b/stateless-asks/program/src/instruction.rs
@@ -1,0 +1,66 @@
+//! Instruction types
+
+#![allow(clippy::too_many_arguments)]
+use {
+    borsh::{BorshDeserialize, BorshSchema, BorshSerialize},
+    solana_program::{
+        instruction::{AccountMeta, Instruction},
+        pubkey::Pubkey,
+        system_program, sysvar,
+    },
+};
+
+/// Instructions supported by the StatelessOffer program.
+#[repr(C)]
+#[derive(Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize, BorshSchema)]
+pub enum StatelessOfferInstruction {
+    ///  Accept a StatelessOffer
+    ///
+    AcceptOffer {
+        maker_size: u64,
+        taker_size: u64,
+        bump_seed: u8,
+    },
+}
+
+/// Creates an 'initialize' instruction.
+pub fn accept_offer(
+    program_id: &Pubkey,
+    maker_src_account: &Pubkey,
+    maker_dst_account: &Pubkey,
+    taker_src_account: &Pubkey,
+    taker_dst_account: &Pubkey,
+    auth_maker_account: &Pubkey,
+    auth_taker_account: &Pubkey,
+    authority: &Pubkey,
+    token_program_id: &Pubkey,
+    bump_seed: u8,
+) -> Instruction {
+    let init_data = StatelessOfferInstruction::AcceptOffer {
+        maker_size,
+        taker_size,
+        bump_seed,
+    };
+    let data = init_data.try_to_vec().unwrap();
+    let mut accounts = vec![
+        AccountMeta::new(*program_id, false),
+
+        AccountMeta::new(*maker_src_account, false),
+        AccountMeta::new(*maker_dst_account, false),
+
+
+        AccountMeta::new(*taker_src_account, false),
+        AccountMeta::new(*taker_dst_account, false),
+
+        AccountMeta::new(*auth_maker_account, false),
+        AccountMeta::new(*auth_taker_account, false),
+
+        AccountMeta::new_readonly(*token_program_id, false),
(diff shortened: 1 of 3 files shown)


jarry-xiao:
This is a cool idea. I briefly read through the slack thread, but I think I now kinda get how you anticipate this working after reading through your code. This is what I think the preprocessing should involve. Please correct me if I'm wrong:

We will follow the steps of Alice (maker) and Bob (taker)
1. Alice has some amount Token A in `mkr_src_account`
2. Alice calls `Approve` on account to some `transfer_authority` owned by this program.
3. Alice initializes an `auth_maker_account` intermediate account and calls `SetAuthority` to the `transfer_authority`

Some time later:

4. Bob initializes `tkr_src_account` (Token B), `tkr_dst_account` (Token A), `mkr_dst_account` (Token B) and calls `Approve` on `tkr_src_account`
5. Bob creates `auth_taker_account` and calls `SetAuthority` to the `transfer_authority`
6. Bob (or anyone) permissionlessly executes `AcceptOffer`

Does this seem right or am I totally missing the point?

I can foreseeably see this as a system as a way to distribute "order books" in a way never done before. (After all a bid is just and ask with the tokens reversed). You're an absolute madman/genius, but I do have some comments about how to change this structure of this program.

stateless-asks/program/src/processor.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        let taker_src_account = next_account_info(account_info_iter)?;
+        let taker_dst_account = next_account_info(account_info_iter)?;
+
+        let auth_maker_account = next_account_info(account_info_iter)?;
+        let auth_taker_account = next_account_info(account_info_iter)?;
+
+        let token_program_info = next_account_info(account_info_iter)?;
+
+        let transfer_authority = next_account_info(account_info_iter)?;
```
jarry-xiao, line 73: How would a client program go about finding these PDA's? The completely stateless idea is cool, but it becomes difficult to discover any of these accounts right? You might be able to see who the delegate is on a token account, but even then, how would you know the seeds?

stateless-asks/program/src/processor.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        let seeds: &[&[_]] = &[
+            &maker_src_account.key.to_bytes(),
+            &maker_dst_account.key.to_bytes(),
+            &taker_src_mint.key.to_bytes(),
```
jarry-xiao, line 78: This account needs to be decoded

stateless-asks/program/src/processor.rs
```diff
@@ -0,0 +1,146 @@
...
+            token_program_info.clone(),
+            maker_src_account.clone(),
+            auth_maker_account.clone(),
+            transfer_authority.clone(),
+            maker_size,
+            seeds,
+        )?;
+        msg!("done tx from maker to temp {}", maker_size);
+
+        Self::token_transfer(
```
jarry-xiao, line 94: I'm not quite following why this intermediate account is necessary. Is there some exploit that you can do if it's just a direct transfer?

stateless-asks/program/src/instruction.rs
```diff
@@ -0,0 +1,66 @@
...
+    bump_seed: u8,
+) -> Instruction {
+    let init_data = StatelessOfferInstruction::AcceptOffer {
+        maker_size,
+        taker_size,
+        bump_seed,
+    };
+    let data = init_data.try_to_vec().unwrap();
+    let mut accounts = vec![
+        AccountMeta::new(*program_id, false),
```
jarry-xiao, line 46: I don't think you need this one

stateless-asks/program/src/instruction.rs
```diff
@@ -0,0 +1,66 @@
...
+
+
+        AccountMeta::new(*taker_src_account, false),
+        AccountMeta::new(*taker_dst_account, false),
+
+        AccountMeta::new(*auth_maker_account, false),
+        AccountMeta::new(*auth_taker_account, false),
+
+        AccountMeta::new_readonly(*token_program_id, false),
+        AccountMeta::new_readonly(*authority, false),
```
jarry-xiao, line 59: Order of the last 2 are reversed

jarry-xiao:
https://github.com/solana-labs/solana-program-library/pull/2476

migrating to here

Kuddah:
i there away, to stop the maker from moving the token to another wallet.

aeyakovenko:
@Kuddah that's just a cancel. These aren't put in a book, you could build a book externally to this that takes custody under its own PDA, and manages a price time queue, but the complexity there is the same a serum.  This is purely to post easy links for p2p offers.

aeyakovenko:
continued by #2476

PR #2476, opened 2021-10-01
Building on top of Toly's PR, v1 compiles

jarry-xiao:
@aeyakovenko Took your code and put it into a new PR. If I have time tomorrow I can look into also building out a really simple JS client. Made a few structural changes/bug fixes


jarry-xiao:
I think it's probably okay for the taker to sign this transaction. The PDA would still be delegated `maker_size` to transfer and the taker won't be required to call `Approve` on their source account. This seems like better UX

aeyakovenko (approved):

mvines:
Consider using a better commit name next time: https://github.com/solana-labs/solana-program-library/commit/290833e76daca5eeaa741c0b2f175e285057a200

"Building on top of Toly's PR, v1 compiles " means nothing from a `git log` perspective

PR #2484, opened 2021-10-05
[token][js]: add missing checked instruction method declarations

kklas:
... to `module.d.ts`

The following methods:
- createApproveCheckedInstruction
- createMintToCheckedInstruction
- createBurnCheckedInstruction
- createSyncNativeInstruction

are missing from the `module.d.ts` file which means they won't be included in
the `lib/index.d.ts` module declaration file and therefore cannot be
referenced to when using the package with typescript.

Added the missing declarations.


token/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@solana/spl-token",
-  "version": "0.1.8",
+  "version": "0.1.9",
```
joncinque, line 3: We'll take care of bumping the version prior to deployment, so there's no need to do that here.

joncinque (commented):
Thanks for the contribution, looks good! Once you take care of the little nit on the version, this should be good to go

PR #2507, opened 2021-10-13
Updating Stateless Offer PDA seeds to use wallet and mints

jarry-xiao:
- This is a small change from the previous merge. Because this contract assumes that the users are using ATA's it makes sense for the seeds to only requires the wallet and mints.

### DIFF
--- a/stateless-asks/program/src/processor.rs
+++ b/stateless-asks/program/src/processor.rs
@@ -76,8 +76,8 @@ fn process_accept_offer(
     msg!("Delegated Amount matches");
     let seeds = &[
         b"stateless_offer",
-        maker_src_account.key.as_ref(),
-        maker_dst_account.key.as_ref(),
+        maker_wallet.key.as_ref(),
+        maker_src_mint.key.as_ref(),
         taker_src_mint.key.as_ref(),
         &maker_size.to_le_bytes(),
         &taker_size.to_le_bytes(),


stateless-asks/program/src/processor.rs
```diff
@@ -167,7 +228,128 @@ fn process_accept_offer(
...
+    is_native: bool,
+    seeds: &[&[u8]],
+) -> Result<u64, ProgramError> {
+    let metadata = Metadata::from_account_info(metadata_info)?;
+    let fees = metadata.data.seller_fee_basis_points;
+    let total_fee = (fees as u64)
+        .checked_mul(size)
+        .ok_or(UtilError::NumericalOverflow)?
+        .checked_div(10000)
+        .ok_or(UtilError::NumericalOverflow)?;
```
joncinque, line 255: nit, and not your fault: seems like the metadata struct should do this for you!

stateless-asks/program/src/processor.rs
```diff
@@ -167,7 +228,128 @@ fn process_accept_offer(
...
         )?;
     }
-    msg!("done tx from taker to maker {}", taker_size);
+    msg!("done tx from taker to maker {}", taker_pay_size);
     msg!("done!");
     Ok(())
 }
+
+#[allow(clippy::too_many_arguments)]
+fn pay_creator_fees<'a>(
```
joncinque, line 237: Seems like this would be a really useful utility for any program using metaplex nfts!

stateless-asks/program/src/processor.rs
```diff
@@ -167,7 +228,128 @@ fn process_accept_offer(
...
+                            )?;
+                        }
+                    }
+                } else if creator_fee > 0 {
+                    if !seeds.is_empty() {
+                        msg!("Maker cannot pay with native SOL");
+                        return Err(ProgramError::InvalidAccountData);
+                    }
+                    match system_program_info {
+                        Some(sys_program_info) => {
```
joncinque, line 325: nit: this could be an `if let...` statement

stateless-asks/program/src/processor.rs
```diff
@@ -167,7 +228,128 @@ fn process_accept_offer(
...
+        .checked_mul(size)
+        .ok_or(UtilError::NumericalOverflow)?
+        .checked_div(10000)
+        .ok_or(UtilError::NumericalOverflow)?;
+    let mut remaining_fee = total_fee;
+    let remaining_size = size
+        .checked_sub(total_fee)
+        .ok_or(UtilError::NumericalOverflow)?;
+    match metadata.data.creators {
+        Some(creators) => {
```
joncinque, line 261: nit: prefer an `if let...` here

stateless-asks/program/src/instruction.rs
```diff
@@ -47,17 +51,69 @@ pub fn accept_offer(
...
+    token_program_id: &Pubkey,
+    metadata: &Pubkey,
+    creators: &[&Pubkey],
+    is_native: bool,
     maker_size: u64,
     taker_size: u64,
     bump_seed: u8,
 ) -> Instruction {
     let init_data = StatelessOfferInstruction::AcceptOffer {
+        has_metadata: true,
```
joncinque, line 110: What happens if people use the wrong instruction? For example, the maker's src token is an NFT, but the taker doesn't use `has_metadata`.  Will there be a failure on transfer or will it succeed and screw over the creators?

stateless-asks/program/src/processor.rs
```diff
@@ -122,42 +188,37 @@ fn process_accept_offer(
...
-        assert_keys_equal(system_program::id(), *system_program_info.key)?;
-        invoke(
-            &system_instruction::transfer(taker_src_account.key, maker_dst_account.key, taker_size),
-            &[
-                taker_src_account.clone(),
-                maker_dst_account.clone(),
-                system_program_info.clone(),
-            ],
-        )?;
+        match system_program_info {
```
joncinque, line 193: nit: `if let` is probably better here too

stateless-asks/program/src/processor.rs
```diff
@@ -56,9 +68,80 @@ fn process_accept_offer(
...
+                taker_wallet,
+                token_program_info,
+                system_program_info,
+                taker_src_mint,
+                taker_size,
+                is_native,
+                &[],
+            )?;
+            (maker_size, taker_remaining_size)
+        } else if *metadata_info.key == taker_metadata_key {
```
joncinque, line 120: Is there any demand to do an NFT for an NFT?

joncinque: oh really, how does that work? i thought both NFTs would operate totally independently, so any transfer of an NFT should pay out to its creators

jarry-xiao: Yeah it totally should, but it's just a function of most NFT's having a supply of 1 and decimals of 0. When one of these is traded for another one, the creator fees will always round down to 0. This is unavoidable

joncinque: Oh of course, that makes perfect sense.  thanks for the explanation!

jarry-xiao: I think an NFT for NFT feeless swap is a reasonable feature to keep in the protocol. Really no way to get around that

steveluscher:
> Because this protocol supports general SPL tokens, you cannot include metadata if you were trading something like Wrapped BTC for Wrapped USDC.

Oh, for sure. I just thought that if there's metadata, the metadata specifies a royalty, and we're not doing an NFT-for-NFT swap, then you might like this program to take the position that creator royalties are required.

PR #2512, opened 2021-10-14
Governance: use voter-weight-addin as dev-dependency

SebastianBor:


governance/program/tests/program_test/addins.rs
```diff
@@ -0,0 +1,25 @@
...
+                .args(&[
+                    "build-bpf",
+                    "--manifest-path",
+                    "../voter-weight-addin/program/Cargo.toml",
+                ])
+                .status()
+                .expect("Failed to build voter-weight-addin program")
+                .success());
+        }
+        std::mem::drop(guard);
```
joncinque, line 23: nit: This shouldn't be necessary, since the guard is dropped right after this line

governance/program/src/state/realm_config.rs
```diff
@@ -6,11 +6,10 @@ use solana_program::{
 };
 
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
+use spl_governance_tools::account::AccountMaxSize;
```
joncinque, line 9: nit: it's confusing to have `spl_governance_tools::account` and `spl_governance_program::tools::account`, would it be possible to move all of `tools::account` into `spl_governance_tools::account`?  Otherwise, there's always the thought of, "is this tools:account or spl_governance_tools::account"?  You could also rename `spl_governance_tools` to `spl_governance_account` and make it totally clear what it's for

governance/program/tests/program_test/addins.rs
SebastianBor: done

governance/program/src/error.rs
```diff
@@ -162,20 +162,6 @@ pub enum GovernanceError {
...
-    InvalidAccountOwner,
-
-    /// Account doesn't exist
-    #[error("Account doesn't exist")]
-    AccountDoesNotExist,
-
-    /// Invalid Account type
-    #[error("Invalid Account type")]
-    InvalidAccountType,
-
```
joncinque, line 178: Will removing these cause any issues for the programs already deployed?

SebastianBor: If there were any programs using the error codes then yes, but I'm not aware of any.
For the UI side I'll make a change to accommodate the breaking change.

PR #2513, opened 2021-10-14
- reimplement prepare_withdraw_accounts according to the documentation

AlexanderRay:
according to documentation and stake-pool-program process_withdraw_stake instruction the withdraw order is: 
 withdraw from preferred validator stake account (if set) or withdraw via an automatic routine in the following order
 - validator stake account
 - transient stake accounts 
 - reserve stake account
 
but the sub-routine prepare_withdraw_accounts in command_withdraw just sort accounts according to their balance 

https://github.com/solana-labs/solana-program-library/blob/53c94ed13d0098ca835672d1831bcd242f50bb2f/stake-pool/cli/src/main.rs#L1093 

which results in program failures, cause the program assumes another order. 

@joncinque

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -1,7 +1,5 @@
 mod client;
 
-use spl_stake_pool::state::ValidatorStakeInfo;
-use std::cmp::Ordering;
 use {
     crate::client::*,
     clap::{
@@ -33,6 +31,7 @@ use {
         transaction::Transaction,
     },
     spl_associated_token_account::{create_associated_token_account, get_associated_token_address},
+    spl_stake_pool::state::ValidatorStakeInfo,
     spl_stake_pool::{
         self, find_stake_program_address, find_transient_stake_program_address,
         find_withdraw_authority_program_address,
@@ -41,6 +40,7 @@ use {
         state::{Fee, FeeType, StakePool, ValidatorList},
         MINIMUM_ACTIVE_STAKE,
     },
+    std::cmp::Ordering,
     std::{process::exit, sync::Arc},
 };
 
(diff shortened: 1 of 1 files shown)


stake-pool/cli/src/main.rs
```diff
@@ -1064,28 +1064,75 @@ struct WithdrawAccount {
...
+            );
+            (
+                stake_account_address,
+                validator.active_stake_lamports,
+                Some(validator.vote_account_address),
+            )
+        })
+        .collect::<Vec<_>>();
+
+    sorted_active_stake_accounts.sort_by(|a, b| b.1.cmp(&a.1));
```
joncinque, line 1096: After the sort, can you move the element for the preferred validator in the first position? Then we'll be totally aligned with the logic as described

joncinque (commented):
Looks really good! This logic has bothered me for awhile, so I'm glad that you're doing it much better.  Just one little thing to add and this should be good to go

stake-pool/cli/src/main.rs
```diff
@@ -1061,31 +1063,93 @@ struct WithdrawAccount {
     pool_amount: u64,
 }
 
+fn sorted_accounts<F>(
+    validator_list: &ValidatorList,
+    stake_pool: &StakePool,
+    get_account: F,
```
joncinque, line 1069: nit: can we call this `get_pubkey`? `get_account` makes it sound like it's actually fetching an account

stake-pool/cli/src/main.rs
```diff
@@ -1,5 +1,7 @@
 mod client;
 
+use spl_stake_pool::state::ValidatorStakeInfo;
+use std::cmp::Ordering;
```
joncinque, line 4: style nit: can you include these in the big `use` statement right underneath?

joncinque:
Thanks so much for the work!

PR #2539, opened 2021-10-23
Rewrite spl-token-js in TypeScript

jordansexton:
Weekend project. The `Token` class has been removed ([Discord chat for context](https://discord.com/channels/428295358100013066/774014770402689065/900056803516239903)). Instructions, state de/serialization, and actions have all been restructured into a simplified, consistent, functional API.

```shell
yarn install
yarn build
yarn docs
yarn test
```

Amounts now use JS native `bigint` instead of `BN.js`. Program IDs have been moved to the end of function arguments and given default values. Errors have been exposed, and the code has been thoroughly refactored to DRY it up. Everything is split up into ES modules to support tree shaking and make the code easier to follow and test.

TypeScript documentation is included, and I plan to add tests and usage docs after initial review. There are `@TODO`s to add a few instructions that aren't implemented in the current version (InitializeAccount2, InitializeAccount3, InitializeMultisig2, InitializeMint2).

This depends on two things that ~~aren't published yet~~  --
- https://github.com/solana-labs/buffer-layout/pull/1 (on @jstarry's excellent TypeScript fork of `buffer-layout`)
- https://github.com/solana-labs/buffer-layout-utils (for PublicKey, u64, and bool de/serialization)

Edit: the tests and example have been ported from the current implementation and are working. Worth noting that the current tests only cover a tiny fraction of the API.

Edit: `@solana/buffer-layout` and `@solana/buffer-layout-utils` are published.


joncinque (dismissed):
Looks good to me, but someone else should probably weigh in too

token/ts/src/state/account.ts
```diff
@@ -0,0 +1,126 @@
+import { struct, u32, u8 } from '@solana/buffer-layout';
+import { publicKey, u64 } from '@solana/buffer-layout-utils';
+import { Commitment, Connection, PublicKey } from '@solana/web3.js';
+import { TOKEN_PROGRAM_ID } from '../constants';
+import { TokenAccountNotFoundError, TokenInvalidAccountOwnerError, TokenInvalidAccountSizeError } from '../errors';
+
+/** Information about a token account */
+export interface Account {
```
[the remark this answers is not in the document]
Arrowana: Account, owner... All those are painful for the beginners, or even painful to navigate through in codebases in general. But ok this isn't the time to change

Jacob-CG:
@jordansexton Running through examples and creating tests/reference material with this, I noticed that the build does generate `csm` and `commonjs` compatible files for browser and node compatibility. Will this be a fast follow?

Building the minified files would be good as well.

Jacob-CG:
Adding to this, I've noticed a lot of the functions have `multiSigners: Signer[]` and use `getSigners` to switch between multisig and single signer on the transaction. Would it make more sense to make `multiSigners` optional? 

I can see the above as difficult and confusing to most new people trying to use this, where they will always be putting `[]` as the parameter.

Arrowana:
> Adding to this, I've noticed a lot of the functions have `multiSigners: Signer[]` and use `getSigners` to switch between multisig and single signer on the transaction. Would it make more sense to make `multiSigners` optional?
> 
> I can see the above as difficult and confusing to most new people trying to use this, where they will always be putting `[]` as the parameter.

I agree that less args in the general case is nice, the old interfaces require passing a lot that is fixed or unnecessary in 99% of cases

jordansexton:
> Would it make more sense to make `multiSigners` optional?

I think so, I'll make this change.

token/ts/src/actions/getOrCreateAssociatedTokenAccount.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @param mint                     Mint associated with the account to set or verify
+ * @param owner                    Owner of the account to set or verify
+ * @param commitment               Desired level of commitment for querying the state
+ * @param confirmOptions           Options for confirming the transaction
+ * @param programId                SPL Token program account
+ * @param associatedTokenProgramId SPL Associated Token program account
+ *
+ * @return Address of the new associated token account
+ */
+export async function getOrCreateAssociatedTokenAccount(
```
surfertas, line 34: Is there a reason why the parameter `allowOwnerOffCurve` is not exposed here, and hard coded to `false` in the call to `getAssociatedTokenAddress`? Would like to use this function to create an associated token account for a PDA.

jordansexton: I think we could expose it, creating an ATA for a PDA is probably a common thing. @joncinque wdyt?

jordansexton: Added in 90e2aecfdb869779a4c0948e261a0cc4eff28d85

jordansexton:
> I've noticed a lot of the functions have `multiSigners: Signer[]` and use `getSigners` to switch between multisig and single signer on the transaction. Would it make more sense to make `multiSigners` optional?

Changed in fe777e57793e3d2eb9f38861b24dabb75b7063e1

jordansexton:
> I noticed that the build does not generate `esm` and `commonjs` files for browser and node compatibility. Will this be a fast follow?

Right now it just uses tsc and outputs ES modules. The expectation is that for browsers, the user will have a bundler that can transpile to CJS, provide polyfills, etc. This is what wallet-adapter does, and there are tradeoffs, and the approach is better suited for a browser-only library.

As you suggested, perhaps the best solution is to follow up on this PR with a build system, and we'll continue to compile with tsc, but also provide a fully loaded CJS build so users can just `require("@solana/spl-token")` regardless of their configuration.

token/ts/src/actions/getOrCreateAssociatedTokenAccount.ts
joncinque: Yep, that makes a lot of sense!

jordansexton:
Merging after discussion with Justin. This will live alongside the current JS implementation for a bit -- it needs a followup PR to prepare the build for release. Thanks for the reviews everyone!

PR #2554, opened 2021-11-02
Allow GC to close empty associated accounts

MarkSackerberg:
Empty associated accounts will only be closed in case the new flag del_associated_accounts is set. Otherwise behaviour is as before.
This can be used tto reclaim SOL that was sent to the associated accounts e.g. for rent exemption

Useage:
spl-token gc --del_associated_accounts

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -2140,7 +2140,7 @@ fn main() {
                 .arg(owner_keypair_arg())
                 .arg(
                     Arg::with_name("del_associated_accounts")
-                    .long("del_associated_accounts")
+                    .long("close-empty-associated-accounts")
                     .value_name("DEL_ASSOCIATED_ACCOUNTS")
                     .takes_value(false)
                     .help("set to true if all empty associated accounts should be deleted")
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1214,7 +1214,7 @@ fn command_multisig(config: &Config, address: Pubkey) -> CommandResult {
     Ok(None)
 }
 
-fn command_gc(config: &Config, owner: Pubkey, del_associated_accounts: bool) -> CommandResult {
+fn command_gc(config: &Config, owner: Pubkey, close_empty_associated_accounts: bool) -> CommandResult {
     println_display(config, "Fetching token accounts".to_string());
     let accounts = config
         .rpc_client
@@ -1295,19 +1295,11 @@ fn command_gc(config: &Config, owner: Pubkey, del_associated_accounts: bool) -> 
         }
 
         for (address, (amount, decimals, frozen, close_authority)) in accounts {
-            if address == associated_token_account {
-                if !del_associated_accounts {
-                    // leave the associated token account alone
-                    continue;
-                } else {
-                    // close non empty associated token accounts
-                    if amount > 0 {
-                        println_display(config, format!("Not closing account {}. Balance > 0", associated_token_account));
-                        continue;
-                    } else {
-                        println_display(config, format!("Closing Account {}", associated_token_account));
-                    }
-                }
+            match (address == associated_token_account, close_empty_associated_accounts, amount > 0) {
+                (true, _, true) => continue, // don't ever close associated token account with amount
+                (true, false, _) => continue, // don't close associated token account if close_empty_associated_accounts isn't set
+                (true, true, false) => println_display(config, format!("Closing Account {}", associated_token_account)),
+                _ => {},
             }
 
             if frozen {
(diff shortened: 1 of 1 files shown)


token/cli/src/main.rs
```diff
@@ -2128,6 +2138,13 @@ fn main() {
             SubCommand::with_name("gc")
                 .about("Cleanup unnecessary token accounts")
                 .arg(owner_keypair_arg())
+                .arg(
+                    Arg::with_name("del_associated_accounts")
+                    .long("del_associated_accounts")
+                    .value_name("DEL_ASSOCIATED_ACCOUNTS")
```
joncinque, line 2144: nit: this isn't needed since there's no value associated

token/cli/src/main.rs
```diff
@@ -2128,6 +2138,13 @@ fn main() {
             SubCommand::with_name("gc")
                 .about("Cleanup unnecessary token accounts")
                 .arg(owner_keypair_arg())
+                .arg(
+                    Arg::with_name("del_associated_accounts")
+                    .long("del_associated_accounts")
```
joncinque, line 2143: nit, use kebab-case for flags 
```suggestion
                    .long("close-empty-associated-accounts")
```

token/cli/src/main.rs
```diff
@@ -1296,8 +1296,18 @@ fn command_gc(config: &Config, owner: Pubkey) -> CommandResult {
...
+                    continue;
+                } else {
+                    // close non empty associated token accounts
+                    if amount > 0 {
+                        println_display(config, format!("Not closing account {}. Balance > 0", associated_token_account));
+                        continue;
+                    } else {
+                        println_display(config, format!("Closing Account {}", associated_token_account));
+                    }
+                }
```
joncinque, line 1310: nit: since there's lots possible combinations, a match statement may be easier to reason, ie something like:
```
match (address == associated_token_account, close_empty_associated_accounts, amount > 0) {
    (true, _, true) => continue, // don't ever close associated token account with amount
    (true, false, _) => continue, // don't close associated token account if close_empty_associated_accounts isn't set
    _ => {},
}
```

token/cli/src/main.rs
```diff
@@ -1296,8 +1296,18 @@ fn command_gc(config: &Config, owner: Pubkey) -> CommandResult {
...
             }
 
             if frozen {
```
joncinque, line 1313: I can't comment down there, but let's add an additional safety check on `if amount > 0` (line 1321) like:
```
if amount > 0 && address == associated_token_account {
    // we shouldn't ever be here, but if we are here, abort!
    continue;
} else if amount > 0 {
...
```

joncinque (commented):
Thanks for the contribution! Just a few minor bits before we can get this merged in

MarkSackerberg:
Hi @joncinque 
thank you for those suggestions!
I've implemented all of them. Can you please have another look?

token/cli/src/main.rs
```diff
@@ -1295,9 +1295,11 @@ fn command_gc(config: &Config, owner: Pubkey) -> CommandResult {
         }
 
         for (address, (amount, decimals, frozen, close_authority)) in accounts {
-            if address == associated_token_account {
-                // leave the associated token account alone
-                continue;
+            match (address == associated_token_account, close_empty_associated_accounts, amount > 0) {
```
joncinque, line 1298: actually, i'm just realizing, this should check `total_amount > 0` instead of `amount > 0`. if we have an auxiliary account with some balance and an empty associated account, we shouldn't close the associated account.

this also makes the sanity check actually valuable :smile:

t-nelson:
What's the motivation here?  I don't see a problem statement

MarkSackerberg:
> What's the motivation here? I don't see a problem statement

Hi @t-nelson,
the idea is to close all the empty associated token accounts to get the rent SOL back. The amounts sums up e.g. in case you do nft airdrops. Therefore the functionality was quite helpful for me.

token/cli/src/main.rs
MarkSackerberg: absolutely makes sense. Will add that.

token/cli/src/main.rs
```diff
@@ -1307,8 +1309,13 @@ fn command_gc(config: &Config, owner: Pubkey) -> CommandResult {
 
             let mut account_instructions = vec![];
 
-            // Transfer the account balance into the associated token account
-            if amount > 0 {
+            
+            if amount > 0 && address == associated_token_account {
+                // Sanity check!
+                // we shouldn't ever be here, but if we are here, abort!
+                continue;
```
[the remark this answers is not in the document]
MarkSackerberg: Sorry, i don't really understand what you are asking for. Is it changing the comment to:
// Assert!
//we shouldn't ever be here, but if we are here, abort!
?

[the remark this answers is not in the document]
joncinque: Since that case is evidence of a programming bug and not a recoverable situation, the proposed change is to instead do:
```
assert!(!(amount > 0 && address == associated_token_account));
```
or:
```
assert!(amount == 0 || address != associated_token_account);
```

joncinque (approved):
Thanks for the work, looks good!

PR #2557, opened 2021-11-03
stake-pool-py: Create and deserialize stake pools

joncinque:
#### Problem

People want to run stake pools, but writing a staking bot in Rust is a daunting task.

#### Solution

Add a python option!  List of things to do:

- [x] create the reserve stake account
- [x] create the mints and token accounts
- [x] create the stake pool / validator list
- [x] initialize the stake pool
- [x] also, CI! the `.cache` used for the solana install is overly aggressive, so this pares back how much is cached to just `.cache/solana` to not interfere with the pip cache
- [x] add / remove validators
- [x] serde on the validator list
- [x] deposit / withdraw sol
- [x] deposit / withdraw stake
- [x] update pool
- [x] increase / decrease stake
- [ ] simple bot to set levels


joncinque:
Will finish the bot in a separate PR, this is already pretty beefy

PR #2558, opened 2021-11-04
Implement multiple output format for spl-stake-pool list-all

JulI0-concerto:
The following output format have been added to spl-stake-pool list-all:
* json
* json-compact
* display


m-reka:
Thanks - Very useful!

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,179 @@
+use std::fmt::{Display, Formatter};
+use solana_cli_output::{QuietDisplay, VerboseDisplay};
+use solana_sdk::pubkey::Pubkey;
+use solana_sdk::stake::state::Lockup;
+use serde::{Serialize, Deserialize};
+use spl_stake_pool::state::{Fee, StakePool, StakeStatus, ValidatorList, ValidatorStakeInfo};
```
CriesofCarrots, line 6:
```suggestion
use {
    std::fmt::{Display, Formatter},
    solana_cli_output::{QuietDisplay, VerboseDisplay},
    solana_sdk::{pubkey::Pubkey, stake::state::Lockup},
    serde::{Serialize, Deserialize},
    spl_stake_pool::state::{Fee, StakePool, StakeStatus, ValidatorList, ValidatorStakeInfo},
};
```

CriesofCarrots: Looks like the style in this crate is for a single use statement

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,179 @@
...
+impl Display for CliStakePools {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        for pool in &self.pools {
+            writeln!(f,
+                     "Address: {}\tManager: {}\tLamports: {}\tPool tokens: {}\tValidators: {}",
+                     pool.address.to_string(),
+                     pool.manager.to_string(),
+                     pool.total_lamports.to_string(),
+                     pool.pool_token_supply.to_string(),
+                     pool.validator_list.len().to_string()
```
CriesofCarrots, line 23: I don't think you need any of these `to_string()` calls, as all these types implement Display

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,179 @@
...
+            ).ok();
+        }
+        writeln!(f, "Total number of pools: {}", &self.pools.len().to_string())
```
CriesofCarrots, line 26:
```suggestion
        writeln!(f, "Total number of pools: {}", &self.pools.len())
```
Ditto

stake-pool/cli/src/main.rs
```diff
@@ -43,10 +45,12 @@ use {
     std::cmp::Ordering,
     std::{process::exit, sync::Arc},
 };
+use crate::output::{CliStakePool, CliStakePools};
```
CriesofCarrots, line 48: Please extend line 5 above

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,179 @@
...
+            active_stake_lamports: v.active_stake_lamports,
+            transient_stake_lamports: v.transient_stake_lamports,
+            last_update_epoch: v.last_update_epoch,
+            transient_seed_suffix_start: v.transient_seed_suffix_start,
+            transient_seed_suffix_end: v.transient_seed_suffix_end,
+            status: match v.status {
+                StakeStatus::Active => "Active",
+                StakeStatus::DeactivatingTransient => "DeactivatingTransient",
+                StakeStatus::ReadyForRemoval => "ReadyForRemoval",
+            }.to_string(),
```
CriesofCarrots, line 94: Did you mean to `impl From<StakeStatus> for CliStakePoolValidatorStakeStatus` and use it here?

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+impl From<Fee> for CliStakePoolFee {
+    fn from(f: Fee) -> Self {
+        Self {
+            denominator: f.denominator,
+            numerator: f.numerator
+        }
+    }
+}
```
CriesofCarrots, line 140: I don't think you need this mapping, since the fields are just u64 and the field names already satisfy camelCase

CriesofCarrots (commented):
The OutputFormat implementation looks good! Couple style nits, plus suggestions in the cli/display area. However, I'm not super familiar with the stake-pool structs, so @joncinque may have some additional comments here.

stake-pool/cli/src/main.rs
```diff
@@ -1626,18 +1630,14 @@ fn command_set_fee(
...
-            stake_pool.pool_token_supply,
-            validator_list.validators.len()
-        );
+    let cli_stake_pool_vec: Vec<CliStakePool> = all_pools.into_iter().map(|x| CliStakePool::from(x)).collect();
+    let cli_stake_pools = CliStakePools { pools: cli_stake_pool_vec };
+    match config.output_format {
+        OutputFormat::Display => println!("{}", config.output_format.formatted_string(&cli_stake_pools)),
+        OutputFormat::Json => println!("{}", config.output_format.formatted_string(&cli_stake_pools)),
+        OutputFormat::JsonCompact => println!("{}", config.output_format.formatted_string(&cli_stake_pools)),
+        _ => {}
```
joncinque, line 1639: Any chance we could also do this on the normal `command_list_pool`?

joncinque (commented):
Agree on all of Tyera's points -- just the one bit from me to extend this to the normal `list` command

stake-pool/cli/src/output.rs
JulI0-concerto: Thx Jon but I don't really understand what you mean. Could you elaborate a bit please? I am newbie in Rust ;-)

stake-pool/cli/src/output.rs
JulI0-concerto: Yes Thx I remove it

stake-pool/cli/src/output.rs
CriesofCarrots: Hi, I'm not Jon, I'm Tyera. You tagged me on Discord :)
If you look above, I made the desired change as a github "suggestion": only one `use`, instead of 6. Does that help?

stake-pool/cli/src/output.rs
JulI0-concerto: I always separate objects between layers even if it's look like duplication of code.
You mean I should directly use Fee struct here?

What append if something is added to the Fee struct which is not camelCase or if an annotation is added which break the api?
```
#[repr(C)]
#[derive(Clone, Copy, Debug, Default, PartialEq, BorshSerialize, BorshDeserialize, BorshSchema)]
pub struct Fee {
    /// denominator of the fee ratio
    pub denominator: u64,
    /// numerator of the fee ratio
    pub numerator: u64,
}
```

stake-pool/cli/src/main.rs
JulI0-concerto: Yep good point. I will do it.

JulI0-concerto: Is it ok for you if I do it in a second PR?

JulI0-concerto: Almost done.

stake-pool/cli/src/output.rs
JulI0-concerto: Ok maybe you are right Fee struct will not be more than a numerator and a denominator after all.

JulI0-concerto: The problem is that Fee struct doesn't implement Deserializer trait.

JulI0-concerto: I stay with CliStakePoolFee

stake-pool/cli/src/main.rs
```diff
@@ -863,145 +867,148 @@ fn command_list(config: &Config, stake_pool_address: &Pubkey) -> CommandResult {
...
+            println!("SOL Deposit Authority: {}", sol_deposit_authority);
+            println!("SOL Withdraw Authority: {}", sol_withdraw_authority);
+            println!("Withdraw Authority: {}", pool_withdraw_authority);
+            println!("Pool Token Mint: {}", stake_pool.pool_mint);
+            println!("Fee Account: {}", stake_pool.manager_fee_account);
+        } else {
+            println!("Stake Pool: {}", stake_pool_address);
+            println!("Validator List: {}", stake_pool.validator_list);
+            println!("Pool Token Mint: {}", stake_pool.pool_mint);
+        }
```
CriesofCarrots, line 888: This approach doesn't seem right to me. I think this should all be implemented as Display and VerboseDisplay on a CliStakePool.
You can add additional information that should only appear in `command_list()` Display and VerboseDisplay responses to CliStakePool with the `#[serde(skip)]` tag, possibly with data wrapped in an Option to prevent extraneous calls in `command_list_all_pools()`. However, it's possible that will be complex enough that you actually need a different struct to do this simply (CliDetailedStakePool?).

JulI0-concerto: Yep Let's do it

JulI0-concerto: Done

stake-pool/cli/src/main.rs
```diff
@@ -2477,12 +2398,26 @@ fn main() {
...
         let verbose = matches.is_present("verbose");
+        let output_format = matches
+            .value_of("output_format")
+            .map(|value| match value {
+                "json" => OutputFormat::Json,
+                "json-compact" => OutputFormat::JsonCompact,
+                _ => unreachable!(),
+            })
+            .unwrap_or(if verbose {
+                println!("ZZZ");
```
CriesofCarrots, line 2409: Remove?

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,435 @@
...
+            &self.stake_referral_fee
+        )?;
+        writeln!(
+            f,
+            "SOL Deposit Referral Fee: {}% of SOL Deposit Fee",
+            &self.sol_referral_fee
+        )?;
+        Ok(())
+    }
+}
```
CriesofCarrots, line 310: Move under CliStakePool definition

stake-pool/cli/src/main.rs
```diff
@@ -852,96 +856,25 @@ fn command_deposit_sol(
 
 fn command_list(config: &Config, stake_pool_address: &Pubkey) -> CommandResult {
     let stake_pool = get_stake_pool(&config.rpc_client, stake_pool_address)?;
+    let reserve_stake_account_address = stake_pool.reserve_stake.clone().to_string();
+    let total_lamports = stake_pool.total_lamports.clone();
+    let last_update_epoch = stake_pool.last_update_epoch.clone();
     let validator_list = get_validator_list(&config.rpc_client, &stake_pool.validator_list)?;
+    let max_number_of_validators = validator_list.header.max_validators.clone();
+    let current_number_of_validators = validator_list.validators.clone().len();
```
CriesofCarrots, line 864: Are all these clones necessary? I think most or all of these types impl Copy.

stake-pool/cli/src/main.rs
```diff
@@ -1,7 +1,9 @@
 mod client;
+mod output;
 
 use {
     crate::client::*,
+    crate::output::{CliStakePool, CliStakePools, CliStakePoolDetails, CliStakePoolStakeAccountInfo},
```
CriesofCarrots, line 6:
```suggestion
    crate::{
        client::*, output::{CliStakePool, CliStakePools, CliStakePoolDetails, CliStakePoolStakeAccountInfo},
    },
```
(might need `cargo fmt`)

stake-pool/cli/src/main.rs
```diff
@@ -953,55 +886,43 @@ fn command_list(config: &Config, stake_pool_address: &Pubkey) -> CommandResult {
...
+        reserve_stake_lamports: reserve_stake.lamports,
+        minimum_reserve_stake_balance,
+        stake_accounts: cli_stake_pool_stake_account_infos,
+        total_lamports,
+        total_pool_tokens,
+        current_number_of_validators: current_number_of_validators as u32,
+        max_number_of_validators,
+        flag: total_flag.to_string(),
+    };
+    cli_stake_pool.details = Option::from(cli_stake_pool_details);
```
CriesofCarrots, line 924:
```suggestion
    cli_stake_pool.details = Some(cli_stake_pool_details);
```

stake-pool/cli/src/main.rs
```diff
@@ -953,55 +886,43 @@ fn command_list(config: &Config, stake_pool_address: &Pubkey) -> CommandResult {
...
-    println!(
-        "Max Number of Validators: {}",
-        validator_list.header.max_validators
-    );
-
+            let flag = if  validator.last_update_epoch != epoch_info.epoch {
+                " [UPDATE REQUIRED]"
+            } else {
+                ""
+            };
```
CriesofCarrots, line 893: I'd recommend these flag strings be moved into the Display implementations, so they don't clutter the json response. I'll make the relevant suggestions in the structs and methods below.

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,435 @@
...
+pub(crate) struct CliStakePoolDetails {
+    pub reserve_stake_account_address: String,
+    pub reserve_stake_lamports: u64,
+    pub minimum_reserve_stake_balance: u64,
+    pub stake_accounts: Vec<CliStakePoolStakeAccountInfo>,
+    pub total_lamports: u64,
+    pub total_pool_tokens: f64,
+    pub current_number_of_validators: u32,
+    pub max_number_of_validators: u32,
+    pub flag: String,
```
CriesofCarrots, line 86:
```suggestion
    #[serde(skip)]
    pub update_required: bool,
```

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,435 @@
...
+                stake_account.vote_account_address,
+                Sol(stake_account.validator_lamports),
+                stake_account.validator_last_update_epoch,
+            )?;
+        }
+        writeln!(
+            f,
+            "Total Pool Stake: {} {}",
+            Sol(*&self.total_lamports),
+            &self.flag,
```
CriesofCarrots, line 110:
```suggestion
            if &self.update_required {
                " [UPDATE REQUIRED]"
            } else {
                ""
            },
```

stake-pool/cli/src/main.rs
```diff
@@ -953,55 +886,43 @@ fn command_list(config: &Config, stake_pool_address: &Pubkey) -> CommandResult {
...
+            }
+        })
+        .collect();
+    let total_pool_tokens = spl_token::amount_to_ui_amount(stake_pool.pool_token_supply, pool_mint.decimals);
+    let mut cli_stake_pool = CliStakePool::from((*stake_pool_address, stake_pool, validator_list, pool_withdraw_authority));
+    let total_flag = if last_update_epoch != epoch_info.epoch {
+        " [UPDATE REQUIRED]"
+    } else {
+        ""
+    };
```
CriesofCarrots, line 912: Ditto regarding this flag

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,435 @@
...
+#[serde(rename_all = "camelCase")]
+pub(crate) struct CliStakePoolStakeAccountInfo {
+    pub vote_account_address: String,
+    pub stake_account_address: String,
+    pub validator_active_stake_lamports: u64,
+    pub validator_last_update_epoch: u64,
+    pub validator_lamports: u64,
+    pub validator_transient_stake_account_address: String,
+    pub validator_transient_stake_lamports: u64,
+    pub flag: String,
```
CriesofCarrots, line 191:
```suggestion
    #[serde(skip)]
    pub update_required: String,
```

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,435 @@
...
+            writeln!(
+                w,
+                "Vote Account: {}\tStake Account: {}\tActive Balance: {}\tTransient Stake Account: {}\tTransient Balance: {}\tLast Update Epoch: {}{}",
+                stake_account.vote_account_address,
+                stake_account.stake_account_address,
+                Sol(stake_account.validator_active_stake_lamports),
+                stake_account.validator_transient_stake_account_address,
+                Sol(stake_account.validator_transient_stake_lamports),
+                stake_account.validator_last_update_epoch,
+                stake_account.flag,
```
CriesofCarrots, line 153:
```suggestion
                if &stake_account.update_required {
                    " [UPDATE REQUIRED]"
                } else {
                    ""
                },
```

stake-pool/cli/src/output.rs
```diff
@@ -0,0 +1,435 @@
...
+            )?;
+        }
+        writeln!(
+            w,
+            "Total Pool Stake: {} {}",
+            Sol(*&self.total_lamports),
+            &self.flag,
```
CriesofCarrots, line 160:
```suggestion
            if &self.update_required {
                " [UPDATE REQUIRED]"
            } else {
                ""
            },
```

stake-pool/cli/src/client.rs
```diff
@@ -10,6 +10,7 @@ use {
     solana_program::{borsh::try_from_slice_unchecked, program_pack::Pack, pubkey::Pubkey, stake},
     spl_stake_pool::state::{StakePool, ValidatorList},
 };
+use spl_stake_pool::find_withdraw_authority_program_address;
```
CriesofCarrots, line 13:
```suggestion
    spl_stake_pool::{
        find_withdraw_authority_program_address, state::{StakePool, ValidatorList},
    },
};
```

CriesofCarrots (commented):
This is looking really close!
A couple last suggestions from me. Also, please run `cargo fmt` and `cargo clippy`. I ran CI on this PR so you can get see a preview of the things clippy will complain about.

stake-pool/cli/src/output.rs
JulI0-concerto: Ok good idea but as boolean I guess

CriesofCarrots: Oops, yes, edited

stake-pool/cli/src/main.rs
JulI0-concerto: ouch thx

stake-pool/cli/src/main.rs
JulI0-concerto: You right it's not. removed

stake-pool/cli/src/output.rs
JulI0-concerto: I don't skip serde. I think it could be a usefull info in json format.

stake-pool/cli/src/main.rs
JulI0-concerto: Yes I agree

stake-pool/cli/src/output.rs
JulI0-concerto: done

stake-pool/cli/src/main.rs
JulI0-concerto: I have changed it by update_required to be consistent.

stake-pool/cli/src/output.rs
JulI0-concerto: Done

stake-pool/cli/src/output.rs
CriesofCarrots: Fair enough. I'll let @joncinque holler if he has any objections.

joncinque: that works for me!

CriesofCarrots (approved):
Awesome, thanks for all the polish! Lgtm

PR #2560, opened 2021-11-04
Typedocs CICD

jacobcreech:
**Problem**:
Typedocs are not generated on the fly for each js/ts library under spl. They end up manually generated or lagged behind the actual releases

**Solution**:
Created a CICD process using actions to generate typedocs for each of the js/ts bindings we have under spl. Docs will be deployed on release.

You can see an example of CICD here: https://github.com/Jacob-CG/solana-program-library/runs/4051236499?check_suite_focus=true

Currently js docs are using [documentation.js](https://documentation.js.org/)


token-lending/js/package.json
```diff
@@ -23,8 +23,7 @@
     ],
     "scripts": {
         "build": "rm -rf lib/* && tsc && rollup -c && rm lib/rollup.config.d.ts",
-        "deploy": "yarn docs && gh-pages -d docs -e token-lending",
-        "docs": "rm -rf docs/* && typedoc",
+        "docs": "rm -rf ./../gh-pages/token-lending/* && typedoc --out ../../gh-pages/token-lending",
```
jordansexton, line 26: Is this path supposed to be `../../gh-pages` like the others?

token/js/package.json
```diff
@@ -35,6 +35,8 @@
   "scripts": {
     "build": "rollup -c",
     "build:browser-test": "rollup -c test/rollup.config.js",
+    "deploy": "yarn docs && gh-pages -d docs",
+    "docs": "rm -rf docs/* && documentation build client/token.js -f html --github -o ../../gh-pages/token --sort-order alpha",
```
jordansexton, line 39: FWIW, these have changed (along with the rest of this file) in master.

jordansexton (commented):
Looks pretty good. One concern is that only the JS docs are being used to deploy the gh-pages here, which will miss the docs from the READMEs for other things.

For the latest `token/js` docs, we `yarn deploy` to the `token/js` subdirectory which retains everything else.

KartikSoneji:
Hi, any updates on this?
The typedocs are lagging quite a bit behind the package.

PR #2570, opened 2021-11-12
name service: allow parent name owner to update records

dr497:


name-service/program/src/processor.rs
```diff
@@ -124,10 +124,23 @@ impl Processor {
...
+        if let Some(parent_name_acc) = parent_name {
+            if name_record_header.parent_name != *parent_name_acc.key {
+                msg!("Invalid parent name account");
+                return Err(ProgramError::InvalidArgument);
+            }
+
+            let parent_name_record_header =
+                NameRecordHeader::unpack_from_slice(&parent_name_acc.data.borrow())?;
+            is_parent_owner = parent_name_record_header.owner == *name_update_signer.key
+        }
```
mvines, line 143: I'd prefer to avoid a `mut` here if it's all the same to you:
```suggestion
        let is_parent_owner = if let Some(parent_name_acc) = parent_name {
            if name_record_header.parent_name != *parent_name_acc.key {
                msg!("Invalid parent name account");
                return Err(ProgramError::InvalidArgument);
            }

            let parent_name_record_header =
                NameRecordHeader::unpack_from_slice(&parent_name_acc.data.borrow())?;
            parent_name_record_header.owner == *name_update_signer.key
        } else { 
            false
        };
```

mvines: Also I prefer the style of reusing the variable name when stripping `Some` for quicker code scanning:
```
if let Some(parent_name) = parent_name { ...
```

mvines (dismissed):
lgtm, just a couple style nits.

name-service/program/src/instruction.rs
```diff
@@ -48,10 +48,12 @@ pub enum NameRegistryInstruction {
     ///   * If account class is `Pubkey::default()`:
     ///   0. `[writeable]` Name record to be updated
     ///   1. `[signer]` Account owner
+    ///   2. `[]` Parent name record (optional)
```
mvines, line 51: I think it'll be clearer to add a third case here rather than adding `2. `[]` Parent name record (optional)` to the two existing cases.

When the parent name record is provided, the signer account must always be the parent owner and not "account owner" or "account class"

name-service/program/src/processor.rs
dr497: I agree it's better this way!

name-service/program/src/instruction.rs
dr497: I agree it's clearer this way

PR #2573, opened 2021-11-12
fix fee denominator in command example to represent 0.3%

JulI0-concerto:


joncinque:
I may be missing something, but 3/100 = 3% and 3/1000 = 0.3%, right?

JulI0-concerto:
> I may be missing something, but 3/100 = 3% and 3/1000 = 0.3%, right?

I am working too late I guess :) We can close it

joncinque:
Hehe no worries!

PR #2579, opened 2021-11-13
Add account diagram to token swap docs

arcticmatt:
* Adds account diagram showing how all the accounts the token swap program uses are related
* Walks through an simple swap example, using the names in the diagram as references


docs/src/token-swap.md
```diff
@@ -301,3 +301,39 @@ to explain fuzzing, but the specific implementation for the program can be found
...
+Below is a diagram showing all the accounts used by the Token Swap program. 
+
+![](../static/img/token-swap.png)
+
+And here's a quick rundown of how things work, using the names in the diagram as references.
+
+### Initialization
+
+- **Swap** is created.
+- **Authority** is verified. There is one **Authority** per **Swap**, and it should be a PDA account that uses the **Swap**'s public key as a seed. Also, it should own **TokenA** and **TokenB**, but not **PoolToken** and **FeeToken**. It needs to own the first two accounts in order to perform transfers.
```
joncinque, line 316: nit: can we clarify all of the requirements on this PDA?  Perhaps we can say,
```
Also, it should:
* own **TokenA** to perform transfers
* own **TokenB** to perform transfers
* be the mint authority on **PoolTokenMint**
* not own **FeeToken** or **PoolToken**
```

docs/src/token-swap.md
```diff
@@ -301,3 +301,39 @@ to explain fuzzing, but the specific implementation for the program can be found
...
+
+### Swap
+
+Let's say the desired swap is token A ⇾ token B.
+
+First, the user that wants to perform the swap creates two new token accounts for **TokenAMint** and **TokenBMint**. We'll call these accounts **UserTokenA** and **UserTokenB** (not pictured in the diagram). The user deposits the tokens they want to swap into **UserTokenA**.
+
+Then, the user creates a throwaway keypair, and approves that keypair to transfer funds from **UserTokenA**. This is covered above:
+
+> Additionally, the user must allow for tokens to be transferred from their source token account. The best practice is to `spl_token::instruction::approve` a precise amount to a new throwaway Keypair, and then have that new Keypair sign the swap transaction. This limits the amount of tokens that can be taken from the user's account by the program.
```
joncinque, line 327: nit: Perhaps just a link to the part rather than copy / pasting from earlier in the doc?

joncinque (commented):
Thanks for the contribution, looks really great! Just a couple of nits, and then we can certainly get this merged.

PR #2593, opened 2021-11-19
spl-token: provide more useful output when using create-token with the output option and small redesign of token cli architecture

### LINKED ISSUE #2592: [spl-token] provide more useful output when using create-token with the output option
currently only the signature is provided. it would be helpful if it also output the created token address and its decimals

paul-schaaf:
closes #2592

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -19,7 +19,7 @@ use solana_clap_utils::{
     memo::memo_arg,
     nonce::*,
     offline::{self, *},
-    ArgConstant,
+    ArgConstant, DisplayError,
 };
 use solana_cli_output::{
     return_signers_data, CliSignOnlyData, CliSignature, OutputFormat, QuietDisplay,
@@ -199,7 +199,7 @@ fn is_multisig_minimum_signers(string: String) -> Result<(), String> {
 pub(crate) type Error = Box<dyn std::error::Error>;
 
 type BulkSigners = Vec<Box<dyn Signer>>;
-pub(crate) type CommandResult = Result<(), Error>;
+pub(crate) type CommandResult = Result<String, Error>;
 
 fn new_throwaway_signer() -> (Box<dyn Signer>, Pubkey) {
     let keypair = Keypair::new();
@@ -329,24 +329,20 @@ fn command_create_token(
         instructions,
     )?;
 
-    match tx_return {
-        TransactionReturnData::CliSignature(cli_signature) => {
-            print_output(
-                CliMint {
-                    address: token.to_string(),
-                    decimals,
-                    transaction_data: cli_signature,
-                },
-                "create-token",
-                config,
-            );
-        }
+    Ok(match tx_return {
+        TransactionReturnData::CliSignature(cli_signature) => format_output(
+            CliMint {
+                address: token.to_string(),
+                decimals,
+                transaction_data: cli_signature,
+            },
+            "create-token",
+            config,
+        ),
         TransactionReturnData::CliSignOnlyData(cli_sign_only_data) => {
-            print_output(cli_sign_only_data, "create-token", config)
+            format_output(cli_sign_only_data, "create-token", config)
         }
-    };
-
-    Ok(())
+    })
 }
 
 fn command_create_account(
@@ -417,17 +413,14 @@ fn command_create_account(
         instructions,
     )?;
 
-    println!(
-        "{}",
-        match tx_return {
-            TransactionReturnData::CliSignature(signature) =>
-                config.output_format.formatted_string(&signature),
-            TransactionReturnData::CliSignOnlyData(sign_only_data) =>
-                config.output_format.formatted_string(&sign_only_data),
+    Ok(match tx_return {
+        TransactionReturnData::CliSignature(signature) => {
+            config.output_format.formatted_string(&signature)
         }
-    );
-
-    Ok(())
+        TransactionReturnData::CliSignOnlyData(sign_only_data) => {
+            config.output_format.formatted_string(&sign_only_data)
+        }
+    })
 }
 
 fn command_create_multisig(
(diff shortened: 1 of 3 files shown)


token/cli/src/main.rs
```diff
@@ -415,19 +471,39 @@ fn command_create_multisig(
...
+            signers: bulk_signers,
+        },
+        config,
+        no_wait,
         minimum_balance_for_rent_exemption,
-        vec![instructions],
-    )))
+        instructions,
+    )?;
+    println!(
```
CriesofCarrots, line 484: What do you think about returning `Result<String>` from each of the the command fns, and handling the print altogether? (This is how solana-cli does it.)

I'll make a comment below about where.

token/cli/src/main.rs
```diff
@@ -2650,81 +3048,88 @@ fn main() {
...
-                    println!("{}", config.output_format.formatted_string(&signature));
-                }
-            }
-        }
-        Ok(())
-    })
     .map_err(|err| {
         eprintln!("{}", err);
         exit(1);
     });
```
CriesofCarrots, line 3058:
```suggestion
    })?;
    println!("{}", result);
```
where result is defined on L2635: `let result = match (sub_command, sub_matches) {`

token/cli/src/main.rs
```diff
@@ -2650,81 +3048,88 @@ fn main() {
...
+{
+    println!(
+        "{}",
+        config.output_format.formatted_string(&CommandOutput {
+            command_name: String::from(command_name),
+            command_output
+        })
+    )
+}
+enum HandleTxReturn {
```
CriesofCarrots, line 3073:
```suggestion
enum TransactionReturnData {
```

token/cli/src/main.rs
```diff
@@ -2650,81 +3048,88 @@ fn main() {
...
-        }
-        Ok(())
-    })
     .map_err(|err| {
         eprintln!("{}", err);
         exit(1);
     });
 }
+
+fn print_output<T>(command_output: T, command_name: &str, config: &Config)
```
CriesofCarrots, line 3061: This would return a String, if you like my suggestion

CriesofCarrots (commented):
I'm quite pleased with how this is looking. I made one suggestion about deduping all the `println`s, consistent with solana-cli. See what you think.

Please give me a bump when this is fully ready for review :)

paul-schaaf:
@CriesofCarrots ready

commands now return an output string whenever their last println was output of the commands but not when it was a status println. e.g. bench lasts line was a println that prints the final status of the bench but I wouldnt really consider that output so I had bench print it itself and then return an empty string. Does that sounds reasonable?

CriesofCarrots:
Sorry for the delay! Will review first thing tomorrow

token/cli/src/main.rs
```diff
@@ -1354,12 +1596,55 @@ fn command_gc(
...
+        )?;
+        result += &match tx_return {
+            TransactionReturnData::CliSignature(signature) => {
+                config.output_format.formatted_string(&signature)
+            }
+            TransactionReturnData::CliSignOnlyData(sign_only_data) => {
+                config.output_format.formatted_string(&sign_only_data)
+            }
+        };
+        result += "\n";
```
CriesofCarrots, line 1620: It would be better if the signature or sign-data list could be compiled into one Cli struct so that it will output one single json array, instead of multiple json maps separated by lines.
Do you mind fixing in a follow-up PR?

token/cli/src/main.rs
```diff
@@ -270,13 +277,16 @@ pub fn signers_of(
...
+#[allow(clippy::too_many_arguments)]
 fn command_create_token(
     config: &Config,
     decimals: u8,
     token: Pubkey,
     authority: Pubkey,
     enable_freeze: bool,
     memo: Option<String>,
+    bulk_signers: Vec<Box<dyn Signer>>,
+    no_wait: bool,
```
CriesofCarrots, line 289: Sort of a nit, but the fact that no_wait is now passed to every command instruction is easy to misread as an indication that the no_wait functionality is available for various operations, when in truth it's only available for `command_transfer`.

Let's remove the line: `let mut no_wait = false;` and pass `matches.is_present("no_wait")` into `command_transfer` directly.

token/cli/src/main.rs
```diff
@@ -2428,6 +2723,8 @@ fn main() {
                 recipient_is_ata_owner,
                 use_unchecked_instruction,
                 memo,
+                bulk_signers,
+                no_wait,
```
CriesofCarrots, line 2727:
```suggestion
                matches.is_present("no_wait"),
```

CriesofCarrots (commented):
Couple small things, otherwise looks great!

token/cli/src/main.rs
```diff
@@ -308,17 +318,40 @@ fn command_create_token(
...
+            CliMint {
+                address: token.to_string(),
+                decimals,
+                transaction_data: cli_signature,
+            },
+            "create-token",
+            config,
+        ),
+        TransactionReturnData::CliSignOnlyData(cli_sign_only_data) => {
+            format_output(cli_sign_only_data, "create-token", config)
```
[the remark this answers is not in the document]
paul-schaaf: agree. I thought we could use an enum for all the command names with a lib like strum so we can use it for all commands and in the cli arg definitions as well. but this seemed like a thing for a separate PR so I left it like this. but I'll adjust this here as quick fix for now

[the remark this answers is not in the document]
CriesofCarrots: Ah, gotcha! Thanks, that sounds good

token/cli/src/main.rs
paul-schaaf: yea I agree this would be better. I just wanted to keep it like it was before this change for this PR so I don't change too much. happy to do it in a follow-up pr yes

CriesofCarrots (approved):
lgtm! Thanks for the polish

PR #2601, opened 2021-11-29
Update comment on state.rs regarding is_native

vpontis:
I was confused when reading this comment and had to read the code to realize that the comment isn't a typo. 

(Btw, let me know if this kind of small changes are helpful or not 👍.)


mvines (approved):
lgtm, thanks for the comment improvement!

PR #2608, opened 2021-12-01
token-2022: Add transfer fee types and instructions

joncinque:
#### Problem

It isn't possible to assess fees on token transfers.

#### Solution

Add new account types for Mints / Accounts with transfer fees, along with an over-type to handle any type, see `UberMint` for more info.

Add instructions for initializing mints with transfer fees, setting fees, transferring tokens, and transferring withheld tokens.

### DIFF
--- a/token/program-2022/src/instruction.rs
+++ b/token/program-2022/src/instruction.rs
@@ -54,7 +54,7 @@ pub enum TokenInstruction {
     /// Otherwise another party can acquire ownership of the uninitialized
     /// account.
     ///
-    /// The mint provided must be a `Mint` instance.
+    /// The mint provided must be a `Mint` instance without any mixins.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -95,8 +95,8 @@ pub enum TokenInstruction {
     /// amounts of SOL and Tokens will be transferred to the destination
     /// account.
     ///
-    /// If the accounts are `PaymentAccount`s, this will fail.  The mint
-    /// is required in order to assess the fee.
+    /// If the accounts contain `AccountTransferFeeMixin`s, this will fail.  Mints with
+    /// the `MintTransferFeeMixin` are required in order to assess the fee.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -205,8 +205,8 @@ pub enum TokenInstruction {
     /// Close an account by transferring all its SOL to the destination account.
     /// Non-native accounts may only be closed if its token amount is zero.
     ///
-    /// Accounts with transfer fees may only be closed if its withheld amount is
-    /// also zero.
+    /// Accounts with the `AccountTransferFeeMixin` may only be closed if the withheld
+    /// amount is zero.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -262,7 +262,7 @@ pub enum TokenInstruction {
     /// decimals value is checked by the caller.  This may be useful when
     /// creating transactions offline or within a hardware wallet.
     ///
-    /// If the accounts are `PaymentAccount`s, the fee is withheld in the
+    /// If the accounts contain `AccountTransferFeeMixin`s, the fee is withheld in the
     /// destination account.
     ///
     /// Accounts expected by this instruction:
@@ -368,7 +368,7 @@ pub enum TokenInstruction {
     /// Cross Program Invocation from an instruction that does not need the owner's
     /// `AccountInfo` otherwise.
     ///
-    /// The mint provided must be a `Mint` instance.
+    /// The mint provided must be a `Mint` instance without any mixins.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -391,7 +391,7 @@ pub enum TokenInstruction {
     SyncNative,
     /// Like InitializeAccount2, but does not require the Rent sysvar to be provided
     ///
-    /// The mint provided must be a `Mint` instance.
+    /// The mint provided must be a `Mint` instance without any mixins.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -427,48 +427,48 @@ pub enum TokenInstruction {
         /// The freeze authority/multisignature of the mint.
         freeze_authority: COption<Pubkey>,
     },
-    /// Initialize a new mint for payment systems.
+    /// Initialize a new mint with all provided mixins.
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
+    /// bytes). If any mixins are provided, there must be 83 bytes of padding,
+    /// 1 byte reserved for the account type, then space required by each mixin.
     ///
     /// Accounts expected by this instruction:
     ///
     ///   0. `[writable]` The mint to initialize.
-    InitializePaymentMint {
+    ///   1. ..1+N `[]` (Optional) Accounts required to initialize the mixins
+    InitializeMintWithMixins {
         /// Number of base 10 digits to the right of the decimal place.
         decimals: u8,
         /// The authority/multisignature to mint tokens.
         mint_authority: Pubkey,
         /// The freeze authority/multisignature of the mint.
         freeze_authority: COption<Pubkey>,
-        /// Pubkey that may update the fees
-        fee_config_authority: COption<Pubkey>,
-        /// Harvest instructions must be signed by this key
-        fee_withdraw_authority: COption<Pubkey>,
-        /// Amount of transfer collected as fees, expressed as basis points of the
-        /// transfer amount
-        transfer_fee_basis_points: u16,
-        /// Maximum fee assessed on transfers
-        maximum_fee: u64,
+        /// Initialization parameters for all mint mixins
+        mint_mixins: Vec<InitializeMintMixinParameters>,
     },
-    /// Initializes a new `PaymentAccount` to hold tokens.  The mint provided must
-    /// be a `PaymentMint`, and the account must have enough data allocated for
-    /// a `PaymentAccount` instance.
+    /// Initializes a new `Account` to hold tokens with all expected mixins.
+    /// The account must have enough data allocated for all mixins included in
+    /// the mint.
     ///
     /// Accounts expected by this instruction:
     ///
     ///   0. `[writable]`  The account to initialize.
-    ///   1. `[]` The mint this account will be associated with. Must be a `PaymentMint`.
-    InitializePaymentAccount {
+    ///   1. `[]` The mint this account will be associated with.
+    InitializeAccountWithMixins {
         /// The new account's owner/multisignature.
         owner: Pubkey,
+        /// Initialization parameters for all account mixins
+        account_mixins: Vec<InitializeAccountMixinParameters>,
     },
     /// Transfer, providing expected mint information and fees
     ///
     /// Accounts expected by this instruction:
     ///
     ///   * Single owner/delegate
-    ///   0. `[writable]` The source account. Must be a PaymentAccount.
-    ///   1. `[]` The token mint. Must be a PaymentMint.
-    ///   2. `[writable]` The destination account. Must be a PaymentAccount.
+    ///   0. `[writable]` The source account. Must include the `AccountTransferFeeMixin`.
+    ///   1. `[]` The token mint. Must include the `AccountMintFeeMixin`.
+    ///   2. `[writable]` The destination account. Must include the `AccountTransferFeeMixin`.
     ///   3. `[signer]` The source account's owner/delegate.
     ///
     ///   * Multisignature owner/delegate
@@ -494,18 +494,19 @@ pub enum TokenInstruction {
     /// Accounts expected by this instruction:
     ///
     ///   * Single owner/delegate
-    ///   0. `[writable]` The token mint. Must be a PaymentMint.
-    ///   1. `[writable]` The fee receiver account. Must be a PaymentAccount
-    ///      associated with the provided PaymentMint.
+    ///   0. `[writable]` The token mint. Must include the `MintTransferFeeMixin`.
+    ///   1. `[writable]` The fee receiver account. Must include the `AccountTransferFeeMixin`.
+    ///      associated with the provided mint.
     ///   2. `[signer]` The mint's `fee_withdraw_authority`
-    ///   3. ..3+N `[writable]` (Optional) The source accounts to harvest from. Must be PaymentAccounts.
+    ///   3. ..3+N `[writable]` (Optional) The source accounts to harvest from.
+    ///      Must include the `AccountTransferFeeMixin`.
     ///
     ///   * Multisignature owner/delegate
-    ///   0. `[writable]` The token mint. Must be a PaymentMint.
+    ///   0. `[writable]` The token mint.
     ///   1. `[writable]` The destination account.
     ///   2. `[]` The source or destination account's multisignature owner/delegate.
     ///   3. ..3+M `[signer]` M signer accounts.
-    ///   3+M+1. ..3+M+N`[writable]` (Optional) The source accounts to harvest from. Must be PaymentAccounts.
+    ///   3+M+1. ..3+M+N`[writable]` (Optional) The source accounts to harvest from.
     HarvestFee,
     /// Close an account by transferring all its SOL to the destination account.
     ///
@@ -526,8 +527,8 @@ pub enum TokenInstruction {
     ///   2. `[writable]` The mint.
     ///   3. `[]` The account's multisignature owner.
     ///   4. ..4+M `[signer]` M signer accounts.
-    ClosePaymentAccount,
-    /// Set transfer fee. Only supported for `PaymentMint`s.
+    CloseAccountWithTransferFee,
+    /// Set transfer fee. Only supported for mints that include the `MintTransferFeeMixin`.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -762,39 +763,6 @@ impl TokenInstruction {
                 buf.extend_from_slice(mint_authority.as_ref());
                 Self::pack_pubkey_option(freeze_authority, &mut buf);
             }
-            &Self::InitializePaymentMint {
-                decimals: _,
-                mint_authority: _,
-                freeze_authority: _,
-                fee_account_owner: _,
-                transfer_fee_basis_points: _,
-                maximum_fee: _,
-            } => {
-                panic!();
-            }
-            &Self::InitializePaymentAccount { owner: _ } => {
-                panic!();
-            }
-            &Self::TransferCheckedWithFee {
-                amount: _,
-                decimals: _,
-                transfer_fee_basis_points: _,
-                maximum_fee: _,
-            } => {
-                panic!();
-            }
-            &Self::HarvestFee => {
-                panic!();
-            }
-            &Self::ClosePaymentAccount => {
-                panic!();
-            }
-            &Self::SetTransferFee {
-                transfer_fee_basis_points: _,
-                maximum_fee: _,
-            } => {
-                panic!();
-            }
         };
         buf
     }
@@ -873,6 +841,38 @@ impl AuthorityType {
             _ => Err(TokenError::InvalidInstruction.into()),
         }
     }
+}
+
+/// Specifies the authority type for SetAuthority instructions
+#[derive(Clone, Debug, PartialEq)]
+pub enum InitializeMintMixinParameters {
+    /// Additional padding required, in case the size of the account is the same
+    /// as a multisig
+    None {
+        padding: u64,
+    },
+    /// Initializes a mint with the `MintTransferFeeMixin`
+    TransferFee {
+        /// Pubkey that may update the fees
+        fee_config_authority: COption<Pubkey>,
+        /// Harvest instructions must be signed by this key
+        fee_withdraw_authority: COption<Pubkey>,
+        /// Amount of transfer collected as fees, expressed as basis points of the
+        /// transfer amount
+        transfer_fee_basis_points: u16,
+        /// Maximum fee assessed on transfers
+        maximum_fee: u64,
+    },
+}
+
+/// Specifies the authority type for SetAuthority instructions
+#[derive(Clone, Debug, PartialEq)]
+pub enum InitializeAccountMixinParameters {
+    /// Additional padding required, in case the size of the account is the same
+    /// as a multisig
+    None,
+    /// Initializes an account with the `AccountTransferFeeMixin`
+    TransferFee,
 }
 
 /// Creates a `InitializeMint` instruction.
--- a/token/program-2022/src/instruction.rs
+++ b/token/program-2022/src/instruction.rs
@@ -29,7 +29,7 @@ pub enum TokenInstruction {
     /// Otherwise another party can acquire ownership of the uninitialized
     /// account.
     ///
-    /// Any mixins must be initialized before this instruction.
+    /// Any extensions must be initialized before this instruction.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -56,7 +56,7 @@ pub enum TokenInstruction {
     /// Otherwise another party can acquire ownership of the uninitialized
     /// account.
     ///
-    /// All required mixins will be initialized at the same time.
+    /// All required extensions will be initialized at the same time.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -97,8 +97,8 @@ pub enum TokenInstruction {
     /// amounts of SOL and Tokens will be transferred to the destination
     /// account.
     ///
-    /// If the accounts contain `AccountTransferFeeMixin`s, this will fail.  Mints with
-    /// the `MintTransferFeeMixin` are required in order to assess the fee.
+    /// If the accounts contain `AccountTransferFeeExtension`s, this will fail.  Mints with
+    /// the `MintTransferFeeExtension` are required in order to assess the fee.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -207,10 +207,10 @@ pub enum TokenInstruction {
     /// Close an account by transferring all its SOL to the destination account.
     /// Non-native accounts may only be closed if its token amount is zero.
     ///
-    /// Accounts with the `AccountTransferFeeMixin` may only be closed if the withheld
+    /// Accounts with the `AccountTransferFeeExtension` may only be closed if the withheld
     /// amount is zero.
     ///
-    /// Mints may be closed if they have the `MintCloseAuthority` mixin and their token
+    /// Mints may be closed if they have the `MintCloseAuthority` extension and their token
     /// supply is zero
     ///
     /// Accounts expected by this instruction:
@@ -267,7 +267,7 @@ pub enum TokenInstruction {
     /// decimals value is checked by the caller.  This may be useful when
     /// creating transactions offline or within a hardware wallet.
     ///
-    /// If the accounts contain `AccountTransferFeeMixin`s, the fee is withheld in the
+    /// If the accounts contain `AccountTransferFeeExtension`s, the fee is withheld in the
     /// destination account.
     ///
     /// Accounts expected by this instruction:
@@ -373,7 +373,7 @@ pub enum TokenInstruction {
     /// Cross Program Invocation from an instruction that does not need the owner's
     /// `AccountInfo` otherwise.
     ///
-    /// All required mixins will be initialized at the same time.
+    /// All required extensions will be initialized at the same time.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -396,7 +396,7 @@ pub enum TokenInstruction {
     SyncNative,
     /// Like InitializeAccount2, but does not require the Rent sysvar to be provided
     ///
-    /// All required mixins will be initialized at the same time.
+    /// All required extensions will be initialized at the same time.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -420,7 +420,7 @@ pub enum TokenInstruction {
     },
     /// Like InitializeMint, but does not require the Rent sysvar to be provided
     ///
-    /// Any mixins must be initialized before this instruction.
+    /// Any extensions must be initialized before this instruction.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -444,14 +444,14 @@ pub enum TokenInstruction {
     ///
     ///   0. `[]` The mint to calculate for
     GetAccountDataSize,
-    /// Initialize the close account mixin on a new mint.
+    /// Initialize the close account authority on a new mint.
     ///
     /// Fails if the mint has already been initialized, so must be called before
     /// `InitializeMint`.
     ///
     /// The mint must have exactly enough space allocated for the base mint (82
     /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
-    /// then space required for the mixin.
+    /// then space required for the extension.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -459,14 +459,14 @@ pub enum TokenInstruction {
     InitializeMintCloseAuthority {
         close_authority: COption<Pubkey>,
     },
-    /// Initialize the transfer fee mixin on a new mint.
+    /// Initialize the transfer fee on a new mint.
     ///
     /// Fails if the mint has already been initialized, so must be called before
     /// `InitializeMint`.
     ///
     /// The mint must have exactly enough space allocated for the base mint (82
     /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
-    /// then space required for the mixin.
+    /// then space required for the extension.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -487,9 +487,9 @@ pub enum TokenInstruction {
     /// Accounts expected by this instruction:
     ///
     ///   * Single owner/delegate
-    ///   0. `[writable]` The source account. Must include the `AccountTransferFeeMixin`.
-    ///   1. `[]` The token mint. Must include the `AccountMintFeeMixin`.
-    ///   2. `[writable]` The destination account. Must include the `AccountTransferFeeMixin`.
+    ///   0. `[writable]` The source account. Must include the `AccountTransferFeeExtension`.
+    ///   1. `[]` The token mint. Must include the `AccountMintFeeExtension`.
+    ///   2. `[writable]` The destination account. Must include the `AccountTransferFeeExtension`.
     ///   3. `[signer]` The source account's owner/delegate.
     ///
     ///   * Multisignature owner/delegate
@@ -507,14 +507,14 @@ pub enum TokenInstruction {
         /// the transfer_fee_basis_points and maximum_fee of the mint.
         fee: u64,
     },
-    /// Transfer all withheld tokens to a fee account. Signed by the mint's
+    /// Transfer all withheld tokens to an account. Signed by the mint's
     /// fee withdraw authority.
     ///
     /// Accounts expected by this instruction:
     ///
     ///   * Single owner/delegate
-    ///   0. `[writable]` The token mint. Must include the `MintTransferFeeMixin`.
-    ///   1. `[writable]` The fee receiver account. Must include the `AccountTransferFeeMixin`.
+    ///   0. `[writable]` The token mint. Must include the `MintTransferFeeExtension`.
+    ///   1. `[writable]` The fee receiver account. Must include the `AccountTransferFeeExtension`.
     ///      associated with the provided mint.
     ///   2. `[signer]` The mint's `fee_withdraw_authority`
     ///
@@ -523,18 +523,20 @@ pub enum TokenInstruction {
     ///   1. `[writable]` The destination account.
     ///   2. `[]` The mint's `fee_withdraw_authority`'s multisignature owner/delegate.
     ///   3. ..3+M `[signer]` M signer accounts.
-    HarvestFeeFromMint,
+    HarvestWithheldTokensFromMint,
     /// Permissionless instruction to transfer all withheld tokens to the mint.
     ///
     /// Succeeds for frozen accounts.
+    ///
+    /// Accounts provided should include the `AccountTransferFeeExtension`. If not,
+    /// the account is skipped.
     ///
     /// Accounts expected by this instruction:
     ///
     ///   0. `[writable]` The mint.
     ///   1. ..1+N `[writable]` The source accounts to harvest from.
-    ///      Must include the `AccountTransferFeeMixin`.
-    HarvestFeeToMint,
-    /// Set transfer fee. Only supported for mints that include the `MintTransferFeeMixin`.
+    HarvestWithheldTokensToMint,
+    /// Set transfer fee. Only supported for mints that include the `MintTransferFeeExtension`.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -818,10 +820,10 @@ pub enum AuthorityType {
     AccountOwner,
     /// Authority to close a token account
     CloseAccount,
-    /// Authority to set the fee
-    FeeConfig,
-    /// Authority to withdraw fees from accounts and mint
-    FeeWithdraw,
+    /// Authority to set the transfer fee
+    TransferFeeConfig,
+    /// Authority to withdraw withheld tokens from a mint
+    WithheldWithdraw,
 }
 
 impl AuthorityType {
--- a/token/program-2022/src/instruction.rs
+++ b/token/program-2022/src/instruction.rs
@@ -90,6 +90,9 @@ pub enum TokenInstruction {
     /// amounts of SOL and Tokens will be transferred to the destination
     /// account.
     ///
+    /// If the accounts are `AccountWithTransferFee`s, this will fail.  The mint
+    /// is required in order to assess the fee.
+    ///
     /// Accounts expected by this instruction:
     ///
     ///   * Single owner/delegate
@@ -196,6 +199,8 @@ pub enum TokenInstruction {
     },
     /// Close an account by transferring all its SOL to the destination account.
     /// Non-native accounts may only be closed if its token amount is zero.
+    /// Accounts with transfer fees may only be closed if its withheld amount is
+    /// also zero.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -250,6 +255,10 @@ pub enum TokenInstruction {
     /// This instruction differs from Transfer in that the token mint and
     /// decimals value is checked by the caller.  This may be useful when
     /// creating transactions offline or within a hardware wallet.
+    ///
+    /// If the accounts are `AccountWithTransferFee`s, the fee is withheld in the
+    /// source account. If either account is owned by the `fee_account_owner`,
+    /// the fee is ignored.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -408,6 +417,102 @@ pub enum TokenInstruction {
         mint_authority: Pubkey,
         /// The freeze authority/multisignature of the mint.
         freeze_authority: COption<Pubkey>,
+    },
+
+    /// Initialize a new mint with a transfer fee
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint to initialize.
+    ///
+    InitializeMintWithTransferFee {
+        /// Number of base 10 digits to the right of the decimal place.
+        decimals: u8,
+        /// The authority/multisignature to mint tokens.
+        mint_authority: Pubkey,
+        /// The freeze authority/multisignature of the mint.
+        freeze_authority: COption<Pubkey>,
+        /// Pubkey that must own any provided fee account
+        fee_account_owner: Pubkey,
+        /// Amount of transfer collected as fees, expressed as basis points of the
+        /// transfer amount
+        transfer_fee_basis_points: u16,
+        /// Maximum fee assessed on transfers
+        maximum_fee: u64,
+    },
+    /// Transfer, providing an explicit fee receiver
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[writable]` The source account. Must be an AccountWithTransferFee.
+    ///   1. `[]` The token mint. Must be a MintWithTransferFee.
+    ///   2. `[writable]` The destination account. Must be an AccountWithTransferFee.
+    ///   3. `[writable]` The fee receiver account. Must be owned by the mint's
+    ///      `fee_account_owner`.
+    ///   4. `[signer]` The source account's owner/delegate.
+    ///
+    ///   * Multisignature owner/delegate
+    ///   0. `[writable]` The source account.
+    ///   1. `[]` The token mint.
+    ///   2. `[writable]` The destination account.
+    ///   3. `[writable]` The fee receiver account. Must be owned by the mint's
+    ///      `fee_account_owner`.
+    ///   4. `[]` The source account's multisignature owner/delegate.
+    ///   5. ..5+M `[signer]` M signer accounts.
+    TransferWithFee {
+        /// The amount of tokens to transfer.
+        amount: u64,
+        /// Expected number of base 10 digits to the right of the decimal place.
+        decimals: u8,
+    },
+    /// Transfer withheld tokens to a fee account. Signed by either the source
+    /// account owner or the fee account owner.
+    ///
+    /// Succeeds for frozen accounts. <-- Does this make sense?
+    ///
+    /// Note: this could be made permissionless, but that may open up attack
+    /// opportunities.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[writable]` The source account. Must be an AccountWithTransferFee.
+    ///   1. `[]` The token mint. Must be a MintWithTransferFee.
+    ///   2. `[writable]` The fee receiver account. Must be owned by the mint's
+    ///      `fee_account_owner`.
+    ///   3. `[signer]` The source account's owner or the mint's `fee_account_owner`
+    ///
+    ///   * Multisignature owner/delegate
+    ///   0. `[writable]` The source account.
+    ///   1. `[]` The token mint.
+    ///   2. `[writable]` The destination account.
+    ///   3. `[]` The source or destination account's multisignature owner/delegate.
+    ///   4. ..4+M `[signer]` M signer accounts.
+    TransferWithheld {
+        /// The amount of tokens to transfer.
+        amount: u64,
+        /// Expected number of base 10 digits to the right of the decimal place.
+        decimals: u8,
+    },
+    /// Set transfer fee. Only supported for `MintWithTransferFee`s.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[signer]` The mint's fee account owner.
+    ///
+    ///   * Multisignature authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[]` The mint's multisignature fee account owner.
+    ///   2. ..2+M `[signer]` M signer accounts.
+    SetTransferFee {
+        /// Amount of transfer collected as fees, expressed as basis points of the
+        /// transfer amount
+        transfer_fee_basis_points: u16,
+        /// Maximum fee assessed on transfers
+        maximum_fee: u64,
     },
 }
 impl TokenInstruction {
@@ -625,6 +730,34 @@ impl TokenInstruction {
                 buf.extend_from_slice(mint_authority.as_ref());
                 Self::pack_pubkey_option(freeze_authority, &mut buf);
             }
+            &Self::InitializeMintWithTransferFee {
+                decimals: _,
+                mint_authority: _,
+                freeze_authority: _,
+                fee_account_owner: _,
+                transfer_fee_basis_points: _,
+                maximum_fee: _,
+            } => {
+                panic!();
+            }
+            &Self::TransferWithFee {
+                amount: _,
+                decimals: _,
+            } => {
+                panic!();
+            }
+            &Self::TransferWithheld {
+                amount: _,
+                decimals: _,
+            } => {
+                panic!();
+            }
+            &Self::SetTransferFee {
+                transfer_fee_basis_points: _,
+                maximum_fee: _,
+            } => {
+                panic!();
+            }
         };
         buf
     }
@@ -674,6 +807,9 @@ pub enum AuthorityType {
     AccountOwner,
     /// Authority to close a token account
     CloseAccount,
+    /// Authority to set the fee and fee owner. Can only be set by the current
+    /// fee owner. <-- does this make sense? Also, how's the name?
+    FeeOwner,
 }
 
 impl AuthorityType {
@@ -683,6 +819,7 @@ impl AuthorityType {
             AuthorityType::FreezeAccount => 1,
             AuthorityType::AccountOwner => 2,
             AuthorityType::CloseAccount => 3,
+            AuthorityType::FeeOwner => 4,
         }
     }
 
@@ -692,6 +829,7 @@ impl AuthorityType {
             1 => Ok(AuthorityType::FreezeAccount),
             2 => Ok(AuthorityType::AccountOwner),
             3 => Ok(AuthorityType::CloseAccount),
+            4 => Ok(AuthorityType::FeeOwner),
             _ => Err(TokenError::InvalidInstruction.into()),
         }
     }
--- a/token/program-2022/src/instruction.rs
+++ b/token/program-2022/src/instruction.rs
@@ -56,7 +56,7 @@ pub enum TokenInstruction {
     /// Otherwise another party can acquire ownership of the uninitialized
     /// account.
     ///
-    /// Any mixins must be initialized before this instruction.
+    /// All required mixins will be initialized at the same time.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -209,6 +209,9 @@ pub enum TokenInstruction {
     ///
     /// Accounts with the `AccountTransferFeeMixin` may only be closed if the withheld
     /// amount is zero.
+    ///
+    /// Mints may be closed if they have the `MintCloseAuthority` mixin and their token
+    /// supply is zero
     ///
     /// Accounts expected by this instruction:
     ///
@@ -370,7 +373,7 @@ pub enum TokenInstruction {
     /// Cross Program Invocation from an instruction that does not need the owner's
     /// `AccountInfo` otherwise.
     ///
-    /// Any mixins must be initialized before this instruction.
+    /// All required mixins will be initialized at the same time.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -393,7 +396,7 @@ pub enum TokenInstruction {
     SyncNative,
     /// Like InitializeAccount2, but does not require the Rent sysvar to be provided
     ///
-    /// Any mixins must be initialized before this instruction.
+    /// All required mixins will be initialized at the same time.
     ///
     /// Accounts expected by this instruction:
     ///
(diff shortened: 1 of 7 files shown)


token/program-2022/src/instruction.rs
```diff
@@ -409,6 +418,102 @@ pub enum TokenInstruction {
...
+    ///   * Multisignature authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[]` The mint's multisignature fee account owner.
+    ///   2. ..2+M `[signer]` M signer accounts.
+    SetTransferFee {
+        /// Amount of transfer collected as fees, expressed as basis points of the
+        /// transfer amount
+        transfer_fee_basis_points: u16,
+        /// Maximum fee assessed on transfers
+        maximum_fee: u64,
```
mvines, line 515:
```suggestion
        maximum_fee: Option<u64>,  /// `None` means there's no maximum fee
```

token/program-2022/src/instruction.rs
```diff
@@ -251,6 +256,10 @@ pub enum TokenInstruction {
     /// decimals value is checked by the caller.  This may be useful when
     /// creating transactions offline or within a hardware wallet.
     ///
+    /// If the accounts are `AccountWithTransferFee`s, the fee is withheld in the
```
mvines, line 259: A `TransferCheckedWithFee` seems needed where the caller can declare the fee they expect to pay upfront.  Otherwise you could get rugged by the fee authority using `SetTransferFee` to dial the fee up to 100% on you unexpectedly

token/program-2022/src/state.rs
```diff
@@ -80,6 +103,55 @@ impl Pack for Mint {
     }
 }
 
+/// MintWithTransferFee data.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct MintWithTransferFee {
```
mvines, line 109: I think we'll need a `withheld_amount: u64` in the mint as well, so that there's somewhere for an `AccountWithTransferFee`'s `withheld_amount` to go when the `Burn` instruction is used.

Notably, a `Burn` of 0 could be used to empty the `withheld_amount` field of an account.

token/program-2022/src/state.rs
```diff
@@ -80,6 +103,55 @@ impl Pack for Mint {
...
+/// the appropriate type.
+/// When writing data back to the buffer, `UberMint::pack` respects the type contained,
+/// and errors if unusable fields have been populated or changed.  For example,
+/// if the fee is set on a normal Mint, it will error.
+///
+/// Name TBD. Other options: MetaMint, AllMint, OverMint, MegaMint.
+/// OR we rename Mint to BaseMint, and then call this Mint.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct UberMint {
```
mvines, line 138: I'm not yet convinced about the utility of the uber structs, I suspect it'll be more flexible to make the caller deal with the different types individually because it may not always be the case that new account types will be a superset of older ones.  But happy to save this chat for later too

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +418,102 @@ pub enum TokenInstruction {
...
+    ///      `fee_account_owner`.
+    ///   3. `[signer]` The source account's owner or the mint's `fee_account_owner`
+    ///
+    ///   * Multisignature owner/delegate
+    ///   0. `[writable]` The source account.
+    ///   1. `[]` The token mint.
+    ///   2. `[writable]` The destination account.
+    ///   3. `[]` The source or destination account's multisignature owner/delegate.
+    ///   4. ..4+M `[signer]` M signer accounts.
+    TransferWithheld {
```
mvines, line 492: Instead of `TransferWithFee`/`TransferWithheld`, what if we had a `HarvestFee` instruction that accepts N input `AccountWithTransferFee` or `MintWithTransferFee` accounts, plus a destination `AccountWithTransferFee` that receives the full `withheld_amount` from all input accounts.  Signed by the `fee_withdraw_authority`

token/program-2022/src/state.rs
```diff
@@ -80,6 +103,55 @@ impl Pack for Mint {
...
+pub struct MintWithTransferFee {
+    /// Base mint data, done as composition
+    pub base_mint: Mint,
+
+    /// ALL NEW DATA MUST START FROM HERE
+    /// account type, always set to "MintWithTransferFee"
+    pub account_type: AccountType,
+    /// Any provided fee account must be owned by this key. If set to `None`, then
+    /// this mint behaves like a normal `Mint`.
+    pub fee_account_owner: COption<Pubkey>,
```
mvines, line 118: Maybe there's two authorities here?
* `fee_withdraw_authority` - who can get at the `withheld_amount` fields
* `fee_config_authority` - who can change the fee

token/program-2022/src/state.rs
```diff
@@ -191,6 +263,34 @@ impl Default for AccountState {
     }
 }
 
+/// AccountWithTransferFee data.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct AccountWithTransferFee {
```
mvines, line 269: Were you thinking there'd be an `InitializeAccountWithTransferFee` instruction, or would `InitializeAccount`/`InitializeAccount2`/`InitializeAccount3` figure `AccountType` based on the length of the account data provided?

mvines (commented):
Thanks, this is a great start. 

Something to 📌: the ATA program will need to change slightly to support the new account type as well.

token/program-2022/src/instruction.rs
joncinque: I was thinking it could be set to u64::MAX for no max fee, but Option is probably clearer

token/program-2022/src/instruction.rs
joncinque: Ah yes, that's a great idea

token/program-2022/src/instruction.rs
```diff
@@ -90,6 +90,9 @@ pub enum TokenInstruction {
     /// amounts of SOL and Tokens will be transferred to the destination
     /// account.
     ///
+    /// If the accounts are `AccountWithTransferFee`s, this will fail.  The mint
```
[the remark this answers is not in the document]
joncinque: Yes please

token/program-2022/src/state.rs
```diff
@@ -10,6 +10,29 @@ use solana_program::{
...
+    /// Nothing, uninitialized
+    Uninitialized,
+    /// Base mint
+    Mint,
+    /// Base token account
+    Account,
+    /// Mints that assess a transfer fee
+    MintWithTransferFee,
+    /// Accounts that require transfer fees
+    AccountWithTransferFee,
```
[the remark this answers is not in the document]
joncinque: Will do

token/program-2022/src/instruction.rs
```diff
@@ -251,6 +256,10 @@ pub enum TokenInstruction {
     /// decimals value is checked by the caller.  This may be useful when
     /// creating transactions offline or within a hardware wallet.
     ///
+    /// If the accounts are `AccountWithTransferFee`s, the fee is withheld in the
+    /// source account. If either account is owned by the `fee_account_owner`,
```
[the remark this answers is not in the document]
joncinque: I went both ways on this one but don't have a strong opinion.  I thought it would be clearer to keep it in the source account, so that the recipient never needs to know about fees withheld.

If I see that I just received $10 in my balance, and also see $1 withheld in my account, I'm going to wonder how to access that, and then get angry when I find out about transfer fees.  But maybe that part is just never shown, so people never ask questions.

As for the additional transfer of all current withheld amount, it certainly does the sweeping in fewer transactions.  On the other hand, I'm writing a script to sweep fees from accounts, there's a chance that those fees will be gone between the time I see the withheld amount and the time I execute the transaction, and then I need to search again to see where those fees went.  That seems unintuitive to me.

token/program-2022/src/instruction.rs
```diff
@@ -251,6 +256,10 @@ pub enum TokenInstruction {
...
+    /// the fee is ignored.
```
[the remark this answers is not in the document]
joncinque: Fair enough, I don't have strong feelings about this.  I was thinking more about defi cases where the fee account can make it impossible remove all liquidity, but this won't be the case anyway with the harvest instruction.

token/program-2022/src/state.rs
joncinque: Yep, that makes more sense

token/program-2022/src/state.rs
joncinque: The latter -- accounts are initialized with 0 withheld amount and all that, so no need for new instructions.  It can be figured out first based on the Mint, and then double-checked on the length of the account data provided. So if I provide a `Mint2` and an account only with length for an `Account`, the instruction fails.  And the opposite fails too `Mint` with `Account2` length.

token/program-2022/src/state.rs
joncinque: This is why I initially had `TransferWithheld`, to make it clear that you can transfer out your own withheld tokens to the fee owner.  I'm not a huge fan of the Burn behavior since it's unclear.

On the other hand, if we store a withheld amount in the mint, then there's no reason to ever withhold amounts in the token accounts.  It's only useful if we want to support unchecked `Transfer` for `Account2`, which isn't possible.  That would resolve pretty much all of the withholding issues that have come up.  What do you think?

token/program-2022/src/state.rs
joncinque: That's exactly the idea! As you say, if `Mint3` does not contain all of the fields in `Mint2`, then `UberMint` has all the fields for Mint3 and Mint2, with sensible defaults.  This means a defi protocol only ever has to deal with `UberMint` to do its work.  Either way, you're right, we don't have to figure this out now since there's no `Mint3`.

It may be worth having conversations with teams to see what works best for them. I'm thinking through the case of a defi protocol saying, "ok cool, new mint type. What do I need to do to support it?" I'm seeing two main approaches:

* have an `unpack` function that returns a `Mint` regardless of what's inside, and another that returns a `Mint2` regardless of what's inside
* have an `unpack` function that returns a trait object / superstruct

Happy to hear any other ideas!

token/program-2022/src/state.rs
joncinque: Although I'm just realizing that `InitializeAccount2` is going to be really confusing if one of the types is called `Account2` :thinking:

token/program-2022/src/instruction.rs
joncinque: The idea of `TransferWithheld` was for the case of wanting to close an account with a withheld amount still stuck in it, but if we store all withheld amounts on the mint, then you're right: we'll only need a `HarvestFee` instruction that takes the mint and an account, signed by the withdraw authority.

token/program-2022/src/state.rs
mvines: Adding instructions is cheap and I feel doing so adds additional clarity to the caller, chain observer, and token program implementation.  

`InitializeAccount2` vs `Account2` is certainly a problem though.  `Account2`/`Mint2` probably need to find a better name to resolve this.  Not sure what that is yet, perhaps we just keep `AccountWithTransferFee`/`MintWithTransferFee` for now then, at which point there could be
`InitializeAccountWithTransferFee`/`InitializeMintWithTransferFee` instructions

token/program-2022/src/instruction.rs
mvines: > And we'll also need a way to move withheld tokens out, whether that's through Burn or another instruction.

I was thinking `Burn(0)` before `Close` could be used to move the withheld amount to the mint before closing.  But practically perhaps that's no different than a new variant of `Close` that accepts a writable mint account

token/program-2022/src/state.rs
mvines: I was thinking of essentially the `From` and `TryFrom` traits.  If one wants a specific `AccountType`, they use `From`.  `TryFrom` can be used if the actual `AccountType` can be converted to the desired type in a sensible way.

But I feel like we don't need to hammer out this detail yet, and I agree that discussing with external teams would be valuable

token/program-2022/src/state.rs
mvines: > I'm not a huge fan of the additional Burn behavior and would prefer it done with a separate instruction.

Yeah, a new `Close` variant is probably cleaner than `Burn(0)`

token/program-2022/src/instruction.rs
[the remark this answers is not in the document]
mvines: Yes good point about the sweep script. If the fee collected at the destination, one could grief the sweep script by moving the withheld amount around continually 😆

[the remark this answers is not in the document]
joncinque: hehe yeah I was thinking that too.  At least it would have to stop eventually since the fees will constantly be left as the amount is transferred around.

I'll change the comment to mention that the transfer fee goes to the destination, that makes more sense.

token/program-2022/src/state.rs
joncinque: Sounds good to me. I'll remove these structs in the meantime. I put them in to kickstart the discussion

token/program-2022/src/state.rs
joncinque: Is your concern that someone who's created a  `Mint2` won't be sure how to initialize accounts because they'll have done `InitializeMintWithTransferFee` and then they'll look for `InitializeAccountWithTransferFee`, not find it, and get confused?  As a corollary, does that mean `InitializeAccount[23]*` always fail for the new account type?

Separately, regarding the name, how about `PaymentMint` and `PaymentAccount` since these features are pretty specific to payment systems?

token/program-2022/src/instruction.rs
joncinque: Let's do a new `Close` instruction.  It would be more composable to have a separate instruction which just moves withheld funds to the mint, but it's probably more abusable / confusing.  We can call it `CloseAccount2`?

token/program-2022/src/state.rs
mvines: > Is your concern that someone who's created a Mint2 won't be sure how to initialize accounts because they'll have done InitializeMintWithTransferFee and then they'll look for InitializeAccountWithTransferFee, not find it, and get confused? 

Three concerns, depending on the context:
1. The client gets confused (or is buggy) and accidentally allocates an `Account`-sized account when they really want a `PaymentAccount`. The `InitializeAccount` works but then they are later confused because they got the wrong thing
2. From an explorer visibility perspective, it's not clear from looking at the Transaction itself what `InitializeAccount` did
3. The `InitializeAccount` implementation now needs more conditional logic based on user input, so more ways to potentially exploit the program.

> As a corollary, does that mean InitializeAccount[23]* always fail for the new account type?

Yes, that's what I was thinking.

> Separately, regarding the name, how about PaymentMint and PaymentAccount since these features are pretty specific to payment systems?

Those are the best names I've heard (or come up with) so far!

token/program-2022/src/instruction.rs
mvines: `ClosePaymentAccount`?

token/program-2022/src/state.rs
mvines: But to be clear, I'm not sure I love PaymentMint/PaymentAccount yet. But we can bike shed that later as well, since it's somewhat independent from the actual interface definition

joncinque: Ok great, thanks for the clarification

joncinque:
All right, I think this covers all of the feedback. Thanks for taking such a quick / close look at all this!

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +427,125 @@ pub enum TokenInstruction {
         /// The freeze authority/multisignature of the mint.
         freeze_authority: COption<Pubkey>,
     },
+    /// Initialize a new mint for payment systems.
```
mvines, line 430: 🤔 The word "payment" is wearing thin already: stables such as USDC and USDT would very much consider themselves to be payment rails and they do not require a transfer fee.

What I think we really want is a mixin concept, there are various independent/optional features that we want to be able to activate for an Account/Mint in the general Token 22 timeframe:
1. Transfer fee
2. Account creation authority (to be able to build allow/deny lists, and charge an account creation fee beyond rent)
3. Confidential token support.  Currently this is implemented as a PDA, but being able to build it directly into SPL Token would be a significant simplification
4. A joint Mint/Account type for NTFs (remove the need to have 1 Mint + 1 Account for each NTF)

What if the extended Mint/Account data was the base mint/account followed by TLV scheme that contained an arbitrary number of extensions.

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +427,125 @@ pub enum TokenInstruction {
...
+    ///   0. `[writable]` The mint to initialize.
+    InitializePaymentMint {
+        /// Number of base 10 digits to the right of the decimal place.
+        decimals: u8,
+        /// The authority/multisignature to mint tokens.
+        mint_authority: Pubkey,
+        /// The freeze authority/multisignature of the mint.
+        freeze_authority: COption<Pubkey>,
+        /// Pubkey that must own any provided fee account
+        fee_account_owner: Pubkey,
```
mvines, line 443: Both  `fee_config_authority` and `fee_withdraw_authority` are set to this pubkey?  Also `COption<Pubkey>` I think

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +427,125 @@ pub enum TokenInstruction {
...
+    TransferCheckedWithFee {
+        /// The amount of tokens to transfer.
+        amount: u64,
+        /// Expected number of base 10 digits to the right of the decimal place.
+        decimals: u8,
+        /// Amount of transfer collected as fees, expressed as basis points of the
+        /// transfer amount
+        transfer_fee_basis_points: u16,
+        /// Maximum fee assessed on transfers
+        maximum_fee: u64,
```
mvines, line 487: How about just `fee: u64`, and the instruction fails if the fee computed by the program doesn't match.  The client likely will compute the fee itself regardless for display to the user

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +427,125 @@ pub enum TokenInstruction {
...
+    /// Succeeds for frozen accounts.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[writable]` The token mint. Must be a PaymentMint.
+    ///   1. `[writable]` The fee receiver account. Must be a PaymentAccount
+    ///      associated with the provided PaymentMint.
+    ///   2. `[signer]` The mint's `fee_withdraw_authority`
+    ///   3. ..3+N `[writable]` The source accounts to harvest from. Must be PaymentAccounts.
```
mvines, line 501: `3..` is optional right?  That is, if no `PaymentAccounts` are provided then any withheld tokens in the token mint are still transferred to the fee receiver account

joncinque: Yep, I'll make that clearer

token/program-2022/src/instruction.rs
joncinque: Much better

token/program-2022/src/instruction.rs
joncinque: Bah thanks, missed that one. And I was thinking it would be weird to create a mint with transfer fees and not set the authorities, since they wouldn't be settable after that. There is something cool to having no config authority and a harvest authority, since then you know the fee will never change.  Works for me!

token/program-2022/src/instruction.rs
joncinque: It's certainly not a perfect name, but I'll take it over `Mint2`.  If you want to go ahead and call it `MixinMint` to start the framework concept, I'm game.

In that case, we might have to nix some of the `TransferFee`-specific instructions in favor of more general ones.  `2. Account creation authority` and `4. joint mint/account` seem pretty simple to tack on, but maybe we should have a chat about `3. Confidential token support` to make sure this design doesn't make it gross later.

Thanks for laying out all of the other features that we're looking for. A TLV scheme is probably our best bet to maintain all of these separate features all in the same account and expand continuously without breaking everyone anytime a new mint type comes out.

It's too bad, I was really hoping that some of those mixin features could get done in separate programs. I'll just hope that a mixin model will make it easier for people to write their own programs in the future :crossed_fingers:

mvines: > It's certainly not a perfect name, but I'll take it over Mint2.

No argument from me on that one! :)

> If you want to go ahead and call it MixinMint to start the framework concept, I'm game.

Here's an idea: we could just use the existing Account and Mint structs and then append TLV data at the end of each.  No need to define new Account/Mint/Multisig structs at all!

If the account data length is not 82, 165, or 355 (legacy Mint/Account, and Multisig accounts), the account data layout would be:
- 0..(81 or 164): existing Mint/Account struct
- 82..164: zero padding only present if this is a Mint.
- 165: a u8 that indicates the account type (Account = 1, Mint = 2).
- TLV entries...

Depending on the TLV entries present, the account data could still end up being 355 bytes long and be confused with a Multisig account. If this happens then a null TLV entry must also be added to ensure the data length never equals the Multisig account data length

> It's too bad, I was really hoping that some of those mixin features could get done in separate programs.

The ZkToken work is an example of exactly that.  It wasn't too painful honestly, but a couple things did feel quite unnatural:
a. One can close the primary token account and orphan the ZkToken account
b. All the ZkToken instructions need the primary token account to look up the owner (so more accounts to pass in/decode)
c. Frozen account handling logic needed to be duplicated in the ZkToken transfer processor

> but maybe we should have a chat about 3. Confidential token support to make sure this design doesn't make it gross later.

For sure.  I think it's pretty straightforward though:
1. For Mints, the TLV value would be the [ZkMint](https://github.com/solana-labs/spl-zk-token/blob/662d0b51105dd7b9cc6aae60f353bb8676f302ae/program/src/state.rs#L13) struct minus the `mint` field (since that's now implicit)
2. For Accounts, the TLV value would be the [ZkAccount](https://github.com/solana-labs/spl-zk-token/blob/662d0b51105dd7b9cc6aae60f353bb8676f302ae/program/src/state.rs#L58) struct minus the `mint` and `token_account` fields.
and then most of the [ZkTokenInstruction](https://github.com/solana-labs/spl-zk-token/blob/662d0b51105dd7b9cc6aae60f353bb8676f302ae/program/src/instruction.rs#L70)s end up moving into SPL Token as well.


The only major downside I currently see with a TLV approach is that it makes RPC `getProgramAccounts` `memcmp` searches harder (impossible?), for example [this](https://spl.solana.com/token#finding-all-token-accounts-for-a-specific-mint), since the byte offset for a particular TLV tag cannot be relied upon.  Maybe that's ok though?  The legacy Account/Mint fields remain searchable, even for extended accounts/mints

joncinque: > Here's an idea: we could just use the existing Account and Mint structs and then append TLV data at the end of each. No need to define new Account/Mint/Multisig structs at all!

I really really like this idea.  It resolves a lot of my hangups with how clients will use this, and all mixins can just have an unpack which tries to find and deserialize, returning an `Option`. In processors, if an instruction is impacted by a particular mixin, it just needs to extract that bit, perform whatever extra logic is needed, and you're good to go!

This means we won't need quite so many instructions in the end and some initializers that take lists of mixins + any required parameters.  I'll update this PR with some interfaces about how this would look.

> For Mints, the TLV value would be the ZkMint struct minus the mint field (since that's now implicit)
> For Accounts, the TLV value would be the ZkAccount struct minus the mint and token_account fields and then most of the ZkTokenInstructions end up moving into SPL Token as well.

Now that I've read through the program, you're absolutely right, this'll be really straightforward to include as a mixin.  My one question: do you still envision having `Deposit` / `Withdraw` logic to "convert" between zk and non-zk tokens?  It seems like that should be done through an external swapping program.

mvines: > My one question: do you still envision having Deposit / Withdraw logic to "convert" between zk and non-zk tokens? 

Yep, that's what I was thinking still.

> It seems like that should be done through an external swapping program.

I guess I don't see a benefit to externalizing that as it's still the same token whether it's on the normal or zk side.

joncinque: oh ok, so that means the ZK token mint would support both normal accounts and ZK accounts? and you would `deposit` from your normal account to your ZK account? or would the same account somehow handle both?

I thought it would be two separate tokens, one for just ZK and another for normal, and have a program in between to do swaps. The idea being that ZK tokens are just confidential wrapped tokens around the base ones.

joncinque:
I've added in the mixin concept as just structs and instructions, let me know if this is clear or needs more detail.

token/program-2022/src/instruction.rs
mvines: Yeah, I was still thinking the same account could handle both.  The zk tokens are confidential, not anonymous, so there doesn’t seem to be much benefit to making users create a separate token account to hold them since normal/zk sides share the same mint, supply, and account owner. 

When the user/wallet creates their token account, they’ll just need to decide if reserving space for the zk mixin is worth it for them (but the ATA program needs to change to support mixins regardless). If they change their mind later, the wallet would need to close and recreate their token account.

joncinque: Gotcha, wfm

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +427,126 @@ pub enum TokenInstruction {
...
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
+    /// bytes). If any mixins are provided, there must be 83 bytes of padding,
+    /// 1 byte reserved for the account type, then space required by each mixin.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint to initialize.
+    ///   1. ..1+N `[]` (Optional) Accounts required to initialize the mixins
+    InitializeMintWithMixins {
```
mvines, line 440: What if we only had `InitializeAccount`/`InitializeMint` still.  When the user wants to adds mixins then they just over-allocate additional account data

Each mixin then has its own initializer, like `InitializeTransferFee`, that follows `InitializeMint`.

Interestingly it seems that an Account doesn't even need an explicit `InitializeTransferFee`.  The Mint is passed into `InitializeAccount`, so if the Mint has a transfer fee initialized then the token program ought to initialize it implicitly for the Account (or fail if the provided account doesn't have enough data storage for the transfer fee mixin)

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +427,126 @@ pub enum TokenInstruction {
...
+    ///   2. `[writable]` The mint.
+    ///   3. `[signer]` The account's owner.
+    ///
+    ///   * Multisignature owner
+    ///   0. `[writable]` The account to close.
+    ///   1. `[writable]` The destination account.
+    ///   2. `[writable]` The mint.
+    ///   3. `[]` The account's multisignature owner.
+    ///   4. ..4+M `[signer]` M signer accounts.
+    CloseAccountWithTransferFee,
```
mvines, line 530: `CloseAccountWithTransferFee` isn't sitting well with me.  What if instead there was a permissionless version of `HarvestFee` that only accepted the writable Mint and writable list of source Accounts to harvest from.

To close an account the client then just needs to add a `HarvestFee` instruction before the standard `CloseAccount` instruction. Some future extension may impose additional/different requirements on account close and then we end up with a zoo of `CloseAccountWith...` variants.

token/program-2022/src/instruction.rs
joncinque: > What if we only had InitializeAccount/InitializeMint still. When the user wants to adds mixins then they just over-allocate additional account data

That was my initial idea, but then I was worried about ensuring that everything is done in one transaction.  If an account requires `InitializeTransferFee`, how are we sure that it's done in the same transaction?  That requires `InitializeMint` to take in the instructions sysvar and check that the correct accounts are passed in, which adds more compute unit usage and complexity.  I worry about undefined behavior / attack opportunities otherwise.  The alternative is to have `InitializeMint` done last, and mixin initializers (ie. `InitializeTransferFee`) fail if any data has been written in the `Mint` part.  But then I worry that it might not be usable if any data from the `Mint` is required when initializing a mixin.

> Interestingly it seems that an Account doesn't even need an explicit InitializeTransferFee. The Mint is passed into InitializeAccount, so if the Mint has a transfer fee initialized then the token program ought to initialize it implicitly for the Account (or fail if the provided account doesn't have enough data storage for the transfer fee mixin)

Absolutely! This was also my initial idea, but you mentioned earlier that you wanted to be explicit about everything, so I put in the explicit instruction.  I'd prefer to just use `InitializeAccount` and a helper function on the mint to give the size of an Account.  That way, things like ATA need only tiny changes.

token/program-2022/src/instruction.rs
joncinque: Haha this was my initial idea too with `TransferWithheld`!  I am worried about a totally permissionless version of `HarvestFee` -- giving anyone the possibility to move other people's tokens doesn't make me feel great.  But as long as the only possible destination is the Mint, I think it should limit the attack possibilities.

Happy to do this, we just need a good name... `TransferWithheldFeeToMint`? `HarvestFeeToMint`? `ConsolidateFees`?

token/program-2022/src/instruction.rs
mvines: > but then I was worried about ensuring that everything is done in one transaction 
> ... 
> any data from the Mint is required when initializing a mixin.

Ok I'm convinced to keep it as for now and we can see how the fully implementation feels

>  I'd prefer to just use InitializeAccount and a helper function on the mint to give the size of an Account. That way, things like ATA need only tiny changes.

Heh, ok deal.  I do think there could be future mixins that do still require Account-level initialization of course

token/program-2022/src/instruction.rs
```diff
@@ -692,11 +836,45 @@ impl AuthorityType {
...
 }
 
+/// Specifies the authority type for SetAuthority instructions
+#[derive(Clone, Debug, PartialEq)]
+pub enum InitializeMintMixinParameters {
+    /// Additional padding required, in case the size of the account is the same
+    /// as a multisig
+    None {
+        padding: u64,
+    },
```
mvines, line 853: I think just the tag name itself should be sufficient padding
```suggestion
    None,
```

token/program-2022/src/instruction.rs
```diff
@@ -692,11 +836,45 @@ impl AuthorityType {
...
+        fee_withdraw_authority: COption<Pubkey>,
+        /// Amount of transfer collected as fees, expressed as basis points of the
+        /// transfer amount
+        transfer_fee_basis_points: u16,
+        /// Maximum fee assessed on transfers
+        maximum_fee: u64,
+    },
+}
+
+/// Specifies the authority type for SetAuthority instructions
```
mvines, line 868: (same)

token/program-2022/src/state.rs
```diff
@@ -10,6 +10,45 @@ use solana_program::{
     pubkey::Pubkey,
 };
 
+/// Different kinds of accounts. Note that `Mint` and `Account` types are determined
```
mvines, line 13:
```suggestion
/// Different kinds of accounts. Note that `Mint`, `Account` and `Multisig` types are determined
```

token/program-2022/src/instruction.rs
```diff
@@ -692,11 +836,45 @@ impl AuthorityType {
...
+        fee_config_authority: COption<Pubkey>,
+        /// Harvest instructions must be signed by this key
+        fee_withdraw_authority: COption<Pubkey>,
+        /// Amount of transfer collected as fees, expressed as basis points of the
+        /// transfer amount
+        transfer_fee_basis_points: u16,
+        /// Maximum fee assessed on transfers
+        maximum_fee: u64,
+    },
+}
```
mvines, line 866: Just for context, I immediately see a need to add a Mint-level mixin:
   `CloseAccount { close_authory:  COption<Pubkey> }`,
which will permit a `Mint` account to be closed via `CloseAccount` when the supply is 0.  This can be used to reclaim future NFT mints (cc: @bartosz-lipinski)

token/program-2022/src/state.rs
```diff
@@ -80,6 +103,55 @@ impl Pack for Mint {
...
+    /// account type, always set to "MintWithTransferFee"
+    pub account_type: AccountType,
+    /// Any provided fee account must be owned by this key. If set to `None`, then
+    /// this mint behaves like a normal `Mint`.
+    pub fee_account_owner: COption<Pubkey>,
+    /// Amount of transfer collected as fees, expressed as basis points of the
+    /// transfer amount, ie.  increments of 0.01%
+    pub transfer_fee_basis_points: u16,
+    /// Maximum fee assessed on transfers, expressed as an amount of tokens
+    pub maximum_fee: u64,
```
t-nelson, line 137: Proposal to epoch-lock the fee before we run into the vote-account commission problem

joncinque: :pray: thank you.  I can't even tell you how disappointed I am in myself for forgetting this.

token/program-2022/src/instruction.rs
mvines: I don't think so, seems like it's just a mint-level setting

token/program-2022/src/instruction.rs
joncinque: I went with `HarvestFeeToMint` and `HarvestFeeFromMint`

joncinque:
Ok, one more round of improvements!

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +431,148 @@ pub enum TokenInstruction {
...
+    /// `InitializeMint`.
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
+    /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
+    /// then space required for the mixin.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint to initialize.
+    InitializeMintTransferFeeMixin {
```
mvines, line 483:
```suggestion
    InitializeMintTransferFee {
```

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +431,148 @@ pub enum TokenInstruction {
...
+    ///
+    /// Fails if the account has already been initialized, so must be called before
+    /// `InitializeAccount`. The account must have enough data allocated for all
+    /// mixins included in the mint.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]`  The account to initialize.
+    ///   1. `[]` The mint this account will be associated with.
+    InitializeAccountTransferFeeMixin,
```
mvines, line 504:
```suggestion
    InitializeAccountTransferFee,
```

token/program-2022/src/instruction.rs
```diff
@@ -251,6 +264,9 @@ pub enum TokenInstruction {
     /// decimals value is checked by the caller.  This may be useful when
     /// creating transactions offline or within a hardware wallet.
     ///
+    /// If the accounts contain `AccountTransferFeeMixin`s, the fee is withheld in the
+    /// destination account.
```
mvines, line 268: I thought we landed on the fee always being withheld in the source account?
```suggestion
    /// source account.
```

token/program-2022/src/instruction.rs
```diff
@@ -54,6 +56,8 @@ pub enum TokenInstruction {
     /// Otherwise another party can acquire ownership of the uninitialized
     /// account.
     ///
+    /// Any mixins must be initialized before this instruction.
```
mvines, line 59: I'm a little squeamish about this initialization flow opening up new attack vectors, for both Mint and Accounts, but maybe it's ok.  The AccountType will still be uninitialized as mixins are getting initialized, so:
a. Mixin initialization must always fail if the account length is 355 (a possible multisig account), or less than 165 (legacy mint account)
b. My comment about "A single mixin type namespace for mints/accounts seems ok to me" probably needs to be stronger. There must be a single mixin type namespace: if I initialize an Account mixin for a Mint, then the Account mixin must be ignored.

mvines (commented):
This is feeling really good.  I have moar comments but I think we're very close now!

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +431,148 @@ pub enum TokenInstruction {
         /// The freeze authority/multisignature of the mint.
         freeze_authority: COption<Pubkey>,
     },
+    /// Gets the required size of an account for the given mint as a little-endian
+    /// `u64`.
```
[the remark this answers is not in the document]
joncinque: It makes for a clearer experience from all platforms, since we always give `space` as a `u64`, ie https://github.com/solana-labs/solana/blob/293629bdb581dd101fb286b73db5615dc3993911/sdk/program/src/system_instruction.rs#L313 -- and it seems fine to use up a `u64` if it's just return data

token/program-2022/src/instruction.rs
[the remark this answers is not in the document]
joncinque: Ah ok, but then for the other ones, ie. ZKToken, that will have an explicit initializer, correct?

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +431,148 @@ pub enum TokenInstruction {
...
+    /// padding.
+    ///
+    /// Fails if the account has already been initialized, so must be called before
+    /// `InitializeAccount`. The account must have enough data allocated for all
+    /// mixins included in the mint.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]`  The account to initialize.
+    InitializeEmptyMixin,
```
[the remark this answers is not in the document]
joncinque: If we do it all in `InitializeAccount`, then yes no need for this

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +431,148 @@ pub enum TokenInstruction {
...
+    /// `InitializeMint`.
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
+    /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
+    /// then space required for the mixin.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint to initialize.
+    InitializeMintCloseMixin,
```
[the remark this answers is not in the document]
joncinque: Oops, yes absolutely

token/program-2022/src/state.rs
```diff
@@ -80,6 +127,53 @@ impl Pack for Mint {
     }
 }
 
+/// Type-Length-Value Entry, used to encapsulate all mixins contained within an account
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct TlvEntry<T, V> {
+    pub account_type: T,
+    pub length: u64, // usize? or u32?
```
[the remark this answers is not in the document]
joncinque: let's do `u32` to be safe

token/program-2022/src/state.rs
```diff
@@ -80,6 +127,53 @@ impl Pack for Mint {
     }
 }
 
+/// Type-Length-Value Entry, used to encapsulate all mixins contained within an account
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct TlvEntry<T, V> {
+    pub account_type: T,
```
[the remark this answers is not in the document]
joncinque: That's fair, I preferred to separate them mainly because the names were annoying as one enum, ie,
```
pub enum MixinType {
    MintTransferFee,
    AccountTransferFee,
    MintCloseAuthority,
}
```
And also, I thought it would be clearer / easier to reason in the code when iterating through the TLV entries.  I'd prefer to keep it for now, and if we need to combine later, it's not a huge deal.

token/program-2022/src/state.rs
```diff
@@ -80,6 +127,53 @@ impl Pack for Mint {
...
+    pub fee_config_authority: COption<Pubkey>,
+    /// Any harvest instruction must be signed by this key. If set to `None`, then
+    /// this mint behaves like a normal `Mint`.
+    pub fee_withdraw_authority: COption<Pubkey>,
+    /// Withheld transfer fee tokens that have been moved to the mint for harvesting
+    pub withheld_amount: u64,
+    /// Old transfer fee, used if the current epoch < new_transfer_fee.epoch
+    pub old_transfer_fee: TransferFee,
+    /// New transfer fee, used if the current epoch >= new_transfer_fee.epoch
+    pub new_transfer_fee: TransferFee,
```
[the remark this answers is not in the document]
joncinque: I had that initially, and then thought it would be annoying to have `transfer_fee[0]` and `transfer_fee[1]` all over the place, when it's totally clear which one is newer and older with explicit names.  Maybe we just need a better name? `older_transfer_fee` and `newer_transfer_fee`? `current_transfer_fee` and `next_transfer_fee`? `current` and `upcoming`? `now` and `later`? I don't like any of these, but `older` and `newer` may be clearest

token/program-2022/src/state.rs
```diff
@@ -4,12 +4,59 @@ use crate::instruction::MAX_SIGNERS;
...
+/// the account data.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum AccountType {
+    /// Nothing, uninitialized
+    Uninitialized,
+    /// Base mint
+    Mint,
+    /// Base token account
+    Account,
```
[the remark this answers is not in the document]
joncinque: :eyes: https://github.com/solana-labs/solana-program-library/blob/0b8961597b3adf7355e48506d7e81b3925bbacd0/token/program-2022/src/lib.rs#L1

Is that worth an `#[allow(missing_docs)]` then?

token/program-2022/src/instruction.rs
joncinque: We went back and forth about it.  Destination is probably better in the end for the clearer closing behavior.  For example, if my account was recently harvested, and I haven't received any transfers since, I can do `transfer` + `close` in one instruction without the permissionless harvest.  What do you think?

token/program-2022/src/instruction.rs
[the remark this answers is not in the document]
joncinque: and for mints, `InitializeMint` will have to do that check last then to sum up all the mixins and make sure the last byte is empty. I initially thought it would be best to write some data to that last byte, but you're right, it doesn't make a difference as long as the empty mixin is always last and implicit

token/program-2022/src/instruction.rs
joncinque: I'm with you on the attacks, but my thinking is that as long as the mint isn't properly initialized, it shouldn't be able to do damage.  For example, if I only initialize a bunch of mixins on a mint, but never do the proper `InitializeMint`, I can't create any token accounts on it.  We'll have to diligently check that the mint is initialized everywhere, but that's a given to me.

Otherwise, you're right, let's put it all in a single mixin namespace.  For now, all of the account mixins are initialized in `InitializeAccount`, so we should be ok from that front, since it's impossible to explicitly initialize an account mixin on its own, but we don't know if that'll hold forever.

token/program-2022/src/instruction.rs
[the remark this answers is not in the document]
mvines: Yes.  In the ZkToken case it's likely that this'll often happen after account initialization though.  Specifically:
1. ATA program creates an account for the wallet via some on-ramp
2. User's wallet later initializes the ZkToken mixin and provides the user's ElGamal pubkey. `InitializeZkToken` internally performs a realloc to make room for the ZkToken mixin, since the ATA program wants to pay as little as possible to initialize the user's account.

token/program-2022/src/instruction.rs
mvines: I agree destination is nicer for the normal case, I would prefer it. 

I thought you had a pretty good argument about griefing the fee collector but moving the fee around continually.  But with `HarvestFeeToMint` the situation seems better.  The fee collector script, in a loop, would:
1. Fetch all known token accounts with a zero withheld tokens 
2. Batch them into calls to `HarvestFeeToMint`
so they'd reclaim the fee eventually!

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +431,148 @@ pub enum TokenInstruction {
...
+    /// Permissionless instruction to transfer all withheld tokens to the mint.
+    ///
+    /// Succeeds for frozen accounts.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint.
+    ///   1. ..1+N `[writable]` The source accounts to harvest from.
+    ///      Must include the `AccountTransferFeeMixin`.
+    HarvestFeeToMint,
```
mvines, line 536: maybe?
```suggestion
    HarvestWithheldTokensToMint,
```

token/program-2022/src/state.rs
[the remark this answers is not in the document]
mvines: Oh, ok.  Fine

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +434,125 @@ pub enum TokenInstruction {
...
+    /// `u64`.
+    ///
+    /// Return data can be fetched using `sol_get_return_data` and deserializing
+    /// the return data as a little-endian `u64`.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[]` The mint to calculate for
+    GetAccountDataSize,
+    /// Initialize the close account mixin on a new mint.
```
mvines, line 447:
```suggestion
    /// Initialize the close account authority on a new mint.
```

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +434,125 @@ pub enum TokenInstruction {
...
+    /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
+    /// then space required for the mixin.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint to initialize.
+    InitializeMintCloseAuthority {
+        close_authority: COption<Pubkey>,
+    },
+    /// Initialize the transfer fee mixin on a new mint.
```
mvines, line 462:
```suggestion
    /// Initialize the transfer fee on a new mint.
```

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +434,125 @@ pub enum TokenInstruction {
...
+    TransferCheckedWithFee {
+        /// The amount of tokens to transfer.
+        amount: u64,
+        /// Expected number of base 10 digits to the right of the decimal place.
+        decimals: u8,
+        /// Expected fee assessed on this transfer, calculated off-chain based on
+        /// the transfer_fee_basis_points and maximum_fee of the mint.
+        fee: u64,
+    },
+    /// Transfer all withheld tokens to a fee account. Signed by the mint's
```
mvines, line 510:
```suggestion
    /// Transfer all withheld tokens to an account. Signed by the mint's
```

token/program-2022/src/instruction.rs
```diff
@@ -674,6 +818,10 @@ pub enum AuthorityType {
     AccountOwner,
     /// Authority to close a token account
     CloseAccount,
+    /// Authority to set the fee
+    FeeConfig,
```
mvines, line 822:
```suggestion
    /// Authority to set the transfer fee
    TransferFeeConfig,
```

token/program-2022/src/instruction.rs
```diff
@@ -674,6 +818,10 @@ pub enum AuthorityType {
...
+    /// Authority to withdraw fees from accounts and mint
+    FeeWithdraw,
```
mvines, line 824:
```suggestion
    /// Authority to withdraw withheld tokens from a mint
    WithheldWithdraw,
```

token/program-2022/src/instruction.rs
mvines, line 822:
```suggestion
    /// Authority to set the transfer fee
    TransferFeeConfig,
```

token/program-2022/src/state.rs
```diff
@@ -4,12 +4,50 @@ use crate::instruction::MAX_SIGNERS;
...
+}
+
+impl Default for AccountType {
+    fn default() -> Self {
+        Self::Uninitialized
+    }
+}
+
+/// Mixins that can be applied to mints or accounts.  Mint mixins must only be
```
mvines, line 35: Ok!  Time to open this can of worms since we're pretty much all done with the important stuff.  Do we want to keep the term "Mixin" or use something more watered down like "Attribute", "Trait", or "Extension"?

mvines (commented):
moar review!!1!

token/program-2022/src/instruction.rs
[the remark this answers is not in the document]
joncinque: Sweet, sounds good to me

token/program-2022/src/instruction.rs
joncinque: To be clear, I don't think the whole wittheld amount should be sent over, just that on one transfer, the wittheld amount stays in the destination.  For example, if:
* transfer fee is 1%
* my account has 10 tokens withheld,
* yours has 3 tokens withheld

I send you 100 tokens. Afterwards:

* my account still has 10 tokens withheld total
* your account receives 99 tokens to your balance
* your withheld amount goes up by 1, to 4

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +431,148 @@ pub enum TokenInstruction {
...
+    ///   3. ..3+M `[signer]` M signer accounts.
+    HarvestFeeFromMint,
+    /// Permissionless instruction to transfer all withheld tokens to the mint.
+    ///
+    /// Succeeds for frozen accounts.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint.
+    ///   1. ..1+N `[writable]` The source accounts to harvest from.
```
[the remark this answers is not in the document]
joncinque: Sounds good

token/program-2022/src/state.rs
joncinque: I've been liking `Attribute` most, and `Extension` is good too.  `Trait` is a loaded word that we may want to avoid.  `Feature` is the only word I can come up with, but that's also loaded.

People seem to have a bizarre relationship with the word mixin. Ex-Django devs?

CriesofCarrots: I've only been following this PR lightly, but fwiw I like Trait or Extension for this.

mvines: Welp that leaves only Extension in the venn diagram

joncinque: Works for me!

joncinque:
One more round!

token/program-2022/src/instruction.rs
```diff
@@ -85,11 +89,17 @@ pub enum TokenInstruction {
...
     },
+    /// NOTE This instruction is deprecated in favor of `TransferChecked` or
+    /// `TransferCheckedWithFee`
+    ///
     /// Transfers tokens from one account to another either directly or via a
     /// delegate.  If this account is associated with the native mint then equal
     /// amounts of SOL and Tokens will be transferred to the destination
     /// account.
     ///
+    /// If the accounts contain `AccountTransferFeeExtension`s, this will fail.  Mints with
```
mvines, line 100:
```suggestion
    /// If either accounts contains an `AccountTransferFee` extension, this will fail.  Mints with
```

That is, can we just drop the `Extension` from the structs in general?  Contrast with `MintCloseAuthority

token/program-2022/src/instruction.rs
```diff
@@ -29,6 +29,8 @@ pub enum TokenInstruction {
     /// Otherwise another party can acquire ownership of the uninitialized
     /// account.
     ///
+    /// Any extensions must be initialized before this instruction.
```
mvines, line 32: nit: I think this comment and the similar ones for the `InitializeAccount` family could be dropped.  This is all true with the current set of extensions but may not be the case going forward.  More generally some extensions will need to be initialized prior to `InitializeMint`/`InitializeAccount` and some can be done later.

mvines (approved):
I'm just down to pedantic nits, which means :shipit:!  This is all feeling pretty tidy to me

@CriesofCarrots - it'd be great if you're able to give this a final once over as well

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +434,127 @@ pub enum TokenInstruction {
...
+    ///   1. `[writable]` The fee receiver account. Must include the `AccountTransferFeeExtension`.
+    ///      associated with the provided mint.
+    ///   2. `[signer]` The mint's `fee_withdraw_authority`
+    ///
+    ///   * Multisignature owner/delegate
+    ///   0. `[writable]` The token mint.
+    ///   1. `[writable]` The destination account.
+    ///   2. `[]` The mint's `fee_withdraw_authority`'s multisignature owner/delegate.
+    ///   3. ..3+M `[signer]` M signer accounts.
+    HarvestWithheldTokensFromMint,
```
CriesofCarrots, line 526: I got confused here. Based on the doc and instruction name -- esp the word "Harvest" -- I thought this instruction was the same as `HarvestWithheldTokensToMint`, except it sends the tokens to an account instead of the mint. But since no source accounts are provided, I must be wrong.

Perhaps this should be `WithdrawWithheldTokensFromMint`?

token/program-2022/src/state.rs
```diff
@@ -4,12 +4,50 @@ use crate::instruction::MAX_SIGNERS;
...
+    clock::Epoch,
     program_error::ProgramError,
     program_option::COption,
     program_pack::{IsInitialized, Pack, Sealed},
     pubkey::Pubkey,
 };
 
+/// Different kinds of accounts. Note that `Mint`, `Account`, and `Multisig` types
+/// are determined exclusively by the size of the account, and are not included in
+/// the account data. These `AccountType`s are only included if extensions have been
```
CriesofCarrots, line 16:
```suggestion
/// the account data. `AccountType` is only included if extensions have been
```

token/program-2022/src/state.rs
```diff
@@ -4,12 +4,50 @@ use crate::instruction::MAX_SIGNERS;
...
+}
+
+impl Default for AccountType {
+    fn default() -> Self {
+        Self::Uninitialized
+    }
+}
+
+/// Extensions that can be applied to mints or accounts.  Mint extensions must only be
+/// applied to mint accounts, and account extensions must only be applied to holding
```
CriesofCarrots, line 36:
```suggestion
/// applied to mint accounts, and account extensions must only be applied to token holding
```

token/program-2022/src/state.rs
```diff
@@ -80,6 +118,52 @@ impl Pack for Mint {
...
+}
+
+/// Transfer fee information
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct TransferFee {
+    /// First epoch where the transfer fee takes effect
+    pub epoch: Epoch,
+    /// Amount of transfer collected as fees, expressed as basis points of the
+    /// transfer amount, ie.  increments of 0.01%
```
CriesofCarrots, line 145:
```suggestion
    /// transfer amount, ie. increments of 0.01%
```

token/program-2022/src/state.rs
```diff
@@ -80,6 +118,52 @@ impl Pack for Mint {
...
+    pub maximum_fee: u64,
+}
+
+/// Transfer fee extension data for mints.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq)]
+pub struct MintTransferFeeExtension {
+    /// Optional authority to set the fee
+    pub transfer_fee_config_authority: COption<Pubkey>,
+    /// Harvest from mint instructions must be signed by this key
```
CriesofCarrots, line 157:
```suggestion
    /// Withdrawal from mint instructions must be signed by this key
```
See my earlier comment. I think it's confusing to call this step harvesting sometimes.

CriesofCarrots (commented):
Awesome, looks good. I think the extension approach works well, and I didn't catch any fundamental issues with the design for fees and mint-closing; just added a few small comments/questions. (Apologies if some/all came up in previous threads! I wish I'd been following more closely)

token/program-2022/src/instruction.rs
```diff
@@ -54,6 +56,8 @@ pub enum TokenInstruction {
     /// Otherwise another party can acquire ownership of the uninitialized
     /// account.
     ///
+    /// All required extensions will be initialized at the same time.
```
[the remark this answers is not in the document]
joncinque: Maybe "are initialized" would be clearest. The idea was to convey that extensions are initialized during this part.  Either way, Michael nitted that comments are not necessary, so the code will speak for itself!

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +434,127 @@ pub enum TokenInstruction {
...
+    ///
+    ///   0. `[]` The mint to calculate for
+    GetAccountDataSize,
+    /// Initialize the close account authority on a new mint.
+    ///
+    /// Fails if the mint has already been initialized, so must be called before
+    /// `InitializeMint`.
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
+    /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
```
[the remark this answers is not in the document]
joncinque: Yes, exactly

token/program-2022/src/instruction.rs
joncinque: That makes a lot of sense, we use the term `WithdrawWithheld` everywhere else.  The idea is there's one instruction to harvest from accounts to the mint, and another to withdraw from the mint to an account.

For better parallelization, we should probably also add a `Withdraw` instruction to withdraw straight from a bunch of accounts and into another one, which wouldn't require a writable mint.  What do you think?

token/program-2022/src/state.rs
joncinque: One of these days we'll come up with a better term than "account"...

token/program-2022/src/state.rs
joncinque: Definitely, the term `Harvest` is now only in the instruction to move withheld tokens from accounts to the mint

token/program-2022/src/instruction.rs
mvines: Yeah I like that idea. The withdraw to mint operation is really just to facilitate emptying all the withheld tokens before closing an account.  If we can assert that an account with 0 tokens will always have 0 withheld tokens then we could remove withdrawing to mint entirely

joncinque: Done.

>  If we can assert that an account with 0 tokens will always have 0 withheld tokens then we could remove withdrawing to mint entirely

I can't come up with a model that guarantees this :thinking:

CriesofCarrots: Excellent, these changes allay all my confusion :)

> I can't come up with a model that guarantees this 🤔

I was going to suggest moving the sender's withheld_amount to the destination account, but I think I just went through the same thought process you two did [here](https://github.com/solana-labs/solana-program-library/pull/2608#discussion_r760681610)

token/program-2022/src/instruction.rs
```diff
@@ -409,6 +428,144 @@ pub enum TokenInstruction {
...
+    ///   3. `[]` The source account's multisignature owner/delegate.
+    ///   4. ..4+M `[signer]` M signer accounts.
+    TransferCheckedWithFee {
+        /// The amount of tokens to transfer.
+        amount: u64,
+        /// Expected number of base 10 digits to the right of the decimal place.
+        decimals: u8,
+        /// Expected fee assessed on this transfer, calculated off-chain based on
+        /// the transfer_fee_basis_points and maximum_fee of the mint.
+        fee: u64,
```
mvines, line 500: If the mint doesn't include an `MintTransferFee` extension, should we still allow this instruction to be used but the `fee` must be 0?

mvines (approved):

CriesofCarrots (dismissed):
lgtm, when clippy is happy!
Sorry about the `#[allow(deprecated)]`s needed

token/program-2022/src/instruction.rs
CriesofCarrots: Seems reasonable to me

joncinque:
Merging this round of changes so we have a base to iterate on, thanks for reviews!

PR #2615, opened 2021-12-03
name-service: allow parent owner to transfer subdomains

dr497:


mvines:
lgtm, this PR just needs a `cargo fmt` run to make CI happen

PR #2616, opened 2021-12-03
docs: token: Added timelock program

imprfekt:
Added another solution for SPL token vesting, with additional customizations, after the review by @joncinque. :)


imprfekt:
cc @joncinque as we discussed

joncinque (approved):
Review happened offline -- looks good!

PR #2624, opened 2021-12-08
Migrate to @solana/buffer-layout

### LINKED ISSUE #2534: Migrate to @solana/buffer-layout
The `buffer-layout` package used here as a dependency lacks proper typing, which causes issues with downstream libraries depending on the token program TypeScript code.

Solana Labs maintains a buffer-layout implementation with types already:
https://github.com/solana-labs/buffer-layout

I suggest to change the dependency to that implementation.

abhimanyusinghgaur:
Fixes #2534.


joncinque:
Thanks for the contribution, looks good!

PR #2633, opened 2021-12-13
token-lending: Migrate to `@solana/buffer-layout`

### LINKED ISSUE #2534: Migrate to @solana/buffer-layout

abhimanyusinghgaur:
Fixes #2534.

### DIFF
--- a/token-lending/js/package.json
+++ b/token-lending/js/package.json
@@ -31,7 +31,7 @@
     "dependencies": {
         "bigint-buffer": "^1.1.5",
         "bignumber.js": "^9.0.1",
-        "buffer-layout": "^1.2.1"
+        "@solana/buffer-layout": "^3.0.0"
     },
     "peerDependencies": {
         "@solana/spl-token": "^0.1.6",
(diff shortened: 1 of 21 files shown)


jordansexton:
Looks like there's a lot of overlap with #2634. I think it may be better to get https://github.com/solana-labs/buffer-layout/pull/1 merged first and then use https://github.com/solana-labs/buffer-layout-utils/ so we don't have to cast things and we can DRY this up.

jordansexton:
Both `@solana/buffer-layout` and `@solana/buffer-layout-utils` have been released.

@abhimanyusinghgaur & @israelidanny, it may be worth figuring out how to combine your efforts on this one.

israelidanny:
@jordansexton from a quick read of this PR, it mostly does the same thing I did in https://github.com/solana-labs/solana-program-library/pull/2634 minus generics. If https://github.com/solana-labs/solana-program-library/pull/2634 is pushed, I think it will cover everything this PR is designed to contribute.

abhimanyusinghgaur:
Thanks @israelidanny. Closing this PR.

PR #2641, opened 2021-12-16
Prepare Token TypeScript client for release

jordansexton:
This PR is to prepare the spl-token TypeScript rewrite from https://github.com/solana-labs/solana-program-library/pull/2539 for release.

It's currently lacking a build system and expects library consumers to use ESM with transpilation, which can be annoying and lead to adoption issues.

We should still make sure it's documented well (https://github.com/solana-labs/solana-program-library/pull/2575) before publishing on NPM.

There will be a lot of users trying to use the old Token class API from tutorials they find, so we'll probably need to provide a migration guide, and maybe roll the new version out under `@solana/spl-token@next` or something for a while.


token/ts/package.json
```diff
@@ -6,19 +6,34 @@
...
-    "main": "lib/index.js",
+    "main": "lib/index.cjs.js",
+    "module": "lib/index.esm.js",
     "esnext": "lib/index.js",
     "types": "lib/index.d.ts",
+    "browser": {
+        "./lib/index.cjs.js": "./lib/index.browser.cjs.js",
+        "./lib/index.esm.js": "./lib/index.browser.esm.js"
+    },
+    "browserslist": [
```
0xCryptoSheik, line 29: As far as I know, and as sad as it is, `rollup` & `@rollup/plugin-typescript` do not take `browserslist` into consideration. `babel` with `@babel/preset-env` do.

You can double check by bundling with and without this configuration and verify that the build outputs are the same.

jordansexton: Ah, good point. I just lifted that from somewhere (probably web3.js). How much does it matter, is it worth adding babel?

token/ts/package.json
```diff
@@ -6,19 +6,34 @@
...
+        "defaults",
+        "not IE 11",
+        "maintained node versions"
+    ],
     "scripts": {
         "clean": "shx rm -rf lib",
-        "build": "yarn clean && tsc",
+        "build": "yarn clean && tsc && rollup -c",
```
0xCryptoSheik, line 36: I invite you to consider not bundling the library, but just to transpile it to the typescript compilation target of your choosing (according to the presumably unused `browserslist` config above, `es5` or `es2015`, although I believe `es2017` is a great & safe target for such a library - end users in the crypto space are heavily incentivized to keep an up to date modern browser, also consumers can always re-compile, which some frameworks do by default).

Indeed, `@solana/spl-token` application consumers **are** bundlers, they will bundle it anyway, might as well allow them to retain some level of control on how it is bundled for their application.

Not bundling also brings in a number of advantages, such as:
- allow consumers to hack directly in `node_modules` and perhaps patch the library themselves locally if needed using tools such as [patch-package](https://www.npmjs.com/package/patch-package)
- improved readability
- simpler build process
- less dependencies related maintenance overhead by Solana maintainers

Seeing the `nodePolyfills()` bit in the rollup configuration, I'm guessing bundling is a way to hide these node.js only dependencies to the end consumer.
I can understand this but I believe it's not the right way, with appropriate documentation, it is trivial for the consumer to configure their bundler to include or not a polyfill for a Node.js API depending on their use case.

Providing 1-1 source/transpiled files also can help with tree-shaking, though that doesn't make a difference if the library is really side-effects free (as opposed to just the `sideEffects` hint in `package.json`.

If you may consider not bundling & just transpiling, please also consider transpiling using Babel rather than TypeScript / `tsc`, even though the cost of maintenance is higher, the control over the compilation output is much better (such as using `browserslist` with `@babel/preset-env`, which will result in smaller bundles, compiled as close as possible to the defined browser targets).

token/ts/package.json
0xCryptoSheik: It depends. It's a trade-off between complexity leading to increased maintenance cost over time & output quality, especially in regard to transforming & (not) polyfill-ing the output as close as possible to the defined browser targets.

Some TS features are also not supported by [`@babel/plugin-transform-typescript`](https://babeljs.io/docs/en/babel-plugin-transform-typescript/)

token/ts/package.json
jordansexton: I can't say I know what the best course here is. As much as we can try to document what they should do, lots of devs don't seem to know how to configure their bundler, and there are so many different ones with different behaviors that its a struggle to maintain.

In general I'm a fan of using ESM and not bundling/polyfilling anything, but I've seen in the wallet-adapter repo that this can lead to a lot of confusion for Next.js/Webpack users, and probably in this case non-ESM Node.js users.

> they will bundle it anyway, might as well allow them to retain some level of control on how it is bundled for their application.

Isn't this possible if their bundler uses the tsc-transpiled JS here? https://github.com/solana-labs/solana-program-library/blob/a7856b7e4e199cf47ee1625fed1b4b6434d2e2dd/token/ts/package.json#L23

I have no idea if various bundlers actually make sense of [this field](https://2ality.com/2017/06/pkg-esnext.html), or how they handle the `module`, `main`, or `browser` fields or that matter.

> please also consider transpiling using Babel rather than TypeScript / tsc

I'll try this.

Ayunk12 (approved):

token/ts/package.json
cogoo: If I can add my two lamports to this. I echo similar sentiments to @0xCryptoSheik that this should remain unbundled. I'm impartial on the use of babel, but you could use TS to produce transpiled (ES6 should be a safe target) ESM and CommonJS, by having two tsconfigs. a `tsconfig.json` and `tsconfig-cjs.json`

`tsconfig-cjs.json` would be something like:

```typescript
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "module": "CommonJS",
    "outDir": "./lib/cjs"
  },
}
```

Then you build with `tsc -p tsconfig.json && tsc -p tsconfig-cjs.json`

jordansexton:
Thanks @cogoo for updating the build. We don't bundle it now, and we'll ship ESM/ESNext and CJS/ES6 versions.

token/ts/src/instructions/approve.ts
```diff
@@ -50,3 +60,44 @@ export function createApproveInstruction(
...
+
+    return {
+        instruction: data.instruction,
+        account,
+        delegate,
+        owner,
+        multiSigners,
+        amount: data.amount,
+    };
+}
```
jordansexton, line 103: @joncinque can I get your thoughts on this? Decoding token instructions to validate them is a need for a few things I'm building, but I want to see if there's a better way to think about doing this.

I'm especially wondering if we should validate the existence account keys at all, and if we should try validate them to be correct in the client side (for example, checking for isWritable and isSigner, and checking for multiSigners if the owner/authority is not a signer).

An argument for is that it would allow someone to generally verify whether an instruction was likely to succeed or fail before simulating it, and it lets us return stricter types from the decode functions. An argument against is that an instruction missing or having the wrong keys is still a valid type of that instruction, even if it will fail.

joncinque: Since I'm not sure where these decodes are ever used, you can define the use-case entirely.  For example, the explorer decodes things in its own way, without relying on any of these decoders.

Correct me if I'm wrong, but it seems like there are two main use-cases:
* validating the instruction is correct, likely before sending
* decoding a sent instruction fetched from somewhere else, which may be incorrect

In that sense, perhaps we should have two of these, `checked` and `unchecked`, where the `unchecked` just does the plain serde, and `checked` actually makes sure that the instruction is well-formed, with correct writable / signer flags and numbers of accounts.  The default is the `checked` version.  What do you think?

jordansexton: Yeah, I think you're right about the use cases. That's a good idea, I'll do that.

grimAgent:
Hey @jordansexton 
I'm trying to use this implementation by adding this dependency: `"@solana/spl-token": "0.2.0-alpha.0"` to my `package.json` but I'm seeing this error on compile:

```
./node_modules/@solana/spl-token/lib/esm/errors.js 11:7
Module parse failed: Unexpected token (11:7)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
| 
| export class TokenAccountNotFoundError extends TokenError {
>   name = 'TokenAccountNotFoundError';
| }
| /** Thrown if a program state account is not owned by the expected token program */
```

any ideas?

jordansexton:
@grimAgent are you using Create React App / Webpack by any chance? That error is usually an issue with how your build system handles ESM imports. Can you post a minimal reproduction repo?

0xCryptoSheik:
Hey, sorry just chiming in quickly.

This isn’t related to how Webpack or another bundler handles ESM, it just
shows that the browser build provided by the package (which happens to be
ESM), contains recent ES syntax, apparently class properties, untranspiled,
and that your bundler (probably webpack again but whatever it is) cannot
parse it.

To fix the issue, either the package should transpile more, or the consumer
should re-transpile further for the bundler to be able to understand the
syntax it’s bundling.

Hope this helps.

grimAgent:
> @grimAgent are you using Create React App / Webpack by any chance? That error is usually an issue with how your build system handles ESM imports. Can you post a minimal reproduction repo?

Here you go! https://github.com/grimAgent/spl-token-bug

jordansexton:
> Hey, sorry just chiming in quickly.

Chime in any time, that's good feedback.

We run into a lot of issues with wallet-adapter and CRA 4/5, and I don't have a great solution here because I've been trying to avoid adding another build tool.

There are some recent changes in WalletAdapter to output strict ESM .mjs files, and this partially solves the issue with Webpack 5, but there's an open PR to handle polyfills.

Parcel and Next handle these things out of the box, but Webpack doesn't. I'm not sure what to recommend.

grimAgent:
> > Hey, sorry just chiming in quickly.
> 
> Chime in any time, that's good feedback.
> 
> We run into a lot of issues with wallet-adapter and CRA 4/5, and I don't have a great solution here because I've been trying to avoid adding another build tool.
> 
> There are some recent changes in WalletAdapter to output strict ESM .mjs files, and this partially solves the issue with Webpack 5, but there's an open PR to handle polyfills.
> 
> Parcel and Next handle these things out of the box, but Webpack doesn't. I'm not sure what to recommend.

well, speaking of Next, I just tried it there as well, and I'm getting these error instead:
```
error - unhandledRejection: Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/node_modules/@solana/spl-token/lib/esm/instructions' is not supported resolving ES modules imported from /node_modules/@solana/spl-token/lib/esm/index.js
```

PR #2642, opened 2021-12-16
token: Extension state packing

joncinque:
#### Problem

Extensions need to be read / written in a way that's simple for the program processor and clients.

#### Solution

Implement a `Pack`-like interface for the TLV entries using bytemuck and Pod types for extensions.

### DIFF
--- a/token/program-2022/src/extension.rs
+++ b/token/program-2022/src/extension.rs
@@ -5,7 +5,6 @@ use {
         pod::*,
         state::{Account, Mint, Multisig},
     },
-    arrayref::{array_mut_ref, array_ref},
     bytemuck::{Pod, Zeroable},
     num_enum::{IntoPrimitive, TryFromPrimitive},
     solana_program::{
@@ -20,7 +19,6 @@ use {
 #[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
 #[repr(transparent)]
 pub struct Length(PodU16);
-
 impl From<Length> for usize {
     fn from(n: Length) -> Self {
         Self::from(u16::from(n.0))
@@ -35,46 +33,142 @@ impl TryFrom<usize> for Length {
     }
 }
 
-// TODO probably need an immutable version of this for clients
+/// Helper struct for returning the indices of the type, length, and value in
+/// a TLV entry
+struct TlvIndices(usize, usize, usize);
+fn get_extension_indices<V: Extension>(
+    tlv_data: &[u8],
+    init: bool,
+) -> Result<TlvIndices, ProgramError> {
+    let mut start_index = pod_get_packed_len::<AccountType>(); // start one byte in to skip the account type
+    while start_index < tlv_data.len() {
+        let type_end_index = start_index.saturating_add(pod_get_packed_len::<ExtensionType>());
+        let length_start_index = type_end_index;
+        let length_end_index = length_start_index.saturating_add(pod_get_packed_len::<Length>());
+        let value_start_index = length_end_index;
+
+        let extension_type =
+            pod_from_bytes::<ExtensionType>(&tlv_data[start_index..type_end_index])?;
+        // got to an empty spot, can init here, or move forward if not initing
+        if *extension_type == ExtensionType::Uninitialized {
+            if init {
+                return Ok(TlvIndices(
+                    start_index,
+                    length_start_index,
+                    value_start_index,
+                ));
+            } else {
+                start_index = length_start_index;
+            }
+        } else if *extension_type == V::TYPE {
+            // found an instance of the extension that we're initializing, abort!
+            if init {
+                return Err(ProgramError::InvalidArgument);
+            } else {
+                return Ok(TlvIndices(
+                    start_index,
+                    length_start_index,
+                    value_start_index,
+                ));
+            }
+        } else {
+            let length = pod_from_bytes::<Length>(&tlv_data[length_start_index..length_end_index])?;
+            let value_end_index = value_start_index.saturating_add(usize::from(*length));
+            start_index = value_end_index;
+        }
+    }
+    Err(ProgramError::InvalidAccountData)
+}
+
+fn check_not_multisig(input: &[u8]) -> Result<(), ProgramError> {
+    if input.len() == Multisig::LEN {
+        Err(ProgramError::InvalidAccountData)
+    } else {
+        Ok(())
+    }
+}
+
+fn tlv_index_unchecked<S: BaseState>(rest_input: &[u8]) -> Result<usize, ProgramError> {
+    if rest_input.is_empty() {
+        Ok(0)
+    } else {
+        Ok(Account::LEN.saturating_sub(S::LEN))
+    }
+}
+
+fn tlv_index<S: BaseState>(rest_input: &[u8]) -> Result<usize, ProgramError> {
+    if rest_input.is_empty() {
+        Ok(0)
+    } else {
+        let tlv_start_index = Account::LEN.saturating_sub(S::LEN);
+        let account_type = AccountType::try_from(rest_input[tlv_start_index])
+            .map_err(|_| ProgramError::InvalidAccountData)?;
+        if account_type != S::ACCOUNT_TYPE {
+            Err(ProgramError::InvalidAccountData)
+        } else {
+            Ok(tlv_start_index)
+        }
+    }
+}
+
+/// Encapsulates immutable base state data (mint or account) with possible extensions
+#[derive(Debug, PartialEq)]
+pub struct StateWithExtensions<'data, S: BaseState> {
+    /// Unpacked base data
+    pub base: S,
+    /// Slice of data containing all TLV data, deserialized on demand
+    tlv_data: &'data [u8],
+}
+impl<'data, S: BaseState> StateWithExtensions<'data, S> {
+    /// Unpack base state, leaving the extension data as a slice
+    pub fn unpack(input: &'data [u8]) -> Result<Self, ProgramError> {
+        check_not_multisig(input)?;
+        let (base_data, rest) = input.split_at(S::LEN);
+        let base = S::unpack(base_data)?;
+        let tlv_start_index = tlv_index::<S>(rest)?;
+        Ok(Self {
+            base,
+            tlv_data: &rest[tlv_start_index..],
+        })
+    }
+
+    /// Unpack a portion of the TLV data as the desired type
+    pub fn get_extension<V: Extension>(&self) -> Result<&V, ProgramError> {
+        if V::ACCOUNT_TYPE != S::ACCOUNT_TYPE {
+            return Err(ProgramError::InvalidAccountData);
+        }
+        let TlvIndices(_, length_start_index, value_start_index) =
+            get_extension_indices::<V>(self.tlv_data, false)?;
+        let length =
+            pod_from_bytes::<Length>(&self.tlv_data[length_start_index..value_start_index])?;
+        let value_end_index = value_start_index.saturating_add(usize::from(*length));
+        pod_from_bytes::<V>(&self.tlv_data[value_start_index..value_end_index])
+    }
+}
+
 /// Encapsulates mutable base state data (mint or account) with possible extensions
 #[derive(Debug, PartialEq)]
-pub struct MutStateWithExtensions<'data, S: BaseState> {
+pub struct StateWithExtensionsMut<'data, S: BaseState> {
     /// Unpacked base data
     pub base: S,
     /// Raw base data
-    pub base_data: &'data mut [u8],
+    base_data: &'data mut [u8],
     /// Slice of data containing all TLV data, deserialized on demand
-    pub tlv_data: &'data mut [u8],
+    tlv_data: &'data mut [u8],
 }
-impl<'data, S: BaseState + Pack + IsInitialized> MutStateWithExtensions<'data, S> {
+impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     /// Unpack the base state portion of the buffer, leaving the extension data as
     /// a serialized slice.
     pub fn unpack(input: &'data mut [u8]) -> Result<Self, ProgramError> {
-        let input_len = input.len();
-        if input_len == Multisig::LEN {
-            return Err(ProgramError::InvalidAccountData);
-        }
+        check_not_multisig(input)?;
         let (base_data, rest) = input.split_at_mut(S::LEN);
         let base = S::unpack(base_data)?;
-        if input_len == S::LEN {
-            Ok(Self {
-                base,
-                base_data,
-                tlv_data: rest, // empty slice
-            })
-        } else {
-            let tlv_start_index = Account::LEN.saturating_sub(S::LEN);
-            let account_type = AccountType::try_from(rest[tlv_start_index])
-                .map_err(|_| ProgramError::InvalidAccountData)?;
-            if account_type != S::ACCOUNT_TYPE {
-                return Err(ProgramError::InvalidAccountData);
-            }
-            Ok(Self {
-                base,
-                base_data,
-                tlv_data: &mut rest[tlv_start_index..],
-            })
-        }
+        let tlv_start_index = tlv_index::<S>(rest)?;
+        Ok(Self {
+            base,
+            base_data,
+            tlv_data: &mut rest[tlv_start_index..],
+        })
     }
 
     /// Unpack the base state portion of the buffer without checking for initialization,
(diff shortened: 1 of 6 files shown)


token/program-2022/src/extension.rs
```diff
@@ -38,30 +229,104 @@ pub enum Extension {
...
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub struct TlvEntry<V> {
-    /// Extension type encoded in the rest of the entry
-    pub extension: Extension,
-    /// Length of the entry, in bytes
-    pub length: u32,
-    /// Deserialized value
-    pub value: V,
+    fn try_from(extension_type: u8) -> Result<Self, Self::Error> {
+        match extension_type {
```
mvines, line 236: Have you tried https://crates.io/crates/enum-ordinalize before? I've used it before in programs to avoid exactly this kind of manual translation code

token/program-2022/src/extension.rs
```diff
@@ -1,6 +1,185 @@
...
+#[derive(Debug, PartialEq)]
+pub struct BaseStateWithExtensions<'data, S: BaseState> {
+    /// Unpacked base data
+    pub base: S,
+    /// Raw base data
+    pub base_data: &'data mut [u8],
+    /// Slice of data containing all TLV data, deserialized on demand
+    pub tlv_data: &'data mut [u8],
+}
+impl<'data, S: BaseState + Pack + IsInitialized> BaseStateWithExtensions<'data, S> {
```
mvines, line 30: I really liked using Pods in zk token: https://github.com/solana-labs/spl-zk-token/blob/master/program/src/pod.rs

It's so much nicer than packing/unpacking: no copy, writes are immediate (no explicit pack needed). I'm game to make the requirement that all extensions be Pods.

token/program-2022/src/extension.rs
```diff
@@ -38,30 +229,104 @@ pub enum Extension {
...
+            1 => Ok(ExtensionType::MintTransferFee),
+            2 => Ok(ExtensionType::AccountTransferFee),
+            3 => Ok(ExtensionType::MintCloseAuthority),
+            _ => Err(ProgramError::InvalidAccountData),
+        }
+    }
+}
+impl ExtensionType {
+    /// Get the data length of the type associated with the enum
+    pub fn get_associated_type_len(&self) -> usize {
```
mvines, line 247: With a `Pod`, this function melts away: https://github.com/solana-labs/spl-zk-token/blob/3cc9415b9dcff3c4e380369023211189f9dfc8f3/program/src/pod.rs#L47-L49

token/program-2022/src/extension.rs
```diff
@@ -1,6 +1,185 @@
...
+        program_error::ProgramError,
+        program_option::COption,
+        program_pack::{IsInitialized, Pack, Sealed},
+        pubkey::Pubkey,
+    },
+    std::convert::{TryFrom, TryInto},
+};
+
+const TYPE_LEN: usize = 1;
+const LENGTH_LEN: usize = 4;
```
mvines, line 17: How about making TYPE_LEN + LENGTH_LEN = 4 to better align the data.  A `u8` for the type feels like we might one day run out of room, and `u32` for length seems more than what we'd ever need. `u16` for both?

token/program-2022/src/extension.rs
joncinque: I haven't, but it looks great, will do!

token/program-2022/src/extension.rs
joncinque: Fine by me

token/program-2022/src/extension.rs
joncinque: Yes please, I was hoping to start playing with bytemuck on this. Thanks for pointing out the helper types that you're using, we may even want to have those available in the SDK.  Would that make C integration really gross, or do we not really care?

token/program-2022/src/extension.rs
joncinque: I may be dense -- how would that work? The idea is to get the required size for an account with a certain set of extensions, denoted by their extension type enum.  This approach creates a link from the enum to the type represented underneath.  For example, `ExtensionType::MintTransferFee` is giving the size of `struct MintTransferFee`.

token/program-2022/src/extension.rs
joncinque: Opting for `num-enum` since it's already being used in the program and works neatly

token/program-2022/src/extension.rs
mvines: > we may even want to have those available in the SDK

Yeah for sure. I've already snuck a couple minimal Pod things into the sdk (e.g. https://github.com/solana-labs/solana/pull/20493). I've been waiting for more adoption before pulling in additional stuff, like these zk-token helpers, since it's hard to take stuff out of the sdk once added

token/program-2022/src/extension.rs
mvines: Fine, I might have been too eager in this case with my Pod shilling.

joncinque:
I moved over to bytemuck, which certainly simplifies some bits.  Biggest annoyance: there's no native support for enums to be treated as `Pod`s, so we have to resort to manually impl'ing the unsafe `Pod` and `Zeroable` traits on them.  Since they're not carrying values and their repr size is declared, this seemed ok.  Let me know if you have a better idea though.

There's also how to deal with options.  After testing around, it seems like it'll be easiest to just impl `Pod` and `Zeroable` on `COption`, but we may need to explicitly set the size on it, ie. through `#[repr(C, u32)]. Again, I'm open to suggestions! I put a few other possibilities out there.

mvines:
> Biggest annoyance: there's no native support for enums to be treated as Pods, so we have to resort to manually impl'ing the unsafe Pod and Zeroable traits on them.

Here's what I did for the instruction type enum in zk-token: https://github.com/solana-labs/spl-zk-token/blob/0f758e8d5537677aee9caa6384a1b90e3b7fb377/program/src/instruction.rs#L301

Options are a bit of a struggle too. Will think about that and respond more tomorrow morning

joncinque:
Yeah, I noticed that, it's a clever way to create rust-style enums as Pods  I'm not sure how we can apply that to the opposite case of doing in-place encoding of the enum into a buffer.

I've been thinking about it more, and as long as the enum uses `#[repr(C)]` or `#[repr(Int)]`, I can't see a situation where we'll get screwed with `Pod` or `Zeroable`.  They're essentially just wrappers to make `transmute` safe, so we're telling the compiler, "it's ok, we want to transmute this type". In our case, we can be sure that the rust-style enum will look more like a union, which means we always reserve the greatest size, which is much more useful than the borsh-style compressing.

https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md#guide-level-explanation gives the best background I've found so far

token/program-2022/src/extension.rs
```diff
@@ -16,19 +260,25 @@ pub enum AccountType {
...
 impl Default for AccountType {
     fn default() -> Self {
         Self::Uninitialized
     }
 }
+// TODO This kind of stinks, but there's no way to automatically implement Pod
+// on an enum, which almost defeats the purpose of in-place serde!
+// Happy to try something else here though.
+#[allow(unsafe_code)]
+unsafe impl Zeroable for AccountType {}
```
mvines, line 272: Using `num-enum` is my pref here with `try_from` mapping to `AccountType:: Uninitialized` on error to avoid the unsafe

token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,186 @@
...
+/// from bytes.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[repr(C, u8)]
+pub enum PodOption<T: Pod> {
+    /// No data, but its size will still be 1 + size_of::<T>()
+    None,
+    /// Contains data value
+    Some(T),
+}
+#[allow(unsafe_code)]
```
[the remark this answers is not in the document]
joncinque: Ah of course, you're right, I knew I was missing something simple and basic.

token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,186 @@
...
+};
+
+// TODO Generic `PodOption` is tricky, since we need to be sure that it'll
+// always take up the same amount of space.  A few ideas, since all of these are
+// pubkeys at the moment:
+//
+//#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+//#[repr(transparent)]
+//struct NonZeroPubkey(Pubkey);
+//impl NonZeroPubkey {
```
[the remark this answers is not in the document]
joncinque: It'll be a bit hacky to make it work, since the normal NonZero* int types get to use special libcore optimizations, but it's looking like the best option

token/program-2022/src/extension.rs
```diff
@@ -1,13 +1,257 @@
...
+}
+
+/// Helper struct for returning the indices of the type, length, and value in
+/// a TLV entry
+struct TlvIndices(usize, usize, usize);
+fn get_extension_indices<V: Extension>(
+    tlv_data: &[u8],
+    init: bool,
+) -> Result<TlvIndices, ProgramError> {
+    let mut start_index = pod_get_packed_len::<AccountType>(); // start one byte in to skip the account type
```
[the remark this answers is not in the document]
joncinque: It certainly could be.  I kept it in since we have to write to it at some point, but I'll separate it out as its own field for more clarity.

joncinque:
We're back to `forbid(unsafe_code)` :tada:

token/program-2022/src/extension.rs
```diff
@@ -1,13 +1,279 @@
...
+
+/// Encapsulates immutable base state data (mint or account) with possible extensions
+#[derive(Debug, PartialEq)]
+pub struct StateWithExtensions<'data, S: BaseState> {
+    /// Unpacked base data
+    pub base: S,
+    /// Unpacked base data
+    pub account_type: AccountType,
+    /// Slice of data containing all TLV data, deserialized on demand
+    tlv_data: &'data [u8],
```
mvines, line 130: For the `Mint` implementation, it looks like the `account_type` and `tlv_data` will end up immediately after the base Mint.  But this means that the length of a Mint with extensions could be the same as Account::LEN. Rather than adding padding, like I think we discussed earlier, how about extending `check_not_multisig` to be `check_not_multisig_or_account`?

token/program-2022/src/extension.rs
```diff
@@ -1,13 +1,279 @@
...
+    base_data: &'data mut [u8],
+    /// Writable account type
+    account_type: &'data mut [u8],
+    /// Slice of data containing all TLV data, deserialized on demand
+    tlv_data: &'data mut [u8],
+}
+impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
+    /// Unpack the base state portion of the buffer, leaving the extension data as
+    /// a serialized slice.
+    pub fn unpack(input: &'data mut [u8]) -> Result<Self, ProgramError> {
```
mvines, line 178: I suspect there'll be cases in the implementation when one starts with a `StateWithExtensions` and then wants to promote it into a `StateWithExtensionsMut`. I ran into this need with zk-token, which spawned [PodAccountInfoData::into_mut](https://github.com/solana-labs/spl-zk-token/blob/e54c8986d9b8d37b4cc1ddc05846bdffc96e50c4/program/src/pod.rs#L108).

I don't think we need to address this here/now, just pointing this out.  One option could be to break up "unpack()"ing the base state from "new()", so that I could just `StateWithExtensionsMut::new()` from an account I've already used ``StateWithExtensions` with, and then jump into calling `StateWithExtensionsMut::get_extension()` without incurring the overhead of unpacking the base state again

joncinque: Ah ok, my thinking was that `StateWithExtensionsMut` would only be used if you knew that you might change the data later.  The ZK helpers are really great, but since we won't be modifying the base data (mint / account) in-place, it'll be pretty awkward to slot them in.  Are you subtly suggesting moving to in-place modifications for mints / tokens?  That could certainly take down the amount of compute units by the program!

token/program-2022/src/extension.rs
joncinque: Couldn't we then have a situation in which we're not totally sure if an account is an `Account` or a `Mint`?  A mint with extensions that takes it past 165 could be indiscernible from an `Account` with an extension, even if we add the account type. For example, let's say:

```
Account: 165 bytes... + [2, 0, 3, 0, 100, ....]
                         ^     ^       ^     ^
                    acct type  extension length data...
```
```
Mint: 82 bytes... + 83 bytes of random other data + [2, 0, 3, 0, 100, ....]
                                                     ^ data in extension just happens to look like this
```

With the current approach, we only start writing the TLV data after `Account::LEN`, which means we always know that the account type is going to be right after that.  We can special case checking for a Multisig, because those aren't extensible under any circumstances.

mvines: hmm, yeah I think you're right.  thanks

token/program-2022/src/extension.rs
```diff
@@ -38,60 +303,218 @@ pub enum Extension {
...
+    let extension_size: usize = extension_types
+        .iter()
+        .map(|e| {
+            e.get_associated_type_len()
+                .saturating_add(size_of::<ExtensionType>())
+                .saturating_add(pod_get_packed_len::<Length>())
+        })
+        .sum();
+    let total_extension_size = if extension_size == Multisig::LEN {
+        extension_size + 1
```
CriesofCarrots, line 343: Comment explaining this magic number?
Actually, I'm a little confused by this, as I though that we'd be using `ExtensionType::Uninitialized` (2 bytes) for this case.

CriesofCarrots (commented):
Looking good! Nothing big from me, although I'm looking forward to seeing a test with multiple extensions (and maybe a custom test extension that puts an Account at multisig length) to help me visualize. (Or maybe I should write such a test? :) )

token/program-2022/src/extension.rs
joncinque: Oh yes, very good find! you're absolutely right

joncinque:
I've addressed your points @CriesofCarrots -- as for the tests, I added one for multisig length, which revealed a bug. so if you want to risk the current implementation, feel free to add a multi-extension test :smile: otherwise, I'll be happy to add it

joncinque:
I ended up adding the last few tests recommended, so this is finally ready by my book!

CriesofCarrots:
Oh shoot, sorry I let you beat me to that test!

joncinque:
> Oh shoot, sorry I let you beat me to that test!

I mainly wanted to spare you from troubleshooting one of my bugs :sweat_smile:

token/program-2022/src/extension.rs
```diff
@@ -453,18 +458,31 @@ impl Extension for MintPaddingTest {
...
 mod test {
     use {
         super::*,
-        crate::state::test::{TEST_MINT, TEST_MINT_SLICE},
+        crate::state::test::{TEST_ACCOUNT, TEST_ACCOUNT_SLICE, TEST_MINT, TEST_MINT_SLICE},
         solana_program::pubkey::Pubkey,
     };
 
     #[test]
     fn mint_with_extension_pack_unpack() {
```
CriesofCarrots, line 480: There are a couple additional things that I was hoping to exemplify in a multi-extension test:
- That extensions can be initialized in any order, and unpacking still works
- Given an opaque data buffer, how to determine which extension(s) are present
This test is getting pretty long, though. Second test?

CriesofCarrots (approved):
Couple nits/suggestions around testing and a comment, but otherwise lgtm!

token/program-2022/src/extension.rs
```diff
@@ -339,14 +344,14 @@ pub fn get_account_len(extension_types: &[ExtensionType]) -> usize {
                 .saturating_add(pod_get_packed_len::<Length>())
         })
         .sum();
-    let total_extension_size = if extension_size == Multisig::LEN {
-        extension_size.saturating_add(size_of::<ExtensionType>())
-    } else {
-        extension_size
-    };
-    total_extension_size
+    let account_size = extension_size
```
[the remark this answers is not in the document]
joncinque: No worries, that's what tests are for!

token/program-2022/src/extension.rs
```diff
@@ -428,6 +433,27 @@ impl Extension for AccountTransferFee {
     const ACCOUNT_TYPE: AccountType = AccountType::Account;
 }
 
+/// Padding a mint account to be exactly Multisig::LEN.
+/// We need to pad 185 bytes, since Multisig::LEN = 355, Account::LEN = 165,
```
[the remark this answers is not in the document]
joncinque: Certainly, will do!

token/program-2022/src/extension.rs
```diff
@@ -517,4 +543,33 @@ mod test {
...
+        extension.padding3 = [1; 9];
+
+        // check raw buffer
+        let mut expect = TEST_MINT_SLICE.to_vec();
+        expect.extend_from_slice(&[0; Account::LEN - Mint::LEN]); // padding
+        expect.push(AccountType::Mint.into());
+        expect.extend_from_slice(&(ExtensionType::MintPaddingTest as u16).to_le_bytes());
+        expect.extend_from_slice(&(pod_get_packed_len::<MintPaddingTest>() as u16).to_le_bytes());
+        expect.extend_from_slice(&vec![1; pod_get_packed_len::<MintPaddingTest>()]);
+        expect.extend_from_slice(&(ExtensionType::Uninitialized as u16).to_le_bytes());
```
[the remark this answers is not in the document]
joncinque: Oops, nice catch!

token/program-2022/src/extension.rs
```diff
@@ -572,4 +656,124 @@ mod test {
...
+
+        // check raw buffer
+        let mut expect = TEST_ACCOUNT_SLICE.to_vec();
+        expect.extend_from_slice(&[0; Account::LEN - Mint::LEN]); // padding
+        expect.push(AccountType::Account.into());
+        expect.extend_from_slice(&(ExtensionType::AccountPaddingTest as u16).to_le_bytes());
+        expect
+            .extend_from_slice(&(pod_get_packed_len::<AccountPaddingTest>() as u16).to_le_bytes());
+        expect.extend_from_slice(&vec![2; pod_get_packed_len::<AccountPaddingTest>()]);
+        expect.extend_from_slice(&(ExtensionType::Uninitialized as u16).to_le_bytes());
```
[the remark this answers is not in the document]
joncinque: Yes, thanks for noticing!

token/program-2022/src/extension.rs
joncinque: Good ideas -- I added some tests which should cover this.  I also added a check on the padding for Account::LEN - Mint::LEN to make sure it's all zeroes.  Not sure if it's useful, but it feels safer

token/program-2022/src/extension.rs
```diff
@@ -97,13 +97,32 @@ fn check_account_type<S: BaseState>(account_type: AccountType) -> Result<(), Pro
...
+///                     acct type  extension length data...
+///
+/// Mint: 82 bytes... + 83 bytes of other extension data + [2, 0, 3, 0, 100, ....]
+///                                                         ^ data in extension just happens to look like this
+///
+/// With this approach, we only start writing the TLV data after Account::LEN,
+/// which means we always know that the account type is going to be right after
+/// that. We do a special case checking for a Multisig length, because those
+/// aren't extensible under any circumstances.
+const BASE_ACCOUNT_LENGTH: usize = Account::LEN;
```
CriesofCarrots, line 117: 🎉  love this, thank you

token/program-2022/src/extension.rs
```diff
@@ -476,6 +500,108 @@ mod test {
...
+        32, 0, // data
+        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+        1, 1,
+    ];
+
+    #[test]
+    fn unpack_opaque_buffer() {
+        let state = StateWithExtensions::<Mint>::unpack(MINT_WITH_EXTENSION).unwrap();
+        assert_eq!(state.base, TEST_MINT);
+        let extension = state.get_extension::<MintCloseAuthority>().unwrap();
```
CriesofCarrots, line 546: So to be clear, a client only knows which `get_extension` type(s) to use for a particular Mint by trying them?

CriesofCarrots (approved):
:shipit: Thanks for all the tests!

token/program-2022/src/extension.rs
joncinque: At the moment, yes.  Since extensions are meant to be additive, I was thinking that clients and instruction processors just do something like:
```
if let Ok(extension) = state.get_extension::<SomeExtension>() {
    /// do stuff here
} else {
    /// do some other stuff
}
```
But now that you mention it, the `get_account_data_length` instruction will have to extract all of the extension types contained in the mint in order to provide an answer, which means there's definitely a use case for something like:
```
let extensions: Vec<ExtensionType> = state.get_extensions();
```
which loops through all of the TLV entries and extracts the type.  Seems like we can do that in another PR though -- what do you think?

CriesofCarrots: Yes! That's exactly where my head was going. I think that method would be super helpful. Separate PR sounds perfect.
On a related topic, do we ever foresee reworking extensions to support multiples of the same ExtensionType?

joncinque: That's a good question too... maybe there's a different API for that? Something like `state.get_extension_index(2)::<Extension>()`? And then some marker on the extensions so that they can allow multiples.  We'll have to rework the internals for that either way, and I think that our current design would allow for it pretty easily

mvines: > do we ever foresee reworking extensions to support multiples of the same ExtensionType?

I sorta hope never myself

PR #2648, opened 2021-12-17
fix account list in token_swap instructions docs

dungph:


joncinque:
Thanks for the contribution, looks good!

PR #2653, opened 2021-12-20
Increase compute max units due to changes in handling of memory ops

dmakarov:
bpf-tools v1.21 changed implementation of memory operations, which results in larger instruction numbers for the same operations. This PR adjusts the compute limits in tests affected by the changes in bpf-tools.

### DIFF
--- a/token/program/tests/assert_instruction_count.rs
+++ b/token/program/tests/assert_instruction_count.rs
@@ -22,7 +22,7 @@ const TRANSFER_AMOUNT: u64 = 1_000_000_000_000_000;
 #[tokio::test]
 async fn initialize_mint() {
     let mut pt = ProgramTest::new("spl_token", id(), processor!(Processor::process));
-    pt.set_bpf_compute_max_units(2_500); // last known 2252
+    pt.set_bpf_compute_max_units(5_000); // last known 2252
     let (mut banks_client, payer, recent_blockhash) = pt.start().await;
 
     let owner_key = Pubkey::new_unique();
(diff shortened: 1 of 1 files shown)


joncinque:
Would you mind also updating the "last known" numbers in the comments? this will help figure out how CU usage evolves over time

dmakarov:
I would gladly update the comments, but not sure how to get this information. Are this counts printed on RUSTLOG set to debug?

joncinque:
Oh right, that part is not very clear, and a bit annoying.  You essentially have to read through the output of `cargo test-bpf` and match up the test to the instruction count printed by default on the console.

In that case, don't worry about it, I'm doing some work around there soon, so I'll update them at that point :smile:

joncinque (approved):

dmakarov:
Actually after the update the tests still fail. For example, for initialize_mint I get this message

```
[2021-12-21T01:44:47.393203000Z DEBUG solana_runtime::message_processor] Program Token22gQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 2500 of 2500 compute units
[2021-12-21T01:44:47.393231000Z DEBUG solana_runtime::message_processor] Program failed to complete: exceeded maximum number of instructions allowed (2500) at instruction #13625
```
Where does this 2500 limit come from if I updated it to 5000?

joncinque:
This is my fault for not pointing it out -- you'll also have to update these numbers in `token/program-2022`: https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/tests/assert_instruction_count.rs

PR #2655, opened 2021-12-21
Increase compute max units in program-2022 assert_instruction_count

dmakarov:


joncinque (approved):
Thanks for doing this!

PR #2657, opened 2021-12-21
Update tests to prepare for `BanksClientError`

joncinque:
#### Problem

Soon, `BanksClient` will only return `BanksClientError`, but many tests in SPL rely on `TransportError` being returned.

#### Solution

Be sneaky, and add a currently unnecessary `into()` conversion, which will be necessary when the return type changes.  This way, we never break downstream builds.


CriesofCarrots (approved):
Lgtm!
(Sorry about all the `#[allow(clippy::useless_conversion)]`s)

PR #2664, opened 2021-12-26
Feature/build system

cogoo:
- build(rollup): remove rollup and related files
- build(tsc): create cjs and esm bundles
- chore(package.json): remove browser and browserlist

I perceive we'll have issues with the @solana/web3.js lib that depends on some node builtins.
An error like process is not defined or buffer is not defined is expected.

Unsure on how best to handle that, we could list those as peer deps, or for the sake of simplicity, bundle the @solana/web3.js package to have those builtins provided, so upstream users don't have to be concerned about it.


jordansexton:
Looks great, thanks!

PR #2666, opened 2021-12-26
Enable rustls-tls-native-roots for spl-token CLI

### LINKED ISSUE #2665: spl-token Does Not Respect System Certificate Store
Hey folks!

I'm behind an HTTP proxy that has a custom HTTPS certificate that must be trusted by applications that access the internet on my machine. I have setup the certificate to be trusted by my system, but the `spl-token` CLI doesn't respect the system cert store. Interestingly the `solana` CLI works fine.

I'm seeing if I can figure out what will make it work, but I wanted to open up the issue in case anybody has any thoughts.

zicklag:
Fixes #2665 .

This is just a quick fix that got `spl-token` to work behind my http proxy with a custom cert.

Let me know if you've got any feedback or if we should do this for all the CLIs, etc.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1424,6 +1424,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "hyper-tls"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d6183ddfa99b85da61a140bea0efc93fdf56ceaa041b37d553518030827f9905"
+dependencies = [
+ "bytes 1.0.1",
+ "hyper",
+ "native-tls",
+ "tokio",
+ "tokio-native-tls",
+]
+
+[[package]]
 name = "idna"
 version = "0.2.2"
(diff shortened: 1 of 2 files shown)


joncinque:
You've definitely stumbled on a strange issue, thanks for finding this!

For some reason, it looks like the monorepo pulls in `native-tls` when including `reqwest`: https://github.com/solana-labs/solana/blob/bb97c8fdcd3ec43979f25c18c49587c91089723d/Cargo.lock#L3661
But meanwhile, the solana-program-library does not: https://github.com/solana-labs/solana-program-library/blob/cd8d79a2b4aa4f90c02514d762ab21023449b6cb/Cargo.lock#L2538

Looking at the monorepo. it's surprising that `native-tls` is pulled in considering `reqwest` is always included as:
```
reqwest = { version = "0.11.6", default-features = false, features = ["blocking", "rustls-tls", "json"] }
```

There may be some transitive dependency that enables it.  I'm not familiar with why `reqwest` is used in this way, so maybe @CriesofCarrots can expand on this?

PR #2667, opened 2021-12-27
fix grcov dependency checking

eventhorizonpl:
Hi,

I have grcov in version 0.8.4 and current version check causes problem here when running coverage.sh

This change fixes version check for me, but in the future this check should be redesigned.

Best regards,
Michal

### DIFF
--- a/coverage.sh
+++ b/coverage.sh
@@ -11,7 +11,7 @@ if ! which grcov; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


joncinque:
Thanks so much for the contribution!  As you've noticed, this script has unfortunately fallen into disrepair.  For now, what do you think about updating the check to be the same one as the solana mono-repo's check? https://github.com/solana-labs/solana/blob/bb97c8fdcd3ec43979f25c18c49587c91089723d/scripts/coverage.sh#L13

eventhorizonpl:
@joncinque Done.

Issue #2672, opened 2021-12-28
[token-swap] `cargo test` fails to compile on Apple M1: static library `python3.9` missing


peyton:
Apple M1
macOS 12.1 (21C52)
aarch64-apple-darwin

`cargo test` fails with 

```
error: could not find native static library `python3.9`, perhaps an -L flag is missing?

error: could not compile `pyo3` due to previous error
```

Appears to be an issue with PyO3 v0.12.3's `build.rs` being a bit too clever with linking libraries.

### Steps to reproduce
1. Apple M1, macOS 12.1 (21C52), aarch64-apple-darwin
2. Unset `$PYTHONFRAMEWORK` env variable.
3. Run `cargo test` in `token-swap/program`.

### Expected result
Tests compile and pass.

### Actual result
Verbose output:
```
error: could not find native static library `python3.9`, perhaps an -L flag is missing?

error: could not compile `pyo3` due to previous error

Caused by:
[...]

PR #2673, opened 2021-12-28
[token-swap] Upgrade PyO3 from 0.12.3 to 0.15.1

peyton:
Addresses #2672


joncinque:
Thanks so much for the contribution!  Looks like this just needs to have formatting run against it.  Could you run `cargo fmt --all` from the root of the repo and then push up those changes? Once it passes CI, I'll merge in right away

peyton:
done, fingers crossed 🤞

joncinque:
Thanks for the contribution, looks great!

PR #2678, opened 2021-12-29
add test for memo

eventhorizonpl:
Hi,

I want to add some tests to have a better code coverage.

Best regards,
Michal

### DIFF
--- a/memo/program/src/lib.rs
+++ b/memo/program/src/lib.rs
@@ -44,12 +44,11 @@ mod tests {
 
     #[test]
     fn test_build_memo() {
-        let program_id = Pubkey::new(&[0; 32]);
+        let signer_pubkey = Pubkey::new_unique();
         let memo = "🐆".as_bytes();
-        let instruction = build_memo(memo, &[&program_id]);
+        let instruction = build_memo(memo, &[&signer_pubkey]);
         assert_eq!(memo, instruction.data);
-        for account in instruction.accounts.iter() {
-            assert_eq!(program_id, account.pubkey);
-        }
+        assert_eq!(instruction.accounts.len(), 1);
+        assert_eq!(instruction.accounts[0].pubkey, signer_pubkey);
     }
 }
--- a/memo/program/src/lib.rs
+++ b/memo/program/src/lib.rs
@@ -46,7 +46,7 @@ mod tests {
     fn test_build_memo() {
         let program_id = Pubkey::new(&[0; 32]);
         let memo = "🐆".as_bytes();
-        let instruction = build_memo(&memo, &[&program_id]);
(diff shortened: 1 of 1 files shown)


memo/program/src/lib.rs
```diff
@@ -37,3 +37,19 @@ pub fn build_memo(memo: &[u8], signer_pubkeys: &[&Pubkey]) -> Instruction {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_build_memo() {
+        let program_id = Pubkey::new(&[0; 32]);
```
CriesofCarrots, line 47: Why do you call this program_id? The list of pubkeys passed to the memo program are signers, not necessarily related to any program id. I think this will be more helpful if it's renamed.

CriesofCarrots: Also:
```suggestion
        let program_id = Pubkey::new_unique();
```

memo/program/src/lib.rs
```diff
@@ -37,3 +37,19 @@ pub fn build_memo(memo: &[u8], signer_pubkeys: &[&Pubkey]) -> Instruction {
...
+        let memo = "🐆".as_bytes();
+        let instruction = build_memo(memo, &[&program_id]);
+        assert_eq!(memo, instruction.data);
+        for account in instruction.accounts.iter() {
+            assert_eq!(program_id, account.pubkey);
+        }
```
CriesofCarrots, line 53:
```suggestion
        assert_eq!(instruction.accounts.len(), 1);
        assert_eq!(instruction.accounts[0], program_id);
```
Iterator is a strange way to test, since there's only one instruction account...

CriesofCarrots (commented):
Sure, I suppose this is somewhat helpful to show how to build a memo instruction. I will approve with a little clean up. However, perhaps it would be even more useful to have a ProgramTest integration test that demonstrates success and failure when signing memos?

eventhorizonpl:
> Sure, I suppose this is somewhat helpful to show how to build a memo instruction. I will approve with a little clean up. However, perhaps it would be even more useful to have a ProgramTest integration test that demonstrates success and failure when signing memos?

Thank you for review. I changed code according to your suggestions.

Regarding "it would be even more useful to have a ProgramTest integration " you are thinking about functional test here?

Best regards,
Michal

CriesofCarrots (approved):
Thanks!
Regarding the integration test, I am thinking of something similar to https://github.com/solana-labs/solana-program-library/blob/master/associated-token-account/program/tests/process_create_associated_token_account.rs (but much simpler, since the memo program is so simple)
Not an urgent need, but feel free to PR if the task appeals to you.

PR #2679, opened 2021-12-29
Governance: Program metadata

SebastianBor:
#### Summary 

`spl-governance` is deployed using multiple instances and `governnace-ui` needs a to know what version of the program it communicates with in case the api is different between versions. 

#### Solution 

`ProgramMetadata` account was introduced to store the given instance version. The version is stored using `UpdatePorgramMetadata` instruction which dumps the program's crate version into the account and also outputs into the logs to make it possible to query the version using Tx simulation

### DIFF
--- a/governance/program/src/instruction.rs
+++ b/governance/program/src/instruction.rs
@@ -7,6 +7,7 @@ use crate::{
             get_account_governance_address, get_mint_governance_address,
             get_program_governance_address, get_token_governance_address, GovernanceConfig,
         },
+        program_metadata::get_program_metadata_address,
         proposal::{get_proposal_address, VoteType},
         proposal_instruction::{get_proposal_instruction_address, InstructionData},
         realm::{get_governing_token_holding_address, get_realm_address, RealmConfigArgs},
@@ -443,6 +444,12 @@ pub enum GovernanceInstruction {
     ///   4. `[signer]` Payer
     ///   5. `[]` System
     CreateTokenOwnerRecord {},
+
+    /// Updates ProgramMetadata account
+    /// The instruction dumps information implied by the program code into a persistent account
+    ///
+    ///   0. `[writable]` ProgramMetadata account. PDA seeds: ['metadata']
+    UpdateProgramMetadata {},
 }
 
 /// Creates CreateRealm instruction
@@ -1373,6 +1380,29 @@ pub fn create_token_owner_record(
     ];
 
     let instruction = GovernanceInstruction::CreateTokenOwnerRecord {};
+
+    Instruction {
+        program_id: *program_id,
+        accounts,
+        data: instruction.try_to_vec().unwrap(),
+    }
+}
+
+/// Creates UpdateProgramMetadata instruction
+pub fn upgrade_program_metadata(
+    program_id: &Pubkey,
+    // Accounts
+    payer: &Pubkey,
+) -> Instruction {
+    let program_metadata_address = get_program_metadata_address(program_id);
+
+    let accounts = vec![
+        AccountMeta::new(program_metadata_address, false),
+        AccountMeta::new(*payer, true),
+        AccountMeta::new_readonly(system_program::id(), false),
+    ];
+
+    let instruction = GovernanceInstruction::UpdateProgramMetadata {};
 
     Instruction {
         program_id: *program_id,
--- a/governance/program/src/instruction.rs
+++ b/governance/program/src/instruction.rs
@@ -446,9 +446,11 @@ pub enum GovernanceInstruction {
     CreateTokenOwnerRecord {},
 
     /// Updates ProgramMetadata account
-    /// The instruction dumps information implied by the program code into a persistent account
+    /// The instruction dumps information implied by the program's code into a persistent account
     ///
-    ///   0. `[writable]` ProgramMetadata account. PDA seeds: ['metadata']
+    ///  0. `[writable]` ProgramMetadata account. PDA seeds: ['metadata']
+    ///  1. `[signer]` Payer
+    ///  2. `[]` System
     UpdateProgramMetadata {},
 }
 
--- a/governance/program/src/processor/process_update_program_metadata.rs
+++ b/governance/program/src/processor/process_update_program_metadata.rs
@@ -1,0 +1,57 @@
+//! Program state processor
+
+use borsh::BorshSerialize;
+use solana_program::{
+    account_info::{next_account_info, AccountInfo},
+    entrypoint::ProgramResult,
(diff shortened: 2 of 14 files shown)


governance/program/src/instruction.rs
```diff
@@ -443,6 +444,14 @@ pub enum GovernanceInstruction {
...
     ///   5. `[]` System
     CreateTokenOwnerRecord {},
+
+    /// Updates ProgramMetadata account
+    /// The instruction dumps information implied by the program's code into a persistent account
+    ///
+    ///  0. `[writable]` ProgramMetadata account. PDA seeds: ['metadata']
+    ///  1. `[signer]` Payer
+    ///  2. `[]` System
+    UpdateProgramMetadata {},
```
joncinque, line 454: nit: Since it could be confusing between the governance program and the governed program, let's make this totally clear and call it `GovernanceProgramMetadata` everywhere
```suggestion
    UpdateGovernanceProgramMetadata {},
```

governance/program/src/processor/process_update_program_metadata.rs
```diff
@@ -0,0 +1,67 @@
...
+    accounts: &[AccountInfo],
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+
+    let program_metadata_info = next_account_info(account_info_iter)?; // 0
+    let payer_info = next_account_info(account_info_iter)?; // 1
+    let system_info = next_account_info(account_info_iter)?; // 2
+
+    let rent = Rent::get().unwrap();
+    let updated_at = Clock::get().unwrap().slot;
```
joncinque, line 32: nit: since the process function is returning a `ProgramResult`, these can be changed to 
```suggestion
    let rent = Rent::get()?;
    let updated_at = Clock::get()?.slot;
```
And that way, if something goes wrong, there's a nicer error than a panic

governance/program/src/processor/process_update_program_metadata.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+    // Put the metadata info into the logs to make it possible to extract it using Tx simulation
```
joncinque, line 36: not-for-now: once 1.9 is released, you can create a specific instruction to just add the version into return data!

governance/program/src/state/program_metadata.rs
```diff
@@ -0,0 +1,77 @@
...
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub struct ProgramMetadata {
+    /// Governance account type
+    pub account_type: GovernanceAccountType,
+
+    /// The slot when the metadata was captured
+    pub updated_at: Slot,
+
+    /// The version of the program
+    pub version: String,
```
joncinque, line 23: You may want to actually use something fixed-length here instead, or comment about the maximum supported size

governance/program/src/state/program_metadata.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    /// Reserved
+    pub reserved: [u8; 64],
+}
+
+impl AccountMaxSize for ProgramMetadata {
+    fn get_max_size(&self) -> Option<usize> {
+        Some(85)
```
joncinque, line 31: At 85, that gives string data of 8, correct?  1 + 8 + 64 for everything else is 73, then the string is 4 bytes for length, then the data, so 85 - 1 - 8 - 64 - 4 = 8.  That means the longest string would be "XX.YY.ZZ".  If you go fixed length, then you can do a "XXX.YYY.ZZZ", which would definitely cover you forever.

governance/program/src/processor/process_update_program_metadata.rs
```diff
@@ -0,0 +1,67 @@
...
+            program_metadata_info,
+            &program_metadata_data,
+            &get_program_metadata_seeds(),
+            program_id,
+            system_info,
+            &rent,
+        )?;
+    } else {
+        let mut program_metadata_data =
+            get_program_metadata_data(program_id, program_metadata_info)?;
```
joncinque, line 58: Be sure to validate that `program_metadata_info.key` is the expected metadata address, otherwise this instruction could be maliciously used to clobber the data in any other account type!

joncinque (commented):
Just a couple of comments, otherwise looks good!

governance/program/src/processor/process_update_program_metadata.rs
SebastianBor: that's cool, done

governance/program/src/processor/process_update_program_metadata.rs
SebastianBor: Super cool!!! thank you for the heads up

governance/program/src/state/program_metadata.rs
SebastianBor: Changed the size to support XXX.YYY.ZZZ

governance/program/src/state/program_metadata.rs
SebastianBor: The version is variable format so I decided to keep it as string and added the max size comment

governance/program/src/processor/process_update_program_metadata.rs
SebastianBor: All the accounts are validated during deserialisation using `get_account_data` and  hence it's not possible to substitute with different account types and the PDA check is not needed  https://github.com/solana-labs/solana-program-library/blob/d6d0b92ae762ac0d4bf726bb28a99cc642e5771b/governance/tools/src/account.rs#L139

Do you think it should check the address anyway?

governance/program/src/instruction.rs
SebastianBor: I renamed it, then thought about it hard and reverted the change :)

The name become too verbose and it was still to close to `GovernanceMetadata` account which I'm planning to add in future versions while `ProgramMetadata` can be used consistently as a pattern accross different programs.

At the same time I want to remove the concrete governance types `ProgramGovernance`, `MintGovernance` and `TokenGovernance`. They are not really needed in the program and they only support UI.

governance/program/src/processor/process_update_program_metadata.rs
joncinque: Ah yes you're right, my mistake.  I forgot that `is_initialized()` checks the account type

governance/program/src/instruction.rs
joncinque: Fair enough

joncinque (approved):
looks good!

PR #2680, opened 2021-12-30
Increase compute budget limits in a test affected by new mem op cost

dmakarov:


bausano:
Hi @dmakarov, could you please provide a link to the CHANGELOG or other documentation regarding this new mem cost? I am struggling with upgrading solana version in my own program because now some endpoints run out of compute units, and I would like to understand what has changed.

Thank you!

dmakarov:
The changes were introduces in this PR https://github.com/solana-labs/solana/pull/22101
We updated the amount of compute units memory operations syscalls consume. Since then
the base cost of mem op was reduced to 10.  If your program does a lot of data/array copying,
it can run over the compute budget limit, because memcpy, memmove, memcmp, memset operations cost more.

PR #2690, opened 2021-12-31
Use Optimized WAD and HALF_WAD Functions For Rate & Decimal Math

bonedaddy:
# Overview

* Optimizes the `wad` and `half_wad` functions to use constant variables.
* optimizes `try_ceil_u64` to use a constant one variable, instead of deriving at runtime.


jordansexton:
Nice!

token-lending/program/src/math/decimal.rs
```diff
@@ -40,14 +40,13 @@ impl Decimal {
         Self(U192::zero())
     }
 
-    // OPTIMIZE: use const slice when fixed in BPF toolchain
     fn wad() -> U192 {
-        U192::from(WAD)
+        const DWAD: U192 = U192([1000000000000000000, 0, 0]);
```
jordansexton, line 44: nit: what about underscoring these for legibility?

bonedaddy: ah yes good catch

PR #2692, opened 2022-01-02
Governance: Native Treasuries

SebastianBor:
#### Summary

At present `spl-governance` supports only `spl-token ` based treasury accounts and it's not possible to have a native SOL treasury. Only `wSOL` is supported which is both confusing and inconvenient for users.

Its also not possible to pay for execution of proposal instructions using DAO treasury and personal wallets a currently used. 

#### Solution

Create a `NativeTreasury` account which is a PDA without any data and make it possible to associate it with `Governance` accounts. This way any proposal instruction executed for the governance can use the funds from the associated SOL treasury account or in the simple case the instruction can be just a native SOL transfer.

### DIFF
--- a/governance/program/src/processor/process_execute_instruction.rs
+++ b/governance/program/src/processor/process_execute_instruction.rs
@@ -48,36 +48,31 @@ pub fn process_execute_instruction(program_id: &Pubkey, accounts: &[AccountInfo]
 
     let instruction_account_infos = account_info_iter.as_slice();
 
+    let mut signer_seeds: Vec<&[&[u8]]> = vec![];
+
+    // Sign the transaction using the governance PDA
     let mut governance_seeds = governance_data.get_governance_address_seeds()?.to_vec();
     let (_, bump_seed) = Pubkey::find_program_address(&governance_seeds, program_id);
     let bump = &[bump_seed];
     governance_seeds.push(bump);
 
-    // TODO: Add treasury seeds only if the treasury account is present instruction_account_infos
+    signer_seeds.push(&governance_seeds[..]);
+
+    // Sign the transaction using the governance treasury PDA if required by the instruction
     let mut treasury_seeds = get_native_treasury_address_seeds(governance_info.key).to_vec();
-    let (_, treasury_bump_seed) = Pubkey::find_program_address(&treasury_seeds, program_id);
+    let (treasury_address, treasury_bump_seed) =
+        Pubkey::find_program_address(&treasury_seeds, program_id);
     let treasury_bump = &[treasury_bump_seed];
-    treasury_seeds.push(treasury_bump);
 
-    //let address = get_native_treasury_address(program_id, governance_info.key);
+    if instruction_account_infos
+        .into_iter()
+        .any(|a| a.key == &treasury_address)
+    {
+        treasury_seeds.push(treasury_bump);
+        signer_seeds.push(&treasury_seeds[..]);
+    }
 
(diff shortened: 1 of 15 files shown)


governance/program/src/processor/process_execute_instruction.rs
```diff
@@ -47,16 +48,31 @@ pub fn process_execute_instruction(program_id: &Pubkey, accounts: &[AccountInfo]
...
+        Pubkey::find_program_address(&treasury_seeds, program_id);
+    let treasury_bump = &[treasury_bump_seed];
+
+    if instruction_account_infos
+        .iter()
+        .any(|a| a.key == &treasury_address)
+    {
+        treasury_seeds.push(treasury_bump);
+        signers_seeds.push(&treasury_seeds[..]);
+    }
```
joncinque, line 73: This is a usability question: is the concept that the native treasury will pay back whoever paid for transaction fees and things like that?  So in your proposal, you would have to put in an instruction to also send some lamports back to the fee payer, and then expect that the fee payer is the the one who actually executes the proposal? Or would separate proposals be created to pay back people who executed transactions in the past?

There could be a neat feature of just sending the lamports back to the fee payer directly in the execution, but that would need to be tightly controlled.

joncinque (dismissed):
Looks good to me! Mostly some nits and questions, but nothing that should stop this

governance/program/src/state/enums.rs
```diff
@@ -56,6 +56,10 @@ pub enum GovernanceAccountType {
 
     /// Program metadata account. It stores information about the particular SPL-Governance program instance
     ProgramMetadata,
+
+    /// NativeTreasury account
+    /// The account has no data and can be used as payer for instructions signed by Governance PDAs or as native SOL treasury
+    NativeTreasury,
```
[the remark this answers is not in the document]
SebastianBor: Since we never deserialise the account the address derivation provides the required safety.
This account is also not even owned by `spl-governance` so it can't write anything to it, and then PDA with non zero data can't spend its lamports either.
I removed the enum value to avoid confusions

governance/program/src/processor/process_execute_instruction.rs
SebastianBor: There are several use cases here

1) The treasury is used as the fee payer for the proposal instructions and hence the the person who executes the proposal doesn't foot the bill except for the outer transaction which is negligible

2) In some cases proposal creation itself requires some expenses. For example a new Mango market setup costs ~2SOL. In that case the proposer can add an extra instruction to the proposal to get reimbursed for that expense. However if the proposal is not successful the a new proposal would have to be created to get just the expenses back

3) Instructions themselves can reimburse the creator. For example `program upgrade` would return the lamports recovered from the disposed program buffer to the proposal creator 

4) Pure native SOL treasury where any instructions like payments or staking could be executed against it

joncinque: > 1. The treasury is used as the fee payer for the proposal instructions and hence the the person who executes the proposal doesn't foot the bill except for the outer transaction which is negligible

Oh interesting, I didn't think this worked! I thought the signature check was done on the outside, before execution, which means that a PDA can't be a fee payer.  If this actually works, it opens up a ton of other use-cases!

SebastianBor: Yes it does work. As long as the PDA has no data and is owned by the system program (and not the PDA owner)  then it can be used as fee payer

PR #2699, opened 2022-01-04
libraries tests p2

eventhorizonpl:
Hi,

This PR adds more libraries tests.

Best regards,
Michal

### DIFF
--- a/libraries/math/src/instruction.rs
+++ b/libraries/math/src/instruction.rs
@@ -167,3 +167,72 @@ pub fn noop() -> Instruction {
         data: MathInstruction::Noop.try_to_vec().unwrap(),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_precise_sqrt() {
+        let instruction = precise_sqrt(u64::MAX);
+        assert_eq!(0, instruction.accounts.len());
+        assert!(!instruction.data.is_empty());
+        assert!(!instruction.program_id.to_string().is_empty());
+    }
+
+    #[test]
+    fn test_sqrt_u64() {
+        let instruction = sqrt_u64(u64::MAX);
+        assert_eq!(0, instruction.accounts.len());
+        assert!(!instruction.data.is_empty());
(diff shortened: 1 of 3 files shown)


libraries/math/src/instruction.rs
```diff
@@ -167,3 +167,72 @@ pub fn noop() -> Instruction {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_precise_sqrt() {
+        let instruction = precise_sqrt(u64::MAX);
+        assert_eq!(0, instruction.accounts.len());
+        assert!(!instruction.data.is_empty());
+        assert!(!instruction.program_id.to_string().is_empty());
```
joncinque, line 180: the `program_id` is set to `crate::id()`, so this can check:
```
assert_eq!(instruction.program_id, crate::id());
```

libraries/math/src/instruction.rs
```diff
@@ -167,3 +167,72 @@ pub fn noop() -> Instruction {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_precise_sqrt() {
+        let instruction = precise_sqrt(u64::MAX);
+        assert_eq!(0, instruction.accounts.len());
+        assert!(!instruction.data.is_empty());
```
joncinque, line 179: For most of the pack / unpack tests, we test against what's actually expected in the buffer, so something like:
```
assert_eq!(instruction.data, [0, 255, 255, 255, 255, 255, 255, 255, 255]);
```
Alternatively, you can do:
```
assert_eq!(instruction.data, MathInstruction::PreciseSquareRoot { radicand: u64::MAX }.try_to_vec().unwrap());
```

joncinque (commented):
Love where this is going! Just a few little ways to improve the tests, and then we can integrate them

eventhorizonpl:
@joncinque PR updated. Thank you for the review.

joncinque:
Great, thanks for those refinements!

PR #2703, opened 2022-01-05
Cargo audit CI job

silas-x:
This PR adds cargo audit as a CI job and consists of two aspects:

1. Action that builds and runs cargo audit against the checked out repo to identify vulnerable crates in the cargo dependency tree
2. audit configuration file that takes arguments as if the job was running as cli. Good for keeping audit trail of ignores etc.

Context:
While building out more security automation for Mango, we noticed that the vulnerable crates found by cargo audit were upstream from Solana. Instead of ignoring them downstream, it'd be better to patch and monitor upstream going forward. Discussed this with SebastionBor before opening this PR.


joncinque:
This looks great! I decided to do this the same as in the Solana monorepo for consistency: https://github.com/solana-labs/solana/blob/master/ci/do-audit.sh

Please check out #2707, and let me know if you have any comments or suggestions

joncinque:
Closing since this was done in #2707

PR #2705, opened 2022-01-05
Decrease the stake pool size due to the new mem op syscall base cost

dmakarov:


dmakarov:
https://github.com/solana-labs/solana/pull/22101

joncinque (approved):
Looks good! I'll start coming up with a new solution for big pools :sweat:

PR #2706, opened 2022-01-06
token-2022: Move each extension into its own module

mvines:


mvines:
It's still yucky that adding a new extension requires modifying `ExtensionType` in two places but I can hold my nose on this.  Overall a single extension.rs is gonna turn into a mess once things like zk-token start getting turned into an extension, so I wanted to setup this kind of precedent early

joncinque (approved):
Looks good to me! I like the idea of including the mint and accompanying account extension in the same file.

PR #2707, opened 2022-01-06
CI: Add cargo audit, just like in the monorepo

joncinque:
#### Problem

We should know if SPL is using insecure crates.

#### Solution

Just like in the monorepo's https://github.com/solana-labs/solana/blob/master/ci/do-audit.sh, run `cargo audit` on every pull request.  Thankfully, SPL has a smaller number of dependencies, which makes it easier to manage.  The `ignore`s here are all from the monorepo, except for the `memmap` one, which comes from `honggfuzz`, only used in testing.

Inspired by #2703 -- @silas-x, what do you think of this?  I had a hard time finding info about some of the extra flags used for `cargo audit` in your PR, so if there's anything that you think is missing in this, please let me know!

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -15,6 +15,7 @@ jobs:
     needs:
       - rustfmt
       - clippy
+      - audit
       - cargo-build-test
     steps:
       - run: echo "Done"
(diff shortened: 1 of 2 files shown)


silas-x:
Hi @joncinque, looks good. Many ways to go about the config, but if you look in #2703 you can see that I created a config file with all the cargo audit arguments: https://github.com/solana-labs/solana-program-library/pull/2703/commits/bc285d38e170d4ae6e7f33c47c9e2b7ade7aa17a

CriesofCarrots (approved):
🎉 lgtm

joncinque:
> Hi @joncinque, looks good. Many ways to go about the config, but if you look in #2703 you can see that I created a config file with all the cargo audit arguments: [bc285d3](https://github.com/solana-labs/solana-program-library/commit/bc285d38e170d4ae6e7f33c47c9e2b7ade7aa17a)

@silas-x I was looking through that exactly -- did you change anything from the defaults? And if so, what / why?

joncinque:
Merging after offline discussion with @silas-x

silas-x:
> 

Just to come back to you on this, it's mainly default, and the audit.toml file includes what default is when running cargo audit vanilla in CLI. In a couple of downstream implementations I have included ignores, but not in this instance.

PR #2712, opened 2022-01-07
Add Confidential Transfer Extension boilerplate

mvines:
Merging the zk-token program into SPL Token will require a token extension that does all the things:
* additional account/mint data
* additional token instructions
* extension instruction processor

This PR attempts to carve out this path, using a simple "Todo" instruction.  Once accepted, follow-up patches will bring in the full zk-token program into this framework

cc: https://github.com/solana-labs/spl-zk-token/issues/91

### DIFF
--- a/token/program-2022/src/extension/confidential_transfer/mod.rs
+++ b/token/program-2022/src/extension/confidential_transfer/mod.rs
@@ -12,23 +12,23 @@ pub mod processor;
 /// Transfer auditor configuration
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
-pub struct MintConfidentialTransferAuditor {
+pub struct ConfidentialTransferAuditor {
     // TODO: inline `zk_token_program::state::Auditor` here
 }
 
-impl Extension for MintConfidentialTransferAuditor {
-    const TYPE: ExtensionType = ExtensionType::MintConfidentialTransferAuditor;
+impl Extension for ConfidentialTransferAuditor {
+    const TYPE: ExtensionType = ExtensionType::ConfidentialTransferAuditor;
     const ACCOUNT_TYPE: AccountType = AccountType::Mint;
 }
 
 /// Confidential account state
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
-pub struct AccountConfidentialState {
+pub struct ConfidentialTransferState {
     // TODO: inline `zk_token_program::state::ZkAccount` here
 }
 
-impl Extension for AccountConfidentialState {
-    const TYPE: ExtensionType = ExtensionType::AccountConfidentialState;
+impl Extension for ConfidentialTransferState {
+    const TYPE: ExtensionType = ExtensionType::ConfidentialTransferState;
     const ACCOUNT_TYPE: AccountType = AccountType::Account;
 }
--- a/token/program-2022/src/extension/confidential_transfer/mod.rs
+++ b/token/program-2022/src/extension/confidential_transfer/mod.rs
@@ -1,0 +1,34 @@
+use {
+    crate::extension::{AccountType, Extension, ExtensionType},
+    bytemuck::{Pod, Zeroable},
+};
+
+/// Confidential Transfer Extension instructions
+pub mod instruction;
+
+/// Confidential Transfer Extension processor
+pub mod processor;
+
+/// Transfer auditor configuration
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct MintConfidentialTransferAuditor {
+    // TODO: inline `zk_token_program::state::Auditor` here
+}
+
+impl Extension for MintConfidentialTransferAuditor {
+    const TYPE: ExtensionType = ExtensionType::MintConfidentialTransferAuditor;
+    const ACCOUNT_TYPE: AccountType = AccountType::Mint;
+}
+
+/// Confidential account state
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct AccountConfidentialState {
+    // TODO: inline `zk_token_program::state::ZkAccount` here
+}
+
+impl Extension for AccountConfidentialState {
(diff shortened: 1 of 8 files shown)


token/program-2022/src/extension/confidential_transfer/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+
+impl Extension for MintConfidentialTransferAuditor {
+    const TYPE: ExtensionType = ExtensionType::MintConfidentialTransferAuditor;
+    const ACCOUNT_TYPE: AccountType = AccountType::Mint;
+}
+
+/// Confidential account state
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct AccountConfidentialState {
```
joncinque, line 27: nit: how about `AccountConfidentialTransferState` or `AccountConfidentialTransfer` for consistency?

token/program-2022/src/instruction.rs
```diff
@@ -569,6 +569,12 @@ pub enum TokenInstruction {
         /// Maximum fee assessed on transfers
         maximum_fee: u64,
     },
+
+    /// The common instruction prefix for Confidential Transfer extension instructions.
+    ///
+    /// See `extension::confidential_transfer::instruction::ConfidentialTransferInstruction` for
+    /// further details about the extended instructions that share this instruction prefix
+    ConfidentialTransferExtension,
```
joncinque, line 577: I initially thought we would directly add all of the confidential instructions here, but your approach neatly encapsulates them.  It makes discoverability of instructions trickier, but not very much.

If we do it here, however, perhaps we should have all of the extension instructions use this same model, so all of the `TransferFee`-related instructions are in their own namespace, and so on.  Once people start forking SPL token, separate namespaces for extensions will make their work much nicer, since they can just quickly throw an `unimplemented!()` for a whole set of instructions if they choose.  What do you think?

joncinque (dismissed):
The general approach looks good to me!  One big question and a nit, but nothing that should stop this from going forward

token/program-2022/src/extension/confidential_transfer/mod.rs
mvines: I struggled with this and found the convention of a `Mint` or `Account` prefix made this naming tough.

I had "Transfer" in there for a while actually.  I settled on removing it because the tokens are being held confidentially in this structure, and adding "Transfer" to me implied that a transfer operation was pending.

I'm not sure I like of the options, even the existing name in the PR, too much frankly

token/program-2022/src/instruction.rs
mvines: > I initially thought we would directly add all of the confidential instructions here, but your approach neatly encapsulates them.

My main motivation was to avoid getting tangled in the manual unpacking code of the top-level instructions.  I had nightmares about having to juggle the ordinals as instructions from multiple extensions merged in at the same time!

Yep for sure, the TransferFee extension instructions probably could be implemented like this as well

token/program-2022/src/extension/confidential_transfer/mod.rs
mvines: `AccountConfidentialTransferAccountState` feels the most accurate, but that 2x Account is no good.

I like naming extensions `ConfidentialTransferAccountState` and `ConfidentialTransferAuditor`, which breaks the `Mint`/`Account` convention. I think I'd get over that though!

mvines: > I like naming extensions ConfidentialTransferAccountState and ConfidentialTransferAuditor

I pushed a new commit, 6eaa6d6, with these to show how that'd look

CriesofCarrots: Why exactly do we need that `Account` in there between `Transfer` and `State`?

token/program-2022/src/instruction.rs
CriesofCarrots: > perhaps we should have all of the extension instructions use this same model

I like that idea

token/program-2022/src/extension/confidential_transfer/mod.rs
CriesofCarrots: I actually don't mind losing the Mint/Account prefixes, since `ACCOUNT_TYPE` is defined for every Extension.
This is a radical suggestion that I'm not sure I love, but we could actually strip those prefixes from the existing extensions as well, leaving something like:
```
pub struct TransferFeeConfig {
    ...
}
impl Extension for TransferFeeConfig {
    const TYPE: ExtensionType = ExtensionType::TransferFeeConfig;
    const ACCOUNT_TYPE: AccountType = AccountType::Mint;
}

pub struct TransferFeeAmount {
    /// Amount withheld during transfers, to be harvested to the mint
    pub withheld_amount: PodU64,
}
impl Extension for TransferFeeAmount {
    const TYPE: ExtensionType = ExtensionType::TransferFeeAmount;
    const ACCOUNT_TYPE: AccountType = AccountType::Account;
}
```

```
pub struct CloseAuthority {
    /// Optional authority to close a mint
    pub close_authority: OptionalNonZeroPubkey,
}
impl Extension for CloseAuthority {
    const TYPE: ExtensionType = ExtensionType::CloseAuthority;
    const ACCOUNT_TYPE: AccountType = AccountType::Mint;
}
```

CriesofCarrots (dismissed):
Just bikeshedding from me. Bones look gtg

token/program-2022/src/extension/confidential_transfer/mod.rs
mvines: > Why exactly do we need that `Account` in there between `Transfer` and `State`?

Not sure! `ConfidentialTransferState` definitely feels better.

I like your radical suggestion, with the exception of `CloseAuthority` as this is now ambiguous due to how an Account already has a close authority.  In this case `MintCloseAuthority` seems quite appropriate.

CriesofCarrots: > I like your radical suggestion

I'll PR it so we can take a look

joncinque: > I actually don't mind losing the Mint/Account prefixes, since ACCOUNT_TYPE is defined for every Extension.

Agreed on this! And we can certainly keep it as `MintCloseAuthority` for that ambiguity

PR #2715, opened 2022-01-07
Strip extension account-type prefixes

CriesofCarrots:
Every extension documents its account type, so Mint/Account prefixes aren't really necessary. (Also, I can envision an extension that may have the same data structure for both Mint and Account implementations.)

This PR removes Mint and Account prefixes from TransferFee extensions.
`MintCloseAuthority` is kept as-is, since token Accounts already have close-authorities in their base state.


mvines (dismissed):
lgtm!

joncinque (approved):
nice!

PR #2716, opened 2022-01-08
token-2022: Harden StateWithExtensions unpacking

mvines:


token/program-2022/src/extension/mod.rs
```diff
@@ -160,26 +150,31 @@ fn type_and_tlv_indices<S: BaseState>(rest_input: &[u8]) -> Result<(usize, usize
...
+        if let Some((account_type_index, tlv_start_index)) = type_and_tlv_indices::<S>(rest)? {
+            let account_type = AccountType::try_from(rest[account_type_index])
+                .map_err(|_| ProgramError::InvalidAccountData)?;
+            check_account_type::<S>(account_type)?;
+            Ok(Self {
+                base,
+                tlv_data: &rest[tlv_start_index..],
+            })
+        } else {
+            Ok(Self {
```
mvines, line 173: A base mint/account is now handled

token/program-2022/src/extension/mod.rs
```diff
@@ -160,26 +150,31 @@ fn type_and_tlv_indices<S: BaseState>(rest_input: &[u8]) -> Result<(usize, usize
...
     /// Slice of data containing all TLV data, deserialized on demand
     tlv_data: &'data [u8],
 }
 impl<'data, S: BaseState> StateWithExtensions<'data, S> {
     /// Unpack base state, leaving the extension data as a slice
+    ///
+    /// Fails if the base state is not initialized.
     pub fn unpack(input: &'data [u8]) -> Result<Self, ProgramError> {
-        check_not_multisig(input)?;
+        check_min_len_and_not_multisig(input, S::LEN)?;
```
mvines, line 161: prevent panic if input is too small

token/program-2022/src/extension/mod.rs
```diff
@@ -209,41 +204,65 @@ pub struct StateWithExtensionsMut<'data, S: BaseState> {
...
 
-    /// Unpack the base state portion of the buffer without checking for initialization,
-    /// leaving the extension data as a serialized slice.
+    /// Unpack an uninitialized base state, leaving the extension data as a mutable slice
     ///
-    /// The base state of the struct may be totally unusable.
-    pub fn unpack_unchecked(input: &'data mut [u8]) -> Result<Self, ProgramError> {
-        check_not_multisig(input)?;
+    /// Fails if the base state has already been initialized.
+    pub fn unpack_uninitialized(input: &'data mut [u8]) -> Result<Self, ProgramError> {
```
mvines, line 238: "unchecked" is not really what we want IMO.  The calling code really just wants to unpack an account/mint that has yet to be initialized, so make this more explicit

token/program-2022/src/extension/mod.rs
```diff
@@ -160,26 +150,31 @@ fn type_and_tlv_indices<S: BaseState>(rest_input: &[u8]) -> Result<(usize, usize
 pub struct StateWithExtensions<'data, S: BaseState> {
     /// Unpacked base data
     pub base: S,
-    /// Unpacked account type
-    pub account_type: AccountType,
```
mvines, line 164: This field seems unnecessary.  The desired `BaseState` is validated on `unpack()` so the caller already knows the account type

joncinque: Yep, definitely

token/program-2022/src/extension/mod.rs
joncinque: Oops, definitely broke that case during a refactor at some point, nice catch

token/program-2022/src/extension/mod.rs
joncinque: You're absolutely right, and a much better name for this.  As I was doing some other work, I kept adding checks in here

token/program-2022/src/extension/mod.rs
```diff
@@ -209,41 +204,65 @@ pub struct StateWithExtensionsMut<'data, S: BaseState> {
...
-        let (account_type, tlv_data) = rest.split_at_mut(tlv_start_index);
-        Ok(Self {
-            base,
-            base_data,
-            account_type: &mut account_type[account_type_index..tlv_start_index],
-            tlv_data,
-        })
+        if base.is_initialized() {
+            return Err(ProgramError::InvalidAccountData);
+        }
```
joncinque, line 244: I went back and forth on including this check here, but I think it's the right call.  We need to have more checks in this function, as I found out while fixing another part.

If we do include the check, we *must* document whether each extension initializer requires a initialized mint or not, and clients must do them in the right order.

For example, in `process_initialize_close_mint_authority`, it'll either do `unpack_uninitialized` or `unpack` before modifying the state, and that will dictate whether the mint *must* be initialized or not.

The alternative is to create some sort of `unpack_either` which tries `unpack` and falls back to `unpack_uninitialized`

joncinque (approved):
When I saw this PR, my heart sank for a bit because I've been working on something similar.  Thankfully, they're mostly orthogonal :sweat_smile: 

My fix is around initialization of mixed types.  For example, someone can currently:
* initialize a mint extension -> initialize account extension
* initialize mint -> initialize account extension
* initialize mint extension -> initialize account
I couldn't finish it this week, but I'll have something out early next week for it.

There's only one big question around `unpack_either` (can't come up with a better name at the moment), but that can get done in a separate PR. Looks good!

token/program-2022/src/extension/mod.rs
mvines: It'd be nice to avoid an `unpack_either` if possible, as that's another level of code complexity that could be exploited.
But yeah let's discuss this more if/when your PR appears instead of in here :)

Conceptually it seems like all the extension initialization should happen before the base initialization, and then extension updates that may need to happen post base initialization are a different instruction

PR #2717, opened 2022-01-08
Fixup tests to use rent-exempt accounts

CriesofCarrots:
Rent-exempt requirement for accounts is coming down the pipe: https://github.com/solana-labs/solana/pull/22292
Several tests in this repository create rent-paying accounts. Since these account balances appear to be arbitrary, this PR fixes those tests up to use rent-exempt amounts.

@mvines , any concerns here? I didn't see any fundamental reason for these rent-paying accounts.

There are also some stake-pool tests that are failing. Will be addressed in a separate PR.


CriesofCarrots:
Merging. Let me know if I've missed something and need to ask for forgiveness and revert...

PR #2720, opened 2022-01-08
Fix up helper to create rent-exempt accounts

CriesofCarrots:
Rent-exempt requirement for accounts is coming down the pipe: solana-labs/solana#22292
The stake-pool helper `create_vote` creates a validator identity account that is rent-paying for no particular reason.


joncinque (approved):
thanks for finding and fixing this!

PR #2723, opened 2022-01-10
token: Add checks for mixed type initialization

joncinque:
#### Problem

Since the token program does not know anything about the buffers passed in before they're initialized, we can run into all sorts of strange situations, mainly arising from initializing mixed types, ie initializing a mint extension on a buffer that actually contains an account extension already.

#### Solution

Add more checks on initializing extensions and unpacking the buffer to make sure that it's correct.  This introduces a mapping from ExtensionType -> AccountType, which was not possible before with just the trait.

In later work, we'll also need to introduce a mapping ExtensionType -> ExtensionType for mint type to account type.

Note that this will mainly save us if we ever allow for account extensions to be initialized on their own.

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -71,12 +71,12 @@ fn get_extension_indices<V: Extension>(
     init: bool,
 ) -> Result<TlvIndices, ProgramError> {
     let mut start_index = 0;
-    let v_account_type = V::TYPE.get_associated_account_type();
+    let v_account_type = V::TYPE.get_account_type();
     while start_index < tlv_data.len() {
         let tlv_indices = get_tlv_indices(start_index);
         let extension_type =
             ExtensionType::try_from(&tlv_data[tlv_indices.type_start..tlv_indices.length_start])?;
-        let account_type = extension_type.get_associated_account_type();
+        let account_type = extension_type.get_account_type();
         // got to an empty spot, can init here, or move forward if not initing
         if extension_type == ExtensionType::Uninitialized {
             if init {
@@ -227,7 +227,7 @@ impl<'data, S: BaseState> StateWithExtensions<'data, S> {
 
     /// Unpack a portion of the TLV data as the desired type
     pub fn get_extension<V: Extension>(&self) -> Result<&V, ProgramError> {
-        if V::TYPE.get_associated_account_type() != S::ACCOUNT_TYPE {
+        if V::TYPE.get_account_type() != S::ACCOUNT_TYPE {
             return Err(ProgramError::InvalidAccountData);
         }
         let TlvIndices {
@@ -311,7 +311,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
                 tlv_data,
             };
             if let Some(extension_type) = state.get_first_extension_type()? {
-                let account_type = extension_type.get_associated_account_type();
+                let account_type = extension_type.get_account_type();
                 if account_type != S::ACCOUNT_TYPE {
                     return Err(TokenError::ExtensionBaseMismatch.into());
                 }
@@ -328,7 +328,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     }
 
     fn init_or_get_extension<V: Extension>(&mut self, init: bool) -> Result<&mut V, ProgramError> {
-        if V::TYPE.get_associated_account_type() != S::ACCOUNT_TYPE {
+        if V::TYPE.get_account_type() != S::ACCOUNT_TYPE {
             return Err(ProgramError::InvalidAccountData);
         }
         let TlvIndices {
@@ -381,7 +381,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     pub fn init_account_type(&mut self) -> Result<(), ProgramError> {
         if !self.account_type.is_empty() {
             if let Some(extension_type) = self.get_first_extension_type()? {
-                let account_type = extension_type.get_associated_account_type();
+                let account_type = extension_type.get_account_type();
                 if account_type != S::ACCOUNT_TYPE {
                     return Err(TokenError::ExtensionBaseMismatch.into());
                 }
@@ -462,7 +462,7 @@ impl From<ExtensionType> for [u8; 2] {
 }
 impl ExtensionType {
     /// Get the data length of the type associated with the enum
-    pub fn get_associated_type_len(&self) -> usize {
+    pub fn get_type_len(&self) -> usize {
         match self {
             ExtensionType::Uninitialized => 0,
             ExtensionType::TransferFeeConfig => pod_get_packed_len::<TransferFeeConfig>(),
@@ -482,7 +482,7 @@ impl ExtensionType {
     }
 
     /// Get the associated account type
-    pub fn get_associated_account_type(&self) -> AccountType {
+    pub fn get_account_type(&self) -> AccountType {
         match self {
             ExtensionType::Uninitialized => AccountType::Uninitialized,
             ExtensionType::TransferFeeConfig
@@ -504,7 +504,7 @@ pub fn get_account_len(extension_types: &[ExtensionType]) -> usize {
     let extension_size: usize = extension_types
         .iter()
         .map(|e| {
-            e.get_associated_type_len()
+            e.get_type_len()
                 .saturating_add(size_of::<ExtensionType>())
                 .saturating_add(pod_get_packed_len::<Length>())
         })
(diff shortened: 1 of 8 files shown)


token/program-2022/src/instruction.rs
```diff
@@ -29,6 +29,8 @@ pub enum TokenInstruction {
     /// Otherwise another party can acquire ownership of the uninitialized
     /// account.
     ///
+    /// All extensions must be initialized before calling this instruction.
```
joncinque, line 32: I now feel very strongly about this for now.  It'll work if we ever need to relax the requirement, but it simplifies a lot of things at the moment.

Consider this situation: I allocate a huge buffer, initialize some mint extensions, and then initialize the main mint.  The mint is now usable, people start making token accounts and using it.  Later on, I initialize another mint extension.  What happens then?

If we do decide to relax this in the future, my idea is to introduce a `FinalizeMintInitialization` instruction, which just writes the `AccountType` into the TLV buffer, and a mint isn't considered usable until that `AccountType` is written.  I had this working for a bit, and then `unpack_uninitialized` came in and simplified this a lot.

mvines: I'm totally onboard.  

The extension init functions doesn't even need to necessary fully initialize the extension, it just needs to reserve space in the TLV buffer and install sensible defaults. I've basically already run into this case with confidential transfers and it was fine --ideally the `ConfigureOmnibusAccount` instruction (from #2725) executes during mint extension initialization, but that's not really possible because one shouldn't be able to create an account for a given mint before the mint itself exists.  Instead `ConfigureOmnibusAccount` just needs to be run after mint init and all confidential transfers would fail until this happens, which is nbd.

token/program-2022/src/extension/mod.rs
```diff
@@ -47,52 +48,99 @@ impl TryFrom<usize> for Length {
...
+        length_start,
+        value_start,
+    }
+}
+
 /// Helper struct for returning the indices of the type, length, and value in
 /// a TLV entry
-struct TlvIndices(usize, usize, usize);
+struct TlvIndices {
+    pub type_start: usize,
```
mvines, line 65: yay, so much more readable!

token/program-2022/src/extension/mod.rs
```diff
@@ -401,6 +480,23 @@ impl ExtensionType {
             ExtensionType::MintPaddingTest => pod_get_packed_len::<MintPaddingTest>(),
         }
     }
+
+    /// Get the associated account type
+    pub fn get_associated_account_type(&self) -> AccountType {
```
mvines, line 485: wdyt about dropping "associated"?  When I see "associated_account" in a token context, my brain jumps to ATA immediately:
```suggestion
    pub fn get_account_type(&self) -> AccountType {
```

mvines (dismissed):

token/program-2022/src/instruction.rs
mvines: Except there's a catch! I do have a current need to add a new TLV extension to a Token Account, reallocing the data if needed to do so.  The use case:
1. User gets their first tokens from somebody, who creates their ATA
2. User now wants to enable confidential transfers.  They need to configure the confidential transfer account extension for their ATA.  It's not reasonable for the somebody from (1) to pay the rent for this extension, and it's also not reasonable to require the user (or their wallet) to close the ATA and recreate it.

So I do think we need a variant of `init_extension()` that reallocs if there's no room.

token/program-2022/src/extension/mod.rs
joncinque: Hehe I had the same thought, and then opted for consistency with `get_associated_type_len()`.  Happy to remove it

token/program-2022/src/instruction.rs
joncinque: >  Instead ConfigureOmnibusAccount just needs to be run after mint init and all confidential transfers would fail until this happens, which is nbd.

That makes sense, extra configuration afterwards is totally legit, especially in the case that requires an existing token account.

> So I do think we need a variant of init_extension() that reallocs if there's no room.

Let's go through the potential dangers with passing in different types:

* a non-extended token account -> no problem, we can check that based on length, then realloc for TLV and init the account type
* an account with extensions -> even easier, the account type is already set
* a mint with extensions -> easy to fail, the mint has the type written
* a non-extended mint -> will fail early based on length check
* an over-allocated mint with exactly 165 bytes -> this one's tricky, but `is_initialized` won't be set in the right place, so `Account::unpack` fails
* a multisig -> fails early based on length check

I went through a few of these cases during this PR, and since `Account::is_initialized` is past the 82nd byte, where a Mint ends, we're safe.

On a totally separate note, I thought we wouldn't need the omnibus account anymore since everything is contained within the token program.  Is the flow actually:
* create a normal token account
* initialize the confidential extension
* during `deposit`, move tokens to the omnibus account and get a confidential deposit into the confidential extension part of my token account
* during `withdraw`, move tokens from the omnibus account back to the non-confidential portion of my token account

mvines: >  I thought we wouldn't need the omnibus account anymore since everything is contained within the token program. Is the flow actually:

Yes that's the new extension-based zk-token flow I'm thinking of.  Some other rejected ideas included:
a. `deposit` and `withdraw` tokens to/from the mint itself (allowing the mint to act as a token account essentially).  It was a nice thought, but the risk of introducing new security issues by opening up an entirely new way to transfer/hold tokens seems pretty high here.  It also means that the confidential transfer extension would not work with legacy mints (USDC or Wrapped SOL)
b. `deposit` burns tokens, and `withdraw` mints them back.  This messes with the supply

joncinque: It seems like the only added benefit of the omnibus account is to know how many tokens are currently in "confidential" mode.  Otherwise, we could just move tokens between the non-confidential and confidential parts of the same account, without doing any mints / burns, right?

mvines: The omnibus account is the pool that all users dip into when they want to withdraw tokens from the confidential side.

joncinque: Maybe I'm misunderstanding something, but it seems like it's there more for bookkeeping, no?  What's the practical difference between:

1. transfer X unencrypted tokens to omnibus, add X encrypted tokens to my encrypted balance
2. directly remove X unencrypted tokens from my unencrypted balance, add X encrypted tokens to my encrypted balance

The advantage of 2) is one less write lock during deposit / withdrawal.

mvines: Ah so (2) does an implicit burn then on `deposit`?  The mint circulating supply remains unchanged, but it is now larger than the sum of all the visible token account balances.  Conversely `withdraw` does an implicit mint?

joncinque: Exactly, yes.  The key question becomes: how important is it that all unencrypted balances sum up to `supply`?

mvines: > how important is it that all unencrypted balances sum up to supply?

Hard to say. This definitely seems like it could be an avenue to easily FUD on a token.  Especially since I want confidential transfers to just work for existing tokens, so suddenly USDC or USDT might get attacked out of left field by the CT tinfoil hat people.

Parallelizing deposits/withdrawals from the confidential side is definitely tempting.  Perhaps allowing confidential transfers to just work for existing tokens is to bold, and we require the mint freeze authority to opt into confidential transfers (which introduces the need to be able to promote a base mint to an extension-enabled mint!).  

@samkim-crypto - wdyt about this proposal to remove the omnibus account entirely?

samkim-crypto: Yes, I agree with all the points. The only real use for the omnibus account is keeping track of how many tokens are encrypted and how many are not. Other than psychological factors, there doesn't seem to be a place that uses this information. In fact, this was how Greg and I were thinking about confidential tokens at first way back. I think it is a more natural solution from a technical perspective.

I do see the point about the actual circulating supply being lower than the mint `supply` could reflect negatively on an existing token though. This is really hard to say. We currently view token and zk-token accounts as two accounts storing the same tokens. But some people could view a token and its confidential variant as two separate tokens even if they are under the same mint. In this case, it is useful to keep track of the total supply of confidential tokens.

mvines: Thanks.  I researched elsewhere as well, I think we can probably go ahead with nuking the omnibus account.  To eliminate surprise, this means that existing tokens don't automatically get confidential token support.  

So in the end this doesn't save us from having an extension that's post initialized though, as `ConfigureOmnibusAccount` just turns into `ConfigureConfidentialTransfers`, which would be used by mints to opt-in 🙃

PR #2725, opened 2022-01-11
Confidential transfer extension

mvines:
Rewrite of the zk-token program as token 2022 extension.

TODO: Only a couple instructions have been implemented so far, to test out the new extension framework.  PRing this early to share what the code of a reasonably complex extension currently looks like

cc: https://github.com/solana-labs/spl-zk-token/issues/91

### DIFF
--- a/token/program-2022/src/extension/confidential_transfer/processor.rs
+++ b/token/program-2022/src/extension/confidential_transfer/processor.rs
@@ -1,19 +1,174 @@
 use {
-    super::instruction::*,
-    solana_program::{account_info::AccountInfo, entrypoint::ProgramResult, msg, pubkey::Pubkey},
+    crate::{
+        check_program_account,
+        extension::{
+            confidential_transfer::{instruction::*, *},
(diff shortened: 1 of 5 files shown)


token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -1,19 +1,183 @@
...
+}
+
+/// Processes an [ApproveAccount] instruction.
+fn process_approve_account(accounts: &[AccountInfo]) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+    let account_to_approve_info = next_account_info(account_info_iter)?;
+    let mint_info = next_account_info(account_info_iter)?;
+
+    check_program_account(account_to_approve_info.owner)?;
+    let account_to_approve_data = &mut account_to_approve_info.data.borrow_mut();
```
mvines, line 121: Having to hold a reference to the account data to keep rustc happy is annoying.  This is the only major wart I've found so far with the extension stuff!

PR #2727, opened 2022-01-11
fix small typo

biw:
Fixed a small typo that refers to a user as a "his" vs. "their" which is used everywhere else in the repo


CriesofCarrots (approved):
Thanks!

PR #2728, opened 2022-01-11
Allow the confidential transfer extension to block `CloseAccount`

mvines:
Token accounts with confidential transfers enabled may not be closed until their confidential side is also closed.  Plumb this hook into the normal `CloseAccount` instruction to motivate more discussion around the current extension API

### DIFF
--- a/token/program-2022/src/error.rs
+++ b/token/program-2022/src/error.rs
@@ -85,6 +85,9 @@ pub enum TokenError {
     /// Extension already initialized on this account
     #[error("Extension already initialized on this account")]
     ExtensionAlreadyInitialized,
+    /// An account can only be closed if its confidential balance is zero"
+    #[error("An account can only be closed if its confidential balance is zero")]
+    ConfidentialTransferStateHasBalance,
 }
 impl From<TokenError> for ProgramError {
     fn from(e: TokenError) -> Self {
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -2,6 +2,10 @@
 
 use crate::{
     error::TokenError,
+    extension::{
+        confidential_transfer::{self, ConfidentialTransferState},
+        StateWithExtensionsMut,
+    },
     instruction::{is_valid_signer_index, AuthorityType, TokenInstruction, MAX_SIGNERS},
     state::{Account, AccountState, Mint, Multisig},
 };
@@ -637,14 +641,18 @@ impl Processor {
         let dest_account_info = next_account_info(account_info_iter)?;
         let authority_info = next_account_info(account_info_iter)?;
 
-        let mut source_account = Account::unpack(&source_account_info.data.borrow())?;
-        if !source_account.is_native() && source_account.amount != 0 {
+        let mut source_account_data = source_account_info.data.borrow_mut();
+        let mut source_account =
+            StateWithExtensionsMut::<Account>::unpack(&mut source_account_data)?;
+        if !source_account.base.is_native() && source_account.base.amount != 0 {
             return Err(TokenError::NonNativeHasBalance.into());
         }
 
         let authority = source_account
+            .base
             .close_authority
-            .unwrap_or(source_account.owner);
+            .unwrap_or(source_account.base.owner);
+
         Self::validate_owner(
             program_id,
             source_account_info.key,
@@ -653,15 +661,23 @@ impl Processor {
             account_info_iter.as_slice(),
         )?;
 
+        if let Ok(confidential_transfer_state) =
+            source_account.get_extension_mut::<ConfidentialTransferState>()
+        {
+            if !confidential_transfer_state.closable() {
+                return Err(TokenError::ConfidentialTransferStateHasBalance.into());
+            }
+        }
+
         let dest_starting_lamports = dest_account_info.lamports();
         **dest_account_info.lamports.borrow_mut() = dest_starting_lamports
             .checked_add(source_account_info.lamports())
             .ok_or(TokenError::Overflow)?;
 
         **source_account_info.lamports.borrow_mut() = 0;
-        source_account.amount = 0;
+        source_account.base.amount = 0;
 
-        Account::pack(source_account, &mut source_account_info.data.borrow_mut())?;
+        source_account.pack_base();
 
         Ok(())
     }
@@ -874,7 +890,7 @@ impl Processor {
                 unimplemented!();
             }
             TokenInstruction::ConfidentialTransferExtension => {
-                crate::extension::confidential_transfer::processor::process_instruction(
+                confidential_transfer::processor::process_instruction(
                     program_id,
                     accounts,
                     &input[1..],
@@ -977,6 +993,9 @@ impl PrintProgramError for TokenError {
             }
             TokenError::ExtensionAlreadyInitialized => {
                 msg!("Error: Extension already initialized on this account")
+            }
+            TokenError::ConfidentialTransferStateHasBalance => {
+                msg!("Error: An account can only be closed if its confidential balance is zero")
             }
         }
     }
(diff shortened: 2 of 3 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -300,8 +329,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     }
 
     /// Packs base state data into the base data portion
-    pub fn pack_base(&mut self, new_base: S) {
-        self.base = new_base;
+    pub fn pack_base(&mut self) {
```
mvines, line 332: I found it strange that `pack_base()` made the caller provider a new base when `unpack()` already populated `self.base`.  The user seems more likely to simply modify `self.base` then copy it elsewhere for modification

token/program-2022/src/extension/mod.rs
```diff
@@ -204,6 +204,35 @@ pub struct StateWithExtensionsMut<'data, S: BaseState> {
     tlv_data: &'data mut [u8],
 }
 impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
+    /// Resolves a single extension
+    ///
+    /// Warning: performs no validation of the `BaseState`
+    pub fn resolve_single_extension_mut<V: Extension>(
```
mvines, line 210: This `fn` name sucks, but essentially the story of me attempting to modify `process_close_account()` as shown in this PR goes like this:
1. Port `process_close_account()` to use `StateWithExtensionsMut`.  This initially motivated the `pack_base()` change in this PR and it worked great.
2. It worked great until I ran `test_close_account_dups` and it failed on me due to a double borrow, as `StateWithExtensionsMut` holds a mut reference to the account data and this breaks when the close authority is the source account.

(2) caused me to reassess (1) -- porting existing token instructions to use `StateWithExtensionsMut` is risky in general, and instead I really just wanted to resolve a single extension out of a base account that had already been validated.  I didn't want to call `StateWithExtensionsMut::unpack()` on an already unpacked account, and burn more compute units.

wdyt @joncinque?

joncinque: I was thinking that every instruction would get ported to use `StateWithExtensionsMut`, and that we'd be careful about double borrows and drop as needed.  I may be missing something though -- how did this function help your particular situation?  Since `resolve` is doing a mutable borrow on some part of the data anyway, it's still possible to double-borrow at some point.  So this would be something like: `unpack`, do stuff with base + extension, `pack_base`, `drop`, borrow data for something else, `resolve`, do stuff with extension?

Overall, I'd prefer to avoid having this, but if there's a situation that can't get around it, then by all means let's do it.

Man, it's going to be tricky to write tests to cover all of these cases.  Welp, something new to keep me up at night!

Also, let's add a borrow check to `validate_owner` to put at least one worry to rest.

token/program-2022/src/extension/mod.rs
joncinque: Makes more sense, definitely!

token/program-2022/src/extension/mod.rs
mvines: > I was thinking that every instruction would get ported to use StateWithExtensionsMut, and that we'd be careful about double borrows and drop as needed

I was hoping we'd only do it on demand actually, and only if strictly needed.  The current token code is battle hardened with some very subtle edge cases. We have decent code coverage but still seems risky.

But for the issue at hand, in the `process_close_account` impl:
1. First the `source_account` is unpacked to figure out if `source_account.owner` should be used as the authority (https://github.com/solana-labs/solana-program-library/blob/0c9f962a7b78be4f99c07a48ea553d3fe76fbb34/token/program-2022/src/processor.rs#L629).  This is the natural place to replace `Account::unpack()` with `StateWithExtensionsMut::unpack()
2. But now `Self::validate_owner` bombs with an immutable borrow of the source_account data.  

This is avoidable in the current `StateWithExtensionsMut`/extensions API by dropping the decoded source account and then unpacking it again, after calling `Self::validate_owner` but that's barf.  Logically at least in this case I'm quite happy to not mess with the bulk of the `process_close_account` code and potentially introduce new bugs, I just want to add one additional simple boring check

joncinque: > I was hoping we'd only do it on demand actually, and only if strictly needed.

Sorry, I should have been clearer.  What I really meant was, "every instruction that is impacted by an extension", and thinking that eventually every instruction would eventually be impacted by an extension.  That may not be the case for awhile though.

> Logically at least in this case I'm quite happy to not mess with the bulk of the process_close_account code and potentially introduce new bugs, I just want to add one additional simple boring check

Sorry, I should have just read the code first.  I late-edited my previous comment instead of adding a new one.  In this particular circumstance, if you use `StateWithExtensionsMut` and move the check up, wouldn't that solve your issue?  As I mentioned I'd prefer to avoid adding this until it's really necessary to avoid a double unpack on the base

mvines: > What I really meant was, "every instruction that is impacted by an extension", and thinking that eventually every instruction would eventually be impacted by an extension.

I realized half way into my bike ride that every instruction will need to use `StateWithExtensions` after all, as the existing `unpack`s will fail for non-base states! Oh well.

I'll push a new version of this PR with `process_close_account` rewritten with `StateWithExtensionsMut` to demo the issue better later tonight.  I thinking what might help is to break up `StateWithExtensionsMut::unpack` into two things: one that just unpacks the base mint, and one that can be later used to grab a mutable reference to one or more extensions.

mvines: This thread has gotten a little long, I'll close it and post fresh comments at the new main point of interest

token/program-2022/src/extension/mod.rs
mvines: I landed this change independently

token/program-2022/src/processor.rs
```diff
@@ -626,30 +630,49 @@ impl Processor {
...
             authority_info,
             account_info_iter.as_slice(),
         )?;
 
+        drop(source_account);
+        drop(source_account_data);
+
+        let mut source_account_data = source_account_info.data.borrow_mut();
+        let mut source_account =
+            StateWithExtensionsMut::<Account>::unpack(&mut source_account_data)?;
```
mvines, line 656: All tests pass with this latest PR, and this section points out the problem: the `source_account` can't be `borrow_mut()` until after `Self::validate_owner()` is called, but needs to be `borrow()` prior.  And this workaround causes the base account to be unpacked again.

I ran into a similar issue with the initial zk-token program, which spawned [PodAccountInfoData](https://github.com/solana-labs/spl-zk-token/blob/c0b843ee9ecfb7e017e1662f5dc486c06e0723ee/program/src/pod.rs#L93-L118) that features an `into_mut()` function to promote it into an [PodAccountInfoDataMut](https://github.com/solana-labs/spl-zk-token/blob/c0b843ee9ecfb7e017e1662f5dc486c06e0723ee/program/src/pod.rs#L150-L167).  Maybe the `get_extension()` function could return a similarly wrapped struct, and `StateWithExtensionsMut::pack_base()` only borrows data mutably internally to pack the base.

joncinque: I put in #2731 to address the validate owner double-borrow annoyance.

On the other hand, an `into_mut()` function could be really useful.  I'm game to have `StateWithExtensions*` take `AccountInfo` instead.

token/program-2022/src/error.rs
```diff
@@ -85,6 +85,9 @@ pub enum TokenError {
     /// Extension already initialized on this account
     #[error("Extension already initialized on this account")]
     ExtensionAlreadyInitialized,
+    /// An account can only be closed if its confidential balance is zero"
+    #[error("An account can only be closed if its confidential balance is zero")]
+    ConfidentialTransferStateHasBalance,
```
mvines, line 90: I don't love how adding extension-specific errors require reaching into `src/error.rs` and `src/processor.rs` but 🙈.

PR #2732, opened 2022-01-12
fix: typo in binary_option.py

eltociear:
Initalize -> Initialize


joncinque:
Thanks for the contribution!

joncinque:
Although now I'm just noticing that `Clinet` is misspelled right next to it :sweat_smile: \

PR #2733, opened 2022-01-13
token: Refactor transfer fee instructions + processor

joncinque:
#### Problem

It's cleaner to refactor all extension-related instructions into their own separate module.

#### Solution

Move all of the transfer fee instructions and processor into new modules.

My main questions here:

* does `TransferCheckedWithFee` belong in the transfer fee module or in the top-level?  I went with putting it in the module
* the instruction packing / unpacking was already done manually, so I opted to not do it with `Pod`s

The rest is boiler-plate, no logic changes


mvines (approved):

CriesofCarrots (approved):
So far, I like extension-specific instructions like `TransferCheckedWithFee` in the extension module.

PR #2734, opened 2022-01-13
Port the remainder of the zk-token program to the ConfidentialTransfer extension

mvines:
This brings over the bulk of the zk-token program.  The removal of the omnibus account and the separate zk-token PDA accounts nicely simplified parts of the original code. 

Follow up PRs will include:
* Resolving the TODO around account reallocation in the `ConfigureAccount` instruction
* Bringing over the zk-token unit tests, and likely fixing bugs as a result

cc: #https://github.com/solana-labs/spl-zk-token/issues/91

### DIFF
--- a/token/program-2022/src/extension/confidential_transfer/processor.rs
+++ b/token/program-2022/src/extension/confidential_transfer/processor.rs
@@ -1,94 +1,63 @@
 use {
     crate::{
         check_program_account,
+        error::TokenError,
         extension::{
             confidential_transfer::{instruction::*, *},
             StateWithExtensions, StateWithExtensionsMut,
         },
-        id,
         processor::Processor,
         state::{Account, Mint},
-        tools::account::create_pda_account,
     },
     solana_program::{
         account_info::{next_account_info, AccountInfo},
         entrypoint::ProgramResult,
+        instruction::Instruction,
         msg,
         program_error::ProgramError,
-        program_pack::Pack,
         pubkey::Pubkey,
-        sysvar::{rent::Rent, Sysvar},
+        sysvar::instructions::get_instruction_relative,
+    },
+    solana_zk_token_sdk::{
+        zk_token_elgamal::{ops, pod},
+        zk_token_proof_program,
     },
 };
 
-/// Processes an [InitializeAuditor] instruction.
-fn process_initialize_auditor(
+fn decode_proof_instruction<T: Pod>(
+    expected: ProofInstruction,
+    instruction: &Instruction,
+) -> Result<&T, ProgramError> {
+    if instruction.program_id != zk_token_proof_program::id()
+        || ProofInstruction::decode_type(&instruction.data) != Some(expected)
+    {
+        msg!("Unexpected proof instruction");
+        return Err(ProgramError::InvalidInstructionData);
+    }
+
+    ProofInstruction::decode_data(&instruction.data).ok_or(ProgramError::InvalidInstructionData)
+}
+
+/// Processes an [InitializeMint] instruction.
+fn process_initialize_mint(
     accounts: &[AccountInfo],
-    auditor: &ConfidentialTransferAuditor,
+    ct_mint: &ConfidentialTransferMint,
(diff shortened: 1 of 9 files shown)


token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -128,18 +173,489 @@ fn process_approve_account(accounts: &[AccountInfo]) -> ProgramResult {
...
+            return Err(TokenError::AccountFrozen.into());
+        }
+
+        if token_account.base.mint != *mint_info.key {
+            return Err(TokenError::MintMismatch.into());
+        }
+
+        // Wrapped SOL deposits are not supported because lamports cannot be vanished.
+        assert!(!token_account.base.is_native());
+        token_account.base.amount = token_account
```
mvines, line 282: It's slightly unsettling to vanish tokens...

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -128,18 +173,489 @@ fn process_approve_account(accounts: &[AccountInfo]) -> ProgramResult {
...
+            return Err(TokenError::AccountFrozen.into());
+        }
+
+        if receiver_token_account.base.mint != *mint_info.key {
+            return Err(TokenError::MintMismatch.into());
+        }
+
+        // Wrapped SOL deposits are not supported because lamports cannot be apparated.
+        assert!(!receiver_token_account.base.is_native());
+        receiver_token_account.base.amount = receiver_token_account
```
mvines, line 424: Likewise for apparating tokens...

mvines:
@samkim-crypto / @joncinque - you're welcome to check this PR out!  It's mostly https://github.com/solana-labs/spl-zk-token/blob/master/program/src/instruction.rs and https://github.com/solana-labs/spl-zk-token/blob/master/program/src/processor.rs in a slightly different form.

I imagine we'll want to do a more full end-to-end review once I plug in the tests and fix up the configure account TODO though

mvines:
> We don't have to worry about it now, but we would ultimately need a new set of confidential token instructions related to fees as well, right? Like `InitializeTransferFeeConfig`, `TransferWithFee`, etc. and analogous zk-token fee extensions to mint and accounts. The proof verification can be a separate `VerifyFee` zk-token-proof instruction.

What additional account and mint state fields will we need for confidential fees?  I was generally thinking we wouldn’t need new instructions, and the confidential transfer extension would detect the TransferFee extension in the mint to determine if fees applied.  You can flesh out all the details here?

samkim-crypto:
Yes, I think transfer is fine. It will get the encrypted transfer amount from the `ProofInstruction::VerifyTransfer` instruction as before and if TransferFee extension is detected, it will additionally check that `ProofInstruction::VerifyFee` is present and get the encrypted fee info from there.

The `ConfidentialTransferInstruction::InitializeMint` processor should perhaps check that if `TransferFee` extension is detected, then withdraw authority account is already extended for confidential tokens?

I think the only real extensions we would actually need are places related to the encrypted withheld fees in the accounts and mints: https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/src/extension/transfer_fee/mod.rs#L52. I do not see front-running issues here, so we would just need a single `pod::ElGamalCiphertext` for withheld_fees.

The way to withdraw withheld tokens will be the same as in regular tokens, but just using ElGamal ciphertext arithmetic. It seems like these should need separate instructions, but maybe there is a more elegant way?

mvines:
There's a little bit of mixing between the TransferFee extension and the ConfidentialTransfer extension for sure, let's move this to different issue to continue as I'm about to close this PR once CI is green!

> I do not see front-running issues here, so we would just need a single `pod::ElGamalCiphertext` for withheld_fees.

One way could be to add the `withheld_fees` field into the `TransferFeeConfig` mint extension, and add the additional instruction into the TransferFee extension. Or into the ConfidentialTransfer extension! I don't see much value in creating a third ConfidentialTransferTransferFee extension though

PR #2736, opened 2022-01-14
Add GetAccountDataSize implementation

CriesofCarrots:
The ATA program is going to need to know the minimum length of new accounts based on required extensions. This PR plumbs in the GetAccountDataSize instruction for ATA to CPI.

Couple notes:
- `get_account_extensions()` doesn't do any duplicate checking of either input Mint extensions or output Account extensions at the moment. We don't intend to support extensions that can exist in a Mint as duplicate, and currently none of the Mint extensions on the horizon have overlapping Account extension requirements, so I chose to make this helper as simple as possible. But we can update later if either of those predicates turns out to be false.
- the test for `process_get_account_data_size` is not complete, but I think the extension cases will be easier to write once InitializeMint is implemented

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -518,6 +518,26 @@ pub fn get_account_len(extension_types: &[ExtensionType]) -> usize {
     }
 }
 
+/// Get the list of required AccountType::Account ExtensionTypes based on a set of
+/// AccountType::Mint ExtensionTypes
+pub fn get_account_extensions(mint_extension_types: &[ExtensionType]) -> Vec<ExtensionType> {
+    let mut account_extension_types = vec![];
+    for extension_type in mint_extension_types {
+        #[allow(clippy::single_match)]
+        match extension_type {
+            ExtensionType::TransferFeeConfig => {
+                account_extension_types.push(ExtensionType::TransferFeeAmount);
+            }
+            #[cfg(test)]
+            ExtensionType::MintPaddingTest => {
+                account_extension_types.push(ExtensionType::AccountPaddingTest);
+            }
+            _ => {}
+        }
+    }
+    account_extension_types
+}
+
 /// Trait for base states, specifying the associated enum
 pub trait BaseState: Pack + IsInitialized {
     /// Associated extension type enum, checked at the start of TLV entries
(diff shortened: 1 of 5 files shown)


token/program-2022/src/processor.rs
```diff
@@ -6251,4 +6286,83 @@ mod tests {
...
+        .unwrap();
+
+        set_expected_data(get_account_len(&[]).to_le_bytes().to_vec());
+        do_process_instruction(
+            get_account_data_size(&program_id, &mint_key).unwrap(),
+            vec![&mut mint_account],
+        )
+        .unwrap();
+
+        // TODO: Extended mint
```
joncinque, line 6315: I'll have this part done tomorrow so we can actually test this part :smile:

token/program-2022/src/extension/mod.rs
```diff
@@ -518,6 +518,26 @@ pub fn get_account_len(extension_types: &[ExtensionType]) -> usize {
...
+            #[cfg(test)]
+            ExtensionType::MintPaddingTest => {
+                account_extension_types.push(ExtensionType::AccountPaddingTest);
+            }
+            _ => {}
+        }
+    }
+    account_extension_types
+}
+
```
joncinque, line 540: nit: wdyt about moving this into `ExtensionType`? I did that for `get_account_len` and it felt cleaner

joncinque (approved):
Looks good!

token/program-2022/src/processor.rs
```diff
@@ -1040,6 +1071,10 @@ mod tests {
             }
             solana_program::entrypoint::SUCCESS
         }
+
+        fn sol_set_return_data(&mut self, data: &[u8]) {
+            assert_eq!(&*EXPECTED_DATA.read().unwrap(), data)
+        }
```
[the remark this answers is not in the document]
CriesofCarrots: Yes, I think so! I'll upstream

token/program-2022/src/extension/mod.rs
CriesofCarrots: I'm not opposed to that. What's the fn signature of your new `get_account_len`?

joncinque: Same as before: `pub fn get_account_len<S: BaseState>(extension_types: &[ExtensionType]) -> usize` Although now that I see it, should've done `Self` instead of `ExtensionType`.
https://github.com/solana-labs/solana-program-library/pull/2739/files#diff-1fd54e31043386c786ab6a3edf57cd653cf69082500a805e9dd50f550d2f260bR483-R505

CriesofCarrots: Yeah, it looks pretty nice. Will push.

PR #2739, opened 2022-01-14
token-2022: Basic integration with rust client

joncinque:
#### Problem

We want to use the rust client to simplify the 2022 tests.

#### Solution

Use it!  A few notes:

* while adding extension support, it borked on the base case of no extensions, so I fixed that up
* the `no-entrypoint` feature took me about an hour of debugging trying to figure out why the program was doing nothing

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -481,7 +481,7 @@ impl ExtensionType {
     }
 
     /// Get the required account data length for the given ExtensionTypes
-    pub fn get_account_len<S: BaseState>(extension_types: &[ExtensionType]) -> usize {
+    pub fn get_account_len<S: BaseState>(extension_types: &[Self]) -> usize {
         if extension_types.is_empty() {
             S::LEN
         } else {
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -480,6 +480,30 @@ impl ExtensionType {
         }
     }
 
+    /// Get the required account data length for the given ExtensionTypes
+    pub fn get_account_len<S: BaseState>(extension_types: &[ExtensionType]) -> usize {
+        if extension_types.is_empty() {
+            S::LEN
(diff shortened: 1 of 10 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -480,6 +480,30 @@ impl ExtensionType {
         }
     }
 
+    /// Get the required account data length for the given ExtensionTypes
+    pub fn get_account_len<S: BaseState>(extension_types: &[Self]) -> usize {
+        if extension_types.is_empty() {
+            S::LEN
```
CriesofCarrots, line 486: Ah, yay, this should unblock my initial ATA PR :)

token/rust/src/token.rs
```diff
@@ -97,40 +97,39 @@ where
...
+    #[allow(clippy::too_many_arguments)]
     pub async fn create_mint<'a, S2: Signer>(
         client: Arc<dyn ProgramClient<T>>,
         payer: S,
         mint_account: &'a S2,
         mint_authority: &'a Pubkey,
         freeze_authority: Option<&'a Pubkey>,
         decimals: u8,
+        extension_types: &'a [ExtensionType],
+        extension_instructions: &'a [Instruction],
```
CriesofCarrots, line 109: Longer term, I don't love the ergonomics of the user/client having to create the extension instructions themselves; ix/tx building is one of the main purposes of the rust client, imo.
Perhaps we move to having a client specific enum that holds ExtensionTypes+required init data?
That can be a separate PR, however.

CriesofCarrots (approved):
One comment from me, but let's land this as-is (plus rebase) and then iterate (so I get the account LEN fix ;) )

token/rust/src/token.rs
joncinque: That makes a lot of sense to me.  I'm down to have a separate enum for extension init data!  If we make separate types for each set of init data, and then bundle them all into a separate enum, it should be pretty slick.

PR #2741, opened 2022-01-14
Add MigrateMultisigNative Instruction

zicklag:
This is a work-in-progress but basically working implementation of a new instruction for the token program that will allow you to wrap native SOL that has been sent directly to a multisig address.

I'm opening an issue to discuss momentarily.


zicklag:
Closing as it this was argued against in the discussion: #2742 .

PR #2743, opened 2022-01-14
token-2022: Support extensions in InitializeAccount variations

CriesofCarrots:
InitializeAccount variations should initialize extensions required by the mint.
This PR adds that support.

Base and native mint cases are covered by existing test suite. The extended-mint cases still need testing, but I wanted to PR this now because: (a) that testing will be a lot easier when InitializeMint implementations for extensions are done; and (b) I wanted to see what you think of `StateWithExtensionsMut::init_extension_from_type()` and the error-type change in `StateWithExtensionsMut::unpack_uninitialized()` now. Thoughts?

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -373,6 +373,38 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
         self.init_or_get_extension(true)
     }
 
+    /// Packs the relevant Extension of an ExtensionType into an open slot if not already
+    /// found in the data buffer, otherwise overwrites the Extension
+    pub fn init_extension_from_type(
+        &mut self,
+        extension_type: ExtensionType,
+    ) -> Result<(), ProgramError> {
+        match extension_type {
+            ExtensionType::Uninitialized => Ok(()),
+            ExtensionType::TransferFeeConfig => {
+                self.init_extension::<TransferFeeConfig>().map(|_| ())
+            }
+            ExtensionType::TransferFeeAmount => {
+                self.init_extension::<TransferFeeAmount>().map(|_| ())
+            }
+            ExtensionType::MintCloseAuthority => {
+                self.init_extension::<MintCloseAuthority>().map(|_| ())
+            }
+            ExtensionType::ConfidentialTransferMint => self
+                .init_extension::<ConfidentialTransferMint>()
+                .map(|_| ()),
+            ExtensionType::ConfidentialTransferAccount => self
+                .init_extension::<ConfidentialTransferAccount>()
+                .map(|_| ()),
+            #[cfg(test)]
+            ExtensionType::AccountPaddingTest => {
+                self.init_extension::<AccountPaddingTest>().map(|_| ())
+            }
+            #[cfg(test)]
+            ExtensionType::MintPaddingTest => self.init_extension::<MintPaddingTest>().map(|_| ()),
+        }
+    }
+
     /// Write the account type into the buffer, done during the base
     /// state initialization
     /// Noops if there is no room for an extension in the account, needed for
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -295,7 +295,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
         let (base_data, rest) = input.split_at_mut(S::LEN);
         let base = S::unpack_unchecked(base_data)?;
         if base.is_initialized() {
-            return Err(ProgramError::InvalidAccountData);
+            return Err(TokenError::AlreadyInUse.into());
         }
         if let Some((account_type_index, tlv_start_index)) = type_and_tlv_indices::<S>(rest)? {
             let account_type = AccountType::try_from(rest[account_type_index])
@@ -783,7 +783,7 @@ mod test {
         // unpack uninitialized will now fail because the Mint is now initialized
         assert_eq!(
             StateWithExtensionsMut::<Mint>::unpack_uninitialized(&mut buffer.clone()),
-            Err(ProgramError::InvalidAccountData),
+            Err(TokenError::AlreadyInUse.into()),
         );
 
         // check unpacking
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -373,9 +373,10 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
         self.init_or_get_extension(true)
     }
 
-    /// If `extension_type` is an Account-associated ExtensionType, this method packs the relevant
-    /// Extension of an ExtensionType into an open slot if not already found in the data buffer.
-    /// For all other ExtensionTypes, this is a no-op.
+    /// If `extension_type` is an Account-associated ExtensionType that requires initialization on
+    /// InitializeAccount, this method packs the relevant Extension of an ExtensionType into an
+    /// open slot if not already found in the data buffer. For all other ExtensionTypes, this is a
+    /// no-op.
     pub fn init_account_extension_from_type(
         &mut self,
         extension_type: ExtensionType,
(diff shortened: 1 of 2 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -295,7 +295,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
         let (base_data, rest) = input.split_at_mut(S::LEN);
         let base = S::unpack_unchecked(base_data)?;
         if base.is_initialized() {
-            return Err(ProgramError::InvalidAccountData);
+            return Err(TokenError::AlreadyInUse.into());
```
joncinque, line 298: Hehe I have this change in my current branch too!

token/program-2022/src/extension/mod.rs
```diff
@@ -373,6 +373,38 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
         self.init_or_get_extension(true)
     }
 
+    /// Packs the relevant Extension of an ExtensionType into an open slot if not already
+    /// found in the data buffer, otherwise overwrites the Extension
```
joncinque, line 377: Currently, `init_extension` doesn't actually overwrite the data part, but maybe it should do a `fill(0)` into the data portion to be safer?
Or even better, what if it initializes it to `Type::default()`?

token/program-2022/src/extension/mod.rs
```diff
@@ -373,6 +373,38 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
...
+    pub fn init_extension_from_type(
+        &mut self,
+        extension_type: ExtensionType,
+    ) -> Result<(), ProgramError> {
+        match extension_type {
```
joncinque, line 382: Since this is focused for account initialization, I'm thinking it should only do something for the Account ones, ie. just `TransferFeeConfig` and `AccountPaddingTest`, what do you think?

token/program-2022/src/processor.rs
```diff
@@ -116,24 +116,29 @@ impl Processor {
...
 
-        account.mint = *mint_info.key;
-        account.owner = *owner;
-        account.delegate = COption::None;
-        account.delegated_amount = 0;
-        account.state = AccountState::Initialized;
+        let required_extensions = Self::get_required_account_extensions(mint_info)?;
+        for extension in required_extensions {
+            account.init_extension_from_type(extension)?;
+        }
```
joncinque, line 122: This isn't part of our original spec, but what do you think about a check to make sure that there's no padding at the end of the account buffer?

token/program-2022/src/processor.rs
```diff
@@ -947,6 +947,16 @@ impl Processor {
         }
         Ok(())
     }
+
+    fn get_required_account_extensions(
+        mint_account_info: &AccountInfo,
```
joncinque, line 952: nit: how about this takes in `StateWithExtensions::<Mint>`, and use the unpack done in line 115? that way we avoid doing two unpacks

joncinque (commented):
Looking good! Just a few comments and questions, but for the most part this is great

token/program-2022/src/extension/mod.rs
CriesofCarrots: Oh yes, I like the idea of initializing to `Type::default()`

token/program-2022/src/extension/mod.rs
CriesofCarrots: Just depends whether this would be useful for Mint initialization too... I'm guessing not, since Mint extension inits will have their own instructions?

token/program-2022/src/processor.rs
CriesofCarrots: Makes a lot of sense.

token/program-2022/src/processor.rs
CriesofCarrots: How about I instead remove this check:
https://github.com/solana-labs/solana-program-library/blob/3e1dd733968a74c19b7df5b0eda1663da79a2377/token/program-2022/src/processor.rs#L114-L116
and map the error type on the unpack in get_required_account_extensions to TokenError::InvalidMint?

That would reduce to one unpack for now. We could reassess if there are other methods that end up calling `get_required_account_extensions` and have other mint needs.

token/program-2022/src/extension/mod.rs
joncinque: Yeah that's the idea right now. I'd prefer to limit this dynamic stuff to avoid any potential vulnerabilities.

token/program-2022/src/processor.rs
joncinque: works for me!

token/program-2022/src/extension/mod.rs
CriesofCarrots: sg

token/program-2022/src/extension/mod.rs
CriesofCarrots: Unfortunately, the confidential-transfer types don't implement Default. I'll go with `fill(0)` for now.

CriesofCarrots: Ohh, thanks!
That said, I'm going to forgo any changes to that behavior in this PR because I think whether we want to remove [this protection](https://github.com/solana-labs/solana-program-library/blob/3e1dd733968a74c19b7df5b0eda1663da79a2377/token/program-2022/src/extension/mod.rs#L89-:90) is a separate conversation

CriesofCarrots:
@joncinque , I think this is ready for another look.
To come separately:
- A bunch of processor.rs and ProgramTest tests, when https://github.com/solana-labs/solana-program-library/pull/2744 is in
- A PR removing the AlreadyInitialized check (and reverting the comment changes), for discussion

token/program-2022/src/extension/mod.rs
joncinque: `Zeroable` is very different from `Default`, since a type can implement `Default` with some non-zero values, which may be better in some cases.

The bytemuck docs specify the difference: https://docs.rs/bytemuck/1.7.3/bytemuck/trait.Zeroable.html
"An all-zeroes value may or may not be the same value as the Default value of the type."

It doesn't need to get done in this PR and with these extensions, but there may be an extension that requires default initialization rather than all-zeroes initialization.

token/program-2022/src/extension/mod.rs
```diff
@@ -368,11 +368,36 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
...
+        if extension_type.get_account_type() != AccountType::Account {
+            return Ok(());
+        }
+        match extension_type {
+            ExtensionType::TransferFeeAmount => {
+                self.init_extension::<TransferFeeAmount>().map(|_| ())
+            }
+            ExtensionType::ConfidentialTransferAccount => self
+                .init_extension::<ConfidentialTransferAccount>()
+                .map(|_| ()),
```
joncinque, line 392: Sorry, I didn't notice this on the first pass, but did we want confidential transfers to be only opt-in post-initialization?  I remember a discussion about not creating Confidential bits by default.

token/program-2022/src/processor.rs
```diff
@@ -102,38 +102,44 @@ impl Processor {
...
             return Err(TokenError::NotRentExempt.into());
         }
 
-        if *mint_info.key != crate::native_mint::id() {
-            let _ = Mint::unpack(&mint_info.data.borrow_mut())
-                .map_err(|_| Into::<ProgramError>::into(TokenError::InvalidMint))?;
+        // get_required_account_extensions checks mint validity
+        let required_extensions = Self::get_required_account_extensions(mint_info)?;
+        if ExtensionType::get_account_len::<Account>(&required_extensions)
+            > new_account_info_data_len
```
joncinque, line 117: What about making this `!=`? That way we're totally covered

joncinque (commented):
Really really close, just two last little questions

token/program-2022/src/processor.rs
CriesofCarrots: Sure. The `<` case should be covered in the extension initialization, but I suppose it is nicer to fail faster.

token/program-2022/src/extension/mod.rs
CriesofCarrots: Yes, I think that's correct, and it should be handled by `ExtensionType::get_account_extensions`. But I don't love having two methods that devs will have to remember to edit when implementing a new extension type. This is why I initially handled all extensions here.

Anyway, I've updated the ConfidentialTransfer case to be a no-op here, updated the name of `get_account_extensions` to be more explicit, and will ponder whether we can improve specifying "required on account initialization" across the two functions.

joncinque (approved):
Thanks for the changes, looks good!

PR #2746, opened 2022-01-16
token-2022: overwrite with init_extension using default state

CriesofCarrots:
As discussed in this thread https://github.com/solana-labs/solana-program-library/pull/2743#discussion_r785221520, this PR explores overwriting any existing data for an ExtensionType when `init_extension()` is called, using `Type::default()`. This enables future extensions to be initialized with non-zero default data.
@joncinque wdyt?

~~This is still marked as draft because of the `default()` implementation. I think ideally Extension becomes `trait Extension: Pod + Default`, but the ConfidentialTransfer state structs will need to have Default impl first. At a glance, it looks to me like it would be okay to add a `Self::zeroed()` default to each of those structs, but need to confirm with @samkim-crypto~~

(edit) Default implemented upstream, so it is now `trait Extension: Pod + Default` 🎉

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -411,16 +411,16 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
         S::pack_into_slice(&self.base, self.base_data);
     }
 
-    /// Packs the extension data into an open slot if not already found in the
-    /// data buffer
+    /// Packs the default extension data into an open slot if not already found in the
+    /// data buffer, otherwise overwrites the existing extension with the default state
     pub fn init_extension<V: Extension>(&mut self) -> Result<&mut V, ProgramError> {
         self.init_or_get_extension(true)
     }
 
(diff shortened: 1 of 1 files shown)


joncinque (commented):
Looks good! Definitely agree that `Extension` should be `Pod + Default` if it makes sense for the confidential types

samkim-crypto:
Yes, it looks good to me. Implementing `Default` for the confidential transfer state structs with `Self::zeroed()` is good. 

Maybe implementing `Default` for the ElGamal and authenticated encryption structs in the zk-token-sdk [PR #22532](https://github.com/solana-labs/solana/pull/22532) and deriving it for the state structs here may be cleaner.

CriesofCarrots:
> Maybe implementing `Default` for the ElGamal and authenticated encryption structs in the zk-token-sdk [PR #22532](https://github.com/solana-labs/solana/pull/22532) and deriving it for the state structs here may be cleaner.

Yes exactly, thank you!

joncinque (approved):
Looks good!

PR #2751, opened 2022-01-17
Governance: Account creation workaround for existing PDA

SebastianBor:
#### Summary 

Any `PDA` account can be created by sending `lamports` to it and then it can't be initialised by program using `create_account` instruction any longer. 

#### Solution

Use `allocate/assign` to initialise PDA when it already exists


ckamm (approved):

joncinque (commented):
Looks good to me too!

PR #2754, opened 2022-01-18
token-2022: Add mint close authority support everywhere

joncinque:
Builds on #2744 , only need to look at 4904516a82545afb7900b5c5b741e5834ab87ef8

#### Problem

The mint close authority must be properly supported in `set_authority` and `close_account`.

#### Solution

Add the support.

Also, update `validate_owner` to accept the data length from the outside, since there were still double-borrow issues.  Now, we pull out the data length right at the beginning, and use that later to see if the authority account is a multisig.  Thankfully, if the authority is a multisig, then it *can't* be used for anything else in processing logic, meaning all double-borrows are avoided.

### DIFF
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -238,6 +238,7 @@ impl Processor {
 
         let dest_account_info = next_account_info(account_info_iter)?;
         let authority_info = next_account_info(account_info_iter)?;
+        let authority_info_len = authority_info.data_len();
 
         let mut source_account = Account::unpack(&source_account_info.data.borrow())?;
         let mut dest_account = Account::unpack(&dest_account_info.data.borrow())?;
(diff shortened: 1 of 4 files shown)


token/program-2022/src/processor.rs
```diff
@@ -412,93 +421,108 @@ impl Processor {
...
                     )?;
-                    mint.freeze_authority = new_authority;
+                    extension.close_authority = new_authority.try_into()?;
                 }
                 _ => {
                     return Err(TokenError::AuthorityTypeNotSupported.into());
                 }
             }
-            Mint::pack(mint, &mut account_info.data.borrow_mut())?;
+            mint.pack_base();
```
mvines, line 523: I guess technically the `pack_base()` is unnecessary for `AuthorityType::CloseAccount` but trying to avoid this is probably more trouble than it's worth

joncinque: that's a good point, I'll move this up for freeze and mint and avoid this call, for sure

mvines (approved):

PR #2757, opened 2022-01-20
token-2022: Add init transfer fee config

joncinque:
#### Problem

We need to initialize fees on a mint.

#### Solution

Add that support.  There's also a couple of field renames:

* `fee_config_authority` -> `transfer_fee_config_authority`
* `withheld_withdraw_authority` -> `withdraw_withheld_authority` (this is for consistency, and because I kept getting it wrong)


mvines (approved):
lgtm!

PR #2760, opened 2022-01-20
DO NOT MERGE associated-token-account: Show how to revert for local dev

joncinque:
#### Problem

If people want to develop against master, CLIs using the ATA program will fail because they're trying to access the normal token program, but the ATA library will be using spl-token-2022.

#### Solution

Do not merge this, but show how to revert back to spl-token for local development for anyone who needs it.


joncinque:
Not needed anymore

PR #2764, opened 2022-01-20
token-2022: Zeroize account data on close

joncinque:
This the token-2022 version of #2763 

#### Problem

Closing accounts doesn't null out all the data.

#### Solution

Zero out all data on account close.

Differences with the non-2022 version:
* Reuse `check_program_account` instead of the other owner check
* `cmp_pubkeys` refactored into lib.rs
* `sol_memset` on close sets the entire `data.len()` to 0, and not `Account:::LEN`

### DIFF
--- a/token/program-2022/src/lib.rs
+++ b/token/program-2022/src/lib.rs
@@ -16,7 +16,12 @@ mod entrypoint;
 
 // Export current sdk types for downstream users building with a different sdk version
 pub use solana_program;
-use solana_program::{entrypoint::ProgramResult, program_error::ProgramError, pubkey::Pubkey};
+use solana_program::{
+    entrypoint::ProgramResult,
+    program_error::ProgramError,
+    program_memory::sol_memcmp,
+    pubkey::{Pubkey, PUBKEY_BYTES},
+};
 
 /// Convert the UI representation of a token amount (using the decimals field defined in its mint)
 /// to the raw amount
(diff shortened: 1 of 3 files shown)


token/program-2022/src/lib.rs
```diff
@@ -38,3 +43,9 @@ pub fn check_program_account(spl_token_program_id: &Pubkey) -> ProgramResult {
     }
     Ok(())
 }
+
+/// Checks two pubkeys for equality in a computationally cheap way using
+/// `sol_memcmp`
+pub fn cmp_pubkeys(a: &Pubkey, b: &Pubkey) -> bool {
```
mvines, line 49: Perhaps we could manually implement `PartialEq` for `Pubkey` to make this optimization available to all Rust programs automatically

token/program-2022/src/processor.rs
```diff
@@ -297,6 +298,11 @@ impl Processor {
             )?,
         };
 
+        if self_transfer || amount == 0 {
```
mvines, line 301: wdyt about unconditionally checking as belts and suspenders?  Slightly more compute for the false case seems to be the only downside

token/program-2022/src/processor.rs
```diff
@@ -574,6 +580,11 @@ impl Processor {
             COption::None => return Err(TokenError::FixedSupply.into()),
         }
 
+        if amount == 0 {
```
mvines, line 583: Same here, unconditionally check?

token/program-2022/src/processor.rs
```diff
@@ -658,6 +669,11 @@ impl Processor {
             )?,
         }
 
+        if amount == 0 {
```
mvines, line 672: ....aaaand here?

token/program-2022/src/lib.rs
joncinque: Definitely. I want to wait until the memcmp costs are sorted out and be sure that it's cheaper before making it the default behavior.

mvines (approved):

token/program-2022/src/processor.rs
joncinque: It depends on how sensitive we want to be about raising compute costs.  I'm a bit worried that the extensions will make things more expensive, and these owner checks are already quite pedantic.

How about this: we remove the checks now, and when we're ready to release, we revisit compute costs and see if the savings are worth it at that point?

mvines: Deal!

PR #2766, opened 2022-01-21
associated-token-account: Bump dependent token version to 3.3

joncinque:
#### Problem

We jumped the gun a little bit with ATA by linking it to spl-token-2022, and we'll need a version of the crate linked to token 3.3.0.

#### Solution

Bump the version and update the program to use `initialize_account3`.  As soon as this is merged, I'll publish the crate and revert the Cargo.toml change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3853,7 +3853,7 @@ dependencies = [
  "solana-client",
  "solana-program-test",
  "solana-sdk",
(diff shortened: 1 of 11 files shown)


CriesofCarrots (approved):
Oh dang, for some reason I thought we had already done that release 😞 Sorry about that!

PR #2767, opened 2022-01-21
Make `spl-token` package compatible with Create React App and Webpack 5

steveluscher:
New constraints on how modules must be imported and laid out in a distribution in CRA/Webpack 5 make it impossible to import the `@solana/spl-token` module.

In this PR, we update the build config to:

* build .mjs files into the dist bundle
* import indexes explicitly, since ESModules no longer support the implicit `import './foo'` syntax.

Note: This depends on https://github.com/solana-labs/buffer-layout-utils/pull/3 being released


jordansexton:
I'll properly review this soon but

> ESModules no longer support the implicit `import './foo'` syntax.

ಠ_ಠ

wtfff

steveluscher:
> For example, if we do an import './directory' from, say, a directory that has an index.js, ES modules do not look for an index.js file in the specified folder, as was the case in CommonJS. Instead, it throws an error.

https://blog.logrocket.com/es-modules-in-node-today/

jordansexton:
Looks great. Thanks! I'll publish a new alpha version once I merge #2687

PR #2770, opened 2022-01-21
token-2022: bump compute limit for close test

CriesofCarrots:
#### Problem

The monorepo downstream job is failing because of compute unit usage in `close_account` test: https://buildkite.com/solana-labs/solana/builds/63871#bb2e1442-b5cb-422f-bcd3-2fe6e67a6819

#### Solution

Bump limit


joncinque (approved):
Bah, sorry about that, I thought 6k would do it since the other tests were using that number.  Thanks for fixing this up

PR #2773, opened 2022-01-21
stake-pool-cli / token-lending-cli: Sign after balance check

joncinque:
#### Problem

To use the new `get_fee_for_message` APIs, the CLIs were wrongly signing the message, and then checking the fee, which means that signatures could remain unused, making @t-nelson 's skin crawl.

#### Solution

No crawling skin! Use `Message::new_with_blockhash`, check the fee, and then sign the transaction with that.

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -31,6 +31,7 @@ use {
     solana_sdk::{
         commitment_config::CommitmentConfig,
         hash::Hash,
+        message::Message,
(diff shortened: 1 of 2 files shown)


joncinque:
> We should really have some Transaction::sign_*() method(s) that assume the blockhash is already set 

Yeah, it would be useful to have something that only takes in the signers and the message.

Although to be honest, I was expecting this PR to look much messier, but `Message::new_with_blockhash` + `Transaction::new` works pretty nicely.

PR #2782, opened 2022-01-22
token-2022: make extension instruction builders consistent with vanilla token

CriesofCarrots:
#### Problem

Started writing some tests using extension ix builders. It's really tedious to have to remember which instructions and which fields take a Pubkey by reference vs not. Also, users shouldn't have to know about COptions at all.

#### Solution

Rewrite ix builders to use `&Pubkey` and `Option` consistently and handle conversions/deref under the hood

### DIFF
--- a/token/rust/src/token.rs
+++ b/token/rust/src/token.rs
@@ -3,7 +3,6 @@ use solana_sdk::{
     account::Account as BaseAccount,
     instruction::Instruction,
     program_error::ProgramError,
-    program_option::COption,
     program_pack::Pack,
     pubkey::Pubkey,
     signer::{signers::Signers, Signer},
@@ -52,11 +51,11 @@ impl PartialEq for TokenError {
 #[derive(Clone, Debug, PartialEq)]
 pub enum ExtensionInitializationParams {
     MintCloseAuthority {
-        close_authority: COption<Pubkey>,
+        close_authority: Option<Pubkey>,
     },
     TransferFeeConfig {
-        transfer_fee_config_authority: COption<Pubkey>,
-        withdraw_withheld_authority: COption<Pubkey>,
+        transfer_fee_config_authority: Option<Pubkey>,
+        withdraw_withheld_authority: Option<Pubkey>,
         transfer_fee_basis_points: u16,
         maximum_fee: u64,
     },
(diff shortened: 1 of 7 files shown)


token/rust/src/token.rs
```diff
@@ -52,11 +51,11 @@ impl PartialEq for TokenError {
...
 pub enum ExtensionInitializationParams {
     MintCloseAuthority {
-        close_authority: COption<Pubkey>,
+        close_authority: Option<Pubkey>,
     },
     TransferFeeConfig {
-        transfer_fee_config_authority: COption<Pubkey>,
-        withdraw_withheld_authority: COption<Pubkey>,
+        transfer_fee_config_authority: Option<Pubkey>,
+        withdraw_withheld_authority: Option<Pubkey>,
```
joncinque, line 58: This makes way more sense, thanks!

joncinque (commented):
I'm wondering if @mvines should take a look too, since I imagine he made the constructors deliberately for confidential tokens. I prefer to pass in values over refs everywhere since `Pubkey` is `Copy`, but consistency is more important.  Do we want to consider making a shift to values over refs with token-2022?

Also, the extension instructions don't pass in the token program id. Is that worth adding in? My sense is no, but I can be swayed.

CriesofCarrots:
> Also, the extension instructions don't pass in the token program id. Is that worth adding in? My sense is no, but I can be swayed.

Ah, thanks for bringing this up. I was wondering about that as well. If we might ever support multiple token program-ids, it seems like we might save ourselves some backward compatibility headache by plumbing it in now. I think I'm for it.

mvines:
I'd love to just drop `Copy` from `Pubkey` myself 😈

joncinque:
Well, consider me totally swayed on all counts then :sweat_smile: apologies for doubting

joncinque (approved):

mvines:
> If we might ever support multiple token program-ids, it seems like we might save ourselves some backward compatibility headache by plumbing it in now. I think I'm for it.

Would us having an upgradable "staging"/"next" token program impact anything if we don't plumb the program_id through?  That's the most immediate case I can think of when we may have two program ids running around

CriesofCarrots:
> Would us having an upgradable "staging"/"next" token program impact anything if we don't plumb the program_id through?

Ah yeah, making a staging program available to users is another good reason to plumb program_id. Will PR separately.

PR #2738, opened 2022-01-14
Handle extended token Accounts in ATA program

CriesofCarrots:
- [x] Needs rebase on #2736 and #2737
- [x] Needs rebase on #2739
- [x] Need support for get_/set_return_data in ProgramTest: https://github.com/solana-labs/solana/pull/22647, #2780
- [x] Needs rebase on  #2800

PR #2783, opened 2022-01-23
token-2022: add more tests

CriesofCarrots:
Now that InitializeMint with extensions is supported, can flesh out TODOs:
- test_get_account_data_size unit test with extended mint
- ProgramTest tests for InitializeAccounts

More tests via rust client to come after ATA support is complete in #2738

Needs rebase on #2782

### DIFF
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -1113,10 +1113,15 @@ impl PrintProgramError for TokenError {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::instruction::*;
+    use crate::{
+        extension::transfer_fee::instruction::initialize_transfer_fee_config, instruction::*,
+    };
     use solana_program::{
-        account_info::IntoAccountInfo, clock::Epoch, instruction::Instruction, program_error,
-        sysvar::rent,
+        account_info::IntoAccountInfo,
+        clock::Epoch,
+        instruction::Instruction,
+        program_error,
+        sysvar::{clock::Clock, rent},
     };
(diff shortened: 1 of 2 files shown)


joncinque (dismissed):
Tests look good!

token/program-2022/src/processor.rs
```diff
@@ -149,6 +149,7 @@ impl Processor {
         };
 
         account.pack_base();
+        account.init_account_type()?;
```
CriesofCarrots, line 152: @joncinque just wanted to point out that I missed this in #2743
Maybe it doesn't matter because the scope of use is so small, but would it make sense to have successful extension inits do this under the hood?

joncinque: Yeah I noticed, did the end-to-end test catch this?

> Maybe it doesn't matter because the scope of use is so small, but would it make sense to have successful extension inits do this under the hood?

With mints, we initialize all extensions *and then* initialize the base + account type, so I'd prefer to keep that flow for accounts too, even though it's all done within the same instruction.  But I don't feel super strongly about this.

CriesofCarrots: > did the end-to-end test catch this or the processor-level test?

The end-to-end, since it depended on full-fledged get-/set-return-data functionality to detect.

Okay. Could maybe be bundled with base init, then. But I'll drop it for now, and readdress if it crops up again later.

joncinque: > Could maybe be bundled with base init, then.

Good call, that'll probably make the most sense

PR #2788, opened 2022-01-24
Temp. ignore thread_local 1.1.3 security warning

SebastianBor:
#### Summary
 
Data race in `Iter` and `IterMut` in thread_local 1.1.3 upstream dependencies 

Date: 2022-01-23
https://rustsec.org/advisories/RUSTSEC-2022-0006
Solution: Upgrade to >=1.1.4
Ingored untill fixed in solana sdk


CriesofCarrots:
We should be able to bump the package without waiting for an sdk release: `cargo update -p thread_local`

PR #2796, opened 2022-01-24
Revert ATA to spl-token v3.3

CriesofCarrots:


joncinque (approved):
Thanks for taking care of this

PR #2797, opened 2022-01-24
token-2022: Implement set transfer fee instruction

joncinque:
#### Problem

The transfer fee config authority needs to be able to set the transfer fee.

#### Solution

Add the implementation and tests for it

### DIFF
--- a/token/program-2022/src/extension/transfer_fee/processor.rs
+++ b/token/program-2022/src/extension/transfer_fee/processor.rs
@@ -76,10 +76,10 @@ fn process_set_transfer_fee(
         account_info_iter.as_slice(),
     )?;
 
-    // When setting the transfer fee, we have a few situations:
+    // When setting the transfer fee, we have two situations:
     // * newer transfer fee epoch <= current epoch:
     //     newer transfer fee is the active one, so overwrite older transfer fee with newer, then overwrite newer transfer fee
-    // * newer transfer fee epoch >= next epoch:
+    // * newer transfer fee epoch == next epoch:
     //     it was never used, so just overwrite next transfer fee
     let epoch = Clock::get()?.epoch;
     let next_epoch = epoch.saturating_add(1);
--- a/token/program-2022/src/extension/transfer_fee/processor.rs
+++ b/token/program-2022/src/extension/transfer_fee/processor.rs
@@ -1,10 +1,12 @@
 use {
     crate::{
         check_program_account,
+        error::TokenError,
         extension::{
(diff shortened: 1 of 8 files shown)


token/program-2022/src/extension/transfer_fee/processor.rs
```diff
@@ -48,6 +50,52 @@ fn process_initialize_transfer_fee_config(
...
+            .ok_or(TokenError::NoAuthorityExists)?;
+    Processor::validate_owner(
+        program_id,
+        &transfer_fee_config_authority,
+        authority_info,
+        authority_info_data_len,
+        account_info_iter.as_slice(),
+    )?;
+
+    // When setting the transfer fee, we have a few situations:
```
CriesofCarrots, line 79:
```suggestion
    // When setting the transfer fee, we have two situations:
```

token/program-2022/src/extension/transfer_fee/processor.rs
```diff
@@ -48,6 +50,52 @@ fn process_initialize_transfer_fee_config(
...
+    // * newer transfer fee epoch <= current epoch:
+    //     newer transfer fee is the active one, so overwrite older transfer fee with newer, then overwrite newer transfer fee
+    // * newer transfer fee epoch >= next epoch:
```
CriesofCarrots, line 82: This is slightly confusing, because newer transfer fee epoch can't ever be greater than next epoch, can it?

CriesofCarrots (approved):
Nice! Couple nits only

token/program-2022/src/processor.rs
```diff
@@ -1107,6 +1107,9 @@ impl PrintProgramError for TokenError {
             TokenError::MintHasSupply => {
                 msg!("Error: Mint has non-zero supply. Burn all tokens before closing the mint")
             }
+            TokenError::NoAuthorityExists => {
+                msg!("No authority exists to perform the desired operation");
```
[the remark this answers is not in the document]
joncinque: Oops, totally missed that, thanks!

token/program-2022/src/extension/transfer_fee/processor.rs
joncinque: Hehe when I started writing it I thought there would be more

token/program-2022/src/extension/transfer_fee/processor.rs
joncinque: I hesitated when writing this, thinking "well, who knows what can happen", but yes you're right, I'll change it to `==`

PR #2799, opened 2022-01-24
Governance: Transfer freeze and close authorities 

SebastianBor:
#### Summary

When mint or token account is transferred to governance then all authorities should be transferred at that time

#### Implementation 

Mint `freeze_authority`  and token account `close_authority` are transferred in addition to `mint` and `owner` authorities when governance is created

### DIFF
--- a/governance/program/src/instruction.rs
+++ b/governance/program/src/instruction.rs
@@ -352,10 +352,11 @@ pub enum GovernanceInstruction {
         config: GovernanceConfig,
 
         #[allow(dead_code)]
-        /// Indicates whether Mint's authority should be transferred to the Governance PDA
+        /// Indicates whether Mint's authorities should be transferred to the Governance PDA
         /// If it's set to false then it can be done at a later time
         /// However the instruction would validate the current mint authority signed the transaction nonetheless
-        transfer_mint_authority: bool,
+        /// If the mint has also freeze_authority then it'll be transferred as well
+        transfer_mint_authorities: bool,
     },
 
     /// Creates Token Governance account which governs a token account
@@ -756,7 +757,7 @@ pub fn create_mint_governance(
     voter_weight_record: Option<Pubkey>,
     // Args
     config: GovernanceConfig,
-    transfer_mint_authority: bool,
+    transfer_mint_authorities: bool,
 ) -> Instruction {
     let mint_governance_address = get_mint_governance_address(program_id, realm, governed_mint);
 
@@ -777,7 +778,7 @@ pub fn create_mint_governance(
 
     let instruction = GovernanceInstruction::CreateMintGovernance {
         config,
-        transfer_mint_authority,
+        transfer_mint_authorities,
     };
 
     Instruction {
--- a/governance/program/src/instruction.rs
+++ b/governance/program/src/instruction.rs
@@ -337,7 +337,7 @@ pub enum GovernanceInstruction {
     ///   0. `[]` Realm account the created Governance belongs to    
     ///   1. `[writable]` Mint Governance account. PDA seeds: ['mint-governance', realm, governed_mint]
     ///   2. `[writable]` Mint governed by this Governance account
-    ///   3. `[signer]` Current Mint Authority
+    ///   3. `[signer]` Current Mint authority (MintTokens and optionally FreezeAccount)
     ///   4. `[]` Governing TokenOwnerRecord account    
     ///   5. `[signer]` Payer
     ///   6. `[]` SPL Token program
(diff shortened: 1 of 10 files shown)


governance/program/src/instruction.rs
```diff
@@ -352,18 +352,18 @@ pub enum GovernanceInstruction {
         config: GovernanceConfig,
 
         #[allow(dead_code)]
-        /// Indicates whether Mint's authority should be transferred to the Governance PDA
+        /// Indicates whether Mint's authorities (MintTokens, FreezeAccount) should be transferred to the Governance PDA
         /// If it's set to false then it can be done at a later time
         /// However the instruction would validate the current mint authority signed the transaction nonetheless
-        transfer_mint_authority: bool,
+        transfer_mint_authorities: bool,
```
joncinque, line 358: The preference would be for separate signing accounts to do this properly, but I imagine breaking the interface is a no-go.  At the very least, a separate flag to also transfer freeze authority would be best, so you have `transfer_mint_authority` and `transfer_freeze_authority`

governance/program/src/instruction.rs
```diff
@@ -378,10 +378,10 @@ pub enum GovernanceInstruction {
         config: GovernanceConfig,
 
         #[allow(dead_code)]
-        /// Indicates whether token owner should be transferred to the Governance PDA
+        /// Indicates whether the token account authorities (AccountOwner and optionally CloseAccount) should be transferred to the Governance PDA
```
joncinque, line 381: The `CloseAccount` authority is rarely the owner, so I'm not sure of the use-case for this

governance/program/src/processor/process_create_mint_governance.rs
```diff
@@ -84,13 +88,28 @@ pub fn process_create_mint_governance(
...
             mint_governance_info.key,
+            AuthorityType::MintTokens,
             spl_token_info,
         )?;
+
+        // If the mint has freeze_authority then transfer it as well
+        let mint_data = Mint::unpack(&governed_mint_info.data.borrow())?;
+        // Note: The code assumes mint_authority==freeze_authority
+        //       If this is not the case then the caller should set freeze_authority accordingly before making the transfer
+        if mint_data.freeze_authority.is_some() {
```
joncinque, line 104: With a separate flag, you can check that the freeze authority is the given `governed_mint_authority_info`, and give a clear error if it isn't

governance/program/src/processor/process_create_token_governance.rs
```diff
@@ -84,13 +86,28 @@ pub fn process_create_token_governance(
...
+        //       If this is not the case then the caller should set close_authority accordingly before making the transfer
+        if token_account_data.close_authority.is_some() {
+            set_spl_token_account_authority(
+                governed_token_info,
+                governed_token_owner_info,
+                token_governance_info.key,
+                AuthorityType::CloseAccount,
+                spl_token_info,
+            )?;
+        }
```
joncinque, line 110: As mentioned earlier, I can't see a use-case for this, so you can probably remove it

joncinque (commented):
A few comments, but the rest looks pretty much there

governance/program/src/instruction.rs
SebastianBor: The reason for this change was to ensure we won't have a partially owned mint or token account in a DAO when this instruction is used. This is why it's all or nothing. You either have to transfer all authorities or you can't do it all. 

Note, this is not a security gate  and it only prevents honest mistakes, because you can still take a mint or a token account and set its authority to any governance without using this instruction, or use this instruction without transferring the authorities. 

Strategically I want to remove the Create Mint/Token/Program Governance instructions and replace it with metadata based annotation and certification where DAO wold be able to approve assets

governance/program/src/processor/process_create_token_governance.rs
SebastianBor: This is to ensure all the authorities are transferred. If the owner can't also have `CloseAuthority` if it's set then the transfer shouldn't be possible.

governance/program/src/processor/process_create_mint_governance.rs
SebastianBor: Agree that would be better, but since the instruction is going to be deprecated anyway I just did the minimum to ensure the partial transfer of authorities is not possible.

governance/program/src/instruction.rs
joncinque: In that case, I would just check that it's `None` and that's it. If it's `None` then the owner is the CloseAuthority by default

governance/program/src/processor/process_create_mint_governance.rs
joncinque: I don't think this is the correct choice, but if it's just a stop-gap, then no problem

PR #2801, opened 2022-01-25
token-2022: implement SetAuthority for TransferFeeConfig and WithheldWithdraw

CriesofCarrots:
#### Problem

The transfer fee config authorities cannot be configured after initialization

#### Solution

Implement SetAuthority for these types

To come later: additional test cases to check that updated withdraw_withheld_authority can withdraw and old cannot; will be much easier to code when the rust token client supports extended Accounts (via ATA) and TransferCheckedWithFee

### DIFF
--- a/token/program-2022/tests/transfer_fee.rs
+++ b/token/program-2022/tests/transfer_fee.rs
@@ -11,6 +11,7 @@ use {
     spl_token_2022::{
         error::TokenError,
         extension::transfer_fee::{TransferFee, TransferFeeConfig},
+        instruction,
     },
     spl_token_client::token::{ExtensionInitializationParams, TokenError as TokenClientError},
     std::convert::TryInto,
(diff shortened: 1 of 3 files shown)


token/program-2022/tests/transfer_fee.rs
```diff
@@ -264,3 +265,234 @@ async fn fail_set_fee_unsupported_mint() {
...
+    let transfer_fee_basis_points = u16::MAX;
+    let maximum_fee = u64::MAX;
+    token
+        .set_transfer_fee(
+            &transfer_fee_config_authority,
+            transfer_fee_basis_points,
+            maximum_fee,
+        )
+        .await
+        .unwrap_err();
```
joncinque, line 339: nit: Could we check the exact error type here?

token/program-2022/tests/transfer_fee.rs
```diff
@@ -264,3 +265,234 @@ async fn fail_set_fee_unsupported_mint() {
...
+                InstructionError::Custom(TokenError::AuthorityTypeNotSupported as u32)
+            )
+        )))
+    );
+
+    // fail update transfer fee config
+    token
+        .set_transfer_fee(&transfer_fee_config_authority, 0, 0)
+        .await
+        .unwrap_err();
```
joncinque, line 386: nit: same here

token/program-2022/tests/transfer_fee.rs
```diff
@@ -264,3 +265,234 @@ async fn fail_set_fee_unsupported_mint() {
...
+        .await
+        .unwrap();
+    let state = token.get_mint_info().await.unwrap();
+    let extension = state.get_extension::<TransferFeeConfig>().unwrap();
+    assert_eq!(
+        extension.withdraw_withheld_authority,
+        Some(new_authority.pubkey()).try_into().unwrap(),
+    );
+
+    // TODO: assert new authority can withdraw withheld fees and old cannot
```
joncinque, line 458: I'll work on transfer today so we can get to the rest of these :smile:

joncinque (dismissed):
Just a couple of nits, looks good!

token/program-2022/tests/transfer_fee.rs
CriesofCarrots: We could. I didn't think it was worth the effort, but I'll do it for you 😅

joncinque: It's just a nit, so feel free to leave it if it's annoying

PR #2803, opened 2022-01-25
Governance: Voter weight action & target

SebastianBor:
#### Summary

Some `voter-weight-addins` need additional information to determine voters' weights, which can be specific to the action a voter takes and what the action pertains to. Some common examples:

- `addin` inspects target `Proposal` instructions to determine whether given voter can cast vote
- `addin` inspects target `Proposal` when voter casts vote to use weight based on the remming voting time to encorage early participation and diminish voting power towards the end of the voting time
- `addin` can use role based approach to determine who can create governances, proposals and cast votes 
- `addin` inspects target `Proposal` when voter casts vote to enforce token lock for the duration of the proposal. This way it doesn't have to depend on `spl-governance` to track unrelinquished votes 

#### Solution 

Add addition fields to `VoterWeightRecord` to allow `addins` to specify the `action` and `target` the provided `voter_weight` pertains to. 
When `action` and `target` are specified then `spl-governance` validates they match the executed instruction.


governance/program/tests/program_test/mod.rs
```diff
@@ -98,6 +102,8 @@ impl GovernanceProgramTest {
 
     #[allow(dead_code)]
     pub async fn start_with_voter_weight_addin() -> Self {
+        // We only ensure the add-in is built but it doesn't detect changes
+        // If the addin is changed it needs to be manually built
```
joncinque, line 106: If this is getting annoying, you want to use a `rerun-if-changed` directive from a build script: https://doc.rust-lang.org/cargo/reference/build-scripts.html

governance/voter-weight-addin/program/src/instruction.rs
```diff
@@ -2,53 +2,48 @@
...
 pub enum VoterWeightAddinInstruction {
-    /// Revises voter weight providing up to date voter weight
+    /// Sets up VoterWeightRecord owned by the program
     ///
     /// 0. `[]` Governance Program Id
     /// 1. `[]` Realm account
     /// 2. `[]` Governing Token mint
     /// 3. `[]` TokenOwnerRecord
     /// 4. `[writable]` VoterWeightRecord
-    Revise {},
```
joncinque, line 21: I imagine that it's ok to break this interface?

governance/voter-weight-addin/program/src/processor.rs
```diff
@@ -29,19 +30,30 @@ pub fn process_instruction(
...
+            voter_weight,
+            voter_weight_expiry,
+            weight_action,
+            weight_action_target,
+        ),
     }
 }
 
 /// Processes Deposit instruction
 pub fn process_deposit(
```
joncinque, line 50: I got a bit confused here -- how about `process_setup_voter_weight_record` or `process_resolve_voter_weight`?

governance/voter-weight-addin/program/src/instruction.rs
```diff
@@ -2,53 +2,48 @@
...
-    /// 0. `[]` Governance Program Id
-    /// 1. `[]` Realm account
-    /// 2. `[]` Governing Token mint
-    /// 3. `[]` TokenOwnerRecord
-    /// 4. `[writable]` VoterWeightRecord
     /// 5. `[signer]` Payer
     /// 6. `[]` System
-    Deposit {
-        /// The deposit amount
+    SetupVoterWeightRecord {
```
joncinque, line 25: I'm a much bigger fan of this! Very clean.  My apologies for not remembering all of this, but previously when you did `deposit` and `withdraw`, you weren't actually moving tokens, correct?  Which is what makes this interface possible?

governance/program/src/addins/voter_weight.rs
```diff
@@ -49,6 +68,17 @@ pub struct VoterWeightRecord {
...
+    /// The governance action the voter's weight pertains to
+    /// It allows to provided voter's weight specific to the particular action the weight is evaluated for
+    /// When the action is provided then the governance program asserts the executing action is the same as specified by the addin
+    pub weight_action: Option<VoterWeightAction>,
+
+    /// The target the voter's weight  action pertains to
+    /// It allows to provided voter's weight specific to the target the weight is evaluated for
+    /// For example when addin supplies weight to vote on a particular proposal then it must specify the proposal as the action target
+    /// When the target is provided then the governance program asserts the target is the same as specified by the addin
+    pub weight_action_target: Option<Pubkey>,
```
joncinque, line 81: So are these fields just a safeguard to be sure that the addin program behaves correctly and is forced to acknowledge the parameters passed in?  I went back and forth on this, but finally it seems like a good idea.  If they're set in correctly, it means the addin program actually deserialized what you passed in, which is better than nothing.

governance/program/src/addins/voter_weight.rs
```diff
@@ -23,7 +23,24 @@ pub enum VoterWeightAccountType {
     VoterWeightRecord,
 }
 
-/// VoterWeight Record account
+/// The governance action VoterWeight is evaluated for
+#[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub enum VoterWeightAction {
```
joncinque, line 28: nit: since all of these actions concern a target, you can make a helper struct that has this enum + pubkey, or you can have the value contained in the enum, ie:
```
enum VoterWeightAction {
    CastVote(Pubkey),
...
}
```

governance/program/tests/program_test/addins.rs
```diff
@@ -0,0 +1,84 @@
+// Governance program has no dependency on the voter-weight-addin program and hence we can't use its instructions here
+// The instructions have to be manually recreated
```
joncinque, line 2: You could save yourself some rewriting by defining the instruction interface and creator in the governance program, and having the sample addin program just process that instruction.

joncinque (approved):
Nothing to stop this from going forward, mainly some questions and nits

governance/program/tests/program_test/mod.rs
SebastianBor: We've tried this before but it didn't work because we have the chicken and egg problem here.
The `addin` depends on `spl-governance` and has to be built after it and not before.
We only need the `voter-weight-addin` as a programable mock program and it doesn't change much so I think it's acceptable for now.

governance/voter-weight-addin/program/src/instruction.rs
SebastianBor: yes, the `addin` is just a programable mock program used for testing only

governance/voter-weight-addin/program/src/processor.rs
SebastianBor: correct, I missed that, renamed now

governance/voter-weight-addin/program/src/instruction.rs
SebastianBor: My initial idea was to make the `addin` a fully working example but I've never implemented that and backtracked from that concept in this PR. Now the `addin` is just a programable mock program which I think is better. It doesn't confuse any longer  and also allows us to test spl-governance with explicit responses from the `addin`

governance/program/src/addins/voter_weight.rs
SebastianBor: It's a reversed relationship. This account is owned by the `addin` and it's the `addin` responsibility to set these fields and then the account is passed to `spl-governance` which validates they match the invoked instruction.

This way the `addin` can for example fetch and inspect `Proposal` and specify in `VoterWeightRecord`  the `voter_weight` was provided for voting on that particular proposal. Then `spl-governance`  checks that in fact `CastVote` was invoked for the `Proposal`

If the `addin` doesn't specify the acton and the target then it means the weight is not context sensitive and can be used for any action and any target

governance/program/tests/program_test/addins.rs
SebastianBor: yes, that would be possible, but since the `addin` is not a real one I didn't want to pollute `governance` program with this mock interface.
I also renamed the `addin` program to `addin-mock` to make the intention explicit

governance/program/src/addins/voter_weight.rs
SebastianBor: The `addin` can specify only the target without any action, or just the action without any target.
For example:
-  It can define voter's weight for anything pertaining to a `Realm` without explicit action.
- It can specify voter's weight for `CastVote` for any proposal

PR #2804, opened 2022-01-25
Add stake-pool browser build commands and documentation

CMEONE:
**Problem**
There are currently no scripts or documentation to build the TypeScript bindings for the stake-pool program for the browser.

**Summary of Changes**
This PR adds scripts and configuration files to build the bindings for the browser, as well as documentation for using the TypeScript bindings in general.

I haven't fully tested the browser files past running the script in the browser and confirming that `solanaStakePool` exists as a global variable in the browser with the relevant functions and constants, but I'll be integrating the browser bindings into the [BlazeStake](https://stake.solblaze.org/) pool and will make another PR if more changes are necessary.

Additionally, I added the UNPKG browser bundles that I assume will be built as soon as a new version is published to NPM, but I can make further changes to this part if UNPKG isn't building the bindings properly.

Fixes #2688 using the solution that @joncinque proposed

PR #2681, opened 2021-12-30
stake-pool-js: Prepare package for release

joncinque:
#### Problem

People want to use the new stake pool JS bindings, but we haven't released it yet.

#### Solution

Prep the package for a first release.  There's an annoyance with buffer-layout between the spl-token version, the borsh version, and the local version, but let's hope that can get resolved later.

Also, finally add the js bindings to CI.

Fixes #2675

PR #2806, opened 2022-01-25
stake-pool: add bindings to js

AlexanderRay:
- add new bindings
  - IncreaseValidatorStake
  - DecreaseValidatorStake
  - UpdateValidatorListBalance
  - UpdateStakePoolBalance
  - CleanupRemovedValidatorEntries
  - DepositStake

- add new function
  - stakePoolInfo

@joncinque


stake-pool/js/package.json
```diff
@@ -1,40 +1,22 @@
 {
```
CMEONE, line 1: Why all the changes to `package.json` and `package-lock.json`? From first glance, it doesn't seem like most of these changes are necessary (other than maybe the `lint` script and possibly `main`). The conflicting changes to these files might make it a little more difficult to merge with the #2804 PR (which was coincidentally made a few hours before this one), and a lot of the removals don't really make sense for me.

CMEONE (changes requested):

joncinque:
As @CMEONE pointed out, it looks like this blasted the changes made during #2681. Could you rebase your changes onto that commit at least?

AlexanderRay:
> As @CMEONE pointed out, it looks like this blasted the changes made during #2681. Could you rebase your changes onto that commit at least?

yea, sorry, my fault, will fix it...

joncinque:
Looks like there are still a few regressions introduced -- could you run this with npm version 7 so that package.lock stays on version 2, and also run the linting / formatting / building?

AlexanderRay:
> Looks like there are still a few regressions introduced -- could you run this with npm version 7 so that package.lock stays on version 2, and also run the linting / formatting / building?

is seems like prettier for some reason has different formation options in your tests ... checking ...

CMEONE:
@AlexanderRay I just wanted to clarify a small thing: in the code, there are references to `stakePoolPubKey`, `stakePoolProgramAddress`, and `stakePoolAddress` (as function parameters). Is there any difference between the three, or is it possible to make it so that the naming is consistent to avoid any confusion?

AlexanderRay:
> @AlexanderRay I just wanted to clarify a small thing: in the code, there are references to `stakePoolPubKey`, `stakePoolProgramAddress`, and `stakePoolAddress` (as function parameters). Is there any difference between the three, or is it possible to make it so that the naming is consistent to avoid any confusion?

fixed

stake-pool/js/package.json
```diff
@@ -15,28 +15,15 @@
   "keywords": [],
   "authors": [
     "Solana Maintainers <maintainers@solana.foundation>",
-    "Lieu Zheng Hong",
+    "Lieu Zheng Hong", 
     "mFactory Team (https://mfactory.ch)"
   ],
-  "homepage": "https://solana.com",
```
CMEONE, line 21: Should these lines be removed from `package.json`? I think most of these removed fields are good to have in the `package.json` file and probably should be added back (unless there's a reason to remove them that I don't know about).

CMEONE (changes requested):

stake-pool/js/package.json
AlexanderRay: yea, sorry about it .. merge issues ... thank you for the note

AlexanderRay: removed react dep from package json"eslint-plugin-react": "^7.28.0",

CMEONE:
Again, great work on these bindings! I am receiving an error with the `depositSol` command from your previous PR:
```
Program log: Error: First update old validator stake account balances and then pool stake balance
```
I assume that `updateStakePool` (added in this PR) would need to be called first to update the stake pool before new stake can be deposited. I also see that `updateRequired` in `stakePoolInfo` would tell me if the stake pool needs to be updated.

Is there any way to create an `updateStakePoolAndDepositSol` function or something that detects automatically if the stake pool needs to be updated (by just checking for the `updateRequired` part and not running all the other computations in `stakePoolInfo`) and adds those update instructions along with deposit instructions, otherwise just returns the deposit instructions if an update is not needed? I don't think that this can be done atomically, but it would be nice to have a function that can achieve this functionality in the most efficient way possible.

CMEONE:
Correction to my previous comment: it may be better to remove the detection step entirely and just have a helper function call the two instruction functions and concatenate the instructions. However, this is probably best to have in a separate, third-party library and not in the main bindings.

AlexanderRay:
It is possible to implement a function like updateAndDepositSol, or add an additional parameter to depositSol function, but I think that is, in general, a bad idea to do an update on the client side. If the stake pool user calls the update, then it pays also for the update, I think it's wrong to update the pool for its costs. Updates should be started from some backend process and paid by the stake pool manager.

CMEONE:
>  If the stake pool user calls the update, then it pays also for the update, I think it's wrong to update the pool for its costs

​I checked by running some test instructions, and when I combine update and deposit instructions in the same transaction, the transaction fee seems to be 0.00001 SOL regardless of whether update instructions are included or not, so the update instructions run at no additional cost to the user.

In my current implementation, I have it so that the update instructions are run in the same transaction as deposit, even if update is not needed (when deposit button is clicked, `updateStakePool` and `depositSol` are called, and the returned instructions are combined like `[...updateInstructions.updateListInstructions, ...updateInstructions.finalInstructions, ...depositInstructions.instructions]`).

I don't think this feature of combining update and deposit instructions is complex enough to warrant it being part of the bindings, developers should be able to implement this feature themselves in just a few lines of code depending on how they want it to be set up (user updates vs. server updates).

CMEONE:
@AlexanderRay are there any other changes that need to be made to this PR, or can @joncinque merge it? The currently deployed bindings have a bug causing many functions to return an error, and that bug is resolved in this PR, so it might be good to get this merged soon if the PR is ready.

stake-pool/js/src/index.ts
```diff
@@ -505,3 +578,331 @@ export async function withdrawSol(
...
+
+/**
+ * Creates instructions required to completely update a stake pool after epoch change.
+ */
+export async function updateStakePool(
+  connection: Connection,
+  stakePool: StakePoolAccount,
+  noMerge = false,
+) {
+  // const stakePool = await getStakePoolAccount(connection, stakePoolAddress)
```
joncinque, line 710: :eyes:

stake-pool/js/src/index.ts
```diff
@@ -505,3 +578,331 @@ export async function withdrawSol(
...
+  };
+}
+
+/**
+ * Retrieves detailed information about the StakePool.
+ */
+export async function stakePoolInfo(connection: Connection, stakePoolAddress: PublicKey) {
+  const stakePool = await getStakePoolAccount(connection, stakePoolAddress);
+  const reserveAccountStakeAddress = stakePool.account.data.reserveStake;
+  const totalLamports = stakePool.account.lamports;
```
joncinque, line 796: The total lamports should be the total lamports of the pool, and not the stake pool account, so something like `stakePool.account.data.totalLamports`

stake-pool/js/src/index.ts
```diff
@@ -505,3 +578,331 @@ export async function withdrawSol(
...
+    epochFee: stakePool.account.data.epochFee,
+    nextEpochFee: stakePool.account.data.nextEpochFee,
+    preferredDepositValidatorVoteAddress:
+      stakePool.account.data.preferredDepositValidatorVoteAddress,
+    preferredWithdrawValidatorVoteAddress:
+      stakePool.account.data.preferredWithdrawValidatorVoteAddress,
+    stakeDepositFee: stakePool.account.data.stakeDepositFee,
+    stakeWithdrawalFee: stakePool.account.data.stakeWithdrawalFee,
+    // CliStakePool the same
+    nextStakeWithdrawalFee: stakePool.account.data.nextSolWithdrawalFee,
```
joncinque, line 886: These mistakes keep coming up in the CLI :sweat: 
```suggestion
    nextStakeWithdrawalFee: stakePool.account.data.nextStakeWithdrawalFee,
```

stake-pool/js/src/instructions.ts
```diff
@@ -133,6 +239,177 @@ export type DepositSolParams = {
...
+
+    const type = STAKE_POOL_INSTRUCTION_LAYOUTS.UpdateStakePoolBalance;
+    const data = encodeData(type);
+
+    const keys = [
+      {pubkey: stakePool, isSigner: false, isWritable: true},
+      {pubkey: withdrawAuthority, isSigner: false, isWritable: false},
+      {pubkey: validatorList, isSigner: false, isWritable: true},
+      {pubkey: reserveStake, isSigner: false, isWritable: false},
+      {pubkey: managerFeeAccount, isSigner: false, isWritable: false},
```
joncinque, line 298:
```suggestion
      {pubkey: managerFeeAccount, isSigner: false, isWritable: true},
```

stake-pool/js/src/utils/stake.ts
```diff
@@ -75,21 +86,21 @@ export async function prepareWithdrawAccounts(
       STAKE_POOL_PROGRAM_ID,
       validator.voteAccountAddress,
       stakePoolAddress,
-      validator.transientSeedSuffixStart,
+      validator?.transientSeedSuffixStart,
```
joncinque, line 89: nit, can remove the `?`

stake-pool/js/src/utils/stake.ts
```diff
@@ -75,21 +86,21 @@ export async function prepareWithdrawAccounts(
...
     );
 
     if (!validator.transientStakeLamports?.isZero()) {
       accounts.push({
         type: 'transient',
         voteAddress: validator.voteAccountAddress,
         stakeAddress: transientStakeAccountAddress,
-        lamports: validator.transientStakeLamports.toNumber(),
+        lamports: validator?.transientStakeLamports.toNumber(),
```
joncinque, line 97: nit: can also remove the `?` here

stake-pool/js/src/utils/token.ts
```diff
@@ -10,6 +11,19 @@ import {AccountLayout} from '../layouts';
...
+  connection: Connection,
+  tokenMintPubkey: PublicKey,
+): Promise<MintInfo | undefined> {
+  // try {
+  // @ts-ignore
+  const token = new Token(connection, tokenMintPubkey, TOKEN_PROGRAM_ID, null);
+  return token.getMintInfo();
+  // } catch (error) {
+  //   console.log(error);
+  // }
```
joncinque, line 24: nit: can get rid of comments around here

joncinque (commented):
The meat of this looks really good! A few bits to clean up and fix -- let me know if you have any questions

CMEONE:
@AlexanderRay small issue: after doing some testing on devnet, I discovered that `withdrawStake` doesn't always properly honor the minimum balance for stake accounts:

This is one example of trying to withdraw lamports in a way that technically looks like it would be allowed given the total stake lamports in the validator account but doesn't honor the minimum amount of lamports that must be stored in a stake pool program validator account:
```
Transaction simulation failed: Error processing Instruction 2: custom program error: 0x17 
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [1]
    Program log: Instruction: Approve
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 2024 of 200000 compute units
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
    Program 11111111111111111111111111111111 invoke [1]
    Program 11111111111111111111111111111111 success
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy invoke [1]
    Program log: Instruction: WithdrawStake
    Program log: Attempting to withdraw 99900000 lamports from validator account with 100000000 stake lamports, 1000000 must remain
    Program log: Error: The lamports in the validator stake account is not equal to the minimum
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy consumed 16664 of 200000 compute units
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy failed: custom program error: 0x17
```

Here's another example, this time the amount requested for withdrawal is above the amount available in the validator account, and although there is another validator account that has enough lamports to make up the difference, the bindings still attempt to withdraw more from the first validator account than possible:

```
Transaction simulation failed: Error processing Instruction 2: custom program error: 0x17 
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [1]
    Program log: Instruction: Approve
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 2024 of 200000 compute units
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
    Program 11111111111111111111111111111111 invoke [1]
    Program 11111111111111111111111111111111 success
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy invoke [1]
    Program log: Instruction: WithdrawStake
    Program log: Attempting to withdraw 102282880 lamports from validator account with 100000000 stake lamports, 1000000 must remain
    Program log: Error: The lamports in the validator stake account is not equal to the minimum
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy consumed 16683 of 200000 compute units
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy failed: custom program error: 0x17
```

The bindings have previously dipped into multiple validator accounts and successfully withdrawn stakes from multiple validator accounts in a single transaction, so I know that can't be the issue.

stake-pool/js/tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
   "compilerOptions": {
-    "target": "es2019",
-    "allowJs": true,
+    "baseUrl": ".",
```
joncinque, line 3: This should probably still be `src`, no?

stake-pool/js/src/buffer-layout.d.ts
```diff
@@ -1,16 +0,0 @@
-declare module 'buffer-layout' {
```
joncinque, line 1: `tsc` is failing without this -- could add it back in?  It might also be better for the `build` step to use `&&` instead of `;` to catch this in the future, ie:
```
"build": "npm run clean && tsc && cross-env NODE_ENV=production rollup -c",
```

stake-pool/js/src/copied-from-solana-web3/layout.ts
```diff
@@ -0,0 +1,78 @@
+import * as BufferLayout from '@solana/buffer-layout';
```
joncinque, line 1: Could you remove this file?

joncinque (commented):
A couple last bits and this should be good, pending @CMEONE 's issue

AlexanderRay:
> @AlexanderRay small issue: after doing some testing on devnet, I discovered that `withdrawStake` doesn't always properly honor the minimum balance for stake accounts:
> 
> This is one example of trying to withdraw lamports in a way that technically looks like it would be allowed given the total stake lamports in the validator account but doesn't honor the minimum amount of lamports that must be stored in a stake pool program validator account:
> 
> ```
> Transaction simulation failed: Error processing Instruction 2: custom program error: 0x17 
>     Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [1]
>     Program log: Instruction: Approve
>     Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 2024 of 200000 compute units
>     Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
>     Program 11111111111111111111111111111111 invoke [1]
>     Program 11111111111111111111111111111111 success
>     Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy invoke [1]
>     Program log: Instruction: WithdrawStake
>     Program log: Attempting to withdraw 99900000 lamports from validator account with 100000000 stake lamports, 1000000 must remain
>     Program log: Error: The lamports in the validator stake account is not equal to the minimum
>     Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy consumed 16664 of 200000 compute units
>     Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy failed: custom program error: 0x17
> ```
> 
> Here's another example, this time the amount requested for withdrawal is above the amount available in the validator account, and although there is another validator account that has enough lamports to make up the difference, the bindings still attempt to withdraw more from the first validator account than possible:
> 
> ```
> Transaction simulation failed: Error processing Instruction 2: custom program error: 0x17 
>     Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [1]
>     Program log: Instruction: Approve
>     Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 2024 of 200000 compute units
>     Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
>     Program 11111111111111111111111111111111 invoke [1]
>     Program 11111111111111111111111111111111 success
>     Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy invoke [1]
>     Program log: Instruction: WithdrawStake
>     Program log: Attempting to withdraw 102282880 lamports from validator account with 100000000 stake lamports, 1000000 must remain
>     Program log: Error: The lamports in the validator stake account is not equal to the minimum
>     Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy consumed 16683 of 200000 compute units
>     Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy failed: custom program error: 0x17
> ```
> 
> The bindings have previously dipped into multiple validator accounts and successfully withdrawn stakes from multiple validator accounts in a single transaction, so I know that can't be the issue.

can you please check now if i works properly on your side

CMEONE:
Ok @AlexanderRay I will check if it works and let you know, also for instances where there are more instructions that can be fit into a transaction (like if it has to withdraw stake from multiple validator accounts), how should the instructions be split most optimally (or how should they be grouped in transactions if the unstake requires multiple transactions)?

stake-pool/js/package.json
```diff
@@ -29,6 +27,7 @@
     "access": "public"
   },
   "browser": "dist.browser/index.js",
+  "module": "dist.browser/index.esm.js",
   "main": "dist/index.js",
```
CMEONE, line 31: Would it be possible to change this to `dist.browser/index.cjs.js` for compatibility with CommonJS?

AlexanderRay: better now?

stake-pool/js/package.json
```diff
@@ -26,17 +26,21 @@
...
   "browserslist": [
     "defaults",
-    "not IE 11"
+    "not IE 11",
+    "maintained node versions"
   ],
   "files": [
     "/dist",
-    "/dist.browser"
+    "/src"
```
joncinque, line 43: probably just need `dist` now and not this, right?

stake-pool/js/package.json
CMEONE: Yes, works perfectly, thank you!

CMEONE:
@AlexanderRay the minimum balance issue seems to be fixed for `withdrawStake`, but I am now sometimes getting another error:

```
ransaction simulation failed: Error processing Instruction 4: custom program error: 0xf 
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [1]
    Program log: Instruction: Approve
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 2024 of 200000 compute units
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
    Program 11111111111111111111111111111111 invoke [1]
    Program 11111111111111111111111111111111 success
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy invoke [1]
    Program log: Instruction: WithdrawStake
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [2]
    Program log: Instruction: Burn
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 2667 of 179733 compute units
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
    Program Stake11111111111111111111111111111111111111 invoke [2]
    Program Stake11111111111111111111111111111111111111 success
    Program Stake11111111111111111111111111111111111111 invoke [2]
    Program Stake11111111111111111111111111111111111111 success
    Program Stake11111111111111111111111111111111111111 invoke [2]
    Program Stake11111111111111111111111111111111111111 success
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA invoke [2]
    Program log: Instruction: Transfer
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA consumed 2781 of 156548 compute units
    Program TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA success
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy consumed 50390 of 200000 compute units
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy success
    Program 11111111111111111111111111111111 invoke [1]
    Program 11111111111111111111111111111111 success
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy invoke [1]
    Program log: Instruction: WithdrawStake
    Program log: Incorrect stake account address for vote vgcDar2pryHvMgPkKaZfh8pQy4BJxv7SpwUG7zinWjG, expected JCds2mFABFyurTV7NCwVSpG1JupXjP2ofjSRn6nCA82h, received HL5ixN3TipmVagAyYxtRW7xVjEwiL9rbJ7nd3RdqoZpY
    Program log: Error: Stake account address not properly derived from the validator address
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy consumed 49910 of 200000 compute units
    Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy failed: custom program error: 0xf
```

The expected address `JCds2mFABFyurTV7NCwVSpG1JupXjP2ofjSRn6nCA82h` is an activated stake account with a very low balance, the address input `HL5ixN3TipmVagAyYxtRW7xVjEwiL9rbJ7nd3RdqoZpY` is an activating stake account with a significantly higher balance, both are pointed to vote `vgcDar2pryHvMgPkKaZfh8pQy4BJxv7SpwUG7zinWjG`. I ran `withdrawStake` after adding stake to the `vgcDar2pryHvMgPkKaZfh8pQy4BJxv7SpwUG7zinWjG` from the CLI as a pool manager, I believe the inputted one was the stake account which was just created from CLI but is still activating and unmerged, while the expected one was the original stake account.

CMEONE:
The issue described above also seems to be the reason that [CI is now failing](https://github.com/solana-labs/solana-program-library/runs/5191491090?check_suite_focus=true).

```
[2022-02-14T22:11:54.411778810Z DEBUG solana_runtime::message_processor::stable_log] Program SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy invoke [1]
[2022-02-14T22:11:54.413795840Z DEBUG solana_runtime::message_processor::stable_log] Program log: Instruction: WithdrawStake
[2022-02-14T22:11:54.432528349Z DEBUG solana_runtime::message_processor::stable_log] Program log: Incorrect stake account address for vote DiyGABdP7sn1Z3Fdhr394nYcteY4dRKBgaRJDWSWmQGq, expected 9qM4cezSUh9jvBL91PboTrG6kph8gc4UoJJ2NCr4vQYt, received 5U3k2hovWDvFrbmRK4wT4ZTgtiwh3Zg4RSesF2aMd8Rw
[2022-02-14T22:11:54.433012780Z DEBUG solana_runtime::message_processor::stable_log] Program log: Error: Stake account address not properly derived from the validator address
```

stake-pool/js/package.json
AlexanderRay: if we check for example https://github.com/solana-labs/solana-web3.js/blob/master/package.json , then we see that "src" is also included. I think that its better to include

joncinque: ah all right, fine by me!

CMEONE:
@AlexanderRay I think the issue is that `withdrawStake` is including accounts which cannot be used for withdrawal such as transient stake accounts which have not yet been merged with the validator stake account. Would it be possible to remove those from the list of valid stake accounts which can be withdrawn, since that seems to be the source of the error?

stake-pool/js/src/utils/stake.ts
```diff
@@ -105,21 +125,26 @@ export async function prepareWithdrawAccounts(
   const withdrawFrom: WithdrawAccount[] = [];
   let remainingAmount = amount;
 
+  const fee = stakePool.stakeWithdrawalFee;
+  const inverseFee: Fee = {
+    numerator: fee.denominator.sub(fee.numerator),
+    denominator: fee.denominator,
+  };
+
   for (const type of ['preferred', 'active', 'transient', 'reserve']) {
```
CMEONE, line 134: I tried removing transient from this list, but then the the error said:

```
Program log: Error withdrawing from reserve: validator stake accounts have lamports available, please use those first.
    Program log: Error: The lamports in the validator stake account is not equal to the minimum
```

Maybe the issue is way back in the `findStakeProgramAddress` function?

CMEONE:
I just don't understand why I'm getting the error `Incorrect stake account address for vote <VOTE_ACCOUNT>, expected <ACTIVE_STAKE_ADDRESS>, received <ACTIVATING_TRANSIENT_STAKE_ADDRESS>` when I feel like the one that's being used/received is in fact the correct one in that context.

CMEONE:
I just don't understand why I'm getting the error `Incorrect stake account address for vote <VOTE_ACCOUNT>, expected <ACTIVE_STAKE_ADDRESS>, received <ACTIVATING_TRANSIENT_STAKE_ADDRESS>` when I feel like the one that's being used/received is in fact the correct one in that context. Maybe @joncinque you might be able to better interpret the reasoning behind this error?

stake-pool/js/src/utils/stake.ts
```diff
@@ -105,21 +125,26 @@ export async function prepareWithdrawAccounts(
...
-      let availableForWithdrawal = Math.floor(
-        calcPoolTokensForDeposit(stakePool, lamports),
-      );
-      if (!stakePool.stakeWithdrawalFee.denominator.isZero()) {
+    for (const { stakeAddress, voteAddress, lamports } of filteredAccounts) {
+      if (lamports <= minBalance) {
+        continue;
+      }
+
+      let availableForWithdrawal = calcPoolTokensForDeposit(stakePool, lamports - minBalance);
```
joncinque, line 142: @CMEONE the issue comes from this line.  Instead of always subtracting `minBalance`, it should only be subtracted from the transient accounts, as done in #2900.  The errors comes from not withdrawing everything possible from the main stake account before withdrawing from the transient stake account.

CMEONE: That makes sense, but when I tried modifying that line to `let availableForWithdrawal = calcPoolTokensForDeposit(stakePool, type == 'transient' ? lamports - minBalance : lamports);`, I still got the same error `Incorrect stake account address for vote <VOTE_ACCOUNT>, expected <ACTIVE_STAKE_ADDRESS>, received <ACTIVATING_TRANSIENT_STAKE_ADDRESS>`.

CMEONE: From the logs, it seems like the issue is not just with not withdrawing everything possible, it's also with trying to withdraw from transient stake accounts before the validator account.

These are the accounts available in the pool on devnet:
- Validator 1 active
- Validator 2 active
- Validator 3 active
- Validator 3 transient

This is the order that the bindings chose for withdrawing stake:
- Validator 1 active
- Validator 3 transient

As you see, the bindings chose to try to withdraw from validator 3 transient before trying validator 3 active.

CMEONE: Also, when `useReserve` is set to true, it tries withdrawing all the balance from the reserve even if there aren't enough lamports in reserve to cover it fully.

AlexanderRay:
@CMEONE @joncinque please check, last commit should fix the issue

CMEONE:
@AlexanderRay Unfortunately, I am still experiencing the same two issues. When `useReserve` is set to false, I am getting the error `Incorrect stake account address for vote <VOTE_ACCOUNT>, expected <ACTIVE_STAKE_ADDRESS>, received <ACTIVATING_TRANSIENT_STAKE_ADDRESS>` (the wrong validator account was used), and when `useReserve` is true, the reserve account is withdrawn from first, causing the error `The lamports in the validator stake account is not equal to the minimum`. I think there might be an issue on how the accounts get prioritized.

stake-pool/js/src/index.ts
```diff
@@ -254,15 +338,15 @@ export async function withdrawStake(
...
+      `Not enough token balance to withdraw ${lamportsToSol(poolAmount)} pool tokens.
+        Maximum withdraw amount is ${lamportsToSol(tokenAccount.amount.toNumber())} pool tokens.`,
     );
   }
 
+  const stakeAccountRentExemption = await connection.getMinimumBalanceForRentExemption(
+    StakeProgram.space,
+  );
+
   const withdrawAuthority = await findWithdrawAuthorityProgramAddress(
```
CMEONE, line 350: A few lines below this one, there is the following code:

```
if (useReserve) {
  withdrawAccounts.push({
    stakeAddress: stakePool.account.data.reserveStake,
    voteAddress: undefined,
    poolAmount,
  });
}
```

Won't this always fail if there is still stake in the validator accounts? In that case, I think it might be better to remove this part and pass `useReserve` into the `prepareWithdrawAccounts` method.

Then, in `prepareWithdrawAccounts`, we can change this:
```
if (reserveStakeBalance > 0) {
  accounts.push({
    type: 'reserve',
    stakeAddress: stakePool.reserveStake,
    lamports: reserveStakeBalance,
  });
}
```
to this:
```
if (useReserve && reserveStakeBalance > 0) {
  accounts.push({
    type: 'reserve',
    stakeAddress: stakePool.reserveStake,
    lamports: reserveStakeBalance,
  });
}
```

AlexanderRay:
are your test on your devnet? can you share your test environment with us?

stake-pool/js/src/index.ts
AlexanderRay: > A few lines below this one, there is the following code:
> 
> ```
> if (useReserve) {
>   withdrawAccounts.push({
>     stakeAddress: stakePool.account.data.reserveStake,
>     voteAddress: undefined,
>     poolAmount,
>   });
> }
> ```
> 
> Won't this always fail if there is still stake in the validator accounts? In that case, I think it might be better to remove this part and pass `useReserve` into the `prepareWithdrawAccounts` method.
> 
> Then, in `prepareWithdrawAccounts`, we can change this:
> 
> ```
> if (reserveStakeBalance > 0) {
>   accounts.push({
>     type: 'reserve',
>     stakeAddress: stakePool.reserveStake,
>     lamports: reserveStakeBalance,
>   });
> }
> ```
> 
> to this:
> 
> ```
> if (useReserve && reserveStakeBalance > 0) {
>   accounts.push({
>     type: 'reserve',
>     stakeAddress: stakePool.reserveStake,
>     lamports: reserveStakeBalance,
>   });
> }
> ```

When processing withdrawals, the order of priority goes:

preferred withdraw validator stake account (if set)
validator stake accounts
transient stake accounts
reserve stake account

CMEONE: Right, that's what it should be. However, when `useReserve` is set to `true`, the order becomes:

reserve stake account

which will never work because: `The lamports in the validator stake account is not equal to the minimum`

The stake pool program won't allow for a stake withdrawal from reserve stake account until all the validator stake accounts have been fully withdrawn.

AlexanderRay: argh ... checking .. anyway to withdraw from reserve account you should use withdrawSol

CMEONE: Right of course, `withdrawSol` is most optimal, `useReserve` seems a bit out of place as a parameter and can probably be removed altogether.

CMEONE: I think I have a solution for the other bug, I'll add another comment once I test that it works (currently dealing with splitting transactions up since transaction sizes can get a bit large sometimes).

stake-pool/js/src/utils/stake.ts
```diff
@@ -71,55 +94,60 @@ export async function prepareWithdrawAccounts(
...
-    const filteredAccounts = accounts.filter(a => a.type == type);
+    const filteredAccounts = accounts.filter((a) => a.type == type);
 
-    for (const {stakeAddress, voteAddress, lamports} of filteredAccounts) {
-      let availableForWithdrawal = Math.floor(
-        calcPoolTokensForDeposit(stakePool, lamports),
-      );
-      if (!stakePool.stakeWithdrawalFee.denominator.isZero()) {
+    for (const { stakeAddress, voteAddress, lamports } of filteredAccounts) {
+      if (lamports <= minBalance) {
```
CMEONE, line 141: I believe this condition should be changed to `(lamports <= minBalance && type == "transient")` because I think @joncinque was saying that the minimum balance part only applies to transient stake accounts.

CMEONE: With this fix, everything seems to be working on devnet.

CMEONE: The reserve part is still problematic, but from inspecting the issue, it seems like there's not much that can be done in the bindings themselves (seems to be an issue in the program itself). For instance, with completely empty validator stake accounts and all the transient stake set to the minimum balance, the program still returns:

```
Program log: Error withdrawing from reserve: validator stake accounts have lamports available, please use those first.
Program log: Error: The lamports in the validator stake account is not equal to the minimum
```

It seems like it might actually be impossible to withdraw stake from the reserve account even when all other accounts have been fully depleted! In that case, it's probably not worth spending too much time trying to fix that on the bindings side.

joncinque: I can confirm the reserve part does have a corner case, since it's impossible to totally drain a transient account.  I think it's possible to relax that requirement, but I have to think through the implications.

AlexanderRay: 'use_reserve' functionality is just reimplemented bases on cli version

https://github.com/solana-labs/solana-program-library/blob/3d92f8f4dc4fab63b9eb68f33185d9c2829abcfe/stake-pool/cli/src/main.rs#L1373

@joncinque  should I also remove it from cli or make it still sense there?

stake-pool/js/src/utils/stake.ts
```diff
@@ -71,55 +94,60 @@ export async function prepareWithdrawAccounts(
...
+        continue;
```
CMEONE, line 142: Please change the above condition to `(lamports <= minBalance && type == "transient")` as the `minBalance` only applies to transient stake accounts. I believe that is all that needs to be fixed (besides `useReserve` which is a larger issue related to reserve accounts in general).

CMEONE (changes requested):

CMEONE (approved):
Thoroughly tested the following methods on devnet: `getStakePoolAccount`, `updateStakePool`, `depositSol`, `withdrawSol`, `withdrawStake`, `stakePoolInfo`

If @joncinque thinks everything looks good, I think this PR can be merged (and pushed to NPM with a minor version bump), and we can revisit the reserve issues in another PR (since that's a larger issue that involves the program itself).

I can't wait to launch [BlazeStake](https://stake.solblaze.org/) on mainnet-beta soon, which would not have been possible without both of you. Thank you so much for your work on these bindings (and for persisting despite all the errors)!

joncinque (approved):
Thanks for your patience, looks great!

PR #2807, opened 2022-01-25
token-2022: Assess transfer fees

joncinque:
#### Problem

Transfer fees need to get assessed during transfers.

#### Solution

Assess those fees!  The commits on this were done purposely:

* commit 1 just reorders the normal transfer logic using `StateWithExtensionsMut`, no logic changes
* commit 2 adds some simple `TokenClient`-based tests, mostly boiler-plate
* commit 3 implements the transfer fees
* commit 4 adds more tests

### DIFF
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -6,7 +6,7 @@ use crate::{
     extension::{
         confidential_transfer::{self, ConfidentialTransferAccount},
         mint_close_authority::MintCloseAuthority,
-        transfer_fee::{self, TransferFeeConfig},
+        transfer_fee::{self, TransferFeeAmount, TransferFeeConfig},
         ExtensionType, StateWithExtensions, StateWithExtensionsMut,
     },
     instruction::{is_valid_signer_index, AuthorityType, TokenInstruction, MAX_SIGNERS},
@@ -15,6 +15,7 @@ use crate::{
 use num_traits::FromPrimitive;
 use solana_program::{
     account_info::{next_account_info, AccountInfo},
+    clock::Clock,
     decode_error::DecodeError,
     entrypoint::ProgramResult,
     msg,
@@ -228,7 +229,7 @@ impl Processor {
         accounts: &[AccountInfo],
         amount: u64,
         expected_decimals: Option<u8>,
-        _expected_fee: Option<u64>,
+        expected_fee: Option<u64>,
     ) -> ProgramResult {
         let account_info_iter = &mut accounts.iter();
 
@@ -253,7 +254,7 @@ impl Processor {
         if source_account.base.amount < amount {
             return Err(TokenError::InsufficientFunds.into());
         }
-        if let Some((mint_info, expected_decimals)) = expected_mint_info {
+        let fee = if let Some((mint_info, expected_decimals)) = expected_mint_info {
             if !cmp_pubkeys(&source_account.base.mint, mint_info.key) {
                 return Err(TokenError::MintMismatch.into());
             }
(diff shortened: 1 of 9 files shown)


token/program-2022/src/processor.rs
```diff
@@ -312,16 +336,43 @@ impl Processor {
...
+        // self-transfer was dealt with earlier, so this *should* be safe
+        let mut dest_account_data = dest_account_info.data.borrow_mut();
+        let mut dest_account = StateWithExtensionsMut::<Account>::unpack(&mut dest_account_data)?;
+
+        if dest_account.base.is_frozen() {
+            return Err(TokenError::AccountFrozen.into());
+        }
+        if !cmp_pubkeys(&source_account.base.mint, &dest_account.base.mint) {
+            return Err(TokenError::MintMismatch.into());
+        }
```
CriesofCarrots, line 348: Huh, I'm sort of surprised that changing this ordering didn't break any tests. Sweet!

token/program-2022/src/processor.rs
```diff
@@ -312,16 +336,43 @@ impl Processor {
             return Ok(());
         }
 
-        source_account.amount = source_account
+        // self-transfer was dealt with earlier, so this *should* be safe
```
CriesofCarrots, line 339: So just confirming, self transfers with fees should not have fees collected? (Though I guess you can still test the expected fee with TransferCheckedWithFee.)
It would be nice to have a test case in transfer_fee.rs demonstrating/reiterating no fees on self-transfers.

token/program-2022/tests/transfer.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    // self transfer is ok
+    token
+        .transfer_checked(&alice_account, &alice_account, &alice, 1, decimals)
+        .await
+        .unwrap();
+}
+
+#[tokio::test]
+async fn transfer_fee() {
```
CriesofCarrots, line 190: nit: I haven't hit on the best alternative, but I found this test name confusing, since there is no fee moving.

CriesofCarrots (dismissed):
Looks correct to me. Just one test case that I'd like to see, assuming that it is correct that no fees move on a self-transfer

token/program-2022/src/processor.rs
joncinque: Yeah that was the nicest surprise of all

token/program-2022/src/processor.rs
joncinque: Correct, the idea here is that self-transfers do not accrue fees, based on the idea that self-transfers signed by a delegate don't eat until the amount delegated.  Ie, I approve 10 tokens to you, and if you self transfer them, your 10 approved tokens don't change.

That's not a great analogy though, so I don't feel strongly about this.  Maybe @mvines or @samkim-crypto have an opinion on the desired behavior?

Either way, I'll certainly add a test case for this.

token/program-2022/tests/transfer.rs
joncinque: `transfer_with_fee_on_mint_without_fee`?

CriesofCarrots: wfm!

CriesofCarrots (approved):
lgtm!

mvines (commented):
lgtm2 once the question regarding self-transfer fee behavior is resolved.

token/program-2022/src/processor.rs
mvines: > the idea here is that self-transfers do not accrue fees

👍🏼

PR #2808, opened 2022-01-26
token-2022: Enforce maximum transfer fee of 10k bps

joncinque:
#### Problem

It's possible to set a transfer fee above 10k basis points, which means that you assess more than all of the amount transferred.  Discovered this one while testing transfer fees.

#### Solution

Enforce a maximum transfer fee of 10,000 basis points.


CriesofCarrots (approved):
Nice one ;)

PR #2810, opened 2022-01-26
token-2022: add DefaultAccountState mint extension

CriesofCarrots:
This PR adds a Mint-targeted extension that allows mints to define the default AccountState for new Accounts. This provides a convenient way for mint operators to establish a whitelist of accounts/users by having all new Accounts default to AccountState::Frozen, and manually thawing approved accounts.

I made the assumption that Mints with Some freeze_authority should be able to update the default Account state of new Accounts after mint initialization, allowing a Mint to stop the whitelist approach in the future.

Currently, there is nothing to prevent a Mint operator from unsetting their freeze authority while the DefaultAccountState is set to Frozen, meaning all new accounts will be frozen and un-thawable. It could be desirable to do this, thus limiting the number of active Accounts for a Mint forever. Or perhaps this is a footgun we should prevent in SetAuthority. Thoughts?

### DIFF
--- a/token/program-2022/src/extension/default_account_state/instruction.rs
+++ b/token/program-2022/src/extension/default_account_state/instruction.rs
@@ -101,7 +101,7 @@ pub fn update_default_account_state(
     check_program_account(token_program_id)?;
     let mut accounts = vec![
         AccountMeta::new(*mint, false),
-        AccountMeta::new_readonly(*freeze_authority, true),
+        AccountMeta::new_readonly(*freeze_authority, signers.is_empty()),
     ];
     for signer_pubkey in signers.iter() {
         accounts.push(AccountMeta::new_readonly(**signer_pubkey, true));
--- a/token/program-2022/src/extension/default_account_state/instruction.rs
+++ b/token/program-2022/src/extension/default_account_state/instruction.rs
@@ -1,0 +1,115 @@
+use {
+    crate::{
+        check_program_account, error::TokenError, instruction::TokenInstruction,
+        state::AccountState,
+    },
+    num_enum::{IntoPrimitive, TryFromPrimitive},
+    solana_program::{
+        instruction::{AccountMeta, Instruction},
+        program_error::ProgramError,
+        pubkey::Pubkey,
+    },
+    std::convert::TryFrom,
+};
+
+/// Default Account State extension instructions
+#[derive(Clone, Copy, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum DefaultAccountStateInstruction {
+    /// Initialize a new mint with the default state for new Accounts.
+    ///
+    /// Fails if the mint has already been initialized, so must be called before
+    /// `InitializeMint`.
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
+    /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
+    /// then space required for this extension, plus any others.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint to initialize.
+    InitializeDefaultAccountState,
+    /// Update the default state for new Accounts. Only supported for mints that include the
+    /// `DefaultAccountState` extension.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[signer]` The mint freeze authority.
+    ///
+    ///   * Multisignature authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[]` The mint's multisignature freeze authority.
+    ///   2. ..2+M `[signer]` M signer accounts.
+    UpdateDefaultAccountState,
+}
+
+pub(crate) fn decode_instruction(
+    input: &[u8],
+) -> Result<(DefaultAccountStateInstruction, AccountState), ProgramError> {
+    if input.len() != 2 {
+        return Err(TokenError::InvalidInstruction.into());
+    }
+    Ok((
+        DefaultAccountStateInstruction::try_from(input[0])
+            .or(Err(TokenError::InvalidInstruction))?,
+        AccountState::try_from(input[1]).or(Err(TokenError::InvalidInstruction))?,
+    ))
+}
+
+fn encode_instruction(
+    token_program_id: &Pubkey,
+    accounts: Vec<AccountMeta>,
+    instruction_type: DefaultAccountStateInstruction,
+    state: &AccountState,
+) -> Instruction {
+    let mut data = TokenInstruction::DefaultAccountStateExtension.pack();
+    data.push(instruction_type.into());
+    data.push((*state).into());
+    Instruction {
+        program_id: *token_program_id,
+        accounts,
+        data,
+    }
+}
+
+/// Create an `InitializeDefaultAccountState` instruction
+pub fn initialize_default_account_state(
+    token_program_id: &Pubkey,
+    mint: &Pubkey,
+    state: &AccountState,
+) -> Result<Instruction, ProgramError> {
+    check_program_account(token_program_id)?;
+    let accounts = vec![AccountMeta::new(*mint, false)];
+    Ok(encode_instruction(
+        token_program_id,
+        accounts,
+        DefaultAccountStateInstruction::InitializeDefaultAccountState,
+        state,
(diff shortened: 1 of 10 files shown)


token/program-2022/src/extension/default_account_state/instruction.rs
```diff
@@ -0,0 +1,115 @@
...
+    token_program_id: &Pubkey,
+    mint: &Pubkey,
+    freeze_authority: &Pubkey,
+    signers: &[&Pubkey],
+    state: &AccountState,
+) -> Result<Instruction, ProgramError> {
+    check_program_account(token_program_id)?;
+    let mut accounts = vec![
+        AccountMeta::new(*mint, false),
+        AccountMeta::new_readonly(*freeze_authority, true),
```
joncinque, line 104: For multisigs, right?
```suggestion
        AccountMeta::new_readonly(*freeze_authority, signers.is_empty()),
```

token/program-2022/src/extension/default_account_state/instruction.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// `InitializeMint`.
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
+    /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
+    /// then space required for this extension, plus any others.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint to initialize.
+    InitializeDefaultAccountState,
```
joncinque, line 31: nit: how about just `Initialize` instead?

token/program-2022/src/extension/default_account_state/instruction.rs
```diff
@@ -0,0 +1,115 @@
...
+    ///
+    ///   * Single authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[signer]` The mint freeze authority.
+    ///
+    ///   * Multisignature authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[]` The mint's multisignature freeze authority.
+    ///   2. ..2+M `[signer]` M signer accounts.
+    UpdateDefaultAccountState,
```
joncinque, line 45: nit: how about just `Update` instead?

token/program-2022/tests/default_account_state.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    let state = token.get_mint_info().await.unwrap();
+    assert_eq!(state.base.decimals, decimals);
+    assert_eq!(
+        state.base.mint_authority,
+        COption::Some(mint_authority.pubkey())
+    );
+    assert_eq!(state.base.supply, 0);
+    assert!(state.base.is_initialized);
+    assert_eq!(state.base.freeze_authority, COption::None);
```
joncinque, line 44: This is really specific, but what do you think about adding in a check to make sure that this situation doesn't happen?  Ie, initializing a mint that has default account state of Frozen, but then no configured freeze authority.  It's either that, or use include a configuration authority in the default account state extension

token/program-2022/tests/program_test.rs
```diff
@@ -52,14 +69,19 @@ impl TestContext {
         let mint_account = Keypair::new();
         let mint_authority = Keypair::new();
         let mint_authority_pubkey = mint_authority.pubkey();
+        let freeze_authority = if enable_freeze {
+            Some(&mint_authority_pubkey)
+        } else {
+            None
+        };
```
joncinque, line 76: nittiest of nits, but how about a separate `Option<Keypair>` field on `TokenContext` for this?  Then we can be sure that we aren't reusing the same authority elsewhere by accident.

joncinque (commented):
Unsetting freeze authority sounds fine to me, since it's the current behavior and does have the added benefit of making it impossible to ever create a new account, as you mention.

token/program-2022/src/extension/default_account_state/instruction.rs
CriesofCarrots: Thanks for catching this!

token/program-2022/tests/default_account_state.rs
CriesofCarrots: Ah, good point. I'll add that check to InitializeMint (has to happen there, since that ix is last).

token/program-2022/src/extension/default_account_state/instruction.rs
```diff
@@ -0,0 +1,115 @@
...
+        program_error::ProgramError,
+        pubkey::Pubkey,
+    },
+    std::convert::TryFrom,
+};
+
+/// Default Account State extension instructions
+#[derive(Clone, Copy, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum DefaultAccountStateInstruction {
```
mvines, line 18: These two instructions could use documentation about the required instruction data (an `AccountState`)

token/program-2022/src/extension/default_account_state/instruction.rs
```diff
@@ -0,0 +1,115 @@
...
+    ///   0. `[writable]` The mint to initialize.
+    Initialize,
+    /// Update the default state for new Accounts. Only supported for mints that include the
+    /// `DefaultAccountState` extension.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[signer]` The mint freeze authority.
```
mvines, line 39: What if we added a DefaultAccountState authority instead?  That could still be set as the freeze authority if the user wants.  But this is an expensive extension (data wise) already so what's another 32bytes

token/program-2022/src/extension/default_account_state/instruction.rs
```diff
@@ -0,0 +1,115 @@
...
+        pubkey::Pubkey,
+    },
+    std::convert::TryFrom,
+};
+
+/// Default Account State extension instructions
+#[derive(Clone, Copy, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum DefaultAccountStateInstruction {
+    /// Initialize a new mint with the default state for new Accounts.
```
mvines, line 19: 🤔 so this is just the base account state right?  That is we can't set default extension state currently.

TBH, I'm slightly worried this is overall more than what we really need, which I think is just the ability to force new accounts to be frozen

token/program-2022/src/extension/default_account_state/instruction.rs
CriesofCarrots: Not sure I'm following why it's expensive, data-wise. The Mint extension is only extension-type (1) + length (2) + AccountState (1) = 4 bytes.
I did consider a separate authority, but adding 32 bytes seemed like too much.

token/program-2022/src/extension/default_account_state/instruction.rs
CriesofCarrots: It could be a bool instead for "init Frozen". Doesn't change the data-size of the extension, but the processing would be slightly simpler (but also less flexible, if there are additional AccountState variants in the future).

token/program-2022/src/extension/default_account_state/instruction.rs
mvines: jfc, my brain swapped `AccountState` for `state::Account` during my previous reading!

CriesofCarrots: Ohhhh 😅  your comment makes perfect sense, then!

token/program-2022/src/extension/default_account_state/instruction.rs
CriesofCarrots: Oh true. I missed your pattern for this.

token/program-2022/src/extension/default_account_state/instruction.rs
CriesofCarrots: ... given that clarification, how is this looking to you @mvines ?

mvines:
> Currently, there is nothing to prevent a Mint operator from unsetting their freeze authority while the DefaultAccountState is set to Frozen, meaning all new accounts will be frozen and un-thawable. It could be desirable to do this, thus limiting the number of active Accounts for a Mint forever.

I think this is kinda neat to leave in as it enables a feature that would otherwise not be possible. Seems like a fair trade-off to leaving the footgun in.

mvines (approved):

PR #2813, opened 2022-01-26
token-2022: remove split_at usages

CriesofCarrots:
`Slice::split_at(n)` panics on slices shorter than `n`. We're using split_at several places in TokenInstruction unpacking, which means that spl_token can panic on malformed instructions. The BPF loader catches and unwinds these panics, but it seems pretty "meh" to depend on that, instead of handling failures gracefully. It also makes debugging -- with ProgramTest and in the wild -- a bit harder than it needs to be.

This PR also adds constants to make slice indexing a little more self-documenting.


joncinque (approved):
nice, looks good!

PR #2821, opened 2022-01-28
token-2022: add extension realloc helper

CriesofCarrots:
In working on the MemoTransfer extension, I found I needed an api to determine if reallocation was needed. Might be useful for confidental transfers as well.

Thoughts?

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -491,6 +491,24 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     fn get_first_extension_type(&self) -> Result<Option<ExtensionType>, ProgramError> {
         get_first_extension_type(self.tlv_data)
     }
+
+    /// Compares the length of an extension with the currently used TLV buffer to determine if
+    /// reallocation is needed. If so, returns Some(v), where v is the difference between current
+    /// space and needed.
+    #[allow(dead_code)]
+    pub(crate) fn realloc_needed(
+        &self,
+        new_extension: ExtensionType,
+    ) -> Result<Option<usize>, ProgramError> {
+        let current_extensions = self.get_extension_types()?;
+        let needed_tlv_len = ExtensionType::get_total_tlv_len(&current_extensions);
+        let new_needed_tlv_len = needed_tlv_len.saturating_add(new_extension.get_type_len());
+        if self.tlv_data.len() >= new_needed_tlv_len {
+            Ok(None)
+        } else {
+            Ok(Some(new_needed_tlv_len - self.tlv_data.len())) // arithmetic safe because of if clause
+        }
+    }
 }
 
 /// Different kinds of accounts. Note that `Mint`, `Account`, and `Multisig` types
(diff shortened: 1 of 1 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -471,6 +471,24 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     fn get_first_extension_type(&self) -> Result<Option<ExtensionType>, ProgramError> {
         get_first_extension_type(self.tlv_data)
     }
+
+    /// Compares the length of an extension with the currently used TLV buffer to determine if
+    /// reallocation is needed. If so, returns Some(v), where v is the difference between current
+    /// space and needed.
+    #[allow(dead_code)]
+    pub(crate) fn realloc_needed(
```
mvines, line 479: wdyt about wrapping this up within `init_extension()` as implicit behavior?  If that existed, the creator of a new account or mint wouldn't even need to bother calculating the needed account data length ahead of time, as each new extension initializer would realloc the room it needs (assuming they don't mind the extra compute units that realloc would consume)

CriesofCarrots: That is appealing.
The one reason I have for keeping it explicit is that the reallocation needs to come with a movement of lamports to cover additional rent-exempt reserve. Doesn't really change the logic of getting the new minimum balance or transferring the diff, just a reminder to do so. Not wedded to that approach, though.

CriesofCarrots: Ha, just realized that we just fall over with an index panic if we try to init_extension on a buffer that's too small for it. So would be good to handle len checks regardless.

mvines: > the reallocation needs to come with a movement of lamports to cover additional rent-exempt reserve

Are you thinking we CPI over to the system program to do this, or make it the callers problem to ensure the affected account will remain rent exempt after the realloc?

CriesofCarrots: > Are you thinking we CPI over to the system program to do this, or make it the callers problem to ensure the affected account will remain rent exempt after the realloc?

I was planning on making it the caller's (my) problem in the extension ix processor.
I'll push my draft of the MemoTransfer extension so you can see what I was thinking: https://github.com/solana-labs/solana-program-library/pull/2822/files#diff-04097469c65d3fc1f76df183abc5e43ed245647e4fecae892a9b607f31ea3015R55

> wrapping this up within `init_extension()` as implicit behavior

Back on this topic, one real reason not to do this may be that StateWithExtensionsMut doesn't carry the AccountInfo that's needed to call the `realloc` syscall

mvines: Got it, that draft looks good! That calling code in the MemoTransfer extension could be factored out into a stand-alone function, it looks like the ConfidentialTransfer extension would essentially need exactly that too. I'm happy to do that in a follow-up when I steal that code for ConfidentialTransfer

joncinque: What if we draw a line between pre-init and post-init extensions, and pre-init extensions always require preallocation, and post-init extensions always realloc?  Then we can also separate `init_extension` and `init_extension_with_realloc` which would also accept the payer `AccountInfo`.

As for the implementation on `StateWithExtensionsMut`, we could always just do the unsafe byte manipulation ourselves and get around requiring `Accountinfo`, ie copy this https://github.com/solana-labs/solana/blob/aab70b3aee2d39ce54842bc5011ffc44c8375cfd/sdk/program/src/account_info.rs#L128

mvines:
> What if we draw a line between pre-init and post-init extensions, and pre-init extensions always require preallocation, and post-init extensions always realloc? Then we can also separate init_extension and init_extension_with_realloc which would also accept the payer AccountInfo.

Seems like that should work fine.

> As for the implementation on StateWithExtensionsMut, we could always just do the unsafe byte manipulation ourselves and get around requiring Accountinfo, ie copy this https://github.com/solana-labs/solana/blob/aab70b3aee2d39ce54842bc5011ffc44c8375cfd/sdk/program/src/account_info.rs#L128

Veto!

joncinque:
> Veto!

Deep down, I think I was unconsciously trolling :smiling_imp:

token/program-2022/src/extension/mod.rs
CriesofCarrots: Does it seem at all useful to merge this as-is (and then the MemoTransfer initial implementation), and then refactor into separate `init_extension`s afterward?

mvines: sgtm

joncinque (approved):

PR #2822, opened 2022-01-28
token-2022: Add MemoTransfer extension

### LINKED ISSUE #2729: token: add extension to support TransferWithMemo
Problem:
Some token mint creators want to require transfers to be accompanied by a memo, and fail those without.

Proposed Solution:
Add a token extension that marks an account as "memo-required"

CriesofCarrots:
Some users would like to require all transfers into particular accounts be accompanied by a memo.
This PR implements a MemoTransfer extension, which stores a byte indicating whether memo-transfers are or are not required. The extension can be initialized on any existing token Account, assuming the account has space for the extension. Callers should call Reallocate to make space, if the account is too small.

Currently, there is only a placeholder where the memo-requirement would be in `_process_transfer()`, pending the syscall that will expose previous sibling instructions to be checked for the memo. Bumping the solana crates and using the syscall will be done in a subsequent PR.

Needs:
- [x] rebase on #2864

Closes #2729

### DIFF
--- a/token/program-2022/src/extension/memo_transfer/instruction.rs
+++ b/token/program-2022/src/extension/memo_transfer/instruction.rs
@@ -1,0 +1,113 @@
+use {
+    crate::{check_program_account, error::TokenError, instruction::TokenInstruction},
+    num_enum::{IntoPrimitive, TryFromPrimitive},
+    solana_program::{
+        instruction::{AccountMeta, Instruction},
+        program_error::ProgramError,
+        pubkey::Pubkey,
+    },
+    std::convert::TryFrom,
+};
+
+/// Default Account State extension instructions
+#[derive(Clone, Copy, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum RequiredMemoTransfersInstruction {
+    /// Require memos for transfers into this Account. Adds the MemoTransfer extension to the
+    /// Account, if it doesn't already exist. In this case,
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The account to update.
+    ///   1. `[signer]` The account's owner.
+    ///
+    ///   * Multisignature authority
+    ///   0. `[writable]` The account to update.
+    ///   1. `[]` The account's multisignature owner.
+    ///   2. ..2+M `[signer]` M signer accounts.
+    ///
+    Enable,
+    /// Stop requiring memos for transfers into this Account.
+    ///
+    /// Fails if the account does not have the extension present.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The account to update.
+    ///   1. `[signer]` The account's owner.
+    ///
+    ///   * Multisignature authority
+    ///   0. `[writable]` The account to update.
+    ///   1. `[]`  The account's multisignature owner.
+    ///   2. ..2+M `[signer]` M signer accounts.
+    ///
+    Disable,
+}
+
+pub(crate) fn decode_instruction(
+    input: &[u8],
+) -> Result<RequiredMemoTransfersInstruction, ProgramError> {
+    if input.len() != 1 {
+        return Err(TokenError::InvalidInstruction.into());
+    }
+    RequiredMemoTransfersInstruction::try_from(input[0])
+        .map_err(|_| TokenError::InvalidInstruction.into())
+}
+
+fn encode_instruction(
+    token_program_id: &Pubkey,
+    accounts: Vec<AccountMeta>,
+    instruction_type: RequiredMemoTransfersInstruction,
+) -> Instruction {
+    let mut data = TokenInstruction::MemoTransferExtension.pack();
+    data.push(instruction_type.into());
+    Instruction {
+        program_id: *token_program_id,
+        accounts,
+        data,
+    }
+}
+
+/// Create an `Enable` instruction
+pub fn enable_required_transfer_memos(
+    token_program_id: &Pubkey,
+    account: &Pubkey,
+    owner: &Pubkey,
+    signers: &[&Pubkey],
+) -> Result<Instruction, ProgramError> {
+    check_program_account(token_program_id)?;
+    let mut accounts = vec![
+        AccountMeta::new(*account, false),
+        AccountMeta::new_readonly(*owner, signers.is_empty()),
+    ];
+    for signer_pubkey in signers.iter() {
+        accounts.push(AccountMeta::new_readonly(**signer_pubkey, true));
+    }
+    Ok(encode_instruction(
+        token_program_id,
+        accounts,
+        RequiredMemoTransfersInstruction::Enable,
+    ))
+}
+
+/// Create a `Disable` instruction
+pub fn disable_required_transfer_memos(
+    token_program_id: &Pubkey,
+    account: &Pubkey,
+    owner: &Pubkey,
+    signers: &[&Pubkey],
+) -> Result<Instruction, ProgramError> {
+    check_program_account(token_program_id)?;
+    let mut accounts = vec![
+        AccountMeta::new(*account, false),
+        AccountMeta::new_readonly(*owner, signers.is_empty()),
+    ];
+    for signer_pubkey in signers.iter() {
+        accounts.push(AccountMeta::new_readonly(**signer_pubkey, true));
+    }
+    Ok(encode_instruction(
+        token_program_id,
+        accounts,
+        RequiredMemoTransfersInstruction::Disable,
+    ))
+}
--- a/token/program-2022/src/extension/memo_transfer/instruction.rs
+++ b/token/program-2022/src/extension/memo_transfer/instruction.rs
@@ -14,7 +14,7 @@ use {
 #[repr(u8)]
 pub enum RequiredMemoTransfersInstruction {
     /// Require memos for transfers into this Account. Adds the MemoTransfer extension to the
-    /// Account, if it doesn't already exist. In this case,
+    /// Account, if it doesn't already exist.
     ///
     /// Accounts expected by this instruction:
     ///
--- a/token/program-2022/src/extension/memo_transfer/mod.rs
+++ b/token/program-2022/src/extension/memo_transfer/mod.rs
@@ -1,0 +1,33 @@
+use {
+    crate::{
+        extension::{Extension, ExtensionType, StateWithExtensionsMut},
+        pod::PodBool,
+        state::Account,
+    },
+    bytemuck::{Pod, Zeroable},
+};
+
+/// Memo Transfer extension instructions
+pub mod instruction;
+
+/// Memo Transfer extension processor
+pub mod processor;
+
+/// Memo Transfer extension for Accounts
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct MemoTransfer {
+    /// Require transfers into this account to be accompanied by a memo
+    pub require_incoming_transfer_memos: PodBool,
+}
+impl Extension for MemoTransfer {
+    const TYPE: ExtensionType = ExtensionType::MemoTransfer;
+}
+
+/// Determine if a memo is required for transfers into this account
+pub fn memo_required(mut account_state: StateWithExtensionsMut<Account>) -> bool {
+    if let Ok(extension) = account_state.get_extension_mut::<MemoTransfer>() {
+        return extension.require_incoming_transfer_memos.into();
+    }
+    false
+}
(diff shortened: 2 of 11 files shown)


CriesofCarrots:
Just fyi that this PR needs a lot of work around reallocation; coming soon

token/program-2022/src/extension/memo_transfer/instruction.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The account to update.
+    ///   1. `[writeable, signer]` The payer account to fund realloc, if needed
+    ///   2. `[signer]` The account's owner.
+    ///
+    ///   * Multisignature authority
+    ///   0. `[writable]` The account to update.
+    ///   1. `[writeable, signer]` The payer account to fund realloc, if needed
+    ///   2. `[]` System program for realloc funding, if needed
```
CriesofCarrots, line 28: These are only needed when reallocation is required to make space for the extension... Currently, they are required for all Require instructions, even if the account won't need reallocation. It's not super clean, but we could make them optional. Thoughts?

token/program-2022/src/processor.rs
```diff
@@ -371,6 +372,12 @@ impl Processor {
             return Err(TokenError::MintMismatch.into());
         }
 
+        if let Ok(extension) = dest_account.get_extension_mut::<MemoTransfer>() {
+            if extension.require_incoming_transfer_memos != 0 {
+                // TODO: use get_processed_instructions syscall to check for memo
```
CriesofCarrots, line 377: Just noting this is where the update will be needed when https://github.com/solana-labs/solana/pull/22859 is available. Until then, all transactions into accounts requiring transfer-memos will succeed.

CriesofCarrots:
Not fully functional, due to syscall, but everything else ready for a look!

token/program-2022/src/extension/memo_transfer/processor.rs
```diff
@@ -0,0 +1,140 @@
...
+            )?;
+
+            // reborrow account state
+            let mut account_data = token_account_info.data.borrow_mut();
+            let mut account = if needs_account_type_init {
+                StateWithExtensionsMut::<Account>::unpack_after_realloc(&mut account_data)
+            } else {
+                StateWithExtensionsMut::<Account>::unpack(&mut account_data)
+            }?;
+            let extension = account.init_extension::<MemoTransfer>()?;
```
CriesofCarrots, line 85: @mvines , this isn't as clean as I originally wrote it, due to handling the mutable borrows. But should still be generalizable on T: ExtensionType

token/program-2022/src/extension/memo_transfer/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+impl Extension for MemoTransfer {
+    const TYPE: ExtensionType = ExtensionType::MemoTransfer;
+}
+
+type PodBool = u8;
+
+/// Determine if a memo is required for transfers into this account
+pub fn memo_required(mut account_state: StateWithExtensionsMut<Account>) -> bool {
+    if let Ok(extension) = account_state.get_extension_mut::<MemoTransfer>() {
```
joncinque, line 30: nit: it might be worth adding a `get_extension` call to `StateWithExtensionsMut` to avoid having to take this by mut

token/program-2022/src/extension/memo_transfer/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub require_incoming_transfer_memos: PodBool,
+}
+impl Extension for MemoTransfer {
+    const TYPE: ExtensionType = ExtensionType::MemoTransfer;
+}
+
+type PodBool = u8;
+
+/// Determine if a memo is required for transfers into this account
+pub fn memo_required(mut account_state: StateWithExtensionsMut<Account>) -> bool {
```
joncinque, line 29: This should probably take a ref instead

token/program-2022/src/extension/memo_transfer/processor.rs
joncinque: Seems like the cleanest solution is to have `StateWithExtensionsMut` hold onto the full `AccountInfo` in the end.  Then you won't need to worry about exposing the drops and `account_type_is_empty()`

token/program-2022/src/extension/memo_transfer/processor.rs
```diff
@@ -0,0 +1,140 @@
...
+    check_program_account(program_id)?;
+
+    let instruction = decode_instruction(input)?;
+    match instruction {
+        MemoTransferInstruction::Require => {
+            msg!("MemoTransferInstruction::Require");
+            process_require_memo_transfers(program_id, accounts)
+        }
+        MemoTransferInstruction::UnRequire => {
+            msg!("MemoTransferInstruction::Require");
```
joncinque, line 136:
```suggestion
            msg!("MemoTransferInstruction::UnRequire");
```

joncinque (commented):
Looks great for the most part! The realloc part is a bit tough, so we might need to refactor `StateWithExtensionsMut`

token/program-2022/src/extension/memo_transfer/instruction.rs
```diff
@@ -0,0 +1,116 @@
...
+#[repr(u8)]
+pub enum MemoTransferInstruction {
+    /// Require memos for transfers into this Account. Adds the MemoTransfer extension to the
+    /// Account, if it doesn't already exist. In this case,
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The account to update.
+    ///   1. `[writeable, signer]` The payer account to fund realloc, if needed
+    ///   2. `[signer]` The account's owner.
```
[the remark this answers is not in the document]
CriesofCarrots: Correct, this is opt in for account owners per individual account.

token/program-2022/src/extension/memo_transfer/instruction.rs
```diff
@@ -0,0 +1,116 @@
...
+    ///
+    ///   0. `[writable]` The account to update.
+    ///   1. `[signer]` The account's owner.
+    ///
+    ///   * Multisignature authority
+    ///   0. `[writable]` The account to update.
+    ///   1. `[]`  The account's multisignature owner.
+    ///   2. ..2+M `[signer]` M signer accounts.
+    ///
+    UnRequire,
```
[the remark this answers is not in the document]
CriesofCarrots: I know 😞 Very open to a better term here

token/program-2022/src/extension/memo_transfer/instruction.rs
```diff
@@ -0,0 +1,116 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum MemoTransferInstruction {
+    /// Require memos for transfers into this Account. Adds the MemoTransfer extension to the
+    /// Account, if it doesn't already exist. In this case,
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The account to update.
+    ///   1. `[writeable, signer]` The payer account to fund realloc, if needed
```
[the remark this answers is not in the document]
CriesofCarrots: Ew, actually that's just broken copy-pasta. Will 🪓

token/program-2022/src/extension/memo_transfer/instruction.rs
CriesofCarrots: Does a 3rd instruction help? On `Init`, the account still may or may not be big enough to hold the new extension, so the payer and system program will still end up optional.

mvines: What if we have a more general `Realloc` instruction, where the caller passes in a list of additional `ExtensionTypes` that it wants to ensure the account has space for.  The account owner doesn't even need to be a signer there, only the funding account.

token/program-2022/src/extension/memo_transfer/instruction.rs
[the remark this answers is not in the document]
mvines: Maybe good old `Enable` and `Disable`?  That is, to enable or disable memo transfers.

token/program-2022/src/extension/memo_transfer/instruction.rs
CriesofCarrots: Oh, that seems nice. I think the account owner should still be a signer, though.
There might be reasons I don't want just anyone to be able to make my account huge, even if they pay the rent for it, eg. if tx-cost-model charges more load large accounts.

CriesofCarrots: I'll PR my vision of a `Realloc` instruction for discussion

mvines: Awesome.

> There might be reasons I don't want just anyone to be able to make my account huge

I don't think I ever checked to see whether realloc allows account data to be shrunk?  But I guess I could still fund your reallocation through a system transfer, while you agree to it by actually calling the `Realloc` instruction

joncinque: > I don't think I ever checked to see whether realloc allows account data to be shrunk?

It does! https://github.com/solana-labs/solana/blob/28442aa9226907ceb926287906cc14135aafd18e/programs/bpf/rust/realloc/src/processor.rs#L78 has an example of that, so we have to be sure that reallocs are always larger, but I think the code already does that

CriesofCarrots: > I don't think I ever checked to see whether realloc allows account data to be shrunk?

I believe it can, yeah. I assume we would want to prevent shrinking in this `Realloc` instruction, though. (jinx, Jon)

token/program-2022/src/extension/memo_transfer/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct MemoTransfer {
+    /// Require transfers into this account to be accompanied by a memo
+    pub require_incoming_transfer_memos: PodBool,
+}
+impl Extension for MemoTransfer {
+    const TYPE: ExtensionType = ExtensionType::MemoTransfer;
+}
+
+type PodBool = u8;
```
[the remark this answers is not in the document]
CriesofCarrots: Can't believe I missed that. Thanks!

token/program-2022/src/extension/memo_transfer/instruction.rs
CriesofCarrots: https://github.com/solana-labs/solana-program-library/pull/2864

CriesofCarrots: Reworked on ^

token/program-2022/src/extension/memo_transfer/processor.rs
CriesofCarrots: This is super simple, now that Reallocate is its own thing

token/program-2022/src/extension/memo_transfer/mod.rs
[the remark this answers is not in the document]
CriesofCarrots: Updated

token/program-2022/src/extension/memo_transfer/instruction.rs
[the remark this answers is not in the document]
CriesofCarrots: I've gone with enable/disable required-memo-transfers... how does that sit with you?

token/program-2022/src/extension/memo_transfer/mod.rs
CriesofCarrots: Done

token/program-2022/src/extension/memo_transfer/instruction.rs
```diff
@@ -0,0 +1,113 @@
...
+    },
+    std::convert::TryFrom,
+};
+
+/// Default Account State extension instructions
+#[derive(Clone, Copy, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum RequiredMemoTransfersInstruction {
+    /// Require memos for transfers into this Account. Adds the MemoTransfer extension to the
+    /// Account, if it doesn't already exist. In this case,
```
joncinque, line 17: nit
```suggestion
    /// Account, if it doesn't already exist.
```

token/rust/src/token.rs
```diff
@@ -750,4 +769,41 @@ where
         )
         .await
     }
+
+    /// Require memos on transfers into this account
+    /// Reallocate first, if needed
```
joncinque, line 774: nit, correct?
```suggestion
```

joncinque (dismissed):
Looks great! Just two nittiest of comment nits

token/program-2022/src/extension/memo_transfer/instruction.rs
CriesofCarrots: 😆 thank you for catching these bad edits!

CriesofCarrots:
Woot! Excited to get this one in

PR #2824, opened 2022-01-28
token-2022: fix panics due to small buffers

CriesofCarrots:
The TLV handling in the extension module uses a lot of direct indexes and `split` methods. These can panic on index out of range when the provided account data is too small.

This PR tries to protect against these panics in as lightweight and compute-minimal way as possible.


joncinque (approved):
Looks good, thanks for fixing this up!

PR #2826, opened 2022-01-28
spl-governance: Add VoteTipping option

ckamm:
This allows two new tipping styles in addition to the previous "Strict" mode:
- Early: Where a yes majority above the yes vote threshold will
  complete the vote, even if below 50% of total possible votes.
- Disabled: Where votes never complete early.

#### Implementation 

- `VoteWeightSource` was removed because it become irrelevant with `voter-weight-addin` and spot taken for `VoteTiping` option 

@SebastianBor

### DIFF
--- a/governance/program/src/state/proposal.rs
+++ b/governance/program/src/state/proposal.rs
@@ -25,7 +25,7 @@ use crate::{
     state::{
         enums::{
             GovernanceAccountType, InstructionExecutionFlags, MintMaxVoteWeightSource,
-            ProposalState, TransactionExecutionStatus, VoteThresholdPercentage,
+            ProposalState, TransactionExecutionStatus, VoteThresholdPercentage, VoteTipping,
         },
         governance::GovernanceConfig,
         proposal_transaction::ProposalTransactionV2,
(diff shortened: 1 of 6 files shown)


governance/program/src/state/proposal.rs
```diff
@@ -473,13 +473,35 @@ impl ProposalV2 {
...
+            VoteTipping::Strict => {
+                if yes_vote_weight >= min_vote_threshold_weight
+                    && yes_vote_weight > (max_vote_weight - yes_vote_weight)
+                {
+                    yes_option.vote_result = OptionVoteResult::Succeeded;
+                    return Some(ProposalState::Succeeded);
+                }
+            }
+            VoteTipping::Early => {
+                solana_program::msg!(
```
SebastianBor, line 487: nit: do we need this msg?

governance/program/src/state/proposal.rs
```diff
@@ -473,13 +473,35 @@ impl ProposalV2 {
...
+                if yes_vote_weight >= min_vote_threshold_weight
+                    && yes_vote_weight > deny_vote_weight
+                {
+                    yes_option.vote_result = OptionVoteResult::Succeeded;
+                    return Some(ProposalState::Succeeded);
+                }
+            }
+        }
+
+        if !matches!(config.vote_tipping, VoteTipping::Disabled)
```
SebastianBor, line 502: nit: `config.vote_tipping != VoteTipping::Disabled` ?
Can you add comments what happens here please? That we never tip early on `Deny` votes and it's always `Strict` mode for `Deny`

governance/program/tests/process_cast_vote.rs
```diff
@@ -578,6 +858,77 @@ async fn test_cast_vote_with_threshold_below_50_and_vote_not_tipped() {
...
+        )
+        .await
+        .unwrap();
+
+    governance_test
+        .mint_community_tokens(&realm_cookie, 20) // total supply: 120
+        .await;
+
+    // Act: no yes tipping
+    let proposal_cookie = governance_test
```
SebastianBor, line 894: nit: `with_signed_off_proposal` is not part of  `Act`. It belongs to `Arrange`

governance/program/tests/process_cast_vote.rs
```diff
@@ -578,6 +858,77 @@ async fn test_cast_vote_with_threshold_below_50_and_vote_not_tipped() {
...
+
+    // Assert
+
+    let proposal_account = governance_test
+        .get_proposal_account(&proposal_cookie.address)
+        .await;
+    assert_eq!(ProposalState::Voting, proposal_account.state);
+
+    // Act: no deny tipping
+    let proposal_cookie = governance_test
```
SebastianBor, line 915: nit: Can we split `Yes` and `No` tipping use cases into separate tests please?

governance/program/tests/process_cast_vote.rs
```diff
@@ -520,6 +520,286 @@ async fn test_cast_vote_with_vote_tipped_to_defeated() {
...
+        .with_community_token_deposit(&realm_cookie)
+        .await
+        .unwrap();
+
+    let token_owner_record_cookie4 = governance_test
+        .with_community_token_deposit(&realm_cookie)
+        .await
+        .unwrap();
+
+    let token_owner_record_cookie5 = governance_test
```
SebastianBor, line 566: nit: Can we capture this use case with just one voter voting Yes? I think we don't need 5 and can make the test simpler

governance/program/tests/process_cast_vote.rs
```diff
@@ -520,6 +520,286 @@ async fn test_cast_vote_with_vote_tipped_to_defeated() {
...
+        .unwrap();
+
+    // 100 votes
+    let token_owner_record_cookie2 = governance_test
+        .with_community_token_deposit(&realm_cookie)
+        .await
+        .unwrap();
+
+    // 100 votes
+    let token_owner_record_cookie3 = governance_test
```
SebastianBor, line 735: nit: Can we capture this use case with just one voter voting No? I think we don't need 3 and can make the test simpler

SebastianBor (commented):
Looks very good, I only added some nits, Thank you

governance/program/src/state/proposal.rs
ckamm: nah, just debugging, will remove

governance/program/tests/process_cast_vote.rs
ckamm: I could, but I think it'd make the test worse. It currently checks that:
- 100 yes does not tip
- 100 yes, 100 no does not tip
- 200 yes, 100 no does tip
- 100 yes, 200 no does not tip
- 200 yes, 200 no does not tip
- 300 yes, 200 no does tip
It should be rewritten to be data-driven, to explicitly list out these test scenarios. But since you seem to prefer one explicit function per test case I was not sure if that would be accepted.

governance/program/tests/process_cast_vote.rs
ckamm: I like that this test confirms that 200 no votes do not tip the vote to defeated, even though it's above 50% and 40% of votes.

SebastianBor: ok

PR #2827, opened 2022-01-28
Governance: Make realm authority check optional

SebastianBor:
#### Summary

Make it possible to set realm authority to non governance accounts.
It's required for scenarios where an external program takes ownership of the realm and controls all interactions

#### Implementation 

`SetRealmAuthorityAction` was introduced to make it possible to specify the intent (`SetChecked`, `SetUnchecked`, `Remove`) of the `SetRealmAuthority` instruction


joncinque (approved):
looks good

PR #2828, opened 2022-01-28
Governance: Allow realm authority create governances

SebastianBor:
#### Summary

Make it possible for realm authority to sign transaction to crate governances
Use cases:

- External program can take ownership of a realm and it's assets 
- Stronger security requirements. A realm can opt to disallow governance creation by members by setting `min_community_tokens_to_create_governance`=`u64::MAX` and only create governances via proposals


governance/program/src/instruction.rs
```diff
@@ -110,11 +110,11 @@ pub enum GovernanceInstruction {
     ///   0. `[]` Realm account the created Governance belongs to
     ///   1. `[writable]` Account Governance account. PDA seeds: ['account-governance', realm, governed_account]
     ///   2. `[]` Account governed by this Governance
-    ///   3. `[]` Governing TokenOwnerRecord account
+    ///   3. `[]` Governing TokenOwnerRecord account (Required only if not signed by RealmAuthority)
```
joncinque, line 113: Could you remove this comment? This makes it sound like the account is optional, which it isn't since you do need to pass it in no matter.  Or you can update the language to `(Used only if not signed by RealmAuthority)`

joncinque (approved):
Looks good, just the nit on the `(Required only if not signed by RealmAuthority)` comment in the instructions

governance/program/src/instruction.rs
SebastianBor: Done

PR #2831, opened 2022-01-28
Governance: Add max_executable_options to VoteType

SebastianBor:
#### Summary

Make it possible to define how many wining options for a proposal can be executed. 

#### Implementation 
This PR only added the required fields to `VoteType` and leaves the implementation for `V3`


joncinque (commented):
Very exciting, can't wait to see this!

PR #2833, opened 2022-01-29
Governance: Multiple instructions per proposal transaction

SebastianBor:
#### Summary

This PR makes it possible to use multiple instructions per single proposal transaction. 
It's required in scenarios where multiple instructions must be executed in a single transaction.

#### Implementation 
-  The main change is in the first commit https://github.com/solana-labs/solana-program-library/pull/2833/commits/e9410a9bb50d7bf130fe82b9ae6dc4d140e04d70 . It changes instruction filed to a vector
- All the other changes are renaming `ProposalInstruction` to `ProposalTransaction`


governance/program/src/state/proposal_transaction.rs
```diff
@@ -95,62 +97,74 @@ pub struct ProposalInstructionV2 {
...
-impl ProposalInstructionV2 {
+impl ProposalTransactionV2 {
     /// Serializes account into the target buffer
     pub fn serialize<W: Write>(self, writer: &mut W) -> Result<(), ProgramError> {
-        if self.account_type == GovernanceAccountType::ProposalInstructionV2 {
+        if self.account_type == GovernanceAccountType::ProposalTransactionV2 {
             BorshSerialize::serialize(&self, writer)?
         } else if self.account_type == GovernanceAccountType::ProposalInstructionV1 {
+            if self.instructions.len() != 1 {
+                panic!("Multiple instructions are not supported by ProposalInstructionV1")
```
joncinque, line 145: This already returns a `Result<(), ProgramError>` so you're better off returning an error here.  Then again, you should probably never arrive at this point, right?

joncinque (commented):
Looks good otherwise!

governance/program/src/state/proposal_transaction.rs
SebastianBor: Correct, we don't mutate `ProposalTransaction`. If we want to change it we replace the entire account, so this check is just in case something would change in future versions

PR #2836, opened 2022-01-29
Governance: Rename AccountGovernance to Governance

SebastianBor:
#### Summary 

Rename `AccountGovernance` to `Governance` to avoid confusion caused by the Account prefix.
`Governance` account  is the core primitive of `spl-governance` and can be used as the authority for any type of asset or multiple  assets. In other words it's not restricted to authority over a single account.

In future versions `ProgramAccount`, `MintAccount` and `TokenAccount` will be deprecated and we will have only `Governance` account.


governance/program/src/instruction.rs
```diff
@@ -682,27 +683,33 @@ pub fn set_governance_delegate(
...
     // Args
     config: GovernanceConfig,
 ) -> Instruction {
-    let account_governance_address =
-        get_account_governance_address(program_id, realm, governed_account);
+    let governed_account_address = if let Some(governed_account) = governed_account {
+        *governed_account
+    } else {
+        // If the governed account is not provided then generate a unique identifier for the Governance account
+        Pubkey::new_unique()
```
joncinque, line 704: What's the use case for this? Keep in mind that `Pubkey::new_unique()` is actually deterministic, and not random, so you could easily end up with many different people creating governances on the same Pubkey.  You may actually prefer to create a random Pubkey instead, or use `Pubkey::default`

Implementation at https://github.com/solana-labs/solana/blob/17b4563a6f4f64a9d0ee321040222f9149fe12a2/sdk/program/src/pubkey.rs#L151

joncinque (commented):
Looks good otherwise!

governance/program/src/instruction.rs
SebastianBor: It just needs `Pubkey` as the seed for the PDA if the `governed_account` is irrelevant. Is there a way to generate a unique `Pubkey` here?

joncinque: You may be better off actually creating a pubkey with random bytes, like the implementation at: https://github.com/solana-labs/solana/blob/17b4563a6f4f64a9d0ee321040222f9149fe12a2/sdk/program/src/pubkey.rs#L147 but then you can't do it on-chain, so there's that

PR #2839, opened 2022-01-30
Governance: Rename tokens_to_create to weight_to_create

SebastianBor:
####

With `voter-weight` and `max-voter-weight ` addins the name `tokens_to_create` was misleading and was changed to `weight_to_create`


joncinque (commented):
Looks good

PR #2840, opened 2022-01-30
Governance: Use saturating sub to decrease voting proposal count

SebastianBor:
#### Summary

After upgrading to `v2` existing realms won't have `voting_proposal_count` and hence transition to completed state would result in underflow

#### Implementation 

Use `saturating_sub` instead of `checked_sub`


joncinque:
Hehe I was going to comment this on the other one!

PR #2841, opened 2022-01-30
Governance: Limit proposal options to 10

SebastianBor:
#### Summary 

- Limit proposal options to 10 to avoid exceeding computation limit
- Increase `Proposal` reserved space to 64 bytes. Even though the runtime would support changing account size in the future version it'll still be very difficult for Governance accounts if other programs read the accounts.

### DIFF
--- a/governance/program/src/instruction.rs
+++ b/governance/program/src/instruction.rs
@@ -232,7 +232,7 @@ pub enum GovernanceInstruction {
     InsertTransaction {
         #[allow(dead_code)]
         /// The index of the option the transaction is for
-        option_index: u16,
+        option_index: u8,
         #[allow(dead_code)]
         /// Transaction index to be inserted at.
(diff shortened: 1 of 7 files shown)


joncinque (commented):
You may be better off staying with `u16` and eventually changing how you deserialize on-chain.  Borsh is quite expensive on expensive compared to `bytemuck`-style transmutes.  With different deserialization, you can definitely support more than 10 options!

SebastianBor:
> You may be better off staying with `u16` and eventually changing how you deserialize on-chain. Borsh is quite expensive on expensive compared to `bytemuck`-style transmutes. With different deserialization, you can definitely support more than 10 options!

I decided to stay with `u8`, it feels like enough options to choose from for users anyway and they dislike voting with just `Yes`/`No` :)

PR #2842, opened 2022-01-31
Governance: add v2 account padding

SebastianBor:
#### Summary 

Add extra padding to `Realm`, `TokenOwnerRecord`, `Governance`, `SignatoryRecord`, `ProposalTransaction` and `VoteRecord`  accounts to future proof the data structures.
 Even though the runtime would support changing account size in the future version it'll still be very difficult for Governance accounts if other programs like its addins read the accounts.


joncinque (commented):
Looks good

PR #2847, opened 2022-01-31
spl-token-cli: Bump to 2.0.16

jdenquin:
In order to have recent changes on `spl-token-cli`, just a simple bump to `2.0.16` to permit an update on crates!


jdenquin:
After trying to use it by building master on my local env, I had some issues trying to creating token, accounts, etc So I think it's not ready to bump!

joncinque:
This was done in #3360

PR #2848, opened 2022-01-31
token-2022: Add freeze / thaw through StateWithExtensions

joncinque:
#### Problem

Freeze / thaw still use `Account::unpack` and `Mint::unpack`, making them incompatible with extensions.

#### Solution

Use `StateWithExtensions{,Mut}` and add a very basic program-test for it.

### DIFF
--- a/token/program-2022/tests/freeze.rs
+++ b/token/program-2022/tests/freeze.rs
@@ -1,0 +1,44 @@
+#![cfg(feature = "test-bpf")]
+
+mod program_test;
+use {
+    program_test::{TestContext, TokenContext},
+    solana_program_test::tokio,
+    solana_sdk::{signature::Signer, signer::keypair::Keypair},
+    spl_token_2022::state::AccountState,
+};
+
+#[tokio::test]
+async fn basic() {
+    let mut context = TestContext::new().await;
+    context.init_token_with_freezing_mint(vec![]).await.unwrap();
+    let TokenContext {
+        freeze_authority,
+        token,
+        alice,
+        ..
+    } = context.token_context.unwrap();
+    let freeze_authority = freeze_authority.unwrap();
+
+    let account = Keypair::new();
+    let account = token
+        .create_auxiliary_token_account(&account, &alice.pubkey())
+        .await
+        .unwrap();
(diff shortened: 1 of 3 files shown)


token/program-2022/tests/freeze.rs
```diff
@@ -0,0 +1,44 @@
...
+        freeze_authority,
+        token,
+        alice,
+        ..
+    } = context.token_context.unwrap();
+    let freeze_authority = freeze_authority.unwrap();
+
+    let account = Keypair::new();
+    let account = token
+        .create_auxiliary_token_account(&account, &alice.pubkey())
```
CriesofCarrots, line 25: Oo, can we do this with ATAs now?

CriesofCarrots (approved):
Lgtm!

token/rust/src/token.rs
```diff
@@ -538,4 +576,25 @@ where
         )
         .await
     }
+
+    /// Withdraw withheld tokens from mint
+    pub async fn withdraw_withheld_tokens_from_mint<S2: Signer>(
```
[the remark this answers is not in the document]
joncinque: Ah yes, it is, sorry about that

token/program-2022/tests/freeze.rs
joncinque: So it is possible, but until 1.0.5 is bundled by default with program-test, we'll have to add it by hand and be sure that the *.so is built. What do you think?

CriesofCarrots: Ohhh that's right, program-test bundling is still a blocker. (Though you mean 1.0.6, right?)
I think that reciprocal dependency between spl-token and spl-ata will ultimately be an issue for publishing to crates.io. So, nm. `create_auxiliary_token_account` seems fine for now.

joncinque: Right yes, 1.0.6, sorry!  I'll revert the commit.  At least we know it can work if needed

CriesofCarrots: Yep! Thanks for poc-ing that :)

PR #2849, opened 2022-01-31
token-2022: Add transfer fee withdrawal from mint

joncinque:
#### Problem

Fees need to be collected from the mint as they're harvested.

#### Solution

Add the capability.

Note: this builds on #2848, only the second commit is needed.

Also note: there are a few TODO tests in the transfer fee, and I'll add those with `withdraw_withheld_tokens_from_accounts`


CriesofCarrots (dismissed):
lgtm! r+ rebase etc

Issue #2640, opened 2021-12-15
Associated token account owner is changeable


ADBalici:
Should the owner of an associated token account be changeable by a call to setAuthority in the SPL Token Program? Otherwise why have the associated token account program in the first place?

I believe it breaks the pattern of deterministically finding a token account given a mint key and a "wallet", and it forces developers to endlessly create multiple token accounts.

Can any check be added to prevent this?

RobertoMalutan:
👍

flashjpr:
This would make so much sense to implement

ADBalici:
All of them basically allow for this. At least the major ones I know of, such as Phantom.

mvines:
I don't see where Phantom allows me to transfer my ATA to another user?  Can you show me the workflow there please, sorry for being dense

ADBalici:
Sure, I would be happy to explain the problem.

I've discovered this problem in the context of listing NFTs for sale on secondary marketplaces. For example, this is a transaction which was generated after listing an item.

https://explorer.solana.com/tx/2hngxBJ2JpwZDQvkXdF24LbwkJMhJE9gk74E8DfsHWx21AMTcV7NX6iq5TnBfg8SzziMLrktuEw7jjnhsLSDnRea

The token is _3jEHeB1kLYNiTAg1rtP9pLEunSjewxiksmXkrM7bFM5W_ (Metaplex token Frakt-2347). 
My wallet (which initiated the listing) is _AeHmkgcz9dGs2sXrjEMrAcDaktUJyVTKknG3CQ9jKRyn_ and the associated token account was _2nkqqfBHkoAinKbNEKDAi6NqGwtikpXEnQ5NF2QE1QA6_.

If you look in the transaction there is a __Token Program: Set Authority__ which sets the new authority to _GUfCR9mK6azb9vcpsxgXyj7XRPAKJd4KMHTTVvtncGgp_ (which is a PDA).

I've done this listing through Phantom and the transaction was allowed with no errors.

ADBalici:
Now I realise that my original phrase "All of them basically allow for this" is a bit misleading. Should have been "All of them basically allow for this, indirectly" :D

mvines:
Oh I see.  You approved a transaction in Phantom from a shady NTF site and they sold your ATA

mvines:
[...]

PR #2852, opened 2022-02-01
token-2022: GetAccountDataSize instruction now accepts user-provided extension types

mvines:
When the user wants to allocate a token account with optional extensions, there's no convenient method to figure the size required for those extensions.  This is in contrast to mandatory extensions, where `GetAccountDataSize` can be used to compute the size.

Extend `GetAccountDataSize` to allow the caller to pass in a list of additional `ExtensionTypes` they intend to initialize on the account.

Context: this'll be used by the ATA program to add an extension in a future PR to ATA accounts to prevent ownership transfers (cc: #2640)

### DIFF
--- a/token/program-2022/src/instruction.rs
+++ b/token/program-2022/src/instruction.rs
@@ -2,8 +2,9 @@
 
 use {
     crate::{
-        check_program_account, error::TokenError,
-        extension::transfer_fee::instruction::TransferFeeInstruction,
+        check_program_account,
+        error::TokenError,
+        extension::{transfer_fee::instruction::TransferFeeInstruction, ExtensionType},
     },
     solana_program::{
         instruction::{AccountMeta, Instruction},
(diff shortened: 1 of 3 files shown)


CriesofCarrots (approved):
👍

mvines:
`withdraw_withheld_tokens_from_mint()` fails at the HEAD of master.  It ain't this PR!

token/program-2022/src/instruction.rs
```diff
@@ -576,7 +580,13 @@ impl TokenInstruction {
                     decimals,
                 }
             }
-            21 => Self::GetAccountDataSize,
+            21 => {
+                let mut extension_types = vec![];
+                for chunk in rest.chunks(size_of::<ExtensionType>()) {
```
mvines, line 585: @CriesofCarrots - I just wanted to point out that I'm not serializing the vector length here.  On deserialization, the code just pulls off `u16`'s until there are no more. I think is is ok though

CriesofCarrots:
> `withdraw_withheld_tokens_from_mint()` fails at the HEAD of master. It ain't this PR!

Naw: https://github.com/solana-labs/solana-program-library/pull/2851

token/program-2022/src/instruction.rs
CriesofCarrots: Yeah, I noticed that. I think it is okay also. Thanks for highlighting it. Hmm, should we add a comment to make it more explicit?

mvines:
Ah, cool!

token/program-2022/src/instruction.rs
mvines: I feel it's prolly ok as is, until somebody trips over it :)

PR #2853, opened 2022-02-01
token-2022: Permit new token accounts to be overallocated

mvines:
It's ok to allocated more account data than what's needed by the required extensions, as maybe the user wants to mix in some optional extensions as well


CriesofCarrots (approved):
Good call, thx!

PR #2855, opened 2022-02-01
Governance V2 cleanup

SebastianBor:
#### Summary

- Use exhaustive account type checks
- Fix `RealmV1` deserialisation

### DIFF
--- a/governance/program/src/state/realm.rs
+++ b/governance/program/src/state/realm.rs
@@ -277,7 +277,7 @@ pub fn get_realm_data(
     let account_type: GovernanceAccountType = try_from_slice_unchecked(&realm_info.data.borrow())?;
 
     // If the account is V1 version then translate to V2
-    if account_type == GovernanceAccountType::ProposalV1 {
+    if account_type == GovernanceAccountType::RealmV1 {
         let realm_data_v1 = get_account_data::<RealmV1>(program_id, realm_info)?;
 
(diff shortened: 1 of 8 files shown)


governance/program/src/state/realm.rs
```diff
@@ -277,7 +298,7 @@ pub fn get_realm_data(
     let account_type: GovernanceAccountType = try_from_slice_unchecked(&realm_info.data.borrow())?;
 
     // If the account is V1 version then translate to V2
-    if account_type == GovernanceAccountType::ProposalV1 {
+    if account_type == GovernanceAccountType::RealmV1 {
```
joncinque, line 301: :eyes:  sorry for missing this!

joncinque (commented):
Looks good

governance/program/src/state/realm.rs
SebastianBor: Yeah, I still have to figure out how to test backward compatibility because it's all manual at the moment. 
I'm thinking about 3 options 
1) Do it from client sdk because the API supports `v1` and `v2` at the same time and we could deploy `v1` setup realm, upgrade and then test using `v2`
2) Dump `v1` accounts in binary format and use that as input for data
3) Setup all in Rust with deploying `v1`, setting up a realm and then upgrading to `v2` and testing

PR #2856, opened 2022-02-01
token-2022: fixup realloc_needed

CriesofCarrots:
Several issues with initial `realloc_needed()` api:
- Does not properly handle accounts with only the base state originally initialized
- Diff returned does not include the space needed for ExtensionType and Length
- Does not check whether the new ExtensionType is already initialized (which does not require reallocation)

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -515,7 +515,9 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
         } else {
             let mut diff = new_needed_tlv_len - self.tlv_data.len(); // arithmetic safe because of if clause
             if self.account_type.is_empty() {
-                diff = diff.saturating_add(size_of::<AccountType>());
+                diff = diff
+                    .saturating_add(size_of::<AccountType>())
+                    .saturating_add(BASE_ACCOUNT_LENGTH.saturating_sub(S::LEN));
             }
             Ok(Some(diff))
         }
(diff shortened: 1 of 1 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -505,13 +505,19 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
...
+        }
+        let new_needed_tlv_len = ExtensionType::get_total_tlv_len(&extensions);
         if self.tlv_data.len() >= new_needed_tlv_len {
             Ok(None)
         } else {
-            Ok(Some(new_needed_tlv_len - self.tlv_data.len())) // arithmetic safe because of if clause
+            let mut diff = new_needed_tlv_len - self.tlv_data.len(); // arithmetic safe because of if clause
+            if self.account_type.is_empty() {
+                diff = diff.saturating_add(size_of::<AccountType>());
+            }
```
joncinque, line 519: Is this being used on mints? If so, we'll have to also add the padding required.

token/program-2022/src/extension/mod.rs
```diff
@@ -505,13 +505,19 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
...
         new_extension: ExtensionType,
     ) -> Result<Option<usize>, ProgramError> {
-        let current_extensions = self.get_extension_types()?;
-        let needed_tlv_len = ExtensionType::get_total_tlv_len(&current_extensions);
-        let new_needed_tlv_len = needed_tlv_len.saturating_add(new_extension.get_type_len());
+        let mut extensions = self.get_extension_types()?;
+        if !extensions.contains(&new_extension) {
+            extensions.push(new_extension);
+        }
+        let new_needed_tlv_len = ExtensionType::get_total_tlv_len(&extensions);
```
joncinque, line 512: `get_total_tlv_len` probably also needs a check to make sure that the return is not equal to `Multisig::LEN - BASE_ACCOUNT_LENGTH - size_of::<AccountType>()`.  Maybe we should just move the multisig check from `get_account_len` to `get_total_tlv_len`. What do you think?

joncinque (commented):
Looks good, just the two extra points to consider

token/program-2022/src/extension/mod.rs
CriesofCarrots: Mm, true. I don't know of an immediate use-case for Mints, aside from my tests, but should support it properly. Thanks!

token/program-2022/src/extension/mod.rs
CriesofCarrots: Good call, moving the multisig check sounds spot on

joncinque (approved):
Looks great!

PR #2857, opened 2022-02-01
token-2022: Add withdraw withheld tokens from accounts

joncinque:
#### Problem

Harvesting tokens + withdrawing from the mint is great, but it's also useful to withdraw directly from the accounts to avoid taking a write lock on the mint too much.

#### Solution

Implement `withdraw_withheld_tokens_from_accounts`!

### DIFF
--- a/token/program-2022/src/extension/transfer_fee/instruction.rs
+++ b/token/program-2022/src/extension/transfer_fee/instruction.rs
@@ -6,6 +6,7 @@ use {
         program_option::COption,
         pubkey::Pubkey,
     },
+    std::convert::TryFrom,
 };
 
 /// Transfer Fee extension instructions
@@ -95,7 +96,10 @@ pub enum TransferFeeInstruction {
     ///   2. `[]` The mint's `withdraw_withheld_authority`'s multisignature owner/delegate.
     ///   3. ..3+M `[signer]` M signer accounts.
     ///   3+M+1. ..3+M+N `[writable]` The source accounts to withdraw from.
-    WithdrawWithheldTokensFromAccounts,
+    WithdrawWithheldTokensFromAccounts {
+        /// Number of multisig signers, in addition to the authority account
+        number_of_additional_signers: u8,
+    },
     /// Permissionless instruction to transfer all withheld tokens to the mint.
     ///
     /// Succeeds for frozen accounts.
@@ -161,7 +165,14 @@ impl TransferFeeInstruction {
                 (instruction, rest)
             }
             2 => (Self::WithdrawWithheldTokensFromMint, rest),
-            3 => (Self::WithdrawWithheldTokensFromAccounts, rest),
+            3 => {
+                let (&number_of_additional_signers, rest) =
+                    rest.split_first().ok_or(InvalidInstruction)?;
+                let instruction = Self::WithdrawWithheldTokensFromAccounts {
+                    number_of_additional_signers,
+                };
+                (instruction, rest)
+            }
             4 => (Self::HarvestWithheldTokensToMint, rest),
             5 => {
                 let (transfer_fee_basis_points, rest) = TokenInstruction::unpack_u16(rest)?;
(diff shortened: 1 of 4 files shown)


token/program-2022/src/extension/transfer_fee/processor.rs
```diff
@@ -187,18 +181,94 @@ fn process_harvest_withheld_tokens_to_mint(accounts: &[AccountInfo]) -> ProgramR
...
+    let mut dest_account_data = dest_account_info.data.borrow_mut();
+    let mut dest_account = StateWithExtensionsMut::<Account>::unpack(&mut dest_account_data)?;
+    if dest_account.base.mint != *mint_account_info.key {
+        return Err(TokenError::MintMismatch.into());
+    }
+    if dest_account.base.is_frozen() {
+        return Err(TokenError::AccountFrozen.into());
+    }
+    // This iterates back over any multisig signers, but since it gracefully
+    // fails on those, we can support them.
```
joncinque, line 238: Note that this is the goofiest part of the implementation.  I don't want to change the behavior of `validate_owner` for multisigs, so this keeps it simple.  The alternative to iterating through everything is to start at the first non-signer in the `account_infos` slice, but this doesn't resolve the issue mentioned earlier of a multisig whose owner is a token account *not* signing the transaction.

CriesofCarrots: Idk, could we avoid some of the iterations in `validator_owner` and goofiness here by passing the number of accounts from which to withdraw as instruction data?

joncinque: That's a great point! For some reason I was thinking that it would be gross to add a new parameter to the instruction creator, but the instruction creator can simply calculate it and tack it on.  And that way, the instruction can create subslices for each part.

joncinque: Ok, done, let me know how it looks

CriesofCarrots: That's the idea, but I guess I was thinking of using the field to specify number of withdrawal accounts so that the instruction remains extension-specific and not mixed up in the multisig handling.
It is really the same functionality either way, just feels slightly cleaner to me.

joncinque: Ah right, makes sense. Deep down, I was just avoiding a subtraction which actually makes the code safer. Done!

CriesofCarrots: Yeah, I feel you. Subtraction is the worst

token/program-2022/src/extension/transfer_fee/instruction.rs
```diff
@@ -95,7 +96,10 @@ pub enum TransferFeeInstruction {
     ///   2. `[]` The mint's `withdraw_withheld_authority`'s multisignature owner/delegate.
     ///   3. ..3+M `[signer]` M signer accounts.
     ///   3+M+1. ..3+M+N `[writable]` The source accounts to withdraw from.
-    WithdrawWithheldTokensFromAccounts,
+    WithdrawWithheldTokensFromAccounts {
+        /// Number of token accounts harvested
+        number_of_token_accounts: u8,
```
CriesofCarrots, line 101:
```suggestion
        num_token_accounts: u8,
```
nitty nit; this is just how we usually write such variables

token/program-2022/src/extension/transfer_fee/processor.rs
```diff
@@ -187,18 +181,98 @@ fn process_harvest_withheld_tokens_to_mint(accounts: &[AccountInfo]) -> ProgramR
...
+    let mint_data = mint_account_info.data.borrow();
+    let mint = StateWithExtensions::<Mint>::unpack(&mint_data)?;
+    let extension = mint.get_extension::<TransferFeeConfig>()?;
+
+    let withdraw_withheld_authority = Option::<Pubkey>::from(extension.withdraw_withheld_authority)
+        .ok_or(TokenError::NoAuthorityExists)?;
+    // This will also iterate over the token accounts for a multisig, but
+    // it will only fail if trying to withdraw from an account that is part
+    // of the multisig, but *not* signing the transaction.
+    // This is an annoyance, but doesn't introduce any security issues.
```
CriesofCarrots, line 224: Update comment?

token/program-2022/tests/transfer_fee.rs
```diff
@@ -629,7 +629,29 @@ async fn set_withdraw_withheld_authority() {
         Some(new_authority.pubkey()).try_into().unwrap(),
     );
 
-    // TODO: assert new authority can withdraw withheld fees and old cannot
+    // new authority can withdraw tokens
```
CriesofCarrots, line 632: ❤️ Thanks for filling these in!

CriesofCarrots (approved):
One tiny nit, take it or leave it.

PR #2859, opened 2022-02-02
token-2022: add StateWithExtensionsMut::unpack_after_realloc

CriesofCarrots:
If a token extension uses `realloc()` to make space for itself on a base-state-only Account or Mint, it needs to be able to unpack the state. `unpack_uninitialized()` doesn't work, because the base-state is already set. `unpack()` doesn't work because the AccountType isn't already set.

This PR adds a new method, `unpack_after_realloc()` that acts like `unpack`, but permits unset AccountType ... and indeed initializes the AccountType under the hood.

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -326,35 +326,17 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     ///
     /// Fails if the base state is not initialized.
     pub fn unpack(input: &'data mut [u8]) -> Result<Self, ProgramError> {
-        check_min_len_and_not_multisig(input, S::LEN)?;
-        let (base_data, rest) = input.split_at_mut(S::LEN);
-        let base = S::unpack(base_data)?;
-        if let Some((account_type_index, tlv_start_index)) = type_and_tlv_indices::<S>(rest)? {
-            // type_and_tlv_indices() checks that returned indexes are within range
-            let account_type = AccountType::try_from(rest[account_type_index])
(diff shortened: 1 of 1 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -350,6 +350,39 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
...
+        let (base_data, rest) = input.split_at_mut(S::LEN);
+        let base = S::unpack(base_data)?;
+        if let Some((account_type_index, tlv_start_index)) = type_and_tlv_indices::<S>(rest)? {
+            // type_and_tlv_indices() checks that returned indexes are within range
+            let mut account_type = AccountType::try_from(rest[account_type_index])
+                .map_err(|_| ProgramError::InvalidAccountData)?;
+            if account_type == AccountType::Uninitialized {
+                rest[account_type_index] = S::ACCOUNT_TYPE.into();
+                account_type = S::ACCOUNT_TYPE;
+            }
```
joncinque, line 368: nit: we could probably create a common private function like `unpack_checked` or `unpacked_internal` which also takes in a flag to do the account type init, since it seems to be the only difference between `unpack` and `unpack_after_realloc`

joncinque (approved):
just a refactoring nit, but otherwise looks good!

token/program-2022/src/extension/mod.rs
CriesofCarrots: Oh, totally! I actually meant to go back and dedupe this way, but then forgot 😞
Thanks!

PR #2860, opened 2022-02-02
token-swap: Changed Wrap SwapCurve's calculator field into Arc instead of Box

Anurag5086:
Solves issue #2718

### DIFF
--- a/token-swap/program/fuzz/src/instructions.rs
+++ b/token-swap/program/fuzz/src/instructions.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use {
     arbitrary::Arbitrary,
     honggfuzz::fuzz,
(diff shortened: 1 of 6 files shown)


joncinque:
Thanks for the contribution!  It looks like there are some CI issues due to the change.

Be sure to run the following on the program at `token-swap/program`:

* `cargo fmt`: formats the code specially for Rust (you can optionally specify the `--all` flag to do the whole repo)
* `cargo +nightly-2021-12-03 clippy -Zunstable-options --all-targets`: linting. currently CI uses nightly-2021-12-03, but that may change in the future
* `cargo test-bpf`: runs all the tests for the program.  token-swap actually works fine with just `cargo test` since it isn't doing any full environment tests, but all of the other programs need it.

For this, you'll probably just need to replace some `Box::new` calls with `Arc::new`

joncinque:
Ah, it looks like the `token-swap/program/fuzz` needs to be updated too.  Once that's done, this should be good to go

joncinque:
Nice job, thanks for the contribution!

Anurag5086:
Thanks for the merge! :)

gitmalong:
```rust
`(dyn spl_token_swap::curve::calculator::CurveCalculator + 'static)` cannot be shared between threads safely
```

Isn't that `Arc` useless when the struct cannot be shared between threads because it is not `sync`?

joncinque:
PRs are still welcome!

PR #2864, opened 2022-02-03
token-2022: Add Reallocate instruction

CriesofCarrots:
Some extensions are opt-in for token account owners at any time after the account has been initialized. In these cases, the owner needs the ability to reallocate their token account data to be large enough to hold the new extension. Previous approaches attempted to do this in the initialization logic of the specific extension, but borrows on the token account were messy, and it was inefficient to include additional reallocation-specific accounts (payer and system-program) on every initialization.

This PR adds a separate Reallocate instruction to handle the data resize and rent-exempt-reserve funding for a provided set of extensions.

First and last commits can be ignored for the purposes of discussion

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -33,6 +33,8 @@ pub mod default_account_state;
 pub mod immutable_owner;
 /// Mint Close Authority extension
 pub mod mint_close_authority;
+/// Utility to reallocate token accounts
+pub mod reallocate;
 /// Transfer Fee extension
 pub mod transfer_fee;
 
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -328,28 +328,13 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     ///
     /// Fails if the base state is not initialized.
     pub fn unpack(input: &'data mut [u8]) -> Result<Self, ProgramError> {
-        Self::_unpack(input, false)
-    }
-    /// Unpack base state, leaving the extension data as a mutable slice
-    /// Checks the account_type, and initializes it if Uninitialized
-    ///
-    /// Fails if the base state is not initialized.
-    pub fn unpack_after_realloc(input: &'data mut [u8]) -> Result<Self, ProgramError> {
-        Self::_unpack(input, true)
-    }
-
-    fn _unpack(input: &'data mut [u8], init_account_type: bool) -> Result<Self, ProgramError> {
         check_min_len_and_not_multisig(input, S::LEN)?;
         let (base_data, rest) = input.split_at_mut(S::LEN);
         let base = S::unpack(base_data)?;
         if let Some((account_type_index, tlv_start_index)) = type_and_tlv_indices::<S>(rest)? {
             // type_and_tlv_indices() checks that returned indexes are within range
-            let mut account_type = AccountType::try_from(rest[account_type_index])
+            let account_type = AccountType::try_from(rest[account_type_index])
                 .map_err(|_| ProgramError::InvalidAccountData)?;
-            if init_account_type && account_type == AccountType::Uninitialized {
-                rest[account_type_index] = S::ACCOUNT_TYPE.into();
-                account_type = S::ACCOUNT_TYPE;
-            }
             check_account_type::<S>(account_type)?;
             let (account_type, tlv_data) = rest.split_at_mut(tlv_start_index);
             Ok(Self {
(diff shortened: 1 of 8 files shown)


token/program-2022/src/extension/reallocate.rs
```diff
@@ -0,0 +1,87 @@
...
+        &[
+            payer_info.clone(),
+            token_account_info.clone(),
+            system_program_info.clone(),
+        ],
+    )?;
+
+    // unpack to set account_type, if needed
+    let mut token_account = token_account_info.data.borrow_mut();
+    StateWithExtensionsMut::<Account>::unpack_after_realloc(&mut token_account)?;
```
CriesofCarrots, line 84: This is the only part that feels a bit icky to me. Thoughts?

token/program-2022/src/extension/reallocate.rs
```diff
@@ -0,0 +1,87 @@
...
+    let token_account_info = next_account_info(account_info_iter)?;
+    let payer_info = next_account_info(account_info_iter)?;
+    let system_program_info = next_account_info(account_info_iter)?;
+    let authority_info = next_account_info(account_info_iter)?;
+    let authority_info_data_len = authority_info.data_len();
+
+    // check that account is the right type and validate owner
+    let mut current_extension_types = {
+        let token_account = token_account_info.data.borrow();
+        let account = StateWithExtensions::<Account>::unpack(&token_account)?;
```
CriesofCarrots, line 35: This only supports Account reallocation atm. Do we need to support Mint reallocation for anything, eg ConfidentialTransfers?

token/program-2022/src/extension/reallocate.rs
```diff
@@ -0,0 +1,87 @@
...
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+    let token_account_info = next_account_info(account_info_iter)?;
+    let payer_info = next_account_info(account_info_iter)?;
+    let system_program_info = next_account_info(account_info_iter)?;
+    let authority_info = next_account_info(account_info_iter)?;
+    let authority_info_data_len = authority_info.data_len();
+
+    // check that account is the right type and validate owner
+    let mut current_extension_types = {
```
joncinque, line 33: This is such a slick way to get the `drop` in

token/program-2022/src/extension/reallocate.rs
joncinque: Is `unpack_after_realloc` used at all anymore? Seems like we just need an `init_account_type` or something

joncinque (approved):
I really like this!  It looks good to me, but I'll let Michael give the final :+1: since Confidentials need this most. Just the one nit on that `unpack_after_realloc`

token/program-2022/src/extension/reallocate.rs
CriesofCarrots: > Seems like we just need an `init_account_type` or something

Yeah, exactly... maybe I can whip up something simple outside of the `StateWith...` impls

token/program-2022/src/extension/reallocate.rs
```diff
@@ -0,0 +1,87 @@
...
+        "account needs realloc, +{:?} bytes",
+        needed_account_len - token_account_info.data_len()
+    );
+    token_account_info.realloc(needed_account_len, false)?;
+
+    // if additional lamports needed to remain rent-exempt, transfer them
+    let rent = Rent::get()?;
+    let new_minimum_balance = rent.minimum_balance(needed_account_len);
+    let lamports_diff = new_minimum_balance.saturating_sub(token_account_info.lamports());
+    invoke(
```
mvines, line 73: How about skipping the `invoke` if `lamports_diff == 0`?  This would permit the caller to not provided a signed payer if the token account can already afford the rent increase

token/program-2022/src/extension/reallocate.rs
mvines: I don't have a mint reallocation use case in mind.  Keeping the mint configuration static after creation has a strong appeal to me right now, not eager to open a new rugging vector

token/program-2022/src/extension/reallocate.rs
CriesofCarrots: Sure, that should work. How would you recommend handling the AccountInfos for that case? ie. do we still require something passed as the payer, even though it won't be used? (optional AccountInfos are tricky with variable-length multisig signers)

mvines: I think it's just an addition to the instruction documentation.  Payer is ignored if the account can already afford the rent.  But to avoid some future client footgun, we should still require the payer be a signer though. 

Hmm, actually that's probably reason enough to leave it as is since we currently leverage the system program here to ensure the payer is a signed system account.  I retract my original suggestion.

token/program-2022/src/extension/reallocate.rs
CriesofCarrots: Okay, see last commit @joncinque ; wdyt?

mvines (approved):

token/program-2022/src/extension/mod.rs
```diff
@@ -511,31 +498,24 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
...
-            let mut diff = new_needed_tlv_len - self.tlv_data.len(); // arithmetic safe because of if clause
-            if self.account_type.is_empty() {
-                diff = diff
-                    .saturating_add(size_of::<AccountType>())
-                    .saturating_add(BASE_ACCOUNT_LENGTH.saturating_sub(S::LEN));
-            }
-            Ok(Some(diff))
+/// If AccountType is uninitialized, set it to the BaseState's ACCOUNT_TYPE;
+/// if AccountType is already set, check is set correctly for BaseState
+pub fn set_account_type<S: BaseState>(input: &mut [u8]) -> Result<(), ProgramError> {
```
joncinque, line 505: Since we only need to consider this for accounts at the moment, we don't need to worry about this case, but I want to get it out there.

In a vacuum, if you pass in a buffer that currently contains just a base mint and call this with `S = Account`, it will succeed, correct? `type_and_tlv_indices` only checks for the opposite case of passing in a buffer containing a base `Account` but call it with `S = Mint`.  I wonder if just to be totally totally safe, this should check the `is_initialized` byte to make sure it's actually an `Account` if `S = Account`.  Doesn't make it too generic though...

Probably doesn't need to get done here, and probably won't become an issue, but this exact case has been in the back of my mind for awhile now.

joncinque (approved):
Looks good, just the one bit on `set_account_type` that we may want to consider if we ever use `set_account_type` before doing a check earlier.

token/program-2022/src/extension/mod.rs
CriesofCarrots: Mm, yeah good point. I originally had `let base = S::unpack(base_data)?;` in there, which would protected against that, but deleted for expense reasons. I'm going to merge this as-is, but will file an issue for us to fix in a follow-up.

PR #2869, opened 2022-02-04
token-2022: Support extensions in burn

joncinque:
#### Problem

`burn` doesn't work with extensions

#### Solution

Support it, and add some tests


CriesofCarrots (approved):
lgtm!

PR #2870, opened 2022-02-04
token-2022: Add approve / revoke support with extensions

joncinque:
Builds directly on #2869

#### Problem

`approve` / `revoke` don't work with extensions

#### Solution

Add the support, and some tests


CriesofCarrots (approved):
Another straightforward one :)

PR #2871, opened 2022-02-04
token-2022: Fail to close account with withheld fees

joncinque:
#### Problem

It's possible to close a token account that still has withheld fees.

#### Solution

Prevent closing an account with withheld fees.

### DIFF
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -841,6 +841,11 @@ impl Processor {
             {
                 confidential_transfer_state.closable()?
             }
+
+            if let Ok(transfer_fee_state) = source_account.get_extension_mut::<TransferFeeAmount>()
+            {
+                transfer_fee_state.closable()?
+            }
         } else if let Ok(mut mint) =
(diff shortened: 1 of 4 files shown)


token/program-2022/src/processor.rs
```diff
@@ -841,6 +841,11 @@ impl Processor {
             {
                 confidential_transfer_state.closable()?
             }
+
+            if let Ok(transfer_fee_state) = source_account.get_extension_mut::<TransferFeeAmount>()
```
CriesofCarrots, line 845: Mebbe use immutable `get_extension` when it lands? https://github.com/solana-labs/solana-program-library/pull/2822/files#diff-1fd54e31043386c786ab6a3edf57cd653cf69082500a805e9dd50f550d2f260bR443

CriesofCarrots (approved):

token/program-2022/src/processor.rs
joncinque: Heck yes, I'll add a TODO for that, thanks!

PR #2872, opened 2022-02-04
token-2022: Add delegate self-revoke

### LINKED ISSUE #2747: Token account delegatee can't revoke itself
Please consider updating token program to allow for token account delegatees to revoke themselves.
It is useful when approval and revoke instructions are performed separately and no user signature is available. 
An example of such a use case would be deposits to an on-chain fund that requires manual approval from the fund manager. If the deposit is rejected or only partially approved, then the user will have a delegate indefinitely unless they explicitly revoke it. This functionality can also be used for conditional withdrawals/distributions.

Another way to achieve similar functionality is to move the user's tokens to custody and later return them if needed. Still, if operations potentially spread in time (or the second part never happens), it is better for users to keep a hold on their tokens. Approve/revoke could be seen as a more flexible and convenient option in this case.

joncinque:
#### Problem

Delegates cannot revoke themselves.  Currently, a delegate can only reset the delegate by transferring out the entire delegated amount, which is not great UX.

#### Solution

Let delegates also revoke.

Fixes #2747

### DIFF
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -471,21 +471,32 @@ impl Processor {
     pub fn process_revoke(program_id: &Pubkey, accounts: &[AccountInfo]) -> ProgramResult {
         let account_info_iter = &mut accounts.iter();
         let source_account_info = next_account_info(account_info_iter)?;
-        let owner_info = next_account_info(account_info_iter)?;
-        let owner_info_data_len = owner_info.data_len();
+        let authority_info = next_account_info(account_info_iter)?;
+        let authority_info_data_len = authority_info.data_len();
 
         let mut source_account = Account::unpack(&source_account_info.data.borrow())?;
         if source_account.is_frozen() {
             return Err(TokenError::AccountFrozen.into());
         }
 
-        Self::validate_owner(
-            program_id,
-            &source_account.owner,
-            owner_info,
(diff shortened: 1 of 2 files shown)


token/program-2022/src/processor.rs
```diff
@@ -471,21 +471,32 @@ impl Processor {
...
         }
 
-        Self::validate_owner(
-            program_id,
-            &source_account.owner,
-            owner_info,
-            owner_info_data_len,
-            account_info_iter.as_slice(),
-        )?;
+        match source_account.delegate {
```
mvines, line 482: Does the borrow checker 🤯 on this form?
```
Self::validate_owner(
    program_id,
    match source_account.delegate {
      COption::Some(ref delegate) if cmp_pubkeys(authority_info.key, delegate) => delegate,
      _ => &source_account.owner,
    },
    authority_info,
    authority_info_data_len,
    account_info_iter.as_slice(),
)?;
```

In my brain this makes it more clear how the check differs across the two code paths.  But I'm just nitting, rest looks great here

joncinque: It's totally happy with it, nice suggestion!

mvines (dismissed):

CriesofCarrots (approved):
That's nice and simple

Issue #2865, opened 2022-02-03
Stake-pool py-test job seems flaky


CriesofCarrots:
Sometimes py-test job fails on first attempt and then succeeds on rerun, on PRs seemingly unrelated to stake-pools.
eg: https://github.com/solana-labs/solana-program-library/actions/runs/1781347781/attempts/1

joncinque:
Sorry about that, I had a hard time getting the timing to work out on the `solana-test-validator` running in the background, required for these tests. If the epoch rolls over during a test, it'll fail with a "stake pool must be updated" error.  I added in some waits in the tests so they start at the beginning of an epoch, but it can still fail randomly.  I can bump up the number of slots and add more waits to stabilize these.

PR #2874, opened 2022-02-04
token-2022: De-flake test

joncinque:
#### Problem

Doing the same transaction twice can make program-test hang, and `harvest_withheld_tokens` does that.

#### Solution

Don't do that.


joncinque:
Gonna merge on red, don't want to be blocked on the other flaky test #2865

PR #2876, opened 2022-02-05
token-2022: complete memo extension functionality

CriesofCarrots:
Once v1.9.6 is released and incorporated into SPL, this will wrap-up the memo extension, using the `get_processed_sibling_instruction` syscall to flesh out the Transfer `if memo_required(&dest_account)` case.

@joncinque , wdyt about what I've done in the rust client here, adding Option<memo> to all the transfer methods? I liked that better than adding `_with_memo` versions of all of them, but open to other approaches. Also, probably it should take the memo as a String instead of bytes...?

### DIFF
--- a/token/program-2022-test/tests/memo_transfer.rs
+++ b/token/program-2022-test/tests/memo_transfer.rs
@@ -70,42 +70,49 @@ async fn test_memo_transfers(
     let bob_state = token.get_account_info(&bob_account).await.unwrap();
     assert_eq!(bob_state.base.amount, 0);
 
-    // attempt to transfer from alice to bob with misplaced memo
-    let mut ctx = context.lock().await;
-    #[allow(deprecated)]
-    let instructions = vec![
-        spl_memo::build_memo(&[240, 159, 166, 150], &[]),
-        system_instruction::transfer(&ctx.payer.pubkey(), &alice.pubkey(), 42),
-        spl_token_2022::instruction::transfer(
-            &spl_token_2022::id(),
-            &alice_account,
-            &bob_account,
-            &alice.pubkey(),
-            &[],
-            10,
-        )
-        .unwrap(),
-    ];
-    let tx = Transaction::new_signed_with_payer(
-        &instructions,
-        Some(&ctx.payer.pubkey()),
-        &[&ctx.payer, &alice],
-        ctx.last_blockhash,
-    );
-    let err: TransactionError = ctx
-        .banks_client
-        .process_transaction(tx)
-        .await
-        .unwrap_err()
-        .unwrap()
-        .into();
-    drop(ctx);
-    assert_eq!(
-        err,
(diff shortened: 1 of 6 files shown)


joncinque:
This seems like the right approach, otherwise we'll have a ridiculous number of variants.  Another option could be to include a `with_memo` function on the token client, which sets in a memo to be consumed and tacked in front during the next call to `process_ixs`, ie:

```
fn with_memo<T: AsRef<str>>(mut &self, memo: Option<T>) {
    self.memo = memo;
}
fn process_ixs(...) {
    if let Some(memo) = self.memo.take() {
        instructions.insert(0, memo);
    }
}
```

This could cause footguns, but also makes it composable with anything else.  I was always a big fan of `with_memo` from the CLI :smile: what do you think?

CriesofCarrots:
> Another option could be to include a `with_memo` function on the token client, which sets in a memo to be consumed and tacked in front during the next call to `process_ixs`

#2901 and rebased. Just waiting on v1.9.6 now

token/program-2022/Cargo.toml
```diff
@@ -20,6 +20,7 @@ num-traits = "0.2"
 num_enum = "0.5.4"
 solana-program = "1.9.9"
 solana-zk-token-sdk = "0.6.0"
+spl-memo = { path = "../../memo/program", features = [ "no-entrypoint" ] }
```
joncinque, line 23: Will this need a version number to publish to crates.io?

token/program-2022-test/tests/memo_transfer.rs
```diff
@@ -38,8 +54,62 @@ async fn test_memo_transfers(
...
+        .into();
+    drop(ctx);
+    assert_eq!(
+        err,
+        TransactionError::InstructionError(2, InstructionError::Custom(TokenError::NoMemo as u32))
+    );
+    let bob_state = token.get_account_info(&bob_account).await.unwrap();
+    assert_eq!(bob_state.base.amount, 0);
+
+    // transfer with memo
```
joncinque, line 110: Feel free to ignore this one, since I don't know too much about the evolution of memo, but is it worth also adding a test that explicitly uses memo v1?

joncinque (dismissed):
Just a couple of nits that you can take or leave.  Looks great!

token/program-2022/Cargo.toml
CriesofCarrots: Probably! Thanks for catching, not sure why/how I missed that.

token/program-2022-test/tests/memo_transfer.rs
CriesofCarrots: Definitely wouldn't hurt! Memo v1 still has a fair bit of use, and is actually a bit cheaper compute-wise, since it doesn't log the memo.

joncinque (approved):
Looks great, thanks for adding those extra bits!

PR #2884, opened 2022-02-07
token-2022: Force refresh blockhash on no-op transaction

joncinque:
#### Problem

The realloc test can be flaky due to the no-op repeated transaction.

#### Solution

Force-refresh the blockhash in the test.  The implementation is stolen from `program-test`'s extension to `BanksClient` of `get_new_latest_blockhash`.


CriesofCarrots (approved):
😞 Thanks for cleaning up after me! lgtm

joncinque:
Anytime :smile:  thanks for the quick review!

Issue #2875, opened 2022-02-04
token-2022: Check initialized base-account in `set_account_type`


CriesofCarrots:
#### Problem

#2864 added a helper function `set_account_type()` that is missing one particular type of safety rails.
While `type_and_tlv_indices()` (called under the hood) protects against calling `set_account_type::<Mint>()` against a base account already initialized as an Account, `set_account_type::<Account>()` will happily initialize a base Mint as an Account.

#### Proposed Solution

Check the is_initialized byte in the base Account data to ensure an Account isn't already initialized as a Mint
(`let base = S::unpack(&input)?;` would also do this, but we don't need to waste cycles deserializing the complete BaseState)

Tvenus:
I am an Ai and blockchain(Solana,Solidity, Smart contract, Economics...) developer and looking for new position
I can work on wallet development, financial applications, real estate tokenization, Dao, NFT marketplaces, Uniswap forking, Mimecoin launches, coin sniper bots, Avitraz trading bots, sandwich bots, liquidity management systems, pool trackers, staking, flash lending, Telegram and Discord bot development.
I can do this on EVM compatible chains and Solana.

Thank you

PR #2885, opened 2022-02-08
token-2022: fixup set_account_type to confirm AccountType

CriesofCarrots:
#### Problem
The extension helper `set_account_type::<Account>()` can set the AccountType incorrectly if the base is initialized as a Mint. See #2875

#### Solution
Check Account initialization with a quick memcmp of one byte when S::ACCOUNT_TYPE == AccountType::Account

@joncinque , I considered adding a method to `trait BaseState`, something like:
```
fn is_initialized_raw(input: &[u8]) -> Result<bool, ProgramError>;
```
that would perform the contents of is_initialized_account for Account, and similar for Mint, even though the Mint check is not strictly necessary. Think I ought to?

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -213,6 +213,16 @@ fn type_and_tlv_indices<S: BaseState>(
     }
 }
 
+/// Checks a base buffer to verify if it is an Account without having to completely deserialize it
+fn is_initialized_account(input: &[u8]) -> Result<bool, ProgramError> {
+    const ACCOUNT_INITIALIZED_INDEX: usize = 108; // See state.rs#L99
+
+    if input.len() != BASE_ACCOUNT_LENGTH {
+        return Err(ProgramError::InvalidAccountData);
+    }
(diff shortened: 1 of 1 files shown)


joncinque:
> I considered adding a method to trait BaseState, that would perform the contents of is_initialized_account for Account, and similar for Mint, even though the Mint check is not strictly necessary. Think I ought to?

Hm that's a really tough one.  Since Mint's `is_initialized` byte is at `[46]`, which is right in the middle of the `owner` of `Account`, maybe we're better keeping this as an `Account`-specific check.  Unless the check for `Mint` *both* checks that `is_initialized` = 1, and that `[106]` = 0.  But that check is already performed earlier.  And also, since the new check only works for slices that are `BASE_ACCOUNT_LENGTH`, it's only meant for Accounts anyway.

All that to say, this looks great! Thanks for looking into it

joncinque (approved):

CriesofCarrots:
Yeah, the mint check would be more extensive than just the one byte. Okay, we can readdress later if we find we need the broader implementation.

PR #2888, opened 2022-02-08
token-22: Update program id to `Tokenz...`

mvines:


mvines:
With this landed I think I want to ship a 0.1.0 version of the spl-token-2022 crate so that we can start plugging in token 22 into the monorepo. For example, on top of https://github.com/solana-labs/solana/pull/23003

Anything terrifying about that idea @joncinque/@CriesofCarrots?

CriesofCarrots (approved):
> Anything terrifying about that idea?

No, seems okay to me!

joncinque:
Go for it!

PR #2892, opened 2022-02-09
Move program-2022 tests to new crate

CriesofCarrots:
#### Problem
A circular dependency with spl-token-client prevents publishing spl-token-2022 to crates.io

#### Solution
- Move tests that need spl-token-client to their own crate.
- Remove various features no longer needed.

Open to bikeshedding on what the test crate should be called; I just followed a precedent from the monorepo


associated-token-account/program/Cargo.toml
```diff
@@ -10,15 +10,11 @@ edition = "2018"
 [features]
 no-entrypoint = []
 test-bpf = []
-# When spl-token depends on spl-token-client, which depends on this with no-entrypoint,
-# all tests fail! Normally, we want no-entrypoint, except when testing spl-token
-no-token-entrypoint = [ "spl-token/no-entrypoint" ]
-default = [ "no-token-entrypoint" ]
```
joncinque, line 16: :heart:

token/rust/Cargo.toml
```diff
@@ -7,18 +7,13 @@ name = "spl-token-client"
 repository = "https://github.com/solana-labs/solana-program-library"
 version = "0.0.1"
 
-# When spl-token-2022 depends on this with no-entrypoint, all tests fail!
-# Normally, we want no-entrypoint, except when testing spl-token-2022
-[features]
-default = [ "spl-token-2022/no-entrypoint", "spl-associated-token-account/no-token-entrypoint" ]
-
```
joncinque, line 14: :heart:

joncinque (approved):
Looks great! Makes me a little embarrassed that I didn't do this from the start

mvines:
> > Makes me a little embarrassed that I didn't do this from the start
> 
> Ditto, wasn't thinking ahead to publishing.

I briefly looked into adapting https://github.com/solana-labs/solana/blob/master/ci/order-crates-for-publishing.py#L4 for the SPL repo, which would have flagged the circular dep right from the start.  The non-uniform crate names in the SPL repo gave me a bit of headache and I gave up for now.  But if this happens again somewhere in SPL then it's prolly worth bringing in that CI extra check

mvines (approved):
thanks for doing this

PR #2899, opened 2022-02-11
Release Token TS client bindings

jordansexton:
This PR makes some minor fixes, moves the existing `js` dir to `js-old`, and moves the `ts` dir to `js`.

We should integrate the necessary changes from the `js-old` directory and delete it before merging.


joncinque:
All right, that was very tedious, but it's done! :crossed_fingers: on CI now

joncinque:
Are there any other "necessary changes" to port over that you know of? If not, I can nuke `js-old`

joncinque (approved):
`js-old` has been blasted as well, this is good to go on my side.

jordansexton:
> Incidentally, is there a reason this was built in a branch on the solana-labs repo, instead of from a fork?

Ah, was just my default since it's easier to manage. I can use a fork in the future if that's preferred?

CriesofCarrots:
> I can use a fork in the future if that's preferred?

Yes, please :) Easier maintenance of the labs repos that way.

joncinque:
Anything holding this up?

jordansexton:
I think we are good to go! I'll release the docs on Github Pages, just waiting to hear back from someone with access.

PR #2900, opened 2022-02-11
stake_pool: fix wrong available_for_withdrawal_wo_fee calculation

AlexanderRay:
`````rust
        if lamports <= min_balance {
            continue;
        }

        let available_for_withdrawal_wo_fee =
            stake_pool.calc_pool_tokens_for_deposit(lamports).unwrap();
`````

available_for_withdrawal_wo_fee dont consider min_balance

@joncinque

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -1257,7 +1257,9 @@ fn prepare_withdraw_accounts(
 
             (
                 transient_stake_account_address,
-                validator.transient_stake_lamports - min_balance,
+                validator
+                    .transient_stake_lamports
+                    .saturating_sub(min_balance),
                 Some(validator.vote_account_address),
             )
         },
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -1257,7 +1257,7 @@ fn prepare_withdraw_accounts(
 
             (
                 transient_stake_account_address,
-                validator.transient_stake_lamports,
+                validator.transient_stake_lamports - min_balance,
                 Some(validator.vote_account_address),
             )
         },
@@ -1265,7 +1265,13 @@ fn prepare_withdraw_accounts(
 
     let reserve_stake = rpc_client.get_account(&stake_pool.reserve_stake)?;
 
-    accounts.push((stake_pool.reserve_stake, reserve_stake.lamports, None));
+    accounts.push((
+        stake_pool.reserve_stake,
+        reserve_stake.lamports
+            - rpc_client.get_minimum_balance_for_rent_exemption(STAKE_STATE_LEN)?
+            - 1,
+        None,
+    ));
 
     // Prepare the list of accounts to withdraw from
     let mut withdraw_from: Vec<WithdrawAccount> = vec![];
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: stake-pool/cli/src/main.rs [hunk 1]
```diff
@@ -1283,8 +1283,9 @@ fn prepare_withdraw_accounts(
             continue;
         }
 
-        let available_for_withdrawal_wo_fee =
-            stake_pool.calc_pool_tokens_for_deposit(lamports).unwrap();
+        let available_for_withdrawal_wo_fee = stake_pool
+            .calc_pool_tokens_for_deposit(lamports - min_balance)
+            .unwrap();
```
joncinque, line 1288: Now that I look through the code, this should only be an issue with transient stake accounts and the reserve. `transient_stake_lamports` includes the rent exemption + minimum, but `active_stake_lamports` doesn't.  Can you instead change line 1260 to read `validator.transient_stake_lamports - min_balance`? And then line 1268 should be ` reserve_stake.lamports - rpc_client
.get_minimum_balance_for_rent_exemption(STAKE_STATE_LEN)? - 1`

But another bit first -- has this been observed for the validator stake accounts too?  If so, we have a potentially larger bug to look at.

AlexanderRay: is there any difference between active stake accounts and transient stake accounts? both of them have  rent_exemption and MINIMUM_ACTIVE_STAKE ?
@CMEONE had that issue - https://github.com/solana-labs/solana-program-library/pull/2806#issuecomment-1033323474 , not sure if that were transient or active stake accounts

joncinque: > is there any difference between active stake accounts and transient stake accounts?

Yes, the lamports reported are *not* the same. `transient_stake_lamports` includes the rent exemption + minimum, but `active_stake_lamports` doesn't.  So for example, if the stake pools says `active_stake_lamports` == 10 SOL, the validator stake account actually has 10.0032.... SOL. But if it says `transient_stake_lamports` == 10 SOL, the transient stake account has exactly 10 SOL.

> had that issue - https://github.com/solana-labs/solana-program-library/pull/2806#issuecomment-1033323474 , not sure if that were transient or active stake accounts

I have a feeling they were trying to withdraw from a transient account.

stake-pool/cli/src/main.rs
```diff
@@ -1257,15 +1257,21 @@ fn prepare_withdraw_accounts(
 
             (
                 transient_stake_account_address,
-                validator.transient_stake_lamports,
```
joncinque, line 1260: We'll need to do `validator.transient_stake_lamports.saturating_sub(min_balance)`, otherwise 0 will give *a lot* of SOL :laughing:

stake-pool/cli/src/main.rs
CMEONE: The stake accounts that I tried to withdraw were transient stake accounts. I moved funds from reserve into validator stake accounts and tried withdrawing stake accounts (transient because they were still activating) from the pool within the same epoch.

joncinque:
Looks good, thanks so much!

PR #2901, opened 2022-02-11
Add rust token client support for memo

CriesofCarrots:
#### Problem
There is no ergonomic way to add a memo to a token transaction in the rust token client

#### Solution
Add `with_memo()` method to stash a memo on the Token object, which will be included with the next Transaction built by `process_ixs`

### DIFF
--- a/token/rust/src/token.rs
+++ b/token/rust/src/token.rs
@@ -22,7 +22,7 @@ use spl_token_2022::{
 };
 use std::{
     fmt, io,
-    sync::Arc,
+    sync::{Arc, RwLock},
     time::{Duration, Instant},
 };
 use thiserror::Error;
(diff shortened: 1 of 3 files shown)


token/rust/src/token.rs
```diff
@@ -299,14 +315,14 @@ where
 
     /// Create and initialize the associated account.
     pub async fn create_associated_token_account(&self, owner: &Pubkey) -> TokenResult<Pubkey> {
-        self.process_ixs(
+        self.process_ixs::<[&dyn Signer; 0]>(
```
joncinque, line 318: Very slick, I had no idea how to resolve this one

joncinque (approved):
Looks great, thanks!

token/rust/src/token.rs
CriesofCarrots: It's a little funny and mind-bending to have to define the type of emptiness 😆

PR #2903, opened 2022-02-11
associated-token-account: Fail create idempotent for non-ata

joncinque:
#### Problem

The `CreateIdempotent` instruction gives a false positive on valid non-associated token accounts.

#### Solution

Move the address derivation check higher to be covered by `CreateIdempotent`.

Thanks @ryoqun for noticing this!


mvines:
> Thanks @ryoqun for noticing this!

@ryoqun - 💪🏼!

PR #2904, opened 2022-02-11
token: Support both token programs in associated-token-account

joncinque:
#### Problem

We want to have two token programs, but associated-token-account currently only supports token-22

#### Solution

A few bits needed:

* token: add `GetAccountDataSize` to return `Account::LEN` and `InitializeImmutableOwner`, which no-ops
* token-22: shuffle around the instructions for token, and remove some program id checks in instruction creators
* ata: use `_with_program_id` variant everywhere


associated-token-account/program/Cargo.toml
```diff
@@ -21,6 +21,7 @@ spl-token = { version = "0.2", path = "../../token/program-2022", package = "spl
 thiserror = "1.0"
 
 [dev-dependencies]
+spl-token-legacy = { version = "3.3", path = "../../token/program", package = "spl-token", features = ["no-entrypoint"] }
```
mvines, line 24: I vote for calling spl-token spl-token, and spl-token-2022 spl-token-2022 myself.  Which would mean this line changes to `spl-token` and the import of token 2022 higher up changes to `spl-token-2022` (as well as in the ATA code)

token/program/src/processor.rs
```diff
@@ -757,6 +758,27 @@ impl Processor {
...
+
+    /// Processes an [InitializeImmutableOwner](enum.TokenInstruction.html) instruction
+    pub fn process_initialize_immutable_owner(accounts: &[AccountInfo]) -> ProgramResult {
+        let account_info_iter = &mut accounts.iter();
+        let token_account_info = next_account_info(account_info_iter)?;
+        let account = Account::unpack_unchecked(&token_account_info.data.borrow())?;
+        if account.is_initialized() {
+            return Err(TokenError::AlreadyInUse.into());
+        }
+        msg!("Not doing anything, instruction support is only for associated-token-account compatibility");
```
mvines, line 778: nit: let's take the opportunity to advertise token 2022 here, as this log message will show up every time a new ATA is created!
```suggestion
        msg!("Please upgrade to SPL Token 2022 for immutable owner support");
```

associated-token-account/program/Cargo.toml
joncinque: Done everywhere, that makes sense

associated-token-account/program/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub fn get_associated_token_address(
     get_associated_token_address_with_program_id(
         wallet_address,
         token_mint_address,
-        &spl_token::id(),
+        &spl_token_2022::id(),
```
mvines, line 43: 🤔 do we really want to do this?  It'll essentially forcibly upgrade all downstream consumers to token-2022 if they upgrade their ATA crate version.  Maybe yes, but if so I think that's a major ATA version bump.  My gut is no for now, to require downstream to deliberately opt-in to token-2022 in all cases until it becomes ubiquitous.

token/program-2022/src/instruction.rs
```diff
@@ -1013,7 +1013,8 @@ pub fn initialize_account3(
     mint_pubkey: &Pubkey,
     owner_pubkey: &Pubkey,
 ) -> Result<Instruction, ProgramError> {
-    check_program_account(token_program_id)?;
+    // NOTE: There is no program account check here for associated-token-account
+    // compatibility with both token programs.
```
CriesofCarrots, line 1017: This feels weird. Can we instead update `check_program_account()` to cover both program_ids?
Or should we use different instruction builders in ATA processor depending on the program_id?

associated-token-account/program/src/lib.rs
```diff
@@ -106,7 +106,7 @@ pub fn create_associated_token_account(
             AccountMeta::new_readonly(*wallet_address, false),
             AccountMeta::new_readonly(*token_mint_address, false),
             AccountMeta::new_readonly(solana_program::system_program::id(), false),
-            AccountMeta::new_readonly(spl_token::id(), false),
+            AccountMeta::new_readonly(spl_token_2022::id(), false),
```
mvines, line 109: same here, I think this ought to remain as spl_token.   Also this is a deprecated method, so another reason to leave it be.

token/program/src/processor.rs
```diff
@@ -757,6 +758,27 @@ impl Processor {
         Ok(())
     }
 
+    /// Processes a [GetAccountDataSize](enum.TokenInstruction.html) instruction
+    pub fn process_get_account_data_size(accounts: &[AccountInfo]) -> ProgramResult {
+        let account_info_iter = &mut accounts.iter();
+        // make sure the mint is passed in
+        let _mint_account_info = next_account_info(account_info_iter)?;
```
CriesofCarrots, line 765: Should we unpack the mint to verify it is a valid mint?

associated-token-account/program/src/lib.rs
joncinque: oops, nope! This is a holdover from before, made clear by the spl-token -> spl-token-2022 crate rename.

associated-token-account/program/src/lib.rs
joncinque: same here

token/program/src/processor.rs
joncinque: Will do

token/program-2022/src/instruction.rs
joncinque: I did this to eventually support generic token programs, but we're likely a ways off from that. We can either use different instruction builders based on program id, or have token-2022 check for either `spl_token_2022::id()` or `spl_token::id()`.  Which do you prefer?

joncinque: I think I like the second one to avoid more branches and to be sure that token-2022 is always compatible with token

CriesofCarrots: I think I prefer the different instruction builders, if it's not too messy.

CriesofCarrots: > I think I like the second one to avoid more branches and to be sure that token-2022 is always compatible with token

Oh, hmm... let me ponder that another minute...

CriesofCarrots: You know, it makes me think... If that's the direction we're going, I wonder if we actually ought to pull spl-token into spl-token-2022 as a dependency, and not duplicate all the stuff we've c+p'ed.

Anyway, I guess I've talked myself down to no preference on the two approaches. I think the extended `check_program_account()` (which I think should be a separate method from the existing `check_program_account()`, incidentally) would require a dependency on spl_token regardless

joncinque: Actually, I'm now leaning strongly towards having spl-token-2022 doing the check for applicable instructions.  If we want other dapps and frontends to upgrade, they'll want backwards compatibility with all the instructions most likely, and not support just the new one. Otherwise they'll have to implement their own branching every time they create a new instruction.

associated-token-account/program/src/instruction.rs
```diff
@@ -9,8 +10,6 @@ use {
     },
 };
 
-use crate::{get_associated_token_address, id};
```
CriesofCarrots, line 12: ❤️

token/program-2022/src/instruction.rs
joncinque: All right, that should be everything. We can reduce the copy-paste between spl-token and spl-token-2022 later. There isn't too much there surprisingly -- only `state.rs` really

associated-token-account/program/tests/spl_token_create.rs
```diff
@@ -0,0 +1,146 @@
...
+
+fn program_test(token_mint_address: Pubkey, use_latest_spl_token: bool) -> ProgramTest {
+    let mut pc = ProgramTest::new(
+        "spl_associated_token_account",
+        id(),
+        processor!(process_instruction),
+    );
+
+    if use_latest_spl_token {
+        // TODO: Remove after Token >3.2.0 is available by default in program-test
```
CriesofCarrots, line 26: Update? Will it be 3.4?

CriesofCarrots (approved):
Nitty nit, lgtm if CI is happy!

associated-token-account/program/tests/spl_token_create.rs
joncinque: Ah right, the change will be in 3.4, so `> 3.3.0`

PR #2908, opened 2022-02-13
Add Soteria analysis job to CI

dmakarov:
This PR adds a CI job running Soteria static analyzer to catch known vulnerabilities. It currently reports over a 100 of vulnerability locations in SPL code. All of these are annotated by ignore directive, so that the CI can pass successfully. However, it may be worth the effort to go over the ignored vulnerabilities and check if they are real of false positive.

### DIFF
--- a/token/js/package.json
+++ b/token/js/package.json
@@ -42,7 +42,7 @@
     "dependencies": {
         "@solana/buffer-layout": "^4.0.0",
         "@solana/buffer-layout-utils": "^0.2.0",
-        "@solana/web3.js": "^1.36.0",
+        "@solana/web3.js": "^1.41.0",
         "start-server-and-test": "^1.14.0"
     },
     "devDependencies": {
--- a/token/js/src/state/mint.ts
+++ b/token/js/src/state/mint.ts
@@ -133,7 +133,8 @@ export async function getMinimumBalanceForRentExemptMintWithExtensions(
 }
 
 /**
- * Get the address of the associated token account for a given mint and owner
+ * Async version of getAssociatedTokenAddressSync
+ * For backwards compatibility
  *
  * @param mint                     Token mint account
  * @param owner                    Owner of the new account
@@ -141,7 +142,7 @@ export async function getMinimumBalanceForRentExemptMintWithExtensions(
  * @param programId                SPL Token program account
  * @param associatedTokenProgramId SPL Associated Token program account
  *
- * @return Address of the associated token account
+ * @return Promise containing the address of the associated token account
  */
 export async function getAssociatedTokenAddress(
     mint: PublicKey,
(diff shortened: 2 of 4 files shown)


mvines (dismissed):
This is an interesting experiment but I think Soteria needs to improve its analysis and suppression directives before we should consider merging this PR

dmakarov:
I asked soteria for a feature to specify ignore statements in an external input file for the tool. Thus we shouldn't need to add the ignore comments in the code, but could add this file somewhere in the repository tree for the tool to use when we run it on CI. Would that be an acceptable alternative, at least until they improve the accuracy of their analysis?

mvines:
Did soteria actually find anything beyond its own limitations?

ci/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN apt-get update -qq && \
+    apt-get install -y --no-install-recommends \
+    ca-certificates \
+    curl \
+    gcc \
+    libc6-dev
+
+RUN curl --proto '=https' --tlsv1.2 -sSf -o soteria-linux-rust.tar.gz https://supercompiler.xyz/y/soteria-linux-rust.tar.gz
+RUN tar xf soteria-linux-rust.tar.gz
+ENV SOTERIA_LICENSE_KEY=2zCz2GgSoSS68eNJENWrYB48dMM1zmH8SZkgYneVDv2G4gRsVfwu5rNXtK5BKFxn7fSqX9BvrBc1rdPAeBEcD6Es
```
mvines, line 11: `SOTERIA_LICENSE_KEY` - 😬 

Using proprietary tools is another strike in my books. This basically amounts to an advertisement.

dmakarov:
> Did soteria actually find anything beyond its own limitations?

This experiment started after the recent wormhole incident.

Soteria claims they would have caught the bug automatically
https://medium.com/coinmonks/the-wormhole-hack-how-soteria-detects-the-vulnerability-automatically-eb0f433e8071

mvines:
> > Did soteria actually find anything beyond its own limitations?
> 
> This experiment started after the recent wormhole incident.
> 
> Soteria claims they would have caught the bug automatically https://medium.com/coinmonks/the-wormhole-hack-how-soteria-detects-the-vulnerability-automatically-eb0f433e8071

Oh sorry, I meant did their proprietary tool find anything in SPL?  I see it generated a ton of noise but I'm looking for what value it added.

dmakarov:
Most of the reported vulnerabilities are "untrustful account issues". For example https://github.com/dmakarov/solana-program-library/runs/5192154992?check_suite_focus=true#step:3:12215
I don't have necessary knowledge to tell whether this is a real vulnerability.

jackcmay:
This has been an interesting experiment, I agree with your concerns @mvines.  I also think that if Soltera can get this tool in good shape it could be helpful for the ecosystem.  I'd be fine if @dmakarov spends some more time to identify what isn't working well or well for us and give that feedback they may be able to come back with something that we can use.  I'm not happy about the proprietary part either, but not sure that should be a blocker if the tool did rock at finding vulnerabilities with a much lower noise ratio.

mvines:
If Soltera wants to sell their tool, they should do the work to see how it works with SPL. Supporting the free version wholeheartedly is great provided they will continue to make it available to the community. I’m pretty opposed to giving free engineering time to people that want to then turn around and charge money for what we helped them build.

Issue #2909, opened 2022-02-13
Option to make Non-Transferable Tokens (NTT)


joxoby:
Non-transferability of tokens can be useful when the token represents achievements, certifications, events, etc. that the owner is related to. If the token can be sold/bought, it becomes meaningless.

The current token program doesn't seem to provide this type of functionality.

I wonder what you all think about adding a "non-transferable" flag to the token initializer function.

Related issue on Ethereum: https://github.com/ethereum/EIPs/issues/1238
Related blog post: https://vitalik.ca/general/2022/01/26/soulbound.html

mvines:
This sounds like a great way to spam a user's wallet with advertisements/scams that they can't get rid of.  This is already happening regularly but at least the victims can currently dispose of the token.

jordansexton:
I actually think this has utility to create [soulbound](https://vitalik.ca/general/2022/01/26/soulbound.html) tokens which have some interesting uses (decentralized identity comes to mind).

Could you make it so the token can be burned but not transfered? This would help with the spam issue.

joxoby:
> This sounds like a great way to spam a user's wallet with advertisements/scams that they can't get rid of. This is already happening regularly but at least the victims can currently dispose of the token.

Valid concern. But in this case, the non-transferable tokens could still be burnt.

mvines:
Wallets need to support burning tokens first.

joxoby:
Draft PR: https://github.com/solana-labs/solana-program-library/pull/2912

jpbogle:
FYI @joxoby we have been exploring a similar thing using a program we call the "token_manager" https://github.com/cardinal-labs/cardinal-token-manager

If you are the mint creator, can you not just utilize freeze authority to freeze everytime you mint someone a token? Also I feel like non-transferability could be modeled more flexibly with a `transfer_authority: COption<Pubkey>`, we have done this in our program by requiring a transfer_authority to create a "transfer_receipt" that someone can use to transfer the token to the approved party. Would be a cool feature in token-program and also solve non-transferability / soul tokens by just setting transfer_authority to None

joncinque:
Added with #3178

fullstackdev0410:
Hello
Can I solve the non-transferability without any contract?

fullstackdev0410:
I was gonna develop the passport mint website and I used metaplex.
Now I need to do non-transferability of nft.
Do I have to develop the smart contract?

joncinque:
[...]

PR #2912, opened 2022-02-13
token-2022: Add support for Non-Transferable Tokens (NTTs) - NonTransferableMint extension

joxoby:
Related issue: https://github.com/solana-labs/solana-program-library/issues/2909

Add support for tokens that represent achievements, certifications, attendances to an event, etc. In short, everything that would lose its intrinsic value if it could be traded in the market.

The PR adds a new extension `NonTransferable` that, when enabled on a mint, disables the ability to transfer the tokens between accounts. The only mechanism to send tokens to an account is by using `MintTo`.

To prevent accounts with non-transferable tokens to change owners, `MintTo` will only succeed for accounts that have immutable ownership, such as associated token accounts.

Also the check for non-transferability can only happen in function `TransferChecked` and not `Transfer` since the latter doesn't receive the mint account information. Fortunately, `Transfer` is now deprecated. Do we expect to totally disable this function anytime soon?

Although not transferable, these tokens can still be burnt. This will help to alleviate airdrop spam, as the victim would still be able to discard any undesired NTT tokens.

As @mvines correctly noted, currently most wallets don't have support for burning. We should time the release of this feature with wallet developers to keep the users able to discard any undesirable tokens.

Thanks to @guibescos for the feedback and discussions that helped shaped this concept.


mvines (commented):
This looks really good, sorry for the delayed review.

Would you mind rebasing and addressing the clippy complaints?

mvines:
I rebased this PR and moved the commits to https://github.com/solana-labs/solana-program-library/pull/3178

PR #2922, opened 2022-02-16
Fix Broken Link to Token Program TypeScript Docs

zicklag:
Just found the broken link and it looks like I found the correct path. :)


jordansexton:
Thanks!

PR #2925, opened 2022-02-17
Fix a typo in instruction.rs

vni:
Hi!

A small typo in solana-program-library/token-swap/program/src/instruction.rs.


mvines:
Thanks!

PR #2926, opened 2022-02-17
token-2022: Support extensions in `SyncNative`

joncinque:
#### Problem

Extensions aren't supported in `SyncNative`, so native accounts created with `ImmutableOwner`, as done with ATA, fail to sync.  This PR assumes that we want native mint accounts to support extensions, which for `ImmutableOwner`, makes sense to me.  We'll have to write a carve-out for the native mint in confidential transfers though.

#### Solution

Use `StateWithExtensionsMut` in `process_sync_native`.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -722,6 +722,15 @@ where
         .await
     }
 
+    /// Sync native account lamports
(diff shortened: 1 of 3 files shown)


CriesofCarrots (commented):
😅 good catch, thanks! lgtm

PR #2928, opened 2022-02-17
Add token Amount/UiAmount conversion Instructions

CriesofCarrots:
The extensions model  in spl-token-2022 gets us most of the way to being able to support interest-bearing tokens: https://github.com/solana-labs/solana/pull/15927
In addition to actual iToken extension definitions, we need on-chain methods to convert between raw token amounts (the number stored in the account) and ui token amount strings. The ui_amount currently only includes the decimal place, but would include the effects of interest for iTokens.

The PR implements those instructions for both token programs, so that RPC handling can be consistent.

### DIFF
--- a/token/program-2022/src/instruction.rs
+++ b/token/program-2022/src/instruction.rs
@@ -28,7 +28,7 @@ const U64_BYTES: usize = 8;
 /// Instructions supported by the token program.
 #[repr(C)]
 #[derive(Clone, Debug, PartialEq)]
-pub enum TokenInstruction {
+pub enum TokenInstruction<'a> {
     /// Initializes a new mint and optionally deposits all the newly minted
     /// tokens in an account.
     ///
@@ -469,6 +469,32 @@ pub enum TokenInstruction {
     ///   None
     ///
     InitializeImmutableOwner,
+    /// Convert an Amount of tokens to a UiAmount `string`, using the given mint.
+    ///
+    /// Fails on an invalid mint.
+    ///
+    /// Return data can be fetched using `sol_get_return_data` and deserialized with
+    /// `String::from_utf8`.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[]` The mint to calculate for
+    AmountToUiAmount {
+        /// The amount of tokens to convert.
+        amount: u64,
+    },
+    /// Convert a UiAmount of tokens to a little-endian `u64` raw Amount, using the given mint.
+    ///
+    /// Return data can be fetched using `sol_get_return_data` and deserializing
+    /// the return data as a little-endian `u64`.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[]` The mint to calculate for
+    UiAmountToAmount {
+        /// The ui_amount of tokens to convert.
+        ui_amount: &'a str,
+    },
     /// Initialize the close account authority on a new mint.
     ///
     /// Fails if the mint has already been initialized, so must be called before
@@ -541,9 +567,9 @@ pub enum TokenInstruction {
     ///
     CreateNativeMint,
 }
-impl TokenInstruction {
+impl<'a> TokenInstruction<'a> {
     /// Unpacks a byte buffer into a [TokenInstruction](enum.TokenInstruction.html).
-    pub fn unpack(input: &[u8]) -> Result<Self, ProgramError> {
+    pub fn unpack(input: &'a [u8]) -> Result<Self, ProgramError> {
         use TokenError::InvalidInstruction;
 
         let (&tag, rest) = input.split_first().ok_or(InvalidInstruction)?;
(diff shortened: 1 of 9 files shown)


token/program/src/processor.rs
```diff
@@ -768,7 +770,8 @@ impl Processor {
         // make sure the mint is valid
         let mint_info = next_account_info(account_info_iter)?;
         Self::check_account_owner(program_id, mint_info)?;
-        let _ = Mint::unpack(&mint_info.data.borrow())?;
+        let _ = Mint::unpack(&mint_info.data.borrow())
+            .map_err(|_| Into::<ProgramError>::into(TokenError::InvalidMint))?;
```
CriesofCarrots, line 774: This change is a little out of the scope of this PR, but happened on it while comparing test outputs between the two token programs, which were a little inconsistent.

CriesofCarrots:
This hacky patch to the ATA program demonstrates how these instructions would be used:

```patch
diff --git a/associated-token-account/program/src/processor.rs b/associated-token-account/program/src/processor.rs
index 58f1e022..3ab568ba 100644
--- a/associated-token-account/program/src/processor.rs
+++ b/associated-token-account/program/src/processor.rs
@@ -155,5 +155,52 @@ fn process_create_associated_token_account(
             wallet_account_info.clone(),
             spl_token_program_info.clone(),
         ],
-    )
+    )?;
+
+    use std::convert::TryInto;
+    invoke(
+        &spl_token_2022::instruction::ui_amount_to_amount(
+            spl_token_program_id,
+            spl_token_mint_info.key,
+            "0.000042",
+        )?,
+        &[spl_token_mint_info.clone(), spl_token_program_info.clone()],
+    )?;
+
+    let amount = solana_program::program::get_return_data()
+        .ok_or(ProgramError::InvalidInstructionData)
+        .and_then(|(key, data)| {
+            if key != *spl_token_program_id {
+                return Err(ProgramError::IncorrectProgramId);
+            }
+            data.try_into()
+                .map(u64::from_le_bytes)
+                .map_err(|_| ProgramError::InvalidInstructionData)
+        })?;
+    msg!("AMOUNT {:?}", amount);
+
+    invoke(
+        &spl_token_2022::instruction::amount_to_ui_amount(
+            spl_token_program_id,
+            spl_token_mint_info.key,
+            42000,
+        )?,
+        &[spl_token_mint_info.clone(), spl_token_program_info.clone()],
+    )?;
+
+    let ui_amount = solana_program::program::get_return_data()
+        .ok_or(ProgramError::InvalidInstructionData)
+        .and_then(|(key, data)| {
+            if key != *spl_token_program_id {
+                return Err(ProgramError::IncorrectProgramId);
+            }
+            data.try_into()
+                .map_err(|_| ProgramError::InvalidInstructionData)
+                .and_then(|data| {
+                    String::from_utf8(data).map_err(|_| ProgramError::InvalidInstructionData)
+                })
+        })?;
+    msg!("UI_AMOUNT {:?}", ui_amount);
+
+    Ok(())
 }
 ```

File: token/program/src/lib.rs [hunk 2]
```diff
@@ -27,6 +27,55 @@ pub fn amount_to_ui_amount(amount: u64, decimals: u8) -> f64 {
...
+/// Convert a raw amount to its UI representation using the given decimals field
+/// Excess zeroes or unneeded decimal point are trimmed.
+pub fn amount_to_ui_amount_string_trimmed(amount: u64, decimals: u8) -> String {
+    let mut s = amount_to_ui_amount_string(amount, decimals);
+    if decimals > 0 {
+        let zeros_trimmed = s.trim_end_matches('0');
+        s = zeros_trimmed.trim_end_matches('.').to_string();
+    }
+    s
+}
```
CriesofCarrots, line 53: The [monorepo methods](https://github.com/solana-labs/solana/blob/5726f42a7cf0faacbceafc19b2293397b99017f1/account-decoder/src/parse_token.rs#L179-L199) can be removed and replaced with these when spl-token is bumped.

joncinque (approved):
Looks great! The test coverage is nice too.  I was worried about weird floating point representations, but the way you did it is totally clear.  I also read through the `FromStr` implementation on integer types, and it's really quite simple thankfully, so we're definitely good.

There's just the one question about including the account, but that could be done with a new instruction / separate PR.

token/program-2022/src/instruction.rs
```diff
@@ -469,6 +469,32 @@ pub enum TokenInstruction {
...
+    /// Convert an Amount of tokens to a UiAmount `string`, using the given mint.
+    ///
+    /// Fails on an invalid mint.
+    ///
+    /// Return data can be fetched using `sol_get_return_data` and deserialized with
+    /// `String::from_utf8`.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[]` The mint to calculate for
```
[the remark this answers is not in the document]
CriesofCarrots: Yeah, I can see a use for that. I wonder if that should/could be part of the iToken extension? Or would it need to be supported by both token programs?
I'll merge this as-is for now, but we can update this interface while building out the extension, if it seems like it would be more ergonomic

PR #2932, opened 2022-02-18
Add zero check to checked_u8_power() and checked_u8_mul() in curve/stable.rs

vni:
Added zero check info few functions in stable.rs.

Actually, I was just lurking the source code. And do not know if it's truly needed.

### DIFF
--- a/token-swap/program/src/curve/stable.rs
+++ b/token-swap/program/src/curve/stable.rs
@@ -36,8 +36,8 @@ fn compute_a(amp: u64) -> Option<u64> {
 
 /// Returns self to the power of b
 fn checked_u8_power(a: &U256, b: u8) -> Option<U256> {
-    let mut result = *a;
-    for _ in 1..b {
+    let mut result = U256::from(1);
+    for _ in 0..b {
         result = result.checked_mul(*a)?;
     }
     Some(result)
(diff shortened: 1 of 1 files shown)


token-swap/program/src/curve/stable.rs
```diff
@@ -36,15 +36,19 @@ fn compute_a(amp: u64) -> Option<u64> {
 
 /// Returns self to the power of b
 fn checked_u8_power(a: &U256, b: u8) -> Option<U256> {
-    let mut result = *a;
-    for _ in 1..b {
+    let mut result = U256::from(1);
+    for _ in 1..=b {
```
joncinque, line 40: if `b = 0`, shouldn't it be better to do ?
```suggestion
    for _ in 0..b {
```

joncinque (commented):
Thanks for the contribution! Any chance we could add a couple of tests for this?

token-swap/program/src/curve/stable.rs
vni: yes, thanks, it looks nicer

joncinque:
Would you mind adding a test for each of these cases that you're fixing?

vni:
Hi!
Thanks about the test suggestions. I will check it in the nearest few days, but don't want to make any promises right now. : )

joncinque:
Closing this since the stable curve was removed

PR #2937, opened 2022-02-22
Add zktoken paper

mvines:


mvines:
@samkim-crypto - I've removed the paper from the zk-token repo as well (https://github.com/solana-labs/spl-zk-token/commit/261302c30a88e9c38e51d1708d2e8daaad04436e).  If you feel there's a better home in the SPL repo for the paper please lmk

samkim-crypto:
Got it. Thanks!

PR #2940, opened 2022-02-23
Custom build.rs for program-2022-test

CriesofCarrots:
#### Problem

Running the program-2022-test suite uses the most recently built spl_token_2022.so, leading to potentially tedious debugging, if you've made changes to the program, but haven't explicitly rebuilt it.

#### Solution

Add custom build.rs to program-2022-test that runs `cargo build-bpf --manifest-path <path/to/program-2022>` if any of the program files have been modified.

@joncinque , can you give this a try and see what you think? Particularly, how do you like the logging in the 2nd commit? I thought it felt weird to not have the .so info and etc, but it is a bit yelly, since the warning was the only way I could figure out how to expose the output. I'm torn about whether to keep or drop that commit.
Incidentally, that logging will appear on every `cargo test-bpf` in program-2022-test, the same way that it does for a direct build or test in `program-2022`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3942,6 +3942,7 @@ dependencies = [
  "spl-associated-token-account 1.0.5",
  "spl-token-2022",
  "spl-token-client",
+ "walkdir",
 ]
 
 [[package]]
--- a/token/program-2022-test/Cargo.toml
+++ b/token/program-2022-test/Cargo.toml
@@ -10,6 +10,9 @@ version = "0.0.1"
 [features]
 test-bpf = []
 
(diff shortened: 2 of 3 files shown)


joncinque (approved):
Looks good! We can eventually make it less yelly by implementing `-vv` for `test-bpf` and `build-bpf` to get all of the output without the `warning (not a warning)`, but it doesn't seem too pressing.

CriesofCarrots:
token-js failure is unrelated

mvines:
Thanks, this makes iterating on test/program changes much easier

PR #2941, opened 2022-02-23
token-2022: Add confidential transfer initialize/update mint tests

mvines:


mvines:
token-js failure is unrelated

PR #2949, opened 2022-02-24
token-js: Update instruction enum for UiAmountToAmount / AmountToUiAmount

joncinque:
#### Problem

The token-js tests keep failing on the `e2e-built` step.

#### Solution

The enum values for the new `UiAmountToAmount` and `AmountToUiAmount` instructions were not included, so `CreateNativeMint` was actually calling `Reallocate` in the JS bindings.


CriesofCarrots:
Aha, thanks for tracking this down! Should we add a comment to either `token` Instruction, `token-2022` Instruction, or the TS instruction types to help prevent this (or at least make it easier to diagnose) in the future?

joncinque:
That's a good point, seems like we should add a comment to `token`'s instruction.rs to update token-2022 and token-js enums.

PR #2970, opened 2022-03-01
Support unpacking token accounts fields partially

lijunwangs:
In certain applications, there is a need to quickly unpacking some subfields of the account without unpacking the whole account to take actions such as creating secondary indexes. The code implemented here was adapted from solana-runtime.

### DIFF
--- a/token/program-2022/src/state.rs
+++ b/token/program-2022/src/state.rs
@@ -1,7 +1,7 @@
 //! State transition types
 
 use {
-    crate::instruction::MAX_SIGNERS,
+    crate::{instruction::MAX_SIGNERS, extension::AccountType},
     arrayref::{array_mut_ref, array_ref, array_refs, mut_array_refs},
     num_enum::{IntoPrimitive, TryFromPrimitive},
     solana_program::{
@@ -290,7 +290,8 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
     }
 }
 
-const ACCOUNTTYPE_ACCOUNT: u8 = 2;
+// `spl_token_program_2022::extension::AccountType::Account` ordinal value
+const ACCOUNTTYPE_ACCOUNT: u8 = AccountType::Account as u8;
 impl GenericTokenAccount for Account {
     fn valid_account_data(account_data: &[u8]) -> bool {
         spl_token::state::Account::valid_account_data(account_data)
--- a/token/program-2022/src/state.rs
+++ b/token/program-2022/src/state.rs
@@ -415,4 +415,56 @@ pub(crate) mod test {
         let unpacked = Multisig::unpack(&packed).unwrap();
         assert_eq!(unpacked, check);
     }
+
+    #[test]
+    fn test_unpack_token_owner() {
+        // Account data length < Account::LEN, unpack will not return a key
+        let src: [u8; 12] = [0; 12];
+        let result = Account::unpack_account_owner(&src);
+        assert_eq!(result, Option::None);
+
+        // The right account data size, unpack will return some key
+        let src: [u8; Account::LEN] = [0; Account::LEN];
+        let result = Account::unpack_account_owner(&src);
+        assert!(result.is_some());
+
+        // Account data length > account data size, but not a valid extension,
+        // unpack will not return a key
+        let src: [u8; Account::LEN + 5] = [0; Account::LEN + 5];
+        let result = Account::unpack_account_owner(&src);
+        assert_eq!(result, Option::None);
+
+        // Account data length > account data size with a valid extension,
+        // expect some key returned
+        let mut src: [u8; Account::LEN + 5] = [0; Account::LEN + 5];
+        src[Account::LEN] = 2;
+        let result = Account::unpack_account_owner(&src);
+        assert!(result.is_some());
+    }
+
+    #[test]
+    fn test_unpack_token_mint() {
+        // Account data length < Account::LEN, unpack will not return a key
+        let src: [u8; 12] = [0; 12];
+        let result = Account::unpack_account_mint(&src);
+        assert_eq!(result, Option::None);
+
+        // The right account data size, unpack will return some key
+        let src: [u8; Account::LEN] = [0; Account::LEN];
+        let result = Account::unpack_account_mint(&src);
+        assert!(result.is_some());
+
+        // Account data length > account data size, but not a valid extension,
+        // unpack will not return a key
+        let src: [u8; Account::LEN + 5] = [0; Account::LEN + 5];
+        let result = Account::unpack_account_mint(&src);
+        assert_eq!(result, Option::None);
+
+        // Account data length > account data size with a valid extension,
+        // expect some key returned
+        let mut src: [u8; Account::LEN + 5] = [0; Account::LEN + 5];
+        src[Account::LEN] = 2;
+        let result = Account::unpack_account_mint(&src);
+        assert!(result.is_some());
+    }
 }
--- a/token/program-2022/src/state.rs
+++ b/token/program-2022/src/state.rs
@@ -1,7 +1,7 @@
 //! State transition types
 
 use {
-    crate::{instruction::MAX_SIGNERS, extension::AccountType},
+    crate::{extension::AccountType, instruction::MAX_SIGNERS},
     arrayref::{array_mut_ref, array_ref, array_refs, mut_array_refs},
     num_enum::{IntoPrimitive, TryFromPrimitive},
     solana_program::{
--- a/token/program-2022/src/state.rs
+++ b/token/program-2022/src/state.rs
@@ -10,6 +10,7 @@ use {
         program_pack::{IsInitialized, Pack, Sealed},
         pubkey::Pubkey,
     },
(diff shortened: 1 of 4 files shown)


token/program/src/state.rs
```diff
@@ -287,6 +287,64 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
...
+        }
+    }
+}
+
+impl Account {
+    /// Return the packed account data length in bytes.
+    pub fn get_packed_len() -> usize {
+        SPL_TOKEN_ACCOUNT_LENGTH
+    }
+}
```
CriesofCarrots, line 340: I don't think we need this method, since the additions in this PR are for callers who are okay with an spl_token dependency. Those callers can use `Account::LEN` instead.

token/program-2022/src/state.rs
```diff
@@ -289,6 +290,18 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
...
 }
 
+// `spl_token_program_2022::extension::AccountType::Account` ordinal value
+const ACCOUNTTYPE_ACCOUNT: u8 = AccountType::Account as u8;
+impl GenericTokenAccount for Account {
+    fn valid_account_data(account_data: &[u8]) -> bool {
+        spl_token::state::Account::valid_account_data(account_data)
+            || ACCOUNTTYPE_ACCOUNT
+                == *account_data
+                    .get(spl_token::state::Account::get_packed_len())
```
CriesofCarrots, line 300:
```suggestion
                    .get(Account::LEN)
```

token/program/src/state.rs
```diff
@@ -287,6 +287,64 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
     }
 }
 
+const SPL_TOKEN_ACCOUNT_MINT_OFFSET: usize = 0;
+const SPL_TOKEN_ACCOUNT_OWNER_OFFSET: usize = 32;
+const SPL_TOKEN_ACCOUNT_LENGTH: usize = 165;
```
CriesofCarrots, line 292: Can remove

token/program/src/state.rs
```diff
@@ -287,6 +287,64 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
...
+
+/// A trait of the generic token account for unpacking an account's fields
+/// partially and returning their references efficiently.
```
CriesofCarrots, line 295:
```suggestion
/// A trait for token Account structs to enable efficiently unpacking various fields 
/// without unpacking the complete state.
```

token/program/src/state.rs
```diff
@@ -287,6 +287,64 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
...
+        Self::unpack_pubkey_unchecked(account_data, SPL_TOKEN_ACCOUNT_MINT_OFFSET)
+    }
+
+    /// Call after account length has already been verified to unpack a Pubkey at
+    /// the specified offset.
+    fn unpack_pubkey_unchecked(account_data: &[u8], offset: usize) -> &Pubkey {
+        bytemuck::from_bytes(&account_data[offset..offset + PUBKEY_BYTES])
+    }
+
+    /// Function for unpacking the account owner from the account data.
```
CriesofCarrots, line 316:
```suggestion
    /// Unpacks an account's owner from opaque account data.
```

token/program/src/state.rs
```diff
@@ -287,6 +287,64 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
...
+    fn unpack_account_owner(account_data: &[u8]) -> Option<&Pubkey> {
+        if Self::valid_account_data(account_data) {
+            Some(Self::unpack_account_owner_unchecked(account_data))
+        } else {
+            None
+        }
+    }
+
+    /// Function for unpacking the account mint from the account data.
```
CriesofCarrots, line 325:
```suggestion
    /// Unpacks an account;s mint from opaque account data.
```

token/program/src/state.rs
CriesofCarrots: This could probably use some more wordsmithing.

token/program/src/state.rs
```diff
@@ -360,4 +418,40 @@ mod tests {
...
+        let src: [u8; 12] = [0; 12];
+        let result = Account::unpack_account_owner(&src);
+        assert_eq!(result, Option::None);
+
+        // The right account data size, unpack will return some key
+        let src: [u8; SPL_TOKEN_ACCOUNT_LENGTH] = [0; SPL_TOKEN_ACCOUNT_LENGTH];
+        let result = Account::unpack_account_owner(&src);
+        assert!(result.is_some());
+
+        // Account data length > account data size, unpack will return some key
```
CriesofCarrots, line 434:
```suggestion
        // Account data length > account data size, unpack will not return a key
```

token/program/src/state.rs
```diff
@@ -360,4 +418,40 @@ mod tests {
...
+        let src: [u8; 12] = [0; 12];
+        let result = Account::unpack_account_mint(&src);
+        assert_eq!(result, Option::None);
+
+        // The right account data size, unpack will return some key
+        let src: [u8; SPL_TOKEN_ACCOUNT_LENGTH] = [0; SPL_TOKEN_ACCOUNT_LENGTH];
+        let result = Account::unpack_account_mint(&src);
+        assert!(result.is_some());
+
+        // Account data length > account data size, unpack will return some key
```
CriesofCarrots, line 452:
```suggestion
        // Account data length > account data size, unpack will not return a key
```

token/program-2022/src/state.rs
```diff
@@ -289,6 +290,18 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
...
+                    .unwrap_or(&0)
+    }
+}
+
```
CriesofCarrots, line 304: Can you please add test cases for `unpack_account_mint()` and `unpack_account_owner` for an extended token account?

lijunwangs: Will do.

lijunwangs: Done

token/program/src/state.rs
lijunwangs: Done

token/program/src/state.rs
lijunwangs: Done

token/program/src/state.rs
lijunwangs: Done

CriesofCarrots (approved):
lgtm!

token/program/src/state.rs
lijunwangs: Done

token/program/src/state.rs
lijunwangs: Done

token/program/src/state.rs
```diff
@@ -287,6 +287,56 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
...
+            Some(Self::unpack_account_mint_unchecked(account_data))
+        } else {
+            None
+        }
+    }
+}
+
+impl GenericTokenAccount for Account {
+    fn valid_account_data(account_data: &[u8]) -> bool {
+        account_data.len() == Account::LEN
```
joncinque, line 336: Sorry for the late review here, but to be safe, this could also check that `is_initialized` is properly set, as in https://github.com/solana-labs/solana-program-library/blob/bbbf250a67777a3dea8717648bf1b78d72fa1a8c/token/program-2022/src/extension/mod.rs#L217-L224

joncinque (commented):
Just a drive-by suggestion, looks good otherwise!

token/program/src/state.rs
lijunwangs: Thanks for the suggestion -- already merged code but I will have another PR to address it.

joncinque: Yeah I noticed, wanted to make you aware either way!

lijunwangs: Hi Jon, the code you pointed to is in the program-2022 module, is that relevant for the regular SPL token? .i.e the ACCOUNT_INITIALIZED_INDEX position and the assumption of it need to be initialized to be valid?

Secondly -- if I understand correctly, the account extension can have more data then Account::Len -- is the following logic still good?

joncinque: > the code you pointed to is in the program-2022 module, is that relevant for the regular SPL token? .i.e the ACCOUNT_INITIALIZED_INDEX position and the assumption of it need to be initialized to be valid?

Yes, it is still relevant for regular SPL token.

> Secondly -- if I understand correctly, the account extension can have more data then Account::Len -- is the following logic still good?

For normal SPL token, it's perfect, but correct, token-2022 can have a longer length, so that would not work.  You'd have to do some logic like you did here:

```
        spl_token::state::Account::valid_account_data(account_data)
            || ACCOUNTTYPE_ACCOUNT == *account_data.get(Account::LEN).unwrap_or(&0)
    }
```

PR #2973, opened 2022-03-01
include version in typedoc for spl-name-service, spl-token, and spl-token-lending

qpwo:
I was using spl-token and had some problems and it turns out I was on the wrong version!

It's just a single argument to typedoc (`--includeVersion`) to make it put the version number in the docs.

### DIFF
--- a/name-service/js/package.json
+++ b/name-service/js/package.json
@@ -25,7 +25,7 @@
     "lint:fix": "yarn pretty:fix && eslint . --fix",
     "pretty": "prettier --check 'src/*.[jt]s'",
     "pretty:fix": "prettier --write 'src/*.[jt]s'",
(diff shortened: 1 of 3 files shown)


qpwo:
See also https://github.com/solana-labs/solana/pull/23428

qpwo:
(Pretty sure the cargo build didn't fail because of the typedoc flag)

CriesofCarrots:
> (Pretty sure the cargo build didn't fail because of the typedoc flag)

Haha, nope. I'm looking at the CI failure now

qpwo:
Very short diff, just adding the flag to typedoc: https://github.com/solana-labs/solana-program-library/pull/2973/files

PR #2975, opened 2022-03-02
Override the constructor to Numberu32 and Numberu64 to resolve #2563

### LINKED ISSUE #2563: token-js: `u64` should implement its own constructor
Try this code:

```
const bn = new BN(5);
const u = new u64(10);
const u2 = new u64(bn);

console.log("bn.toBuffer", bn.toBuffer);
console.log("u.toBuffer", u.toBuffer);
console.log("u2.toBuffer", u2.toBuffer);
```

Logging `u.toBuffer` shows a function. But logging `u2.toBuffer` shows undefined, even though `toBuffer` is defined right here! https://github.com/solana-labs/solana-program-library/blob/master/token/js/client/token.js#L54-L65

Why? It's because of `BN`'s constructor: https://github.com/indutny/bn.js/blob/master/lib/bn.js#L21-L24

This behavior is very misleading in TypeScript, because `u2` is typed as a `u64`, which you expect to have a `toBuffer` function.

In order to fix this, I think we should override `BN`'s constructor in `u64`.

tian000:
…63 by overriding the constructor of BN sub classes

This PR resolves https://github.com/solana-labs/solana-program-library/issues/2563

The following code:
```
const bn = new BN(5);
const u = new Numberu64(10);
const u2 = new Numberu64(bn);

console.log("bn.toBuffer", bn.toBuffer);
console.log("u.toBuffer", u.toBuffer);
console.log("u2.toBuffer", u2.toBuffer);
```

Now results in the following output:
```
bn.toBuffer [Function: toBuffer]
u.toBuffer [Function: toBuffer]
u2.toBuffer [Function: toBuffer]
```

### DIFF
--- a/name-service/js/src/utils.ts
+++ b/name-service/js/src/utils.ts
@@ -15,6 +15,19 @@ import { HASH_PREFIX, NAME_PROGRAM_ID } from './bindings';
 import { NameRegistryState } from './state';
 
 export class Numberu32 extends BN {
+  constructor(
+      number: number | string | number[] | Uint8Array | Buffer | BN,
+      base?: number | 'hex',
+      endian?: BN.Endianness
+  ) {
+    if (BN.isBN(number)) {
+      const nbn = number as BN
+      super(nbn.toArray())
+    } else {
+      super(number, base, endian)
+    }
+  }
+
   /**
    * Convert to Buffer representation
    */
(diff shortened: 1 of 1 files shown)


jordansexton:
@tian000 I closed #2563 because the linked issue in the token client lib no longer applies. I see that this PR is for the nameservice though?

PR #2980, opened 2022-03-03
fuzz: Fix fuzz build

joncinque:
#### Problem

Since upgrading to 1.59.0, the fuzz build doesn't work anymore, with an error just like https://github.com/rust-fuzz/honggfuzz-rs/issues/61

#### Solution

Apply a workaround to use the old LLVM pass manager, which requires Rust nightly.


CriesofCarrots (approved):
Thank you!

PR #2984, opened 2022-03-04
anchor: Updates for token-2022

joncinque:
#### Problem

Verifiable builds and publishing source code doesn't work currently, since not all required programs are included in the Anchor workspace.

#### Solution

Add required programs to publish token and token-2022.

While we're at it, upgrade the solana version to 1.9.9.

### DIFF
--- a/Anchor.toml
+++ b/Anchor.toml
@@ -1,10 +1,13 @@
 anchor_version = "0.20.1"
-solana_version = "1.9.5"
+solana_version = "1.9.9"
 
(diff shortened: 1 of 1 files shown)


CriesofCarrots:
Oh hey, should we do something to get update-solana-dependencies.sh to bump this as well, so it stays in sync with the repo? Or is it best left manual?

joncinque:
Yes, great call! #2996

PR #2985, opened 2022-03-04
token-2022: Make program and crate deployable

joncinque:
#### Problem

People want to start using mint close authorities afforded by token-2022, but the program hasn't been deployed, and cannot be released until reallocate, sibling instruction, and zk ops are available.

#### Solution (?)

Comment out just the processor bits that rely on future syscalls. The rest of the crate is usable, and the program build optimizes away unused processor code.

Also, since there was a dependency on new parts of spl-token, copy those bits over.

This PR will be reverted and version bumped to 0.3.0 after the crate is published.


token/program-2022/src/lib.rs
```diff
@@ -35,6 +35,55 @@ pub fn amount_to_ui_amount(amount: u64, decimals: u8) -> f64 {
     amount as f64 / 10_usize.pow(decimals as u32) as f64
 }
 
+/// Convert a raw amount to its UI representation (using the decimals field defined in its mint)
```
CriesofCarrots, line 38: Just to be explicit, these functions are duplicated so that we can move forward with the release/deploy before releasing a new spl_token?

token/program-2022/src/state.rs
```diff
@@ -290,15 +289,6 @@ fn unpack_coption_u64(src: &[u8; 12]) -> Result<COption<u64>, ProgramError> {
     }
 }
 
-// `spl_token_program_2022::extension::AccountType::Account` ordinal value
-const ACCOUNTTYPE_ACCOUNT: u8 = AccountType::Account as u8;
-impl GenericTokenAccount for Account {
-    fn valid_account_data(account_data: &[u8]) -> bool {
-        spl_token::state::Account::valid_account_data(account_data)
```
CriesofCarrots, line 297: Can we duplicate the `valid_account_data` logic (spl_token dependency) instead of removing this? That way we don't have to remember to re-add this either.

CriesofCarrots (dismissed):
The feature approach turned out nice!
One nit

token/program-2022/src/lib.rs
joncinque: Yes exactly. I know it's not ideal, but releasing the token crate is such a hassle that duplicating was by far easier than deploying and all that.

token/program-2022/src/state.rs
joncinque: Yeah sure, I'll duplicate the trait too, that way we'll just have to delete that trait declaration and impl the one from spl_token once we're ready.

And what's really neat, there won't be anything to re-add!

PR #2988, opened 2022-03-04
token-2022: Add confidential transfer with fee instruction

samkim-crypto:
I added a separate instruction for confidential transfer with fee. 

Confidential transfer with and without fees require different zkps. I was originally considering having a single confidential transfer instruction for either with or without fees. The processor would decode the zk verify instruction and distinguish the fee/no fee case. But since the two zkps require different inputs, I think the client code will be a bit simpler to have two separate instructions.


samkim-crypto:
@mvines Can you look at this briefly to check if this looks reasonable? 

The part that I am unsure about is the new instruction `TransferWithFee` and the `confidential_transfer_transfer_with_fee` in `client` that takes in an epoch information.

The changes in `program-2022-test` are going to be updated with decryption once zk-token-sdk 0.8.0 becomes available, so you can probably take a more detailed look in a future PR.

samkim-crypto:
Hm, for some reason, the test-bpf passes on my local machine, but it seems to fail here. I must done some commit wrong. I will figure it out soon...

samkim-crypto:
Hm... I seems to always get EqualityProof error, but all the proofs seem to verify on the client side https://github.com/solana-labs/solana-program-library/pull/2988/commits/734d3f7a92efc77d435f1e0a73561642f486a371. Not sure what is going on...

mvines:
> Hm... I seems to always get EqualityProof error, but all the proofs seem to verify on the client side [734d3f7](https://github.com/solana-labs/solana-program-library/commit/734d3f7a92efc77d435f1e0a73561642f486a371). Not sure what is going on...

You could try instrumenting the code around https://github.com/solana-labs/solana/blob/949006b5a2559b57dd5967963b44a232277f63e2/programs/zk-token-proof/src/lib.rs#L62 to see if something different is getting passed into the actual proof program perhaps?

token/client/src/token.rs
```diff
@@ -1148,6 +1150,95 @@ where
...
+        )?;
+
+        let mint_state = self.get_mint_info().await.unwrap();
+        let transfer_fee_config = mint_state
+            .get_extension::<transfer_fee::TransferFeeConfig>()
+            .unwrap();
+
+        // Zkp generator requires transfer fee parameters as input
+        let fee_parameters =
+            if u64::from(transfer_fee_config.newer_transfer_fee.epoch) < epoch_info.epoch {
```
mvines, line 1185: It looks like you could use `transfer_fee_config.get_epoch_fee(epoch_info.epoch)` to obtain the appropriate `TransferFee`, which'll save a couple lines of code here

samkim-crypto: Yep this is applied now, thanks!

PR #2990, opened 2022-03-06
add getMultipleAccounts for Token

gebob19:
impl of `getMultipleAccounts` allows to request **multiple token accounts in a single RPC call** compared to `getAccount` which can only get 1 account per RPC call

### DIFF
--- a/token/js/src/state/account.ts
+++ b/token/js/src/state/account.ts
@@ -82,7 +82,6 @@ export const AccountLayout = struct<RawAccount>([
 /** Byte length of a token account */
 export const ACCOUNT_SIZE = AccountLayout.span;
 
-
 /**
  * Retrieve information about a token account
  *
@@ -113,7 +112,7 @@ export async function getAccount(
  *
  * @return Token account information
  */
- export async function getMultipleAccounts(
+export async function getMultipleAccounts(
     connection: Connection,
     addresses: PublicKey[],
     commitment?: Commitment,
--- a/token/js/src/state/account.ts
+++ b/token/js/src/state/account.ts
@@ -82,59 +82,6 @@ export const AccountLayout = struct<RawAccount>([
 /** Byte length of a token account */
 export const ACCOUNT_SIZE = AccountLayout.span;
 
-function unpackTokenAccount(info: AccountInfo<Buffer> | null, address: PublicKey, programId: PublicKey) {
-    if (!info) throw new TokenAccountNotFoundError();
-    if (!info.owner.equals(programId)) throw new TokenInvalidAccountOwnerError();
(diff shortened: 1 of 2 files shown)


token/js/src/state/account.ts
```diff
@@ -82,23 +82,11 @@ export const AccountLayout = struct<RawAccount>([
...
- * @param connection Connection to use
- * @param address    Token account
- * @param commitment Desired level of commitment for querying the state
- * @param programId  SPL Token program account
- *
- * @return Token account information
- */
-export async function getAccount(
-    connection: Connection,
+function info2Account(
```
joncinque, line 85: nit: can we call this `unpackTokenAccount` or `fromAccountInfo`? The `2` is unfortunately not in our conventions

joncinque (commented):
Thanks for the contribution!  Looks really good, just a comment about the conversion function name.  Also, I'll kick off the CI, but I think the linter might yell at this, so be sure to run `yarn lint:fix`

joncinque:
Ah sorry, looks like you'll need to rebase on https://github.com/solana-labs/solana-program-library/pull/2995 to fix the audit issue

joncinque:
There's a build issue when updating to the new `@solana/web3.js`, so I'll take care of that in #3001, then you'll have to rebase once more before we can merge.  We're close!

token/js/src/state/account.ts
```diff
@@ -82,23 +82,7 @@ export const AccountLayout = struct<RawAccount>([
...
- * @return Token account information
- */
-export async function getAccount(
-    connection: Connection,
-    address: PublicKey,
-    commitment?: Commitment,
-    programId = TOKEN_PROGRAM_ID
-): Promise<Account> {
-    const info = await connection.getAccountInfo(address, commitment);
+function unpackTokenAccount(info: AccountInfo<Buffer> | null, address: PublicKey, programId: PublicKey) {
```
jordansexton, line 85: nits: possible to call this `unpackAccount` for consistency, and move it down to the bottom of the file since it's an internal function?

token/js/src/state/account.ts
```diff
@@ -127,6 +111,51 @@ export async function getAccount(
...
+ * Retrieve information about multiple token accounts in a single RPC call
+ *
+ * @param connection Connection to use
+ * @param addresses  Token accounts
+ * @param commitment Desired level of commitment for querying the state
+ * @param programId  SPL Token program account
+ *
+ * @return Token account information
+ */
+export async function getMultipleAccounts(
```
jordansexton, line 124: nit: possible to move this after `getAccount`? Just looking to have the file flow logically

jordansexton (commented):
Looks good! Only some nonblocking nits from me.

joncinque:

PR #2995, opened 2022-03-08
Bump regex to v1.5.5

CriesofCarrots:
`cargo audit`


CriesofCarrots:
I bumped farms-rpc because I was in the area, but I think the farms stuff is rotting, since it's excluded from everything.

PR #2996, opened 2022-03-09
scripts: Update solana_version in Anchor.toml with dependencies

joncinque:
#### Problem

The `solana_version` in `Anchor.toml` isn't updated with the rest of the dependencies during `update-solana-dependencies.sh`, which leads to awkward updates later on bundled up with other changes.

#### Solution

Also update `solana_version` during `update-solana-dependencies.sh`


CriesofCarrots (approved):
Awesome

PR #2997, opened 2022-03-09
Update twoxtx patch

CriesofCarrots:
One type in rpc.rs changed upstream, breaking the patch.


mvines (approved):
ty

CriesofCarrots:
Past the twoxtx test, merging

PR #2998, opened 2022-03-09
Skip patching zk-token-sdk in twoxtx tests

CriesofCarrots:
There may be breaking changes coming to solana-zk-token-sdk... we don't want to bust token CI for everyone when they do. Currently, twoxtx-setup.sh pulls the monorepo master and patches all the solana crates, pulling any changes merged upstream and breaking the confidential-transfer tests if they haven't been updated yet.

This PR skips patching the zk-token-sdk crate for twoxtx tests so it depends on a manual bump and stays in sync with the extension.

### DIFF
--- a/patch.crates-io.sh
+++ b/patch.crates-io.sh
@@ -3,7 +3,24 @@
 # Patches the SPL crates for developing against a local solana monorepo
 #
 
-solana_dir=$1
+maybe_patch_zk_token_sdk=true
+positional_args=()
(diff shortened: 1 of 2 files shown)


CriesofCarrots:
Hrm, this now passes for me locally...

CriesofCarrots:
This isn't going to work because solana-program-test pulls in the master solana-zk-token-sdk via solana-runtime. I guess we're back to pinning.

PR #3010, opened 2022-03-14
Fix typo in token-swap/program/src/curve/stable.rs

codewithgun:
Fix typo remove_liquidty -> remove_liquidity


joncinque:
Thanks for the contribution!

PR #3011, opened 2022-03-16
token-2022: Refactor extension instruction encoding / decoding

joncinque:
#### Problem

There are a lot of copies of instruction encoding / decoding in the extensions.

#### Solution

Unify most of them!  A few notes:

* the default account extension can't easily follow this model because the data passed in is `AccountState`, and not `Pod`, so I excluded it from this
* this changes the confidential token instruction to use `num_enum` instead of `num_traits`, which has a bit more in terms of type and size safety


mvines (approved):
nice!

PR #3012, opened 2022-03-18
Add Anchor version of the record - MLH

Anurag5086:


record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[account(mut)]
+    pub authority: Signer<'info>,
+    pub system_program: Program<'info, System>
+}
+
+#[derive(Accounts)]
+pub struct SetAuthority<'info> {
+    #[account(mut, has_one = authority)]
+    pub record_account: Account<'info, RecordData>,
+    #[account(mut)]
```
joncinque, line 95: This probably doesn't need to be `mut` since the data and lamports aren't changing

record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// CHECK: This is not dangerous because we don't read or write from this account
+    pub new_authority: UncheckedAccount<'info>,
+    pub system_program: Program<'info, System>
+}
+
+#[derive(Accounts)]
+pub struct CloseAccount<'info> {
+    #[account(mut, has_one = authority)]
+    pub record_account: Account<'info, RecordData>,
+    #[account(mut)]
```
joncinque, line 106: Same here, `mut` probably not required

record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// CHECK: This is not dangerous because we don't read or write from this account
+    pub reciever: UncheckedAccount<'info>,
+    pub system_program: Program<'info, System>
+}
+
+#[derive(Accounts)]
+pub struct Write<'info> {
+    #[account(mut, has_one = authority)]
+    pub record_account: Account<'info, RecordData>,
+    #[account(mut)]
```
joncinque, line 117: And here as well

record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+const DATA_SIZE: usize = 8;
+const CURRENT_VERSION: u8 = 1;
+const WRITABLE_START_INDEX: usize = 33;
+
+#[account]
+pub struct RecordData {
+    /// Struct version, allows for upgrades to the program
+    pub version: u8,
```
joncinque, line 129: How does this interact with the Account discriminator used by Anchor? I'm wondering if it's even needed considering Anchor does some of this work for us!

record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[account(mut, has_one = authority)]
+    pub record_account: Account<'info, RecordData>,
+    #[account(mut)]
+    pub authority: Signer<'info>,
+    pub system_program: Program<'info, System>
+}
+
+const DATA_SIZE: usize = 8;
+const CURRENT_VERSION: u8 = 1;
+const WRITABLE_START_INDEX: usize = 33;
```
joncinque, line 124: Does this work with the extra discriminator added by Anchor? Or are the indexes properly shifted in the deserialized account?

record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+        
+        **data_info.to_account_info().lamports.borrow_mut() = 0;
+        **destination_info.to_account_info().lamports.borrow_mut() = destination_starting_lamports
+            .checked_add(data_lamports)
+            .ok_or(ProgramError::Custom(0))?;
+        account_data.data = Data::default();
+        Ok(())
+    }
+
+    pub fn write(ctx: Context<Write>, offset: u64, data: Vec<u8>) -> Result<()> {
```
joncinque, line 63: Can `data` be a slice `&[u8]` instead of a `Vec<u8>`? That would save at least one copy

joncinque (commented):
A few points to consider at first -- otherwise, I'd love to see some tests whenever they're ready!  Also, we're trying to find the right place for this to land so that it'll be properly supported going forward, stay tuned.

record/program-anchor/programs/record-anchor/src/lib.rs
Anurag5086: Makes sense, done!

record/program-anchor/Anchor.toml
```diff
@@ -0,0 +1,14 @@
...
+seeds = false
+[programs.localnet]
+record_anchor = "FGQLqP4smSphjS1WRamE3zEezL8tCFZrsu4Wg5z2Kq4r"
+
+[registry]
+url = "https://anchor.projectserum.com"
+
+[provider]
+cluster = "localnet"
+wallet = "/home/anurag/.config/solana/id.json"
```
joncinque, line 11: So that everyone can use it :smile: 
```suggestion
wallet = "~/.config/solana/id.json"
```

record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+const DATA_SIZE: usize = 8;
+const CURRENT_VERSION: u8 = 1;
+const WRITABLE_START_INDEX: usize = 0;
+
+#[account]
+#[derive(Default)]
+pub struct RecordData {
+    /// Struct version, allows for upgrades to the program
+    pub version: u8,
```
joncinque, line 106: Since anchor already adds an 8-byte discriminator for us, no need for this!  Unless you want to not use the `#[account]` macro

record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+pub struct Write<'info> {
+    #[account(mut, has_one = authority)]
+    pub record_account: Account<'info, RecordData>,
+    pub authority: Signer<'info>,
+    pub system_program: Program<'info, System>
+}
+
+const DATA_SIZE: usize = 8;
+const CURRENT_VERSION: u8 = 1;
+const WRITABLE_START_INDEX: usize = 0;
```
joncinque, line 100: The purpose of this variable is to show where the program should allow data writing.  If it's `0`, then it's possible to blast the authority without using `SetAuthority`.  So if you keep the version, it should be `33`, otherwise `32`.

record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+#[derive(Accounts)]
+pub struct CloseAccount<'info> {
+    #[account(mut, has_one = authority)]
+    pub record_account: Account<'info, RecordData>,
+    pub authority: Signer<'info>,
+    #[account(mut)]
+    /// CHECK: This is not dangerous because we don't read or write from this account
+    pub reciever: UncheckedAccount<'info>,
```
joncinque, line 86: nit:
```suggestion
    pub receiver: UncheckedAccount<'info>,
```

record/program-anchor/programs/record-anchor/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+        
+        **data_info.to_account_info().lamports.borrow_mut() = 0;
+        **destination_info.to_account_info().lamports.borrow_mut() = destination_starting_lamports
+            .checked_add(data_lamports)
+            .ok_or(ProgramError::Custom(0))?;
+        data_info.data = RecordData::default().data;
+        Ok(())
+    }
+
+    pub fn write(ctx: Context<Write>, offset: u64, data: Vec<u8>) -> Result<()> {
```
joncinque, line 46: So no way to change this to a slice?

joncinque (commented):
A few more bits, this is getting really close!

record/program-anchor/programs/record-anchor/src/lib.rs
Anurag5086: Yeah, so removed the version and setting the `WRITABLE_START_INDEX` to `32`. It throws error saying `account data too small for instruction`.

record/program-anchor/programs/record-anchor/src/lib.rs
Anurag5086: Yeah, we'll be using the `#[account]` macro, so we won't need the version.

record/program-anchor/programs/record-anchor/src/lib.rs
joncinque: Right, in your case, you don't need this variable at all! Since Anchor nicely deserializes everything for you, you just write into `data` directly as you're doing and remove this

Anurag5086: Yeah right, removed the variable.

jnappy:
@Anurag5086 are you still working on this?

Anurag5086:
> @Anurag5086 are you still working on this?

The PR is completed and was also approved by @joncinque . Jon said, he'll be adding this to a different repository.

PR #3026, opened 2022-03-23
Governance: Update addin accounts discriminators to Anchor compatible

SebastianBor:
#### Summary

`MaxVoterWeightRecord` and `VoterWeightRecord` are defined in `spl-governance-addin-api` but they are owned by the addin programs. In order to make them easier to use in `Anchor` based programs their discriminators were updated to match the ones automatically generated by `Anchor`.

For backward compatibility with existing addins the legacy discriminator are also considered as valid ones.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3678,7 +3678,7 @@ dependencies = [
 
 [[package]]
 name = "spl-governance"
-version = "2.2.3"
+version = "2.2.4"
 dependencies = [
  "arrayref",
  "assert_matches",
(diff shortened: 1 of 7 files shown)


ckamm:
When I looked at this last time, the attempt to be compatible with Anchor didn't work because the discriminator field was part of the struct itself in spl-gov, while it's implicit for Anchor accounts structs. That was the reason the struct still needs to be wrapped and we never ended up testing the account discriminator value: https://github.com/blockworks-foundation/voter-stake-registry/blob/master/programs/voter-stake-registry/src/governance.rs

How do you use the struct with anchor without wrapping it?

SebastianBor:
> How do you use the struct with anchor without wrapping it?

I redefine the structs in Anchor program without the descriptor

PR #3029, opened 2022-03-24
transferFee Amount Template

ronald-hood-jr:
A template for how the transfer fee bindings will be implemented in the token 2022 library


token/js/src/instructions/initializeTransferFee.ts
```diff
@@ -0,0 +1,90 @@
...
+    TokenInvalidInstructionTypeError,
+} from '../errors';
+import { TokenInstruction } from './types';
+
+/** TODO: docs */
+
+export interface InitializeTransferFeeAmountInstructionData {
+    instruction: TokenInstruction.TransferFeeExtension;
+    transferFeeAmount: Number;
+}
```
joncinque, line 17: There's no instruction that does this unfortunately -- perhaps you meant to map `InitializeTransferFeeConfig`? https://github.com/solana-labs/solana-program-library/blob/24baf875e9e19c26d694d28c557d33848c3a9180/token/program-2022/src/extension/transfer_fee/instruction.rs#L28

token/yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
joncinque, line 4: This file can be deleted

token/js/src/extensions/transferFee.ts
```diff
@@ -0,0 +1,24 @@
+import { Layout, struct } from '@solana/buffer-layout';
+import { Mint } from '../state/mint';
+import { ExtensionType, getExtensionData } from './extensionType';
+
+
+/** TransferFeeAmount as stored by the program */
+export interface TransferFeeAmount {
+    feeAmount: Number;
```
joncinque, line 8: Since the fee is given by a `u64` in the code, we should use a `bigint` here, similar to the `amount` in `Account`

token/js/src/extensions/transferFee.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+export declare const feeAmount: (property?: number | undefined) => Layout<Number>;
+
+/** Buffer layout for de/serializing a mint */
+export const TransferFeeAmountLayout = struct<TransferFeeAmount>([feeAmount(0)]);
```
joncinque, line 13:
```suggestion
export const TransferFeeAmountLayout = struct<TransferFeeAmount>([u64('withheld_amount')]);
```

PR #3045, opened 2022-03-28
Immutable owner bindings Template 

ronald-hood-jr:
Template for immutable owner bindings


token/js/src/extensions/immutableOwner.ts
```diff
@@ -0,0 +1,24 @@
+import { struct } from '@solana/buffer-layout';
+import { publicKey } from '@solana/buffer-layout-utils';
+import { PublicKey } from '@solana/web3.js';
+import { Mint } from '../state/mint';
+import { ExtensionType, getExtensionData } from './extensionType';
+
+/** MintCloseAuthority as stored by the program */
+export interface ImmutableOwner {
+    owner: PublicKey;
```
joncinque, line 9: This is tricky -- there is no data in the immutable owner extension, so you really only need to check its existence

token/js/src/extensions/immutableOwner.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+/** Buffer layout for de/serializing a mint */
+export const ImmutableOwnerLayout = struct<ImmutableOwner>([publicKey('owner')]);
+
+export const IMMUTABLE_OWNER_SIZE = ImmutableOwnerLayout.span;
+
+export function getImmutableOwner(mint: Mint): ImmutableOwner | null {
```
joncinque, line 17: The `ImmutableOwner` extension only works on an account, and not a mint: https://github.com/solana-labs/solana-program-library/blob/24baf875e9e19c26d694d28c557d33848c3a9180/token/program-2022/src/extension/mod.rs#L695 -- and essentially all we might need to know is if the account's owner is immutable or not

joncinque (commented):
An e2e test for this would be great!  Also, you'll need to create a different branch for this work, since it includes the transfer fee bits from #3029

PR #3046, opened 2022-03-29
token-cli-bench: Remove RPC utils in favor of Tpu Client

joncinque:
#### Problem

The token CLI is still running a custom version of `send_and_confirm_messages_with_spinner` in its bench, when the TPU client has this already built in.

#### Solution

Remove it in favor of the `TpuClient`'s version.

Note: I tested this with the bench `create-account` against a local test validator.


CriesofCarrots (approved):
Lgtm!
However, to be clear, this only updates the cli bench to use TpuClient. The cli bin itself uses RpcClient directly.

joncinque:
>  However, to be clear, this only updates the cli bench to use TpuClient. The cli bin itself uses RpcClient directly.

Correct, I'll update the description to be clearer

PR #3070, opened 2022-04-08
token-cli: Add tests for all cases from docs

joncinque:
#### Problem

The token CLI has no tests, making it hard to see if token-2022 will also be supported from the CLI.

#### Solution

Refactor the CLI to be testable, and add tests inspired by the docs.

Note: due to the build issue on `solana-test-validator` with the yanked lru crate, this depends on the upgrade to 1.10.7.  Once the upgrade is pushed, this can use `TestValidator` everywhere.  Currently, in my own testing environment, I have it run against a test validator.


token/cli/src/main.rs
```diff
@@ -2603,15 +2628,27 @@ fn main() -> Result<(), Error> {
...
-        ("bench", Some(arg_matches)) => bench_process_command(
+fn process_command(
+    sub_command: &str,
+    sub_matches: &ArgMatches<'_>,
+    config: &Config,
+    mut wallet_manager: Option<Arc<RemoteWalletManager>>,
+    mut bulk_signers: Vec<Box<dyn Signer>>,
+) -> CommandResult {
+    match (sub_command, sub_matches) {
+        ("bench", arg_matches) => bench_process_command(
```
CriesofCarrots, line 2645: I spent too long trying to figure out why this went from `Some(arg_matches)` to `arg_matches`... wow, our handling of matches/arg_matches/sub_matches was dumb. Thanks for cleaning it up!

CriesofCarrots (commented):
Maybe I missed something, but these don't actually get run in CI, do they?
Refactor and tests look good, though!

joncinque:
> Maybe I missed something, but these don't actually get run in CI, do they?

They get run as part of `cargo-build-test`. You can see the run starting at https://github.com/solana-labs/solana-program-library/runs/6007493531?check_suite_focus=true#step:9:1767 -- is that ok?

CriesofCarrots:
Aha, yep, did miss it. I think the log didn't expand fully when I was searching for test names. Thanks!

CriesofCarrots (approved):

PR #3071, opened 2022-04-08
token-cli: Add compatibility for token-2022

joncinque:
#### Problem

The token CLI only works with `Tokenkeg...` and not `Tokenz...`.

#### Solution

Support both of them at once!  Most of the changes here involve deserializing accounts with extensions, and passing in the program id correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6167,6 +6167,7 @@ dependencies = [
  "spl-associated-token-account 1.0.5",
  "spl-memo 3.0.1",
  "spl-token 3.3.0",
+ "spl-token-2022 0.4.1",
  "strum",
  "strum_macros",
  "tempfile",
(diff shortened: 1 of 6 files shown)


joncinque:
@CriesofCarrots feel free to use this branch to test RPC changes! All of the tests marked with TODO for token-2022 do not pass with token-2022, but should pass with token-2022 RPC work.

joncinque:
@samkim-crypto if you want to use this branch, you just need to patch your SPL repo by running `./patch-crates.io path/to/your/solana/repo` and then you can start using the `program_client` in the Config object

joncinque:
@CriesofCarrots this one's finally ready for a look too!

CriesofCarrots (approved):
Awesome, lgtm!

PR #3072, opened 2022-04-09
made lending market authority param optional in deposit reserve liquidity

tjkyner:
### **Problem**
The TypeScript implementation of `depositReserveLiquidityInstruction` currently requires supplying the `lendingMarketAuthority` pubkey. This stands in contrast with the Rust program where the `lendingMarketAuthority` is derived within the instruction itself as seen below. As a result, developers must redundantly derive the `lendingMarketAuthority` pubkey ahead of time.

```rust
/// Creates a 'DepositReserveLiquidity' instruction.
pub fn deposit_reserve_liquidity(
    program_id: Pubkey,
    liquidity_amount: u64,
    source_liquidity_pubkey: Pubkey,
    destination_collateral_pubkey: Pubkey,
    reserve_pubkey: Pubkey,
    reserve_liquidity_supply_pubkey: Pubkey,
    reserve_collateral_mint_pubkey: Pubkey,
    lending_market_pubkey: Pubkey,
    user_transfer_authority_pubkey: Pubkey,
) -> Instruction {
    let (lending_market_authority_pubkey, _bump_seed) = Pubkey::find_program_address(
        &[&lending_market_pubkey.to_bytes()[..PUBKEY_BYTES]],
        &program_id,
    );
    Instruction {
        program_id,
        accounts: vec![
            AccountMeta::new(source_liquidity_pubkey, false),
            AccountMeta::new(destination_collateral_pubkey, false),
            AccountMeta::new(reserve_pubkey, false),
            AccountMeta::new(reserve_liquidity_supply_pubkey, false),
            AccountMeta::new(reserve_collateral_mint_pubkey, false),
            AccountMeta::new_readonly(lending_market_pubkey, false),
            AccountMeta::new_readonly(lending_market_authority_pubkey, false),
            AccountMeta::new_readonly(user_transfer_authority_pubkey, true),
            AccountMeta::new_readonly(spl_token::id(), false),
        ],
        data: LendingInstruction::DepositReserveLiquidity { liquidity_amount }.pack(),
    }
}
```

---

### **Summary of Changes**
These changes make supplying the `lendingMarketAuthority` parameter optional.

### DIFF
--- a/token-lending/js/src/instructions/depositReserveLiquidity.ts
+++ b/token-lending/js/src/instructions/depositReserveLiquidity.ts
@@ -20,8 +20,8 @@ export const depositReserveLiquidityInstruction = (
     reserveLiquiditySupply: PublicKey,
     reserveCollateralMint: PublicKey,
     lendingMarket: PublicKey,
-    lendingMarketAuthority: PublicKey,
-    transferAuthority: PublicKey
+    transferAuthority: PublicKey,
+    lendingMarketAuthority?: PublicKey
 ): TransactionInstruction => {
     const data = Buffer.alloc(DataLayout.span);
     DataLayout.encode(
@@ -39,11 +39,14 @@ export const depositReserveLiquidityInstruction = (
         { pubkey: reserveLiquiditySupply, isSigner: false, isWritable: true },
         { pubkey: reserveCollateralMint, isSigner: false, isWritable: true },
         { pubkey: lendingMarket, isSigner: false, isWritable: false },
-        { pubkey: lendingMarketAuthority, isSigner: false, isWritable: false },
         { pubkey: transferAuthority, isSigner: true, isWritable: false },
         { pubkey: SYSVAR_CLOCK_PUBKEY, isSigner: false, isWritable: false },
         { pubkey: TOKEN_PROGRAM_ID, isSigner: false, isWritable: false },
     ];
+
+    if (lendingMarketAuthority) {
+        keys.push({pubkey: lendingMarketAuthority, isSigner: false, isWritable: false})
+    }
 
     return new TransactionInstruction({
         keys,
--- a/token-lending/js/src/instructions/depositReserveLiquidity.ts
+++ b/token-lending/js/src/instructions/depositReserveLiquidity.ts
@@ -45,7 +45,7 @@ export const depositReserveLiquidityInstruction = (
     ];
 
     if (lendingMarketAuthority) {
-        keys.push({pubkey: lendingMarketAuthority, isSigner: false, isWritable: false})
+        keys.push({ pubkey: lendingMarketAuthority, isSigner: false, isWritable: false });
     }
 
     return new TransactionInstruction({


token-lending/js/src/instructions/depositReserveLiquidity.ts
```diff
@@ -39,12 +39,15 @@ export const depositReserveLiquidityInstruction = (
...
-        { pubkey: lendingMarketAuthority, isSigner: false, isWritable: false },
         { pubkey: transferAuthority, isSigner: true, isWritable: false },
         { pubkey: SYSVAR_CLOCK_PUBKEY, isSigner: false, isWritable: false },
         { pubkey: TOKEN_PROGRAM_ID, isSigner: false, isWritable: false },
     ];
 
+    if (lendingMarketAuthority) {
+        keys.push({pubkey: lendingMarketAuthority, isSigner: false, isWritable: false})
+    }
+
```
tjkyner, line 50: Not sure if this is actually needed but wanted to maintain backwards compatibility just in case. Happy to remove if deemed unnecessary!

tjkyner:
@t-nelson my apologies if you're not the correct person to reach out to but I was wondering if you could take a look at this? I don't seem to be able to tag anyone for review and I'm not sure what the next steps are (or perhaps I'm just being impatient lol)

token-lending/js/src/instructions/depositReserveLiquidity.ts
joncinque: We're better off just removing the parameter totally -- flipping the parameters will cause strange behavior, so let's instead break people totally when they upgrade.

joncinque (commented):
Thanks for your contribution! This won't work unfortunately, because we need to derive the proper lending market authority and pass that in to the instruction, as demonstrated in the Rust code that you linked.

tjkyner:
@joncinque Thanks for taking the time to review it! Just to clarify, the reason the `lendingMarketAuthority` needs to be included in the list of accounts for the instruction is so the runtime can parallelize it appropriately, correct?

I saw the below code within the `process_deposit_reserve_liquidity` function in the processor and initially figured if it was forcing users to utilize the derived authority anyways, it may as well not require it in the list of accounts in the first place. But it makes much more sense in the context of needing it for appropriate parallelization!

```rust
let authority_signer_seeds = &[
    lending_market_info.key.as_ref(),
    &[lending_market.bump_seed],
];

let lending_market_authority_pubkey = Pubkey::create_program_address(authority_signer_seeds, program_id)?;

if &lending_market_authority_pubkey != lending_market_authority_info.key {
    msg!(
        "Derived lending market authority {} does not match the lending market authority provided {}",
        &lending_market_authority_pubkey.to_string(),
        &lending_market_authority_info.key.to_string(),
    );
    return Err(LendingError::InvalidMarketAuthority.into());
}
```

joncinque:
> Thanks for taking the time to review it! Just to clarify, the reason the lendingMarketAuthority needs to be included in the list of accounts for the instruction is so the runtime can parallelize it appropriately, correct?

Correct, every account must be specified in the instruction, so it isn't possible to dynamically load accounts on-chain.  The logic you're seeing is checking that the proper derivation was used for the program-derived address.  Later on, `lending_market_authority_info` is passed into `invoke_signed` (I think).

tjkyner:
Appreciate it! I'll go ahead and close this PR 👍

Issue #1653, opened 2021-05-02
js: Add option to `skipPreflight` for token and token-swap clients


michaelhly:
`skipPreflight` is hardcoded to `false` in the token and token-swap client where a transaction is sent. I remember this use to be hardcoded to `true`. Can we make this configurable? 

https://github.com/solana-labs/solana-program-library/search?q=skipPreflight

joncinque:
Sounds good, go for it!

joncinque:
This has been fixed with #3094

PR #3073, opened 2022-04-10
js: Add option to skipPreflight for token and token-swap clients - MLH

Anurag5086:
Solves #1653


token-swap/js/src/util/send-and-confirm-transaction.ts
```diff
@@ -10,10 +13,11 @@ export function sendAndConfirmTransaction(
   title: string,
   connection: Connection,
   transaction: Transaction,
+  options: ConfirmOptions,
   ...signers: Array<Account>
 ): Promise<TransactionSignature> {
   return realSendAndConfirmTransaction(connection, transaction, signers, {
```
joncinque, line 19: `realSendAndConfirmTransaction` already takes in `ConfirmOptions` as the final parameter, which means this function is pretty much exactly the same.  I'm thinking we can remove this version of `sendAndConfirmTransaction` and always use the official version from `@solana/web3.js`.  What do you think?

token-swap/js/src/index.ts
```diff
@@ -428,11 +428,16 @@ export class TokenSwap {
       curveParameters,
     );
 
+    const options = {
+      skipPreflight: false,
+    };
```
joncinque, line 433: Instead of hard-coding these everywhere, we can add `options: ConfirmOptions` as a parameter to all of these functions, similar to how the token JS bindings do it.  Here's an example for `createMint`: https://github.com/solana-labs/solana-program-library/blob/dd8f4d138fe5c51067737124cf8feb47daf75254/token/js/src/actions/createMint.ts#L36

This might cause more churn since you'll also have to update the tests calling these functions, but it'll help functionality a lot

joncinque (commented):
This is a great start!  It looks like your editor added a bunch of spurious changes, so be sure to revert those

Anurag5086:
@joncinque I added the new changes and regarding the linting changes they came in when I ran `lint:fix` and `pretty:fix` script. So it can be that it got linted properly now? Not sure tho.

token-swap/js/src/util/send-and-confirm-transaction.ts
Anurag5086: Yeah sounds good, added this.

token-swap/js/src/index.ts
Anurag5086: Made the changes!

token-swap/js/src/index.ts
```diff
@@ -461,9 +466,10 @@ export class TokenSwap {
     userTransferAuthority: Account,
     amountIn: number | Numberu64,
     minimumAmountOut: number | Numberu64,
+    confirmOptions: ConfirmOptions,
```
joncinque, line 469: nit: let's keep these consistent
```suggestion
    confirmOptions?: ConfirmOptions,
```

token-swap/js/src/index.ts
```diff
@@ -429,12 +434,12 @@ export class TokenSwap {
     );
 
     transaction.add(instruction);
+    // 'createAccount and InitializeSwap'
```
joncinque, line 437: nit: you can remove all these comments

token-swap/js/src/index.ts
```diff
@@ -560,9 +566,10 @@ export class TokenSwap {
     poolTokenAmount: number | Numberu64,
     maximumTokenA: number | Numberu64,
     maximumTokenB: number | Numberu64,
+    confirmOptions: ConfirmOptions,
```
joncinque, line 569: same here
```suggestion
    confirmOptions?: ConfirmOptions,
```

token-swap/js/src/index.ts
```diff
@@ -659,9 +666,10 @@ export class TokenSwap {
     poolTokenAmount: number | Numberu64,
     minimumTokenA: number | Numberu64,
     minimumTokenB: number | Numberu64,
+    confirmOptions: ConfirmOptions,
```
joncinque, line 669:
```suggestion
    confirmOptions?: ConfirmOptions,
```

token-swap/js/src/index.ts
```diff
@@ -756,9 +764,10 @@ export class TokenSwap {
     userTransferAuthority: Account,
     sourceTokenAmount: number | Numberu64,
     minimumPoolTokenAmount: number | Numberu64,
+    confirmOptions: ConfirmOptions,
```
joncinque, line 767:
```suggestion
    confirmOptions?: ConfirmOptions,
```

token-swap/js/src/index.ts
```diff
@@ -846,9 +855,10 @@ export class TokenSwap {
     userTransferAuthority: Account,
     destinationTokenAmount: number | Numberu64,
     maximumPoolTokenAmount: number | Numberu64,
+    confirmOptions: ConfirmOptions,
```
joncinque, line 858:
```suggestion
    confirmOptions?: ConfirmOptions,
```

token-swap/js/src/util/send-and-confirm-transaction.ts
```diff
@@ -1,4 +1,7 @@
-import {sendAndConfirmTransaction as realSendAndConfirmTransaction} from '@solana/web3.js';
+import {
```
joncinque, line 1: nit: I think we can fully delete this file now

joncinque (commented):
It's looking really close! Just a few nits.  There's also the issue with the whitespace showing up tons of changes still, so we'll need to revert those.  I can't really review the test changes since the whole file is shown as changed

Anurag5086:
This PR really got messed up :( 
Created a new one #3094

PR #3076, opened 2022-04-11
deps: Upgrade Solana crates to 1.10.8

joncinque:
#### Problem

Testnet is on 1.10, but SPL is still on 1.9.

#### Solution

Upgrade the crates!


CriesofCarrots (approved):
Thanks! r+ ci

PR #3089, opened 2022-04-18
Remove duplciate output of 'Stake Accounts' header

bshumygora:
**Problem:**

'list -v' command prints 'Stake accounts' header twice.
It looks as follows:

Stake Deposit Referral Fee: 50% of Stake Deposit Fee
SOL Deposit Referral Fee: 50% of SOL Deposit Fee

Stake Accounts
--------------

Stake Accounts
--------------
Reserve Account: GYEzowUBH3XeLjke9z38To1HP8T88K1MGaCq7GJYSso7	Available Balance: ◎0.062412670

**Solution:**

Remove duplicate output of 'Stake Accounts' header

### DIFF
--- a/stake-pool/cli/src/output.rs
+++ b/stake-pool/cli/src/output.rs
@@ -167,8 +167,7 @@ impl VerboseDisplay for CliStakePool {
             &self.sol_referral_fee
         )?;
         writeln!(w)?;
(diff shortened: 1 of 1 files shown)


joncinque:
Thanks for the find and the contribution!

PR #3090, opened 2022-04-18
F32 Normal CDF Instruction and Implementation

ronald-hood-jr:


libraries/math/src/approximations.rs
```diff
@@ -41,6 +41,31 @@ pub fn sqrt<T: PrimInt + CheckedShl + CheckedShr>(radicand: T) -> Option<T> {
...
+pub fn f32_normal_cdf(argument: f32) -> Option<f32> {
+    const PI: f32 = std::f32::consts::PI;
+
+    let mod_argument = if argument < 0.0 {
+        -1.0 * argument
+    } else {
+        argument
+    };
+    let tabulation_numerator: f32 =
+        (1.0 / (1.0 * (2.0 * PI).sqrt())) * (-1.0 * (mod_argument * mod_argument) / 2.0).exp();
```
joncinque, line 61: It looks like the paper is inconsistent about this, and in the "calculate implementation" also suggests:
```suggestion
        (1.0 / (2.0 * (2.0 * PI).sqrt())) * (-1.0 * (mod_argument * mod_argument) / 2.0).exp();
```
Any idea which is right?

libraries/math/src/approximations.rs
```diff
@@ -41,6 +41,31 @@ pub fn sqrt<T: PrimInt + CheckedShl + CheckedShr>(radicand: T) -> Option<T> {
     Some(result)
 }
 
+/// Calculate the normal cdf of the given number
+///
+/// Code lovingly adapted from the excellent work at:
+///
+/// <https://www.hrpub.org/download/20140305/MS7-13401470.pdf>
+///
+/// The algorithm is based on the implementation in the paper above.
```
joncinque, line 50: In numerical computing, it's crucial to provide some guarantees about how accurate the approximation is, ie. out to 3 digits according to the paper.

libraries/math/src/approximations.rs
```diff
@@ -41,6 +41,31 @@ pub fn sqrt<T: PrimInt + CheckedShl + CheckedShr>(radicand: T) -> Option<T> {
...
+#[inline(never)]
+pub fn f32_normal_cdf(argument: f32) -> Option<f32> {
```
joncinque, line 52: nit: This function returns an `Option` but doesn't do any checked math.  Seems like it can just return `f32` in that case, right?

libraries/math/src/approximations.rs
```diff
@@ -67,4 +92,19 @@ mod tests {
             check_square_root(a as u128);
         }
     }
+
+    fn check_normal_cdf_f32(argument: f32) {
+        let result = f32_normal_cdf(argument).unwrap();
+        let abs_difference: f32 = (result - 0.841_344_8).abs();
+        dbg!(result);
```
joncinque, line 99: nit: We can remove the debug for release

libraries/math/src/approximations.rs
```diff
@@ -67,4 +92,19 @@ mod tests {
...
+        assert!(abs_difference <= 0.001);
+    }
+
+    #[test]
+    fn test_normal_cdf_f32() {
+        let test_arguments: [f32; 1] = [1.0];
```
joncinque, line 105: This is a great opportunity to use the excellent `proptest` crate! Since the paper claims a maximum difference of `.0001922`, you can compare the output against `.5 * libm::erff(-1 * x / f32::SQRT_2)`, which should be the closest number. Check out libm https://rust-random.github.io/rand/libm/index.html.  Otherwise, you can check out how proptest is used right above this.

joncinque (commented):
Great start! With a proptest, this'll be amazing

libraries/math/src/approximations.rs
joncinque: Any response here?

ronald-hood-jr:
Hey @joncinque thank you for your feedback! When I included the libm crate in the Cargo.lock, the build failed because of the duplicate crate reference, as seen in the CI/CD tests of f6803fb. Additionally, whenever I ran the prop tests with the "calculate implementation" suggestion, the prop tests began to fail. How do you recommend that I proceed on these two issues?

joncinque:
> When I included the libm crate in the Cargo.lock, the build failed because of the duplicate crate reference

Ah great, thanks for checking that

> Additionally, whenever I ran the prop tests with the "calculate implementation" suggestion, the prop tests began to fail.

In that case, the test absolutely served its purpose, and you implemented it correctly!

PR #3094, opened 2022-04-20
js: Add option to skipPreflight for token and token-swap clients - MLH

Anurag5086:
Solves #1653


joncinque:
Looks great, thanks for your patience!

PR #3102, opened 2022-04-22
token-js: Bindings For Transfer Fee Extension IXs

### LINKED ISSUE #2959: token-js: Support `TransferFee` extension
#### Problem

`token-2022` includes new extension instructions to use transfer fees, but it isn't supported in the JS bindings.  The relevant top-level instruction definition can be found here: https://github.com/solana-labs/solana-program-library/blob/0268d76d306d86a9fb03a17d5ede18144bfad684/token/program-2022/src/instruction.rs#L514-L518

and the sub-instructions here: https://github.com/solana-labs/solana-program-library/blob/0268d76d306d86a9fb03a17d5ede18144bfad684/token/program-2022/src/extension/transfer_fee/instruction.rs#L15-L134

#### Solution

Following the model in #2951 , add support for these new instructions, state, and extension type, including some end-to-end tests.

ronald-hood-jr:
This pr is still under development and not yet ready for serious review. The first commit includes the JS bindings for the generation of transfer fee extension instructions. I've attempted testing, but I still need to do a lot of debugging.

Fixes #2959


joncinque:
I got antsy and ended up adding the tests and actions.  Great job, and thanks for the contribution!

Issue #994, opened 2020-12-31
lending: Support closing obligation accounts


jstarry:
#### Problem
When obligations are fully repaid, a user has no way of reclaiming the SOL used to cover rent for the obligation account

#### Proposed Change
Add a `CloseObligation` instruction which allows closing obligations that are fully repaid

allanperlee:
Hi jstarry,

Is the obligation account initialized in `process_init_reserve`?

jordansexton:
It's initialized in `process_init_obligation` 🙂

allanperlee:
Thanks, is that on line 671, within the `obligation` struct?

jordansexton:
https://github.com/solana-labs/solana-program-library/blob/70973850d3105c6765b42a35d1a67b36d0684c9a/token-lending/program/src/processor.rs#L637

allanperlee:
@jordansexton How do I pass the workflow tests by the way? Even after running `cargo clippy`, the PR failed.

jordansexton:
Hmm, it looks like it needs to have master merged in and the conflicts resolved. If you can do that, I can then trigger them to run again.

allanperlee:
Gitbash is the end of me.
On git I commanded:
`git fetch upstream`
`git merge master/upstream` 
`git commit` then `git push` since Git asked me to complete the merge this way.

PR #3105, opened 2022-04-22
Added a test macro in the instructions.rs file for the close obligati…

allanperlee:
…on account instruction


allanperlee:
@joncinque Hi Jon, this pull request addresses issue #994

token-lending/program/src/processor.rs
```diff
@@ -1696,6 +1700,88 @@ fn process_flash_loan(
...
+        return Err(LendingError::InvalidMarketAuthority.into());    
+    }
+
+    spl_close_account(TokenCloseAccountParams{
+        token_program: token_program_id.clone(),
+        authority: lending_market_authority_info.clone(),
+        source: obligation_owner_info.clone(),
+        destination: destination_info.clone(),
+        authority_signer_seeds,
+    })?;
```
joncinque, line 1780: In this case, `obligation_owner_info` isn't an SPL token account, but rather the actual user interacting with the system, which shouldn't be closed. Instead, you'll want to zero-out the data in `obligation_info` and send the lamports over to destination, implementing similar logic to the token program: https://github.com/solana-labs/solana-program-library/blob/51ece68cec173197d84c19630e3207410db7f58d/token/program/src/processor.rs#L695-L702

token-lending/program/src/processor.rs
```diff
@@ -1696,6 +1700,88 @@ fn process_flash_loan(
...
+        return Err(LendingError::InvalidAccountInput.into());
+    }
+    if &obligation.owner != obligation_owner_info.key {
+        msg!("Obligation owner does not match the obligation owner provided");
+        return Err(LendingError::InvalidObligationOwner.into());
+    }
+    if !&obligation.borrows.is_empty() {
+        msg!("Obligation borrows must be zero");
+        return Err(LendingError::NonZeroObligationBorrow.into());
+    }
```
joncinque, line 1754: We'll also need to check that there aren't deposits

joncinque, line 1754: We'll also need to check that there aren't deposits

token-lending/program/src/processor.rs
```diff
@@ -1696,6 +1700,88 @@ fn process_flash_loan(
...
+
+    let obligation = Obligation::unpack(&obligation_info.data.borrow())?;
+    if obligation_info.owner != program_id {
+        msg!("Obligation provided is not owned by the lending program");
+        return Err(LendingError::InvalidAccountOwner.into());
+    }
+    if &obligation.lending_market != lending_market_info.key {
+        msg!("Obligation lending market does not match the lending market provided");
+        return Err(LendingError::InvalidAccountInput.into());
+    }
```
joncinque, line 1746: Probably don't need to check this, or pass in the lending market at all

token-lending/program/src/processor.rs
```diff
@@ -1696,6 +1700,88 @@ fn process_flash_loan(
...
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+    let obligation_info = next_account_info(account_info_iter)?;
+    let obligation_owner_info = next_account_info(account_info_iter)?;
+    let destination_info = next_account_info(account_info_iter)?;
+    let reserve_info = next_account_info(account_info_iter)?;
+    let lending_market_info = next_account_info(account_info_iter)?;
+    let lending_market_authority_info = next_account_info(account_info_iter)?;
```
joncinque, line 1715: These probably aren't needed

token-lending/program/src/processor.rs
```diff
@@ -1696,6 +1700,88 @@ fn process_flash_loan(
...
+    let token_program_id = next_account_info(account_info_iter)?;
```
joncinque, line 1716: These won't  be needed

joncinque, line 1716: These won't  be needed

token-lending/program/tests/repay_obligation_liquidity.rs
```diff
@@ -144,3 +144,141 @@ async fn test_success() {
...
+    assert_eq!(
+        liquidity_supply_balance,
+        initial_liquidity_supply_balance + USDC_BORROW_AMOUNT_FRACTIONAL
+    );
+
+    let obligation_balance = get_token_balance(&mut banks_client, test_obligation.pubkey).await;
+    assert!(obligation_balance < 1);
+
+    let obligation = test_obligation.get_state(&mut banks_client).await;
+    assert_eq!(obligation.borrows.len(), 0);
```
joncinque, line 283: You'll also want to check that the rent lamports from the obligation went to destination

joncinque, line 283: You'll also want to check that the rent lamports from the obligation went to destination

joncinque (commented):
Thanks for taking this on! The direction is slightly off, but since you've added a test, it shouldn't be too bad to change.

token-lending/program/src/processor.rs
allanperlee: Hi Jon, thanks for the feedback. `process_close_obligation_account` implementation uses processor.rs's `process_close_account`. Should this implementation disregard its use?

allanperlee: Hi Jon, thanks for the feedback. process_close_obligation_account implementation uses processor.rs's process_close_account. Should this implementation disregard its use and/or only make changes to the obligation information?

allanperlee: Hi Jon, thanks for the feedback. process_close_obligation_account implementation uses processor.rs's process_close_account. Should this implementation disregard its use and/or only make changes to the obligation information?  What is the lamport source?

joncinque: Your desire to reuse code is great, but in this case, the runtime won't accept it unfortunately, since it'll look like the token program is trying to move lamports that belong to the lending program.  So you'll make changes to the obligation data yourself, and the `obligation_info` is the source for the lamports.

allanperlee: Got It, I'll find a different method instead of using 'process_close_account`. The source of lamports was misunderstood. The `obligation_info` is the object in obligation.rs correct?

joncinque: Kind of -- `obligation_info` is the name of the variable that you chose for the `AccountInfo` representing the obligation in your processor function

allanperlee: The change is made, and it uses the memset method as well, which zeros the respective data in an `obligation`.  I'll create tests by the end of this week.

token-lending/program/src/processor.rs
```diff
@@ -1696,6 +1701,52 @@ fn process_flash_loan(
...
+        return Err(LendingError::NonZeroObligationDeposit.into());
+    }
+    
+    if obligation_info.owner != program_id {
+        msg!("Obligation provided is not owned by the lending program");
+        return Err(LendingError::InvalidAccountOwner.into());
+    }
+    if !obligation_owner_info.is_signer {
+        msg!("Obligation owner provided must be a signer");
+        return Err(LendingError::InvalidSigner.into());
```
joncinque, line 1735: nit: we could return a `ProgramError::MissingRequiredSignature` instead to be clearer

token-lending/program/src/processor.rs
```diff
@@ -1696,6 +1701,52 @@ fn process_flash_loan(
...
+    }
+
+    let destination_starting_lamports = destination_account_info.lamports();
+    **destination_account_info.lamports.borrow_mut() = destination_starting_lamports
+        .checked_add(obligation_info.lamports())
+        .ok_or(LendingError::MathOverflow)?;
+    
+    **obligation_info.lamports.borrow_mut() = 0;
+
+    sol_memset(*obligation_info.data.borrow_mut(), 0, Account::LEN);
```
joncinque, line 1745: This will only set the first `Account::LEN` bytes to 0, which could create some very strange or exploitable situations.  Be sure to set the whole slice to 0!
```suggestion
    sol_memset(*obligation_info.data.borrow_mut(), 0, Obligation::LEN); 
```

token-lending/program/src/instruction.rs
```diff
@@ -982,6 +996,27 @@ pub fn flash_loan(
...
+    obligation_owner_pubkey: Pubkey,
+    destination_pubkey: Pubkey,
+) -> Instruction {
+    Instruction {
+        program_id,
+        accounts: vec![
+            AccountMeta::new(obligation_pubkey, false),
+            AccountMeta::new(obligation_owner_pubkey, false),
+            AccountMeta::new(destination_pubkey, false),
+            AccountMeta::new(spl_token::id(), false),
```
joncinque, line 1012: You can remove the spl token id from this instruction

joncinque (commented):
We're getting closer on this!

token-lending/program/tests/close_obligation_account.rs
```diff
@@ -0,0 +1,261 @@
...
+    );
+    transaction.sign(&[&payer, &user_accounts_owner], recent_blockhash);
+    assert_eq!(
+        banks_client
+            .process_transaction(transaction)
+            .await
+            .unwrap_err()
+            .unwrap(),
+        TransactionError::InstructionError(
+            1,
```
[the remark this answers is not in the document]
allanperlee: Maybe, I was unsure how the errors were numbered.

token-lending/program/src/processor.rs
allanperlee: I'll implement your suggestions. Thanks for rearing this pull-request!

token-lending/program/src/instruction.rs
```diff
@@ -308,6 +308,16 @@ pub enum LendingInstruction {
...
     },
+    ///14
+    /// Closes obligation account to retrieve SOL rent from the obligation account to loanee
+    /// 
+    /// Accounts expected by this instruction: 
+    /// 
+    /// 0. `[writable]` Obligation account
+    /// 1. `[signer]` Obigation owner
+    /// 2. `[writable]` Destination account
+    /// 6. `[]` Token program id
```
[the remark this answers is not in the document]
allanperlee: You mean item 6?

token-lending/program/tests/close_obligation_account.rs
```diff
@@ -0,0 +1,261 @@
...
+    assert_eq!(
+        liquidity_supply_balance,
+        initial_liquidity_supply_balance + USDC_BORROW_AMOUNT_FRACTIONAL
+    );
+
+    let obligation_balance = get_token_balance(&mut banks_client, test_obligation.pubkey).await;
+    assert!(obligation_balance < 1);
+
+    let obligation = test_obligation.get_state(&mut banks_client).await;
+    assert_eq!(obligation.borrows.len(), 0);
```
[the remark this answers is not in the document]
allanperlee: Totally, this one is still in progress :)

token-lending/program/src/instruction.rs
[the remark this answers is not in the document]
joncinque: Correct

token-lending/program/tests/close_obligation_account.rs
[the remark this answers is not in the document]
joncinque: That first number gives the index of the instruction in the transaction.  In this case, you're looking to have the close instruction fail, which is at index 2 in your transaction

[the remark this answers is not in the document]
allanperlee: i see! Thanks for clarifying!

PR #3106, opened 2022-04-22
token-2022: Add compatibility test for instructions

joncinque:
#### Problem

While doing the work to parse token-2022 and token-v3 instructions, it became clear that we're not totally sure about the compatibility between token-2022 and token-v3.

#### Solution

Since token-2022 is meant to be a superset of token-v3, be sure that all of the instruction creators make the same exact instructions.  This way, when testing instruction parsing on the validator, we can be sure that we only need to support the token-2022 variants.

This actually uncovered a bug!


CriesofCarrots (approved):
Perfect, thank you for doing this!

PR #3111, opened 2022-04-25
Add AuthorityType variant to use for mint close-authority adjustment

CriesofCarrots:
Mints with the `mint_close_authority` extension active expect `SetAuthority` instructions to use the `AuthorityType::CloseAccount` variant. This works, but means that it is not possible to tell if an account being modified is an Account or a Mint by inspecting the instruction. (eg. https://github.com/solana-labs/solana/pull/24537#discussion_r857848860)

This PR adds an `AuthorityType::CloseMint` variant to disambiguate. This is essentially just a cosmetic change, but follows the theme of `TransferChecked` etc exposing context within each instruction.


joncinque (approved):
Looks good!

PR #3112, opened 2022-04-26
Make spl-token-2022 instruction decoders pub

CriesofCarrots:
If we want to parse spl-token-2022 instructions in Rpc, we need the instruction decoders for `default_account_state` and `confidential_token_transfer` instructions (or have to duplicate them by hand 🤢 ). This PR makes them available downstream.

As an aside: I looked at removing `default_account_state`'s custom decoder to use the generic ones, as I had obviously forgotten why I didn't use them in the first place. Unfortunately, a `repr(u8)` enum cannot derive `Pod`. It arguably doesn't need to, but the trait bound prevents calling something like `decode_instruction_data::<AccountState>(input)`

### DIFF
--- a/token/program-2022/src/extension/default_account_state/instruction.rs
+++ b/token/program-2022/src/extension/default_account_state/instruction.rs
@@ -53,7 +53,7 @@ pub enum DefaultAccountStateInstruction {
     Update,
 }
 
-pub(crate) fn decode_instruction(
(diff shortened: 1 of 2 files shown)


joncinque (approved):
Sorry about that, I should have put a comment next to the default account state one when I did this work.  Looks good!

PR #3113, opened 2022-04-26
Bump token-2022

CriesofCarrots:
Preparation for new spl-token-2022 release

It looks like since `solana-send-transaction-service` now depends on `solana-client`, and indirectly on spl-token, in order to publish a bumped spl-token-2022, we'll need to update to a version of solana crates with [this in it](https://github.com/solana-labs/solana/pull/22649). ie, this process for all future spl-token bumps -> token-2022:
1. Bump spl-token and publish
2. Bump spl-token in solana crates and release
3. Bump solana crates in SPL


joncinque (approved):
Gonna bookmark the process.... looks great!

PR #3116, opened 2022-04-26
Add feature-wrapping for new ConfidentialTransferInstruction::TransferWithFee

CriesofCarrots:
We need to build spl-token-2022 without default features to deploy to current public clusters. The new `ConfidentialTransferInstruction::TransferWithFee` instruction doesn't include this feature-wrapping.

cc @samkim-crypto


joncinque (approved):
Thanks for fixing this

PR #3120, opened 2022-04-27
token-2022: Set transfer fee two epochs ahead

joncinque:
#### Problem

Transfer fees are currently set one epoch ahead, in order to avoid a malicious transfer-fee-configurer from imposing a massive fee out of nowhere, potentially making the token unusable.

This still has an issue however, since a truly malicious configurer can simply set the fee right before the end of the epoch, essentially achieving the same rug pull.

#### Solution

Set the epoch fee two epochs forward, giving a minimum of 1 epoch for people to react to the change.

### DIFF
--- a/token/program-2022-test/tests/transfer_fee.rs
+++ b/token/program-2022-test/tests/transfer_fee.rs
@@ -267,6 +267,18 @@ async fn set_fee() {
         }])
         .await
         .unwrap();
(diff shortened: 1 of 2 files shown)


CriesofCarrots (approved):
Nice, such a slim changeset needed :)

PR #3121, opened 2022-04-27
binary-option: Fixed three integer overflows

zhaogang92:
This PR fixed three integer overflows that have been confirmed. If there're any other actions I need to take, please let me know. Thanks!

### DIFF
--- a/binary-option/program/src/state.rs
+++ b/binary-option/program/src/state.rs
@@ -32,7 +32,7 @@ impl BinaryOption {
         self.circulation = self
             .circulation
             .checked_add(n)
-            .ok_or(BinaryOptionError::InvalidSupply)?;
+            .ok_or(BinaryOptionError::AmountOverflow)?;
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


binary-option/program/src/state.rs
```diff
@@ -28,8 +28,12 @@ impl BinaryOption {
         Ok(binary_option)
     }
 
-    pub fn increment_supply(&mut self, n: u64) {
-        self.circulation += n;
+    pub fn increment_supply(&mut self, n: u64) -> ProgramResult {
+        self.circulation = self
+            .circulation
+            .checked_add(n)
+            .ok_or(BinaryOptionError::InvalidSupply)?;
```
joncinque, line 35: Should this be
```suggestion
            .ok_or(BinaryOptionError::AmountOverflow)?;
```

joncinque (commented):
Thanks for the contribution! Any chance you can also commit the tests you used to confirm these issues?

PR #3127, opened 2022-04-29
stake-pool: Update program to work with minimum delegation

joncinque:
#### Problem

The minimum delegation will be raised with https://github.com/solana-labs/solana/pull/24603, and the minimum reserve amount will be lowered with https://github.com/solana-labs/solana/pull/24670 -- the stake pool program needs to be ready for that!

#### Solution

Update the minimum active stake amount and the minimum reserve amount as required.  This was tested against the changes to raise the minimum delegation to 1 SOL, but not the second PR.  The hope is that the only change will be to update `MINIMUM_RESERVE_LAMPORTS` to `0`.

The best option will be to call `get_minimum_delegation` on the stake program, but that hasn't landed in any of the Solana crates yet.

We can land this PR to unblock https://github.com/solana-labs/solana/pull/24603 if needed, or keep it as a draft.  We won't redeploy the program until the new instruction makes it anyway.

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -15,16 +15,14 @@ use {
     borsh::{BorshDeserialize, BorshSerialize},
     num_traits::FromPrimitive,
     solana_program::{
-        account_info::next_account_info,
-        account_info::AccountInfo,
+        account_info::{next_account_info, AccountInfo},
         borsh::try_from_slice_unchecked,
         clock::{Clock, Epoch},
         decode_error::DecodeError,
         entrypoint::ProgramResult,
         msg,
         program::{invoke, invoke_signed},
-        program_error::PrintProgramError,
-        program_error::ProgramError,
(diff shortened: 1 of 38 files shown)


stake-pool/program/src/processor.rs
```diff
@@ -747,7 +745,7 @@ impl Processor {
         stake_pool.manager_fee_account = *manager_fee_info.key;
         stake_pool.token_program_id = *token_program_info.key;
         stake_pool.total_lamports = total_lamports;
-        stake_pool.pool_token_supply = 0;
+        stake_pool.pool_token_supply = total_lamports;
```
joncinque, line 748: The tests uncovered this little oopsie!

brooksprumo (approved):
Lgtm! I did not go in-depth on all the tests; once `MINIMUM_RESERVE_LAMPORTS` drops to 0, that'll effectively double check the tests as well 😊

PR #3133, opened 2022-05-02
Allow immediate proposal execution

holykol:
### Motivation
Provide ability to execute withdrawal/any other action in single transaction when single entity has more than 50% of voting power. In our case, this reduces administration complexity when governance token has not yet been fully distributed to voters, making voting meaningless, since one party always holds deciding vote.
This is not currently possible, because governance program denies execution when `created_at + hold_up_time >= now`, so if `hold_up_time` is zero and time has not changed, error will be raised.

### Proposed changes
Change error condition to `created_at + hold_up_time > now`, so proposal may be created and executed in same transaction, but only if `hold_up_time == 0` 

### Security considerations
In theory, some malicious (mainly flash loan related) attacks can be executed on governance's with vote tipping enabled and more than 50% of community token supply is available on open market.
But, when hold_up_time for governance is set to 0, this attack potential should be in mind, and any DAO's at risk should increase this time to avoid this concern.

### DIFF
--- a/governance/program/src/state/proposal.rs
+++ b/governance/program/src/state/proposal.rs
@@ -671,7 +671,7 @@ impl ProposalV2 {
             .unwrap()
             .checked_add(proposal_transaction_data.hold_up_time as i64)
             .unwrap()
-            >= current_unix_timestamp
+            > current_unix_timestamp
         {
             return Err(GovernanceError::CannotExecuteTransactionWithinHoldUpTime.into());
         }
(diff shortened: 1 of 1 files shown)


joncinque:
@SebastianBor can you take a look at this? Seems like a reasonable change, but you'll have better insight into the potential issues

SebastianBor:
> @SebastianBor can you take a look at this? Seems like a reasonable change, but you'll have better insight into the potential issues

Yes, it makes sense, 0 hold up time indicates no time restrictions on execution after proposal was voted on.
@holykol please create a unit test covering this scenario

holykol:
@SebastianBor done

askibin:
maybe we should check if execution is the last instruction in the transaction if proposal is created and executed within one transaction? this way we would avoid flash loan attack

PR #3135, opened 2022-05-03
Governance: Council Vote Threshold

SebastianBor:
#### Summary 

When council and community tokens have significantly different voting populations then a single `vote_threshold` is not sufficient and both need separate settings for the threshold.

For example NFT voting DAO can have 2-5% approval quorum for NFT votes while 60% for the council.

#### Implementation Details
- Unused `proposal_cool_off_time` field space on `GovernanceConfig` struct was repurposed for `council_vote_threshold`
- For backward compatibility with existing accounts `council_vote_threshold` is set to `community_vote_threshold` during `Governance` accounts deserialisation 
- `VoteThreshold::Disabled` was introduced to explicitly prevent community or council votes on given `Governance` proposals


governance/program/src/error.rs
```diff
@@ -393,6 +393,14 @@ pub enum GovernanceError {
     /// GovernanceConfig change not allowed
     #[error("GovernanceConfig change not allowed")]
     GovernanceConfigChangeNotAllowed,
+
+    /// All vote thresholds can't be disabled
+    #[error("All vote thresholds can't be disabled")]
+    AllVoteThresholdsDisabled,
```
joncinque, line 399: nit: `AllVoteThresholdsDisabled` means something a bit different than `all vote thresholds can't be disabled`, how about `AtLeastOneVoteThresholdRequired`?

governance/program/src/state/enums.rs
```diff
@@ -152,6 +154,15 @@ pub enum VoteThreshold {
     /// Once the quorum is achieved a simple majority (50%+1) of Yes votes is required for the vote to succeed
     /// Note: Quorum is not implemented in the current version
     QuorumPercentage(u8),
+
+    /// Disabled vote threshold indicates the given voting population (community or council) is not allowed to vote
+    /// on proposals for the given Governance
+    Disabled,
+    //
+    // Absolute vote threshold expressed in the voting mint units
+    // It can be implemented once Solana runtime supports accounts resizing to accommodate u64 size extension
```
joncinque, line 163: Should get enabled on mainnet within the next couple of epochs!

governance/program/src/state/enums.rs
```diff
@@ -152,6 +154,15 @@ pub enum VoteThreshold {
     /// Once the quorum is achieved a simple majority (50%+1) of Yes votes is required for the vote to succeed
     /// Note: Quorum is not implemented in the current version
     QuorumPercentage(u8),
+
+    /// Disabled vote threshold indicates the given voting population (community or council) is not allowed to vote
+    /// on proposals for the given Governance
+    Disabled,
```
joncinque, line 160: I'm not sure how sensitive you are to account layout on the frontend, but there might be an issue since this is only size 1 while the others are size 2.  When borsh serializes, it'll pack it in to 1 byte.  It may be worth adding a dummy `u8` value.

governance/program/src/state/governance.rs
```diff
@@ -377,20 +428,181 @@ pub fn assert_valid_create_governance_args(
...
+    }
+
+    Ok(())
+}
+
+/// Asserts the provided vote_threshold is valid
+pub fn assert_is_valid_vote_threshold(vote_threshold: &VoteThreshold) -> Result<(), ProgramError> {
+    match *vote_threshold {
         VoteThreshold::YesVotePercentage(yes_vote_threshold_percentage) => {
             if !(1..=100).contains(&yes_vote_threshold_percentage) {
```
joncinque, line 451: It turns out really well that `0` is invalid!

joncinque (approved):
Nothing to stop this from going forward, mainly small questions, great job!

governance/program/src/error.rs
SebastianBor: done

governance/program/src/state/enums.rs
SebastianBor: It needs a custom serialisation logic and I accounted for that in the client package https://github.com/solana-labs/oyster/blob/fe749f0190f54b5b3a4c6ed5decf6100dfb58a9c/packages/governance-sdk/src/governance/serialisation.ts#L159

governance/program/src/state/governance.rs
SebastianBor: yeah :)

PR #3139, opened 2022-05-04
Token client improvements

CriesofCarrots:
It's not that easy to use spl-token-client with an RpcClient (to point at a test-validator or live cluster RPC).

Changes:
- (First commit is purely cosmetic, and can be ignored in review)
- Use the nonblocking RpcClient instead of the blocking one (we're already async here)
- Arc-wrap the RpcClient in `ProgramRpcClient` so that it's easier to use for multiple calls without `will not live long enough` compile errors
- Change RpcClient `send` impl method from `send_transaction` to `send_and_confirm_transaction` - I think this is the most controversial of the changes. This approach seems the most straightforward to me, but we could instead require the caller to do any confirming they need/want. If so, we probably need to rework some things in client/src/token.rs. For instance, `Token::create_mint()` doesn't return the Output (signature) from `process_ixs()`, so there isn't currently a way for a caller to confirm that signature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5710,7 +5710,7 @@ dependencies = [
 
 [[package]]
 name = "spl-token-client"
-version = "0.0.1"
+version = "0.1.0"
 dependencies = [
  "async-trait",
  "solana-client",
--- a/token/client/Cargo.toml
+++ b/token/client/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2018"
 license = "Apache-2.0"
 name = "spl-token-client"
 repository = "https://github.com/solana-labs/solana-program-library"
-version = "0.0.1"
+version = "0.1.0"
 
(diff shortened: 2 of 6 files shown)


joncinque (approved):
Looks great! Changing to `send_and_confirm_transaction` is more aligned with `process_transaction` on the banks client anyway.  If there's a need to send transactions without confirmation, we can implement a new trait for that with `Output = Signature`, right?  This client is really forward-looking!

CriesofCarrots:
Here's a little example of using this client against a solana-test-validator, btw: https://gist.github.com/CriesofCarrots/0539fec5c5601db36d5b7a0c0ead385a

CriesofCarrots:
CI failure seems due to changes upstream in web3.js/transaction.ts (twotx patch), but I'll fix that before merging this

PR #3148, opened 2022-05-08
Use target_os = "solana" in place of all BPF specific cfg annotations

dmakarov:


joncinque:
What if we added a new `cfg` flag added to all compilations with `cargo build-bpf`, in order to avoid doing an arch check? It could be something like:
```
#[cfg(solana_onchain_program)]
```
 That way, when `sbfv2` is added, programs don't need to redo all the work of adding one more target arch.  what do you think?  Happy to bikeshed on a better name!

dmakarov:
I think “sbf” arch is essentially such a cfg flag. This change also need to be done only once, since we’re migrating from bpf to sbf architecture. Sbfv2 is still going to be target_arch = “sbf”.

joncinque:
I'm more worried about users (and me) getting this wrong in the future, which is why if we had some sort of a `#[cfg(one_thing)]` it would be easier.  If the triple is going to be `sbf-solana-solana`, how about using `target_os = "solana"`?  Or would that just make things more confusing do you think?

dmakarov:
It looks like we don't define target_os feature for sbf and bpf targets in rustc, but instead we have vendor feature and that is set to "solana" both for bpf and sbf architectures, so it seems to be using target_arch = "sbf" is more appropriate and currently the only way to distinguish between the old and new toolchains. I think even if in the future we introduce binary incompatible versions of SBF (like V2 etc) they are not going to be noticeable on the source level. This change is to transition from the old BPF toolchain to the new SBF toolchain.

dmakarov:
I updated the cfg annotations to be target_vendor = "solana" instead. This should work the same with the old and the new toolchains.

joncinque:
Oh very nice! That'll be an easy one to adopt.  Apparently `target_vendor` will be deprecated but still work in the future. If this works for all versions currently, I'm game.  Paging some people that might have opinions about this since we'll want to do the same thing in the monorepo: @CriesofCarrots @mvines @t-nelson

dmakarov:
> If this works for all versions currently, I'm game.

It does.

mvines:
wfm, I'm curious if you considered using `target_os` or `target_family` as an alternative?

dmakarov:
I can add target_os to the bpf and sbf target specifications, but it will be picked up only with the next release of sbf-tools.

mvines:
Cool.  `target_os` feels more natural to me than `target_vendor` in general, although I'm certainly less tuned into what's customary in clang than you are.

joncinque (approved):
Very cool, thanks for making this change

PR #3156, opened 2022-05-11
Governance: Veto vote 

SebastianBor:
#### Summary

This change makes it possible for `Council` to directly veto `Community` proposals. It gives an explicit option for the `Council` to defend DAOs from `Community` governance attacks.

`GovernanceConfig` makes it possible to create a configuration where `Council` can't vote on proposals but still can veto them.

#### Implementation Details
- `council_veto_vote_threshold` was introduced. It defines the required number of `Veto` votes to block a `Proposal` 
- Reserved space from `GovernanceConfig` was used for `council_veto_vote_threshold` and it's defaulted to `council_vote_threshold` for existing configurations. It makes the `Veto` option enabled by default for all existing DAOs
- `veto_vote_weight` on `Proposal` account was changed from `Option<u64>` to `u64` as it's not an optional field. Note: It was replaced with `reserved:u8` and moved to the end of the account struct.

#### Out Of Scope
- The `Veto` vote implementation is symmetrical and either `Council` or `Community` can cast `Veto` vote but reserved space from `GovernanceConfig` would have to be taken for the additional `VoteThreshold` configuration. 
Since the `Community` `Veto` vote is not as important as the `Council` one and should rather be disabled by default the support for it was not included in order to preserve the reserved account space.

### DIFF
--- a/governance/chat/program/src/state.rs
+++ b/governance/chat/program/src/state.rs
@@ -8,7 +8,6 @@ use solana_program::{
 use spl_governance_tools::account::{assert_is_valid_account_of_type, AccountMaxSize};
 
 /// Defines all GovernanceChat accounts types
-#[repr(C)]
 #[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
 pub enum GovernanceChatAccountType {
     /// Default uninitialized account state
--- a/governance/program/src/instruction.rs
+++ b/governance/program/src/instruction.rs
@@ -30,7 +30,6 @@ use solana_program::{
 
 /// Instructions supported by the Governance program
 #[derive(Clone, Debug, PartialEq, BorshDeserialize, BorshSerialize, BorshSchema)]
-#[repr(C)]
 #[allow(clippy::large_enum_variant)]
 pub enum GovernanceInstruction {
     /// Creates Governance Realm account which aggregates governances for given Community Mint and optional Council Mint
--- a/governance/program/src/instruction.rs
+++ b/governance/program/src/instruction.rs
@@ -285,10 +285,10 @@ pub enum GovernanceInstruction {
     ///   1. `[writable]` Governance account
     ///   2. `[writable]` Proposal account
     ///   3. `[writable]` TokenOwnerRecord of the Proposal owner
-    ///   4. `[writable]` TokenOwnerRecord of the voter. PDA seeds: ['governance',realm, governing_token_mint, governing_token_owner]
+    ///   4. `[writable]` TokenOwnerRecord of the voter. PDA seeds: ['governance',realm, vote_governing_token_mint, governing_token_owner]
     ///   5. `[signer]` Governance Authority (Token Owner or Governance Delegate)
     ///   6. `[writable]` Proposal VoteRecord account. PDA seeds: ['governance',proposal,governing_token_owner_record]
-    ///   7. `[]` Voting Token Mint
+    ///   7. `[]` The Governing Token Mint which is used to cast the vote (vote_governing_token_mint)
     ///           The voting token mint is the governing_token_mint of the Proposal for Approve, Deny and Abstain votes
     ///           For Veto vote the voting token mint is the mint of the opposite voting population
     ///           Council mint to veto Community proposals and Community mint to veto Council proposals
@@ -323,9 +323,9 @@ pub enum GovernanceInstruction {
     ///   0. `[]` Realm account
     ///   1. `[]` Governance account
     ///   2. `[writable]` Proposal account
-    ///   3. `[writable]` TokenOwnerRecord account. PDA seeds: ['governance',realm, voting_token_mint, governing_token_owner]
+    ///   3. `[writable]` TokenOwnerRecord account. PDA seeds: ['governance',realm, vote_governing_token_mint, governing_token_owner]
     ///   4. `[writable]` Proposal VoteRecord account. PDA seeds: ['governance',proposal, governing_token_owner_record]
-    ///   5. `[]` Voting Token Mint. The Governing Token Mint which was used to cast the vote
+    ///   5. `[]` The Governing Token Mint which was used to cast the vote (vote_governing_token_mint)
     ///   6. `[signer]` Optional Governance Authority (Token Owner or Governance Delegate)
     ///       It's required only when Proposal is still being voted on
     ///   7. `[writable]` Optional Beneficiary account which would receive lamports when VoteRecord Account is disposed
@@ -1024,7 +1024,7 @@ pub fn cast_vote(
     proposal_owner_record: &Pubkey,
     voter_token_owner_record: &Pubkey,
     governance_authority: &Pubkey,
-    voting_token_mint: &Pubkey,
+    vote_governing_token_mint: &Pubkey,
     payer: &Pubkey,
     voter_weight_record: Option<Pubkey>,
     max_voter_weight_record: Option<Pubkey>,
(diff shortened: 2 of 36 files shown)


governance/program/src/state/proposal.rs
```diff
@@ -465,40 +475,42 @@ impl ProposalV2 {
         account_info_iter: &mut Iter<AccountInfo>,
         realm: &Pubkey,
         realm_data: &RealmV2,
+        vote: Option<&Vote>,
```
joncinque, line 478: I'm not a huge fan of these kinds of parameters, since they can be silently ignored in some cases, ie. a veto on a community mint.  I know this isn't supported currently, but it'll help you in the long run.  I can think of a couple of options:

* Keep this, but don't make it an Option, and check that it makes sense. this will force the `None` calls into something valid, which will be weird in some situations
* Split this into `resolve_max_veto_voter_weight` and `resolve_max_governing_voter_weight`, which don't take the vote parameter.  The name of the second one kind of stinks, but it's the best I could come up with.  You could have them call this internally if you like

governance/program/src/state/proposal.rs
```diff
@@ -414,10 +417,12 @@ impl ProposalV2 {
     fn get_max_voter_weight_from_mint_supply(
         &mut self,
         realm_data: &RealmV2,
+        governing_token_mint: &Pubkey,
         governing_token_mint_supply: u64,
+        vote: Option<&Vote>,
```
joncinque, line 422: Same note here as in `resolve_max_voter_weight`, since it's silently ignored for the council mint, which is extra confusing.  Consider splitting this into `get_max_veto_voter_weight`, and removing the parameter from this one

governance/program/src/state/governance.rs
```diff
@@ -205,21 +209,30 @@ impl GovernanceV2 {
         governing_token_mint: &Pubkey,
     ) -> Result<(), ProgramError> {
         // resolve_vote_threshold() asserts the vote threshold exists for the given governing_token_mint and is not disabled
-        let _ = self.resolve_vote_threshold(realm_data, governing_token_mint)?;
+        let _ = self.resolve_vote_threshold(realm_data, governing_token_mint, None)?;
```
joncinque, line 212: The `None` here looks like a potential bug because of the optional vote parameter.  It makes it seem like you're not checking that a veto is proper

governance/program/src/state/governance.rs
```diff
@@ -205,21 +209,30 @@ impl GovernanceV2 {
...
         Ok(())
     }
 
-    /// Resolves VoteThreshold for the given realm and governing token
+    /// Resolves VoteThreshold for the given realm, governing token and optional Vote
     pub fn resolve_vote_threshold(
         &self,
         realm_data: &RealmV2,
         governing_token_mint: &Pubkey,
+        vote: Option<&Vote>,
```
joncinque, line 222: Please split this one up as well or make the parameter mandatory, as noted above.

governance/program/src/state/governance.rs
```diff
@@ -43,15 +43,19 @@ pub struct GovernanceConfig {
...
-    pub reserved: [u8; 2],
+    /// The threshold for Council Veto votes
+    pub council_veto_vote_threshold: VoteThreshold,
 
     /// Minimum council weight a governance token owner must possess to be able to create a proposal
     pub min_council_weight_to_create_proposal: u64,
+    //
+    // The threshold for Community Veto votes
+    // Note: Community Veto vote is not supported in the current version
+    // In order to use this threshold the space from GovernanceV2.reserved must be taken to expand GovernanceConfig size
```
joncinque, line 54: We're finally live for realloc on mainnet!

governance/program/src/state/realm.rs
```diff
@@ -177,6 +175,34 @@ impl RealmV2 {
...
+        voting_token_mint: &Pubkey,
+    ) -> Result<Pubkey, ProgramError> {
+        if *vote != Vote::Veto {
+            return Ok(*voting_token_mint);
+        }
+
+        // When Community veto Council proposal then return council_token_mint as the Proposal governing_token_mint
+        if self.community_mint == *voting_token_mint {
+            return Ok(self.config.council_mint.unwrap());
+        }
```
joncinque, line 196: But this isn't supported yet, correct?

joncinque (commented):
Looks really good! The main substance is ready to go, just a few small points about helping readability and avoiding some coding footguns in the future.

governance/chat/program/src/state.rs
```diff
@@ -8,7 +8,6 @@ use solana_program::{
 use spl_governance_tools::account::{assert_is_valid_account_of_type, AccountMaxSize};
 
 /// Defines all GovernanceChat accounts types
-#[repr(C)]
```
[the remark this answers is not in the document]
SebastianBor: Yes, my understanding is it's redundant and unnecessary option

governance/program/src/instruction.rs
```diff
@@ -289,7 +288,11 @@ pub enum GovernanceInstruction {
     ///   4. `[writable]` TokenOwnerRecord of the voter. PDA seeds: ['governance',realm, governing_token_mint, governing_token_owner]
```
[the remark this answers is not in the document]
SebastianBor: Updated

governance/program/tests/use_veto_vote.rs
```diff
@@ -0,0 +1,653 @@
...
+    // Assert
+    let proposal_account = governance_test
+        .get_proposal_account(&proposal_cookie.address)
+        .await;
+
+    assert_eq!(proposal_account.state, ProposalState::Vetoed);
+}
+
+// TODO: Once Veto for Community or plugin support for Council is implemented write Veto tests with plugin
+// The tests should cover scenarios where Veto voter_weight and/or max_voter_weight is resolved using the plugins
```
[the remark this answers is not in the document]
SebastianBor: Done

governance/program/src/state/proposal.rs
```diff
@@ -200,6 +199,9 @@ pub struct ProposalV2 {
 
     /// Link to proposal's description
     pub description_link: String,
+
+    /// The total weight of Veto votes
+    pub veto_vote_weight: u64,
```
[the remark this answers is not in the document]
SebastianBor: The reason I moved it to the end of the struct and replaced with `reserved1: u8` was because it was unused  `Option<u64>` before. It means it was always serialised as `0:u8` and I couldn't change it to `u64` because it wouldn't be able to deserialise existing accounts where it would always wanted to use 8 consecutive bytes.


The only risk with this was if the actual serialised size of  `Proposal` account could ever be shrank because I don't think we zero the reaming empty account space, so we could end up  with `veto_vote_weight` being read from some leftovers. But the shrinking can never happen in the current design

governance/program/src/state/proposal.rs
SebastianBor: Agree, I struggled with that myself. 
I fixed it by introducing `VoteKind` enum which indicates whether it's an `Electorate` vote (`Approve`, `Deny`, `Abstain`) or `Veto` vote for the opposite voting population

governance/program/src/state/governance.rs
SebastianBor: Fixed with `VoteKind`

governance/program/src/state/proposal.rs
SebastianBor: Fixed with `VoteKind`

governance/program/src/state/governance.rs
SebastianBor: Fixed with `VoteKind`

governance/program/src/state/realm.rs
SebastianBor: Correct. This code was unreachable but better to throw an error here. I fixed it.

governance/program/src/state/proposal.rs
```diff
@@ -627,6 +644,23 @@ impl ProposalV2 {
         None
     }
 
+    /// Checks if vote can be tipped and transitioned to Vetoed state
+    /// If yes then Some(ProposalState::Vetoed) is returned and None otherwise
+    #[allow(clippy::float_cmp)]
```
joncinque, line 649: nit: are these needed? I don't see any float comparisons in these functions

joncinque (dismissed):
Looks good! Just one tiny question that shouldn't stop this

governance/program/src/state/proposal.rs
SebastianBor: Good spot! removed

PR #3166, opened 2022-05-13
token-cli: produce error if invalid --config option

### LINKED ISSUE #3159: token-cli: don't silently ignore invalid `--config` option
`spl-token` CLI commands accept a `-C`/`--config` flag which specifies the configuration file to use:

```    -C, --config <PATH>
            Configuration file to use [default: ~/.config/solana/cli/config.yml]
```

# Expected Behavior:

If the passed in configuration file does not exist, I expect the command to fail immediately or at least warn me.

# Observed Behavior

If you pass in a path to a file that does not exist, the flag is silently ignored and the command may succeed without error.

I have verified this with the `mint` sub-command, though all other sub-commands probably suffer the same issue.

dboures:
If a nonexistent config file is provided, the CLI will now fail and display an error.

closes #3159


joncinque:
This looks really great, thanks for the contribution!  Also, thanks for adding that `assert_cmd` crate, we should change up all the CLI tests to use that instead.  It'll make debugging test failures much easier.

PR #3176, opened 2022-05-17
Bump solana to v1.10.15

dmakarov:


dmakarov:
@joncinque is it ok to bump Solana version? I'd like to use updated bpf-tools that are available in v1.10.15 of Solana.

joncinque (approved):
Absolutely, looks good, thanks!

joncinque:
Ah sorry, this is my fault.  I forgot that you should also at least run `cargo check` to update the lockfile.  Would you mind running a PR to update the lockfile?

dmakarov:
Will do it now. No worries. I'm sorry, I didn't think about updating the lock files.

PR #3178, opened 2022-05-17
token-2022: Add support for Non-Transferable Tokens (NTTs) - NonTransferableMint extension

mvines:
Rebased version of @joxoby's work from #2912


mvines:
Thanks for the contribution @joxoby!

PR #3182, opened 2022-05-19
token-swap-js: Bump buffer-layout and web3.js

joncinque:
#### Problem

People want updated dependencies in token-swap-js.

#### Solution

Update those dependencies!

cc @JamesMDunn


joncinque:
Do you also need a new version published?

JamesMDunn:
@joncinque yeah can you publish a new version? Thanks!

PR #3185, opened 2022-05-19
token-swap-js: Bump package version for release

joncinque:
#### Problem

@JamesMDunn wants a new version of token-swap-js.

#### Solution

Bump the version for the release!


joncinque:
Enjoy! https://www.npmjs.com/package/@solana/spl-token-swap

PR #3186, opened 2022-05-20
token-js: move start-server-and-test dep to devDependencies

beeman:
This patch moves the `start-server-and-test` dependencies to devDependencies which should shave off a few deps when using this library :)


joncinque:
Thanks for the contribution!

PR #3187, opened 2022-05-20
Confusion with decimals

S1nus:
If decimals for the mint are set to "9", then to mint 100 tokens, you'd need to enter 100000000000

unless I am mistaken?


joncinque:
You are absolutely right, thanks for the contribution!

PR #3195, opened 2022-05-24
bump solana to v1.10.19

samkim-crypto:
Updating solana to v1.10.19.


joncinque:
@samkim-crypto just wondering, how did you make the changes for this?  When upgrading to 1.10.29 in #3303 I noticed that not all of them were updated after running `./update-solana-dependencies.sh 1.10.29`, because some of them were left behind to 1.10.15 during this PR.  Did something go wrong with the script?

PR #3208, opened 2022-06-01
token-2022: limit incoming transfers for confidential transfer extension

samkim-crypto:
Decrypting ElGamal ciphertexts that encrypt large numbers (> 2^32) is computationally very expensive. When a confidential transfer extended account receives incoming transfers of large amounts that are accumulated into the `pending_balance` ciphertext, then it may become very expensive for users to decrypt and recover the exact amount of tokens that is held as `pending_balance`. 

To cope with this, a confidential transfer instruction is structured such that the transfer amount (a 64-bit number) is encrypted as two ciphertexts: a `lo` ciphertext that encrypts the low 32-bits of the transfer amount and a `hi` ciphertext that encrypts the high 32-bits. Since each of the `lo` and `hi` ciphertexts encrypt 32-bit numbers, they can be efficiently decrypted. Then to recover an amount that is held in the `pending_balance` ciphertext in an account, users can look up the relevant incoming transactions data, decrypt the `lo` and `hi` ciphertexts, and then update the account data (i.e. `decryptable_ciphertext`). Although this solution could be made to work, searching for and decrypting each of the relevant incoming transfer data is cumbersome and difficult to work with on the client. 

An alternative approach is to break up the `pending_balance` into two ciphertexts:
- `pending_balance_lo` that encrypts the low 32-bits of the pending balance
- `pending_balance_hi` that encrypts the high 32-bits of the pending balance
As long as these pending balance ciphertexts are guaranteed to encrypt 32-bit values, users can easily look up and decrypt the pending balance from accounts.

The main problem with having two pending balance ciphertexts is that they can overflow. If an account receives a large number of incoming transactions, then the amount that `pending_balance_lo` encrypts could grow to be larger than a 32-bit number, in which case, the ciphertext becomes hard to decrypt.

One way to cope with the overflow problem is to introduce the following changes:
1. Introduce a 2^48 [upper bound](https://github.com/solana-labs/solana-program-library/pull/3208/files#diff-56624b0356d2f9104ef0926d6f8e00090a52ba82e0c1da2ed376d3b69da05e67R15) on the transfer amount for a confidential transfer.
2. Introduce an [upper bound](https://github.com/solana-labs/solana-program-library/pull/3208/files#diff-56624b0356d2f9104ef0926d6f8e00090a52ba82e0c1da2ed376d3b69da05e67R111) (e.g. 2^16) on the number of incoming transfers that a single account can accept. Any additional incoming transfers or deposits to an account are rejected before the owner of the account submits an `ApplyPendingBalance`.

With a 2^48 upper bound, a transfer amount must be encrypted as two ciphertexts in a transfer instruction such that:
- the `lo` ciphertext encrypts the low 16-bits of the transfer amount
- the `hi` ciphertext encrypts the high 32-bits of the transfer amount
and similarly, `pending_balance_lo` and `pending_balance_hi` ciphertexts encrypt 16 and 32 bit amounts respectively.

With an upper bound of 2^16 incoming transfers to an account, the `pending_balance_lo` will overflow to  at most 2^32, which is still efficiently decryptable. `pending_balance_hi` can also overflow to encrypt numbers larger up to 2^48. However, for overflows to occur in `pending_balance_hi`, an account must receive a high volume of large transactions. Users that expect large transactions can either submit `ApplyPendingBalance` more often to flush out pending balance ciphertexts or they can maintain multiple accounts to divide up the load in receiving large transactions. Furthermore, with enough time and/or hardware, moderate amounts of overflows (encrypted amount <2^42) in `pending_balance_hi` can still be decrypted. Users that typically deal with high volume of large transactions may have access to more advanced hardware.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -51,7 +51,7 @@ pub enum TokenError {
     AccountInvalidMint,
     #[error("proof error: {0}")]
     Proof(ProofError),
-    #[error("illegal amount")]
+    #[error("maximum deposit transfer amount exceeded")]
     MaximumDepositTransferAmountExceeded,
 }
 impl PartialEq for TokenError {
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -52,7 +52,7 @@ pub enum TokenError {
     #[error("proof error: {0}")]
     Proof(ProofError),
     #[error("illegal amount")]
-    IllegalAmount,
+    MaximumDepositTransferAmountExceeded,
 }
 impl PartialEq for TokenError {
     fn eq(&self, other: &Self) -> bool {
@@ -1074,7 +1074,7 @@ where
         decimals: u8,
     ) -> TokenResult<T::Output> {
         if amount >> confidential_transfer::MAXIMUM_DEPOSIT_TRANSFER_AMOUNT_BIT_LENGTH != 0 {
-            return Err(TokenError::IllegalAmount);
+            return Err(TokenError::MaximumDepositTransferAmountExceeded);
         }
 
         self.process_ixs(
@@ -1149,7 +1149,7 @@ where
         new_source_decryptable_available_balance: AeCiphertext,
     ) -> TokenResult<T::Output> {
         if amount >> confidential_transfer::MAXIMUM_DEPOSIT_TRANSFER_AMOUNT_BIT_LENGTH != 0 {
-            return Err(TokenError::IllegalAmount);
+            return Err(TokenError::MaximumDepositTransferAmountExceeded);
         }
 
         let source_state = self.get_account_info(source_token_account).await.unwrap();
@@ -1213,7 +1213,7 @@ where
         epoch_info: &EpochInfo,
     ) -> TokenResult<T::Output> {
         if amount >> confidential_transfer::MAXIMUM_DEPOSIT_TRANSFER_AMOUNT_BIT_LENGTH != 0 {
-            return Err(TokenError::IllegalAmount);
+            return Err(TokenError::MaximumDepositTransferAmountExceeded);
         }
 
         let source_state = self.get_account_info(source_token_account).await.unwrap();
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -51,6 +51,8 @@ pub enum TokenError {
     AccountInvalidMint,
     #[error("proof error: {0}")]
     Proof(ProofError),
+    #[error("illegal amount")]
+    IllegalAmount,
 }
 impl PartialEq for TokenError {
     fn eq(&self, other: &Self) -> bool {
@@ -1071,6 +1073,10 @@ where
         amount: u64,
         decimals: u8,
     ) -> TokenResult<T::Output> {
+        if amount >> confidential_transfer::MAXIMUM_DEPOSIT_TRANSFER_AMOUNT_BIT_LENGTH != 0 {
+            return Err(TokenError::IllegalAmount);
+        }
+
         self.process_ixs(
             &[confidential_transfer::instruction::deposit(
                 &self.program_id,
@@ -1142,6 +1148,10 @@ where
         source_elgamal_keypair: &ElGamalKeypair,
         new_source_decryptable_available_balance: AeCiphertext,
     ) -> TokenResult<T::Output> {
+        if amount >> confidential_transfer::MAXIMUM_DEPOSIT_TRANSFER_AMOUNT_BIT_LENGTH != 0 {
+            return Err(TokenError::IllegalAmount);
+        }
+
         let source_state = self.get_account_info(source_token_account).await.unwrap();
         let source_extension =
             source_state.get_extension::<confidential_transfer::ConfidentialTransferAccount>()?;
@@ -1202,6 +1212,10 @@ where
         new_source_decryptable_available_balance: AeCiphertext,
         epoch_info: &EpochInfo,
     ) -> TokenResult<T::Output> {
+        if amount >> confidential_transfer::MAXIMUM_DEPOSIT_TRANSFER_AMOUNT_BIT_LENGTH != 0 {
+            return Err(TokenError::IllegalAmount);
+        }
+
         let source_state = self.get_account_info(source_token_account).await.unwrap();
         let source_extension =
             source_state.get_extension::<confidential_transfer::ConfidentialTransferAccount>()?;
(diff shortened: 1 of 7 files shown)


token/client/src/token.rs
```diff
@@ -51,6 +51,8 @@ pub enum TokenError {
     AccountInvalidMint,
     #[error("proof error: {0}")]
     Proof(ProofError),
+    #[error("illegal amount")]
+    IllegalAmount,
```
mvines, line 55: Perhaps an error that's more precise here, as the amount isn't really "illegal". `MaximumDepositTransferAmountExceeded`?

samkim-crypto: Yes, that is a more decryptive and much better name indeed.

samkim-crypto:
@mvines I updated the PR with more background on the changes. We did have a discussion thread on slack some time ago on this, but it seems like it is gone now. When you get a chance, it would be nice if you can skim over the changes to see if it looks reasonable.

mvines (dismissed):
lgtm overall!

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -330,9 +339,23 @@ fn process_deposit(
             return Err(TokenError::ConfidentialTransferDepositsAndTransfersDisabled.into());
         }
 
-        destination_confidential_transfer_account.pending_balance = ops::add_to(
-            &destination_confidential_transfer_account.pending_balance,
-            amount,
+        if destination_confidential_transfer_account.pending_balance_credit_counter
+            == destination_confidential_transfer_account.maximum_pending_balance_credit_counter
```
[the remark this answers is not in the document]
samkim-crypto: Yep, agreed.

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -700,9 +723,20 @@ fn process_destination_for_transfer(
...
+    }
+
+    let new_destination_pending_balance_lo = ops::add(
+        &destination_confidential_transfer_account.pending_balance_lo,
         destination_ciphertext_lo,
+    )
+    .ok_or(ProgramError::InvalidInstructionData)?;
+
+    let new_destination_pending_balance_hi = ops::add(
+        &destination_confidential_transfer_account.pending_balance_hi,
```
joncinque, line 739: This may be a silly question, but since you only do the validation that the transfer amount fits in 48 bytes on the client, how are you sure that it fits over here on the program side?

joncinque (approved):
Just a few nits and a question -- also, I second what Michael said

token/client/src/token.rs
```diff
@@ -51,6 +51,8 @@ pub enum TokenError {
     AccountInvalidMint,
     #[error("proof error: {0}")]
     Proof(ProofError),
+    #[error("illegal amount")]
```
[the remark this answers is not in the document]
samkim-crypto: Thanks!

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: Yep that's a valid question. The range is checked via zkp's. The proof program will check the validity of the range proof that is attached with a transfer instruction, which will guarantee that the transfer amount is 48 bits. This update to zkp generation should already be checked into the zk-token-sdk.

joncinque (dismissed):

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -157,7 +160,8 @@ fn process_configure_account(
 
         This should just be encoded as [0; 64]
     */
-    confidential_transfer_account.pending_balance = EncryptedBalance::zeroed();
+    confidential_transfer_account.pending_balance_lo = EncryptedBalance::zeroed();
+    confidential_transfer_account.pending_balance_hi = EncryptedBalance::zeroed();
```
CriesofCarrots, line 164: Update comment preceding this?

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -330,9 +339,25 @@ fn process_deposit(
             return Err(TokenError::ConfidentialTransferDepositsAndTransfersDisabled.into());
         }
 
-        destination_confidential_transfer_account.pending_balance = ops::add_to(
-            &destination_confidential_transfer_account.pending_balance,
-            amount,
+        if u64::from(destination_confidential_transfer_account.pending_balance_credit_counter)
+            >= u64::from(
```
CriesofCarrots, line 343:
```suggestion
            > u64::from(
```
nit, but to me `maximum_pending_balance_credit_counter` implies the max that is acceptable

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -700,9 +725,22 @@ fn process_destination_for_transfer(
         return Err(TokenError::ConfidentialTransferElGamalPubkeyMismatch.into());
     }
 
-    let new_destination_pending_balance = ops::add_with_lo_hi(
-        &destination_confidential_transfer_account.pending_balance,
+    if u64::from(destination_confidential_transfer_account.pending_balance_credit_counter)
+        >= u64::from(
```
CriesofCarrots, line 729: Ditto nit above

CriesofCarrots (commented):
Just nits. Code changes look correct to me.
Would be nice to have tests that cover the max-transfer error case, as well as `pending_balance_hi` population

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: Thanks! Updated now.

token/program-2022/src/extension/confidential_transfer/processor.rs
CriesofCarrots: Fyi, this one doesn't look changed

token/program-2022/src/extension/confidential_transfer/processor.rs
CriesofCarrots: I don't see this change either

samkim-crypto: Ah terribly sorry. The changes should be pushed now.

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: Should be updated 🙏

samkim-crypto:
I pushed all the unpushed changes and the tests have also been updated to check that (1) deposit or transfer fails if max bound on pending credit is reached and (2) the `pending_hi` ciphertexts are populated on deposits.

CriesofCarrots (approved):

PR #3216, opened 2022-06-03
token-2022: prevent an already configured confidential account to be configured again

samkim-crypto:
The `confidential_transfer::ConfigureAccount`, should fail if run on an already configured account as stated in the [description](https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/src/extension/confidential_transfer/instruction.rs#L62-L63). However, this is currently not enforced in the processor logic, allowing users to overwrite the account state.

This PR adds a check in the `ConfigureAccount` processor logic to prevent accounts from being configured again.

Thank you @joncinque for the catch!

### DIFF
--- a/token/program-2022/src/extension/confidential_transfer/processor.rs
+++ b/token/program-2022/src/extension/confidential_transfer/processor.rs
@@ -54,7 +54,7 @@ fn process_initialize_mint(
     check_program_account(mint_info.owner)?;
     let mint_data = &mut mint_info.data.borrow_mut();
     let mut mint = StateWithExtensionsMut::<Mint>::unpack_uninitialized(mint_data)?;
-    *mint.init_extension::<ConfidentialTransferMint>()? = *confidential_transfer_mint;
+    *mint.init_extension::<ConfidentialTransferMint>(false)? = *confidential_transfer_mint;
 
     Ok(())
 }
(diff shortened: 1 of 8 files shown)


token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -122,6 +122,15 @@ fn process_configure_account(
...
+    if token_account
+        .get_extension::<ConfidentialTransferAccount>()
+        .is_ok()
+    {
+        return Err(TokenError::ExtensionAlreadyInitialized.into());
+    }
+
     // Note: The caller is expected to use the `Reallocate` instruction to ensure there is
     // sufficient room in their token account for the new `ConfidentialTransferAccount` extension
     let mut confidential_transfer_account =
```
joncinque, line 136: nit: what do you think about adding a flag to `init_extension` to fail if the extension already exists? because otherwise this behavior will be a footgun for any extension that's initialized after the account

samkim-crypto: Yes, I was actually just about to ask you about that. `get_extension` also makes unnecessary copies if we just care about whether an extension is initialized or not. I think adding a flag to `init_extension` is a better and more natural way.

joncinque: Ok great! I hope there won't be too much churn because of the change :crossed_fingers:

samkim-crypto:
Added the `overwrite` flag on `initialize_extension`. If the flag is set, then `initialize_extension` will overwrite an existing extension (if it exists) with the default extension state. If the flag is not set, then `initialize_extension` will return an error if the extension was already initialized.

I ended up just dividing the logic in `init_or_get_extension` to `get_extension` and `init_extension` since the `overwrite` flag felt unnatural in `init_or_get_extension`. I hope this change does not break the original readability of the code.

samkim-crypto:
One alternative to having a boolean flag `overwrite` is to have two separate functions `init_extension_overwrite` and `init_extension` for better readability.

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -54,7 +54,7 @@ fn process_initialize_mint(
     check_program_account(mint_info.owner)?;
     let mint_data = &mut mint_info.data.borrow_mut();
     let mut mint = StateWithExtensionsMut::<Mint>::unpack_uninitialized(mint_data)?;
-    *mint.init_extension::<ConfidentialTransferMint>()? = *confidential_transfer_mint;
+    *mint.init_extension::<ConfidentialTransferMint>(false)? = *confidential_transfer_mint;
```
joncinque, line 57: Since this is being done on an uninitialized mint, it should be ok to overwrite, no?

joncinque (dismissed):
Looks good to me, but let's see what @CriesofCarrots thinks too, since I think we used to have `init` and `get` separated, but there was a good reason to unify them.

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: Right, good catch!

CriesofCarrots (approved):
Lgtm, I can't remember any specific reason for the old api.

I do find the bool a little unintuitive to read, but I don't want 2 different functions, and can't really think of a reasonable alternative, so let's go with it

samkim-crypto:
Great! Thanks for the review.

PR #3221, opened 2022-06-06
feat: update `solana airdrop` amount

kootsZhin:
`solana airdrop` request is capped at 2 now, exceeding the cap will lead to failed transactions.


joncinque:
Thanks for the contribution! It looks like this PR also deletes almost all of the repo, so could you reduce the changes to just the docs changes?

PR #3229, opened 2022-06-08
Support InitializeAccount2 instruction

kunaljaydesai:


joncinque:
Thanks for your contribution! Could you add a little test similar to https://github.com/solana-labs/solana-program-library/blob/5b1b87ed484b94a5f5d30f2fd733c29f34816f5a/token/js/test/unit/index.test.ts#L35 to create the instruction and check that it matches expectations? Then we can merge this quickly

kunaljaydesai:
added!

joncinque:
We're good to go, thanks again!

PR #3236, opened 2022-06-10
spl-token: Enums for Cli command names

### LINKED ISSUE #2598: [spl-token] add enum for cli command names
problem: currently, all cli commands use strings like "create-token" which are hardcoded at different places. difficult to maintain

fix: introduce command name enum that can be turned into string

cc: @CriesofCarrots

atharmohammad:
Fixes #2598 
Added Enums for Command Names and replaced hard coded command names
@joncinque

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -49,7 +49,10 @@ use spl_token::{
     native_mint,
     state::{Account, Mint, Multisig},
 };
-use std::{collections::HashMap, fmt::Display, process::exit, str::FromStr, sync::Arc};
+use std::{
+    collections::HashMap, fmt::Display, process::exit, str::FromStr, string::ToString, sync::Arc,
+};
+use strum_macros::{EnumString, IntoStaticStr, ToString};
 
 mod config;
 use config::Config;
@@ -100,7 +103,33 @@ pub const MULTISIG_SIGNER_ARG: ArgConstant<'static> = ArgConstant {
     help: "Member signer of a multisig account",
 };
 
-pub const CREATE_TOKEN: &str = "create-token";
+#[derive(Debug, Clone, Copy, PartialEq, EnumString, IntoStaticStr, ToString)]
+#[strum(serialize_all = "kebab-case")]
+pub enum CommandName {
+    CreateToken,
+    Close,
+    Bench,
+    CreateAccount,
+    CreateMultisig,
+    Authorize,
+    Transfer,
+    Burn,
+    Mint,
+    Freeze,
+    Thaw,
+    Wrap,
+    Unwrap,
+    Approve,
+    Revoke,
+    Balance,
+    Supply,
+    Accounts,
+    Address,
+    AccountInfo,
+    MultisigInfo,
+    Gc,
+    SyncNative,
+}
 
 pub fn owner_address_arg<'a, 'b>() -> Arg<'a, 'b> {
     Arg::with_name(OWNER_ADDRESS_ARG.name)
@@ -334,11 +363,11 @@ fn command_create_token(
                 decimals,
                 transaction_data: cli_signature,
             },
-            CREATE_TOKEN,
+            &CommandName::CreateToken,
             config,
         ),
         TransactionReturnData::CliSignOnlyData(cli_sign_only_data) => {
-            format_output(cli_sign_only_data, CREATE_TOKEN, config)
+            format_output(cli_sign_only_data, &CommandName::CreateToken, config)
         }
     })
 }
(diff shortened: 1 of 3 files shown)


token/cli/src/main.rs
```diff
@@ -2535,6 +2568,7 @@ fn main() -> Result<(), Error> {
     let mut multisigner_ids = Vec::new();
 
     let (sub_command, sub_matches) = app_matches.subcommand();
+    let sub_command_ = CommandName::from_str(sub_command).unwrap();
```
joncinque, line 2571: nit: call this `sub_command`

token/cli/src/main.rs
```diff
@@ -100,7 +101,39 @@ pub const MULTISIG_SIGNER_ARG: ArgConstant<'static> = ArgConstant {
...
+    Approve,
+    Revoke,
+    Balance,
+    Supply,
+    Accounts,
+    Address,
+    AccountInfo,
+    MultisigInfo,
+    Gc,
+    SyncNative
```
joncinque, line 130: Eventually, the idea for these is to be completely expressive with the values required for the command encoded as well.

For example, check out the CLI commands in the main repo: https://github.com/solana-labs/solana/blob/c419845cfeb1c7ab9ca959de66e90b9dacbc25b7/cli/src/cli.rs#L47

But we can definitely start here :smile:

token/cli/src/main.rs
```diff
@@ -100,7 +101,39 @@ pub const MULTISIG_SIGNER_ARG: ArgConstant<'static> = ArgConstant {
...
+}
+
+pub fn to_str(command:CommandName) -> &'static str {
+    let command_str : &'static str = command.into();
+    command_str
+}
```
joncinque, line 136: Let's remove this function and just call `.into()` on the `CommandName` directly

token/cli/src/main.rs
```diff
@@ -3050,7 +3084,7 @@ fn process_command(
                 bulk_signers,
             )
         }
-        ("sync-native", arg_matches) => {
+        (CommandName::SyncNative, arg_matches) => {
             let address = config.associated_token_address_for_token_or_override(
                 arg_matches,
                 "address",
```
joncinque, line 3090: I can't comment lower, but can you remove the final case?
```
_ => unreachable!(),
```
With this change, we can be explicit!

joncinque (commented):
Great work!  A few small points, then we should be able to get this integrated

token/cli/src/main.rs
```diff
@@ -3264,19 +3291,24 @@ mod tests {
             &multisig_member_help,
         )
         .get_matches_from(args);
-        let (sub_command, sub_matches) = app_matches.subcommand();
+        let (sub_command_, sub_matches) = app_matches.subcommand();
+        let sub_command = CommandName::from_str(sub_command_).unwrap();
```
joncinque, line 3295: nittiest of nits: in Rust, you can rebind an old variable name, so no need to have the underscore.  You can do
```
 let (sub_command, sub_matches) = app_matches.subcommand();
let sub_command = CommandName::from_str(sub_command).unwrap();
```

token/cli/src/main.rs
```diff
@@ -2534,7 +2561,8 @@ fn main() -> Result<(), Error> {
     let mut bulk_signers: Vec<Box<dyn Signer>> = Vec::new();
     let mut multisigner_ids = Vec::new();
 
-    let (sub_command, sub_matches) = app_matches.subcommand();
+    let (sub_command_, sub_matches) = app_matches.subcommand();
+    let sub_command = CommandName::from_str(sub_command_).unwrap();
```
joncinque, line 2565: same here

token/cli/src/main.rs
```diff
@@ -3060,16 +3088,15 @@ fn process_command(
...
 }
 
-fn format_output<T>(command_output: T, command_name: &str, config: &Config) -> String
+fn format_output<T>(command_output: T, command_name: &CommandName, config: &Config) -> String
 where
     T: Serialize + Display + QuietDisplay + VerboseDisplay,
 {
     config.output_format.formatted_string(&CommandOutput {
-        command_name: String::from(command_name),
+        command_name: <String as From<&str>>::from(command_name.into()),
```
joncinque, line 3099: nit: how about:
```suggestion
        command_name: command_name.to_string()
```
Should be possible according to https://docs.rs/strum/latest/strum/derive.ToString.html

joncinque (commented):
Just two little nits, then this is good to go!

token/cli/src/main.rs
atharmohammad: Didn't knew this, cool :)

joncinque:
The changes look good, so now we'll have to get it through CI :smile: 

First, can you reformat the code by running `cargo fmt --all` at the root of the repo?

Second, can you rebase your branch onto master? We had to bump a crate in order to pass the cargo audit as part of CI.

atharmohammad:
> The changes look good, so now we'll have to get it through CI smile
> 
> First, can you reformat the code by running `cargo fmt --all` at the root of the repo?
> 
> Second, can you rebase your branch onto master? We had to bump a crate in order to pass the cargo audit as part of CI.
Done the formatting part, was confused why does do-audit.sh fails thanks for letting me know

joncinque:
Great work! Thanks for the contribution

Issue #3163, opened 2022-05-12
math: Add fixed point approximation for natural log


joncinque:
#### Problem

Natural logs are not possible to calculate in fixed-point terms.

#### Solution

Add an approximation for `ln` using a Taylor approximation (probably only possible with `PreciseNumber` around 1), or using linear interpolation.

zrthxn:
I'd like to work on this

joncinque:
Great, feel free to submit a draft pull request if you want to discuss anything

PR #3240, opened 2022-06-11
Approximation of Natural Log

zrthxn:
## Description
Added an approximation for `ln` using a Taylor series approximation 
Closes: #3163 

Ideas were taken from https://math.stackexchange.com/a/977836
The following form of the Taylor series expansion of $ln(x)$ was used
$$ ln(x) = (n-1)*ln(10) + 2 * \sum{ y^{2i+1} \over 2i+1 } $$
$$ \forall i \epsilon 0, 1, 2, ... p $$
where $y= {A−1 \over A+1}$, such that $x = A ×10^{n−1}$ and $0≤A<10$
Checking for convergence was not an issue because this form of the Taylor series is known to converge quickly.

## Code
The tests for this still need to be written properly, although the function does perform satisfactorily.
The traits for arguments might be incorrect but I can change those if I'm told what sorts of inputs need to be taken.

## Results
Here I've tabulated the performance of the function with increasing inputs
| Input | Approx. | Actual | Error |
| ----- | ------- | ------ | ----- |
| 0.56276 | -0.5749020294731245 | -0.5749020294731245 | 0 |
| 0.35213 | -1.0437548533833947 | -1.043754853383395 | 0.0000000000000002220446049250313 |
| 0.6 | -0.5108256237659905 | -0.5108256237659907 | 0.0000000000000002220446049250313 |
| 1 | 0 | 0 | 0 |
| 3.5242 | 1.2596534601524871 | 1.2596534601524931 | -0.000000000000005995204332975845 |
| 12.487 | 2.5246881351084447 | 2.524688103133008 | 0.000000031975436520781386 |
| 3248.152436 | 8.085841728372603 | 8.085841632446295 | 0.00000009592630867416574 |
| 21348.1328491 | 9.968719688359453 | 9.968719560457707 | 0.00000012790174608312554 |
| 542436.43243 | 13.203826818605346 | 13.203826182286956 | 0.0000006363183899793512 |

The error slowly increases as the input becomes larger.

### DIFF
--- a/libraries/math/src/approximations.rs
+++ b/libraries/math/src/approximations.rs
@@ -49,7 +49,7 @@ pub fn sqrt<T: PrimInt + CheckedShl + CheckedShr>(radicand: T) -> Option<T> {
 /// $$ ln(x) = (n-1)*ln(10) + 2 * \sum{ y^{2k+1} \over 2k+1 } $$
 /// where x = A * 10^(n-1) such that 0 <= A < 10
 /// and y = (A-1)/(A+1)
-pub fn ln<T: Float + Copy>(input: T) -> Option<T> {
+pub fn ln<T: Float + Clone>(input: T) -> Option<T> {
     if input.le(&T::zero()) {
         return None; // fail for less than or equal to 0
     }
--- a/libraries/math/src/approximations.rs
+++ b/libraries/math/src/approximations.rs
@@ -1,11 +1,7 @@
 //! Approximation calculations
 
-use std::fmt::Display;
-
-use num_traits::{Float, Unsigned};
-
 use {
-    num_traits::{CheckedShl, CheckedShr, PrimInt},
+    num_traits::{CheckedShl, CheckedShr, PrimInt, Float},
     std::cmp::Ordering,
 };
 
@@ -45,10 +41,6 @@ pub fn sqrt<T: PrimInt + CheckedShl + CheckedShr>(radicand: T) -> Option<T> {
     Some(result)
 }
 
-/// Accurate value of natural log of 10.
-/// Used in calculating approximations of ln(x).
-pub const LN10: f32 = 2.302585092994046;
-
 /// Calculate approximate natural log of a number
 /// using Taylor series of Log_e(x)
 ///     
@@ -57,34 +49,34 @@ pub const LN10: f32 = 2.302585092994046;
 /// $$ ln(x) = 2 * \sum{ y^{2k+1} \over 2k+1 } $$
 /// where y = (A-1)/(A+1) and x = A * 10^(n-1) 
 /// such that 0 <= A < 10
-pub fn ln<T: Float>(input: T) -> Option<T> {
+pub fn ln<T: Float + Copy>(input: T) -> Option<T> {
     if input.le(&T::zero()) {
         return None; // fail for less than or equal to 0
     }
 
     // x = A * 10^(n-1)
     let mut mantissa = input.clone();
-    while mantissa.floor() > T::from(10 as f32).unwrap() {
-        mantissa = mantissa.div(T::from(10 as f32).unwrap());
+    while mantissa.floor() > T::from(10_f64).unwrap() {
+        mantissa = mantissa.div(T::from(10_f64).unwrap());
     }
     
     // number of digits of input before decimal
     let mut temp_mantx = mantissa.clone();
     let mut n = 1;
     while temp_mantx.floor() != input.floor() {
-        temp_mantx = temp_mantx.mul(T::from(10 as f32).unwrap());
+        temp_mantx = temp_mantx.mul(T::from(10_f64).unwrap());
         n += 1;
     };
 
-    let y: T = (mantissa - T::from(1 as f32).unwrap()) / (mantissa + T::from(1 as f32).unwrap());
+    let y: T = (mantissa - T::from(1_f64).unwrap()) / (mantissa + T::from(1_f64).unwrap());
     
-    let mut sum: T = T::from(0 as f32).unwrap();
+    let mut sum: T = T::from(0_f64).unwrap();
     for k in 0..25 {
         sum = sum.add(y.powi(2*k + 1).div(T::from(2*k + 1).unwrap()));
     }
 
-    sum = sum.mul(T::from(2 as f32).unwrap());
-    Some(sum.add( T::from(LN10 * (n - 1) as f32).unwrap() ))
+    sum = sum.mul(T::from(2_f64).unwrap());
+    Some(sum.add( T::from(std::f64::consts::LN_10 * (n - 1) as f64).unwrap() ))
 }
 
 /// Calculate the normal cdf of the given number
--- a/libraries/math/src/approximations.rs
+++ b/libraries/math/src/approximations.rs
@@ -69,7 +69,7 @@ pub fn ln<T: Float + Clone>(input: T) -> Option<T> {
         sum += 2_f64 * y.powi(2*k + 1) / ((2*k + 1) as f64);
     }
 
-    T::from(sum + ( std::f64::consts::LN_10 * (n - 1) as f64 ))
+    T::from(sum + (std::f64::consts::LN_10 * (n - 1) as f64))
 }
 
 /// Calculate the normal cdf of the given number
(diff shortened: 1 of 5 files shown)


joncinque (dismissed):
This is a great start!  Now that the base implementation is there, we need a few things to check its validity:

* proptests to make sure that the approximation is good to a certain number of digits.  You can do this by comparing the output of your function do the actual library implementation of `ln` and making sure that the result is within some epsilon.  The proptests for sqrt in PreciseNumber are a good example of this: https://github.com/solana-labs/solana-program-library/blob/aaca5353bc7bd273031f5792854902dbe22601e0/libraries/math/src/precise_number.rs#L596
* Let's see the compute cost!  You can make a test similar to other instruction count tests by adding an instruction, a processor for it, and a test calling it: https://github.com/solana-labs/solana-program-library/blob/aaca5353bc7bd273031f5792854902dbe22601e0/libraries/math/tests/instruction_count.rs#L167  I'm curious to see how much cheaper it is than the native implementation

Edit: "requested changes" is a bit aggressive considering it's a draft, but I can't change it to a normal comment.  Sorry about that!

zrthxn:
Added proptest for values from 0 to INT32_MAX. Error for INT32_MAX is 0.0000076293945, which I think is the worst error since errors have been observed  to grow with size of input. This makes this function reliable unto 5 decimal places.

askibin:
> But the number of instructions is too high unfortunately. I'd appreciate some help on how to reduce this.
I would suggest not to loop 0..50 but until the max error threshold is reached, this way it will need much less iterations most of the time

askibin:
right, thats what I meant, increase number of loops until you get an error that is below threshold, but it sounds like you've already done that.
you can also replace recalculation of y.powi(2*k + 1) in the loop with a single mutable variable and two multiplications

zrthxn:
> you can also replace recalculation of y.powi(2*k + 1) in the loop with a single mutable variable and two multiplications
I tried this, of course the precision is still there, but I still run out of compute budget.

What would really help is some way of running a debugger such that I can check the compute budget after every line. 
Is that possible? Dyou know of anything like that?

joncinque:
> What would really help is some way of running a debugger such that I can check the compute budget left after every line.
Is that possible? Dyou know how I could do that?

There's no way to get the compute budget as a number from a syscall unfortunately.  We've had https://github.com/solana-labs/solana/issues/18134 open for awhile, and people have asked for it, but it still isn't done

joncinque:
> But the number of instructions is too high unfortunately. I'd appreciate some help on how to reduce this.

You can set a higher number for your compute budget in the test, 3500 is just the baseline number for calculating `ln(1)` in BPF "natively".  The hope is that the approximation costs less than the exact calculation, but perhaps the approximation has better compute unit performance for higher numbers.  You can try different input numbers and compute unit numbers in the instruction count test to compare the "native" implementation and your approximation.

zrthxn:
I've narrowed the problem down to this: the division on [line 69](https://github.com/solana-labs/solana-program-library/pull/3240/files#diff-439b112da95e93d7ec549b1af2d654078a0fcf79a4f8235c8ad918f96c2abb61R69) has a very high cost. Around 1100 units. I think this is because its a `f64` divided by another `f64`.

askibin:
there are other ways to approximate you can probably try, like this one: https://stackoverflow.com/questions/39821367/very-fast-approximate-logarithm-natural-log-function-in-c
but what problem are you trying to solve, built-in ln() is too slow?

zrthxn:
The main problem being solved here, I think, is to reduce the number of compute units consumed to calculate $ln(x)$ with the native implementation. This only makes sense if we can get a reasonably good approximation while using less compute units.

I will try and use `PreciseNumber` instead of `f64` and see if that makes a difference. If not, I will later try a different method, like the one you mentioned.

askibin:
compute units do not precisely represent the complexity (i.e. amount of cpu time spent), but rather number of instructions executed and their cost. f64::ln() was about 100 CU last time I checked, around the same as f64 addition or subtraction. So it would be impossible to beat native implementation (from CU perspective).

zrthxn:
Correct me if I'm wrong but it seems that the native implementation takes around 3114 CUs?
I just ran `cargo test-bpf` to check this.

Here are the relevant logs
```
[2022-06-21T19:23:30.447222000Z DEBUG solana_runtime::message_processor::stable_log] Program Math111111111111111111111111111111111111111 invoke [1]
[2022-06-21T19:23:30.447415445Z DEBUG solana_runtime::message_processor::stable_log] Program log: Calculating f32 Natural Log
[2022-06-21T19:23:30.447441442Z DEBUG solana_runtime::message_processor::stable_log] Program consumption: 3097 units remaining
[2022-06-21T19:23:30.448121472Z DEBUG solana_runtime::message_processor::stable_log] Program consumption: 879 units remaining
[2022-06-21T19:23:30.448267270Z DEBUG solana_runtime::message_processor::stable_log] Program log: 1
[2022-06-21T19:23:30.448297177Z DEBUG solana_rbpf::vm] BPF instructions executed (interp): 2714
[2022-06-21T19:23:30.448310581Z DEBUG solana_rbpf::vm] Max frame depth reached: 10
[2022-06-21T19:23:30.448322615Z DEBUG solana_runtime::message_processor::stable_log] Program Math111111111111111111111111111111111111111 consumed 3114 of 3500 compute units
[2022-06-21T19:23:30.448404361Z DEBUG solana_runtime::message_processor::stable_log] Program Math111111111111111111111111111111111111111 success
```

askibin:
not sure, I guess test environment is different. I just dropped this program to mainnet:
```
    msg!("Log calc test");
    let x = 17234.5f64;
    sol_log_compute_units();
    let log = x.ln();
    sol_log_compute_units();

    let x = 0.5f64;
    sol_log_compute_units();
    let log2 = x.ln();
    sol_log_compute_units();

    let x = 1.0f64;
    sol_log_compute_units();
    let log3 = x.ln();
    sol_log_compute_units();

    msg!("log: {}", log);
    msg!("log2: {}", log2);
    msg!("log3: {}", log3);
    return Ok(());
```
you can see result here:
https://explorer.solana.com/tx/9LrJ1Lm2wRAXaj5GsWQypGwk2xwMrXRX9fLgCntVdWoqWXyijLRJSQFzugjU9235PrxoBwcFqFMRB6zCakzb5gT
```
> Program consumption: 1399261 units remaining
> Program consumption: 1399258 units remaining
> Program consumption: 1399257 units remaining
> Program consumption: 1399254 units remaining
> Program consumption: 1399253 units remaining
> Program consumption: 1399250 units remaining
```
So, I was wrong, it actually consumes just 3 CU per ln() call. Or we are testing different things. Let me know what you think.

askibin:
Oh, I realized my test was dumb, calculations probably got optimized out by the compiler. My apologies. I got it re-done and here numbers are more realistic: https://explorer.solana.com/tx/5e5USBBebqBpVK6kH58HWRmkbWkv1tmPw4zEHCG7NdwtTSkwQZtBfCHj89DvTVC18LwpCcpBcvyou2B3TP5r1uWJ
so, depending on input it takes somewhere between 20 - 3000 CU. hence there could be some room for improvement

askibin:
since I was on it I ran it for other basic operations:
```
u32	
add:	14
sub:	13
mul:	3
div: 10
pow: 79
	
f32	
add:	151
sub:	76
mul:	118
div:	497
pow: 2
ln: 2091
	
u64	
add:	8
sub:	52
mul:	4
div:	9
pow: 292
	
f64	
add:	100
sub:	105
mul:	181
div:	902
pow: 2
ln: 2972
	
u128	
add:	6
sub:	107
mul:	43
div: 5
pow: 758
```
results heavily depend on the input, but still give some idea. As you can see, even though ln() is the heaviest it would be really hard to approximate it with anything else cause you will spend the limit in just a dozen iterations

joncinque:
@zrthxn For the PreciseNumber implementation, can you use a PreciseNumber everywhere? That is, as both the input and the result.  If it goes over the float compute unit amount, don't worry too much.  As long as it isn't ridiculous, I'll be happy to merge the change in

joncinque:
@zrthxn Are you still working on this?

zrthxn:
I have a few things I've done but there aren't any major improvements.
Someone else can work on it if they want to, I'll just keep my changes in
my branch for reference.

PR #3275, opened 2022-06-19
Check if amount+supply not greater than u64 supply limit: Issue 1983 error if more than max supply is minted

SUMEETRM:
Fixed issue #1983
Total supply implemented in u64, which means that total tokens that can be minted is 18,446,744,073,709,551,615

This means that the amount requested to be minted and the total current supply, summed up, should be less than 2^64

As of this commit, an error is thrown if this sum exceeds 2^64. This happens in the check_mint function. However, the next commits in this pull request will add a flag to overmint tokens and work on the following edge cases:

Edge case 1:
After the decimals have been checked and already stated, 
Do 20-decimals = y
Now a number with y digits is the maximum token supply that one can get

Edge case 2:
In fact, another nuance would be that we’d do min(check3, 18,446,744,073,709,551,615/(10^decimals))

Edge case 3 (Not important):
Number of decimals not greater than 20

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -985,6 +985,7 @@ fn command_mint(
     let max_denom =10;
     let max_denom = u64::pow(max_denom, decimals.into());
     let max_supply = u64::MAX/max_denom;
+    //For decimals:  max(u64)/(10^decimals) with integer divsion will be the max token supply we can get
     if amount>max_supply {
         println!("WARNING: Max supply will be limited to {}",max_supply);
     };
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -950,6 +950,9 @@ fn check_if_amount_overflows_supply(
     token: Pubkey,
     amount : u64,
 ) -> bool {
+    if config.sign_only {
+        return false;
+    };
     let supply = config.rpc_client.get_token_supply(&token).unwrap();
     let amount = amount;
     let supply: u64 = supply.amount.parse().unwrap();
@@ -964,7 +967,7 @@ fn check_if_amount_overflows_supply(
 fn command_mint(
     config: &Config,
     token: Pubkey,
-    ui_amount: u64,
+    amount: u64,
     recipient: Pubkey,
     mint_decimals: Option<u8>,
     mint_authority: Pubkey,
@@ -975,17 +978,17 @@ fn command_mint(
         config,
         format!(
             "Minting {} tokens\n  Token: {}\n  Recipient: {}",
-            ui_amount, token, recipient
+            amount, token, recipient
         ),
     );
     let (_, decimals) = resolve_mint_info(config, &recipient, None, mint_decimals)?;
     let max_denom =10;
     let max_denom = u64::pow(max_denom, decimals.into());
     let max_supply = u64::MAX/max_denom;
-    if ui_amount>max_supply {
+    if amount>max_supply {
         println!("WARNING: Max supply will be limited to {}",max_supply);
     };
-    let amount = spl_token::ui_amount_to_amountmint(ui_amount, decimals);
+    let amount = spl_token::ui_amount_to_amountmint(amount, decimals);
 
     let instructions = if use_unchecked_instruction {
         vec![mint_to(
@@ -2187,8 +2190,8 @@ fn app<'a, 'b>(
                         ),
                 )
                 .arg(
-                    Arg::with_name("overmint")
-                        .long("overmint-tokens")
+                    Arg::with_name("allow-supply-overflow")
+                        .long("allow-supply-overflow-u64")
                         .value_name("FORMAT")
                         .global(false)
                         .takes_value(false)
@@ -2918,7 +2921,7 @@ fn process_command(
             );
             let mint_decimals = value_of::<u8>(arg_matches, MINT_DECIMALS_ARG.name);
             let use_unchecked_instruction = arg_matches.is_present("use_unchecked_instruction");
-            let overmint_flag = arg_matches.is_present("overmint");
+            let overmint_flag = arg_matches.is_present("allow-supply-overflow");
             //Disable overflow check - change flag name.
 
             if overmint_flag {
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -305,6 +305,18 @@ pub fn signers_of(
     }
 }
 
+fn check_decimals(  
+    config: &Config,
+    token: Pubkey,
+    decimals: u8,
+) -> bool {
+    let d = decimals;
+    if d> 20 {
+    return true;
+    };
+    return false;
+}
+
 #[allow(clippy::too_many_arguments)]
 fn command_create_token(
     config: &Config,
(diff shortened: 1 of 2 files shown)


token/cli/src/main.rs
```diff
@@ -2885,6 +2900,10 @@ fn process_command(
             );
             let mint_decimals = value_of::<u8>(arg_matches, MINT_DECIMALS_ARG.name);
             let use_unchecked_instruction = arg_matches.is_present("use_unchecked_instruction");
+            if check_mint(config, token, amount) {
```
joncinque, line 2903: How about a flag like "allow-supply-overflow" to disable the check?

token/cli/src/main.rs
```diff
@@ -943,6 +943,21 @@ fn command_burn(
     })
 }
 
+
+fn check_mint(
+    config: &Config,
+    token: Pubkey,
+    ui_amount : f64,
```
joncinque, line 950: The UI amount can't be trusted totally, since an `f64` can't contain the same precision as a `u64`.  You'll have to convert the `ui_amount` into an `amount` and also get a fixed-point amount for the supply.  This means that you might have to move this check inside of `command_mint`

token/cli/src/main.rs
```diff
@@ -943,6 +943,21 @@ fn command_burn(
...
+) -> bool {
+    let supply = config.rpc_client.get_token_supply(&token).unwrap();
+    let amount = ui_amount;
+    if amount + supply.ui_amount.unwrap()> 18446744073709551615.0 {
```
joncinque, line 954: When using fixed-point numbers, to easily check for an overflow, you can use `checked_add` and error if that fails: https://doc.rust-lang.org/std/primitive.u64.html#method.checked_add

token/cli/src/main.rs
```diff
@@ -943,6 +943,21 @@ fn command_burn(
...
+    return true;
+    };
+    return false;
```
joncinque, line 957: nit: you don't need explicit return statements in Rust, so this can be done as:

```
amount.checked_add(supply).is_err()
```

token/cli/src/main.rs
```diff
@@ -943,6 +943,21 @@ fn command_burn(
     })
 }
 
+
+fn check_mint(
```
joncinque, line 947: nit: 
```suggestion
fn check_if_amount_overflows_supply(
```
Or something more descriptive

joncinque (commented):
This is on the right track! Keep it up

token/cli/src/main.rs
SUMEETRM: Done

token/cli/src/main.rs
```diff
@@ -305,6 +305,16 @@ pub fn signers_of(
...
 
+fn check_decimals(  
+    decimals: u8,
+) -> bool {
+    let d = decimals;
+    if d> 20 {
+    return true;
+    };
+    return false;
+}
```
joncinque, line 316: Sorry, I'm not sure that we need this check in the end -- can we remove it and consider adding it in a new PR with a flag to toggle the check?

token/cli/src/main.rs
```diff
@@ -943,11 +953,34 @@ fn command_burn(
     })
 }
 
+
+fn check_if_amount_overflows_supply(  
+    config: &Config,
+    token: Pubkey,
+    ui_amount : u64,
```
joncinque, line 960: nit: let's have this be a real `amount`, and not ui-related

token/cli/src/main.rs
```diff
@@ -943,11 +953,34 @@ fn command_burn(
...
+) -> bool {
+    let supply = config.rpc_client.get_token_supply(&token).unwrap();
+    let amount = ui_amount;
+    if amount > u64::MAX {
+        return true;
+    };
```
joncinque, line 966: nit: this can be removed since it'll be covered by the later check

token/cli/src/main.rs
```diff
@@ -943,11 +953,34 @@ fn command_burn(
     })
 }
 
+
+fn check_if_amount_overflows_supply(  
+    config: &Config,
+    token: Pubkey,
+    ui_amount : u64,
+) -> bool {
+    let supply = config.rpc_client.get_token_supply(&token).unwrap();
```
joncinque, line 962: if we're offline, we can't do this check, so consider just returning `false` if `config.sign_only`

token/cli/src/main.rs
```diff
@@ -943,11 +953,34 @@ fn command_burn(
...
+    let amount = ui_amount;
+    if amount > u64::MAX {
+        return true;
+    };
+    let supply_val = supply.amount;
+    let supply_u64: u64 = supply_val.parse().unwrap();
```
joncinque, line 968: nit: can condense these into
```suggestion
    let supply: u64 = supply.amount.parse().unwrap();
```

token/cli/src/main.rs
```diff
@@ -943,11 +953,34 @@ fn command_burn(
...
+    if amount.checked_add(supply_u64).is_none() {
+        return true;
+    };
+    if amount + supply_u64 > u64::MAX {
+    return true;
+    };
+    return false;
```
joncinque, line 975: nit: The check here is now being done twice -- you can condense this all down to just
```suggestion
    amount.checked_add(supply).is_none()
```

token/cli/src/main.rs
```diff
@@ -943,11 +953,34 @@ fn command_burn(
...
+}
+    
+
 #[allow(clippy::too_many_arguments)]
 fn command_mint(
     config: &Config,
     token: Pubkey,
-    ui_amount: f64,
+    ui_amount: u64,
```
joncinque, line 983: nit: 
```suggestion
    amount: u64,
```

token/cli/src/main.rs
```diff
@@ -2877,14 +2923,19 @@ fn process_command(
             let token = pubkey_of_signer(arg_matches, "token", &mut wallet_manager)
                 .unwrap()
                 .unwrap();
-            let amount = value_t_or_exit!(arg_matches, "amount", f64);
+            let amount = value_t_or_exit!(arg_matches, "amount", u64);
```
joncinque, line 2926: Be careful, this won't be correct if someone wants to mint `1.5` tokens.  The best way to do this is convert from the string in `amount` to a u64 amount using `spl_token::try_ui_amount_into_amount`

token/program/src/lib.rs
```diff
@@ -22,6 +22,11 @@ pub fn ui_amount_to_amount(ui_amount: f64, decimals: u8) -> u64 {
     (ui_amount * 10_usize.pow(decimals as u32) as f64) as u64
 }
 
+/// test
+pub fn ui_amount_to_amountmint(ui_amount: u64, decimals: u8) -> u64 {
+    (ui_amount * 10_usize.pow(decimals as u32) as u64) as u64
+}
+
```
joncinque, line 29: Be sure to remove this

joncinque:
@SUMEETRM are you still working on this?

SUMEETRM:
> @SUMEETRM are you still working on this?

Yes, will make add some commits to my PR before the maintainer meeting

token/cli/src/main.rs
SUMEETRM: Need to work on this

Issue #887, opened 2020-11-24
token-cli: unexpected behavior when closing a wrapped SOL account with another wrapped SOL account as recipient


RichardWarfield:
When closing a wrapped SOL account (`spl-token close <acctA> <acctB>`) where acctB is another wrapped SOL token account, the token program will happily send unwrapped SOL to acctB, resulting in a situation where acctB has a balance of SOL that cannot easily be accessed.  I'm not entirely sure this qualifies as a bug but it seems to me that either the `close` instruction should fail when `acctA` is non-empty, or the token program should update the wrapped SOL amount to reflect the new balance.

mvines:
A check could be added in `spl-token close` to ensure that `<acctB>` is a system account, with a `--force` option to override.   This behaviour is aligned with `solana transfer`, which will happily send SOL to any account even if it's a non-system account. 

So it's not a bug per se.  Close is in fact fully honouring your request and working as initially designed.  But it certainly could be made more user friendly.

joncinque:
Since we have `sync-native` available, the CLI can tack on a `sync-native` instruction if the destination is a wrapped account.

zrthxn:
@joncinque  I'll be working on this next.

PR #3278, opened 2022-06-20
token-cli: Fix unexpected behavior when closing a wrapped SOL account with another wrapped SOL account as recipient

zrthxn:
Fixes: #887

Tacked on a sync-native instruction if the destination is a wrapped account.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -3648,6 +3648,48 @@ mod tests {
     }
 
     #[test]
+    fn close_wrapped_sol_account() {
+        let (test_validator, payer) = validator_for_test();
+        let config = test_config(&test_validator, &payer, &spl_token::id());
+        let bulk_signers: Vec<Box<dyn Signer>> = vec![Box::new(clone_keypair(&payer))];
+
+        let token = create_token(&config, &payer);
+        let source = create_associated_account(&config, &payer, token);
+        let ui_amount = 10.0;
+        command_wrap(
+            &config,
+            ui_amount,
+            payer.pubkey(),
+            Some(source),
+            bulk_signers,
+        )
+        .unwrap();
+
+        let recipient = get_associated_token_address_with_program_id(
+            &payer.pubkey(),
+            &native_mint::id(),
+            &config.program_id,
+        );
+        let _result = process_test_command(
+            &config,
+            &payer,
+            &[
+                "spl-token",
+                CommandName::Close.into(),
+                &source.to_string(),
+                &recipient.to_string(),
+            ],
+        );
+
+        let ui_account = config
+            .rpc_client
+            .get_token_account(&recipient)
+            .unwrap()
+            .unwrap();
+        assert_eq!(ui_account.token_amount.amount, "10");
+    }
+
+    #[test]
     fn disable_mint_authority() {
         let (test_validator, payer) = validator_for_test();
         let config = test_config(&test_validator, &payer, &spl_token::id());
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -39,7 +39,7 @@ use solana_sdk::{
     pubkey::Pubkey,
     signature::{Keypair, Signer},
     system_instruction, system_program,
-    transaction::Transaction,
+    transaction::Transaction
 };
 use spl_associated_token_account::{
     get_associated_token_address_with_program_id, instruction::create_associated_token_account,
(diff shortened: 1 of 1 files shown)


zrthxn (commented):
Need to know how to check what type of account was provided in `destination`

token/js/src/actions/closeAccount.ts
```diff
@@ -41,5 +42,14 @@ export async function closeAccount(
         createCloseAccountInstruction(account, destination, authorityPublicKey, multiSigners, programId)
     );
 
+    try {
+        const destinationAccount = await getAccount(connection, destination, undefined, TOKEN_PROGRAM_ID);
```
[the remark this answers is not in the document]
joncinque: Ah, so the change for this should be in the CLI, and not in the JS module.  You'll want to look at https://github.com/solana-labs/solana-program-library/blob/57c5fd80f60ed4d6053989e64882f9002e352910/token/cli/src/main.rs#L1168 and https://github.com/solana-labs/solana-program-library/blob/57c5fd80f60ed4d6053989e64882f9002e352910/token/cli/src/main.rs#L1347 for this specific change

zrthxn:
I'd appreciate just some direction on whether or not I'm doing this right... I kinda used the same logic as I had with the JS client changes. 

Check if the recipient is a token account, if that fails we know its a wallet address and we dont need to do anything. If it is a token account and `recipient_accout.is_native` is `true`, then append a `sync-native` instruction.

token/cli/src/main.rs
```diff
@@ -1377,13 +1377,22 @@ fn command_close(
...
     )?];
+
+    let recipient_account = config
+        .rpc_client
+        .get_token_account(&recipient)?;
+
+    if !recipient_account.is_none() && recipient_account.unwrap().is_native {
+        instructions.push(sync_native(&config.program_id, &recipient)?);
+    }
+
```
joncinque, line 1395: These checks need to be moved up in the `if !config.sign_only` block, since in offline mode, we can't fetch the recipient account.

token/cli/src/main.rs
```diff
@@ -1377,13 +1377,22 @@ fn command_close(
...
         &recipient,
         &close_authority,
         &config.multisigner_pubkeys,
     )?];
+
+    let recipient_account = config
+        .rpc_client
+        .get_token_account(&recipient)?;
+
+    if !recipient_account.is_none() && recipient_account.unwrap().is_native {
```
joncinque, line 1392: Be sure to run clippy to do this check in a clearer way with: `$ cargo +nightly-2022-02-24 clippy -Zunstable-options --all-targets`

joncinque (commented):
This is definitely on the right track!

zrthxn:
Maybe another issue to add a test for this, i.e. recipient account when closing is a wrapped sol account, would also be a good idea?

token/cli/src/main.rs
```diff
@@ -1375,15 +1376,23 @@ fn command_close(
             )
             .into());
         }
+
+        let recipient_account = config.rpc_client.get_token_account(&recipient)?;
+        is_recipient_wrapped = recipient_account.is_some() && recipient_account.unwrap().is_native;
```
joncinque, line 1381: nit for more functional clarity: 
```suggestion
        is_recipient_wrapped = recipient_account.and_then(|x| x.is_native).unwrap_or(false)
```

token/cli/src/main.rs
```diff
@@ -1351,6 +1351,7 @@ fn command_close(
     recipient: Pubkey,
     bulk_signers: BulkSigners,
 ) -> CommandResult {
+    let mut is_recipient_wrapped = false;
```
joncinque, line 1354: Nit to avoid mutables:
```suggestion
    let is_recipient_wrapped = if config.sign_only {
        false
    } else {
    ... the rest of the block as normal ...
    };
```

joncinque (commented):
Very slick and easy -- just be sure to add a test and this will be good to go!

zrthxn:
I've added a first draft of a test for this. Please let me know if its correct/going in the right direction at all.

token/cli/src/main.rs
```diff
@@ -3636,6 +3647,48 @@ mod tests {
...
+        let source = create_associated_account(&config, &payer, token);
+        let ui_amount = 10.0;
+        command_wrap(
+            &config,
+            ui_amount,
+            payer.pubkey(),
+            Some(source),
+            bulk_signers,
+        )
+        .unwrap();
```
joncinque, line 3666: This won't do exactly what you expect -- you're creating a new mint and account, and then trying to wrap that account that was just created.  To do this simply. you can just wrap the default account by doing
```suggestion
        let token = create_token(&config, &payer);
        let source = create_associated_account(&config, &payer, token);
        let ui_amount = 10.0;
        command_wrap(
            &config,
            ui_amount,
            payer.pubkey(),
            None,
            bulk_signers,
        )
        .unwrap();
```

token/cli/src/main.rs
```diff
@@ -3636,6 +3647,48 @@ mod tests {
...
+                &recipient.to_string(),
+            ],
+        );
+
+        let ui_account = config
+            .rpc_client
+            .get_token_account(&recipient)
+            .unwrap()
+            .unwrap();
+        assert_eq!(ui_account.token_amount.amount, "10");
```
joncinque, line 3689: After the closing from `source` into `recipient`, with your change, it should be something like "10.00203928"

joncinque (commented):
The changes look great, just need to get the test working, and this'll be good to go

joncinque:
Looks like you still need to run `cargo fmt` on your changes -- you can do that from `token/cli`.  Also, there are some test failures, so be sure to fix those.

joncinque:

PR #3311, opened 2022-07-04
Remove deposit constraint

dankelleher:
Now that token deposits can be used as inputs to vote plugins, remove the constraint that prevents token deposits when a vote plugin is used.


SebastianBor (approved):

joncinque (approved):
Looks good!

PR #3322, opened 2022-07-07
CI: Bump openssl for audit

joncinque:
#### Problem

Audit is currently failing in CI.

#### Solution

Same as https://github.com/solana-labs/solana/pull/26410, we need to bump openssl.


joncinque:
Merging early!

PR #3330, opened 2022-07-10
Fixed typing error

fskroes:


CriesofCarrots:
Actually, this isn't a typo. `AToken` is short for the Associated Token (Program). There might be a clearer way to introduce the short-hand, or maybe it should just be spelled out entirely. Please feel free to re-open this PR with changes along those lines.

PR #3335, opened 2022-07-12
[stake-pool] instruction to add metadata for pool token

betterclever:
#### Problem

The ecosystem has abandoned the token-list and encourages using the Metaplex Metadata standard for tokens.  This requires a signature from the mint authority, which means that it's currently impossible to create token metadata for stake pools.

#### Solution

Add new `create_metadata` and `update_metadata` instructions to the stake pool program to allow working with metaplex metadata accounts.  This provides a signature from the stake pool withdraw authority, which is the mint authority.  The instruction must be signed by the stake pool manager.


stake-pool/program/tests/create_pool_token_metadata.rs
```diff
@@ -0,0 +1,125 @@
...
+
+#[tokio::test]
+async fn fail_manager_did_not_sign() {
+    let (mut context, stake_pool_accounts) = setup().await;
+
+    let name = "test_name";
+    let symbol = "SYM";
+    let uri = "test_uri";
+
+    let ix = instruction::create_token_metadata(
```
joncinque, line 92: after creating the instruction, can you also modify the manager by hand to not be a signer? ie:
```
ix.accounts[1].is_signer = false;
```
Otherwise, this will fail before it even gets to the program

stake-pool/program/tests/update_pool_token_metadata.rs
```diff
@@ -0,0 +1,108 @@
...
+    let metadata = get_metadata_account(
+        &mut context.banks_client,
+        &stake_pool_accounts.pool_mint.pubkey(),
+    )
+    .await;
+
+    assert_eq!(metadata.data.name.to_string(), puffed_name);
+    assert_eq!(metadata.data.symbol.to_string(), puffed_symbol);
+    assert_eq!(metadata.data.uri.to_string(), puffed_uri);
+}
```
joncinque, line 108: Just as one more sanity check, can you add a test when the wrong manager signs and one with the wrong metadata program id?

stake-pool/program/tests/create_pool_token_metadata.rs
```diff
@@ -0,0 +1,125 @@
...
+        .unwrap();
+
+    match error {
+        TransactionError::InstructionError(_, InstructionError::Custom(error_index)) => {
+            let program_error = crate::error::SignatureMissing as u32;
+            assert_eq!(error_index, program_error);
+        }
+        _ => panic!("Wrong error occurs while signing with the wrong manager"),
+    }
+}
```
joncinque, line 125: Can you add one more test failing to create it twice?

joncinque (commented):
Lots of really great work! This is absolutely on the right track, and I'm looking forward to integrating this soon.

stake-pool/program/src/instruction.rs
```diff
@@ -374,6 +379,48 @@ pub enum StakePoolInstruction {
...
     ///  12. `[s]` (Optional) Stake pool sol withdraw authority
     WithdrawSol(u64),
+
+    /// Create token metadata for the stake-pool token in the
+    /// metaplex-token program
+    /// 0. `[]` Stake pool
+    /// 1. `[]` Manager
+    /// 2. `[]` Stake pool withdraw authority
+    /// 3. `[]` Pool token mint account
+    /// 4. `[]` Payer for creation of token metadata account
```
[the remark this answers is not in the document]
joncinque: That's a good question.  It may be because the specified `payer` is the same as the fee payer in your test transactions.  You can try a test with a different fee payer and see what happens!  I'm surprised that it lets this through though :thinking:

betterclever:
@joncinque I have addressed all the comments highlighted above. Please re-review whenever possible :D

joncinque:
This looks good to me, you'll need to run `cargo fmt` on `stake-pool/program` to pass CI, and run clippy to make sure everything's ok there.  Let's wait to see if any of the other stake pool devs want to take a look too

stake-pool/program/tests/update_pool_token_metadata.rs
```diff
@@ -0,0 +1,108 @@
...
```
joncinque, line 108: Oh sorry, I'm just noticing at the end of this pass, but can you also add a test to make sure that the correct manager must sign? Same as with `fail_manager_did_not_sign` but for this instruction

betterclever: Will do.

CMEONE (approved):
I just skimmed through the changes, and it looks all good to me (other than adding those tests and resolving the CI as @joncinque has mentioned). I'm looking forward to seeing this get merged soon!

joncinque (dismissed):
Looks great! I'll give one more day for any external reviews, and merge tomorrow

joncinque:
@betterclever can you look at the CI test failure and fix the issues? Looks like an unused import that needs to be removed

stake-pool/program/src/processor.rs
```diff
@@ -2676,6 +2722,170 @@ impl Processor {
...
+        let pool_mint_info = next_account_info(account_info_iter)?;
+        let payer_info = next_account_info(account_info_iter)?;
+        let metadata_info = next_account_info(account_info_iter)?;
+        let mpl_token_metadata_program_info = next_account_info(account_info_iter)?;
+        let system_program_info = next_account_info(account_info_iter)?;
+        let rent_sysvar_info = next_account_info(account_info_iter)?;
+
+        check_system_program(system_program_info.key)?;
+        check_rent_sysvar(rent_sysvar_info.key)?;
+        check_account_owner(payer_info, &system_program::id())?;
```
joncinque, line 2746: Could you also check that `payer_info` is a signer?  That way we're sure that a PDA for this program can't possibly get spoofed in this position.

joncinque (commented):
Sorry for the late addition, but with this I'll feel a little more comfortable and ready to merge!

joncinque:
Thanks so much for all of your work!

Issue #822, opened 2020-11-11
token-swap: Token error messages are printed incorrectly


jstarry:
#### Problem
The token-swap program prints all `ProgramError`s returned from the processor as if they are `SwapError`s but they actually may be `TokenError`s. This means errors are only identified by custom error index and so a `TokenError::NotRentExempt` error would be printed as `SwapError::AlreadyInUse`.

#### Proposed Changes
- Create new `SwapError` variants for failed token transfers, burns, and mints
- Log invoked token instruction errors
- Map invoked token instruction errors to those variants instead of passing through

throwbackjams:
Hey Jon / Justin - I made a [pull request](https://github.com/solana-labs/solana-program-library/pull/3338) which addresses this issue. In the PR description, you will find a summary of changes made, in case helpful. Please let me know your thoughts - I'm open to feedback!

Thanks!

PR #3338, opened 2022-07-12
token-swap: (fix) Convert token program CPI errors to correct SwapError variants

throwbackjams:
# Issue & Change Summary
**Issue**: Addresses [issue 822](https://github.com/solana-labs/solana-program-library/issues/822)
For more background / a refresher of the issue, please see below.

**Summary of changes**: Please let me know your thoughts - Very open to feedback!
-   Created `SwapError` variants that correspond to the errors returned by the token program for the Transfer, Burn, and Mint instructions
-   Defined a function `convert_token_error_to_swap_error` within `error.rs`, which takes the`ProgramError` returned by calls to the token program and converts it into the corresponding (newly created) `SwapError` variant, giving appropriate information on the type of error received from the token program call
-  Use `map_err` on calls to the token program to convert the token program error to the appropriate `SwapError`
-   Modified tests within `proccessor.rs` to remove references to `TokenError` and use the new `SwapError` variants instead

# Change Detail: Background, Approach, and Notes
**Background**: When the token-swap program makes a cross-program invocation to the token program and an error occurs, the error that gets returned is mapped to the _position_ of the `SwapError` variant instead of getting mapped to the correct _corresponding_ `SwapError` variant
- Example: If the CPI returned `TokenError::NotRentExempt`, `Err(Custom(0))` in the _token_ program, then the error displayed would be `SwapError::AlreadyInUse`, `Err(Custom(0))` in the _token-swap_ program. The **position** of the error enum variant is interpreted instead of the error variant itself.

**Approach - Steps**:
1.   Modify tests in `processor.rs` to remove TokenError references and replace with new to-be defined `SwapError` variants
2.   Identify spots in token-swap that make a CPI to the token program (Transfer, Burn, Mint instructions)
3.   Add `SwapError` variants based on the possible errors returned by a CPI to the token program via a Transfer, Burn, or Mint instruction
4.   Implement a function (`convert_token_error_to_swap_error`) to convert token program errors to the corresponding `SwapError` enum variant, using a match on custom error received. E.g. `Custom(1) => SwapError::TokenProgramErrorInsufficientFunds`
5.   In spots with a CPI to the token program (`invoke`), add a `map_err`, which logs the error in place of occurence with `msg!` and calls the conversion function
6.   Modify implementation of `PrintProgramError` for `SwapError` to include new `SwapError` variants

**Notes**: 
- Because `convert_token_error_to_swap_error` uses a match internally, I used `SwapError::InvalidInstruction` for the catch-all at the end of the match (since the conversion only needs to cover some of the `TokenError` enum variants). This arm should never be reached unless the Token Program Transfer/Burn/Mint instruction processing changes. However, another reasonable approach could be to create a new generic `SwapError` variant for CPI call failures and use that variant for the catch-all instead. Open to thoughts!

### DIFF
--- a/token-swap/program/src/error.rs
+++ b/token-swap/program/src/error.rs
@@ -1,7 +1,9 @@
 //! Error types
 
 use num_derive::FromPrimitive;
-use solana_program::{decode_error::DecodeError, program_error::ProgramError};
+use solana_program::{
+    decode_error::DecodeError, program_error::ProgramError, program_error::ProgramError::Custom,
+};
 use thiserror::Error;
 
 /// Errors that may be returned by the TokenSwap program.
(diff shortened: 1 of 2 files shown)


joncinque:
Thanks for your contribution!  In the end, we preferred an approach of wrapping the CPI calls and printing the error from there, rather than copying all of the error types from `TokenError` into `SwapError`.  You can see the PR #3357.  If you're looking for any other token-swap issues, let me know!

throwbackjams:
Thank you for linking that PR, Jon! That approach makes sense and I really appreciate the feedback.

I'm interested in contributing generally! Can be but certainly doesn't need to be only token-swap issues (I also took a stab recently at token-lending with a [PR](https://github.com/solana-labs/solana-program-library/pull/3349)). Let me know where I can plug in!

joncinque:
Definitely! The best is to start with the "good first issue"s tagged in the repo, and I've given you an overdue review on your other contribution.  Thanks so much for that :smile:

PR #3345, opened 2022-07-14
[token-js] :  Support for UiAmountToAmount and AmountToUiAmount instructions

### LINKED ISSUE #3264: token-js: Support `UiAmountToAmount` and `AmountToUiAmount` instructions
#### Problem

As part of token-2022, we added new instructions to convert both ways between amount and ui amount, but there's no support for them in JS.  For reference:

* https://github.com/solana-labs/solana-program-library/blob/056948bbd1d40c240bcdecb78cee3afeb9903ede/token/program-2022/src/instruction.rs#L477-L490
* https://github.com/solana-labs/solana-program-library/blob/056948bbd1d40c240bcdecb78cee3afeb9903ede/token/program-2022/src/instruction.rs#L491-L502

#### Solution

Add implementations for these instructions with corresponding actions.

atharmohammad:
This PR adds :
1). support for ```amount to uiamount``` instruction
2). support for ```uiamount to amount ``` instruction
3). upgrades solana/web3 from 1.41.0 to 1.47.4
4). e2e test for instructions

fixes #3264 

cc : @joncinque


token/js/package.json
```diff
@@ -52,6 +52,7 @@
         "@types/eslint-plugin-prettier": "^3.1.0",
         "@types/mocha": "^9.1.0",
         "@types/node": "^16.11.21",
+        "@types/node-fetch": "^2.6.2",
```
atharmohammad, line 55: Added ```node-fetch``` as i was not able to build the lib getting error for not having node-fetch as it was imported in ```@solana/web3.js```

token/js/package.json
```diff
@@ -43,7 +43,7 @@
     "dependencies": {
         "@solana/buffer-layout": "^4.0.0",
         "@solana/buffer-layout-utils": "^0.2.0",
-        "@solana/web3.js": "^1.41.0"
+        "@solana/web3.js": "^1.47.4"
```
atharmohammad, line 46: needed to upgrade for getting ```returnData``` in simulateTransactionResponse

token/js/src/instructions/uiAmountToAmount.ts
```diff
@@ -0,0 +1,123 @@
...
+import { TokenInstruction } from './types';
+
+/** TODO: docs */
+export interface UiAmountToAmountInstructionData {
+    instruction: TokenInstruction.UiAmountToAmount;
+    amount: string;
+}
+
+/** TODO: docs */
+export const uiAmountToAmountInstructionData = struct<UiAmountToAmountInstructionData>([u8('instruction'),cstr('amount')]);
```
atharmohammad, line 19: When taking ```cstr``` for amount i am getting error ``` RangeError [ERR_INVALID_ARG_VALUE]: The argument 'size' is invalid. Received -1```, what should i use for representing as ```uiamount``` is string

token/js/test/e2e-2022/Amount.test.ts
```diff
@@ -0,0 +1,45 @@
+import chai, { expect } from 'chai';
```
joncinque, line 1: nit: for consistency, can have the file name lower-case? ie `amount.test.ts`

token/js/test/e2e-2022/Amount.test.ts
```diff
@@ -0,0 +1,45 @@
...
+            TEST_TOKEN_DECIMALS,
+            mintKeypair,
+            undefined,
+            TEST_PROGRAM_ID
+        );
+    });
+    it('amountToUiAmount', async () => {
+        const amount = BigInt(5245);
+        const uiAmount = await amountToUiAmount(connection, payer, mint, amount, TEST_PROGRAM_ID);
+        expect(uiAmount).to.be.not.null;
```
joncinque, line 37: nit: It's not necessary to test not null since you're testing its equality right after

token/js/test/e2e-2022/Amount.test.ts
```diff
@@ -0,0 +1,45 @@
...
+        expect(uiAmount).to.eql("52.45");
+    });
+    it.only('uiAmountToAmount', async () => {
+        const uiAmount = await uiAmountToAmount(connection, payer, mint, "52.45", TEST_PROGRAM_ID);
+        expect(uiAmount).to.be.not.null;
```
joncinque, line 42: Same here

token/js/src/instructions/uiAmountToAmount.ts
```diff
@@ -0,0 +1,123 @@
...
+ * @param programId    SPL Token program account
+ *
+ * @return Instruction to add to a transaction
+ */
+export function createUiAmountToAmountInstruction(
+    mint: PublicKey,
+    amount: string,
+    programId = TOKEN_PROGRAM_ID
+): TransactionInstruction {
+    const keys = [{ pubkey: mint, isSigner: false, isWritable: true }];
```
joncinque, line 35: The mint isn't being written to here, so be sure to keep it read-only to help with parallelization of transactions on the network
```suggestion
    const keys = [{ pubkey: mint, isSigner: false, isWritable: false }];
```

token/js/src/instructions/uiAmountToAmount.ts
joncinque: I think that `utf8` will be a better option, since `cstr` will null-terminate the bytes, and the program isn't expecting that

token/js/src/instructions/index.ts
```diff
@@ -22,8 +22,10 @@ export * from './initializeAccount3'; //  18
 export * from './initializeMultisig2'; // 19
 export * from './initializeMint2'; //     20
 export * from './initializeImmutableOwner'; //  22
-export * from './initializeMintCloseAuthority'; // 23
-export * from './createNativeMint'; //    29
+export * from './amountToUiAmount'; // 23
+export * from './uiAmountToAmount'; //24
+export * from './initializeMintCloseAuthority'; // 25
```
joncinque, line 27: Thanks for fixing these up!

token/js/src/instructions/decode.ts
```diff
@@ -21,6 +21,7 @@ import { DecodedSyncNativeInstruction, decodeSyncNativeInstruction } from './syn
 import { DecodedThawAccountInstruction, decodeThawAccountInstruction } from './thawAccount';
 import { DecodedTransferInstruction, decodeTransferInstruction } from './transfer';
 import { DecodedTransferCheckedInstruction, decodeTransferCheckedInstruction } from './transferChecked';
+import { DecodedAmountToUiAmountInstruction, decodeAmountToUiAmountInstruction } from './amountToUiAmount';
```
joncinque, line 24: Can you also add support for UiAmountToAmount in this file?

token/js/src/instructions/amountToUiAmount.ts
```diff
@@ -0,0 +1,124 @@
...
+ * @param programId    SPL Token program account
+ *
+ * @return Instruction to add to a transaction
+ */
+export function createAmountToUiAmountInstruction(
+    mint: PublicKey,
+    amount: number | bigint,
+    programId = TOKEN_PROGRAM_ID
+): TransactionInstruction {
+    const keys = [{ pubkey: mint, isSigner: false, isWritable: true }];
```
joncinque, line 36: Mint isn't writable in this instruction either
```suggestion
    const keys = [{ pubkey: mint, isSigner: false, isWritable: false }];
```

token/js/src/actions/uiAmountToAmount.ts
```diff
@@ -0,0 +1,44 @@
...
+    connection: Connection,
+    payer: Signer,
+    mint: PublicKey,
+    amount: string,
+    programId = TOKEN_PROGRAM_ID
+): Promise<number> {
+    const transaction = new Transaction().add(
+        createUiAmountToAmountInstruction(mint,amount,programId)
+    );
+    try{
```
joncinque, line 31: No need to try / catch here

token/js/src/actions/uiAmountToAmount.ts
```diff
@@ -0,0 +1,44 @@
...
+        const {returnData,err} = (await connection.simulateTransaction(transaction, [payer] , false)).value;
+        console.log(err)
+        if(returnData?.data){
+            const x = Buffer.from(returnData.data[0],returnData.data[1]);
+            console.log(x);
+            return 5245;
```
joncinque, line 37: You'll have to decode a `u64` from `x`, something like `u64.decode(x)` should do it from `@solana/buffer-layout-utils`

token/js/src/actions/uiAmountToAmount.ts
```diff
@@ -0,0 +1,44 @@
...
+): Promise<number> {
+    const transaction = new Transaction().add(
+        createUiAmountToAmountInstruction(mint,amount,programId)
+    );
+    try{
+        const {returnData,err} = (await connection.simulateTransaction(transaction, [payer] , false)).value;
+        console.log(err)
+        if(returnData?.data){
+            const x = Buffer.from(returnData.data[0],returnData.data[1]);
+            console.log(x);
```
joncinque, line 36: Please remove extra logging calls

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -0,0 +1,41 @@
...
+    amount: number | bigint,
+    programId = TOKEN_PROGRAM_ID
+): Promise<string> {
+    const transaction = new Transaction().add(
+        createAmountToUiAmountInstruction(mint,amount,programId)
+    );
+    try{
+        const {returnData} = (await connection.simulateTransaction(transaction, [payer] , false)).value;
+        if(returnData?.data){
+            return Buffer.from(returnData.data[0],returnData.data[1]).toString();
```
joncinque, line 34: nit: let's be explicit about the encoding:
```suggestion
            return Buffer.from(returnData.data[0],returnData.data[1]).toString('utf8');
```

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -0,0 +1,41 @@
...
+    connection: Connection,
+    payer: Signer,
+    mint: PublicKey,
+    amount: number | bigint,
+    programId = TOKEN_PROGRAM_ID
+): Promise<string> {
+    const transaction = new Transaction().add(
+        createAmountToUiAmountInstruction(mint,amount,programId)
+    );
+    try{
```
joncinque, line 31: Can remove this try/catch block

joncinque (commented):
This is in pretty good shape already! Just some notes and questions to make this ready.  Also, be sure to run `yarn lint` or `yarn lint:fix` to get the formatting standard and fix the lint errors

token/js/src/instructions/uiAmountToAmount.ts
atharmohammad: So i found that the Error is due to ```Buffer.alloc(uiAmountToAmountInstructionData.span)``` the span generated here is -1 so it is throwing error. i tried using utf8 but still i am getting -1 as span . I tried something like 
```
export interface UiAmountToAmountInstructionData {
    instruction: TokenInstruction.UiAmountToAmount;
    amount: string | LayoutObject;
}

export const uiAmountToAmountInstructionData = struct<UiAmountToAmountInstructionData([u8('instruction'),utf8(128,'amount')]);
```

joncinque: Ah right my apologies, yes, you'll need to specify the size somehow, otherwise `span` will give you that `-1` result.  Instead of doing `Buffer.alloc(uiAmountToAmountInstructionData.span);`, you can do:

```
const buf = Buffer.from(amount, 'utf8');
const uiAmountToAmountInstructionData = struct<UiAmountToAmountInstructionData([u8('instruction'), blob(buf.length)]);
const data = Buffer.alloc(uiAmountToAmountInstructionData.span);
```

And then the rest as normal

token/js/src/instructions/uiAmountToAmount.ts
```diff
@@ -0,0 +1,132 @@
...
+ * @return Decoded, valid instruction
+ */
+export function decodeUiAmountToAmountInstruction(
+    instruction: TransactionInstruction,
+    programId = TOKEN_PROGRAM_ID
+): DecodedUiAmountToAmountInstruction {
+    if (!instruction.programId.equals(programId)) throw new TokenInvalidInstructionProgramError();
+    const uiAmountToAmountInstructionData = struct<UiAmountToAmountInstructionData>([
+        u8('instruction'),
+        blob(instruction.data.length - 1, 'amount'),
```
atharmohammad, line 78: Since length would be depending on the size of amount string , for decoding I assigned it data.length -1 , since u8 instruction would be taking 1 size, i checked it's if it's decoding correctly or not, is it fine or is there some better way for this

token/js/src/instructions/uiAmountToAmount.ts
```diff
@@ -0,0 +1,132 @@
...
+ * @return Decoded, non-validated instruction
+ */
+export function decodeUiAmountToAmountInstructionUnchecked({
+    programId,
+    keys: [mint],
+    data,
+}: TransactionInstruction): DecodedUiAmountToAmountInstructionUnchecked {
+    const uiAmountToAmountInstructionData = struct<UiAmountToAmountInstructionData>([
+        u8('instruction'),
+        blob(data.length - 1, 'amount'),
```
atharmohammad, line 123: same here

atharmohammad:
@joncinque is the test failing because of upgrading @solana/web3.js ?

joncinque:
@atharmohammad Ah yeah, that's an annoying bug -- since two packages depend on `@solana/web3.js`, when you update, you'll have two separate entries in `yarn.lock`.  You need to combine those to both resolve to the newer web3.js.

I should have noticed it earlier, but you'll also need to commit the change to `yarn.lock`

atharmohammad:
@joncinque So i have resolved web3 conflict issue, now i think test are failing because solana version for ci is ```v1.10.29``` whereas for ```returnData``` we need version ```v1.11.4``` so should i change solana version in ci ?

joncinque:
@atharmohammad you're correct -- I thought we had backported those changes to 1.10, but it isn't the case.  We'll upgrade SPL to 1.11 once the release branch gets branched off of master, so not just yet.  Which means we'll need to hold off on this a little bit longer.

2501babe:
@atharmohammad master is now on 1.11.6

joncinque (commented):
Looks great, just two little things, then we can get this merged!

token/js/src/actions/uiAmountToAmount.ts
```diff
@@ -0,0 +1,30 @@
...
+    payer: Signer,
+    mint: PublicKey,
+    amount: string,
+    programId = TOKEN_PROGRAM_ID
+): Promise<bigint | TransactionError | null> {
+    const transaction = new Transaction().add(createUiAmountToAmountInstruction(mint, amount, programId));
+    const { returnData, err } = (await connection.simulateTransaction(transaction, [payer], false)).value;
+    if (returnData) {
+        const data = Buffer.from(returnData.data[0], returnData.data[1]);
+        return u64().decode(data);
```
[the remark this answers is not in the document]
atharmohammad: it's not implementing ```decode```  without it

jordansexton:
The resolved conflicts in this PR undo a bunch of the changes from https://github.com/solana-labs/solana-program-library/pull/3508. This will need to be rebased against the current master.

joncinque:
@jordansexton was the rebase done correctly? Because otherwise this is ready to go from my side

token/js/package.json
```diff
@@ -49,7 +49,7 @@
         "deploy:docs": "npm run docs && gh-pages --dist token/js --dotfiles"
     },
     "peerDependencies": {
-        "@solana/web3.js": "^1.20.0"
+        "@solana/web3.js": "^1.47.4"
```
jordansexton, line 52: Is a reason these minimum versions changed? It doesn't matter hugely, but in general it seems ideal to declare a lower minimum version if it's not needed (1.20 is kind of a magic number I saw in the wild a bunch). Just wondering if this relies on some specific API that became available as of this version.

jordansexton (approved):
Thanks for rebasing! I added a commit to fix the .js file extensions in the imports and exports in the new code, and organize the imports. I just have one question about the web3.js version.

token/js/package.json
atharmohammad: so i was upgrading it because of ```returnData``` in ```SimulatedTransactionResponse``` which was added just a month ago , it was not showing before in lower version of ```web3.js``` but idk why when installing lower version now i checked it is having ```returnData``` so i think we can just let it be at ```1.20.0```. Maybe by mistake i saw something else at that time.

jordansexton: Ah, no that's a good point. If we need that API let's require a newish version that has it!

atharmohammad: i am actually confused why i am getting ```returnData``` in ```1.20.0```  even though here is the PR that merged for it https://github.com/solana-labs/solana/pull/26499 , do all merge appears in lower version also 🤔 or is there some problem with my build ?

jordansexton: So if your version is set with ^ (as it is in this PR's dev dependencies) that will still resolve to the most recent matching version of 1.x.x.

You can see what version is actually being resolved to in the package-lock.json file.

atharmohammad: oh yes, correct. Thanks for this 🙂 , spend a lot of time in figuring this out.

jordansexton: I think this is good to merge as is then?

atharmohammad: yes :)

jordansexton:
I'll release to npm tomorrow! Thanks for this!

jordansexton:
Published: @solana/spl-token@0.3.3

PR #3346, opened 2022-07-14
patches test_lamport_transfer

behzadnouri:
https://github.com/solana-labs/solana/pull/26606
prevents crediting a rent paying account if the account stays rent
paying. As a result, test_lamport_transfer fails.

The commit updates destination account lamports so that after transfer
it is no longer rent paying.


CriesofCarrots (approved):
Sure, this works for me, thanks!
Possibly we should come through and make all the accounts in these tests rent-exempt, but not necessary at the moment.

Issue #1000, opened 2020-12-31
lending: Add instruction for configuring reserve params


jstarry:
#### Problem
Reserve config values (LTV ratio, liquidation params, APY curve params, etc) can not be adjusted after the reserve is initialized.

#### Proposed Changes
Add authorized instruction for configuring reserves

joncinque:
Added with #3349

PR #3349, opened 2022-07-15
lending: Add instruction for changing reserve configuration after initialization

throwbackjams:
# Issue & Change Summary
**Issue**: Addresses [issue #1000](https://github.com/solana-labs/solana-program-library/issues/1000)
Reserve configuration values (LTV ratio, liquidation params, APY curve params, etc.) can not be adjusted after the reserve is initialized.

**Summary of changes**:

-  Created an instruction variant `ModifyReserveConfig`, with unpack and pack implementations
- In the processor, created `process_modify_reserve_config` to handle the instruction
- Wrote tests in `modify_reserve_config.rs` to test that the reserve config can be changed and only by the correct lending market owner

# Change Detail & Notes (Not required - one can also just examine code. Only read if additional context is helpful)
**Approach - Steps**:
1. Created `ModifyReserveConfig` instruction variant, with pack & unpack implementations, and created a function to create the instruction. The instruction requires 3 accounts: Reserve account(writable), Lending market account, Lending market owner(signer)
2. In the processor, created `process_modify_reserve_config` which:
    1. Checks the config parameters, with the same rules as those in `init_reserve`
    2. Checks that the provided lending market & owner is correct and corresponds to the reserve to be modified
    3. Changes the reserve state by replacing the old `ReserveConfig` within `Reserve` with the newly provided `ReserveConfig`
3. Refactor the config parameter checks (duplicated in both `process_init_reserve` and `process_modify_reserve_config`) into a helper function `validate_reserve_configs`. Modify both previous functions to use `validate_reserve_configs`
4. Wrote tests in `modify_reserve_config.rs`, which specifically test the following:
    1. Reserve configs can be changed with the correct lending market owner
    2. Reserve configs cannot be changed by a different lending market owner
    3. Reserve configs cannot be changed if the same owner provides a lending market that corresponds to a different `Reserve` state

**Note**: Tangentially, it seems that the `init_reserve.rs` tests (unmodified by this PR) are not passing. The error given is `InstructionError(8, NotEnoughAccountKeys)` by the MintTo call within `process_init_reserve`. However, I was not able to identify what account was missing. More [logs on the error here if helpful](https://gist.github.com/throwbackjams/47c598badc1700adc64c3443b409cf09)

### DIFF
--- a/token-lending/program/src/processor.rs
+++ b/token-lending/program/src/processor.rs
@@ -3,7 +3,7 @@
 use crate::{
     error::LendingError,
     instruction::LendingInstruction,
-    math::{Decimal, Rate, TryAdd, TryDiv, TryMul, WAD},
+    math::{Decimal, Rate, TryAdd, TryDiv, TryMul},
     pyth,
     state::{
         CalculateBorrowResult, CalculateLiquidationResult, CalculateRepayResult,
@@ -100,7 +100,7 @@ pub fn process_instruction(
             process_flash_loan(program_id, amount, accounts)
         }
         LendingInstruction::ModifyReserveConfig { new_config } => {
-            msg!("Instruction: Modify ReserveConfig");
+            msg!("Instruction: Modify Reserve Config");
             process_modify_reserve_config(program_id, new_config, accounts)
         }
     }
@@ -178,7 +178,7 @@ fn process_init_reserve(
         return Err(LendingError::InvalidAmount.into());
     }
 
-    validate_reserve_configs(&config)?;
+    config.validate()?;
 
     let account_info_iter = &mut accounts.iter().peekable();
     let source_liquidity_info = next_account_info(account_info_iter)?;
@@ -1671,7 +1671,7 @@ fn process_modify_reserve_config(
     new_config: ReserveConfig,
     accounts: &[AccountInfo],
 ) -> ProgramResult {
-    validate_reserve_configs(&new_config)?;
+    new_config.validate()?;
 
     let account_info_iter = &mut accounts.iter().peekable();
     let reserve_info = next_account_info(account_info_iter)?;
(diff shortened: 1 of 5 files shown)


token-lending/program/src/processor.rs
```diff
@@ -1821,6 +1836,50 @@ fn get_pyth_price(pyth_price_info: &AccountInfo, clock: &Clock) -> Result<Decima
     Ok(market_price)
 }
 
+/// Validate the reserve configs, when initializing or modifying the reserve configs
+fn validate_reserve_configs(config: &ReserveConfig) -> ProgramResult {
```
joncinque, line 1840: nit: let's rename this to `validate_reserve_config` to reflect the exact same type name.  Or we can add a `validate()` function to `ReserveConfig` directly, your choice!

token-lending/program/tests/modify_reserve_config.rs
```diff
@@ -0,0 +1,380 @@
...
+        .process_transaction(transaction)
+        .await
+        .map_err(|e| e.unwrap())
+        .unwrap();
+
+    let reserve_info = sol_test_reserve.get_state(&mut banks_client).await;
+    assert_eq!(
+        reserve_info.config.optimal_utilization_rate,
+        TEST_RESERVE_CONFIG.optimal_utilization_rate - OPTIMAL_UTILIZATION_RATE_CHANGE
+    );
```
joncinque, line 99: nit: how about testing the whole thing? ie
```suggestion
    assert_eq!(
        reserve_info.config,
        new_config
    );
```

token-lending/program/tests/modify_reserve_config.rs
```diff
@@ -0,0 +1,380 @@
...
+}
+
+#[tokio::test]
+#[should_panic(expected = "Transaction::sign failed with error KeypairPubkeyMismatch")]
+async fn wrong_signer_of_lending_market_cannot_change_reserve_config() {
```
joncinque, line 104: This test is checking the signing logic from the SDK, and not your signer check in the program.

How about changing the instruction to not have the owner as a signer, and check that it trips your signer check? e.g.
```
let mut instruction = modify_reserve_config(
            spl_token_lending::id(),
            new_config,
            sol_test_reserve.pubkey,
            lending_market.pubkey,
            lending_market.owner.pubkey(),
        );
instruction.account_metas[2].is_signer = false;
...
transaction.sign(&[&payer], recent_blockhash);
```

Then run the transaction and make sure that the correct error is tripped

token-lending/program/tests/modify_reserve_config.rs
```diff
@@ -0,0 +1,380 @@
...
+            0,
+            InstructionError::Custom(LendingError::InvalidAccountInput as u32)
+        )
+    );
+
+    let reserve_info = sol_test_reserve.get_state(&mut banks_client).await;
+    assert_eq!(
+        reserve_info.config.optimal_utilization_rate,
+        TEST_RESERVE_CONFIG.optimal_utilization_rate
+    );
```
joncinque, line 288: nit: full check here also
```suggestion
    assert_eq!(
        reserve_info.config,
        TEST_RESERVE_CONFIG
    );
```

token-lending/program/tests/modify_reserve_config.rs
```diff
@@ -0,0 +1,380 @@
...
```
joncinque, line 379: nit: full check
```suggestion
    assert_eq!(
        reserve_info.config,
        TEST_RESERVE_CONFIG
    );
```

token-lending/program/tests/modify_reserve_config.rs
```diff
@@ -0,0 +1,380 @@
...
+        },
+    };
+
+    let mut transaction = Transaction::new_with_payer(
+        &[modify_reserve_config(
+            spl_token_lending::id(),
+            new_config,
+            sol_test_reserve.pubkey,
+            other_lending_market.pubkey,
+            other_lending_market.owner.pubkey(),
```
joncinque, line 355: If you want to test wrong lending market with correct owner, shouldn't this be `lending_market.owner.pubkey()`?  Then you can trip `InvalidMarketOwner` in your test.

Otherwise, is this exactly the same as the previous test?

joncinque (commented):
This is a really great start! Most importantly, the program logic is correct, and the rest is mostly nits.

From a security perspective, there's a potential issue of someone refreshing obligations, updating the config, and then liquidating based on the earlier config (since it was updated that slot), but since this is a privileged operation, I think it'll be ok.

@jordansexton are there any high level issues with reconfiguring the reserve? ie. potential attacks or bits that might be out of sync

token-lending/program/tests/modify_reserve_config.rs
throwbackjams: So `lending_market.owner.pubkey() == other_lending_market.owner.pubkey()` due to the fact that the `add_lending_market` helper, which creates both lending markets in this test, uses a [static keypair file](https://github.com/solana-labs/solana-program-library/blob/d1e842ca3f5e1752a2e436e711d9fa09b486f119/token-lending/program/tests/helpers/mod.rs#L96) as the owner for every invocation. The error would still be `InvalidAccountInput` since the owner is indeed valid (the same) but the lending market is not.

But I agree with you - if line 355 in the above just uses `lending_market.owner.pubkey()` as the argument, then the test case of "correct owner but incorrect market" would be more clear. I'll make this change. Thank you!

Note: The previous test, `owner_of_different_lending_market...` should be testing the `InvalidMarketOwner` error. I'll modify the previous test accordingly.

throwbackjams: To summarize, the four (updated) tests should test:
1. happy path - right market, right owner can change ReserveConfig
2. invalid signer - right market, right owner but owner is not a signer -> error `InvalidSigner`
3. invalid owner - right market, wrong owner -> error `InvalidMarketOwner`
4. invalid market - wrong market, right owner -> error `InvalidAccountInput`

token-lending/program/src/processor.rs
throwbackjams: I like the method approach!

throwbackjams:
Thanks Jon! Really appreciate the feedback!

All of your suggestions should be incorporated in commit `a9d01a3` (with one clarifying note above for your question on the tests). Please let me know if you have any other thoughts!

token-lending/program/tests/modify_reserve_config.rs
joncinque: This is all perfect, well done!

joncinque:
Thanks for your contribution, great work!

PR #3353, opened 2022-07-18
token-cli: Move everything to async

joncinque:
#### Problem

We need to move the token cli to async in order to use the rust token-client with token-2022 extensions.

#### Solution

Move it all to async!  No functional changes here, but some of the tests were slightly rewritten to more easily slot in token-2022 later.

Note: the async TpuClient will only come in with 1.11, so `spl-token bench` is still blocking.


CriesofCarrots (approved):
Thank you for splitting this out!

PR #3354, opened 2022-07-18
token-cli: Figure out program_id dynamically

joncinque:
This is a reimagining of #3307 with separated commits to better understand the different pieces.

#### Problem

In order to more easily integrate with token-2022, the entire CLI needs to figure out the program id dynamically.

#### Solution

These changes were strongly coupled to https://github.com/solana-labs/solana-program-library/pull/3071, so I tried to decouple them as much as possible. The idea is that there aren't any functional changes, just figuring out the correct `program_id` based on the mentioned mint or token account, rather than using `config.program_id`. That'll help with the token-2022 support work.

### DIFF
--- a/token/cli/src/bench.rs
+++ b/token/cli/src/bench.rs
@@ -266,15 +266,15 @@ fn get_token_addresses_with_seed(
         .collect()
 }
 
-async fn is_valid_token(rpc_client: &RpcClient, token: &Pubkey) -> Result<(), Error> {
+async fn is_valid_token(rpc_client: &RpcClient, token: &Pubkey) -> Result<Pubkey, Error> {
     let mint_account = rpc_client
-        .get_account_data(token)
+        .get_account(token)
         .await
         .map_err(|err| format!("Token mint {} does not exist: {}", token, err))?;
 
-    Mint::unpack(&mint_account)
-        .map(|_| ())
-        .map_err(|err| format!("Invalid token mint {}: {}", token, err).into())
+    Mint::unpack(&mint_account.data)
+        .map_err(|err| format!("Invalid token mint {}: {}", token, err))?;
+    Ok(mint_account.owner)
 }
 
 async fn command_create_accounts(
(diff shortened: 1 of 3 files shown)


token/cli/src/bench.rs
```diff
@@ -266,15 +266,15 @@ fn get_token_addresses_with_seed(
         .collect()
 }
 
-async fn is_valid_token(rpc_client: &RpcClient, token: &Pubkey) -> Result<(), Error> {
+async fn is_valid_token(rpc_client: &RpcClient, token: &Pubkey) -> Result<Pubkey, Error> {
```
CriesofCarrots, line 259: Now that this method returns data, should we rename to be more self-descriptive? (I realize this is only used in `bench.rs`, so not a huge deal)

CriesofCarrots (dismissed):
Thanks for separating these out. Much easier to review!
Nitty nits only

token/cli/src/main.rs
```diff
@@ -3343,7 +3321,11 @@ mod tests {
             mint,
             ui_amount,
             recipient,
-            Some(native_mint::DECIMALS),
+            MintInfo {
+                program_id: config.program_id,
+                address: mint,
+                decimals: TEST_DECIMALS,
```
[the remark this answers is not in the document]
joncinque: This was actually a bug in the tests! We were assuming it was a mint with 0 decimals when looking at the output, but saying it was `native_mint::DECIMALS`.  So either we explicitly say 0 decimals, or change all of the outputs to be 9 decimals.  I opted for the easier path :smile:

token/cli/src/bench.rs
joncinque: You're totally right, I punted because I couldn't come up with a good name. Best I got is `get_program_id_from_valid_mint`  or `get_valid_mint_program_id` -- you?

token/cli/src/main.rs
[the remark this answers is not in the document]
CriesofCarrots: Oh, lol. Good to know!

token/cli/src/bench.rs
CriesofCarrots: I like `get_valid_mint_program_id` well enough :)

PR #3355, opened 2022-07-18
Issue #1805 - reverse logic for --recipient-is-ata-owner

SUMEETRM:
Pull request for issue #1805 

Retained --recipient-is-ata-owner as a hidden flag and, if passed, warn the user that the behavior is now default, the flag is deprecated and can be removed.

New flag called --no-recipient-is-ata-owner to match convention of other CLI tools. Reverse the logic for the flag.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -305,6 +305,18 @@ pub fn signers_of(
     }
 }
 
+fn check_decimals(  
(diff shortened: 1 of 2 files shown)


joncinque:
Could you rebase your branch onto master and also create a separate branch for this PR?  It looks like your changes for #3275 also got pulled in

PR #3357, opened 2022-07-19
Improve token error messages in token-swap

### LINKED ISSUE #822: token-swap: Token error messages are printed incorrectly

jmalcic:
This PR resolves #822 by wrapping the `invoke_signed` function with another function `invoke_signed_wrapper` that takes a generic. This allows a custom error type to be provided to `PrintProgramError::print<E>`, so that the `ProgramError` can be converted back to the custom type for logging.


joncinque (dismissed):
Great job!

joncinque:
Looks like a little CI failure on the tests, since you need to use `solana_program::program_error::PrintProgramError` in the processor tests.

joncinque:

PR #3362, opened 2022-07-20
token-cli: Make BPF program builds in build.rs dependent on env var

joncinque:
#### Problem

The token-cli `build.rs` helpfully builds dependent BPF programs required for tests (token, token-2022, and associated-token-account). It's currently behind a check for the `DEBUG` environment variable, which means that it only runs on debug builds.  This is fine for the most part, but `cargo build` and even `cargo publish` do debug builds before the publish, which means that we currently can't publish the token-cli crate.

#### Solution

Add a new environment variable for building the dependent programs.

I'm not very happy with this solution, but there are literally no special environment variables set during `cargo test` which are available to `build.rs`, to tell us if we're in a testing environment.

Other options are to eliminate the file entirely or add crate feature to build the programs.  The crate feature seemed worse to me, but I can be swayed.

cc @jstarry (sorry for not publishing just yet, will do once this is sorted)

### DIFF
--- a/ci/cargo-build-test.sh
+++ b/ci/cargo-build-test.sh
@@ -15,8 +15,8 @@ set -x
 make -C examples/c
 
 # Build/test all host crates
-cargo +"$rust_stable" build
-BUILD_DEPENDENT_PROGRAMS=1 cargo +"$rust_stable" test -- --nocapture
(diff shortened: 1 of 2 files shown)


CriesofCarrots (approved):
Doesn't seem terrible for now. Thanks for figuring this one out

PR #3368, opened 2022-07-22
Add IDL for governance program

gagliardetto:
This is a draft of an IDL for the governance program.

NOTES:
----------

- There might be some things that are in `accounts` and should go to `types` instead (and viceversa).
- Instruction accounts are generated from the comments (such as `// 0. `[writable]` The account to reallocate.`)
- Names of accounts could be improved (maybe too long, or not clear enough).

### DIFF
--- a/governance/idl.json
+++ b/governance/idl.json
@@ -892,12 +892,6 @@
 			"args": []
 		},
 		{
-			"name": "",
-			"docs": null,
-			"accounts": [],
-			"args": null
-		},
-		{
 			"name": "CancelProposal",
 			"docs": [
 				"Cancels Proposal by changing its state to Canceled"
@@ -945,12 +939,6 @@
 				}
 			],
 			"args": []
-		},
-		{
-			"name": "",
-			"docs": null,
-			"accounts": [],
-			"args": null
 		},
 		{
 			"name": "SignOffProposal",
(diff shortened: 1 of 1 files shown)


governance/idl.json
```diff
@@ -0,0 +1,4865 @@
...
+						]
+					}
+				]
+			},
+			"docs": [
+				"Program metadata account. It stores information about the particular SPL-Governance program instance"
+			]
+		},
+		{
+			"name": "ProposalOption",
```
ngundotra, line 2648: This should be in `types` not `accounts`

ngundotra:
@gagliardetto this looks great! Missing a few new accounts however:

- ProgramGovernanceV2
- MintGovernanceV2
- TokenGovernanceV2
- SignatoryRecordV2

I think it may be better to go with auto generating the IDL using something like [MPL's shank](https://github.com/metaplex-foundation/shank)

ngundotra:
Might be helpful to just annotate using Shank once this PR is in: https://github.com/metaplex-foundation/shank/pull/32

gagliardetto:
> Might be helpful to just annotate using Shank

Yep! Was thinking the same. It would avoid needing ad-hoc updates to the IDL json file, and remove the need to mentally understand what/how to update.

PR #3371, opened 2022-07-23
token-cli: Downgrade to crates.io versions for release

joncinque:
#### Problem

It's time to release a new version of the token CLI to crates.io, but since the current codebase is using newer versions of the libraries which are *not* on crates.io, we have to downgrade dependencies and update the code.

#### Solution

Downgrade dependencies and update code as needed.  I'll revert this commit once the release is done.


CriesofCarrots:
Are we going to have to do this every time we release token-cli?

joncinque:
No not every time -- this will get resolved once I release the new token and associated-token-account programs to mainnet along with their crates.  Someone needed a newer version of the cli released, so I opted to do the temporary downgrade instead of waiting.

PR #3373, opened 2022-07-25
Add Anchor IDL for token lending program

acheroncrypto:
I've added the IDL in the root directory of `token-lending`. 

Open to feedback! I'll add other IDLs as well if this format is appropriate.


acheroncrypto:
https://github.com/solana-labs/solana-program-library/pull/3468#issuecomment-1225122410

PR #3380, opened 2022-07-25
add_license_file

c3pko:


joncinque (approved):
May need to update the copyright year, but all good!

PR #3382, opened 2022-07-25
create-setup-file

c3pko:


joncinque:
I did some digging around, and it looks like there's a more general way to define packages for any build tool: https://packaging.python.org/en/latest/tutorials/packaging-projects/

Since it's on python.org, I'm inclined to trust it -- what do you think?

c3pko:
> I did some digging around, and it looks like there's a more general way to define packages for any build tool: https://packaging.python.org/en/latest/tutorials/packaging-projects/
> 
> Since it's on python.org, I'm inclined to trust it -- what do you think?

ok seems like the pyproject.toml replaces the setup.py file (https://stackoverflow.com/questions/62983756/what-is-pyproject-toml-file-for) so I will create that instead!

c3pko:
> > I did some digging around, and it looks like there's a more general way to define packages for any build tool: https://packaging.python.org/en/latest/tutorials/packaging-projects/
> > Since it's on python.org, I'm inclined to trust it -- what do you think?
> 
> ok seems like the pyproject.toml mostly replaces the setup.py file (https://stackoverflow.com/questions/62983756/what-is-pyproject-toml-file-for) so I will create that instead!

replaced the setup.py file with a pyproject.toml. added a few items--lmk if it looks good/you have suggestions on names/description.

On the packages front, I copied the example structure used here: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html. Note it does say package discovery is in beta for pyproject.toml.

c3pko:
@joncinque I got the latest stake-pool repo, added the pyproject.toml file and uploaded it to the testPyPi here: https://test.pypi.org/project/stake-pool/1.0/. If you could download it and test that it works as intended for you, that would be super helpful!

stake-pool/py/pyproject.toml
```diff
@@ -0,0 +1,67 @@
+[project]
+name = 'stake-pool'
```
joncinque, line 2: how about calling this `spl-stake-pool` to line up with other packages?

stake-pool/py/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+version = '1.0'
+authors = [
+  { name= 'joncinque'},
```
joncinque, line 5: Can this be the solana foundation? ie
```suggestion
  { name="Solana Maintainers", email="maintainers@solana.foundation" },
```
Actually, we use that address everywhere, and I'm not sure if it's even valid :laughing: could you check to see that it's still good?

stake-pool/py/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+]
+description = 'solana labs stake pool reference implementation'
```
joncinque, line 7: Since this is only python bindings and not the program, how about:
```suggestion
description = 'Solana Program Library Stake Pool Program Client Library'
```

stake-pool/py/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+        'toml==0.10.2',
+        'typing-extensions==3.10.0.2',
+        'urllib3==1.26.7'
+      ]
+
+[tool.setuptools]
+packages = ['bot', 'vote', 'stake', 'system', 'spl_token', 'stake_pool']
+
+[project.urls]
+'Homepage' = 'https://github.com/solana-labs/solana-program-library/new/master/stake-pool/py'
```
joncinque, line 58: This page actually creates a new file in the repo, so let's do https://github.com/solana-labs/solana-program-library/tree/master/stake-pool/py instead

stake-pool/py/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+        'packaging==21.2',
+        'pluggy==1.0.0',
+        'py==1.10.0',
+        'pycodestyle==2.8.0',
+        'pycparser==2.20',
+        'pyflakes==2.4.0',
+        'PyNaCl==1.4.0',
+        'pyparsing==2.4.7',
+        'pytest==6.2.5',
+        'pytest-asyncio==0.16.0',
```
joncinque, line 43: I recently bumped this, sorry!
```suggestion
        'pytest-asyncio==0.19.0',
```

stake-pool/py/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+description = 'solana labs stake pool reference implementation'
+readme = 'README.md'
+license = { file='LICENSE' }
+requires-python = '>=3.7'
+classifiers = [
+    'Programming Language :: Python :: 3',
+    'License :: OSI Approved :: Apache Software License',
+    'Operating System :: OS Independent',
+]
+dependencies = [
```
joncinque, line 16: It would be better to split these up into dependencies and optional-dependencies, to separate what's needed to use the package vs build / test it. But we can do that another time

stake-pool/py/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+'Homepage' = 'https://github.com/solana-labs/solana-program-library/new/master/stake-pool/py'
+
+[build-system]
+requires = ['setuptools>=61.0']
+build-backend = 'setuptools.build_meta'
+
+
+
+
+
```
joncinque, line 67: nit: can remove this whitespace

joncinque (commented):
Can you remove requirements.txt with this? You'll also need to update the CI  at https://github.com/solana-labs/solana-program-library/blob/674b0ac70e1a319b0d97e78a50b84f79b803be2a/ci/py-test-stake-pool.sh#L10 to do `pip3 install .` instead (I think)

stake-pool/py/pyproject.toml
c3pko: happy to add that if you can share which dependencies are optional

joncinque: For sure, I just need to do a little sweep of the dependencies, I'll also update a bunch of them while I'm at it

PR #3384, opened 2022-07-26
token-2022: Add intro docs, structure for more docs

joncinque:
#### Problem

Token-2022 exists, but few people know why or how to use it.

#### Solution

Add the intro docs for token-2022 and eventual structure for more.

### DIFF
--- a/docs/src/token-2022.md
+++ b/docs/src/token-2022.md
@@ -109,8 +109,8 @@ To get started with Token-2022:
 
 - [Install the Solana Tools](https://docs.solana.com/cli/install-solana-cli-tools)
 - [Extension Guide](token-2022/extensions.md)
-- [Wallet Migration Guide](token-2022/wallet-migration.md)
-- [On-Chain Program Migration Guide](token-2022/onchain-migration.md)
+- [Wallet Guide](token-2022/wallet.md)
+- [On-Chain Program Guide](token-2022/onchain.md)
 
 For existing functionality in the Token Program, see the [token docs](token.mdx).
 The Token functionality will always apply to Token-2022.
--- a/docs/src/token-2022.md
+++ b/docs/src/token-2022.md
@@ -1,0 +1,129 @@
+---
+title: Token-2022 Program
+---
+
+A token program on the Solana blockchain, defining a common implementation for
+fungible and non-fungible tokens.
+
+The Token-2022 Program is a superset of functionality provided by the
+[Token Program](token.mdx), deployed to all networks.
+
+| Information | Account Address |
+| --- | --- |
+| Token-2022 Program | `TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb` |
+
+## Motivation
+
+The existing Token Program serves most needs for fungible and non-fungible tokens
+on Solana through a simple set of interfaces and structures. It has been rigorously
+audited since its initial deployment in 2020.
+
+As more developers have come to Solana with new ideas, however, they have forked the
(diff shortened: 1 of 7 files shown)


docs/src/token-2022.md
```diff
@@ -0,0 +1,129 @@
...
+The Token-2022 Program's source is available on
+[GitHub](https://github.com/solana-labs/solana-program-library/tree/master/token/program-2022).
+
+For information about the types and instructions, the Rust docs are available at
+[docs.rs](https://docs.rs/spl-token-2022/latest/spl_token_2022/).
+
+## Security Audits
+
+The Token-2022 Program is currently under multiple audits to ensure safety of
+funds. all audits will be published here as they are completed.
```
CriesofCarrots, line 129:
```suggestion
funds. All audits will be published here as they are completed.
```

docs/src/token-2022.md
```diff
@@ -0,0 +1,129 @@
+---
+title: Token-2022 Program
+---
+
+A token program on the Solana blockchain, defining a common implementation for
+fungible and non-fungible tokens.
+
+The Token-2022 Program is a superset of functionality provided by the
```
CriesofCarrots, line 8:
```suggestion
The Token-2022 Program is a superset of the functionality provided by the
```

docs/src/token-2022.md
```diff
@@ -0,0 +1,129 @@
...
+New fields are added in the form of extensions.
+
+Mint creators and account owners can opt-in to Token-2022 features. Extension data
+is written after the end of the `Account` in Token, which is the byte at index
+`165`.  This means it is always possible to differentiate mints and accounts.
+
+You can read more about how this is done at the
+[source code](https://github.com/solana-labs/solana-program-library/tree/master/token/program-2022/src/extensions/mod.rs).
+
+Mint extensions include:
```
CriesofCarrots, line 86:
```suggestion
Mint extensions currently include:
```
? Not sure how open we want to leave the door for future extensions. Some of the text above sort of implies new functionality will be on-going.

CriesofCarrots (approved):
Looks like a great starting place to me! Couple nits

docs/src/token-2022.md
joncinque: Yeah that makes sense to me

docs/src/token-2022/onchain-migration.md
```diff
@@ -0,0 +1,5 @@
+---
+title: On-chain Program Migration Guide
```
jacobcreech, line 2: For both this and `wallet-migration`, are they really considered migrations or additions?

docs/src/token-2022/extensions.md
```diff
@@ -0,0 +1,5 @@
+---
+title: Extension Guide
```
jacobcreech, line 2: Is the plan to have this guide on usage of each of the different extensions within token-2022? Are these added to the CLI?

joncinque: Correct, and no, they aren't added to the CLI yet, just JS

docs/src/token-2022/onchain-migration.md
joncinque: Good point... let's see, if you're on-chain, you need to deserialize accounts differently if you want to support both and use a different library, spl-token-2022 instead of spl-token, so migration makes sense there.

For wallets though, it would be more of an addition, so I can rename that to "Wallet Support Guide".  What do you think?

jacobcreech: I think `Wallet Guide` and `On-chain Program Guide` sounds good 👍

PR #3385, opened 2022-07-27
Bump solana crates to v1.10.33

CriesofCarrots:
In order to pick up [these implementations](https://github.com/solana-labs/solana/pull/26784) in the monorepo, we need a new spl-token-2022 with them. This PR does the necessary solana-crate bump.

~~...this is still WIP, because locally it looked like pulling in the new zk-token-sdk caused one of the confidential-transfer tests to fail~~ Passed here, so I guess the issue was mine; maybe stale build.


CriesofCarrots:
Incidentally, `ci/py-test-stake-pool.sh` hung, ie ran for 4h23m until I canceled it.

joncinque (approved):
Looks good, and no worries about the python test failure, but that is strange...

PR #3386, opened 2022-07-27
token-cli: Add token-client dependency for token-2022 support

joncinque:
#### Problem

We want to integrate token-2022 extensions into the token-cli, and a lot of the work is already done in the token-client.

#### Solution

Add a `program_client` field to the token cli config, which can be used when using token-2022 extensions or instructions.

This uses `config.program_client` just to get rent exemptions, which was an easy way to show that the new field exists and can be used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6285,6 +6285,7 @@ dependencies = [
  "spl-memo 3.0.1",
  "spl-token 3.3.0",
  "spl-token-2022 0.4.2",
+ "spl-token-client",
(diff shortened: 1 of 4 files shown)


CriesofCarrots (approved):

CriesofCarrots:
Sorry for the conflict!

samkim-crypto (dismissed):

joncinque:
Merging on red, the twotx failure can be sorted out separately

PR #3387, opened 2022-07-27
docs: Fix links

joncinque:
#### Problem

When adding the new token docs, some of the filenames were changed, but not in the sidebars, causing a build error.

#### Solution

Fix the links, and in the future, consider doing a docs build in CI for PRs on docs. I thought this was happening already, but maybe we need to add one more step.


joncinque:
Will fix up the CI in a separate PR, landing this now since the build failure is on master.

PR #3396, opened 2022-07-29
token-cli: Run tests in batches

joncinque:
#### Problem

The token cli tests have become flaky, as an outcome of running so many test validators in parallel.

#### Solution

Run the tests in two big batches using `serial_test::parallel`.  If needed, we can instead do it in three batches.  This seemed like an easy solution, but does require setting a new `parallel` for each test added, which could be annoying.  Alternatives are welcome!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4217,7 +4217,20 @@ checksum = "e0bccbcf40c8938196944a3da0e133e031a33f4d6b72db3bda3cc556e361905d"
 dependencies = [
  "lazy_static",
  "parking_lot 0.11.2",
- "serial_test_derive",
+ "serial_test_derive 0.5.1",
+]
+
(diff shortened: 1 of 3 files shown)


CriesofCarrots:
Are there test assumptions we could change to make the tests not interfere with each other instead?
I've no strong opposition to this, but curious which approach would be less brittle.

joncinque:
The tests don't really interfere with each other, it's just depleting resources on the system running it, which causes errors.  An alternative is bundling the tests together so each one tests more conditions, but this could still break in the future if we add too many tests.

CriesofCarrots:
Ohh, that kind of flaky. Okay, let's roll with this, then

CriesofCarrots (approved):

PR #3399, opened 2022-08-01
Confidential transfer encryption doc

samkim-crypto:
Adding SPL doc for the confidential extension encryption part.


joncinque (commented):
Great read, nice job!

PR #3410, opened 2022-08-02
token-2022: merge instruction data for confidential Transfer and TransferWithFee

samkim-crypto:
The instruction data that is needed for `Transfer` and `TransferWithFee` is the same since the additional expected fee amount in `TransferWithFee` is encrypted and included in the proof verification instruction.

Currently, we have two separate structs for these two instructions. This pr merges these two data structs.


samkim-crypto:
Merging on red. Hopefully, the twotx failure can be sorted out separately.

PR #3412, opened 2022-08-02
stake-pool-py: Clean up dependencies

joncinque:
#### Problem

We want to separate the dependencies to avoid bundling too much in the stake pool python package.

#### Solution

Split up the deps between `requirements.txt` and `optional-requirements.txt`.  @c3pko You can use everything in `requirements.txt` as the `dependencies`, and everything in `optional-requirements.txt` in `optional-dependencies.  After this is merged, you can rebase your PR, and remove both of these requirements files.


joncinque:
The stake pool python tests passed, and that's what matters most -- merging on red!

PR #3415, opened 2022-08-03
token: Reassign and reallocate accounts on close

joncinque:
#### Problem

Closing token accounts sets data to 0, but it's still possible to re-use closed accounts by initializing again in the same transaction.

#### Solution

Completely close the account by reassigning it back to the system program and reallocating the account to 0, as if it were totally new.


joncinque:
This contains reverts for #3416 and #3417 in order to match what's currently deployed to mainnet.  Once this lands, we'll reapply those patches.

PR #3416, opened 2022-08-03
guards

aeyakovenko:


CriesofCarrots:
I'll merge early, but I want to see the all token tests run here or locally first

steviez:
Just checked - all other instances `.split_at()` in this repo have length checks before actually doing the length check

CriesofCarrots:
`twotx` CI failure is known and unrelated

token/program/src/instruction.rs
```diff
@@ -1689,4 +1704,12 @@ mod test {
         let unpacked = TokenInstruction::unpack(&expect).unwrap();
         assert_eq!(unpacked, check);
     }
+
+    #[test]
+    fn test_instruction_unpack_panic() {
```
mvines, line 1709: As a follow-up, perhaps something more exhaustive?
```rust
    #[test]
    fn test_instruction_unpack_panic() {
        for i in 0..255u8 {
            for j in 1..10 {
                let mut data = vec![0;j];
                data[0] = i;
                let _no_panic = TokenInstruction::unpack(&data);
            }
        }
    }
```

CriesofCarrots: Yep, on my list already :)

aeyakovenko: I am fuzzing it with 
```
    #[test]
    fn test_instruction_unpack_fuzz() {
        for _ in 0..10000000 {
            let r: usize = rand::thread_rng().gen_range(0..10);
            let data: Vec<u8> = (0..r).map(|_| rand::thread_rng().gen()).collect();
            _ = TokenInstruction::unpack(&data);
        }
    }
 ```

PR #3417, opened 2022-08-03
Refactor unpack and make test more robust

CriesofCarrots:
#3416 added some guard rails, but we already have an elegant way of doing this in token-2022.
This PR refactors those checks to match, and makes the unpack unit test exhaustive.

### DIFF
--- a/token/program/src/instruction.rs
+++ b/token/program/src/instruction.rs
@@ -1677,8 +1677,11 @@ mod test {
     #[test]
     fn test_instruction_unpack_panic() {
         for i in 0..255u8 {
-            let expect = Vec::from([i, 1, 0, 0, 0, 0, 0, 0, 0, 2]);
-            _ = TokenInstruction::unpack(&expect[0..2]);
+            for j in 1..10 {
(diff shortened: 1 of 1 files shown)


token/program/src/instruction.rs
```diff
@@ -1708,8 +1677,11 @@ mod test {
     #[test]
     fn test_instruction_unpack_panic() {
         for i in 0..255u8 {
-            let expect = Vec::from([i, 1, 0, 0, 0, 0, 0, 0, 0, 2]);
-            _ = TokenInstruction::unpack(&expect[0..2]);
+            for j in 1..10 {
+                let mut data = vec![0;j];
```
mvines, line 1681:
```suggestion
                let mut data = vec![0; j];
```

mvines (approved):

token/program/src/instruction.rs
CriesofCarrots: ty

CriesofCarrots:
Will bump spl-token to v3.3.1 before merging this

PR #3422, opened 2022-08-03
ci: Separate twoxtx build step, match Rust version

joncinque:
#### Problem

The token CI is failing on the twoxtx step, because spl is still using Rust 1.59.0, but the monorepo is at 1.60.0, and depends on features stabilized in 1.60.0

#### Solution

Split out the twoxtx build step, and use the correct version of Rust there.

### DIFF
--- a/.github/workflows/pull-request-token.yml
+++ b/.github/workflows/pull-request-token.yml
@@ -80,6 +80,8 @@ jobs:
           if-no-files-found: error
 
   cargo-test-bpf-twoxtx:
+    runs-on: ubuntu-latest
+    steps:
       - uses: actions/checkout@v2
 
       - name: Set env vars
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request-token.yml
```diff
@@ -82,6 +79,43 @@ jobs:
...
           if-no-files-found: error
 
+  cargo-test-bpf-twoxtx:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set env vars
+        run: |
+          echo "RUST_STABLE_VERSION=1.60.0" >> $GITHUB_ENV
```
CriesofCarrots, line 89: How is this used?

CriesofCarrots: Ahhh, thanks. I should have just looked at that file.
I wonder if we can find a way to pull this from `solana` instead of having to hard code it. Let's go with this for now, though, to unblock CI and set up the separate job

CriesofCarrots (approved):

PR #3423, opened 2022-08-03
Add slippage constraint to borrow instruction

### LINKED ISSUE #988: lending: Add slippage constraint for borrow instruction
#### Problem
When processing a borrow instruction which is denoted by deposited collateral, there is currently no way to specify the minimum number of borrowed tokens to receive.

Processing borrows denoted by requested borrow tokens does not need an explicit slippage constraint because collateral tokens can be restricted by using SPL token's delegate instruction. But it seems nice to have a slippage constraint for this direction too.

#### Proposed Changes
Add minimum amount to borrow instruction params

jmalcic:
This PR resolves #988 by adding a slippage constraint to the borrow instruction. As outlined in #1441 this only applies if the borrow is 100% of available borrowing power (`u64::MAX`).


token-lending/program/src/instruction.rs
```diff
@@ -219,7 +219,8 @@ pub enum LendingInstruction {
     BorrowObligationLiquidity {
         /// Amount of liquidity to borrow - u64::MAX for 100% of borrowing power
         liquidity_amount: u64,
-        // @TODO: slippage constraint - https://git.io/JmV67
+        /// Minimum amount of liquidity to receive, if borrowing 100% of borrowing power
+        slippage_limit: u64,
     },
```
joncinque, line 224: How about separating this into two instructions? We've got `BorrowObligationLiquidity` to specify exactly how much to borrow, and then a new instruction like `BorrowAllObligationLiquidity` which uses 100% of borrowing power and specifies just a slippage.  In the implementation, you call out to the same processor function, but this allows for clearer params, similar to how `Transfer` and `TransferChecked` work in the token program. What do you think?

token-lending/program/tests/borrow_obligation_liquidity.rs
```diff
@@ -408,3 +411,351 @@ async fn test_borrow_too_large() {
         )
     );
 }
+
+#[tokio::test]
+async fn test_borrow_max_receive_less_than_slippage() {
```
joncinque, line 416: nit: is this supposed to be `test_borrow_max_receive_more_than_slippage`?

token-lending/program/tests/borrow_obligation_liquidity.rs
```diff
@@ -408,3 +411,351 @@ async fn test_borrow_too_large() {
...
+
+    const FEE_AMOUNT: u64 = 5000;
+    const HOST_FEE_AMOUNT: u64 = 1000;
+
+    const USDC_DEPOSIT_AMOUNT_FRACTIONAL: u64 =
+        2_000 * FRACTIONAL_TO_USDC * INITIAL_COLLATERAL_RATIO;
+    const SOL_BORROW_AMOUNT_LAMPORTS: u64 = 50 * LAMPORTS_TO_SOL;
+    const USDC_RESERVE_COLLATERAL_FRACTIONAL: u64 = 2 * USDC_DEPOSIT_AMOUNT_FRACTIONAL;
+    const SOL_RESERVE_LIQUIDITY_LAMPORTS: u64 = 2 * SOL_BORROW_AMOUNT_LAMPORTS;
+    const SLIPPAGE_LIMIT: u64 = SOL_BORROW_AMOUNT_LAMPORTS - FEE_AMOUNT - 1;
```
joncinque, line 434: How about making this exactly at the limit to test your `<` condition, so instead this becomes:
```suggestion
    const SLIPPAGE_LIMIT: u64 = SOL_BORROW_AMOUNT_LAMPORTS - FEE_AMOUNT;
```
Then the test name would be `test_borrow_max_receive_minimum` or something

joncinque (commented):
The logic and tests look great! Let me know what you think about separating into two instructions.  I know it's a bit more work, but it'll be much clearer for users

token-lending/program/src/instruction.rs
```diff
@@ -217,12 +217,35 @@ pub enum LendingInstruction {
...
+    ///   7. `[signer]` Obligation owner.
+    ///   8. `[]` Clock sysvar.
+    ///   9. `[]` Token program id.
+    ///   10 `[optional, writable]` Host fee receiver account.
+    BorrowAllObligationLiquidity {
+        /// Minimum amount of liquidity to receive as a whole number
+        slippage_limit: u64,
+    },
+
+    // 12
```
jordansexton, line 248: This changes the index of all the subsequent instructions, a breaking change. I think new instructions should be appended?

However, I _think_ I disagree that this should be a separate instruction at all, since `liquidity_amount == u64::MAX` is already special-cased in the implementation.

This makes it possible to call `BorrowObligationLiquidity` with `u64::MAX` and it has different behavior than `BorrowAllObligationLiquidity`  while having the same semantic intent.

jordansexton: > I didn't think we needed to worry about breaking changes for a reference implementation

This is reasonable, I don't have a strong opinion, and it does keep the code cleaner this way. The only concern I have is that it makes it harder for downstream forks, because it's a breaking change for them. But I don't know if any of the major ones that started as forks (Solend, Port, Larix, Tulip, etc) actually pull anything from here.

> The idea is that special-casing `u64::MAX` at the borrow instruction is unclear.

Also reasonable, this design choice is owing to me not knowing what I'm doing so trying to keep things simple 🙃

joncinque: > But I don't know if any of the major ones that started as forks (Solend, Port, Larix, Tulip, etc) actually pull anything from here.

Ah right, in that case, we can do them a solid and add it to the end.

> Also reasonable, this design choice is owing to me not knowing what I'm doing so trying to keep things simple

Too modest! What do you think is the best approach to make this clearer? Note that "keeping it as it is" is a valid answer :smile:

jordansexton: > we can do them a solid and add it to the end.

Side note -- is there a way to manually set the enum "keys" so that it can appear lexically in order but use the index at the end?

> What do you think is the best approach to make this clearer?

I like _the idea_ of not using `u64::MAX` if there's a better option, though it's probably not worth changing. Keep it as is I guess!

I personally prefer the original implementation without adding an instruction:
https://github.com/jmalcic/solana-program-library/blob/63b6e9a307487c86c06f5b9a8fb5ecb846d89602/token-lending/program/src/processor.rs#L1153-L1156

Since this is the only business logic change, adding an instruction + creator to get to the same processor code path is a bigger change than just adding an arg to the same instruction.

I would probably approach this by using the same instruction, and making this optional: https://github.com/jmalcic/solana-program-library/blob/63b6e9a307487c86c06f5b9a8fb5ecb846d89602/token-lending/program/src/instruction.rs#L375

Parse a u64 if it's there, and if not, default to 0. This makes it a nonbreaking change for the instruction data.

jmalcic: Okay, I've reverted the splitting of the instruction and made the parsing of the `u64` optional. https://github.com/solana-labs/solana-program-library/blob/ede3bdfdc993b07ec7fc5fb126f41253b117b680/token-lending/program/src/instruction.rs#L375

token-lending/program/tests/borrow_obligation_liquidity.rs
```diff
@@ -408,3 +411,351 @@ async fn test_borrow_too_large() {
...
+
+    // limit to track compute unit increase
+    test.set_compute_max_units(60_000);
+
+    const USDC_DEPOSIT_AMOUNT_FRACTIONAL: u64 =
+        2_000 * FRACTIONAL_TO_USDC * INITIAL_COLLATERAL_RATIO;
+    const SOL_BORROW_AMOUNT_LAMPORTS: u64 = 50 * LAMPORTS_TO_SOL;
+    const USDC_RESERVE_COLLATERAL_FRACTIONAL: u64 = 2 * USDC_DEPOSIT_AMOUNT_FRACTIONAL;
+    const SOL_RESERVE_LIQUIDITY_LAMPORTS: u64 = 2 * SOL_BORROW_AMOUNT_LAMPORTS;
+    const SLIPPAGE_LIMIT: u64 = SOL_BORROW_AMOUNT_LAMPORTS + 1;
```
joncinque, line 574: To be at the limit, should this be `SOL_BORROW_AMOUNT_LAMPORTS - FEE_AMOUNT + 1`?

token-lending/program/tests/borrow_obligation_liquidity.rs
joncinque: Sorry, it looks like this one got missed -- is it possible for this to be the exact limit?

token-lending/program/tests/borrow_obligation_liquidity.rs
```diff
@@ -408,3 +411,351 @@ async fn test_borrow_too_large() {
...
+                Some(usdc_test_reserve.liquidity_host_pubkey),
+            ),
+        ],
+        Some(&payer.pubkey()),
+    );
+
+    transaction.sign(&[&payer, &user_accounts_owner], recent_blockhash);
+
+    // check that transaction succeeds
+    assert!(banks_client.process_transaction(transaction).await.is_ok());
```
joncinque, line 760: nit: same as the in the other PR, in order to immediately surface the error when running the tests, we should do `banks_client.process_transaction(transaction).await.unwrap()`

joncinque (commented):
Just a couple of little bits, then this is good to go!

token-lending/program/tests/borrow_obligation_liquidity.rs
jmalcic: Whoops, sorry! 😅

joncinque:
And thanks also for this one, nice job!

PR #3424, opened 2022-08-03
token-cli: don't error when missing default config

2501babe:
fixes a regression introduced in #3166. prior to that, the config argument would default to the value of `solana_cli_config::CONFIG_FILE`, and loading the config would fall back to a default if it failed. changing the to an error protected against a missing file provided by `--config`, but also errored in the default case if they lacked the default file

this change distinguishes the three cases, erroring if the user provided a bad `--config` option, but falling back to defaults if they didnt

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1848,20 +1848,15 @@ fn app<'a, 'b>(
         .about(crate_description!())
         .version(crate_version!())
         .setting(AppSettings::SubcommandRequiredElseHelp)
-        .arg({
-            let arg = Arg::with_name("config_file")
+        .arg(
+            Arg::with_name("config_file")
                 .short("C")
                 .long("config")
                 .value_name("PATH")
                 .takes_value(true)
                 .global(true)
-                .help("Configuration file to use");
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Nice catch, thanks

joncinque:
of course it fails right before your new test... you can rebase on top of #3425, if you want to take your chances

joncinque:
#3425 landed, so if you don't mind rebasing once more, we can be sure that this is good to go

2501babe:
taking the cowards way out and deleting the test i attempted to add, it turns out `spl-token address` (and, in fact, every non-help command) attempts to load the signer keypair from its expected path and fails if its not there, and i cant tell it to use a different path in a test with a non-default config because thats... what the config does. and i dont want to write a key out to the default location and pollute a shared resource, if later tests end up depending on something else being there

there should be more comprehensive integration tests for the binary (pre this patch it crashed unconditionally on my machine) but thats a different thing imo

joncinque:
I think #3293 will at least address the config issue.  Definitely happy to come up with a better testing strategy!

PR #3425, opened 2022-08-03
token-cli: Split out CI into separate step, run tests in serial

joncinque:
#### Problem

The token-cli tests keep failing CI.

#### Solution

Serialize all the tests.  This takes a lot longer, so also split off the tests into a separate CI step.


CriesofCarrots (approved):
lgtm! We can readdress if the time to complete ends up feeling too painful

PR #3428, opened 2022-08-04
fix reference to the field hold_up_time in the governance README

Arrowana:
https://github.com/solana-labs/solana-program-library/blob/3d3b32dcf6b64bfe609600735d067ebe735ed36d/governance/program/src/state/proposal_transaction.rs#L100-L101


joncinque:
Thanks!

PR #3436, opened 2022-08-05
token-upgrade: Define new conversion program

joncinque:
#### Problem

If people want to move fungible tokens from token to token-2022, they'll need a way to trustlessly destroy the old tokens and create the new ones.  This can come in many forms, but it would be a good gesture towards the community to provide a reference implementation that can be tweaked.

#### Solution

Let's get it started with state + interface definitions!  The basic concept goes:

* create token-2022 mint
* set mint authority to the defined PDA
* create the factory
* let people upgrade
* reclaim mint authority when you need to mint tokens yourself

PDAs were chosen to have only one factory per mint, that *cannot* be closed / recreated afterwards, this way you lock in the old mint to always point to the new mint.  And it's easy to discover the new mint.

Also, since factories can't be closed, after you "reclaim" mint authority, you can always assign it back to the factory and let it work once again, giving lots of flexibility.


jordansexton:
Mind if I review this whenever it's ready? I've been thinking about how people will handle this, so it'll be cool to look into your implementation.

joncinque:
> Mind if I review this whenever it's ready? I've been thinking about how people will handle this, so it'll be cool to look into your implementation.

Absolutely! I want to get the interface + state merged with this before starting on the implementation, so all comments are appreciated

mvines:
Relinquishing the mint authority is a little sucky, even though it can be reclaimed. There could be another way:
USDC tends to "premint" tokens because they don't want to be living in the faraday cage all day as new end-user "mint" requests come in. But "preminting" screws up their circulating supply so what's on explorer.solana.com doesn't always match their dashboards. If we formalized the concept of preminting in token22 then that could work here as well:
1. premint a bunch of new tokens to give to the upgrader
2. upgrader burns the old tokens as it swaps them for the new tokens
Circulating supply is now always correct and you get to keep your mint authority!

joncinque:
Ah, the preminting idea is really cool, and nicely gets around the biggest issue with this approach.  It'll probably be a better solution for everyone.  I'm sold.

My main and very specific question: does the program enforce that the provided bag has `old_mint.supply` tokens inside of it to be fair? I don't think we need that restriction at all, since old tokens could always be minted, and new tokens could be sent to the bag as needed.

mvines:
Yeah I dont think the program needs to care at all

CriesofCarrots:
> If we formalized the concept of preminting in token22

Is a `Premint` an extension on a mint that holds a u64 `amount` and an authority?

mvines:
> > If we formalized the concept of preminting in token22
> 
> Is a `Premint` an extension on a mint that holds a u64 `amount` and an authority?

heh mebbe. needs design!

A normal token account with a non-circulating bit set on it could be another way, but that may be too invasive of a change.

joncinque:
I'm a fan of the Premint living on the mint with some other authority that can be a hot wallet or PDA.  And the extra preminted amount doesn't count against the supply until it's transferred from the mint.  We'll have to take a write lock on the mint to update the supply anyway, so it seems ok to store that extra amount on the mint too.

mvines:
The main downside using the Mint account is that it forces a restriction of a single premint bucket. That's my only concern about this approach.

A "non-circulating" bit on a Token account is slightly more flexible in this way but certainly introduces additional edge cases in the code to prevent transfers between circulating/non-circulating accounts. But in this case we could still have a Mint extension to track the "total supply", plus instructions to flag an account as circulating (and perhaps non-circulating?)

joncinque:
I've updated the proposal to work through a token account instead, which means that the pre-minted extension on an account could happen later and not impact this at all, which feels nice.

token-upgrade/program/src/instruction.rs
```diff
@@ -0,0 +1,48 @@
...
+    /// Creates an upgrade factory for the given mint.
+    ///
+    /// Returns an error if the account exists.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writeable,signer]` Funding account (must be a system account)
+    ///   1. `[writeable]` PDA factory account to be created
+    ///   2. `[]` Source token mint
+    ///   3. `[signer]` Source token mint authority
```
mvines, line 18: Why does the mint authority need to be a signer here?

token-upgrade/program/src/instruction.rs
```diff
@@ -0,0 +1,48 @@
...
+    ///
+    CreateFactory,
+    /// Burns all of the source tokens in the user's account, and transfers the same
+    /// amount of pre-minted tokens into another account.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writeable]` Source token account to burn from
+    ///   1. `[writeable]` Source token mint
+    ///   2. `[writeable]` Pre-minted token account
```
mvines, line 33: What's this account?

token-upgrade/program/src/instruction.rs
```diff
@@ -0,0 +1,48 @@
...
+#[repr(u8)]
+pub enum TokenUpgradeInstruction {
+    /// Creates an upgrade factory for the given mint.
+    ///
+    /// Returns an error if the account exists.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writeable,signer]` Funding account (must be a system account)
+    ///   1. `[writeable]` PDA factory account to be created
```
mvines, line 16: Maybe this account can be avoided entirely since it only seems to hold `pre_minted_token_account`, which is already being passed into `UpdateTokens`?

token-upgrade/program/src/instruction.rs
```diff
@@ -0,0 +1,48 @@
...
+    ///   0. `[writeable]` Source token account to burn from
+    ///   1. `[writeable]` Source token mint
+    ///   2. `[writeable]` Pre-minted token account
+    ///   3. `[writeable]` Destination token account to transfer into
+    ///   4. `[]` Destination token mint
+    ///   5. `[]` PDA upgrade factory account for the source mint, given by:
+    ///        `get_factory_address(source_mint, program_id)`. Must be initialized.
+    ///   6. `[signer]` Source token account transfer authority (owner or delegate)
+    ///   7. `[]` PDA token account authority for the factory, given by:
+    ///        `get_factory_token_account_authority_address(factory_address, program_id)`
```
mvines, line 40: Nit: I'd drop the `PDA` prefix here.  That's already implied by the references to `get_factory_address()`/`get_factory_token_account_authority_address()`

token-upgrade/program/src/instruction.rs
joncinque: That's a good point, maybe the entire thing is stateless, and literally just does a burn on one side, and a transfer on the other. We simply define a standard for how to derive the owner of the pre-minted token account, as `PDA(old_mint, new_mint)`, so that way you can always find the accounts to use as long as you know the two mint addresses.

The entire idea of the factory was to give a pointer from the old mint to the new mint, so you can easily discover the new mint.  And to avoid scams, every mint only has one factory, so there's only one possible upgrade path. Since there's only one factory, we force the mint authority to sign to avoid address squatting.

But the more we talk about this, the more I realize that this is trying too hard, and we're better off making it entirely stateless.  What do you think?

mvines: Yep! Glad we arrived at the same place independently.

joncinque:
Ok, this should be good for another pass. It's much simpler now since the whole thing is stateless, so let me know if you need any more info anywhere

token-upgrade/program/src/instruction.rs
```diff
@@ -0,0 +1,30 @@
...
+pub enum TokenUpgradeInstruction {
+    /// Burns all of the source tokens in the user's account, and transfers the same
+    /// amount of tokens from an account owned by a PDA into another account.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writeable]` Source token account to burn from
+    ///   1. `[writeable]` Source token mint
+    ///   2. `[writeable]` Destination token upgrade account, owned by:
+    ///       `get_token_upgrade_authority_address(source_mint, destination_mint, program_id)`
```
mvines, line 17:
```suggestion
    ///   2. `[writeable]` Fount of destination tokens held in program escrow
```

token-upgrade/program/src/instruction.rs
```diff
@@ -0,0 +1,30 @@
...
+    ///   3. `[writeable]` Destination token account to transfer into
+    ///   4. `[]` Destination token mint
+    ///   5. `[signer]` Source token account transfer authority (owner or delegate)
+    ///   6. `[]` PDA owner of token upgrade account, given by:
+    ///       `get_token_upgrade_authority_address(source_mint, destination_mint, program_id)`
```
mvines, line 22:
```suggestion
    ///   6. `[]` Transfer authority (owner or delegate) of destination tokens held in program escrow, must be:
    ///       `get_token_upgrade_authority_address(source_mint, destination_mint, program_id)`
```

Specifically I think this'll still work if the program is only a delegate, which gives the owner a nice way to yank back any remaining tokens from the program if ever needed

token-upgrade/program/src/instruction.rs
mvines: I found the term "token upgrade account" a little confusing so I'm trying to work around that a little here

token-upgrade/program/src/instruction.rs
```diff
@@ -0,0 +1,30 @@
...
+    ///   7. `[]` SPL Token program for source mint
+    ///   8. `[]` SPL Token program for destination mint
+    ///
+    /// Data expected by this instruction:
+    ///   None
+    ///
+    UpgradeTokens,
```
mvines, line 29: perhaps
```suggestion
    Exchange,
```
or just `Update`?

I'm bothered by how `TokenUpgradeInstruction::UpgradeTokens` repeats Token

token-upgrade/program/src/instruction.rs
```diff
@@ -0,0 +1,30 @@
...
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writeable]` Source token account to burn from
+    ///   1. `[writeable]` Source token mint
+    ///   2. `[writeable]` Destination token upgrade account, owned by:
+    ///       `get_token_upgrade_authority_address(source_mint, destination_mint, program_id)`
+    ///   3. `[writeable]` Destination token account to transfer into
+    ///   4. `[]` Destination token mint
+    ///   5. `[signer]` Source token account transfer authority (owner or delegate)
```
mvines, line 20: This should be the last address, since it may be a multisig in which case additional signers will added.

mvines (commented):
super! just a couple nits

token-upgrade/program/src/instruction.rs
joncinque: I had a hard time with naming... "Fount" might imply that it's generating tokens, how about just "bag"?

token-upgrade/program/src/instruction.rs
joncinque: great idea!

token-upgrade/program/src/instruction.rs
joncinque: Ah, good call

CriesofCarrots:
> I've updated the proposal to work through a token account instead

Sorry if I'm missing something obvious, but is there a written proposal somewhere? I've read the instruction docs, but since the processor isn't implemented yet, I'm not sure I quite follow what you are envisioning.

token-upgrade/program/src/instruction.rs
joncinque: ...and it repeats `Upgrade` :sweat_smile: `Exchange` works for me

joncinque:
> Sorry if I'm missing something obvious, but is there a written proposal somewhere? I've read the instruction docs, but since the processor isn't implemented yet, I'm not sure I quite follow what you are envisioning.

That's my fault, I used the wrong term, and I meant interface. My hope was that the interface made things clear enough about how it would be implemented.

At this point though, since it's just one instruction, I'll add the implementation in this PR too. Unless it's better to commit the interface here and add the implementation separately.

CriesofCarrots:
> At this point though, since it's just one instruction, I'll add the implementation in this PR too. Unless it's better to commit the interface here and add the implementation separately.

Here works for me!

> My hope was that the interface made things clear enough about how it would be implemented.

The main thing that isn't clear to me is whether there's a separate "mint" step on the part of the destination mint_authority (or however they want to deliver tokens) to the escrow account? I assume there has to be (and it was sort of implied by the preminting conversation), but didn't see this made explicit anywhere.

joncinque:
> The main thing that isn't clear to me is whether there's a separate "mint" step on the part of the destination mint_authority (or however they want to deliver tokens) to the escrow account? I assume there has to be (and it was sort of implied by the preminting conversation), but didn't see this made explicit anywhere.

From what we're discussing here, the idea is to burn on one side, and transfer from a bag on the other side, so there's no minting at all in this protocol. This allows for a mint authority to mint a bunch of tokens to the bag owned by the PDA, or delegate tokens to the PDA, and then allow users to claim their upgraded tokens as they wish.

joncinque:
@CriesofCarrots I've got a first-pass on the implementation if you want to see generally how it'll work. I'll get some tests up next

CriesofCarrots:
> I've got a first-pass on the implementation if you want to see generally how it'll work. I'll get some tests up next

Cool, thanks! Implementation makes sense.

Some bike-shedding:
- I don't love "bag"; could we use "escrow" everywhere instead? Seems like you're leaning that way for the authority already
- I think "source" and "destination" are overloaded; as prefixes, they mean original mint+program and new mint+program, but you're also using them in `token_burn` and `token_transfer` as their original token-program api meanings. If we can think of better prefixes, great. Otherwise, unless you have more general use-cases for `token_burn` and `token_transfer`, I'd be inclined to use the program-specific variable names.

As an aside, it's interesting that someone could also use this program to offer 1:1 conversions of tokens within the same token program. Should probably document that.

joncinque:
Welp, I ended up going the whole way on this one.  #3565 will add a README with examples and CLI

joncinque:
There's no a CLI and docs at #3593, which are both ready for review.  Let me know if you need anything else!

CriesofCarrots (approved):
Lgtm! Take or leave the nits

token-upgrade/program/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+    new_mint: &'a Pubkey,
+) -> [&'a [u8]; 3] {
+    [
+        TOKEN_ESCROW_AUTHORITY_SEED,
+        original_mint.as_ref(),
+        new_mint.as_ref(),
+    ]
+}
+
+pub(crate) fn get_token_upgrade_authority_signer_seeds<'a>(
```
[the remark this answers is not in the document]
joncinque: makes sense to me!

token-upgrade/program/src/instruction.rs
```diff
@@ -0,0 +1,77 @@
...
+#[repr(u8)]
+pub enum TokenUpgradeInstruction {
+    /// Burns all of the original tokens in the user's account, and transfers the same
+    /// amount of tokens from an account owned by a PDA into another account.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writeable]` Original token account to burn from
+    ///   1. `[writeable]` Original token mint
+    ///   2. `[writeable]` Escrow of new tokens held by or delegated to PDA:
```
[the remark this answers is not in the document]
joncinque: Certainly

token-upgrade/program/src/processor.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        let token_amount = original_account.base.amount;
+        if new_escrow.base.amount < token_amount {
+            msg!(
+                "Escrow only has {} tokens, needs at least {}",
+                new_escrow.base.amount,
+                token_amount
+            );
+            return Err(ProgramError::InsufficientFunds);
+        }
```
2501babe, line 148: should we enforce `original_mint.base.decimals == new_mint.base.decimals` before here? since we dont provide a setup instruction, theres nothing to prevent the person who sets up the upgrade from making them different, and finding out the hard way when their conversion runs out of tokens partway through. enforcing it prevents any burn of the original from happening, giving the mint controller time to contemplate their words and deeds

i guess there might be a usecase for using this to change decimals, but i think exchanging base amounts 1:1 thereby scaling ui amounts feels... weird, and id expect (as a naive user) the behavior to be the ui amount is the same but the base amount scales. but idk if we want to deal with the math if theres no users for it

token-upgrade/program/tests/functional.rs
```diff
@@ -0,0 +1,305 @@
...
+        .unwrap();
+    token
+}
+
+#[test_case(spl_token::id(), spl_token_2022::id() ; "upgrade to token-2022")]
+#[test_case(spl_token_2022::id(), spl_token::id() ; "downgrade to token")]
+#[test_case(spl_token::id(), spl_token::id() ; "token to token")]
+#[test_case(spl_token_2022::id(), spl_token_2022::id() ; "token-2022 to token-2022")]
+#[tokio::test]
+async fn success(original_program_id: Pubkey, new_program_id: Pubkey) {
```
2501babe, line 92: this is really cool i had no idea this was a thing

2501babe:
this looks good to me! im a little iffy on not requiring the original mint authority to endorse the upgrade as legitimate somehow, but since randos could only grief people by burning tokens rather than actually gaining any value themselves from this, im inclined to say the simplicity of having one instruction and no state is probably worth it...

token-upgrade/program/src/processor.rs
joncinque: I was split on this too, and decided to punt by doing less.  This will probably be used more as a reference implementation for token creators to fork, make tiny changes, and deploy as needed.  But for whatever version we deploy, it probably makes more sense to put in a little more protection by ensuring that the mint has the same number of decimals.

Ok, you've convinced me!

joncinque:
>  im a little iffy on not requiring the original mint authority to endorse the upgrade as legitimate somehow

I hear you on that one. Even with only one "legitimate" upgrade mint though, it's still possible to engineer a social attack, likely through a similar program or something.  In the spirit of deeming this a reference implementation, the simplicity wins out, and people forking it can add whatever checks or special logic they want.

token-upgrade/program/tests/functional.rs
joncinque: hehe it's the cornerstone to testing the upgrade to token-swap

2501babe (approved):

Issue #945, opened 2020-12-13
Example: Add example of deserializing the SlotHash sysvar while remaining within the BPF compute limit


mvines:
Normal decoding of the SlotHash sysvar will exceed the current BPF compute limit. This PR demonstrates how to manually decode the sysvar to avoid this issue: https://github.com/mvines/solana-bpf-program-template/pull/4

This PR should be added to the examples/

PR #3440, opened 2022-08-08
Add example of deserializing the SlotHash sysvar while remaining with…

SUMEETRM:
Normal decoding of the SlotHash sysvar will exceed the current BPF compute limit. This PR demonstrates how to manually decode the sysvar to avoid this issue: https://github.com/mvines/solana-bpf-program-template/pull/4

It has been added to fix issue #945 in the examples folder and can be located by examples/rust/sysvar/src/slothash.rs


joncinque (commented):
Rather than adding a separate program in this file, how about adding the decoding to the existing `process_instruction` in `processor.rs` in this same directory? then you can merge this test with the functional test that already exists, just adding the slot hashes sysvar and file.

examples/rust/sysvar/src/slothash.rs
```diff
@@ -0,0 +1,104 @@
...
+        );
+
+        // Replace the SlotHashes sysvar will a fully populated version that was grabbed off Mainnet
+        // Beta by running:
+        //      solana account SysvarS1otHashes111111111111111111111111111 -o slot_hashes.bin
+        program_test.add_account_with_file_data(
+            sysvar::slot_hashes::id(),
+            sol_to_lamports(1.),
+            Pubkey::default(),
+            "slot_hashes.bin",
```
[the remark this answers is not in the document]
SUMEETRM: Tried to fix it in my most recent commit but not sure if that's what needs to be done. Couldn't find the tests/fixtures directory in rust examples so I added slot_hashes.bin to the same directory as processor.rs

I've added the code from slothashes.rs to processor.rs and have changed the function name. Is this how it's supposed to be done?

examples/rust/sysvar/src/processor.rs
```diff
@@ -43,3 +43,98 @@ pub fn process_instruction(
...
+
+
+//Code for deserializing the SlotHash sysvar while remaining within the BPF compute limit 
+
+//entrypoint!(process_instruction2);
+fn process_instruction2(
+    _program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    _instruction_data: &[u8],
+) -> ProgramResult {
```
joncinque, line 55: What I meant was to have all this code happen after line 43, in the normal `process_instruction` function.  Written this way, the processor will never be triggered

examples/rust/sysvar/src/processor.rs
```diff
@@ -43,3 +43,98 @@ pub fn process_instruction(
...
+            instruction::{AccountMeta, Instruction},
+            native_token::sol_to_lamports,
+            sysvar,
+        },
+        solana_program_test::*,
+        solana_sdk::{signature::Signer, transaction::Transaction},
+    };
+
+    #[tokio::test]
+    async fn test_transaction() {
```
joncinque, line 107: The idea is to merge this test with the one in  `examples/rust/sysvar/tests/functional.rs` since they exercise the same code underneath.  You can change the transaction to also include the slot hashes sysvar and rest should work very nicely

examples/rust/sysvar/src/processor.rs
```diff
@@ -43,3 +43,98 @@ pub fn process_instruction(
...
+        );
+
+        // Replace the SlotHashes sysvar will a fully populated version that was grabbed off Mainnet
+        // Beta by running:
+        //      solana account SysvarS1otHashes111111111111111111111111111 -o slot_hashes.bin
+        program_test.add_account_with_file_data(
+            sysvar::slot_hashes::id(),
+            sol_to_lamports(1.),
+            Pubkey::default(),
+            "slot_hashes.bin",
```
joncinque, line 123: Once this test is moved to `examples/rust/sysvar/tests/functional.rs`, then you can move the slot hashes file to `examples/rust/sysvar/tests/fixtures/slot_hashes.bin`

joncinque (commented):
I've clarified these points, let me know if you need anything else!

PR #3447, opened 2022-08-09
token-js: Add reallocate instruction implementation

### LINKED ISSUE #2956: token-js: Support `Reallocate` instruction
#### Problem

`token-2022` includes a new instruction to reallocate the space in an Account, but it isn't supported in the JS bindings.  The relevant instruction definition can be found at https://github.com/solana-labs/solana-program-library/blob/0268d76d306d86a9fb03a17d5ede18144bfad684/token/program-2022/src/instruction.rs#L529-L550

#### Solution

Following the model in #2951 , add support for this new instruction, including some tests.

joncinque:
#### Problem

`Reallocate` is an available instruction in token-2022 to add more space to token accounts, but the JS bindings don't use it yet.

#### Solution

Implement the reallocate instruction and add a test.

Fixes #2956

### DIFF
--- a/token/js/src/instructions/index.ts
+++ b/token/js/src/instructions/index.ts
@@ -23,7 +23,8 @@ export * from './initializeMultisig2'; // 19
 export * from './initializeMint2'; //     20
 export * from './initializeImmutableOwner'; //  22
 export * from './initializeMintCloseAuthority'; // 23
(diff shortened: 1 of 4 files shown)


staccDOTsol:
<3 cheers

PR #3448, opened 2022-08-09
token-js: Version bump to 0.3.0

joncinque:
#### Problem

We've got loads of new functionality in the spl-token JS package, but it's not published!

#### Solution

Bump the version, and improve the README with some more info for people upgrading from 0.1.x and 0.2.0


joncinque:
cc @jacobcreech for awareness that we're publishing this new package

Issue #3442, opened 2022-08-09
token-cli: --blockhash does nothing


2501babe:
this is a regression introduced when the cli was converted to async. `BlockhashQuery` is a client wrapper that can be instantiated with a static blockhash which it always returns, a client that fetches a blockhash, or a static blockhash that it verifies before returning. it was removed for the nonblocking client and an unconditional call to `get_latest_blockhash`

i found this by accident when dumping a transaction, and im opening an issue instead of fixing it because:
* do we even need this? im not sure if there is a usecase for producing offline non-nonce transactions considering theyre good for like 30 seconds
* a quick fix here is pointless because if we decide to keep supporting it im going to have to work it into the client pr im working on anyway

@joncinque thoughts?

joncinque:
Thanks for writing this up, I definitely broke this :sweat: 

We definitely need this, since even offline nonce transactions require a blockhash provided from the outside.  Check out the example and writeup here: https://spl.solana.com/token#example-offline-signing-with-multisig

Let's fix this the right way, by adding a nonblocking version of `BlockhashQuery` https://github.com/solana-labs/solana/blob/master/client/src/blockhash_query.rs -- since the implementation is pretty small, and a lot of the functions are deprecated, I think we're fine mostly copy-pasting with `async` and `await` wherever needed.

Alternatively, if you want to do it the DRY way, you can reuse the runtime from within the blocking RPC client in the blocking BlockhashQuery, and on the inside call the nonblocking `BlockhashQuery`, ie https://github.com/solana-labs/solana/blob/b6d38aad6979e7aca8fd431904af3cb90d87ebec/client/src/tpu_client.rs#L141-L146

But that seems like overkill.  What do you think?

2501babe:
cooool, im gonna look into copying the nonblocking client approach for a nonblocking `BlockhashQuery` then

this is another thing we are gonna have to pass in to `create_mint` et al so maybe we should have the complicated interface that the cli needs and then a wrapper with defaults (no nonce/offline stuff, at least) for normal consumers and expose them both...

joncinque:
Hm, might be worth implementing that no-op client to make this a bit simpler... I can get a draft PR up with it to show what I mean.

2501babe:
```
#[derive(Debug, PartialEq, Eq)]
pub enum BlockhashQuery {
    None(Hash),
    FeeCalculator(Source, Hash),
    All(Source),
}
```
[...]

PR #3449, opened 2022-08-09
token-cli: Add no-op client

joncinque:
#### Problem

In order to make offline signing work with `TokenClient`, we need to avoid going out to the network, the token client will by default try to send off transactions.

#### (Potential) Solution

Implement a no-op / offline client type that can be substituted in if `--sign-only` is specified.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -3138,7 +3138,7 @@ async fn handle_tx<'a>(
 ) -> Result<TransactionReturnData, Error> {
     let fee_payer = Some(&config.fee_payer);
 
-    let recent_blockhash = config.rpc_client.get_latest_blockhash().await?;
+    let recent_blockhash = config.program_client.get_latest_blockhash().await?;
     let message = if let Some(nonce_account) = config.nonce_account.as_ref() {
         let mut message = Message::new_with_nonce(
             instructions,
@@ -3151,9 +3151,9 @@ async fn handle_tx<'a>(
     } else {
         Message::new_with_blockhash(&instructions, fee_payer, &recent_blockhash)
     };
-    let fee = config.rpc_client.get_fee_for_message(&message).await?;
 
     if !config.sign_only {
+        let fee = config.rpc_client.get_fee_for_message(&message).await?;
         check_fee_payer_balance(config, minimum_balance_for_rent_exemption + fee).await?;
     }
 
--- a/token/client/src/client.rs
+++ b/token/client/src/client.rs
@@ -62,8 +62,14 @@ pub trait SendTransactionRpc: SendTransaction {
 #[derive(Debug, Clone, Copy, Default)]
 pub struct ProgramRpcClientSendTransaction;
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RpcClientResponse {
+    Signature(Signature),
+    Transaction(Transaction),
+}
+
 impl SendTransaction for ProgramRpcClientSendTransaction {
-    type Output = Signature;
+    type Output = RpcClientResponse;
 }
 
 impl SendTransactionRpc for ProgramRpcClientSendTransaction {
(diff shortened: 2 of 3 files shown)


2501babe:
oki i pushed my own commit that uses your noop client and fixes #3442. i also removed the maybe wrapper on blockhash in favor of a default. --blockhash is a hard requirement for --sign-only so this case should never occur anyway

one thing i kinda think would be cool is if `send_transaction` for `ProgramOfflineClient` returned a `Transaction`. doing this by making `ST` free seems back because then `Config<'a>` would need to be parametrized over it. maybe making what is now `ProgramRpcClientSendTransaction` a two-value enum? actually i dont know if thats possible... maybe `type Output = ` a two-value wrapper enum?

the thing im trying to do here is give the token client the ability to return a `Transaction` or a `Signature` depending on the mode of the `program_client`, which could be by having `send_transaction` return out of `create_mint` et al as is presently done. or it could be by giving the token client some way to `match` off a type from the program client to determine what to return...

hmm ill play with it tomorrow afternoon

joncinque:
> i also removed the maybe wrapper on blockhash in favor of a default. --blockhash is a hard requirement for --sign-only so this case should never occur anyway

Makes perfect sense, thanks for that

> maybe type Output = a two-value wrapper enum?

This seems like the best option to me.  We're in a weird position of some templating and some `dyn` bits with the `Arc<dyn ProgramClient<ProgramRpcClientSendTransaction>>`. The templated approach, of having the noop client always return a Transaction, and the rpc client always return a signature, makes things worse in the CLI, since it'll need to separate implementations for `ST: SendTransaction<Output = Signature>` and `ST: SendTransaction<Output = Transaction>`.  Whereas if we just match on `ST: SendTransaction<Output = TransactionOrSignature>`, we have a much easier time of everything.

2501babe:
incidentally i just realized i guess we dont need `BlockhashQuery` after all

2501babe:
i pushed the change to add the enum. id like to merge this and rebase my close-mint branch on it and then i think i can move all send logic into the client

however i see the dependency audit fails now. im not sure what we are supposed to do for this, since it looks like something pulled in indirectly via the monorepo?

2501babe:
amusingly i cant add you as a reviewer because its your pr... also if pushing to your branch is Rude lmk, i assumed not because you made it for me to use

joncinque:
#3458 ~fixed~ ignored the audit issue for now, so rebase and you'll be good

token/cli/src/main.rs
```diff
@@ -3138,7 +3138,7 @@ async fn handle_tx<'a>(
 ) -> Result<TransactionReturnData, Error> {
     let fee_payer = Some(&config.fee_payer);
 
-    let recent_blockhash = config.rpc_client.get_latest_blockhash().await?;
+    let recent_blockhash = config.program_client.get_latest_blockhash().await?;
```
joncinque, line 3141: Pretty awesome to see the fix boil down to this

joncinque (commented):
It looks good to me, but maybe @CriesofCarrots can also take a peek and make sure we're not too far off the mark.

token/client/src/client.rs
```diff
@@ -62,8 +62,14 @@ pub trait SendTransactionRpc: SendTransaction {
 #[derive(Debug, Clone, Copy, Default)]
 pub struct ProgramRpcClientSendTransaction;
```
2501babe, line 63: tbh the name `ProgramRpcClientSendTransaction` bothers me a lil bit now, since its used for rpc and offline, but i didnt think of anything better and the types here are already pretty complex

2501babe (approved):

token/client/src/client.rs
joncinque: I almost made the same comment, but I think it's technically ok since the base of both is meant to look like an RPC client, as opposed to a banks client.

CriesofCarrots (approved):
Lgtm! It's a little weird that the spl-token-cli constructs an RpcClient even for offline actions, but we do the same thing in solana-cli, so oh well.
Take WIP out of the title before merge?

PR #3450, opened 2022-08-10
token-swap: Support pools with mixed token and token-2022 mints

joncinque:
#### Problem

Token-2022 is out, but people may not know how to support both Token and Token-2022 within the same program.  For example, how do you swap Token tokens for Token-2022 tokens?

#### Solution

Provide another guide with step by step instructions, using the token-swap program as a testing ground.  This is part 2 of 3, so we're almost there!

### DIFF
--- a/docs/src/token-2022/onchain.md
+++ b/docs/src/token-2022/onchain.md
@@ -230,6 +230,250 @@ do more work.
 
 ## Part II: Support Mixed Token Programs: trading a Token for a Token-2022
 
+In Part I, we looked at the minimal amount of work to support Token-2022 in your
+program. This work won't cover all cases, however. Specifically, in the token-swap
+program, most instructions involve multiple token types. If those token types are
+from different token programs, then our current implementation will fail.
+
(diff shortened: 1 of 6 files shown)


docs/src/token-2022/onchain.md
```diff
@@ -230,6 +230,249 @@ do more work.
...
+    pub amount_in: u64,
+    pub minimum_amount_out: u64
+}
+```
+
+Note the clarification on `9.`, and the new inputs of `10.` and `11.`.
+
+All of these additional accounts may make you wonder: how big will transactions
+get with these new accounts? If you are using both Token and Token-2022,
+the additional Token-2022 program pubkey will take up space in the transaction.
```
jacobcreech, line 300: Probably good to mention that is +32bytes for those that are not aware yet

jacobcreech (changes requested):
Just one nit. Otherwise lgtm!

docs/src/token-2022/onchain.md
joncinque: Makes sense, thanks!

PR #3456, opened 2022-08-11
stake-pool: Bump versions to 0.7.0

joncinque:
#### Problem

We have new instructions in the stake pool program to create / update pool token metadata, but the program hasn't been released anywhere.

#### Solution

First step: tag a new release so we can build a program from it!


joncinque:
Merging on red since the audit issue won't be fixed until we update to 1.11

PR #3457, opened 2022-08-11
docs: Add token-2022 extension docs and JS examples

joncinque:
#### Problem

Token-2022 is out on all networks, and all extensions have JS bindings, but people don't know how to use the new extensions.

#### Solution

Add docs and examples for all the extensions, in JS to start.  Once CLI support is added everywhere, we'll add those too.


samkim-crypto:
Looks good to me! I think it is a nice read and the example are easy to follow.

docs/src/token-2022/extensions.md
```diff
@@ -2,4 +2,997 @@
...
+        TOKEN_2022_PROGRAM_ID
+    );
+```
+
+  </TabItem>
+</Tabs>
+
+#### Example: Withdraw withheld tokens from accounts
+
+As users transfer their tokens, transfer fees accumulate in the various recipient
```
jacobcreech, line 313: An additional example that would be useful: How do I get a list of accounts holding fees?

docs/src/token-2022/extensions.md
```diff
@@ -2,4 +2,997 @@
...
+    );
+```
+
+  </TabItem>
+</Tabs>
+
+### Non-Transferable Tokens
+
+To accompany immutably owner token accounts, the `NonTransferable` mint extension
+allows for "soul-bound" tokens that cannot be moved to any other entity. For
```
jacobcreech, line 636: I know "soul-bound" is the recent common term used in the Eth ecosystem, but do we want to use it here too? Maybe a #terminology thing

docs/src/token-2022/extensions.md
```diff
@@ -2,4 +2,997 @@
...
+With the Token-2022 extension model, however, we have the possibility to change
+how the UI amount of tokens are represented. Using the `InterestBearingMint`
+extension and the `amount_to_ui_amount` instruction, you can set an interest
+rate on your token and fetch its amount with interest at any time.
+
+Interest is continuously compounded based on the timestamp in the network. Due
+to drift that may occur in the network timestamp, the accumulated interest could
+be lower than the expected value. Thankfully, this is rare.
+
+Note that no new tokens are ever created, the UI amount returns the amount of tokens
```
jacobcreech, line 923:
```suggestion
**Note**: No new tokens are ever created, the UI amount returns the amount of tokens
```

jacobcreech: Also, does this mean that there needs to be a separate program that creates the tokens based on the interest accumulated? This statement seems like this feature is just cosmetic for UI vs actually creating new tokens as they are held.

jacobcreech (changes requested):
Looking great! Love the jsx examples. A few nits + a question on interest bearing tokens

docs/src/token-2022/extensions.md
```diff
@@ -2,4 +2,997 @@
 title: Extension Guide
 ---
 
-Coming soon!
+```mdx-code-block
```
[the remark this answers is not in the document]
joncinque: Huh, weird. It worked on the build, but you're probably right that it should be `.mdx` to be safe

docs/src/token-2022/extensions.md
joncinque: Up to you! I always think of "soul-bound" from my MMO days :laughing:

docs/src/token-2022/extensions.md
joncinque: There's no really easy way unfortunately... you'd have to iterate through the accounts and check them one by one.  Either way, I'll add it.

docs/src/token-2022/extensions.md
joncinque: The feature is in fact entirely cosmetic.  I'll make that clearer

docs/src/token-2022/extensions.md
jacobcreech: We can stick with it until we have a better recommendation imo

joncinque:
Ok, everything should be addressed now

jacobcreech (approved):

PR #3458, opened 2022-08-11
ci: Add rocksdb to ignore audit (for now)

joncinque:
#### Problem

There's a security advisory out for rocksdb, and audit is failing.

#### Solution

Ignore RUSTSEC-2022-0046 until we update to the newest Solana 1.11 crates.


joncinque:
Audit passed, gonna merge early

PR #3460, opened 2022-08-11
ci: Disable twoxtx for now

joncinque:
#### Problem

Since the new rocksdb lib is incompatible with Rust 1.59.0, we can't properly build the programs while `build-bpf` is on an older version of Rust.

#### Solution

Not the best option, but just disable the job until we have Rust 1.60 available everywhere.


joncinque:
The step was properly marked as skipped, merging early

PR #3465, opened 2022-08-12
ci: Re-enable twoxtx by using local `cargo-test-sbf`

### LINKED ISSUE #3461: ci: Re-enable token twoxtx build step
#### Problem

The twoxtx CI step was disabled in #3460 due to an incompatibility between rocksdb and Rust <1.60.0

#### Solution

Once all of the tools are on at least Rust 1.60.0, re-enable the job.  The main blocker at this point is `cargo build-sbf`.

joncinque:
#### Problem

We disabled the twoxtx CI job because of incompatible rust versions, but we can actually pick up the correct version if we just use `cargo-test-sbf` from the twoxtx solana repo.

#### Solution

Just use that!

Fixes #3461

### DIFF
--- a/.github/workflows/pull-request-token.yml
+++ b/.github/workflows/pull-request-token.yml
@@ -114,7 +114,7 @@ jobs:
       - name: Build and test token-2022 twoxtx (TEMPORARY)
         run: |
           ./token/twoxtx-setup.sh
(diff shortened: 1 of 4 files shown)


joncinque:
Closing in favor of #3553

PR #3468, opened 2022-08-13
Add associated token account IDL

acheroncrypto:
Follow up on #3373, added Anchor IDL for associated-token-account program.

Please let me know if this is appropriate. I have been waiting to submit the remaining IDLs until I receive a feedback.

### DIFF
--- a/associated-token-account/program/idl.json
+++ b/associated-token-account/program/idl.json
@@ -1,0 +1,126 @@
+{
+  "version": "1.1.1",
+  "name": "spl_associated_token_account",
+  "instructions": [
+    {
+      "name": "create",
+      "accounts": [
+        {
+          "name": "fundingAddress",
+          "isMut": true,
+          "isSigner": true
+        },
+        {
+          "name": "associatedAccountAddress",
+          "isMut": true,
+          "isSigner": false
+        },
+        {
+          "name": "walletAddress",
+          "isMut": false,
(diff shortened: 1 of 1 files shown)


joncinque:
Thanks for your contribution.  Are you sure that these IDLs work? I was trying to read into how anchor generates instructions and couldn't find it in the documentation, but it looks like it might hash the instruction name in order to generate a part of the instruction data, whereas the SPL instruction data is just one byte counting up from 0.

acheroncrypto:
Yes you are correct, Anchor programs have 8 byte discriminator up front, that's why we also use a coder. I've made a repository of SPL Anchor clients and tests here: https://github.com/acheroncrypto/native-to-anchor

Here are the tests for `associated-token-address` https://github.com/acheroncrypto/native-to-anchor/blob/master/client/tests/src/spl/associated-token-account.ts

I wasn't sure where to put all of these so made a repo for it, most of the SPL is tested. Curious to know what you think!

joncinque:
Would this IDL be usable without the custom coder, which is provided in a separate library?  If it requires an additional outside library, it would probably do better to live in that other library

acheroncrypto:
Yes I agree, I think there was a miscommunication for where to submit these IDLs but now that they have been merged into Anchor https://github.com/coral-xyz/anchor/pull/2143 this PR can be closed.

Issue #3403, opened 2022-08-01
token-cli: Support `MintCloseAuthority` extension


joncinque:
#### Problem

In token-2022, it's possible to set a close authority on mints using the `MintCloseAuthority` extension, but the CLI doesn't support this.

#### Solution

Using the `program_client` on the token cli `Config`, fully support the `MintCloseAuthority` extension, which entails the following features:

* add an optional `--close-authority` parameter during `create-mint`, and instead of creating the instructions by hand, use `create_mint` on the client, adding in the close authority if specified https://github.com/solana-labs/solana-program-library/blob/674b0ac70e1a319b0d97e78a50b84f79b803be2a/token/client/src/token.rs#L293
* support `close` with a mint passed in (if mint authority is specified)
* support `authorize` with `close-mint` to set a new close authority on the mint

2501babe:
im going to add a `close-mint` command instead of expanding the scope of `close`. `close` by default takes the mint address and closes the users associated account, so there is no ux here that wouldnt potentially be a catastrophic footgun

incidentally i dont like how some of these help messages use `TOKEN_ADDRESS` to mean different or ambiguous and i might do a separate pr to change that

Poletgole:
> #### Problem
> In token-2022, it's possible to set a close authority on mints using the `MintCloseAuthority` extension, but the CLI doesn't support this.
> 
> #### Solution
> Using the `program_client` on the token cli `Config`, fully support the `MintCloseAuthority` extension, which entails the following features:
> 
> * add an optional `--close-authority` parameter during `create-mint`, and instead of creating the instructions by hand, use `create_mint` on the client, adding in the close authority if specified https://github.com/solana-labs/solana-program-library/blob/674b0ac70e1a319b0d97e78a50b84f79b803be2a/token/client/src/token.rs#L293
> * support `close` with a mint passed in (if mint authority is specified)
> * support `authorize` with `close-mint` to set a new close authority on the mint

If i may suggest a solution it will be talking to a customer agent through live chat https://direct.lc.chat/14390013/ they got the best, straightforward and  transparent solution to most programming blockchain related issues

2501babe:
for point 1 i went with an `--enable-close` flag instead that doesnt take a parameter, to mirror the behavior of `--enable-freeze`

also thanks to the refactor work, implementing this was just adding the flag and pushing the struct into the vec. magic!

2501babe:
[...]

PR #3469, opened 2022-08-13
Refactor command_create_token to use token client

2501babe:
draft because:
* ~i had a small question about the memo handling~
* ~i need to rebase again when #3449 lands~
* ~i need to reverse merge out the close authority stubs~
* ~we need a new solana monorepo version to bring in the `Signers` impl (so this actually wont build at all for you as-is)~

but (other than that!!) this is basically done

this pr implements a full flow for `command_create_token` to go through the token client, including support for nonce mode offline signing. this is prerequisite work to #3403

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -46,6 +46,7 @@ use spl_token_2022::{
     instruction::*,
     state::{Account, Mint, Multisig},
 };
+use spl_token_client::{client::RpcClientResponse, token::Token};
 use std::{
     collections::HashMap, fmt::Display, process::exit, str::FromStr, string::ToString, sync::Arc,
 };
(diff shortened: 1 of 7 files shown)


token/client/src/token.rs
```diff
@@ -210,12 +221,29 @@ where
...
         Token {
             client: Arc::clone(&self.client),
             pubkey: self.pubkey,
             payer,
             program_id: self.program_id,
+            nonce_account: self.nonce_account,
+            nonce_authority: self.nonce_authority,
+            // XXX FIXME this is wrong but i dont understand the implications of clone on RwLock
+            // in fact i dont understand why we use write at all why not just Option<String> and clone
+            memo: Arc::new(RwLock::new(None)),
```
2501babe, line 234: actually i think this is the only issue i had left with this stuff. memo is thrown away here but needs to be retained, but im not sure what all the `write()` and `RwLock` stuff is for, is there a reason we dont want to just clone strings like normal?

token/cli/src/main.rs
```diff
@@ -3184,6 +3258,43 @@ async fn handle_tx<'a>(
...
+                &transaction,
+                &ReturnSignersConfig {
+                    dump_transaction_message: config.dump_transaction_message,
+                },
+            )))
+        }
+        RpcClientResponse::Signature(signature) if no_wait => {
+            Ok(TransactionReturnData::CliSignature(CliSignature {
+                signature: signature.to_string(),
+            }))
```
joncinque, line 3278: At first, I wrote that you should send here, before realizing that the send is already being done in the client -- nice!

token/client/src/token.rs
```diff
@@ -210,12 +221,29 @@ where
...
+        }
+    }
+
+    pub fn with_nonce(&self, nonce_account: &Pubkey, nonce_authority: &Pubkey) -> Token<T> {
```
joncinque, line 238: Careful here, there's an inconsistency in the interface between `with_payer` and `with_memo`, where `with_payer` creates a new `Token`, but `with_memo` modifies the existing one.  Because of that, this is incorrectly used in the CLI.

At first, I wrote that this should work like `with_memo` and modify the underlying client, but after reading this over, I think we're better off making this a factory-style constructor as you have it, so you can do `let token = with_payer().with_nonce()`.

token/client/src/token.rs
```diff
@@ -210,12 +221,29 @@ where
...
+        Token {
+            client: Arc::clone(&self.client),
+            pubkey: self.pubkey,
+            payer: self.payer.clone(),
+            program_id: self.program_id,
+            // XXX FIXME as above. actually why cant these just take &mut self? seems silly
```
joncinque, line 244: These can't take a normal `&mut self` because you're in an async context, where the underlying `self` can be accessed in different async tasks.  It's goofy, but if you need mutability in a struct that's shared async, you need to wrap some part in `Arc<RwLock<T>>` -- either the whole struct, or the fields.  For the memo, we wrapped the field, but in this case, since it's set in the constructor, you shouldn't need to do that

token/client/src/token.rs
```diff
@@ -259,85 +287,116 @@ where
         ))))
     }
 
-    pub async fn process_ixs<S2: Signers>(
+    pub async fn construct_tx<S2: Signers>(
```
joncinque, line 290: nit: for this and `process_ixs`, since we've removed the `S` type on the struct, we can rename `S2` to `S`

token/client/src/token.rs
joncinque: The idea is to do `with_memo` and send the transaction immediately after, so something like `token.with_memo("hiiii").create_mint(....)`, and that way it gets cleared out during the send, and it won't be there for your next transaction.

Since we're in an async context, we can't just use a normal `Option<String>`, as mentioned later

joncinque (commented):
This is looking great, definitely on the right track.  I left some comments -- let me know what you think!

token/client/src/token.rs
2501babe: yes youre right i use `with_nonce` (my own function!!) wrong h no

token/client/src/token.rs
2501babe: ah yea understand `&mut self` now. also understand why `Option<String>` isnt allowed, it works fine for a constructor but i didnt realize `take()` actually deleted the string off the object. this makes sense to me now, you can use the same mechanism for transfer memos

part of me wants to clone the memos in the other `with_` functions or name the functions something different or something but i guess i use this in one and only one place so it isnt really important

2501babe:
alright this branch is good to go once a new monorepo drops and we update to 1.11. ill ping you for a final once-over once thats done. the commit history is a bit fucked because i havent figured out how to cleanly rebase against master when i pull in commits to my branch from a branch and then rebase the original branch... alas. ill just squash all this when its time

token/client/src/token.rs
```diff
@@ -377,7 +432,7 @@ where
     /// Create and initialize a new token account.
     pub async fn create_auxiliary_token_account(
         &self,
-        account: &S,
+        account: &dyn Signer,
```
2501babe, line 435: id made this `Arc<dyn Signer>` but it broke every test, but this function is never used outside of tests so we can beautify this later. i feel like theres some cute shit that could be done with `AsRef`

2501babe:
if at first you dont succeed, force push and force push again. this is good to go now

token/client/src/token.rs
joncinque, line 435: Whoa, we should've done this the whole time

joncinque (approved):
Looks great! I'm glad the churn wasn't too bad on this one

token/client/src/token.rs
2501babe: its actually the same thing, trait objects are always dynamically sized. nothing outside of tests uses this but if we need to pass an `Arc` through ever i might see if i can use `AsRef`...

PR #3471, opened 2022-08-15
token-js: expose unpack methods

### LINKED ISSUE #3217: token: methods to decode account layouts for mint and token are not publicly accessible
Currently all methods to access a token account or mint are coupled to a connection object. In order to work with data from other transport streams, e.g. when fetching accounts via geyser plugins or other indexers i can't use them and so just copied all the layout decoding logic. A simple fix like exporting unpack* wouldn't help either as it expects AccountInfo rather then a plain Buffer.

### LINKED ISSUE #3472: spl-token js: Unpack methods are no exposed to clients
### Problem
`@solana/spl-token` does not export unpack functions

They are essential since clients might do all sort of account batching and need to be able to mix fetches of various account types.

A basic example is a wallet, it would do a `connection.getTokenAccountByOwner` then deserialize them all.

Then on update later on it might batch fetch token accounts and stake accounts, well everything is possible

### Solution

export `unpackAccount` and `unpackMint`

Fix
#3471

Arrowana:
Fixes #3217 fixes #3472

### DIFF
--- a/token/js/src/state/account.ts
+++ b/token/js/src/state/account.ts
@@ -157,7 +157,14 @@ export async function getMinimumBalanceForRentExemptAccountWithExtensions(
     return await connection.getMinimumBalanceForRentExemption(accountLen, commitment);
 }
 
-function unpackAccount(info: AccountInfo<Buffer> | null, address: PublicKey, programId: PublicKey) {
+/**
+ * Unpacks a token account
+ * @param info the token account on-chain account
+ * @param address TokenAccount
+ * @param programId SPL Token program account
+ * @returns
+ */
+export function unpackAccount(info: AccountInfo<Buffer> | null, address: PublicKey, programId: PublicKey): Account {
     if (!info) throw new TokenAccountNotFoundError();
     if (!info.owner.equals(programId)) throw new TokenInvalidAccountOwnerError();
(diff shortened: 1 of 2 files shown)


token/js/src/state/account.ts
```diff
@@ -157,7 +157,14 @@ export async function getMinimumBalanceForRentExemptAccountWithExtensions(
...
 
-function unpackAccount(info: AccountInfo<Buffer> | null, address: PublicKey, programId: PublicKey) {
+/**
+ * Unpacks a token account
+ * @param info the token account on-chain account
+ * @param address TokenAccount
+ * @param programId SPL Token program account
+ * @returns
+ */
+export function unpackAccount(info: AccountInfo<Buffer> | null, address: PublicKey, programId: PublicKey): Account {
```
jordansexton, line 167: These `programId` args can default to `TOKEN_PROGRAM_ID` since we're exposing them now.

jordansexton (commented):
Looks good!

token/js/src/state/account.ts
Arrowana: Yes, I realized this, added it but forgot to push...

joncinque (approved):
Thanks for your contribution!

PR #3484, opened 2022-08-16
Handle invalid fee account in `token-swap`

### LINKED ISSUE #3411: token-swap: Handle invalid fee account
#### Problem

A token swap pool has an associated fee account used for collecting trade / deposit / withdrawal fees, but if that account is ever closed from the outside using `spl_token::instruction::close_account`, the pool is bricked.

#### Solution

Just as in stake pools, gracefully handle the situation where a fee account might be closed or invalid with logic like this:

https://github.com/solana-labs/solana-program-library/blob/5d8cb5588b5abe893691f395211f2423e0560347/stake-pool/program/src/processor.rs#L1868-L1875

cc @jmalcic

jmalcic:
This PR resolves #3411 by handling cases where the pool fee account is closed when processing withdraw and swap instructions.


token-swap/program/src/processor.rs
```diff
@@ -7421,4 +7436,460 @@ mod tests {
...
+            .unwrap(),
+            vec![
+                &mut pool_account,
+                &mut SolanaAccount::default(),
+                &mut SolanaAccount::default(),
+            ],
+        )
+        .unwrap();
+
+        assert!(do_process_instruction(
```
joncinque, line 7550: nit: so we can surface the error info more easily on an error, let's do `do_process_instruction(....).unwrap()` rather than asserting that it's ok

token-swap/program/src/processor.rs
```diff
@@ -7421,4 +7436,460 @@ mod tests {
...
```
joncinque, line 7700: same here

token-swap/program/src/processor.rs
```diff
@@ -7421,4 +7436,460 @@ mod tests {
...
+            .unwrap(),
+            vec![
+                &mut accounts.pool_fee_account,
+                &mut destination,
+                &mut SolanaAccount::default(),
+            ],
+        )
+        .unwrap();
+
+        assert!(do_process_instruction_with_fee_constraints(
```
joncinque, line 7854: same here

token-swap/program/src/state.rs
```diff
@@ -164,6 +176,25 @@ impl SwapState for SwapV1 {
         &self.pool_fee_account
     }
 
+    fn check_pool_fee_info(&self, pool_fee_info: &AccountInfo) -> Result<(), ProgramError> {
+        let data = &pool_fee_info.data.borrow();
+        let token_account =
+            StateWithExtensions::<Account>::unpack(data).map_err(|err| match err {
```
joncinque, line 182: Thanks for using the token-2022 bits right away :bow:

joncinque (commented):
Just some micro-nits, then this is good to go

joncinque:
Thanks a bunch, looks great!

PR #3486, opened 2022-08-16
token/js - make unpack functions consistent

jordansexton:
The new unpack functions haven't been published to npm yet. This gets their argument order consistent with other lib functions, and updates to the latest spl-memo before releasing.

### DIFF
--- a/token/js/package.json
+++ b/token/js/package.json
@@ -46,7 +46,7 @@
         "@solana/web3.js": "^1.41.0"
     },
     "devDependencies": {
-        "@solana/spl-memo": "^0.1.0",
+        "@solana/spl-memo": "^0.2.0",
         "@types/chai-as-promised": "^7.1.4",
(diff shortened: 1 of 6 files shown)


token/js/package.json
```diff
@@ -46,7 +46,7 @@
         "@solana/web3.js": "^1.41.0"
     },
     "devDependencies": {
-        "@solana/spl-memo": "^0.1.0",
+        "@solana/spl-memo": "^0.2.0",
```
joncinque, line 49: no `yarn.lock` changes to go with this?

joncinque (dismissed):
Looks good! What other functions were these made to be consistent with?

token/js/package.json
jordansexton: Good point, just pushed lockfile.

jordansexton:
Published `@solana/spl-token@0.3.1`

PR #3487, opened 2022-08-16
token-cli: support `MintCloseAuthority`

### LINKED ISSUE #3403: token-cli: Support `MintCloseAuthority` extension

2501babe:
* ~add `--enable-close` flag to `create-token` that sets up a close authority~
* ~implement `close-mint` to close a mint that has a close authority~
* ~support `close-mint` in `authorize` to change a mint close authority~

~this needs to be rebased after #3469 lands~
~this is now based on the quite erroneously named `close-mint` branch so a rebase to master is trivial~
this is based on master

closes #3403

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -385,7 +385,6 @@ async fn command_create_token(
         .await?;
 
     let tx_return = finish_tx(config, &res, false).await?;
-
     Ok(match tx_return {
         TransactionReturnData::CliSignature(cli_signature) => format_output(
             CliMint {
@@ -681,6 +680,10 @@ async fn command_authorize(
         (COption::None, config.program_id)
     };
 
+    if program_id != config.program_id {
+        return Err(format!("Token program `{}` provided, but token belongs to `{}`", config.program_id, program_id).into());
+    }
+
     println_display(
         config,
         format!(
@@ -697,28 +700,9 @@ async fn command_authorize(
         ),
     );
 
-    // XXX TODO NEXT i need to change set_authority in the client to return properly
-    // and then i can cut out handle_tx here... cool yea good this actually works as is
+    let res = token.set_authority(&account, &authority, new_authority.as_ref(), authority_type, &bulk_signers).await?;
 
-    let instructions = vec![set_authority(
-        &program_id,
-        &account,
-        new_authority.as_ref(),
-        authority_type,
-        &authority,
-        &config.multisigner_pubkeys,
-    )?];
-    let tx_return = handle_tx(
-        &CliSignerInfo {
-            signers: bulk_signers,
-        },
-        config,
-        false,
-        0,
-        instructions,
-    )
-    .await?;
-
+    let tx_return = finish_tx(config, &res, false).await?;
     Ok(match tx_return {
         TransactionReturnData::CliSignature(signature) => {
             config.output_format.formatted_string(&signature)
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -4280,4 +4280,48 @@ mod tests {
             assert_eq!(ui_account.delegated_amount, None);
         }
     }
+
+    #[tokio::test]
+    #[serial]
+    async fn close_mint() {
+        let (test_validator, payer) = new_validator_for_test().await;
+        let config = test_config(&test_validator, &payer, &spl_token_2022::id());
+
+        let token_keypair = Keypair::new();
+        let token_pubkey = token_keypair.pubkey();
+        let bulk_signers: Vec<Arc<dyn Signer>> =
+            vec![Arc::new(clone_keypair(&payer)), Arc::new(token_keypair)];
+
+        command_create_token(
+            &config,
+            TEST_DECIMALS,
+            token_pubkey,
+            payer.pubkey(),
+            false,
+            true,
+            None,
+            bulk_signers,
+        )
+        .await
+        .unwrap();
+
+        let account = config.rpc_client.get_account(&token_pubkey).await.unwrap();
+        let test_mint = StateWithExtensionsOwned::<Mint>::unpack(account.data);
+        assert!(test_mint.is_ok());
+
+        process_test_command(
+            &config,
+            &payer,
+            &[
+                "spl-token",
+                CommandName::CloseMint.into(),
+                &token_pubkey.to_string(),
+            ],
+        )
+        .await
+        .unwrap();
+
+        let account = config.rpc_client.get_account(&token_pubkey).await;
+        assert!(account.is_err());
+    }
 }
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1474,24 +1474,64 @@ async fn command_close(
     })
 }
 
-// HANA new command impl
-// XXX OK COOL WHAT NOW
-// use command_close as a guide
-// the first thing is we want to do sanity checks in online mode
-// that means fetch the mint, check supply is zero
-// check there exists a close authority? or, check the authority matches the one we expect
-// code to do the parse is in set_authority
-// the first thing to check is how close actually works
-// if it goes through the same instruction maybe reuse that
-// otherwise it might have its own shit idk
-//
-// alright cool it does in fact go through close account, this is easy
-// first step make command_close go through the client
-//
-// done. i needed to refactor some shit but close_account now handles native recipient and multisig
-// so close_mint is easy. i just need different sanity checks and then the close_account call
-async fn command_close_mint(config: &Config<'_>) -> CommandResult {
-    Ok("".to_string())
+async fn command_close_mint(
+    config: &Config<'_>,
+    token_pubkey: Pubkey,
+    close_authority: Pubkey,
+    recipient: Pubkey,
+    bulk_signers: BulkSigners,
+) -> CommandResult {
+    let token = token_client_from_config(config, &token_pubkey);
+
+    if !config.sign_only {
+        let mint_account = config.rpc_client.get_account(&token_pubkey).await?;
+        config.check_owner(&token_pubkey, &mint_account.owner)?;
+
+        let mint_state = StateWithExtensionsOwned::<Mint>::unpack(mint_account.data)
+            .map_err(|_| format!("Could not deserialize token mint {}", token_pubkey))?;
+        let mint_supply = mint_state.base.supply;
+
+        if mint_supply > 0 {
+            return Err(format!(
+                "Mint {} still has {} outstanding tokens; these must be burned before closing the mint.",
+                token_pubkey, mint_supply,
+            )
+            .into());
+        }
+
+        if let Ok(mint_close_authority) = mint_state.get_extension::<MintCloseAuthority>() {
+            let mint_close_authority_pubkey =
+                Option::<Pubkey>::from(mint_close_authority.close_authority);
+
+            if mint_close_authority_pubkey != Some(close_authority) {
+                return Err(format!(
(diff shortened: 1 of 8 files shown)


token/client/src/token.rs
```diff
@@ -536,26 +537,28 @@ where
...
             &[instruction::set_authority(
                 &self.program_id,
                 account,
                 new_authority,
                 authority_type,
-                &owner.pubkey(),
+                authority,
+                // XXX i think this is wrong and we need to supply all bulk signers for multisig to work
+                // leaving for now, theres a working example in close_account
                 &[],
```
2501babe, line 557: .

2501babe, line 557: im 90% sure this would cause multisig shit to fail because of `validate_owners` in `program-2022/src/processor.rs`, theres a (in theory???) working but ugly example below to compare

token/client/src/token.rs
```diff
@@ -765,23 +768,39 @@ where
...
+            .map(Borrow::borrow)
+            .collect::<Vec<&Pubkey>>();
+
+        let mut instructions = vec![instruction::close_account(
+            &self.program_id,
+            account,
+            destination,
+            authority,
+            signer_pubkey_refs,
+        )?];
```
2501babe, line 791: this works tho i think

2501babe:
@joncinque do you have opinions about the above notes? no rush im not blocked or anything

token/client/src/token.rs
```diff
@@ -765,23 +768,39 @@ where
...
-            )?],
-            &[authority],
-        )
-        .await
+        // XXX i know this is stupid. but maybe it would feel less stupid buried in a helper function!
+        let signer_pubkeys = signing_keypairs.pubkeys();
+        let signer_pubkey_refs: &[&Pubkey] = &signer_pubkeys
+            .iter()
+            .map(Borrow::borrow)
+            .collect::<Vec<&Pubkey>>();
```
joncinque, line 783: nit: You can gather all the refs in a vec without the `map(Borrow::borrow)` bit and just do `signer_pubkeys.iter().collect::<Vec<_>>()`

I know it looks goofy, but it's the only way

token/client/src/token.rs
joncinque: This API is a bit tricky -- the idea is that the last parameter `signer_pubkeys` provides the signers *of the multisig*. So the last parameter should only be used in a multisig situation.

Ideally, the `authority` parameter would either be a Signer or a Pubkey, but that's tricky to do.

As an alternative, we could check the signers, and if there's just one whose pubkey matches the authority, then we pass empty, otherwise we pass them all in.  What do you think?

token/client/src/token.rs
joncinque: You got it right -- this definitely doesn't work with multisigs unfortunately.

token/client/src/token.rs
2501babe: oooh thats cool

token/client/src/token.rs
```diff
@@ -765,23 +772,39 @@ where
...
-            )?],
-            &[authority],
-        )
-        .await
+        let signing_pubkeys = signing_keypairs.pubkeys();
+        let multisig = if signing_pubkeys == &[*authority] {
+            signing_pubkeys.iter().collect::<Vec<_>>()
+        } else {
+            vec![]
+        };
```
2501babe, line 787: oki this is somewhat better, and in line with your suggestion to pass through the empty array when not a multisig

2501babe: haha jk this is egregiously wrong:
* i got the branch arms backwards
* `process_command` pushes the config signer onto `bulk_signers` without checking if its there yet, so in many cases the vec is the payer twice
* having multiple signers in the array is necessary *but not sufficient* evidence that this is a multisig transaction. the simplest case is where payer and authority differ

following the flow from `main`, `bulk_signers` gets multisig signers first, then the fee payer, then possibly a nonce authority, then (in the case of `Command::Close`, eg) close authority, which by default is the payer a second time

there are all kind of weird edge cases possible. eg, if we dedupe payers and it reduces to two addresses, one of which is the authority, we cant know whether the non-authority is payer or whether authority is payer and non-authority is a 1/1 multisig. we can check payer on the `Token` object, but then it depends on whether the user initialized it properly

im inclined to pass all signers to the instruction and let god (token) sort them out

2501babe: i thought a bit more about this and decided the fixed version of this is actually fine. its not the token functions responsibility to make sure the keypairs array is carefully assembled, it just needs to sign with whats given, and expecting to see just the authority key, or some multisig/nonce stuff, is reasonable. its the *callers* responsibility to pass a good array

2501babe: one thing id like is if i could make it a helper function but the lifetime shit is too weird for me

joncinque: Yeah I think that's reasonable. The only footgun with this is a weird situation in which a multisig is a signer for itself. As in, I create a multisig at `multsig111111111111....` and inside, one of its signers is `multisig1111111...`. But maybe we just decide to make life hard for someone like that?

token/cli/src/main.rs
```diff
@@ -324,51 +328,66 @@ pub fn signers_of(
     }
 }
 
+fn token_client_from_config(
+    config: &Config<'_>,
+    token_pubkey: &Pubkey,
+) -> Token<ProgramRpcClientSendTransaction> {
+    let token = Token::new(
+        config.program_client.clone(),
+        &config.program_id,
```
joncinque, line 337: Careful, this might not always be correct -- it defaults to `Tokenkeg...` but almost all of the commands figure out the correct program id to use based on the mint. The only exception is `create_token`, where we can't figure it out

token/cli/src/main.rs
```diff
@@ -550,10 +571,11 @@ async fn command_authorize(
         AuthorityType::WithheldWithdraw => "withdraw withheld authority",
         AuthorityType::InterestRate => "interest rate authority",
     };
-    let (previous_authority, program_id) = if !config.sign_only {
+
+    let previous_authority = if !config.sign_only {
         let target_account = config.rpc_client.get_account(&account).await?;
         config.check_owner(&account, &target_account.owner)?;
-        let program_id = target_account.owner;
```
joncinque, line 556: Be sure to keep this and update the token client based on the owner, since the default is just `Tokenkeg...`

token/cli/src/main.rs
```diff
@@ -631,32 +666,29 @@ async fn command_authorize(
             auth_str,
             previous_authority
                 .map(|pubkey| pubkey.to_string())
-                .unwrap_or_else(|| "disabled".to_string()),
+                .unwrap_or_else(|| if config.sign_only {
+                    "unknown".to_string()
```
joncinque, line 670: nit: I don't think it'll be possible to get to this line since it's always `None` in sign only mode

token/cli/src/main.rs
```diff
@@ -1423,35 +1457,73 @@ async fn command_close(
...
+    config: &Config<'_>,
+    token_pubkey: Pubkey,
+    close_authority: Pubkey,
+    recipient: Pubkey,
+    bulk_signers: BulkSigners,
+) -> CommandResult {
+    let token = token_client_from_config(config, &token_pubkey);
+
+    if !config.sign_only {
+        let mint_account = config.rpc_client.get_account(&token_pubkey).await?;
```
joncinque, line 1487: You'll need to update the program id in the config based on the owner of this account

token/cli/src/main.rs
```diff
@@ -1408,10 +1440,12 @@ async fn command_close(
...
     bulk_signers: BulkSigners,
 ) -> CommandResult {
-    let (is_recipient_wrapped, program_id) = if config.sign_only {
-        (false, config.program_id)
-    } else {
+    let token = token_client_from_config(config, &Pubkey::default());
+
+    if !config.sign_only {
         let source_account = config.rpc_client.get_account(&account).await?;
+        config.check_owner(&account, &source_account.owner)?;
```
joncinque, line 1447: You'll need to update the program id in the config based on this owner

joncinque (commented):
Looks good for the most part! There's just the bit about figuring out the correct program id rather than relying on the user-provided one.

token/cli/src/main.rs
2501babe: ty i didnt even think of this, id gotten used to "providing the correct program id on invocation is mandatory"

token/cli/src/main.rs
2501babe: unless im misunderstanding which line, it should _always_ get this line in sign only. its to clear up a tiny display inaccuracy: in normal mode, `None` means there is no authority, but in sign only, we dont know

token/client/src/token.rs
2501babe: im not sure what the issue would be even in this case. the flow is "if `signing_keypairs` is one key that matches the authority, the authority param will mark it as a signer of the instruction. if its anything else, everything is marked as a signer. then all signers sign the transaction unconditionally, in addition to whatever payer may be designated on the `Token` object"

2501babe:
ok i cleared up the issues flagged

token/client/src/token.rs
2501babe: incidentally im moving the if/else into a function in my next pr instead of copying this logic around everywhere

token/cli/src/main.rs
joncinque: No worries, it's really not clear in the code... we may want to put a note on `config.program_id` to discourage its use as much as possible

token/cli/src/main.rs
joncinque: Ah sorry, yes you're absolutely right, I reviewed this a little too quickly, and didn't see `unwrap_or_else`, and thought it was happening on `map`.  All good!

joncinque (approved):
Thanks for driving this forward, looks great!

token/cli/src/main.rs
```diff
@@ -1423,35 +1458,82 @@ async fn command_close(
             )
             .into());
         }
-        config.check_owner(&account, &source_account.owner)?;
 
-        let recipient_account = config.rpc_client.get_token_account(&recipient).await?;
-        let is_recipient_wrapped = recipient_account.map(|x| x.is_native).unwrap_or(false);
-        (is_recipient_wrapped, source_account.owner)
```
CriesofCarrots, line 1430: Wow, @2501babe , just discovered that `close` has been totally broken for non-wrapped recipients (like a vanilla system account) since [this](https://github.com/solana-labs/solana-program-library/pull/3278), until your change here. Thanks for fixing!
spl-token v2.0.17 is affected, so it'll be cool to get your new cli release out :)

2501babe: oh, thank you! i tried to do a cli release recently but me and jon scuttled it due to token22 dependency version stuff between spl and monorepo, since there wasnt a pressing need to resolve at the time. ill prioritize this with jon and sam since theres regressions to fix now

PR #3489, opened 2022-08-16
Add on-chain compression program that uses CMT

ngundotra:
Add on-chain program called `spl-account-compression` (formerly known as `gummyroll`) that provides a developer-friendly interface to compression account data using the `ConcurrentMerkleTree` data structure.

cc @jarry-xiao @austbot @samwise2


account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,549 @@
+//! Gummyroll is an on-chain Merkle tree that supports concurrent writes.
+//!
+//! A buffer of proof-like changelogs is stored on-chain that allow multiple proof-based writes to succeed within the same slot.
+//! This is accomplished by fast-forwarding out-of-date (or possibly invalid) proofs based on information stored in the changelogs.
+//! See a copy of the whitepaper [here](https://google.com)
```
jarry-xiao, line 5: Let's fix these comments for when the crate gets published

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,549 @@
...
+//! one exemplified use-case is the [Bubblegum](https://google.com) contract,
+//! which uses Gummyroll trees to store encoded information about NFTs.
+//! The use of Gummyroll within Bubblegum allows for:
+//! - up to 1 billion NFTs to be stored in a single account on-chain (10,000x decrease in on-chain cost)
+//! - (by default) up to 1024 concurrent updates per slot (this number is not correct)
+//!
+//! Operationally, Gummyroll trees **must** be supplemented by off-chain indexers to cache information
+//! about leafs and to power an API that can supply up-to-date proofs to allow updates to the tree.
+//! All modifications to Gummyroll trees are settled on the Solana ledger via instructions against the Gummyroll contract.
+//! A production-ready indexer (Plerkle) can be found in the [Metaplex program library](https://google.com)
```
jarry-xiao, line 17: Here as well

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,549 @@
...
+    Ok(())
+}
+
+fn fill_in_proof_from_canopy(
+    canopy_bytes: &mut [u8],
+    max_depth: u32,
+    index: u32,
+    proof: &mut Vec<Node>,
+) -> Result<()> {
+    // 26 is hard coded as it is the current max depth that Gummyroll supports
```
jarry-xiao, line 167: I forgot to change this comment :)

account-compression/Anchor.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[registry]
+url = "https://anchor.projectserum.com"
+
+[provider]
+cluster = "testnet"
+wallet = "~/.config/solana/id.json"
+
+[scripts]
+test = "yarn run ts-mocha -t 1000000 tests/**/*-test.ts"
```
jarry-xiao, line 19: Do we want to add the tests to this repo?

account-compression/README.md
```diff
@@ -0,0 +1,13 @@
+# Smart Contracts
+
+`gumdrop` will be managed by solana-labs
```
jarry-xiao, line 3: We probably need to significantly modify this README. I (and @jnwng ) will help you with dividing up some of these tasks.

account-compression/Anchor.toml
ngundotra: Yes but they depend on setting up the SDK, which I want to add separately, because I want it to use Solita

account-compression/programs/spl-compression/src/lib.rs
ngundotra: Updated

account-compression/programs/spl-compression/src/lib.rs
ngundotra: Updated

ngundotra: Actually this whole paragraph needs to be rewritten

account-compression/README.md
```diff
@@ -1,13 +1,18 @@
-# Smart Contracts
+# Account Compression
 
-`gumdrop` will be managed by solana-labs 
+This on-chain program provides an interface for composing smart-contracts to create and use
+SPL ConcurrentMerkleTrees. The primary application of using SPL ConcurrentMerkleTrees is
+to make edits to off-chain data with on-chain verification. 
 
-`bubblegum`, `gumball-machine` and `gumdrop` are managed by metaplex-foundation
+Using this program requires an indexer to parse transaction information and write relevant information to an off-chain database.
```
jnwng, line 7: can you expand on what "relevant information" means here?

jnwng (approved):

account-compression/README.md
ngundotra: This is a note for application devs. 

Relevant information refers to application-level information that gets hashed into the leaf.

This information has to be logged via cpi to `wrapper` in order for off-chain indexers to update properly.

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+        previous_leaf: [u8; 32],
+        new_leaf: [u8; 32],
+        index: u32,
+    ) -> Result<()> {
+        let mut merkle_roll_bytes = ctx.accounts.merkle_roll.try_borrow_mut_data()?;
+        let (header_bytes, rest) =
+            merkle_roll_bytes.split_at_mut(size_of::<ConcurrentMerkleTreeHeader>());
+
+        let header = Box::new(ConcurrentMerkleTreeHeader::try_from_slice(header_bytes)?);
+        assert_eq!(header.authority, ctx.accounts.authority.key());
```
joncinque, line 416: It looks like this can fail under normal operation if someone signs with a different authority -- best to return an error rather than panicking

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+        let merkle_roll_size = merkle_roll_get_size!(header)?;
+        let (roll_bytes, canopy_bytes) = rest.split_at_mut(merkle_roll_size);
+
+        let mut proof = vec![];
+        for node in ctx.remaining_accounts.iter() {
+            proof.push(node.key().to_bytes());
+        }
```
joncinque, line 423: I'm surprised to see the nodes provided as accounts in the instruction rather than as data.  Does that have to do with the program that calls into the compression program?

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+        let mut header = Box::new(ConcurrentMerkleTreeHeader::try_from_slice(&header_bytes)?);
+        header.initialize(
+            max_depth,
+            max_buffer_size,
+            &ctx.accounts.authority.key(),
+            Clock::get()?.slot,
+        );
+        header.serialize(&mut header_bytes)?;
+        let merkle_roll_size = merkle_roll_get_size!(header)?;
+        let (roll_bytes, canopy_bytes) = rest.split_at_mut(merkle_roll_size);
```
joncinque, line 336: Unless I've missed something obvious, could you provide definitions of the "roll" and "canopy" somewhere? And are the terms "roll" and "header" interchangeable?

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub fn replace_leaf(
+        ctx: Context<Modify>,
+        root: [u8; 32],
+        previous_leaf: [u8; 32],
+        new_leaf: [u8; 32],
+        index: u32,
+    ) -> Result<()> {
+        let mut merkle_roll_bytes = ctx.accounts.merkle_roll.try_borrow_mut_data()?;
+        let (header_bytes, rest) =
+            merkle_roll_bytes.split_at_mut(size_of::<ConcurrentMerkleTreeHeader>());
```
joncinque, line 413: The documentation say that this is checked in the instruction, but only the authority and signature is checked.  An owner check may be useful as well in case it's possible to no-op and trick someone into thinking that it worked

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// Authority that validates the content of the trees.
+    /// Typically a program, e.g., the Bubblegum contract validates that leaves are valid NFTs.
+    pub authority: Signer<'info>,
+
+    /// Program used to emit changelogs as instruction data.
+    /// See `WRAPYChf58WFCnyjXKJHtrPgzKXgHp6MD9aVDqJBbGh`
+    pub candy_wrapper: Program<'info, Wrapper>,
+}
+
+/// Context for inserting, appending, or replacing a leaf in the tree
```
joncinque, line 64: Could you also document the use of remaining accounts?

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub fn transfer_authority(
+        ctx: Context<TransferAuthority>,
+        new_authority: Pubkey,
+    ) -> Result<()> {
+        let mut merkle_roll_bytes = ctx.accounts.merkle_roll.try_borrow_mut_data()?;
+        let (mut header_bytes, _) =
+            merkle_roll_bytes.split_at_mut(size_of::<ConcurrentMerkleTreeHeader>());
+
+        let mut header = Box::new(ConcurrentMerkleTreeHeader::try_from_slice(header_bytes)?);
+        assert_eq!(header.authority, ctx.accounts.authority.key());
```
joncinque, line 454: Same here -- an error is better than panicking

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+        root: [u8; 32],
+        leaf: [u8; 32],
+        index: u32,
+    ) -> Result<()> {
+        let mut merkle_roll_bytes = ctx.accounts.merkle_roll.try_borrow_mut_data()?;
+        let (header_bytes, rest) =
+            merkle_roll_bytes.split_at_mut(size_of::<ConcurrentMerkleTreeHeader>());
+        let header = Box::new(ConcurrentMerkleTreeHeader::try_from_slice(header_bytes)?);
+        let merkle_roll_size = merkle_roll_get_size!(header)?;
+        let (roll_bytes, canopy_bytes) = rest.split_at_mut(merkle_roll_size);
```
joncinque, line 476: Up to you, but none of this is checked, so someone could spoof a merkle tree.  Probably not a big deal, but better to check to be sure

joncinque (commented):
Great job! Mainly some small questions / cleanup items.  Were there any tests to go along with any of this?

account-compression/README.md
```diff
@@ -0,0 +1,18 @@
...
+This on-chain program provides an interface for composing smart-contracts to create and use
+SPL ConcurrentMerkleTrees. The primary application of using SPL ConcurrentMerkleTrees is
+to make edits to off-chain data with on-chain verification. 
+
+Using this program requires an indexer to parse transaction information and write relevant information to an off-chain database.
+
+
+## SDK 
+
+The typescript SDK for this contract will be generated using Metaplex Foundation's [Solita](https://github.com/metaplex-foundation/solita/).
```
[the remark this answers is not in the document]
ngundotra: Yes the SDK will be presented in a separate PR

account-compression/README.md
```diff
@@ -0,0 +1,18 @@
...
+Using this program requires an indexer to parse transaction information and write relevant information to an off-chain database.
+
+
+## SDK 
+
+The typescript SDK for this contract will be generated using Metaplex Foundation's [Solita](https://github.com/metaplex-foundation/solita/). 
+
+## Testing
+
+Testing contracts locally requires the SDK to be built. Then you can run: `anchor test`
```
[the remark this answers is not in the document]
ngundotra: It is possible to test this code as is but it's a little bit cumbersome. I don't mind adding banks tests against this code after we add the typescript SDK & tests.

account-compression/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@types/node": "^18.0.5",
+    "chai": "^4.3.4",
+    "mocha": "^9.0.3",
+    "ts-mocha": "^10.0.0",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.7.4"
+  },
+  "scripts": {
+    "update-deps": "cd deps/metaplex-program-library; git fetch; git checkout 7e2810a; cd ../solana-program-library; git fetch; cd ../..",
+    "build": "yarn update-deps && cargo build-bpf --manifest-path deps/metaplex-program-library/token-metadata/program/Cargo.toml && cargo build-bpf --manifest-path deps/solana-program-library/token/program/Cargo.toml && cargo build-bpf --manifest-path deps/solana-program-library/token/program-2022/Cargo.toml && cargo build-bpf --manifest-path deps/solana-program-library/associated-token-account/program/Cargo.toml",
```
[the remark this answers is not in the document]
ngundotra: Forgot to strip this out, this is no longer necessary

account-compression/programs/spl-compression/src/data_wrapper/mod.rs
```diff
@@ -0,0 +1,30 @@
+//! # Data Wrapper
+//! We use CPI calls to circumvent the 10kb log limit on Solana transactions.
+//! Instead of logging events to the runtime, we execute a CPI to the `wrapper` program
+//! where the log data is serialized into the instruction data.
+//!
+//! This works because CPI instruction data is never truncated. Logging information is
+//! vital to the functioning of compression. When compression logs are truncated, indexers can fallback to 
+//! deserializing the CPI instruction data.
```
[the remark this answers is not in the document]
ngundotra: Maximum number of bytes needed is `32 * (MAX_DEPTH + 1) == 32 * 32 == 1024` per operation.

It might be possible to slim this down (cc @jarry-xiao )

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+    pub candy_wrapper: Program<'info, Wrapper>,
+}
+
+/// Context for validating a provided proof against the SPL ConcurrentMerkleTree.
+/// Throws an error if provided proof is invalid.
+#[derive(Accounts)]
+pub struct VerifyLeaf<'info> {
+    /// CHECK: This account is validated in the instruction
+    pub merkle_roll: UncheckedAccount<'info>,
+}
```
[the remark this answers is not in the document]
ngundotra: This instruction is allows composing programs to check if a leaf exists in the tree

account-compression/programs/spl-compression/src/lib.rs
ngundotra: Will provide definition of `canopy`.

`Roll` is a catch-all term that refers to account data that holds the SPL-CMT (ie header + merkle tree). Open to replacing the `roll` with plain old `merkle_tree`.

account-compression/programs/spl-compression/src/data_wrapper/mod.rs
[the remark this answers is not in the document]
joncinque: Ah ok, how about this: transaction return data is set to be exactly 1024 bytes, and should be logged by validators.  The only issue is that you can run into an issue if another program overwrites it.  Would you be able to use that?

account-compression/programs/spl-compression/src/lib.rs
joncinque: Let's go with `merkle_tree` everywhere

account-compression/programs/spl-compression/src/lib.rs
jarry-xiao: Yeah might be good to add that the owner of the merkle_tree is the correct program. Practically doesn’t matter for any of the instructions that actually modify the data

account-compression/programs/spl-compression/src/data_wrapper/mod.rs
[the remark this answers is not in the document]
jarry-xiao: Practically inconvenient to wait for the runtime to implement this new solution. I agree that this is a hack, but logs getting dropped silently should be 100% unacceptable IMO. I would be more okay with building a separate undroppable log buffer or somehow injecting a Noop program in every VM so the CPI overhead is as low as possible. I’m pretty sure Ethereum does not drop events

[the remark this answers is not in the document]
jarry-xiao: Also I think we need a little more data that that @ngundotra (Node index, tree_id). It’s always less than the 1280 byte CPI instruction size limit so there’s no reason to worry unless that changes.

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+#[program]
+pub mod spl_compression {
+    use super::*;
+
+    /// Creates a new merkle tree with maximum leaf capacity of `power(2, max_depth)`
+    /// and a minimum concurrency limit of `max_buffer_size`.
+    ///
+    /// Concurrency limit represents the # of replace instructions that can be successfully
+    /// executed with proofs dated for the same root. For example, a maximum buffer size of 1024
+    /// means that a minimum of 1024 replaces can be executed before a new proof must be
```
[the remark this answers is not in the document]
ngundotra: Actually... this is a minimum.

SPL CMTs by default will fast-forward proofs through the entire changelog buffer if it can't find the original root that proof was submitted for. This can create scenarios where proofs can be more then BUFFER_SIZE (1024) operations old, yet still be fast forwarded to a valid proof...

This feels a little bit unsafe, but the operating thesis was that so long as you know what the leaf data is, you should be able to modify it.

account-compression/programs/spl-compression/src/state/concurrent_merkle_tree_header.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub fn initialize(
+        &mut self,
+        max_depth: u32,
+        max_buffer_size: u32,
+        authority: &Pubkey,
+        creation_slot: u64,
+    ) {
+        // Check header is empty
+        assert_eq!(self.max_buffer_size, 0);
+        assert_eq!(self.max_depth, 0);
```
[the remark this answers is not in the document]
ngundotra: Actually, we need to add an AccountDiscriminator / AccountType here and check system ownership

account-compression/Cargo.toml
```diff
@@ -0,0 +1,5 @@
+[workspace]
+members = [
+    "programs/spl-compression",
```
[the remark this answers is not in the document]
ngundotra: Updated

account-compression/programs/spl-compression/src/lib.rs
ngundotra: Added

account-compression/programs/spl-compression/src/lib.rs
ngundotra: Added

account-compression/programs/spl-compression/src/lib.rs
ngundotra: Added

account-compression/programs/spl-compression/src/lib.rs
ngundotra: Added

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+        emit!(*change_log);
+        update_canopy(canopy_bytes, header.max_depth, None)
+    }
+
+    /// Note:
+    /// Supporting this instruction open a security vulnerability for indexers.
+    /// This instruction has been deemed unusable for publicly indexed compressed NFTs.
+    /// Indexing batched data in this way requires indexers to read in the `uri`s onto physical storage
+    /// and then into their database. This opens up a DOS attack vector, whereby this instruction is
+    /// repeatedly invoked, causing indexers to fail.
```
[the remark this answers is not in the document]
ngundotra: Eventually this can be secure. Using this requires some complex on-chain verification however

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+                msg!("Error zero copying merkle roll: {}", err);
+                err!(AccountCompressionError::ZeroCopyError)
+            }
+        }
+    }
+}
+
+/// This applies a given function on a merkle roll by
+/// allowing the compiler to infer the size of the tree based
+/// upon the header information stored on-chain
```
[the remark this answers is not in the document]
ngundotra: Good catch, removed

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+macro_rules! merkle_roll_get_size {
```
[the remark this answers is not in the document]
ngundotra: Rewrote as a templated function

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+        ctx: Context<Initialize>,
+        max_depth: u32,
+        max_buffer_size: u32,
+    ) -> Result<()> {
+        let mut merkle_roll_bytes = ctx.accounts.merkle_roll.try_borrow_mut_data()?;
+
+        let (mut header_bytes, rest) =
+            merkle_roll_bytes.split_at_mut(size_of::<ConcurrentMerkleTreeHeader>());
+
+        let mut header = Box::new(ConcurrentMerkleTreeHeader::try_from_slice(&header_bytes)?);
```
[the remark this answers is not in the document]
ngundotra: No benefit, removed

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// and a minimum concurrency limit of `max_buffer_size`.
+    ///
+    /// Concurrency limit represents the # of replace instructions that can be successfully
+    /// executed with proofs dated for the same root. For example, a maximum buffer size of 1024
+    /// means that a minimum of 1024 replaces can be executed before a new proof must be
+    /// generated for the next replace instruction.
+    ///
+    /// Concurrency limit should be determined by empirically testing the demand for
+    /// state built on top of SPL Compression.
+    pub fn init_empty_gummyroll(
```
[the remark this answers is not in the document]
ngundotra: Addressed

account-compression/programs/spl-compression/src/lib.rs
ngundotra: I'm double checking this

account-compression/programs/spl-compression/src/lib.rs
ngundotra: Added

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// CHECK: This account will be zeroed out, and the size will be validated
+    pub merkle_roll: UncheckedAccount<'info>,
+
+    /// Authority that validates the content of the trees.
+    /// Typically a program, e.g., the Bubblegum contract validates that leaves are valid NFTs.
+    pub authority: Signer<'info>,
+
+    /// Program used to emit changelogs as instruction data.
+    /// See `WRAPYChf58WFCnyjXKJHtrPgzKXgHp6MD9aVDqJBbGh`
+    pub candy_wrapper: Program<'info, Wrapper>,
```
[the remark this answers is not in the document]
ngundotra: Fixed! All candy references have been scrubbed

libraries/concurrent-merkle-tree/src/hash.rs
```diff
@@ -1,6 +1,9 @@
 use crate::node::{empty_node, Node};
 use solana_program::keccak::hashv;
 
+#[cfg(feature = "sol-log")]
+use solana_program::msg;
```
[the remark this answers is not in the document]
ngundotra: Thanks!  Updated

account-compression/programs/spl-compression/src/lib.rs
ngundotra: The benefit of passing proof nodes as accounts is that it can deduplicate proof data needed for transactions that perform multiple Compression instructions

account-compression/programs/spl-compression/src/lib.rs
ngundotra: Added documentation for `canopy` under `canopy.rs`

account-compression/programs/spl-compression/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+    /// repeatedly invoked, causing indexers to fail.
+    pub fn init_gummyroll_with_root(
+        ctx: Context<Initialize>,
+        max_depth: u32,
+        max_buffer_size: u32,
+        root: [u8; 32],
+        leaf: [u8; 32],
+        index: u32,
+        _changelog_db_uri: String,
+        _metadata_db_uri: String,
```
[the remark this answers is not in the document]
ngundotra: Removed

account-compression/programs/spl-compression/src/state/concurrent_merkle_tree_header.rs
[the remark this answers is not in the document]
ngundotra: Added account type key

account-compression/programs/spl-compression/src/data_wrapper/mod.rs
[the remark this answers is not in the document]
joncinque: Return data is currently available in the runtime on all networks if you just use `program::set_return_data` https://github.com/solana-labs/solana/blob/9703855a80d6f5991a093c98c626cbbdb7e628e1/sdk/program/src/program.rs#L322 and validators can log if they choose with the `--enable-extended-tx-metadata-storage` flag.

Either way, the more I think about it, that would just be another hack.  How many bytes can you log currently with normal `msg`?

account-compression/programs/account-compression/src/canopy.rs
```diff
@@ -0,0 +1,127 @@
+//! Canopy is way to cache the upper `N` levels of a SPL ConcurrentMerkleTree.
```
joncinque, line 1: This is great, thanks for providing the info

account-compression/programs/wrapper/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "wrapper"
```
joncinque, line 2: is this meant to be published? If so, maybe spl-data-wrapper or something?

account-compression/programs/wrapper/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+crate-type = ["cdylib", "lib"]
+name = "wrapper"
```
joncinque, line 9: same as with the other, may as well get rid of this

account-compression/programs/spl-compression/src/lib.rs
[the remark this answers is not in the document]
joncinque: Does that mean you literally *can't* generate another proof before 1024 replaces?

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+            _ => {
+                msg!("Failed to apply {} on concurrent merkle tree with max depth {} and max buffer size {}", stringify!($func), $header.max_depth, $header.max_buffer_size);
+                err!(AccountCompressionError::ConcurrentMerkleTreeConstantsError)
+            }
+        }
+    };
+}
+
+#[program]
+pub mod spl_compression {
```
joncinque, line 205: Does this name need to change to `spl_account_compression` or is it ok like this?

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+
+    /// Program used to emit changelogs as instruction data.
+    /// See `WRAPYChf58WFCnyjXKJHtrPgzKXgHp6MD9aVDqJBbGh`
+    pub log_wrapper: Program<'info, Wrapper>,
+}
+
+/// Context for inserting, appending, or replacing a leaf in the tree
+///
+/// Modification instructions also require the proof to the leaf to be provided
+/// as base58-encoded nodes as "remaining accounts".
```
joncinque, line 73: nittiest of nits: if it's remaining accounts, it's not base58-encoded, just 32 bytes

account-compression/programs/spl-compression/src/lib.rs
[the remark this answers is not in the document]
ngundotra: I think "at least" is the better term. 

It's 1024 replaces before "cache entry" is flushed. However based on access pattern of nearby leaves, we may be able to recreate the cache entry as needed.

account-compression/programs/spl-compression/src/data_wrapper/mod.rs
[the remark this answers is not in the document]
jarry-xiao: Yeah I've known about the return data feature for a while, but ultimately we went with the option that seemed to show up most consistently in RPC calls. The problem with msg is that I can prepend a transaction to maliciously overflow the log buffer and mess up the indexer, so even logging a single byte would be unsafe

account-compression/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@types/mocha": "^9.0.0",
+    "@types/node": "^18.0.5",
+    "chai": "^4.3.4",
+    "mocha": "^9.0.3",
+    "ts-mocha": "^10.0.0",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.7.4"
+  },
+    "test": "anchor test"
+  }
```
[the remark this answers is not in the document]
ngundotra: Fixed

account-compression/programs/account-compression/src/canopy.rs
ngundotra: 👍

account-compression/programs/wrapper/Cargo.toml
ngundotra: Removed name

account-compression/programs/account-compression/src/lib.rs
ngundotra: it's okay like this, renamed package to `spl_account_compression`

account-compression/programs/wrapper/Cargo.toml
ngundotra: This is not meant to be published. Would prefer to limit the "official" visibility of this package / program because, like you mentioned, this is currently a hack.

ngundotra:
> Return data is currently available in the runtime on all networks if you just use program::set_return_data https://github.com/solana-labs/solana/blob/9703855a80d6f5991a093c98c626cbbdb7e628e1/sdk/program/src/program.rs#L322 and validators can log if they choose with the --enable-extended-tx-metadata-storage flag.

Return data is limited per programId in a single transaction. Even if we could fit our `ChangelogEvent` in that, it would make this program a lot less useful. For example, this would prevent multiple compressed CandyMachine `mint` instructions from being able to be executed in the same instruction. 

Unless validators have the ability to voluntarily truncate CPI args in their tx storage, I think this `wrapper` solution is easier to reason about and use than having to identify which validators stored the extended logs for this program.

joncinque:
> I think this wrapper solution is easier to reason about and use than having to identify which validators stored the extended logs for this program.

For sure, in the end you've come up with a great solution. It's just too bad that the runtime has to spin up a vm to run a no-op.  We may need to come up with a more solid "log" syscall that will always work, regardless of log limits, specifically to be used for events.

joncinque (approved):
Great job! I'll let you do the honors again

PR #3491, opened 2022-08-17
token-swap: Migrate to `transfer_checked`

joncinque:
#### Problem

If people follow the on-chain program migration docs, they may get stuck if their program uses `transfer` instead of `transfer_checked` everywhere.

#### Solution

Provide docs and example for how to migrate to `transfer_checked`, using token-swap example once again.

There will be one more doc after this, for using specific extensions, and after that we'll be good from the on-chain side, I hope.


samkim-crypto (approved):
This part is all clear to me!

PR #3494, opened 2022-08-18
Gov: add shank for spl governance program

ngundotra:
Automatically generate an Anchor IDL by annotating the Account structs and Instruction enum with shank macros.

### DIFF
--- a/governance/program/src/entrypoint.rs
+++ b/governance/program/src/entrypoint.rs
@@ -3,9 +3,12 @@
 
 use crate::{error::GovernanceError, processor};
 use solana_program::{
-    account_info::AccountInfo, entrypoint, entrypoint::ProgramResult,
+    account_info::AccountInfo, declare_id, entrypoint, entrypoint::ProgramResult,
     program_error::PrintProgramError, pubkey::Pubkey,
 };
+
+// Todo: Feature-gate this to only run when publishing shank IDL
+declare_id!("GovER5Lthms3bLBqWub97yVrMmEogzX7xNjdXpPPCVZw");
 
 entrypoint!(process_instruction);
 fn process_instruction(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5957,6 +5957,7 @@ dependencies = [
  "proptest",
  "serde",
  "serde_derive",
+ "shank",
  "solana-program",
  "solana-program-test",
  "solana-sdk",
--- a/governance/program/Cargo.toml
+++ b/governance/program/Cargo.toml
@@ -19,6 +19,7 @@ num-derive = "0.3"
 num-traits = "0.2"
 serde = "1.0.130"
 serde_derive = "1.0.103"
+shank = { version = "~0.0.5" }
 solana-program = "1.10.35"
 spl-token = { version = "3.5", path = "../../token/program", features = [ "no-entrypoint" ] }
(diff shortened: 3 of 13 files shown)


ngundotra:
Blocked by: https://github.com/metaplex-foundation/shank/pull/32

governance/program/src/entrypoint.rs
```diff
@@ -3,10 +3,13 @@
 
 use crate::{error::GovernanceError, processor};
 use solana_program::{
-    account_info::AccountInfo, entrypoint, entrypoint::ProgramResult,
+    account_info::AccountInfo, declare_id, entrypoint, entrypoint::ProgramResult,
     program_error::PrintProgramError, pubkey::Pubkey,
 };
 
+// Todo: Feature-gate this to only run when publishing shank IDL
+declare_id!("GovER5Lthms3bLBqWub97yVrMmEogzX7xNjdXpPPCVZw");
```
thlorenz, line 11: We may also extend shank to allow you to supply the program ID manually via the command line.
I just assumed people always wanna expose this id via `declare_id`, but it seems that assumption is not correct.

ngundotra:
Blocked on https://github.com/metaplex-foundation/solita/issues/71. Due to enum types like `VoteThreshold`, the SPL Governance IDL cannot be used by Anchor to deserialize its own accounts. Specifically this throws when initializing the `BorshAccountsCoder` object.

One other really nice thing that's missing is a metadata mapping for accounts, that allows Anchor / Solita to deserialize accounts using a prefixed enum to the accounts (rather than an Anchor Account Discriminator). Adding something like a `{metadata: mapping: {}}` to map between the prefix and the account type is the final step in using this IDL to deserialize governance accounts in the Explorer.

This would make the Anchor IDL spec more useable for SPL IDLs.

thlorenz:
> Adding something like a {metadata: mapping: {}} to map between the prefix and the account type is the final step in using this IDL to deserialize governance accounts in the Explorer.

Would love to understand more. Please file an issue in the solita repo with an example of what you need.
Alternatively you might be able to work around your issue using the [idl hook](https://github.com/metaplex-foundation/solita#hooking-into-idl-creation) (at least until we fix this properly).

governance/program/src/entrypoint.rs
SebastianBor: `programId` can't be hardcoded here because there are multiple instances of `spl-governance` deployed and hence the generated IDL/SDK should take `programId` as input param and not hardcore it.

SebastianBor:
> 
> To generate the Anchor idl: `cargo install shank-cli && shank idl` from inside `governance/programs`

Would the generated sdk work with any instance of `spl-governance`? In other words can `programId` be provided for api calls?

ngundotra:
@SebastianBor yes shank can work for any program id deployment. I've removed the `declare_id!` statement.

Now the anchor idl must be generated with `shank idl -p GovER5Lthms3bLBqWub97yVrMmEogzX7xNjdXpPPCVZw`

thlorenz:
FYI even with the `declare_id!` statement solita will only generate a default for the program id in the generated code that is overridable when calling instructions.
But overriding during the _shanking_ process via CLI is a viable option as well to get the default to the desired value.

ngundotra:
Thoughts on merging this @SebastianBor ?

PR #3495, opened 2022-08-19
fix incorrect docstrings in stake pool client instruction libraries

0xEdgar:
the python and ts client stake pool instructions have some incorrect docstrings


joncinque (approved):
Thanks for your contribution!

PR #3498, opened 2022-08-20
Governance: Use mint authority to revoke membership tokens

SebastianBor:
#### Summary

Use `governing_token_mint_authority` instead of `realm_authority` to authorise `Membership` tokens revocation.

The `mint_authority` can mint the `Membership` tokens and hence has ultimate power over the governance tokens because it could dilute any existing governance structure by minting more tokens. It makes then more sense to use the same authority to authorise revocation (burn) of those tokens instead of the `realm_authority` which doesn't have to be the same.


joncinque (approved):
Looks great!

PR #3499, opened 2022-08-20
Governance: Do not tip on max vote weight when tipping Disabled

SebastianBor:
#### Summary

When vote tipping is disabled the vote should not end early even if `max_voter_weight` worth of `vote_weight` have been cast.


joncinque (approved):
Makes sense, looks good!

PR #3500, opened 2022-08-20
Governance: Add council_vote_tipping

SebastianBor:
#### Summary 

Make it possible to have different `vote_tipping` settings for `Community` and `Council`.

It allows to create a governance structure where both `Community` and `Council` can control the same assets but with different voting rules. For example a governance to control program upgrades could use the following settings:

 - `Community` `vote_tipping` set to `Disabled`  meaning its vote can't end early and always must last the configured amount to time.
 - `Council` `vote_tipping` set to `Early` meaning it has power to create and execute fast forward emergency proposals to upgrade security vulnerabilities. 

#### Implementation details
- `council_vote_tipping` was added to `GovernanceConfig` struct. It used reserved space making the change backward compatible with previous versions and the binary account layout.

### DIFF
--- a/governance/program/src/state/governance.rs
+++ b/governance/program/src/state/governance.rs
@@ -248,6 +248,23 @@ impl GovernanceV2 {
 
         Ok(vote_threshold.clone())
     }
+
+    /// Returns VoteTipping for the given governing_token_mint
+    pub fn get_vote_tipping(
+        &self,
+        realm_data: &RealmV2,
+        governing_token_mint: &Pubkey,
+    ) -> Result<&VoteTipping, ProgramError> {
+        let vote_tipping = if *governing_token_mint == realm_data.community_mint {
+            &self.config.community_vote_tipping
+        } else if Some(*governing_token_mint) == realm_data.config.council_mint {
+            &self.config.council_vote_tipping
+        } else {
+            return Err(GovernanceError::InvalidGoverningTokenMint.into());
+        };
+
+        Ok(vote_tipping)
+    }
 }
 
 /// Deserializes Governance account and checks owner program
--- a/governance/program/src/state/governance.rs
+++ b/governance/program/src/state/governance.rs
@@ -36,8 +36,8 @@ pub struct GovernanceConfig {
     /// Time limit in seconds for proposal to be open for voting
     pub max_voting_time: u32,
 
-    /// Conditions under which a vote will complete early
-    pub vote_tipping: VoteTipping,
+    /// Conditions under which a Community vote will complete early
+    pub community_vote_tipping: VoteTipping,
 
     /// The type of the vote threshold used for council vote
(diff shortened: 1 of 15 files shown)


joncinque (commented):
How does this impact existing governances with `Early` or `Disable` vote tipping? Here they'll get set to a "default" of `Strict` after the change -- would that change behavior for some council token votes?

Other than that, looks good!

SebastianBor:
> How does this impact existing governances with `Early` or `Disable` vote tipping? Here they'll get set to a "default" of `Strict` after the change -- would that change behavior for some council token votes?
> 

For the existing accounts it's set to the same as it was before as a single setting 

```
        // For legacy accounts default Council VoteTipping to the Community
        governance_data.config.council_vote_tipping =
            governance_data.config.community_vote_tipping.clone()
```

governance/program/src/state/governance.rs
```diff
@@ -294,6 +314,10 @@ pub fn get_governance_data(
         // The assumption here is that council should have Veto vote enabled by default and equal to council_vote_threshold
         governance_data.config.council_veto_vote_threshold =
             governance_data.config.council_vote_threshold.clone();
+
+        // For legacy accounts default Council VoteTipping to the Community
+        governance_data.config.council_vote_tipping =
+            governance_data.config.community_vote_tipping.clone()
```
joncinque, line 320: I may be missing something in the upgrade path, but this is only triggered for legacy accounts where `governance_data.config.council_vote_threshold == VoteThreshold::YesVotePercentage(0)`.

Are there accounts where `council_vote_threshold` is set to something (ie, *not* `YesVotePercentage(0)`) that would use the default tipping rather than the community one?

SebastianBor: That condition triggers the upgrade path because it indicates we deserialised a legacy `V1` or `V2` account which didn't have settings for:
- `community_vote_threshold`
- `council_vote_threshold`
- `community_veto_vote_threshold`
- `council_veto_vote_threshold`
-  `community_vote_tipping`
-  `council_vote_tipping`

All these settings were introduced in `V3` and the code initialises them as follows:

- `community_vote_threshold`  replaces the legacy single `vote_threshold` field - the legacy field was renamed
- `council_vote_threshold` is set to legacy `vote_threshold` 
- `council_veto_vote_threshold ` is set to legacy `vote_threshold`
- `community_veto_vote_threshold` is set to `Disabled`
-  `community_vote_tipping` replaces the legacy singe `vote_tipping` field - the legacy field was renamed
- `council_vote_tipping` is set to the legacy `vote_tipping`

Note, it's not possible to set `council_vote_threshold` to `VoteThreshold::YesVotePercentage(0)` in `V3` and at the same time it's always that in the legacy versions and hence it can be used as the upgrade trigger.

joncinque: Ah ok, great, thanks for the explanation!

PR #3501, opened 2022-08-20
Governance: Explicit disabled weight thresholds

SebastianBor:
#### Summary 

Make it possible to explicitly disable voter weight thresholds to create governances and proposals. 
The purpose of this change is to have strong guaranteed to prevent governance or proposal creation by given governance population. 

#### Implementation 

Ideally `Option<u64>` should be used to indicate the optionality of the value but In order to preserve backward compatibility with existing realms the value of `u64::MAX` was used to denote Disabled thresholds.

### DIFF
--- a/governance/program/src/error.rs
+++ b/governance/program/src/error.rs
@@ -434,6 +434,10 @@ pub enum GovernanceError {
     /// Cannot change community TokenType to Memebership
     #[error("Cannot change community TokenType to Memebership")]
     CannotChangeCommunityTokenTypeToMemebership, // 604
+
+    /// Voter weight threshold disabled
+    #[error("Voter weight threshold disabled")]
+    VoterWeightThresholdDisabled, // 605
 }
 
 impl PrintProgramError for GovernanceError {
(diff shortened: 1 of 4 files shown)


joncinque (approved):
Looks good -- is the main reason for this change to give a better error message? If the threshold is `u64::MAX`, then no one has enough tokens to create a proposal anyway, correct?

SebastianBor:
> Looks good -- is the main reason for this change to give a better error message? If the threshold is `u64::MAX`, then no one has enough tokens to create a proposal anyway, correct?

The reason is to make a strong guarantee it can't be used in any circumstances and not just for the error message. 
Normally it should never happen, but if there was a bug in a `voter_weight` plugin which would allow a bad actor to get `u64:MAX` `voter_weight` then it would be able to circumvent this logic and potentially create and execute a malicious proposal.

```
 if voter_weight < min_weight_to_create_proposal {
            return Err(GovernanceError::NotEnoughTokensToCreateProposal.into());
 }
```

joncinque:
Ah smart, I didn't think about the weight plugins

Issue #3502, opened 2022-08-20
Error to import `@solana/spl-token` ESM module


grhbit:
Importing `@solana/spl-token` occurs errors like below such as next.js or browser.
```
error - ../node_modules/.pnpm/@solana+spl-token@0.3.1/node_modules/@solana/spl-token/lib/esm/extensions/extensionType.mjs:5:0
Module not found: Can't resolve './memoTransfer.mjs'

Import trace for requested module:
../node_modules/.pnpm/@solana+spl-token@0.3.1/node_modules/@solana/spl-token/lib/esm/extensions/index.mjs
../node_modules/.pnpm/@solana+spl-token@0.3.1/node_modules/@solana/spl-token/lib/esm/index.mjs
./src/pages/index.tsx
```

It seems to need that some ts files in `src` would be updated like below.

```diff
diff --git a/src/extensions/extensionType.ts b/src/extensions/extensionType.ts
index cbc0f4760f6c325d092d5d4b0441046b93b1d0f9..e4b4a3aa267b26cfe83b02292bd74219434343a5 100644
--- a/src/extensions/extensionType.ts
+++ b/src/extensions/extensionType.ts
@@ -2,11 +2,11 @@ import { ACCOUNT_SIZE } from '../state/account';
 import { Mint, MINT_SIZE } from '../state/mint';
 import { MULTISIG_SIZE } from '../state/multisig';
 import { ACCOUNT_TYPE_SIZE } from './accountType';
-import { MEMO_TRANSFER_SIZE } from './memoTransfer';
-import { DEFAULT_ACCOUNT_STATE_SIZE } from './defaultAccountState';
+import { MEMO_TRANSFER_SIZE } from './memoTransfer/index';
+import { DEFAULT_ACCOUNT_STATE_SIZE } from './defaultAccountState/index';
 import { MINT_CLOSE_AUTHORITY_SIZE } from './mintCloseAuthority';
 import { IMMUTABLE_OWNER_SIZE } from './immutableOwner';
-import { TRANSFER_FEE_CONFIG_SIZE, TRANSFER_FEE_AMOUNT_SIZE } from './transferFee';
+import { TRANSFER_FEE_CONFIG_SIZE, TRANSFER_FEE_AMOUNT_SIZE } from './transferFee/index';
 import { NON_TRANSFERABLE_SIZE } from './nonTransferable';
[...]

PR #3507, opened 2022-08-22
Fix Memo program ESM exports

jordansexton:
This is the first of two PRs to address https://github.com/solana-labs/solana-program-library/issues/3502, this one for the Memo program JS (which the Token program JS depends on). This fixes the issues that build systems have with .mjs files by exporting .js files. It also restructures the project to be similar to Wallet Adapter / the TypeScript monorepo template I've been using. I'll update Token in parallel.


memo/js/.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 4
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
```
joncinque, line 9: Can you remove this?

memo/js/package.json
```diff
@@ -1,70 +1,77 @@
...
-    "typescript": "^4.7.3",
-    "typescript-esm": "^2.0.0"
-  },
-  "dependencies": {
-    "@solana/web3.js": "^1.41.0",
-    "buffer": "^6.0.3"
-  }
+    "name": "@solana/spl-memo",
+    "description": "SPL Memo Program JS API",
+    "version": "0.3.1",
```
joncinque, line 4: Not a big deal, but 0.3.0 should probably be next -- was this version copy-pasted from Token?

memo/js/.editorconfig
jordansexton: I can, but the value of it is that it makes an IDE use the same spacing settings as Prettier is configured to here. Thoughts?

memo/js/package.json
jordansexton: Oh, you're right. This should be 0.2.1, this is a nonbreaking change.

jordansexton:
> Why is there no lockfile at all now?

Fixed!

memo/js/.editorconfig
joncinque: It's just spacing, so I'd prefer not.  Many IDEs infer from prettier files what to do as well. Finally, CI checks for it anyway, so let's axe it

jordansexton: Got it. Removed!

jordansexton:
Published in @solana/spl-memo@0.2.1

PR #3508, opened 2022-08-22
Fix Token program ESM exports

### LINKED ISSUE #3502: Error to import `@solana/spl-token` ESM module

jordansexton:
This is the second of two PRs to address https://github.com/solana-labs/solana-program-library/issues/3502, this one for the Token program JS (https://github.com/solana-labs/solana-program-library/pull/3507 for Memo, which it builds on top of). This fixes the issues that build systems have with .mjs files by exporting .js files. It also restructures the project to be similar to Wallet Adapter / the TypeScript monorepo template I've been using.

Fixes #3502


token/js/src/actions/approve.ts
```diff
@@ -1,15 +1,8 @@
...
-    TransactionSignature,
-} from '@solana/web3.js';
-import { TOKEN_PROGRAM_ID } from '../constants';
-import { createApproveInstruction } from '../instructions/index';
-import { getSigners } from './internal';
+import type { ConfirmOptions, Connection, PublicKey, Signer, TransactionSignature } from '@solana/web3.js';
+import { sendAndConfirmTransaction, Transaction } from '@solana/web3.js';
+import { TOKEN_PROGRAM_ID } from '../constants.js';
+import { createApproveInstruction } from '../instructions/approve.js';
+import { getSigners } from './internal.js';
```
jordansexton, line 5: What's changing across the project is that we are using imports with explicit .js extensions, and we're limiting use of index imports to descendants of any given directory. Each instruction can be imported directly without circular references.

token/js/package.json
```diff
@@ -1,74 +1,86 @@
...
+        "deploy:docs": "npm run docs && gh-pages --dist token/js --dotfiles"
+    },
+    "peerDependencies": {
+        "@solana/web3.js": "^1.20.0"
     },
     "dependencies": {
         "@solana/buffer-layout": "^4.0.0",
         "@solana/buffer-layout-utils": "^0.2.0",
-        "@solana/web3.js": "^1.41.0"
+        "@solana/spl-memo": "^0.2.0",
```
joncinque, line 57: Shouldn't this be a dev dependency? Also, may as well specify 0.3.0 for the next version, right?

token/js/src/actions/approve.ts
joncinque: what's the reason for explicit .js extensions? Is that specifically for those build systems?  And what do you mean by

> we're limiting use of index imports to descendants of any given directory

I didn't notice any differences in the import statements in index files

joncinque (commented):
Just the one question, but otherwise looks good. Also, the node runner for the GitHub action needs to be updated to version 16

token/js/package.json
jordansexton: This is an unbundled runtime dependency, so not a dev dependency. But I did change the version to 0.2.1 to match the next one.

joncinque: How is it a runtime dependency? It's only ever used in a test

token/js/src/actions/approve.ts
jordansexton: > what's the reason for explicit .js extensions? Is that specifically for those build systems?

TypeScript (somewhat infamously) [will not add `.js` extensions to imports](https://github.com/microsoft/TypeScript/issues/16577) when compiling, but _will_ resolve `.js` files to `.ts` files.

Next.js, Webpack 5, and some other build systems won't automatically resolve `.mjs` files when importing without an extension, which is what is causing https://github.com/solana-labs/solana-program-library/issues/3502.

Explicit declaring the file extensions here is the way to get them explicitly declared in the output. I will probably need to add an eslint rule to enforce this.

> I didn't notice any differences in the import statements in index files

For example, this file change we are commenting on replaces:

```ts
import { createApproveInstruction } from '../instructions/index';
```
with 
```ts
import { createApproveInstruction } from '../instructions/approve.js';
```

Instead of using an index import, which would need to become `'../instructions/index.js'` anyway, we just import from the specific file we need. This practice is generally good because it helps avoid circular references. Plus, we don't need to resolve the whole index to get one or two exports from it.

joncinque: Ah gotcha, I missed that part, thanks for all the explanation!

token/js/package.json
jordansexton: Oh, I was mistaken. I thought it was used by the new `MemoTransferInstruction` but it isn't. Fixed!

token/js/src/actions/approve.ts
jordansexton: Np!

jordansexton:
> Also, the node runner for the GitHub action needs to be updated to version 16

Done, thanks! Running now.

joncinque (approved):

jordansexton:
Published in @solana/spl-token@0.3.2

raymondfeng:
This change is breaking TypeScript node.js import of `@solana/spl-token`:

```
✖ ERROR: Error [ERR_REQUIRE_ESM]: require() of ES Module node_modules/@solana/spl-token/lib/cjs/index.js from solana.client.js not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in node_modules/@solana/spl-token/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

jordansexton:
@raymondfeng I'm not sure what `solana.client.js` or where this error is coming from. Can you open a new issue with details on how to reproduce the issue?

PR #3510, opened 2022-08-22
token-cli: Support the `InterestBearingMint` extension

### LINKED ISSUE #3406: token-cli: Support the `InterestBearingMint` extension
#### Problem

Token-2022 supports interest-bearing tokens through the `InterestBearingMint` extension, but the CLI can't interact with the extension yet.

#### Solution

Support the interest-bearing mint extension by adding features to:

* initialize a mint with interest using a flag: `--interest-bearing <RATE_AUTHORITY_ADDRESS> <RATE_IN_BASIS_POINTS>`
* set the interest rate using a new command: `set-interest-rate <NEW_RATE_IN_BASIS_POINTS>` with an optional `--rate-authority <SIGNER>` flag

udbhav1:
resolves #3406 

i used `handle_tx` for `command_set_interest_rate` because i was getting a trait bound error trying to pass an `Arc<dyn Signer>` into the token client's `update_interest_rate`

not super happy with how i handled the two params for `--interest-bearing <RATE_AUTHORITY> <RATE_BPS>` - i made the former optional by checking if it was "NONE" because i couldn't find a way to do this natively in clap

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1,6 +1,6 @@
 use clap::{
-    crate_description, crate_name, crate_version, value_t, value_t_or_exit, App,
-    AppSettings, Arg, ArgMatches, SubCommand,
+    crate_description, crate_name, crate_version, value_t, value_t_or_exit, App, AppSettings, Arg,
+    ArgMatches, SubCommand,
 };
 use serde::Serialize;
 use solana_account_decoder::{
@@ -41,7 +41,10 @@ use spl_associated_token_account::{
     get_associated_token_address_with_program_id, instruction::create_associated_token_account,
 };
 use spl_token_2022::{
-    extension::{interest_bearing_mint, mint_close_authority::MintCloseAuthority, StateWithExtensionsOwned},
+    extension::{
+        interest_bearing_mint, interest_bearing_mint::InterestBearingConfig,
+        mint_close_authority::MintCloseAuthority, StateWithExtensionsOwned,
+    },
     instruction::*,
     state::{Account, Mint, Multisig},
 };
@@ -362,7 +365,6 @@ async fn command_create_token(
     enable_freeze: bool,
     enable_close: bool,
     memo: Option<String>,
-    rate_authority: Option<Pubkey>,
     rate_bps: Option<i16>,
     bulk_signers: Vec<Arc<dyn Signer>>,
 ) -> CommandResult {
@@ -379,13 +381,12 @@ async fn command_create_token(
             close_authority: Some(authority),
         });
     }
+
     if let Some(rate_bps) = rate_bps {
-        extensions.push(
-            spl_token_client::token::ExtensionInitializationParams::InterestBearingConfig {
-                rate_authority,
-                rate: rate_bps,
-            },
-        )
+        extensions.push(ExtensionInitializationParams::InterestBearingConfig {
+            rate_authority: Some(authority),
+            rate: rate_bps,
+        })
     }
 
     if let Some(text) = memo {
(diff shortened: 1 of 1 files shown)


2501babe:
nice work! ill be able to look at this in more detail later

as for the token client, ive been in the process of reworking it to implement `close-mint`. you can see in full the approach ive been taking in #3487, but the tldr is we probably want to change the client interface to look like this:
```
    /// Update interest rate
    pub async fn update_interest_rate<S: Signers>(
        &self,
        authority: &Pubkey,
        new_rate: i16,
        signing_keypairs: &S,
    ) -> TokenResult<T::Output> {
```

token/cli/src/main.rs
```diff
@@ -1961,9 +2026,45 @@ fn app<'a, 'b>(
                             "Enable the mint authority to freeze associated token accounts."
                         ),
                 )
+                .arg(
+                    Arg::with_name("interest_bearing")
+                        .long("interest-bearing")
+                        .value_names(&["ADDRESS", "RATE"])
+                        .help("Specify the interest rate authority (or NONE) and rate in basis points"),
+                )
```
2501babe, line 2034: youre right that this is a tricky interface. i dont like the double parameter, it feels confusing and un-unixy. im thinking the most reasonable thing to do is make the flag `--interest-rate`, accept just the rate, and make the authority same as mint authority (as is done with `--enable-freeze` and `--enable-close` already). users can use `set-authority` to change this (it turns out `command_authorize` doesnt support all the new types yet, but ill open a separate issue, dont worry about this for your own pr)

token/cli/src/main.rs
```diff
@@ -2702,6 +2803,23 @@ async fn process_command<'a>(
...
+                    is_valid_pubkey(rate_addr)?;
+                    rate_authority = Some(config.pubkey_or_default(
+                        arg_matches,
+                        "interest_bearing",
+                        &mut wallet_manager,
+                    ));
+                }
+                rate_bps = Some(values[1].parse::<i16>().unwrap());
+            }
+
```
2501babe, line 2822: this becomes a much more pleasant `values_t!` call if we change the flag to one param

token/cli/src/main.rs
```diff
@@ -385,6 +397,59 @@ async fn command_create_token(
...
+    );
+
+    let minimum_balance_for_rent_exemption = if !config.sign_only {
+        config
+            .program_client
+            .get_minimum_balance_for_rent_exemption(Account::LEN)
+            .await?
+    } else {
+        0
+    };
```
2501babe, line 422: this is unnecessary, no account is being created so the arg passed to `handle_tx` can always be zero. however we should do basic sanity checks inside the `if !config.sign_only` block. off the top of my head i would... fetch mint, check that its owned by token22, check that it has the interest extension, check that the rate authority is correct. you dont need to write this, i can add it after if you prefer

token/cli/src/main.rs
```diff
@@ -356,13 +359,22 @@ async fn command_create_token(
...
     let freeze_authority = if enable_freeze { Some(authority) } else { None };
+    let mut extensions = vec![];
+    if let Some(rate_bps) = rate_bps {
+        extensions.push(
+            spl_token_client::token::ExtensionInitializationParams::InterestBearingConfig {
+                rate_authority,
+                rate: rate_bps,
+            },
+        )
+    }
```
2501babe, line 370: youll have to rebase against master because we create this vector now (as part of the close-authority pr), but you shouldnt have to do anything different otherwise

2501babe:
good work! i think if you could rebase against master and change the flag as described, this will be good to merge. i can handle the sanity checks on the mint if youd prefer (if so do you mind if i push to your branch?). im still in the process of converting client interfaces so we can use `handle_tx` for now

udbhav1:
fixed the flag and hopefully i rebased properly. i think i'd rather have you do the mint sanity checks and i don't mind if you use my branch. i left a `// TODO` at the appropriate location. thanks for the review!

2501babe:
thanks for your contribution!

Issue #3509, opened 2022-08-22
token-cli: bulk_signers has too much stuff


2501babe:
theres two things here:
* ~keys are duplicated all over the place. this is trivial, wrap `push()` in `contains()` checks~
* payer shouldnt be in the vec if its not also an authority on something, because in this case it doesnt need to be a signer on the instruction, and `construct_tx` handles signing on the transaction. this can be fixed after everything is refactored to go through the client

2501babe:
to expand on the rationale, since i lost the place where i orginally explained it: the token client takes `signing_keypairs` (which is called `bulk_signers` in the cli) and checks to see if it the the signer for the `authority: &Pubkey` arg that almost all functions take. if so, it passes an empty array to the token instruction constructor, and the the instruction will be signed with authority only. if not, it presumes the array contains signers for a multisig, and provides them to the instruction so their pubkeys can all be marked on the instruction

this means we want to strip down `bulk_signers` to just the authority in the non-multisig case, so it doesnt demand signatures from accounts it doesnt need them from

PR #3511, opened 2022-08-22
token-cli: dont push duplicates to bulk_signers

2501babe:
half of #3509


joncinque (approved):
Unasked-for-flyby! Looks good -- up to you if you want to wrap these in a separate function

PR #3518, opened 2022-08-23
token-cli: remove #[allow(deprecated)] from toplevel

### LINKED ISSUE #3497: token-cli: remove `#![allow(deprecated)]`
`// TODO: Remove when SPL upgrades to Solana 1.8` says it all. this covers up a handful of legitimate deprecations that need to be resolved

2501babe:
i allowed unchecked transfer to be deprecated because the purpose of using it is to make a deprecated thing available (although i do wonder whether anyone actually needs this). everything else is straightforward

closes #3497

### DIFF
--- a/token/cli/src/bench.rs
+++ b/token/cli/src/bench.rs
@@ -12,12 +12,7 @@ use {
     },
     solana_remote_wallet::remote_wallet::RemoteWalletManager,
     solana_sdk::{
-        message::Message,
-        native_token::Sol,
-        program_pack::Pack,
-        pubkey::Pubkey,
-        signature::{Signature, Signer, SignerError},
-        signers::Signers,
+        message::Message, native_token::Sol, program_pack::Pack, pubkey::Pubkey, signature::Signer,
         system_instruction,
     },
     spl_associated_token_account::*,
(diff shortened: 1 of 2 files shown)


token/cli/src/bench.rs
```diff
@@ -484,16 +484,9 @@ async fn send_messages(
...
-        .get_recent_blockhash_with_commitment(config.rpc_client.commitment())
-        .await?
-        .value;
-
-    lamports_required += messages
-        .iter()
-        .map(|message| fee_calculator.calculate_fee(message))
-        .sum::<u64>();
+    for message in messages.iter() {
+        lamports_required += config.rpc_client.get_fee_for_message(message).await?;
```
joncinque, line 488: This will probably slow things down unfortunately.  Check out the recent discussion in new-hire-learning-corner on Discord.  It looks like this is always called with the same kind of message, so you can fetch the fee for the first message, and then assume that all others will cost the same amount.

Also, this will probably weirdly error because you need a blockhash provided on the message in order to fetch a fee. I've been had on that *a lot*, so be sure to fetch a blockhash, get one fee, and multiply to greatness.

What do you think?

joncinque (commented):
Let me know what you think about the `get_fee_for_message` bit, otherwise looks perfect

token/cli/src/bench.rs
2501babe: after reviewing what the code is doing in more detail, yep this all makes sense to me

2501babe:
oki i redid the fee calculation and confirmed it works with `create-accounts` and `close-accounts`. `messages` is always homogenous so just grabbing one and multiplying should be safe. also i noticed some leftover code from the `Arc<dyn Signer>` switchover in here to remove

joncinque (approved):
Awesome, and thanks for the extra cleanup for the Arc<Signer>

PR #3523, opened 2022-08-24
token-swap: Add transfer fee and mint close authority support

joncinque:
#### Problem

On-chain programs that use Token-2022 could see some surprising behavior if mints have extensions.  Two situations in particular:

* if a mint has transfer fees, then token-swap slippage calculations are wrong
* if the pool token mint has a close authority, it's possible to brick a pool

#### Solution

Cover these situations, and also add some docs for it.  This builds on #3491 and #3524

### DIFF
--- a/token-swap/program/src/processor.rs
+++ b/token-swap/program/src/processor.rs
@@ -31,7 +31,10 @@ use solana_program::{
 use spl_token_2022::{
     check_spl_token_program_account,
     error::TokenError,
-    extension::{transfer_fee::TransferFeeConfig, StateWithExtensions},
+    extension::{
+        mint_close_authority::MintCloseAuthority, transfer_fee::TransferFeeConfig,
+        StateWithExtensions,
+    },
     state::{Account, Mint},
 };
 use std::{convert::TryInto, error::Error};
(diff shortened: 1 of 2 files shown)


2501babe:
is this the main pr for the token swap processor.rs changes, or is this based on another branch/pr? wasnt sure with the `docs:` label. i think i want to look at the swap changes there are some things in here i dont understand and make me frown

joncinque:
This is based on #3491 which contains more refactoring to use `transfer_checked`.  I wanted to put the docs and program changes in one place so they could be seen together, especially because the steps in the docs directly refer to commits, but I can separate the PRs too.

I went with "docs" since the main reason to make these changes is for token-2022 docs, but that does hide the underlying program changes, hm... It would probably make more sense to separate these PRs. huh

2501babe:
i dont think its worth the extra work to separate it out, id just spotlight the code change in the message and note the doc change as extra

token-swap/program/src/processor.rs
```diff
@@ -357,6 +393,8 @@ impl Processor {
         let destination_info = next_account_info(account_info_iter)?;
         let pool_mint_info = next_account_info(account_info_iter)?;
         let pool_fee_account_info = next_account_info(account_info_iter)?;
+        let source_token_mint_info = next_account_info(account_info_iter)?;
+        let destination_token_mint_info = next_account_info(account_info_iter)?;
```
2501babe, line 397: ill review the rest of the pr in more detail but i think im ok with this. the main thing that bothered me was the addition of the mints without checking their validity, but i see they flow through `transfer_checked` in all cases. still it might be nice to compare their addresses to the mints on the token accounts after parsing those on line 445, for redundancy

joncinque: Exactly, the idea is to let the token program perform the mint checks.  If we were optimizing this program for compute units, like Token (sometimes), I would keep the checks out, but since this is a reference implementation, I'm happy to add them in.

token-swap/program/src/processor.rs
```diff
@@ -414,16 +473,62 @@ impl Processor {
...
+                &source_mint_data,
+                source_token_mint_info.owner,
+                token_swap.token_program_id(),
+            )?;
+            let amount = if let Ok(transfer_fee_config) =
+                source_mint.get_extension::<TransferFeeConfig>()
+            {
+                source_amount_swapped.saturating_add(
+                    transfer_fee_config
+                        .calculate_epoch_fee_with_out(Clock::get()?.epoch, source_amount_swapped)
```
samkim-crypto, line 499: I was initially confused by `calculate_epoch_fee_with_out` for a bit. I now understand what it is doing, but it feels more natural to me to have a function that computes the gross amount from the net amount than the fee from the net amount. It seems like what we ultimately need here is to compute the gross amount from the net amount as well.

I think the math could also be simplified a little bit, so let me meditate on this a little bit...

samkim-crypto (approved):
Overall, it looks good to me! Docs are easy to follow and the code changes all look reasonable. I think computing the fee from the net fee is a little weird since we go from the net fee --> gross fee --> fee and then adding the fee to net fee to get the gross again. But simplifying the fee calculation could be for a separate PR.

jacobcreech (approved):
Docs were pretty easy to follow. Loved that it went over a bunch of gotchas and edge cases to avoid others messing up in the future

2501babe (dismissed):

PR #3524, opened 2022-08-24
token-2022: Add inverse fee calculation

joncinque:
#### Problem

While working on #3523 for token swap, the program needed to figure out how much to send in order to get a particular output.  This seems like it could be generally useful.

#### Solution

Add the inverse fee calculation.  Getting this to work with the ceiling calculation in the normal transfer fee was tricky, but the proptest works with really good precision, one part in 10^15, which seems pretty darn good!

Also, happy to bikeshed a better name for this.

### DIFF
--- a/token/program-2022/src/extension/transfer_fee/mod.rs
+++ b/token/program-2022/src/extension/transfer_fee/mod.rs
@@ -33,12 +33,12 @@ pub struct TransferFee {
 }
 impl TransferFee {
     /// Calculate the transfer fee
-    pub fn calculate(&self, amount: u64) -> Option<u64> {
+    pub fn calculate_fee(&self, pre_fee_amount: u64) -> Option<u64> {
         let transfer_fee_basis_points = u16::from(self.transfer_fee_basis_points) as u128;
-        if transfer_fee_basis_points == 0 || amount == 0 {
+        if transfer_fee_basis_points == 0 || pre_fee_amount == 0 {
             Some(0)
         } else {
-            let numerator = (amount as u128).checked_mul(transfer_fee_basis_points)?;
+            let numerator = (pre_fee_amount as u128).checked_mul(transfer_fee_basis_points)?;
             let mut raw_fee = numerator.checked_div(ONE_IN_BASIS_POINTS)?;
             let remainder = numerator.checked_rem(ONE_IN_BASIS_POINTS)?;
             if remainder > 0 {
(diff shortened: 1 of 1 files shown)


CriesofCarrots:
> Also, happy to bikeshed a better name for this.

Just a flyby, but I do find the use of `_in` and `_out` in these method and variable names very confusing, especially `_with_out`... without what?
Perhaps use something like net and gross?

joncinque:
> Just a flyby, but I do find the use of _in and _out in these method and variable names very confusing, especially _with_out... without what?
Perhaps use something like net and gross

Totally fair, I was doing a shorter (read: less clear) version of the token-swap functions, which are `DepositSingleTokenTypeExactAmountIn` and `WithdrawSingleTokenTypeExactAmountOut`.

How about: `calculate_fee` with `pre_fee_amount` and `calculate_inverse_fee` with `post_fee_amount` as better names? Otherwise it might have to be too much of a mouthful, with something like `calculate_fee_with_exact_amount_in` and `calculate_fee_with_exact_amount_out`. I worry that net and gross may not help too much

CriesofCarrots:
> How about: `calculate_fee` with `pre_fee_amount` and `calculate_inverse_fee` with `post_fee_amount` as better names?

Would have to see how it plays out in all the usages, but so far I think that works a lot better!

joncinque:
> what about calculate_from_base and calculate_from_total?

Sorry, I can't figure out which is which unfortunately...

I went with `calculate_fee` and `calculate_inverse_fee` in the end

2501babe:
sorry, i deleted because i realized i was confused about what exactly these are doing

token/program-2022/src/extension/transfer_fee/mod.rs
```diff
@@ -49,6 +49,46 @@ impl TransferFee {
             Some(cmp::min(raw_fee, u64::from(self.maximum_fee)))
         }
     }
+
+    /// Calculate the fee that would produce the given output
+    pub fn calculate_inverse_fee(&self, post_fee_amount: u64) -> Option<u64> {
```
2501babe, line 54: ok it took me some notes to self and some review of https://github.com/solana-labs/solana-program-library/pull/3523 to understand the use case. the concept here is... transfer fees are deducted from the total credit to the recipient, not added to the total debit from the sender. the inverse fee is working backwards from a net amount, to determine the gross. and this is used in token-swap to take the amount the pool received and calculate the amount that must have been taken from the user. right?

joncinque: Yes exactly! And I was thinking that this could be useful more generally, similar to payment platforms that allow you to pay transfer fees for the recipient so they receive an exact amount.

2501babe (approved):

PR #3525, opened 2022-08-24
token-cli: Memo transfer extension

### LINKED ISSUE #3407: token-cli: Support the `MemoTransfer` extension
#### Problem

Token-2022 adds the ability to require a memo to be included in all incoming transfers to an account.  The CLI does not support this yet.

#### Solution

Add the support with a new command: `{enable,disable}-required-transfer-memos <ACCOUNT_OR_MINT>`, figure out the account address if it's a mint, with an optional param for the owner: `--owner <SENDER_TOKEN_OWNER_KEYPAIR>`.

If the account doesn't have the extension already, and there isn't enough room in the account, don't forget to add a reallocation!

AndoroidX:
Apart from required transfer memos, I've wrapped `reallocate_if_needed` into function. This would ease addition of other extensions to cli.

resolves #3407

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -4423,4 +4423,23 @@ mod tests {
             assert_eq!(ui_account.delegated_amount, None);
         }
     }
+
+    #[tokio::test]
+    #[serial]
+    async fn enable_required_transfer_memos() {
+        let (test_validator, payer) = new_validator_for_test().await;
+        let program_id = spl_token_2022::id();
+        let config = test_config(&test_validator, &payer, &program_id);
+        let token = create_token(&config, &payer).await;
+        let result = process_test_command(
+            &config,
+            &payer,
+            &[
+                "spl-token",
+                CommandName::EnableRequiredTransferMemos.into(),
+                &token.to_string(),
+            ],
+        ).await;
+        result.unwrap();
+    }
 }
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -50,6 +50,7 @@ use spl_token_client::{client::RpcClientResponse, token::Token};
 use std::{
     collections::HashMap, fmt::Display, process::exit, str::FromStr, string::ToString, sync::Arc,
 };
+use solana_account_decoder::parse_token_extension::UiExtension;
 use strum_macros::{EnumString, IntoStaticStr, ToString};
 
 mod config;
@@ -63,6 +64,7 @@ use sort::{is_supported_program, sort_and_parse_token_accounts};
 
 mod bench;
 use bench::*;
+use spl_token_2022::extension::{ExtensionType};
 
 struct CliSignerInfo {
     pub signers: Vec<Arc<dyn Signer>>,
@@ -99,6 +101,12 @@ pub const MINT_ADDRESS_ARG: ArgConstant<'static> = ArgConstant {
     name: "mint_address",
     long: "mint-address",
     help: "Address of mint that token account is associated with. Required by --sign-only",
+};
+
+pub const ACCOUNT_OR_MIN_ADDRESS_ARG: ArgConstant<'static> = ArgConstant {
+    name: "account-or-mint-address",
+    long: "account-or-mint-address",
+    help: "Either a token account address or mint address",
 };
 
 pub const MINT_DECIMALS_ARG: ArgConstant<'static> = ArgConstant {
@@ -145,6 +153,8 @@ pub enum CommandName {
     MultisigInfo,
     Gc,
     SyncNative,
+    EnableRequiredTransferMemos,
+    DisableRequiredTransferMemos
 }
 
 pub fn owner_address_arg<'a, 'b>() -> Arg<'a, 'b> {
@@ -167,6 +177,15 @@ pub fn owner_keypair_arg_with_value_name<'a, 'b>(value_name: &'static str) -> Ar
 
 pub fn owner_keypair_arg<'a, 'b>() -> Arg<'a, 'b> {
     owner_keypair_arg_with_value_name("OWNER_KEYPAIR")
+}
+
+pub fn account_or_mint_address_arg<'a, 'b>() -> Arg<'a, 'b> {
+    Arg::with_name(ACCOUNT_OR_MIN_ADDRESS_ARG.name)
+        .long(ACCOUNT_OR_MIN_ADDRESS_ARG.long)
+        .takes_value(true)
+        .value_name("ACCOUNT_OR_MINT_ADDRESS")
+        .validator(is_valid_pubkey)
+        .help(ACCOUNT_OR_MIN_ADDRESS_ARG.help)
 }
 
 pub fn mint_address_arg<'a, 'b>() -> Arg<'a, 'b> {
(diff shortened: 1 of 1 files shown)


AndoroidX:
@2501babe can I get a review on my PR pls

AndoroidX:
@joncinque , can you review my PR, please

token/cli/src/main.rs
```diff
@@ -101,6 +104,12 @@ pub const MINT_ADDRESS_ARG: ArgConstant<'static> = ArgConstant {
     help: "Address of mint that token account is associated with. Required by --sign-only",
 };
 
+pub const ACCOUNT_OR_MIN_ADDRESS_ARG: ArgConstant<'static> = ArgConstant {
```
2501babe, line 107: typo `MIN` but im not sure we want to keep this as per below

token/cli/src/main.rs
```diff
@@ -2771,6 +2949,32 @@ fn app<'a, 'b>(
...
+                .arg(
+                    account_or_mint_address_arg()
+                        .index(1)
+                        .takes_value(true)
+                        .required(true)
+                )
+                .arg(
+                    owner_address_arg()
+                )
+        )
```
2501babe, line 2964: this and the below should also provide
```
                .arg(multisig_signer_arg())
                .nonce_args(true)
                .offline_args()
```

token/cli/src/main.rs
```diff
@@ -2771,6 +2949,32 @@ fn app<'a, 'b>(
                         .help("Specify the specific token account address to sync"),
                 ),
         )
+        .subcommand(
+            SubCommand::with_name(CommandName::EnableRequiredTransferMemos.into())
+                .about("Enable required transfer memos for token account")
+                .arg(
+                    account_or_mint_address_arg()
```
2501babe, line 2956: i have mixed feelings about this. on the one hand, id prefer the interface to be consistent with existing functions, and nothing else behaves like this. unfortunately the existing functions are themselves somewhat inconsistent, because (i assume) a lot of the early work on the cli was before atas were adopted as standard, and (i further assume) we cant break backwards compat anytime in the near future

the typical ways to do this are:

accept the mint address and derive the ata address, providing an --address flag for an override (close, balance, address)
accept the account address only (burn, freeze, thaw)
im inclined to say we should go with the latter option for consistency rather than trying to be clever, and save a refactor of the various other commands for another day

token/cli/src/main.rs
```diff
@@ -1891,6 +1911,164 @@ async fn command_sync_native(
     })
 }
 
+// Both enable-required_transfer-mesos and disable-required_transfer-mesos, switches with enable bool
+async fn command_required_transfer_memos(
```
2501babe, line 1915: i like this being one function

token/cli/src/main.rs
```diff
@@ -1891,6 +1911,164 @@ async fn command_sync_native(
...
+    token_account_address: Pubkey,
+    owner: Pubkey,
+    bulk_signers: BulkSigners,
+    enable: bool,
```
2501babe, line 1919: maybe rename this `enable_memos`?

token/cli/src/main.rs
```diff
@@ -1891,6 +1911,164 @@ async fn command_sync_native(
...
+    };
+    let mut instructions: Vec<Instruction> = Vec::new();
+    let reallocate_instruction = reallocate_if_needed(
+        &program_id,
+        &token_account_address,
+        &owner,
+        vec![ExtensionType::MemoTransfer],
+        config,
+    )
+    .await?;
```
2501babe, line 1945: this should go inside the block that executes when `!config.sign_only` because it depends on rpc. however, both rpc calls inside the `reallocate_if_needed` function are extraneous. we should fetch the token account once and use/clone it as needed

it might be easier to eliminate the function and do the work inline, and we can save refactoring out for the day when its needed by other things (the only other extension i can think of is confidential transfer, and that is going to be a major effort to support)

token/cli/src/main.rs
```diff
@@ -1891,6 +1911,164 @@ async fn command_sync_native(
...
+            .get_account(&token_account_address)
+            .await
+            .map_err(|err| {
+                format!(
+                    "Token account {} does not exist: {}",
+                    token_account_address, err
+                )
+            })?
+            .owner
+    };
```
2501babe, line 1936: there are some helpful sanity checks to do here... particularly we like to call `check_owner`, and it would be nice to halt if they are trying to enable when enabled or disable when disabled. this is kind of rote, you can look at the recent commit 38a9e3af72c156cf10b48b979c963c4787e720a3 to get an idea, or i could just handle it

token/cli/src/main.rs
```diff
@@ -1891,6 +1911,164 @@ async fn command_sync_native(
...
+    token_account_pubkey: &Pubkey,
+    owner_pubkey: &Pubkey,
+    new_extension_types: Vec<ExtensionType>,
+    config: &Config<'_>,
+) -> Result<Option<Instruction>, Error> {
+    let token_account = config
+        .rpc_client
+        .get_token_account(token_account_pubkey)
+        .await?
+        .expect("Invalid token account pubkey");
```
2501babe, line 2000: as noted above, whether we keep the function or not, we shouldnt fetch

token/cli/src/main.rs
```diff
@@ -1891,6 +1911,164 @@ async fn command_sync_native(
...
+    let mut existing_extensions: Vec<ExtensionType> = token_account
+        .extensions
+        .iter()
+        .map(parse_ui_extension)
+        .collect();
```
2501babe, line 2005: instead of all this, we can use `StateWithExtensionsOwned::<Account>::unpack` on the raw account data and call `get_extension_types` on that

token/cli/src/main.rs
```diff
@@ -1891,6 +1911,164 @@ async fn command_sync_native(
...
+        UiExtension::ConfidentialTransferMint(_) => ExtensionType::ConfidentialTransferMint,
+        UiExtension::ConfidentialTransferAccount(_) => ExtensionType::ConfidentialTransferAccount,
+        UiExtension::DefaultAccountState(_) => ExtensionType::DefaultAccountState,
+        UiExtension::ImmutableOwner => ExtensionType::ImmutableOwner,
+        UiExtension::MemoTransfer(_) => ExtensionType::MemoTransfer,
+        UiExtension::NonTransferable => ExtensionType::NonTransferable,
+        UiExtension::InterestBearingConfig(_) => ExtensionType::InterestBearingConfig,
+        UiExtension::UnparseableExtension => panic!("Unparsable extension"),
+    }
+}
```
2501babe, line 2044: we can delete this when `get_extension_types` replaces the current parse method

token/cli/src/main.rs
```diff
@@ -1891,6 +1911,164 @@ async fn command_sync_native(
...
+    } else {
+        // It's a mint account
+        Ok(get_associated_token_address_with_program_id(
+            owner,
+            &mint_or_token_account_pubkey,
+            program_id,
+        ))
+    }
+}
+
```
2501babe, line 2071: we can delete this if we just take `TOKEN_ACCOUNT_ADDRESS`

token/cli/src/main.rs
```diff
@@ -1891,6 +1911,164 @@ async fn command_sync_native(
...
+        instructions.push(
+            spl_token_2022::extension::memo_transfer::instruction::disable_required_transfer_memos(
+                &program_id,
+                &token_account_address,
+                &owner,
+                &config.multisigner_pubkeys,
+            )?,
+        );
+    }
+    let tx_return = handle_tx(
```
2501babe, line 1968: note to jon when i tag him in that im merging extension implementations that use this as-is, and will do a second batch of token client conversions later

token/cli/src/main.rs
```diff
@@ -4371,4 +4619,38 @@ mod tests {
...
+            &config,
+            &payer,
+            &[
+                "spl-token",
+                CommandName::EnableRequiredTransferMemos.into(),
+                &token.to_string(),
+            ],
+        )
+        .await;
+        result.unwrap();
```
2501babe, line 4642: i would fetch the account and confirm the extension exists in its types here, and that it doesnt after the disable call

2501babe:
whew! thanks for your work, and sorry for the long review. im happy to handle whatever subset of issues you would rather me to myself. if youd like me to do them, let me know when youre done tackling the ones youd like to, and if it ok that i push to your branch (or maybe i can fork your branch to preserve your name on your history in my repo? i havent tried this on github before)

AndoroidX:
Thank you for your review. I briefly looked over the issues that you have mentioned, and it seems that I can handle all of those. I'll look deeper into that later today and tomorrow. I'll let you know if I need any help, thank you.

Also as a note for myself, I need to update the docs with these subcommands.

token/cli/src/main.rs
AndoroidX: resolved with 7e157dd

token/cli/src/main.rs
```diff
@@ -175,6 +186,15 @@ pub fn owner_keypair_arg<'a, 'b>() -> Arg<'a, 'b> {
...
 
+pub fn account_or_mint_address_arg<'a, 'b>() -> Arg<'a, 'b> {
+    Arg::with_name(ACCOUNT_OR_MIN_ADDRESS_ARG.name)
+        .long(ACCOUNT_OR_MIN_ADDRESS_ARG.long)
+        .takes_value(true)
+        .value_name("ACCOUNT_OR_MINT_ADDRESS")
+        .validator(is_valid_pubkey)
+        .help(ACCOUNT_OR_MIN_ADDRESS_ARG.help)
+}
+
```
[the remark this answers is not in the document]
AndoroidX: resolved with e4dc1d8

token/cli/src/main.rs
AndoroidX: resolved with e4dc1d8

token/cli/src/main.rs
AndoroidX: resolved with e4dc1d8

token/cli/src/main.rs
AndoroidX: resolved with d37c851

token/cli/src/main.rs
AndoroidX: resolved with d37c851

token/cli/src/main.rs
AndoroidX: resolved with d37c851

token/cli/src/main.rs
AndoroidX: resolved with d37c851

token/cli/src/main.rs
AndoroidX: resolved with d37c851

token/cli/src/main.rs
AndoroidX: owner sanity check resolved in 0eb49c3
halting command resolved in d37c851

AndoroidX:
I've fixed all of the issues that were pointed out. The only thing that I'm wondering about is what to do if config is sign only, the problem is with reallocation. Since we can't fetch rpc, we can't know what extensions are already in the account, and how much space we need to get after reallocation. Right now it just panics if config is sign only.
Of course, I can just include one instrution (enable/disable memo transfers) and forget about reallocation, but that seems a little bit off for me.
The other thing that I can do, is add additional arguments for the cli command. Something like 
```--without-reallocation``` 
and 
```--existing-extensions [list of existing extensions]```
Only one of those would be required of config is sign only. 
Or maybe I'm missing something...
@2501babe , what are your thoughts on this?

**Comment by 2501babe** [comment]
hmm on second thought maybe just forget the signer args for now. when i convert this to use the token client, what will happen is it tries to do a fetch through the `SednTransaction` interface and errors anyway. its better to just not have the feature for now, and add it later if a usecase shows up. as far as i know, downstream users only really depend on offline signing for `mint` and `transfer`

**Comment by AndoroidX** [comment]
If working with sign only is not necessary, the PR is ready for review and merge.
@2501babe @joncinque

File: token/cli/src/main.rs [hunk 15]
```diff
@@ -1916,6 +1919,104 @@ async fn command_sync_native(
...
 }
 
+// Both enable_required_transfer_mesos and disable_required_transfer_mesos
+// Switches with enable_memos bool
+async fn command_required_transfer_memos(
+    token_account_address: Pubkey,
+    owner: Pubkey,
+    bulk_signers: BulkSigners,
+    enable_memos: bool,
+    config: &Config<'_>,
```
joncinque, line 1929: nit: can you move this param to the top for consistency?

token/cli/src/main.rs
```diff
@@ -1916,6 +1919,104 @@ async fn command_sync_native(
     })
 }
 
+// Both enable_required_transfer_mesos and disable_required_transfer_mesos
+// Switches with enable_memos bool
+async fn command_required_transfer_memos(
+    token_account_address: Pubkey,
+    owner: Pubkey,
+    bulk_signers: BulkSigners,
+    enable_memos: bool,
```
joncinque, line 1928: nit: can just call this `enable`

token/cli/src/main.rs
```diff
@@ -1916,6 +1919,104 @@ async fn command_sync_native(
...
+        .get_account(&token_account_address)
+        .await
+        .map_err(|err| {
+            format!(
+                "Token account {} does not exist: {}",
+                token_account_address, err
+            )
+        })?;
+    let program_id = account_fetch.owner;
+    config.check_owner(&token_account_address, &program_id)?;
```
joncinque, line 1945: This should probably be `config.get_account_checked` now, and the program id can be pulled from the `config` safely

token/cli/src/main.rs
```diff
@@ -1916,6 +1919,104 @@ async fn command_sync_native(
...
+    let mut instructions: Vec<Instruction> = Vec::new();
+    // Reallocation (if needed)
+    let current_account_len = account_fetch.data.len();
+    let state_with_extension =
+        StateWithExtensionsOwned::<Account>::unpack(account_fetch.data.clone())?;
```
joncinque, line 1950: nit: since you're not using `account_fetch` later, there's no need to clone the data, just move the value in

token/cli/src/main.rs
```diff
@@ -1916,6 +1919,104 @@ async fn command_sync_native(
...
+                "Required memo transfer was already {}",
+                if extension_data {
+                    "enabled"
+                } else {
+                    "disabled"
+                }
+            ));
+        }
+    } else {
+        existing_extensions.extend_from_slice(&[ExtensionType::MemoTransfer]);
```
joncinque, line 1968: nit: prefer a simple `push` on the vec

joncinque (commented):
This looking really close! Just a couple of nits on my side then it's ready. Since there's been some changes around here, be sure to rebase and make sure everything's still ok.

token/cli/src/main.rs
2501babe: haha he had that before and i made the opposite suggestion

joncinque: Bah sorry, I hoped that wasn't the case. I'll retract this and go with yours

token/cli/src/main.rs
AndoroidX: resolved with 484279b

token/cli/src/main.rs
AndoroidX: resolved with 484279b

token/cli/src/main.rs
AndoroidX: resolved with 484279b

token/cli/src/main.rs
AndoroidX: resolved with 484279b

AndoroidX:
@joncinque thank you for your review. I have made the changes according to your review and fetched with the current version of master. How does it look like now?

joncinque (approved):
Looks good to me! I'll let @2501babe give the final OK and merge

2501babe:
thank you for your contribution!

PR #3527, opened 2022-08-24
token-cli: Removed usage of Account::LEN as a verificaton for token account

### LINKED ISSUE #3526: token-cli: Usage of Account::LEN as a verificaton for token account
The current version of token cli uses `account.data.len() == Account::LEN` is a `command_transfer` function for verifying that account is actually a token account.
```rust
let recipient_account_info = config
            .rpc_client
            .get_account_with_commitment(&recipient, config.rpc_client.commitment())
            .await?
            .value
            .map(|account| {
                (
                    account.owner == mint_info.program_id && account.data.len() == Account::LEN,
                    account.owner == system_program::id(),
                )
            });
        if let Some((recipient_is_token_account, recipient_is_system_account)) =
            recipient_account_info
```
With the introduction of extensions, accounts might have larger size and this verification will fail.
I assume that `account.owner == mint_info.program_id` is enough for verifying that this account is in fact token account.

I'll fix this issue by removing `&& account.data.len() == Account::LEN`

AndoroidX:
closes #3526

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -797,7 +797,7 @@ async fn command_transfer(
             .value
             .map(|account| {
                 (
-                    account.owner == mint_info.program_id && account.data.len() == Account::LEN,
+                    account.owner == mint_info.program_id,
                     account.owner == system_program::id(),
                 )
             });
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -66,6 +66,7 @@ use sort::{is_supported_program, sort_and_parse_token_accounts};
 
 mod bench;
 use bench::*;
+use spl_token_2022::generic_token_account::GenericTokenAccount;
 
 struct CliSignerInfo {
     pub signers: Vec<Arc<dyn Signer>>,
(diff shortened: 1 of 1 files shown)


joncinque:
Thanks for your contribution! This will relax the check a bit too strongly, however, so please use `valid_account_data()` from https://github.com/solana-labs/solana-program-library/blob/3fe92ea05ffa09b545fd88e8f226a10ee995e6c7/token/program-2022/src/state.rs#L304, which covers for all the situations in which an account is actually a token account.

AndoroidX:
Thank you. I'll check it out

AndoroidX:
Used the function mentioned by @joncinque, thank you.
Previously failed test passes locally.

AndoroidX:
@joncinque , seems that PR is ready for merge

token/cli/src/main.rs
```diff
@@ -66,6 +66,7 @@ use sort::{is_supported_program, sort_and_parse_token_accounts};
 
 mod bench;
 use bench::*;
+use spl_token_2022::generic_token_account::GenericTokenAccount;
```
joncinque, line 69: micro-nit for next time: we typically prefer to keep the `use` statements together, so it would have been better to put this up at line 44-48 with the rest of the `spl_token_2022` imports

joncinque (commented):
Thanks for your contribution!

token/cli/src/main.rs
AndoroidX: Got it. Thank you.

Issue #3482, opened 2022-08-16
token-cli: refactor more to go through token client


2501babe:
once #3469 lands, we can refactor more of the `command_` functions in the cli to go through the token client and eventually once they all do we can remove `handle_tx`

ill probably do a couple of these to see how rote it is and how much time it takes, then either do the rest or make issues for it

edit, copied from a pr, this is what a conversion looks like:
* the `S` on a client method type signature changes from `Signer` to `Signers`
* `authority` param changes from `&S` to `&Pubkey`
* methods take an additional param `signing_keypairs: &S`
* `get_multisig_signers` is called up top to gather pubkeys to mark as signers on token instructions
* return types change to `TokenResult<T::Output>` if not already, to enable returning signatures or (possibly partially) signed transactions
* cli flow changes, generally to:
  - create token client with `token_client_from_config`
  - call relevant token client method
  - produce `tx_return` from this result with `finish_tx`, removing all instruction creation logic and `handle_tx`
  - `match` on `tx_return` to produce user output remains unchanged
* update tests in `client/tests/` and `program-2022-tests/tests/` for the client interface change

2501babe:
based on my experience doing this with `set_authority` and `close_account`, its trivial to update the client, easy to update the cli, somewhat annoying to update the tests

2501babe:
the only remaining non-client cli commands are: create multisig, transfer, gc, transfer memos, and interest rate

PR #3528, opened 2022-08-24
run more cli stuff through client

2501babe:
part of #3482. i started with `mint_to`, will also do `burn` and `transfer` since we discussed unifying the client interface of those to take `decimals: Option<u8>` and get rid of checked/unchecked variants

i dunno if im gonna do all the cli stuff but this will provide a blueprint at least as for what the changes need to look like. in short:
* the `S` on a client method type signature changes from `Signer` to `Signers`
* `authority` param changes from `&S` to `&Pubkey`
* methods take an additional param `signing_keypairs: &S`
* `get_multisig_signers` is called up top to gather pubkeys to mark as signers on token instructions
* return types change to `TokenResult<T::Output>` if not already, to enable returning signatures or (possibly partially) signed transactions
* cli flow changes, generally to:
  - create token client with `token_client_from_config`
  - call relevant token client method
  - produce `tx_return` from this result with `finish_tx`, removing all instruction creation logic and `handle_tx`
  - `match` on `tx_return` to produce user output remains unchanged
* update tests in `client/tests/` and `program-2022-tests/tests/` for the client interface change

remembering that in our bright future we are not the only consumer of the token client so its interface should remain suitable for general use

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1152,23 +1152,12 @@ async fn command_freeze(
         ),
     );
 
-    let instructions = vec![freeze_account(
-        &mint_info.program_id,
-        &account,
-        &mint_info.address,
-        &freeze_authority,
-        &config.multisigner_pubkeys,
-    )?];
-    let tx_return = handle_tx(
-        &CliSignerInfo {
-            signers: bulk_signers,
-        },
-        config,
-        false,
-        0,
-        instructions,
-    )
-    .await?;
+    let token = token_client_from_config(config, &mint_info.program_id, &mint_info.address);
+    let res = token
+        .freeze(&account, &freeze_authority, &bulk_signers)
+        .await?;
+
+    let tx_return = finish_tx(config, &res, false).await?;
     Ok(match tx_return {
         TransactionReturnData::CliSignature(signature) => {
             config.output_format.formatted_string(&signature)
@@ -1197,23 +1186,12 @@ async fn command_thaw(
         ),
     );
 
-    let instructions = vec![thaw_account(
-        &mint_info.program_id,
-        &account,
-        &mint_info.address,
-        &freeze_authority,
-        &config.multisigner_pubkeys,
-    )?];
-    let tx_return = handle_tx(
-        &CliSignerInfo {
-            signers: bulk_signers,
-        },
-        config,
-        false,
-        0,
-        instructions,
-    )
-    .await?;
+    let token = token_client_from_config(config, &mint_info.program_id, &mint_info.address);
+    let res = token
+        .thaw(&account, &freeze_authority, &bulk_signers)
+        .await?;
+
+    let tx_return = finish_tx(config, &res, false).await?;
     Ok(match tx_return {
         TransactionReturnData::CliSignature(signature) => {
             config.output_format.formatted_string(&signature)
@@ -1478,22 +1456,10 @@ async fn command_revoke(
         return Err(format!("No delegate on account {}", account).into());
     }
 
-    let instructions = vec![revoke(
-        &program_id,
-        &account,
-        &owner,
-        &config.multisigner_pubkeys,
-    )?];
-    let tx_return = handle_tx(
-        &CliSignerInfo {
-            signers: bulk_signers,
-        },
-        config,
-        false,
-        0,
-        instructions,
-    )
-    .await?;
+    let token = token_client_from_config(config, &program_id, &Pubkey::default());
+    let res = token.revoke(&account, &owner, &bulk_signers).await?;
+
+    let tx_return = finish_tx(config, &res, false).await?;
     Ok(match tx_return {
         TransactionReturnData::CliSignature(signature) => {
             config.output_format.formatted_string(&signature)
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1378,38 +1378,18 @@ async fn command_approve(
     let mint_address = config.check_account(&account, mint_address).await?;
     let mint_info = config.get_mint_info(&mint_address, mint_decimals).await?;
     let amount = spl_token::ui_amount_to_amount(ui_amount, mint_info.decimals);
-
-    let instructions = if use_unchecked_instruction {
-        vec![approve(
-            &mint_info.program_id,
-            &account,
-            &delegate,
-            &owner,
-            &config.multisigner_pubkeys,
-            amount,
-        )?]
+    let decimals = if use_unchecked_instruction {
+        None
     } else {
-        vec![approve_checked(
-            &mint_info.program_id,
-            &account,
-            &mint_info.address,
-            &delegate,
-            &owner,
-            &config.multisigner_pubkeys,
-            amount,
-            mint_info.decimals,
-        )?]
+        Some(mint_info.decimals)
     };
-    let tx_return = handle_tx(
-        &CliSignerInfo {
-            signers: bulk_signers,
-        },
-        config,
-        false,
-        0,
-        instructions,
-    )
-    .await?;
+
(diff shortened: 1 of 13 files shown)


2501babe:
actually im gonna update the transfer client interface how we want it but leave the cli as it is, the command is complicated and i wanna give it proper attention

2501babe:
alright cb67630ec6b54b49fa05e4dce3ab4ccd95368a52 converts burn to the new regime and does nothing else so it serves as a good example of what a typical conversion will look like

2501babe:
i converted all "standard" token functions (mint, burn, transfer, freeze, thaw, approve, revoke) on the client to the new interface that supports multisigning, used the client for all of these commands in the cli except transfer (which is complex enough to merit special attention), added support for memo to mint, and fixed an issue where the token client used owner instead of payer for paying fees

token/cli/src/main.rs
```diff
@@ -1409,22 +1342,10 @@ async fn command_revoke(
...
-        &CliSignerInfo {
-            signers: bulk_signers,
-        },
-        config,
-        false,
-        0,
-        instructions,
-    )
-    .await?;
+    let token = token_client_from_config(config, &program_id, &Pubkey::default());
```
joncinque, line 1345: nit: The final parameter of `&Pubkey::default()` absolutely works but is kinda hacky. I don't feel too strongly about it, but how about also pulling out the mint pubkey in the first branch higher up, and using `Pubkey::default()` in the offline signing case with a comment explaining why it's ok.  Or at the very least, a comment explaining why this parameter is ok

token/program-2022-test/tests/transfer_fee.rs
```diff
@@ -793,7 +799,14 @@ async fn transfer_checked() {
 
     // fail unchecked always
     let error = token
-        .transfer_unchecked(&alice_account, &bob_account, &alice, maximum_fee)
+        .transfer(
+            &alice_account,
+            &alice_account,
```
joncinque, line 804: nit: Let's keep this as `&bob_account`

token/program-2022-test/tests/transfer_fee.rs
```diff
@@ -808,12 +821,13 @@ async fn transfer_checked() {
 
     // fail because amount too high
     let error = token
-        .transfer_checked(
+        .transfer(
             &alice_account,
-            &bob_account,
-            &alice,
+            &alice_account,
```
joncinque, line 826: nit: same here

token/program-2022-test/tests/delegate.rs
```diff
@@ -114,26 +129,57 @@ async fn run_basic(
...
+            &bob.pubkey(),
+            1,
+            Some(decimals),
+            &vec![&bob],
+        )
         .await
         .unwrap();
 
     // wrong signer
+    let keypair = &Keypair::new();
```
joncinque, line 173: TIL: I didn't know you could directly grab a reference to something during creation like this

token/client/src/token.rs
```diff
@@ -544,12 +558,7 @@ where
         authority_type: instruction::AuthorityType,
         signing_keypairs: &S,
     ) -> TokenResult<T::Output> {
-        let signing_pubkeys = signing_keypairs.pubkeys();
-        let multisig_signing_pubkeys = if signing_pubkeys == [*authority] {
-            vec![]
-        } else {
-            signing_pubkeys.iter().collect::<Vec<_>>()
-        };
+        let multisig_signers = self.get_multisig_signers(authority, signing_keypairs);
```
joncinque, line 561: Take it or leave it, but if you want to avoid the `collect` everywhere, you can have `get_multisig_signers` return `Vec<&Pubkey>` by doing something like:

```
        let signing_pubkeys = signing_keypairs.pubkeys();
        let multisig_signers = self.get_multisig_signers(authority, &signing_pubkeys);
```

and then `get_multisig_signers` looks like:

```
if signing_pubkeys == [*authority] {
    vec![]
} else {
    signing_pubkeys.iter().collect::<Vec<_>>()
}

**Review by joncinque — COMMENTED** [review]
Looks really great! Just a few nits / questions, but nothing major

File: token/program-2022-test/tests/transfer_fee.rs (hunk 2 above)
Comment (2501babe) [reply token/program-2022-test/tests/transfer_fee.rs:804]: aaah thank you, this was a legitimate mistake from overzealous copy-paste while fixing the many test invocations

File: token/client/src/token.rs (hunk 5 above)
Comment (2501babe) [reply token/client/src/token.rs:561]: this is smarter yes. i got hung up on not being able to create the pubkeys inside the function

File: token/cli/src/main.rs (hunk 1 above)
Comment (2501babe) [reply token/cli/src/main.rs:1345]: this probably sounds stupid but the lack of comments on anything has made me wonder if we arent supposed to include comments as a style thing

Comment (joncinque) [reply token/cli/src/main.rs:1345]: I don't think there's any style in particular.  At the very least, I like to require comments where the code may not be clear, and beyond that, it's up to the person. In the monorepo, there are loads of PRs just adding comments to unclear code, so it's certainly appreciated.

Comment (CriesofCarrots) [reply token/cli/src/main.rs]: It'd actually be really cool to add robust rust-docs to spl-token-client (like we've done for solana-client), since that's a dependency that's likely to be pulled in by external rust users.

**Comment by 2501babe** [comment]
took the suggestion with `get_multisig_signers` (and wrote my own lifetime annotations for the first time! it turns out theres just like three trivial rules that dictate when you need to annotate things for the compiler lol, for some reason i had the idea that lifetimes were this complex arcane thing that let you do dark magic like override scoping rules). also fixed the other places where i used `Pubkey::default()` in previous prs

File: token/cli/src/main.rs [hunk 6]
```diff
@@ -749,17 +751,20 @@ async fn command_authorize(
...
+            Ok((token_account.base.mint, previous_authority))
         } else {
             Err("Unsupported account data format".to_string())
         }?;
 
-        (program_id, previous_authority)
+        (program_id, mint_pubkey, previous_authority)
     } else {
-        (config.program_id, COption::None)
+        // default is safe here because revoke doesnt use it
```
joncinque, line 763: micro-nit: comment should say "authorize" instead of "revoke"

token/cli/src/main.rs
```diff
@@ -1480,12 +1486,13 @@ async fn command_close(
             .into());
         }
 
-        program_id
+        (program_id, source_state.base.mint)
     } else {
-        config.program_id
+        // default is safe here because revoke doesnt use it
```
joncinque, line 1491: micro-nit: "close" instead of "revoke"

token/client/src/token.rs
```diff
@@ -281,17 +281,15 @@ where
...
         &self,
-        authority: &Pubkey,
-        signing_keypairs: &S,
-    ) -> Vec<Pubkey> {
-        let signing_pubkeys = signing_keypairs.pubkeys();
-
-        if signing_pubkeys == [*authority] {
+        authority: &'a Pubkey,
+        signing_pubkeys: &'a Vec<Pubkey>,
+    ) -> Vec<&'a Pubkey> {
```
joncinque, line 288: nit only because you mentioned this was your first time annotating lifetimes:

since the return is a reference to `signing_pubkeys`, you can do a slightly more flexible version
```
fn get_multisig_signers<'a, 'b>(
        &self,
        authority: &'b Pubkey,
        signing_pubkeys: &'a Vec<Pubkey>,
    ) -> Vec<&'a Pubkey> {
```

This means that the lifetimes for `authority` and `signing_pubkeys` don't need to be the same. Most likely this would almost never be useful since it's called internally, but thought you like the info :smile:

joncinque (approved):
R+ a few last little bits

Issue #3522, opened 2022-08-24
debug: @solana/spl-token@0.3.2 breaks TypeScript imports


raymondfeng:
`@solana/spl-token@0.3.2` is published as an ESM and it breaks existing Node.js usage with TypeScript.

To reproduce the issue, please use the repo below and follow instructions in README.md.

https://github.com/raymondfeng/spl-token-import-issue

jordansexton:
This is because your build config is wacky.

Add this to package.json:
```json
  "type": "module",
```

Add this to your tsconfig.json `compilerOptions`:
```json
    "module": "ES2020",
    "target": "ES2020",
    "moduleResolution": "Node",
```

Build and run:
```shell
tsc
node .

###

PublicKey {
  _bn: <BN: 6ddf6e1d765a193d9cbe146ceeb79ac1cb485ed5f5b37913a8cf5857eff00a9>
}
```

jordansexton:
I'll try to get this working with CJS though.

raymondfeng:
> Add this to package.json:

  "type": "module",

This is not feasible for us at the moment as we still use CommonJS for our applications/packages.

0xdeepmehta:
also gibbing this error in `"@solana/spl-token": "latest"` package. The previous versions are working fine.
`TypeError: Unknown file extension ".ts" for /Users/0xdeep/Dev/web3.0/test/sold/tests/sold.ts`

jordansexton:
I think I have a fix for this.

It works by adding the following package.json to `node_modules/@solana/spl-token/lib/cjs`:
```json
{
    "type": "commonjs"
}
```

jordansexton:
I'll ship an alpha release with this, please stand by to test!

jordansexton:
Published as
```
alpha: 0.3.4-alpha.0
```

Please test this out!

jordansexton:
Works for me with https://github.com/raymondfeng/spl-token-import-issue and also works with the ESM config from https://github.com/solana-labs/solana-program-library/issues/3522#issuecomment-1226274881

jordansexton:
Will release as 0.3.4 when CI @ https://github.com/solana-labs/solana-program-library/pull/3529 succeeds.

jordansexton:
Published
- @solana/spl-memo@0.2.2
- @solana/spl-token-0.3.4

raymondfeng:
👍

PR #3529, opened 2022-08-24
spl-token: fix CJS imports

jordansexton:
Fixes https://github.com/solana-labs/solana-program-library/issues/3522


jordansexton:
Published
- @solana/spl-memo@0.2.2
- @solana/spl-token-0.3.4

Issue #3519, opened 2022-08-24
[token-2022] Restrict deposit source and withdraw destination in confidential extension to base accounts


samkim-crypto:
In the confidential extension, it is currently possible to deposit and withdraw from the confidential extension from different source and destination accounts other than the base account for the extension. Deposit and withdraw can basically act as an alternate way to transfer tokens from accounts.

This is fine when the mint is only extended for confidential transfers, but it causes problems when the mint is extended for other extensions. For instance, using deposit for transfers could be used to bypass transfer fees.

We can add custom logic for some of the extensions for the deposit and withdraw, but I think the best way forward is to simply require that deposits/withdraws to/from confidential extensions are always directed to/from the base account. 

Wdyt @mvines ? I just want to double check that I am not missing anything about the usefulness of deposit/withdraw from different accounts.

cc @joncinque @CriesofCarrots @2501babe

mvines:
Yes I'm ok with restricting to the base account. Not really a big loss of functionality to add this restriction and it is nice to disarm this footgun for future extensions

joncinque:
Fixed with #3540

PR #3540, opened 2022-08-27
confidential-extension: Restrict deposit source and withdraw dest to base accounts

samkim-crypto:
Currently, the deposit source and withdraw destination accounts for the confidential extension instructions can differ from the base account for the extension, which can act as an alternate way to transfer tokens (issue [#3519](https://github.com/solana-labs/solana-program-library/issues/3519)). 

Summary:
- limit the deposit source and withdraw destination accounts to base accounts only
- limit deposit amounts to a 48-bit number

### DIFF
--- a/token/program-2022/src/extension/confidential_transfer/instruction.rs
+++ b/token/program-2022/src/extension/confidential_transfer/instruction.rs
@@ -160,17 +160,17 @@ pub enum ConfidentialTransferInstruction {
     /// Accounts expected by this instruction:
     ///
     ///   * Single owner/delegate
-    ///   1. `[writable]` The SPL Token account.
-    ///   2. `[]` The token mint.
-    ///   3. `[]` Instructions sysvar.
-    ///   4. `[signer]` The single source account owner.
+    ///   0. `[writable]` The SPL Token account.
+    ///   1. `[]` The token mint.
+    ///   2. `[]` Instructions sysvar.
+    ///   3. `[signer]` The single source account owner.
     ///
     ///   * Multisignature owner/delegate
-    ///   1. `[writable]` The SPL Token account.
-    ///   2. `[]` The token mint.
-    ///   3. `[]` Instructions sysvar.
-    ///   4. `[]` The multisig  source account owner.
-    ///   5.. `[signer]` Required M signer accounts for the SPL Token Multisig account.
+    ///   0. `[writable]` The SPL Token account.
+    ///   1. `[]` The token mint.
+    ///   2. `[]` Instructions sysvar.
+    ///   3. `[]` The multisig  source account owner.
+    ///   4.. `[signer]` Required M signer accounts for the SPL Token Multisig account.
     ///
     /// Data expected by this instruction:
     ///   `WithdrawInstructionData`
--- a/token/program-2022/src/extension/confidential_transfer/instruction.rs
+++ b/token/program-2022/src/extension/confidential_transfer/instruction.rs
@@ -131,23 +131,25 @@ pub enum ConfidentialTransferInstruction {
     /// The account owner can then invoke the `ApplyPendingBalance` instruction to roll the deposit
     /// into their available balance at a time of their choosing.
     ///
+<<<<<<< HEAD
     /// Fails if the source or destination accounts are frozen.
     /// Fails if the associated mint is extended as `NonTransferable`.
+=======
+    /// Fails if the account is frozen.
+>>>>>>> 1668db8e (restrict deposit source account to base account)
     ///
     /// Accounts expected by this instruction:
     ///
     ///   * Single owner/delegate
-    ///   0. `[writable]` The source SPL Token account.
-    ///   1. `[writable]` The destination SPL Token account with confidential transfers configured.
-    ///   2. `[]` The token mint.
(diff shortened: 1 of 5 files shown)


2501babe:
ive been backlogged on reviews but i should be able to do this and your other one before the long weekend

token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -155,20 +153,18 @@ pub enum ConfidentialTransferInstruction {
...
-    ///   1. `[writable]` The destination SPL Token account.
+    ///   1. `[writable]` The SPL Token account.
     ///   2. `[]` The token mint.
     ///   3. `[]` Instructions sysvar.
     ///   4. `[signer]` The single source account owner.
     ///
     ///   * Multisignature owner/delegate
-    ///   0. `[writable]` The source SPL Token account with confidential transfers configured.
-    ///   1. `[writable]` The destination SPL Token account.
+    ///   1. `[writable]` The SPL Token account.
```
2501babe, line 167: these should be zero-indexed for consistency with the rest of the instruction descriptions. also i notice the accounts in the comments for `Transfer` and `TransferWithFee` are one-indexed

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -279,94 +279,74 @@ fn process_deposit(
...
-    // Process destination account
-    {
-        check_program_account(destination_token_account_info.owner)?;
-        let destination_token_account_data = &mut destination_token_account_info.data.borrow_mut();
-        let mut destination_token_account =
-            StateWithExtensionsMut::<Account>::unpack(destination_token_account_data)?;
+    // A deposit amount must be a 48-bit number
+    if amount >> MAXIMUM_DEPOSIT_TRANSFER_AMOUNT_BIT_LENGTH > 0 {
+        return Err(TokenError::MaximumDepositAmountExceeded.into());
+    }
```
2501babe, line 308: is this sufficient to remediate HAL-03? legitimate question, i dont understand a lot of what happens in this extension to be honest

their comments in their report, and the documentation they linked [here](https://spl.solana.com/confidential-token/deep-dive/overview#dealing-with-discrete-log), seem to indicate balances must never exceed u48::MAX. but the math down below with `pending_balance_lo` and `pending_balance_hi` makes it seem like values up to u64::MAX are expected?

also, their comments state that HAL-03 breaks the decryption primitive via an "overflow," but the protocol overview doc claims a 48-bit limit only exists for performance reasons? i am doubly unsure about all of this after reading the zk token paper because it describes splitting the balance into two 32-bit numbers for performance reasons, so im not sure what the 16/48 split here is meant to be 

anyway, you dont need to explain the cryptography to me, i lack the background to appreciate it. assuming the HAL-03 report is correct, and this block is the only new safeguard against it, here are my attacks:
* i break my own account by depositing u48::MAX and then depositing again
* i break my own account by depositing u48::MAX and rolling that pending balance into a nonzero active balance. equivalently, i transfer to someone else and attempt to convince them to roll the pending into their nonzero active
* i break someone elses account by depositing u48::MAX into my own account, transfering to them, depositing again, transfering again. skip the second cycle if their balance was nonzero

do these work? or am i missing something?

samkim-crypto: You are understanding things correctly. All of these are valid points. Restricting the amounts to 48-bit values was added on after the paper was written, so it is a little confusing. I discuss this a little bit in the [spl doc](https://spl.solana.com/confidential-token/deep-dive/encryption), which is more up-to-date than the paper.

All the attacks that you mentioned are valid. I do want to mention though that if the funds do overflow, it is not that they are lost forever. It is just that decryption become noticably slower. If the funds overflow, then (1) one can spend more time decrypting it or (2) it can also look up individual transaction data that sent tokens to the confidential account and then decrypt the associated amounts individually (the amounts associated with the instructions are guaranteed to be 48-bits by zkp).

I can comment on the attacks in order:
- There is no mechanism to protect against this attack. If this happens by accident, then the owner can just do the costly decryption or decrypts the instruction data individually.
- Only pending balance is not decryptable. Once pending balance is rolled over to available balance, then the balances are decryptable because available balance is encrypted as decryptable balance.
- There is no mechanism to protect against this attack. If this happens, then the receiver has to do the costly decryption (after disabling incoming transfers if needed) or decrypt the individual instruction data. This is a more serious attack than the above two because a malicious sender can genuinely impact another user's accounts. Unfortunately, no mechanism is possible to prevent this without significantly impacting performance. Practically speaking, it would be a very costly attack for the malicious sender only to overflow an account which the receiver can still recover the funds for (albeit in a more costly way).

2501babe: ahhh i see. their line:
>Doing so will cause the
 Elgamal encryption addition function to malfunction/overflow and the
 users available_balance will not be decryptable thus leading to a loss
 of funds.

led me to believe this was more serious than merely slowing it down

2501babe (dismissed):

samkim-crypto:
Great, thanks for the reviews!

PR #3553, opened 2022-08-30
ci: Re-enable twotx job

joncinque:
#### Problem

This is another attempt at #3465 -- we disabled the twoxtx CI job because of incompatible rust versions, but we can actually pick up the correct version if we just use cargo-test-sbf from the twoxtx solana repo.

#### Solution

Just use that!

Fixes https://github.com/solana-labs/solana-program-library/issues/3461

### DIFF
--- a/.github/workflows/pull-request-token.yml
+++ b/.github/workflows/pull-request-token.yml
@@ -80,14 +80,13 @@ jobs:
           if-no-files-found: error
 
   cargo-test-sbf-twoxtx:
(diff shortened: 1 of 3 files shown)


joncinque:
Wow, looks like it just worked! I'll go ahead and merge then. :crossed_fingers: that it CI doesn't just start breaking randomly

samkim-crypto:
Great, thanks so much!

PR #3556, opened 2022-08-31
token-swap: Remove stable swap curve

joncinque:
#### Problem

The stable swap curve needs to removed.

#### Solution

Remove it and all references to it in the code.


elliotkennedy:
> 

@joncinque - why did this need to be removed? Was there an issue with the implementation?

Issue #3557, opened 2022-08-31
[tests] Tests using chai-as-promised are missing the return keyword


mwrites:
### Problem
Tests using [chai-as-promised](https://www.chaijs.com/plugins/chai-as-promised/) are missing the `return` keyword, consequently, some expectations are not awaited and **ignored**.

> either return or notify(done) must be used with promise assertions. This can be a slight departure from the existing format of assertions being used on a project or by a team. Those other assertions are likely synchronous and thus do not require special handling

### Example

I noticed the issue by trying to make one of the test fail.

On this [line](https://github.com/solana-labs/solana-program-library/blob/master/token/js/test/e2e/create.test.ts#L144), I switched the `rejected` for a `fulfilled` to purposely make the test fail, but it just passed:
```jsx
expect(
    createAccount(
        connection,
        payer,
        mint,
        owner.publicKey,
        undefined, // uses ATA by default
        undefined,
        TEST_PROGRAM_ID
    )
).to.be.fulfilled;
```
<img width="389" alt="image" src="https://user-images.githubusercontent.com/20499416/187747244-a389cb92-9515-434a-9c9b-cf064404259a.png">

By adding the `return` it fails as expected:
```jsx
return expect(
    createAccount(
        connection,
        payer,
        mint,
        owner.publicKey,
        undefined, // uses ATA by default
        undefined,
        TEST_PROGRAM_ID
    )
).to.be.fulfilled;
```
<img width="856" alt="image" src="https://user-images.githubusercontent.com/20499416/187747499-fb05c368-07ef-49a7-bfac-6784fb6cf589.png">

By the way, to reproduce, you might need to increase the timeout `mocha --timeout` since we rely on `await sendAndConfirmTransaction()` to know if a promise has been fulfilled or rejected.

`

PR #3558, opened 2022-08-31
[token-js] : JS binding for CreateIdempotent instruction

mwrites:
Added JS binding for [CreateIdempotent](https://github.com/solana-labs/solana-program-library/pull/2883).

See [issue/3557](https://github.com/solana-labs/solana-program-library/issues/3557) for the use of `return expect` in the tests.


token/js/src/actions/createAccount.ts
```diff
@@ -26,10 +26,11 @@ export async function createAccount(
     owner: PublicKey,
     keypair?: Keypair,
     confirmOptions?: ConfirmOptions,
-    programId = TOKEN_PROGRAM_ID
+    programId = TOKEN_PROGRAM_ID,
+    idemPotent = false,
```
joncinque, line 30: We already have a lot of null / undefined parameters in this function, so to simplify usage, let's have a separate `createAssociatedTokenAccountIdempotent` action which only does a create idempotent call in `creteAssociatedTokenAccount.ts`, and not have the flag here.

token/js/src/actions/createAssociatedTokenAccount.ts
```diff
@@ -24,7 +24,8 @@ export async function createAssociatedTokenAccount(
     owner: PublicKey,
     confirmOptions?: ConfirmOptions,
     programId = TOKEN_PROGRAM_ID,
-    associatedTokenProgramId = ASSOCIATED_TOKEN_PROGRAM_ID
+    associatedTokenProgramId = ASSOCIATED_TOKEN_PROGRAM_ID,
+    idemPotent = false,
```
joncinque, line 28: Same here -- let's not have this parameter and create a separate function `createAssociatedTokenAccountIdempotent`.

token/js/src/instructions/associatedTokenAccount.ts
```diff
@@ -20,7 +22,8 @@ export function createAssociatedTokenAccountInstruction(
     owner: PublicKey,
     mint: PublicKey,
     programId = TOKEN_PROGRAM_ID,
-    associatedTokenProgramId = ASSOCIATED_TOKEN_PROGRAM_ID
+    associatedTokenProgramId = ASSOCIATED_TOKEN_PROGRAM_ID,
+    idempotent = false,
```
joncinque, line 26: And lastly here, let's have a separate function.  Since the implementation is mostly the same, you can have them call into the same private function underneath

joncinque (commented):
Thanks for your contribution! Since a lot of these APIs are overcomplicated as it is, let's break out the idempotent bit, then this should be good to go

mwrites:
> Sorry for the long wait, and thanks for your contribution! Since a lot of these APIs are overcomplicated as it is, let's break out the idempotent bit, then this should be good to go

Thanks for taking a look! I have refactored the Idempotent into its own set of functions. 👏

token/js/src/instructions/associatedTokenAccount.ts
```diff
@@ -21,6 +22,57 @@ export function createAssociatedTokenAccountInstruction(
...
+        associatedToken,
+        owner,
+        mint,
+        Buffer.from(new Uint8Array([1])),
+        programId,
+        associatedTokenProgramId
+    );
+}
+
+function _createAssociatedTokenAccountInstruction(
```
joncinque, line 68: nit: we don't use the `_functionName` convention in this repo for non-exported functions, so could you rename this to something like `buildAssociatedTokenAccountInstruction`? That way it'll match the rust implementation too

joncinque (commented):
We're almost there! Just a couple of micro-nits, then we can merge this right away

joncinque:
Also, you'll need to run `prettier` to format the files properly.  You can do that with `npm run lint:fix`

mwrites:
@joncinque done!
- [x] lint
- [x] rename
- [x] Buffer.from(1)

joncinque:
CI failure is unrelated, merging on red. Thanks for your patience with this!

PR #3564, opened 2022-09-01
token-cli: infer program_id when possible

### LINKED ISSUE #3532: token-cli: program_id inference does not happen
the cli is designed so that, ideally, program_id is determined like this:
* if the user uses the `-p` flag, their choice is enforced
* otherwise, for a command that takes an account, the owner of it is used
* otherwise, for commands that take nothing (only `create-token`?), tokenkeg is used (presumably someday tokenz)
* if `-p` was provided and a differently owned mint/account is ever fetched, the program halts

unfortunately this is what happens:
* if the user uses the `-p` flag, their choice is enforced, otherwise it defaults to tokenkeg
* if you do anything with a tokenz mint/account and didnt specify it with the `-p` flag, the program halts

the logic for an updated `program_id` to flow through the program exists, but `check_owner` functions as an assert that prevents any of it from ever happening

i think what i want to do here is look for a pubkey in the cli args that isnt a flag (almost all commands have one), fetch that inside config creation, check the owner but dont parse data, and use _that_ as our default

2501babe:
first step is to normalize what our args are named (details in commit message, for the sake of our futute descendents) and then its pretty easy to handle this inside config creation

closes #3532

### DIFF
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -10,7 +10,10 @@ use solana_clap_utils::{
 use solana_cli_output::OutputFormat;
 use solana_client::nonblocking::rpc_client::RpcClient;
 use solana_remote_wallet::remote_wallet::RemoteWalletManager;
-use solana_sdk::{commitment_config::CommitmentConfig, pubkey::Pubkey, signature::Signer};
+use solana_sdk::{
+    account::Account as RawAccount, commitment_config::CommitmentConfig, pubkey::Pubkey,
+    signature::Signer,
+};
 use spl_associated_token_account::*;
 use spl_token_2022::{
     extension::StateWithExtensionsOwned,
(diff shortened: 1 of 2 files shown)


2501babe:
it turns out im not allowed to do tis but i thought it was v cute
```
        } else if !sign_only {
            value_of(matches, "token")
                .or_else(|| value_of(matches, "account"))
                .or_else(|| value_of(matches, "address"))
                .and_then(async |address| rpc_client.get_account(&address).await.ok())
                .map(|account| account.owner)
                .unwrap_or(default_program_id)
        } else {
```

2501babe:
alright going back to code reviews before i finish this but note to self that i want to write a few tests for program id inference and mismatch, and i also want to get rid of all the `check_owner` calls and replace them witha config method `get_owned_account` that does this implicitly

2501babe:
ok nevermind the line about tests, this cant be unit tested, and we dont have an integration test flow

joncinque (dismissed):
Really just tiny nits, this is good to go -- thanks!

PR #3570, opened 2022-09-06
token-2022: add calculate post and pre fee amounts

samkim-crypto:
An attempt at simplifying transfer fee calculation.

- Simplified fee calculation using ceiled division `ceil[ a / b ]` as `(a + b - 1) / b`.
- Added `calculate_post_fee_amount` and `calculate_pre_fee_amount`.

It seems like for most applications, computing (1) the net transfer amount from the gross amount or (2) the gross amount from the net amount is very useful.

@joncinque What do you think? The updated `calculate_inverse_fee` breaks a couple of the edge case tests, but it seems to maintain the precision. It also seems like `calculate_post_fee_amount` and `calculate_pre_fee_amount` can be used directly for the swap program.

### DIFF
--- a/token/program-2022/src/extension/transfer_fee/mod.rs
+++ b/token/program-2022/src/extension/transfer_fee/mod.rs
@@ -51,7 +51,7 @@ impl TransferFee {
         } else {
             let numerator = (pre_fee_amount as u128).checked_mul(transfer_fee_basis_points)?;
             let raw_fee = Self::ceil_div(numerator, ONE_IN_BASIS_POINTS)?
-                .try_into() // guaranteed to be oky
+                .try_into() // guaranteed to be okay
                 .ok()?;
 
             Some(cmp::min(raw_fee, u64::from(self.maximum_fee)))
--- a/token/program-2022/src/extension/transfer_fee/mod.rs
+++ b/token/program-2022/src/extension/transfer_fee/mod.rs
@@ -35,7 +35,10 @@ pub struct TransferFee {
     pub transfer_fee_basis_points: PodU16,
 }
 impl TransferFee {
-    /// Calculate ceiled division
+    /// Calculate ceiling-division
+    ///
+    /// Ceiling-division `ceil[ dividend / divisor ]` can be represented as a floor-division
+    /// `floor[ dividend + (divisor - 1) / divisor ]`
     fn ceil_div(dividend: u128, divisor: u128) -> Option<u128> {
         dividend
             .checked_add(divisor)?
--- a/token/program-2022/src/extension/transfer_fee/mod.rs
+++ b/token/program-2022/src/extension/transfer_fee/mod.rs
@@ -58,7 +58,7 @@ impl TransferFee {
         }
     }
 
-    /// Calculate the net transfer amount after deducting fees
+    /// Calculate the gross transfer amount after deducting fees
     pub fn calculate_post_fee_amount(&self, pre_fee_amount: u64) -> Option<u64> {
         pre_fee_amount.checked_sub(self.calculate_fee(pre_fee_amount)?)
     }
--- a/token/program-2022/src/extension/transfer_fee/mod.rs
+++ b/token/program-2022/src/extension/transfer_fee/mod.rs
@@ -314,10 +314,12 @@ pub(crate) mod test {
         };
         let maximum_fee = u64::from(transfer_fee.maximum_fee);
         // hit maximum fee
-        // assert_eq!(
-        //     maximum_fee,
-        //     transfer_fee.calculate_inverse_fee(u64::MAX).unwrap()
-        // );
+        assert_eq!(
+            maximum_fee,
+            transfer_fee
+                .calculate_inverse_fee(u64::MAX - maximum_fee)
+                .unwrap()
+        );
         // at exactly the max
         assert_eq!(
             maximum_fee,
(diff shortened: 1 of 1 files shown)


token/program-2022/src/extension/transfer_fee/mod.rs
```diff
@@ -32,63 +35,71 @@ pub struct TransferFee {
     pub transfer_fee_basis_points: PodU16,
 }
 impl TransferFee {
+    /// Calculate ceiled division
```
joncinque, line 38: nit: could you provide a little more info in the comment about this implementation?

token/program-2022/src/extension/transfer_fee/mod.rs
```diff
@@ -32,63 +35,71 @@ pub struct TransferFee {
...
+                .try_into() // guaranteed to be okay
+                .ok()?;
+
             Some(cmp::min(raw_fee, u64::from(self.maximum_fee)))
         }
     }
 
-    /// Calculate the fee that would produce the given output
-    pub fn calculate_inverse_fee(&self, post_fee_amount: u64) -> Option<u64> {
+    /// Calculate the net transfer amount after deducting fees
```
joncinque, line 61: nit, right?
```suggestion
    /// Calculate the gross transfer amount after deducting fees
```

token/program-2022/src/extension/transfer_fee/mod.rs
```diff
@@ -300,10 +311,10 @@ pub(crate) mod test {
...
         let maximum_fee = u64::from(transfer_fee.maximum_fee);
         // hit maximum fee
-        assert_eq!(
-            maximum_fee,
-            transfer_fee.calculate_inverse_fee(u64::MAX).unwrap()
-        );
+        // assert_eq!(
+        //     maximum_fee,
+        //     transfer_fee.calculate_inverse_fee(u64::MAX).unwrap()
+        // );
```
joncinque, line 317: Can you use a different value that works to hit the maximum fee here?

token/program-2022/src/extension/transfer_fee/mod.rs
```diff
@@ -32,63 +35,71 @@ pub struct TransferFee {
...
+    ///
+    /// The original transfer amount may not always exist on large net transfer amounts due to
+    /// overflow. In this case, `None` is returned.
+    /// e.g. The net fee amount is `u64::MAX` with a positive fee rate.
+    pub fn calculate_pre_fee_amount(&self, post_fee_amount: u64) -> Option<u64> {
+        let maximum_fee = u64::from(self.maximum_fee);
         let transfer_fee_basis_points = u16::from(self.transfer_fee_basis_points) as u128;
         if transfer_fee_basis_points == 0 {
+            Some(post_fee_amount)
+        } else if transfer_fee_basis_points == ONE_IN_BASIS_POINTS || post_fee_amount == 0 {
```
joncinque, line 82: If post fee amount is 0, that doesn't necessarily mean that the pre fee amount is 0. If I transfer 1 token and the fee is 1bp, that'll have a fee of 1.

token/program-2022/src/extension/transfer_fee/mod.rs
```diff
@@ -351,7 +362,7 @@ pub(crate) mod test {
             transfer_fee.calculate_inverse_fee(one).unwrap()
         );
         // zero is 1 token
-        assert_eq!(minimum_fee, transfer_fee.calculate_inverse_fee(0).unwrap());
+        // assert_eq!(minimum_fee, transfer_fee.calculate_inverse_fee(0).unwrap());
```
joncinque, line 365: This should still work too

joncinque (commented):
Great job! Mainly some nits.

I'm glad that the extra ceiling bit is totally unnecessary with your improved calc.

token/program-2022/src/extension/transfer_fee/mod.rs
samkim-crypto: Yeah, so this is something that I thought about for a bit and is related to one of the commented out tests below as well. The gross amount is not uniquely determined due to rounding. So I thought the function should naturally provide the minimum gross amount that will produce a specified net amount.

Suppose a sender wants a receiver to receive a certain number of tokens after fees. Then it could use this function to calculate how much to send in total before fees are deducted. Then it is natural to always choose the smaller amount since the sender does not want to overpay when possible.

If the post fee amount is 0, then both 0 and 1 should work. If the fee rate is 10,000, then any amount below max fee should also produce 0. In all these cases, the amount 0 is the minimum pre fee amount that will produce 0 post fee amount.

token/program-2022/src/extension/transfer_fee/mod.rs
samkim-crypto: Yep, this is related to the comment above. Assume that bp is positive. If post fee amount is 0, then `calculate_pre_fee_amount` could produce either 0 or 1, which results in fee of either 0 or 1.

Currently, `calculate_pre_fee_amount` produces pre fee of 0 on post fee 0, so this test should fail since `minimum_fee` is 1. If the current behavior of `calculate_pre_fee_amount` is fine, then `minimum_fee` should be replaced by 0 in this test.

token/program-2022/src/extension/transfer_fee/mod.rs
joncinque: We had a discussion about this offline, and you're right, we'll go with this implementation!

token/program-2022/src/extension/transfer_fee/mod.rs
joncinque: Per our discussion, let's have this be 0 then!

token/program-2022/src/extension/transfer_fee/mod.rs
```diff
@@ -302,7 +316,9 @@ pub(crate) mod test {
         // hit maximum fee
         assert_eq!(
             maximum_fee,
-            transfer_fee.calculate_inverse_fee(u64::MAX).unwrap()
+            transfer_fee
+                .calculate_inverse_fee(u64::MAX - maximum_fee)
```
joncinque, line 320: perfect choice!

joncinque (approved):
Thanks for improving this

PR #3573, opened 2022-09-06
Added prototype for managed-token program

jarry-xiao:
Spent a few hours today hacking away at this new primitive smart contract. It uses Metaplex's shank library to autogenerate the client.

This program is a wrapper on the token program to allow for controlled operations on specific token types. The intended use case is for `upstream_authority` to be a PDA of some higher level program. This way that program can have precise control over what types of token transfers are permitted (can manage whitelist for example). The frozen token accounts managed by this contract cannot be modified without signing with the `upstream_authority` key.

It works by wrapping every action with a `thaw` and `freeze`. Every token of a mint created through this contract will be frozen unless it explicitly interacts with `permissioned-token`.

Example contract - "Gift card":
Suppose there were a large online retailer that wanted to allow users to purchase goods/collect funds using the Solana blockchain as a settlement layer. Their contract could do the following:
- User deposits USDC to gift card smart contract. Contract CPI's into `permissioned-token` to mint the user gift card points in a frozen token account.
- User once again interacts with the smart contract to make a purchase. Their gift card points are burned, and the retailer can now claim USDC stored from the first step.
- If a user wants to convert back to USDC, they can burn their points and withdraw from the USDC vault (maybe the retailer charges a fee)
- Users can transfer point to other users (probably would need to be some whitelist process to manage registrations)

TODO's
- [ ] Figure out a way to manage delegations (would function like chargebacks). Seems tricky, maybe doable
- [ ] Add Token 2022 support


permissioned-token/program/build.sh
```diff
@@ -0,0 +1,6 @@
+cargo build-bpf
+shank idl -o ../sdk/idl
```
thlorenz, line 2: This step is not needed as `yarn solita` will run `solita` which will find the `.solitarc.js` and download a matching version of the shank-cli and run it to produce the IDL before generating the code from it.
This makes sure that the `shank` version you use in your program is compatible with the shank-cli version used to extract the IDL.

jarry-xiao: oh cool! learned something new thanks :)

joncinque (commented):
The program logic looks good, and the interface is easy to understand, nice job!

I'm probably missing some context, but I'm wondering: why is this done through an on-chain program rather than a client library?

Chargebacks through an escrow would be an awesome feature and definitely make sense to have a separate program, but otherwise I'm not sure I totally understand the need for this.

permissioned-token/program/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+        &system_instruction::create_account(
+            payer.key,
+            mint.key,
+            Rent::get()?.minimum_balance(space),
+            space as u64,
+            token_program.key,
+        ),
+        &[payer.clone(), mint.clone(), system_program.clone()],
+    )?;
+    let (mint_authority, _) = Pubkey::find_program_address(&[mint.key.as_ref()], program_id);
```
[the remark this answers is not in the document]
jarry-xiao: You want to make sure that the only way for a mint to occur is through this program (closed loop). Otherwise you could mint to a non-frozen account.

permissioned-token/program/src/instruction.rs
```diff
@@ -0,0 +1,195 @@
...
+    #[account(4, name = "mint")]
+    #[account(5, name = "system_program", desc = "System program")]
+    #[account(6, name = "rent", desc = "Rent sysvar")]
+    #[account(
+        7,
+        name = "associated_token_program",
+        desc = "Associated Token program"
+    )]
+    #[account(8, name = "token_program", desc = "Token program")]
+    InitializeAccount,
```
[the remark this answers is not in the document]
jarry-xiao: This is true, but they can never receive tokens because the mint can only be created or sent to accounts that are frozen!

jarry-xiao:
Yeah the freeze authority should almost always be a pda. The cool thing here is that it makes it easy to whitelist, blacklist, and have tight controls over flows and transfers

permissioned-token/program/src/instruction.rs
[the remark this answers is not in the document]
jarry-xiao: My claim is that any creation that doesn’t go through this is useless

joncinque:
This isn't a totally closed loop because the `update_authority` is not a PDA of this program.  This means that it's always possible for `update_authority` to thaw accounts outside of this program.  The freeze authority of the mint needs to be something like `find_program_address([update_authority], program_id)`.

jarry-xiao:
Yeah this is true, I guess there needs to be 1 more layer of indirection for this to be truly safe. Ultimately the security is required at the upstream program with the current impl

jarry-xiao:
@joncinque I added in the additional redirection layer. Now the tokens can never be thawed outside this program

permissioned-token/program/src/instruction.rs
```diff
@@ -0,0 +1,215 @@
...
+    #[account(3, signer, name = "upstream_authority")]
+    #[account(4, name = "freeze_authority")]
+    #[account(5, name = "token_program", desc = "Token program")]
+    MintTo { amount: u64 },
+
+    #[account(0, writable, name = "mint")]
+    #[account(1, writable, name = "account")]
+    #[account(2, signer, name = "owner")]
+    #[account(3, signer, name = "upstream_authority")]
+    #[account(4, signer, name = "freeze_authority")]
```
joncinque, line 61: looks like this one got missed:

```suggestion
    #[account(4, name = "freeze_authority")]
```

permissioned-token/program/src/instruction.rs
```diff
@@ -0,0 +1,215 @@
...
+    #[account(2, name = "mint")]
+    #[account(3, signer, name = "owner")]
+    #[account(4, signer, name = "upstream_authority")]
+    #[account(5, name = "freeze_authority")]
+    #[account(6, name = "token_program", desc = "Token program")]
+    Transfer { amount: u64 },
+
+    #[account(0, writable, name = "mint")]
+    #[account(1, writable, name = "account")]
+    #[account(2, signer, name = "mint_authority")]
```
joncinque, line 51: is this correct?
```suggestion
    #[account(2, name = "mint_authority")]
```

permissioned-token/program/src/instruction.rs
```diff
@@ -0,0 +1,215 @@
...
+    #[account(4, signer, name = "freeze_authority")]
+    #[account(5, name = "token_program", desc = "Token program")]
+    Burn { amount: u64 },
+
+    #[account(0, writable, name = "account")]
+    #[account(1, writable, name = "destination")]
+    #[account(2, name = "mint")]
+    #[account(3, signer, name = "owner")]
+    #[account(4, signer, name = "upstream_authority")]
+    #[account(5, signer, name = "freeze_authority")]
```
joncinque, line 70: Same here
```suggestion
    #[account(5, name = "freeze_authority")]
```

permissioned-token/program/src/instruction.rs
```diff
@@ -0,0 +1,215 @@
...
+
+pub fn create_initialize_account_instruction(
+    mint: &Pubkey,
+    owner: &Pubkey,
+    payer: &Pubkey,
+    upstream_authority: &Pubkey,
+) -> Result<Instruction, ProgramError> {
+    let account = get_associated_token_address(owner, mint);
+    let (freeze_authority, _) =
+        Pubkey::find_program_address(&[upstream_authority.as_ref()], &crate::id());
```
joncinque, line 102: nit: probably better to reuse `get_freeze_authority` in all of these

permissioned-token/program/src/instruction.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+pub fn create_mint_to_instruction(
+    mint: &Pubkey,
+    owner: &Pubkey,
+    upstream_authority: &Pubkey,
+    amount: u64,
+) -> Result<Instruction, ProgramError> {
+    let account = get_associated_token_address(owner, mint);
+    let (mint_authority, _) = Pubkey::find_program_address(&[mint.as_ref()], &crate::id());
```
joncinque, line 128: nit: it'd be better to refactor this into a separate function, and to put in a prefix seed like `b"mint-authority"` to totally separate the domain from the upstream authority

permissioned-token/program/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+        expected_key == &freeze_key,
+        ProgramError::InvalidInstructionData,
+        "Invalid freeze authority",
+    )?;
+    Ok(seeds)
+}
+
+#[inline]
+fn get_freeze_authority(upstream_authority: &Pubkey) -> (Pubkey, Vec<Vec<u8>>) {
+    let mut seeds = vec![upstream_authority.as_ref().to_vec()];
```
joncinque, line 49: nit: it'd be better to separate the domain for the freeze authority by adding some additional seed like `b"freeze-authority"`

permissioned-token/program/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+pub fn process_initialize_account(accounts: &[AccountInfo]) -> ProgramResult {
+    let InitializeAccount {
+        token_account,
+        owner,
+        payer,
+        upstream_authority,
+        freeze_authority,
+        mint,
+        system_program,
+        rent,
```
joncinque, line 129: You can remove the rent account, it isn't needed in the ATA program!

permissioned-token/program/tests/test.rs
```diff
@@ -0,0 +1,143 @@
...
+    let mut tx = Transaction::new_with_payer(&instructions, Some(&signers[0].pubkey()));
+    tx.partial_sign(&signers, client.get_latest_blockhash().await?);
+    let sig = tx.signatures[0];
+    client
+        .process_transaction_with_commitment(tx, CommitmentLevel::Confirmed)
+        .await?;
+    Ok(sig)
+}
+
+pub async fn airdrop(
```
joncinque, line 36: nit: how about calling this `transfer`?

permissioned-token/sdk/.solitarc.js
```diff
@@ -0,0 +1,14 @@
+const path = require('path');
+const programDir = path.join(__dirname, '..', 'program');
+const idlDir = path.join(__dirname, 'idl');
+const sdkDir = path.join(__dirname, 'src', 'generated');
+const binaryInstallDir = path.join(__dirname, '.crates');
```
joncinque, line 5: We did this for compression, and I thought it was a nice idea to avoid bloating in random directories -- how about putting the binaries in `target/solita`?

permissioned-token/program/src/lib.rs
```diff
@@ -0,0 +1,221 @@
...
+        token_program,
+    } = MintOrBurn::load(accounts)?;
+    let seeds = get_freeze_authority_seeds_checked(upstream_authority.key, freeze_authority.key)?;
+    thaw(freeze_authority, mint, token_account, token_program, &seeds)?;
+    let (mint_authority, bump) = Pubkey::find_program_address(&[mint.key.as_ref()], program_id);
+    assert_with_msg(
+        &mint_authority == owner.key,
+        spl_token::error::TokenError::OwnerMismatch,
+        "Mint authority must be a PDA of the Permissioned Token Program",
+    )?;
```
joncinque, line 188: This check confused me for awhile until I realized that `owner` is the mint authority in this context.  Can it be removed?

joncinque (commented):
This is on the right track, just a few bits before it's ready to be included

permissioned-token/program/src/lib.rs
jarry-xiao: Didn’t want to create 2 separate account context objects

permissioned-token/program/src/accounts.rs
```diff
@@ -0,0 +1,336 @@
...
+        assert_with_msg(
+            ctx.payer.is_writable,
+            ProgramError::InvalidInstructionData,
+            "Payer account must be writable (lamport balance will change)",
+        )?;
+        assert_with_msg(
+            ctx.payer.is_signer,
+            ProgramError::MissingRequiredSignature,
+            "Payer must sign for initialization",
+        )?;
```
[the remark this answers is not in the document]
jarry-xiao: Agree, but I think it makes it a lot easier to debug errors related to client side issues. The CPI error is super unintuitive to the untrained eye

permissioned-token/program/src/lib.rs
jarry-xiao: Is this a recent change? I feel like it used to be needed

permissioned-token/program/src/lib.rs
jarry-xiao: But I can alias the variable

permissioned-token/program/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
```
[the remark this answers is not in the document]
jarry-xiao: done

Cargo.toml
```diff
@@ -22,6 +22,7 @@ members = [
   "libraries/merkle-tree-reference",
   "memo/program",
   "name-service/program",
+  "permissioned-token/program",
```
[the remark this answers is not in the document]
jarry-xiao: not too sure what you mean here. can you check that this is done correctly on the most recent update?

jarry-xiao:
@joncinque lmk if you have any final thoughts

permissioned-token/program/src/accounts.rs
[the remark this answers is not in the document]
joncinque: Gotcha, makes sense

permissioned-token/program/src/lib.rs
joncinque: You can still get into weird situations that go against the spirit of the two separate PDAs.  For example, if you generate a keypair, create a mint on it, and also use that keypair as the upstream authority, then your derived mint authority and freeze authorities will be the same address.

This isn't exploitable in this case, but it's still a source of confusion.  If you want to keep things simple, I'd recommend just having one PDA that mints and freezes.

permissioned-token/program/src/lib.rs
joncinque: I think it was needed for ATA version 1.0.4 and before

managed-token/sdk/.crates/.crates.toml
```diff
@@ -0,0 +1,2 @@
+[v1]
+"shank-cli 0.0.5 (registry+https://github.com/rust-lang/crates.io-index)" = ["shank"]
```
joncinque, line 2: Does this need to be checked in?

managed-token/sdk/.crates/.crates2.json
```diff
@@ -0,0 +1 @@
+{"installs":{"shank-cli 0.0.5 (registry+https://github.com/rust-lang/crates.io-index)":{"version_req":"^0.0.5","bins":["shank"],"features":[],"all_features":false,"no_default_features":false,"profile":"release","target":"aarch64-apple-darwin","rustc":"rustc 1.60.0 (7737e0b5c 2022-04-04)\nbinary: rustc\ncommit-hash: 7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c\ncommit-date: 2022-04-04\nhost: aarch64-apple-darwin\nrelease: 1.60.0\nLLVM version: 14.0.0\n"}}}
```
joncinque, line 1: Same with this

Cargo.toml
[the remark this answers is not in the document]
joncinque: Essentially, you can remove `managed-token/program/Cargo.lock`, since you've included it in the top-level workspace.  Typically you do one or the other.

managed-token/sdk/.crates/.crates2.json
jarry-xiao: @thlorenz probably knows more about what gets autogenerated. I think this can probably get deleted, but I don't think it really matters that much?

jarry-xiao: I'll remove it thought

permissioned-token/program/src/lib.rs
jarry-xiao: I'll make the change the unify the PDAs

permissioned-token/sdk/.solitarc.js
joncinque: Could you do this one to avoid accidentally checking in the `.crates` folder in the future?

jarry-xiao: done

managed-token/sdk/.solitarc.js
```diff
@@ -2,7 +2,7 @@ const path = require("path");
 const programDir = path.join(__dirname, "..", "program");
 const idlDir = path.join(__dirname, "idl");
 const sdkDir = path.join(__dirname, "src", "generated");
-const binaryInstallDir = path.join(__dirname, ".crates");
+const binaryInstallDir = path.join(__dirname, "..", "target", "solita");
```
joncinque, line 5: Sorry, could this actually go into `target/solita` from the top-level? meaning:
```
const binaryInstallDir = path.join(__dirname, "..", "..", "target", "solita");
```

joncinque (approved):
r+ one last bit, thanks for your patience!

managed-token/sdk/package.json
```diff
@@ -0,0 +1,12 @@
+{
```
thlorenz, line 1: @jarry-xiao I just noticed (too late) that the dependencies here need some changes.

- `rustbin` + `solita` should be `devDependencies`
- deps that the generated code is using should be added as well

Extra deps needed:

- @metaplex-foundation/beet
- @solana/web3.js

Maybe we can fix this in a separate PR.

jarry-xiao: ah sad, i can potentially look into adding. Could that all be generated with shank/solita or are there issues there?

thlorenz: I've thought of generating this, but don't want to mess with the user's `package.json` file.
Specifically they might already have a `web3.js` dep in there for instance and I wouldn't want to change that and/or add a duplicate.

However if we come up with a frictionless way to do this for users I'm open to add this as a feature to solita.

PR #3576, opened 2022-09-07
Update token-2022.md

jnwng:
just a couple of updates here as i was reading through the docs!


joncinque (approved):
Thanks for the contribution!

PR #3582, opened 2022-09-08
token-2022: Add support for non transferable and immutable ownership in token client

samkim-crypto:
- Extended `ExtensionInitializationParams` for the non-transferable extension so that non-transferable extension can be easily initialized
- Tweaked `create_auxiliary_token_account_with_extension_space` in token client so that immutable owner accounts can be easily initialized
- Added basic tests for the non-transferable extension. Tests for the non-transferable confidential extension will be added in a separate PR.

### DIFF
--- a/token/program-2022-test/tests/non_transferable.rs
+++ b/token/program-2022-test/tests/non_transferable.rs
@@ -5,29 +5,19 @@ use {
     program_test::{TestContext, TokenContext},
     solana_program_test::tokio,
     solana_sdk::{
-        instruction::InstructionError, program_option::COption, pubkey::Pubkey, signature::Signer,
-        signer::keypair::Keypair, transaction::TransactionError, transport::TransportError,
+        instruction::InstructionError, signature::Signer, signer::keypair::Keypair,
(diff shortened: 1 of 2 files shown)


token/program-2022-test/tests/non_transferable.rs
```diff
@@ -0,0 +1,295 @@
...
+            TransactionError::InstructionError(
+                0,
+                InstructionError::Custom(TokenError::NonTransferable as u32)
+            )
+        )))
+    );
+}
+
+#[tokio::test]
+async fn transfer_checked_with_fee() {
```
joncinque, line 133: It's pretty funny to imagine a mint with nontransferable tokens and transfer fees, but thanks for adding the test :laughing:

joncinque (dismissed):
Looks good to me! Sorry for sleeping on this one

token/program-2022-test/tests/non_transferable.rs
samkim-crypto: Yep, you knever know! 😆

PR #3583, opened 2022-09-08
add stake pool support support for hardware wallets in set-manager

0xEdgar:
currently, `set-manager` cannot be used to change the stake pool manager to a hardware wallet, due to the use of `keypair_of`, which does not support a hardware wallet path. We instead use `get_signer` to get the signer, which allows hardware wallets to sign the transaction

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -1718,7 +1718,7 @@ fn command_withdraw_sol(
 fn command_set_manager(
     config: &Config,
     stake_pool_address: &Pubkey,
-    new_manager: &Box<dyn Signer>,
+    new_manager: Box<dyn Signer>,
     new_fee_receiver: &Option<Pubkey>,
 ) -> CommandResult {
     if !config.no_update {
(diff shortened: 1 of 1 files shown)


stake-pool/cli/src/main.rs
```diff
@@ -1718,7 +1718,7 @@ fn command_withdraw_sol(
 fn command_set_manager(
     config: &Config,
     stake_pool_address: &Pubkey,
-    new_manager: &Option<Keypair>,
+    new_manager: Box<dyn Signer>,
```
joncinque, line 1721: Can we make this an `Option<Box<dyn Signer>>`? This command allows you to set either the manager or the fee receiver, but the underlying instruction updates both of them at once, which means that if someone only updates the fee receiver, we need to fetch and set the old value.

joncinque (changes requested):
Thanks for your contribution! This is a great and needed fix, we just need to maintain the old functionality.  Let me know if you have any questions

stake-pool/cli/src/main.rs
0xEdgar: done. would have been cleaner if `get_signer` returned an option or a result type so i just wrapped the call lol

0xEdgar:
are the cargo-test-sbf tests flaky? [this test run](https://github.com/solana-labs/solana-program-library/runs/8275889366?check_suite_focus=true) seems to have failed on a test unrelated to what i changed. I've re-triggered CI so 🤞🏻

0xEdgar:
> are the cargo-test-sbf tests flaky? [this test run](https://github.com/solana-labs/solana-program-library/runs/8275889366?check_suite_focus=true) seems to have failed on a test unrelated to what i changed. I've re-triggered CI so 🤞🏻

ok it was flakey yay ✅✅✅

PR #3584, opened 2022-09-08
token-client: Change create_account to return T::Output

joncinque:
#### Problem

The token client's `create_associated_account` and `create_auxiliary_account` functions return the pubkey of the created account, which is nifty in tests, but not usable in CLIs where we want a transaction signature.

I came across this issue while writing a CLI for the token-upgrade program. One of the commands just creates a token account, but I wasn't able to display the transaction info because of the previous return type.

#### Solution

Change the output of the the create account functions to `T::Output`, and fixup all tests that use it.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -458,7 +458,7 @@ where
     }
 
     /// Create and initialize the associated account.
-    pub async fn create_associated_token_account(&self, owner: &Pubkey) -> TokenResult<Pubkey> {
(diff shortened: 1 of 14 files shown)


joncinque:
Bah sorry, I'll fixup the rest of the CI stuff tomorrow, it's late over here.

joncinque:
All right, this should be good for a review now @2501babe whenever you have a chance

2501babe (approved):

PR #3585, opened 2022-09-09
token-cli: add `spl-token display`

### LINKED ISSUE #3578: token-cli: new command to replace `account-info`
right now the only prettyprint function in the cli is `account-info`, which has a number of deficiencies:
* it only prints accounts, not mints
* its default behavior is to accept a mint and look up the client keypair ata for that mint, so adding backwards compat ui to print mints would be torturous
* it doesnt support token22 and we need a way to show extensions and all the various new authorities

so far when debugging token22 features ive just used `solana account` and looked at the hexdumps. but i am not most users

id like to add a new subcommand (tenatively `spl-token display`) that takes a mint or account (or multisig? im still not entirely sure how spl multisigs work or where they come from) address and prints whatever is there with no gymnastics, and deprecate `account-info` by making it hidden once this is better

cc @joncinque (just lmk if this sounds unreasonable otherwise im gonna do it)

2501babe:
closes #3578

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -4,7 +4,7 @@ use clap::{
 };
 use serde::Serialize;
 use solana_account_decoder::{
-    parse_token::{parse_token, TokenAccountType, UiAccountState},
+    parse_token::{get_token_account_mint, parse_token, TokenAccountType, UiAccountState},
     UiAccountData,
 };
 use solana_clap_utils::{
@@ -149,6 +149,7 @@ pub enum CommandName {
     Address,
     AccountInfo,
     MultisigInfo,
+    Display,
     Gc,
     SyncNative,
 }
@@ -1623,40 +1624,47 @@ async fn command_address(
     Ok(config.output_format.formatted_string(&cli_address))
 }
 
-async fn command_account_info(config: &Config<'_>, address: Pubkey) -> CommandResult {
-    let account = config
-        .rpc_client
-        .get_token_account(&address)
-        .await
-        .map_err(|_| format!("Could not find token account {}", address))?
-        .unwrap();
-    let mint = Pubkey::from_str(&account.mint).unwrap();
-    let owner = Pubkey::from_str(&account.owner).unwrap();
-    let program_id = config.rpc_client.get_account(&address).await?.owner;
-    let is_associated =
-        get_associated_token_address_with_program_id(&owner, &mint, &program_id) == address;
-    let cli_token_account = CliTokenAccount {
-        address: address.to_string(),
-        is_associated,
-        account,
-    };
-    Ok(config.output_format.formatted_string(&cli_token_account))
-}
-
 async fn command_display(config: &Config<'_>, address: Pubkey) -> CommandResult {
-    let account = config.get_account_checked(&address).await?;
-    let token_data = parse_token(&account.data, None);
+    let account_data = config.get_account_checked(&address).await?;
 
-    if let Ok(TokenAccountType::Multisig(multisig)) = token_data {
-        let cli_output = CliMultisig {
-            address: address.to_string(),
-            program_id: config.program_id.to_string(),
-            multisig,
-        };
-
-        Ok(config.output_format.formatted_string(&cli_output))
+    let decimals = if let Some(mint_address) = get_token_account_mint(&account_data.data) {
+        Some(config.get_mint_info(&mint_address, None).await?.decimals)
     } else {
-        panic!("unimplemented");
+        None
+    };
+
+    let token_data = parse_token(&account_data.data, decimals);
+
+    match token_data {
+        Ok(TokenAccountType::Account(account)) => {
+            let mint_address = Pubkey::from_str(&account.mint)?;
+            let owner = Pubkey::from_str(&account.owner)?;
+            let associated_address = get_associated_token_address_with_program_id(
+                &owner,
+                &mint_address,
+                &config.program_id,
+            );
+
+            let cli_output = CliTokenAccount {
+                address: address.to_string(),
+                program_id: config.program_id.to_string(),
+                is_associated: associated_address == address,
+                account,
+            };
+
+            Ok(config.output_format.formatted_string(&cli_output))
+        }
+        Ok(TokenAccountType::Mint(_mint)) => unimplemented!(),
+        Ok(TokenAccountType::Multisig(multisig)) => {
+            let cli_output = CliMultisig {
+                address: address.to_string(),
+                program_id: config.program_id.to_string(),
+                multisig,
+            };
+
+            Ok(config.output_format.formatted_string(&cli_output))
+        }
+        Err(_) => panic!("err here"),
     }
 }
 
@@ -2744,7 +2752,7 @@ fn app<'a, 'b>(
         )
         .subcommand(
             SubCommand::with_name(CommandName::MultisigInfo.into())
-                .about("Query details about and SPL Token multisig account by address (DEPRECATED: use `spl-token display`)")
+                .about("Query details of an SPL Token multisig account by address (DEPRECATED: use `spl-token display`)")
                 .setting(AppSettings::Hidden)
                 .arg(
                     Arg::with_name("address")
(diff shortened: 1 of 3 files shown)


2501babe:
spl-display
alright this is what this looks like now. i just need to impl more extensions

token/cli/src/output.rs
```diff
@@ -399,6 +481,108 @@ impl fmt::Display for CliTokenAccounts {
...
+            writeln_name_value(
+                f,
+                "    Rate authority:",
+                rate_authority.as_ref().unwrap_or(&String::new()),
+            )
+        }
+        // ExtensionType::Uninitialized is a hack to ensure a mint/acocunt is never the same length as a multisig
+        UiExtension::Uninitialized => Ok(()),
+        UiExtension::UnparseableExtension => {
+            panic!("Failed to parse extension; token data may be malformed")
```
2501babe, line 581: this seemed more reasonable than returning an error, because this is for a `Display` impl, which expects to see `std::fmt::Error`, but this isnt a format error...

token/cli/src/output.rs
```diff
@@ -399,6 +481,108 @@ impl fmt::Display for CliTokenAccounts {
...
+        }
+        UiExtension::MintCloseAuthority(UiMintCloseAuthority { close_authority }) => {
+            writeln_name_value(
+                f,
+                "  Close authority:",
+                close_authority.as_ref().unwrap_or(&String::new()),
+            )
+        }
+        UiExtension::ConfidentialTransferMint(_) => unimplemented!(),
+        UiExtension::ConfidentialTransferAccount(_) => unimplemented!(),
```
2501babe, line 543: i think im going to punt this for now because the confidential code is complex and im not sure what we want to do for like, possibly decrypting and displaying values

token/cli/src/output.rs
```diff
@@ -176,6 +187,9 @@ impl fmt::Display for CliMultisig {
 #[serde(rename_all = "camelCase")]
 pub(crate) struct CliTokenAccount {
     pub(crate) address: String,
+    pub(crate) program_id: String,
+    pub(crate) epoch: u64,
```
2501babe, line 191: i was split on whether to include `epoch` here, because we only need it for mint, but getting it for account is trivial and passing an `Option<u64>` into `display_ui_extension` feels silly

2501babe:
@joncinque here is a script you can use to see what the output looks like:
```
#!/bin/bash
set -uo pipefail

BIN="../../target/debug/spl-token"

KEG_MINT=$($BIN create-token -u l --output json | sed -n 's/^ *"address": "\([a-zA-Z0-9]\+\)",/\1/p')
Z_MINT=$($BIN create-token -u l -p TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb --enable-close --interest-rate 100 --output json | sed -n 's/^ *"address": "\([a-zA-Z0-9]\+\)",/\1/p')

$BIN create-account -u l $KEG_MINT > /dev/null
KEG_ACCT=$($BIN address -u l --verbose --token $KEG_MINT | sed -n 's/^Associated token address: \([a-zA-Z0-9]\+\)/\1/p')
$BIN approve -u l $KEG_ACCT 100 $(solana-keygen new --no-outfile --no-bip39-passphrase | sed -n 's/pubkey: \(.*\)/\1/p') > /dev/null

$BIN create-account -u l $Z_MINT > /dev/null
Z_ACCT=$($BIN address -u l --verbose --token $Z_MINT | sed -n 's/^Associated token address: \([a-zA-Z0-9]\+\)/\1/p')

KEG_MSIG=$($BIN create-multisig -u l 2 $(solana-keygen new --no-outfile --no-bip39-passphrase | sed -n 's/pubkey: \(.*\)/\1/p') $(solana-keygen new --no-outfile --no-bip39-passphrase | sed -n 's/pubkey: \(.*\)/\1/p') $(solana-keygen new --no-outfile --no-bip39-passphrase | sed -n 's/pubkey: \(.*\)/\1/p') | sed -n 's/Creating 2\/3 multisig \([a-zA-Z0-9]\+\)/\1/p')

$BIN display -u l $KEG_MINT
echo "--"

$BIN display -u l $Z_MINT 
echo "--"

$BIN display -u l $KEG_ACCT
echo "--"

$BIN display -u l $Z_ACCT 
echo "--"

$BIN display -u l $KEG_MSIG
```

it only uses the extensions that presently have cli support in master. the screenshot above is representative

token/cli/src/output.rs
joncinque: My take is to not have the `epoch` at all and just show both transfer fees and their epoch.  This way you can easily spot an upcoming rug pull, or see how it's been increased

token/cli/src/main.rs
```diff
@@ -408,7 +409,7 @@ async fn command_create_token(
     let tx_return = finish_tx(config, &res, false).await?;
     Ok(match tx_return {
         TransactionReturnData::CliSignature(cli_signature) => format_output(
-            CliMint {
+            CliCreateToken {
```
joncinque, line 412: This is much better, thanks for separating these types

token/cli/src/output.rs
```diff
@@ -176,6 +187,9 @@ impl fmt::Display for CliMultisig {
 #[serde(rename_all = "camelCase")]
 pub(crate) struct CliTokenAccount {
     pub(crate) address: String,
+    pub(crate) program_id: String,
```
joncinque, line 190: As another note, these are a breaking changes on the CLI output, which may cause issues for people who use json or json-compact and parse the data with another program.

We'll have to be careful in the release notes to specify this change

token/cli/src/output.rs
```diff
@@ -176,6 +187,9 @@ impl fmt::Display for CliMultisig {
...
+    pub(crate) epoch: u64,
+    pub(crate) decimals: Option<u8>,
```
joncinque, line 192: Unless I'm misreading something, this should never be `None` during creation, so you should be able to `unwrap()` and have this just as a `u8`

token/cli/src/output.rs
joncinque: @samkim-crypto will probably have some nice ideas :smile:

token/cli/src/output.rs
```diff
@@ -399,6 +481,107 @@ impl fmt::Display for CliTokenAccounts {
...
+            current_rate,
+            ..
+        }) => {
+            writeln!(f, "  {}", style("Interest-bearing:").bold())?;
+            writeln!(
+                f,
+                "    {} {}bps",
+                style("Interest rate:").bold(),
+                current_rate
+            )?;
```
joncinque, line 570: Take it or leave it, but it could be nice to see the average rate too

token/cli/src/output.rs
joncinque: If someone has an old version of the token-cli, and we've since added some new extensions, and they try to fetch their account, this could cause a panic, which would stink.  How about we print an error message that an unparseable extension was found, and that they likely need to update `spl-token-cli` to see it?

joncinque (commented):
Nice job! This is close to ready, just a few nits and questions

token/cli/src/output.rs
2501babe: its also created in sort.rs#60, which has less information to work with but uses almost none of it. oh which reminds me, i wanted to improve `spl-token accounts` too...

token/cli/src/output.rs
2501babe: right, showing both does make sense. i think i still want it so i can display which is active and whether the other is upcoming or historical tho

token/cli/src/output.rs
2501babe: ahh i didnt think of adding more extensions youre right, in my head this case was a just-in-case that was never hit

token/cli/src/output.rs
joncinque: Ah right, my mistake!

token/cli/src/output.rs
joncinque: Mm true... I'll leave the final choice to you

CriesofCarrots: Sorry for the fly-by, but I think it will be confusing for consumers to see `epoch` anywhere. If you need it for extension display, can you add the `#[serde(skip_serializing)]` tag so the field doesn't appear in json output?

2501babe: thank you! i forgot that these structures were dumped as json and had been using them as if they were purely internal

2501babe:
ok i think this should be good now. ill do `accounts` in a separate pr

joncinque (approved):
Looks good, and thanks for the flyby @CriesofCarrots !

token/cli/src/output.rs
samkim-crypto: Yep, the confidential part requires key derivation and decryption, so it should probably be tackled in a separate PR. I do have ideas on displaying it, so I can build on this later. Thanks!

PR #3586, opened 2022-09-09
token-cli: remove unnecessary program_id logic

2501babe:
remembered while working on something else that none of the `program_id` fetch logic does anything, because inference happens when `Config` is created and any mismatch will hard error anyway. ie, nothing in this pr changes any behavior

note that `get_mint_info` is a `Token` method rather than a `Config` method, and thus doesnt go through `get_account_checked`, but _does_ error on a `program_id` mismatch itself too

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -340,12 +340,11 @@ pub fn signers_of(
 
 fn token_client_from_config(
     config: &Config<'_>,
-    program_id: &Pubkey,
     token_pubkey: &Pubkey,
 ) -> Token<ProgramRpcClientSendTransaction> {
     let token = Token::new(
         config.program_client.clone(),
-        program_id,
+        &config.program_id,
         token_pubkey,
         config.fee_payer.clone(),
     );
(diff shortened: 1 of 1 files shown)


token/cli/src/main.rs
```diff
@@ -1583,18 +1570,16 @@ async fn command_accounts(
...
+        validate_mint(config, token).await?;
+    }
     let accounts = config
         .rpc_client
         .get_token_accounts_by_owner(
             &owner,
             match token {
                 Some(token) => TokenAccountsFilter::Mint(token),
-                None => TokenAccountsFilter::ProgramId(program_id),
+                None => TokenAccountsFilter::ProgramId(config.program_id),
```
2501babe, line 1582: i left this as-is but dislike it. in my prettyprint pr i might make this ignore `-p` and print all accounts, grouped by program

joncinque: Yes please, that would be much better for usability

joncinque (approved):
Much better, thanks for cleaning this up

PR #3587, opened 2022-09-09
token-2022: integrate memo into confidential extension

samkim-crypto:
Confidential transfers without memo do not fail even if the destination accounts are extended to require fees. This PR fixes this so that confidential extension can work with the memo extension.

### DIFF
--- a/token/program-2022-test/tests/confidential_transfer.rs
+++ b/token/program-2022-test/tests/confidential_transfer.rs
@@ -127,9 +127,10 @@ impl ConfidentialTokenAccountMeta {
     where
         T: SendTransaction,
     {
-        let token_account = token
+        let token_account_keypair = Keypair::new();
+        token
             .create_auxiliary_token_account_with_extension_space(
-                &Keypair::new(),
+                &token_account_keypair,
                 &owner.pubkey(),
                 vec![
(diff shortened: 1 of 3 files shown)


token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -121,6 +121,46 @@ impl ConfidentialTokenAccountMeta {
         }
     }
 
+    async fn new_with_memo<T>(token: &Token<T>, owner: &Keypair) -> Self
```
joncinque, line 124: nit: I thought this created something with a memo, so how about calling this `new_with_required_memo_transfers`?

joncinque (approved):
The tests and logic looks good! Just the one nit on a function name

token/program-2022-test/tests/confidential_transfer.rs
samkim-crypto: Yep, `new_with_required_memo_transfers` is much better. Thanks!

samkim-crypto:
@mvines Can you take a quick look at the changes here for sanity? Since a memo instruction must immediately precede the token instruction, I moved the proof instructions to be immediately after the token instruction rather than immediately before.

PR #3588, opened 2022-09-09
add overflow checks on release profile

samkim-crypto:
By default, overflow checks are disabled in optimized release builds. If there is an overflow in release builds, it is silenced and could potentially lead to unexpected behavior.

This PR adds `overflow-checks = true` under release profile in Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,3 +55,6 @@ exclude = [
 
 [profile.dev]
 split-debuginfo = "unpacked"
+
+[profile.release]
+overflow-checks = true


joncinque:
Were you able to determine the impact of adding this on compute units?

samkim-crypto:
Yes, I was actually going to ask you whether you had some advice on the best way to measure the difference in the compute units. It seems like `overflow-checks` is by default set to true on the `dev` and `test` profile and is set to false on `release` profile. I compared the logs for the tests with all profiles set with `overflow-checks = true` and all set to `overflow-checks = false` and the logs of compute units were identical for the tests associated with fees.

samkim-crypto:
So the reason why I was not getting any changes in the compute unit logs was that everything in the token program uses checked operations. After I changed the transfer fee calculation to use unchecked arithmetic operations and turned off overflow checks, the compute units were saved by 219 operations for instructions related to transfer fees.

In the scheme of things, it seems like overflow checks are not completely expensive. But perhaps it is best to use checked arithemtic rather than adding `overflow-checks = true` in the release profile since adding it in cargo will impact the entire SPL repo.

joncinque:
Hm yeah, seems like the best option is to follow the monorepo:

* add `--deny=clippy::integer_arithmetic` to our clippy run
* tag offending files with `#![allow(clippy::integer_arithmetic)]`
* fix things bit by bit with some "good-first-issue"s

It ended up being less than expected, so I put in #3606 to address most things

samkim-crypto:
Subsumed by [#3606](https://github.com/solana-labs/solana-program-library/pull/3606).

PR #3590, opened 2022-09-12
Update token-js with rent-less instructions

stegaBOB:
Implemented the `InitializeMint2` instruction and updated the `createAssociatedTokenAccount` instruction builder to not use rent anymore. In addition, I defaulted the createMint instruction to use the new `InitializeMint2` handler. Not sure if there's a good reason not to do that? Would love to hear any reasons if they do exist.

### DIFF
--- a/token/js/src/instructions/decode.ts
+++ b/token/js/src/instructions/decode.ts
@@ -247,10 +247,3 @@ export function isUiamountToAmountInstruction(
 // ): decoded is DecodedInitializeMultisig2Instruction {
 //     return decoded.data.instruction === TokenInstruction.InitializeMultisig2;
 // }
-
-/** TODO: docs, implement */
-// export function isInitializeMint2Instruction(
-//     decoded: DecodedInstruction
-// ): decoded is DecodedInitializeMint2Instruction {
-//     return decoded.data.instruction === TokenInstruction.InitializeMint2;
-// }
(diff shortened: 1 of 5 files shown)


token/js/src/instructions/decode.ts
```diff
@@ -120,6 +120,13 @@ export function isInitializeMintInstruction(decoded: DecodedInstruction): decode
     return decoded.data.instruction === TokenInstruction.InitializeMint;
 }
 
+/** TODO: docs */
+export function isInitializeMint2Instruction(
+    decoded: DecodedInstruction
+): decoded is DecodedInitializeMint2Instruction {
+    return decoded.data.instruction === TokenInstruction.InitializeMint2;
+}
+
```
jordansexton, line 129: nit: the order of these in all the places in this file is intentional, following https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/instructions/types.ts

stegaBOB: Gotcha! I'll fix this

stegaBOB: fixed!

stegaBOB:
Looking into the failed test now.

jordansexton:
Published in `@solana/spl-token@0.3.5`

castle055:
On createAssociatedTokenAccount(), the Associated Token program returns a NotEnoughAccountKeys error since rent is not provided in the instruction when using v0.3.5 of the SDK. I don't know what you guys have in mind with these rentless instructions, but as of now the only way to get that instruction working is by manually pushing the rent sysvar public key into the keys array before signing and sending the transaction.

I checked some of your unit tests and they don't seem to actually send the transactions, so integration with the program is not tested and thus this error doesn't happen.

jordansexton:
Thanks @castle055, would you be willing to open an issue for this? Happy to look into it and follow up.

PR #3593, opened 2022-09-12
token-upgrade: Add CLI and docs

joncinque:
#### Problem

The token-upgrade program will be added in #3436, but there's no usable client or usage information.

#### Solution

Building on #3436, add a CLI based on https://github.com/mvines/solana-cli-template along with documentation of how to use the program.

PR #3594, opened 2022-09-12
token-client: Add fmt::Display implementation to RpcClientResponse

joncinque:
#### Problem

While working on #3593, I wanted to print the result from the token client, but the Debug printer for `Signature` gives `Signature(signature_in_base_58)`, which is not great.

#### Solution

Implement `fmt::Display` on `RpcClientResponse`, and add a feature for output that can be disable if needed.

### DIFF
--- a/update-solana-dependencies.sh
+++ b/update-solana-dependencies.sh
@@ -47,4 +47,5 @@ crates=(
 set -x
 for crate in "${crates[@]}"; do
   sed -E -i'' -e "s:(${crate} = \")(=?)${old_solana_ver}\".*:\1\2${solana_ver}\":" "${tomls[@]}"
+  sed -E -i'' -e "s:(${crate} = \{ version = \")(=?)${old_solana_ver}(\".*):\1\2${solana_ver}\3:" "${tomls[@]}"
 done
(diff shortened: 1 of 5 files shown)


update-solana-dependencies.sh
```diff
@@ -47,4 +47,5 @@ crates=(
 set -x
 for crate in "${crates[@]}"; do
   sed -E -i'' -e "s:(${crate} = \")(=?)${old_solana_ver}\".*:\1\2${solana_ver}\":" "${tomls[@]}"
+  sed -E -i'' -e "s:(${crate} = \{ version = \")(=?)${old_solana_ver}(\".*):\1\2${solana_ver}\3:" "${tomls[@]}"
```
2501babe, line 50: hahaha, i wasnt fast enough to sneakily commit my driveby, i wanted to test it first
```
-  sed -E -i'' -e "s:(${crate} = \")(=?)${old_solana_ver}\".*:\1\2${solana_ver}\":" "${tomls[@]}"
+  sed -E -i'' -e "s:(${crate} = )(\{ version = ){0,1}(\")(=?)${old_solana_ver}\".*:\1\2\3\4${solana_ver}\":" "${tomls[@]}"
```

joncinque: Hehe, thanks! looks like we pretty much got to the same place, right?

2501babe (approved):

update-solana-dependencies.sh
2501babe: ya its equivalent

PR #3595, opened 2022-09-12
token-wrap: Add interface description for wrapping program between two mints and/or token programs

joncinque:
#### Problem

Token-2022 adoption will take a long time since many different systems need to make invasive changes.

#### Proposed changes

Add an SPL token wrapping program which allows for easy interoperability of tokens, in case some systems never upgrade.  It's possible to wrap tokens in either direction, token -> token-2022 or token-2022 -> token, and then use them with any existing ecosystem.

The only piece of state required is a backpointer to the unwrapped mint address, which allows clients to figure out if their token is wrapped, and then how to unwrap it.

The functionality for unwrap will look very similar to the token-upgrade program #3436, since you're burning on one side and transferring from another.  `Wrap` will be the opposite: transfer then mint.

Note that since the wrapped mint also uses the wrapped token program as a seed, this program can work with more than two token programs, in the hope that the ecosystem gets there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7067,6 +7067,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "spl-token-wrap"
+version = "0.1.0"
+dependencies = [
+ "bytemuck",
+ "num_enum 0.5.11",
+ "solana-program",
+ "spl-associated-token-account 2.0.0",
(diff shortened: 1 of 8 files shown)


joncinque:
We've got some people interested in hacking on this, so going to merge the base instruction set. We can always revisit later!

PR #3600, opened 2022-09-13
account-compression: Rename packages, prep for release

joncinque:
#### Problem

People want the account compression crates, but they're not released yet.

#### Summary of changes

Get the crates ready for publishing.  As part of that, we need to make everything consistent with the rest of the repo:

* add `spl-` prefix to JS package
* rename `spl_compression` to `spl_account_compression`
* rename `wrapper` to `spl-noop`
* add version specifiers to local dependencies in Cargo.toml files so that crates.io versions can be used during publish

I ran through all of the tests in the READMEs, and everything worked out fine except for the steps at: https://github.com/solana-labs/solana-program-library/tree/master/account-compression#testing to `yarn link ...` and `yarn test`.


joncinque:
Merging this despite the audit failure, which is addressed in  #3601

PR #3606, opened 2022-09-14
clippy: Deny integer arithmetic, add allows where needed

joncinque:
#### Problem

Per #3588, adding overflow checks might not be the best idea due to performance and potentially dangerous panics, but we should avoid doing integer math wherever possible.

#### Solution

Following the example of the monorepo add `--deny=clippy::integer_arithmetic` to the clippy run, and add the proper `#[allow(clippy::integer_arithmetic)]` at the top of libraries and tools that currently break this.  This way, we can remove those allows little by little.

While doing this, I noticed that we weren't actually running clippy on most program tests because they're behind the `test-sbf` feature, so now we'll clippy the program tests too.

And then I made a bunch of little fixes where it seemed easy.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -77,7 +77,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -Zunstable-options --workspace --all-targets -- --deny=warnings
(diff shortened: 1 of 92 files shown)


samkim-crypto (approved):
Yep, I agree that this is a better solution than enabling `overflow-checks`. All the changes look good to me. We can start from here and remove the allows incrementally. It is also nice that clippy works on tests now!

PR #3607, opened 2022-09-15
token-cli: support immutable owner extension

### LINKED ISSUE #3405: token-cli: Support the `ImmutableOwner` extension
#### Problem

Token-2022 supports fixed owners for token accounts using the `ImmutableOwner` extension, but the CLI doesn't support this feature.

#### Solution

Support the extension by adding a flag `--immutable` on `create-account` and `wrap` to tack on the immutable owner initialization -- mind that there's enough space allocated in the account first!

2501babe:
closes #3405

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -499,77 +499,43 @@ async fn command_set_interest_rate(
 
 async fn command_create_account(
     config: &Config<'_>,
-    token: Pubkey,
+    token_pubkey: Pubkey,
     owner: Pubkey,
     maybe_account: Option<Pubkey>,
     bulk_signers: Vec<Arc<dyn Signer>>,
 ) -> CommandResult {
-    let minimum_balance_for_rent_exemption = if !config.sign_only {
-        config
-            .program_client
-            .get_minimum_balance_for_rent_exemption(Account::LEN)
-            .await?
+    let token = token_client_from_config(config, &token_pubkey);
+
+    let (account, associated) = if let Some(account) = maybe_account {
+        (account, false)
     } else {
-        0
+        (token.get_associated_token_address(&owner), true)
     };
 
-    let mint_info = config.get_mint_info(&token, None).await?;
-    let (account, system_account_ok, instructions) = if let Some(account) = maybe_account {
-        println_display(config, format!("Creating account {}", account));
-        (
-            account,
-            false,
-            vec![
-                system_instruction::create_account(
-                    &config.fee_payer.pubkey(),
-                    &account,
-                    minimum_balance_for_rent_exemption,
-                    Account::LEN as u64,
-                    &mint_info.program_id,
-                ),
-                initialize_account(&mint_info.program_id, &account, &token, &owner)?,
-            ],
-        )
-    } else {
-        let account =
-            get_associated_token_address_with_program_id(&owner, &token, &mint_info.program_id);
-        println_display(config, format!("Creating account {}", account));
-        (
-            account,
-            true,
-            vec![create_associated_token_account(
-                &config.fee_payer.pubkey(),
-                &owner,
-                &token,
-                &mint_info.program_id,
-            )],
-        )
-    };
+    println_display(config, format!("Creating account {}", account));
 
     if !config.sign_only {
-        if let Some(account_data) = config
-            .rpc_client
-            .get_account_with_commitment(&account, config.rpc_client.commitment())
-            .await?
-            .value
-        {
-            if !(account_data.owner == system_program::id() && system_account_ok) {
+        if let Some(account_data) = config.program_client.get_account(account).await? {
+            if account_data.owner != system_program::id() || !associated {
                 return Err(format!("Error: Account already exists: {}", account).into());
             }
         }
     }
 
-    let tx_return = handle_tx(
-        &CliSignerInfo {
-            signers: bulk_signers,
-        },
-        config,
-        false,
-        minimum_balance_for_rent_exemption,
-        instructions,
-    )
-    .await?;
+    let res = if associated {
+        token.create_associated_token_account(&owner).await
+    } else {
+        let signer = bulk_signers
+            .iter()
+            .find(|signer| signer.pubkey() == account)
+            .unwrap_or_else(|| panic!("No signer provided for account {}", account));
 
+        token
+            .create_auxiliary_token_account_with_extension_space(&**signer, &owner, vec![])
+            .await
+    }?;
+
+    let tx_return = finish_tx(config, &res, false).await?;
     Ok(match tx_return {
         TransactionReturnData::CliSignature(signature) => {
             config.output_format.formatted_string(&signature)
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -363,6 +363,22 @@ fn token_client_from_config(
     }
 }
 
+fn native_token_client_from_config(config: &Config<'_>) -> Token<ProgramRpcClientSendTransaction> {
+    let token = Token::new_native(
+        config.program_client.clone(),
+        &config.program_id,
+        config.fee_payer.clone(),
+    );
+
+    if let (Some(nonce_account), Some(nonce_authority)) =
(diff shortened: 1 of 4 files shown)


token/client/src/token.rs
```diff
@@ -894,6 +916,88 @@ where
         .await
     }
 
+    /// Wrap lamports into native account
+    pub async fn wrap<S: Signers>(
```
2501babe, line 920: i liked the idea of making this immutable by default and exposing a second "you asked for it, you got it" function, in favor of taking a bool or making mutable owner the default, but id be open to either of those interfaces if you prefer one

token/client/src/token.rs
```diff
@@ -894,6 +916,88 @@ where
...
+        owner: &Pubkey,
+        lamports: u64,
+        signing_keypairs: &S,
+    ) -> TokenResult<T::Output> {
+        // mutable owner for Tokenkeg, immutable otherwise
+        let immutable_owner = self.program_id != spl_token::id();
+        let instructions = self.wrap_ixs(account, owner, lamports, immutable_owner)?;
+
+        self.process_ixs(&instructions, signing_keypairs).await
+    }
```
2501babe, line 932: i liked the idea of making wrapped accounts immutable by default and exposing a second "you asked for it, you got it" function, in favor of taking a bool for that, or making mutable owner the default, but id be open to either of those interfaces if you prefer one

token/cli/src/main.rs
```diff
@@ -1861,25 +1831,20 @@ async fn command_gc(
     Ok(result)
 }
 
-async fn command_sync_native(
-    native_account_address: Pubkey,
-    bulk_signers: Vec<Arc<dyn Signer>>,
-    config: &Config<'_>,
-) -> CommandResult {
+async fn command_sync_native(config: &Config<'_>, native_account_address: Pubkey) -> CommandResult {
```
2501babe, line 1834: note the removal of `bulk_signers`. im 95% sure from looking at `processor.rs` that `sync_native` is permissionless, so signing only with the payer is acceptable in all cases, but i havent tested this by hand

token/cli/src/main.rs
```diff
@@ -1204,90 +1205,57 @@ async fn command_thaw(
...
-            {
-                if account_data.owner != system_program::id() {
-                    return Err(format!("Error: Account already exists: {}", account).into());
-                }
+    println_display(config, format!("Wrapping {} SOL into {}", sol, account));
+
+    if !config.sign_only {
+        if let Ok(account_data) = config.get_account_checked(&account).await {
+            if account_data.owner != system_program::id() {
+                return Err(format!("Error: Account already exists: {}", account).into());
```
joncinque, line 1232: I'm a bit confused... `get_account_checked` makes sure that the owner of the account is `config.program_id`, so would we ever trigger the error? Seems like you might prefer to do a raw `get_account` rather than `get_account_checked`

token/client/src/token.rs
```diff
@@ -894,6 +916,88 @@ where
...
+
+    /// Wrap lamports into a native account that can always have its ownership changed
+    pub async fn wrap_transferable<S: Signers>(
```
joncinque, line 935: nit: how about full clarity with the name and say `wrap_with_mutable_ownership`? `transferable` is ambiguous because of the non-transferable token extension

token/cli/src/main.rs
```diff
@@ -499,77 +515,62 @@ async fn command_set_interest_rate(
...
-                    &account,
-                    minimum_balance_for_rent_exemption,
-                    Account::LEN as u64,
-                    &mint_info.program_id,
-                ),
-                initialize_account(&mint_info.program_id, &account, &token, &owner)?,
-            ],
-        )
+    let (account, is_associated) = if let Some(account) = maybe_account {
+        (account, false)
```
joncinque, line 528: The token client checks to see if it's an associated address when wrapping, how about doing the same thing here?  I only bring this up because you did good work elsewhere :smile:

token/cli/src/main.rs
```diff
@@ -499,77 +515,62 @@ async fn command_set_interest_rate(
...
+        if config.program_id == spl_token::id() {
+            return Err(format!(
+                "Specified --immutable, but token program {} does not support the extension",
+                config.program_id
+            )
+            .into());
+        } else if is_associated {
+            println_display(
+                config,
+                "Note: --immutable specified, but Token-2022 ATAs are always immutable".to_string(),
```
joncinque, line 553: take it or leave it:
```suggestion
                "Note: --immutable specified, but Token-2022 ATAs are always immutable, ignoring".to_string(),
```

joncinque (commented):
Looks really good, just a few questions / small points

token/client/src/token.rs
```diff
@@ -217,6 +217,16 @@ impl<T> fmt::Debug for Token<T> {
     }
 }
 
+fn native_mint(program_id: &Pubkey) -> Pubkey {
+    if program_id == &spl_token_2022::id() {
+        spl_token_2022::native_mint::id()
+    } else if program_id == &spl_token::id() {
+        spl_token::native_mint::id()
+    } else {
+        panic!("Unrecognized token program id: {}", program_id);
```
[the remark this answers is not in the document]
2501babe: hm, maybe i should just catch that at the cli with a predicate. the thing is the new `native_mint` is only used in `new_native` (a constructor) and `is_native` (an object predicate) neither of which should return `Result` imo

token/cli/src/main.rs
```diff
@@ -499,77 +515,66 @@ async fn command_set_interest_rate(
...
-                system_instruction::create_account(
-                    &config.fee_payer.pubkey(),
-                    &account,
-                    minimum_balance_for_rent_exemption,
-                    Account::LEN as u64,
-                    &mint_info.program_id,
-                ),
-                initialize_account(&mint_info.program_id, &account, &token, &owner)?,
-            ],
+            token.get_associated_token_address(&owner) == account,
```
2501babe, line 530: >The token client checks to see if it's an associated address when wrapping, how about doing the same thing here?
this was what you meant by this comment, right? i wasnt entirely sure

2501babe: this is the full block, it looks weird in the diff
```
    let (account, is_associated) = if let Some(account) = maybe_account {
        (       
            account,
            token.get_associated_token_address(&owner) == account,
        )       
    } else {
        (token.get_associated_token_address(&owner), true)
    };
```

joncinque: Yep, exactly!

token/client/src/token.rs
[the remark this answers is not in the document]
joncinque: You're right -- I like catching it at the CLI level

token/cli/src/main.rs
```diff
@@ -3633,6 +3629,16 @@ mod tests {
...
 
+    fn native_mint(program_id: &Pubkey) -> Result<Pubkey, Error> {
+        if program_id == &spl_token_2022::id() {
+            Ok(spl_token_2022::native_mint::id())
+        } else if program_id == &spl_token::id() {
+            Ok(spl_token::native_mint::id())
+        } else {
+            Err(format!("Error: unknown token program id {}", program_id).into())
+        }
+    }
```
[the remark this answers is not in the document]
joncinque: You took the better approach, thanks!

joncinque (dismissed):
Good to go, up to you if you want to do that additional check when doing `--program-id <my_bogo_token_id> wrap`.

Edit: actually, we can make it work! The native mint is a PDA of the program id for token-2022, so we can use that derivation with whatever program id is provided and never have to error: https://github.com/solana-labs/solana-program-library/blob/54695b233484722458b18c0e26ebb8334f98422c/token/program-2022/src/native_mint.rs#L10

2501babe:
i think i prefer constraining `-p` because its likely to protect against many other silly edge cases. the pda thing is clever tho

2501babe:
hmm theres some networking issue with ci or something, im gonna sleep and see if it magically resolves itself in the morning

PR #3610, opened 2022-09-15
ci: Add clap-v3-utils to patch and update scripts

joncinque:
#### Problem

We're not including the `solana-clap-v3-utils` crate in the patch / update scripts, causing the downstream build project to fail in the monorepo.

#### Solution

Update the script.


joncinque:
Yolo'ing the change since it's not actually used in CI anywhere in this repo

CriesofCarrots:
Thank you!

PR #3627, opened 2022-09-21
Governance: Implement Absolute max voter weight source

SebastianBor:
#### Summary 

Implement `Absolute` option for `community_mint_max_voter_weight_source` to support static values not related to the governance token supply

### DIFF
--- a/governance/program/src/state/proposal.rs
+++ b/governance/program/src/state/proposal.rs
@@ -432,26 +432,24 @@ impl ProposalV2 {
             return Ok(governing_token_mint_supply);
         }
 
-        match realm_data.config.community_mint_max_vote_weight_source {
+        let max_voter_weight = match realm_data.config.community_mint_max_vote_weight_source {
             MintMaxVoteWeightSource::SupplyFraction(fraction) => {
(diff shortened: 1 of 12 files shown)


governance/program/src/state/proposal.rs
```diff
@@ -1834,6 +1832,57 @@ mod test {
         assert_eq!(proposal.max_vote_weight, Some(100));
     }
 
+    #[test]
+    fn test_try_tip_vote_with_reduced_absolute_community_mint_max_vote_weight() {
+        // Arrange
+        let mut proposal = create_test_proposal();
+
+        proposal.options[0].vote_weight = 60;
```
joncinque, line 1840: As another test, can you try vote with more than the absolute voter weight?  For example, just like this one, but instead have `vote_weight = 120`.  Just to make sure that overflows are correctly handled

joncinque (commented):
The logic looks good, but I'd like to see another test for the "overflow" case of a low absolute max and a lot of tokens used.

governance/program/src/state/proposal.rs
SebastianBor: Done

joncinque (approved):
Perfect, thanks!

PR #3629, opened 2022-09-21
token-cli: finish client conversion

### LINKED ISSUE #3482: token-cli: refactor more to go through token client

2501babe:
this will be the last pr for converting the token cli to use the token client for everything. we will convert:
* ~transfer~
* ~create multisig~
* ~enable/disable transfer memo~
* ~set interest rate~
* ~gc~

closes #3482

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -28,14 +28,11 @@ use solana_cli_output::{
 use solana_client::rpc_request::TokenAccountsFilter;
 use solana_remote_wallet::remote_wallet::RemoteWalletManager;
 use solana_sdk::{
-    instruction::Instruction,
-    message::Message,
     native_token::*,
     program_option::COption,
     pubkey::Pubkey,
     signature::{Keypair, Signer},
     system_program,
-    transaction::Transaction,
 };
 use spl_associated_token_account::get_associated_token_address_with_program_id;
 use spl_token_2022::{
@@ -65,25 +62,6 @@ use sort::{is_supported_program, sort_and_parse_token_accounts};
 mod bench;
 use bench::*;
 use spl_token_2022::generic_token_account::GenericTokenAccount;
-
-struct CliSignerInfo {
-    pub signers: Vec<Arc<dyn Signer>>,
-}
-
-impl CliSignerInfo {
-    pub fn signers_for_message(&self, message: &Message) -> Vec<&dyn Signer> {
-        self.signers
-            .iter()
-            .filter_map(|k| {
-                if message.signer_keys().contains(&&k.pubkey()) {
-                    Some(k.as_ref())
-                } else {
-                    None
-                }
-            })
-            .collect()
-    }
-}
 
 pub const OWNER_ADDRESS_ARG: ArgConstant<'static> = ArgConstant {
     name: "owner",
@@ -3416,63 +3394,6 @@ where
 enum TransactionReturnData {
     CliSignature(CliSignature),
     CliSignOnlyData(CliSignOnlyData),
-}
-
-// XXX this goes away once everything is converted to token client
-async fn handle_tx<'a>(
-    signer_info: &CliSignerInfo,
-    config: &Config<'a>,
-    no_wait: bool,
-    minimum_balance_for_rent_exemption: u64,
-    instructions: Vec<Instruction>,
-) -> Result<TransactionReturnData, Error> {
-    let fee_payer_pubkey = config.fee_payer.pubkey();
-    let fee_payer = Some(&fee_payer_pubkey);
-
-    let recent_blockhash = config.program_client.get_latest_blockhash().await?;
-    let message = if let Some(nonce_account) = config.nonce_account.as_ref() {
-        let mut message = Message::new_with_nonce(
-            instructions,
-            fee_payer,
-            nonce_account,
-            config.nonce_authority.as_ref().unwrap(),
-        );
-        message.recent_blockhash = recent_blockhash;
-        message
-    } else {
-        Message::new_with_blockhash(&instructions, fee_payer, &recent_blockhash)
-    };
-
-    if !config.sign_only {
-        let fee = config.rpc_client.get_fee_for_message(&message).await?;
-        check_fee_payer_balance(config, minimum_balance_for_rent_exemption + fee).await?;
-    }
-
-    let signers = signer_info.signers_for_message(&message);
-    let mut transaction = Transaction::new_unsigned(message);
-
-    if config.sign_only {
-        transaction.try_partial_sign(&signers, recent_blockhash)?;
-        Ok(TransactionReturnData::CliSignOnlyData(return_signers_data(
-            &transaction,
-            &ReturnSignersConfig {
-                dump_transaction_message: config.dump_transaction_message,
-            },
-        )))
-    } else {
-        transaction.try_sign(&signers, recent_blockhash)?;
-        let signature = if no_wait {
-            config.rpc_client.send_transaction(&transaction).await?
-        } else {
-            config
-                .rpc_client
-                .send_and_confirm_transaction_with_spinner(&transaction)
-                .await?
-        };
-        Ok(TransactionReturnData::CliSignature(CliSignature {
-            signature: signature.to_string(),
-        }))
-    }
 }
 
 async fn finish_tx<'a>(
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1676,18 +1676,16 @@ async fn command_gc(
                 continue;
             }
 
-            // Sanity check!
-            // we shouldn't ever be here, but if we are here, abort!
-            if is_associated && amount > 0 {
-                panic!("gc should NEVER attempt to close a nonempty ata");
-            }
-
-            if close_authority == owner {
-                let res = if is_associated || amount == 0 {
+            // this logic is quite fiendish, but its more readable this way than if/else
+            let maybe_res = match (close_authority == owner, is_associated, amount == 0) {
+                // owner authority, associated or auxiliary, empty -> close
+                (true, _, true) => Some(
                     token
                         .close_account(&address, &owner, &owner, &bulk_signers)
-                        .await
-                } else {
+                        .await,
+                ),
+                // owner authority, auxiliary, nonempty -> empty and close
+                (true, false, false) => Some(
                     token
                         .empty_and_close_auxiliary_account(
                             &address,
@@ -1696,10 +1694,40 @@ async fn command_gc(
                             decimals,
                             &bulk_signers,
                         )
-                        .await
-                }?;
+                        .await,
+                ),
+                // separate authority, auxiliary, nonempty -> transfer
+                (false, false, false) => Some(
+                    token
+                        .transfer(
+                            &address,
+                            &associated_token_account,
+                            &owner,
+                            amount,
+                            Some(decimals),
+                            Some(owner),
+                            &bulk_signers,
+                        )
+                        .await,
+                ),
+                // separate authority, associated or auxiliary, empty -> print warning
+                (false, _, true) => {
+                    println_display(
+                        config,
+                        format!(
+                            "Note: skipping {} due to separate close authority {}; \
+                             revoke authority and rerun gc, or rerun gc with --owner",
+                            address, close_authority
+                        ),
+                    );
+                    None
+                }
+                // anything else, including a nonempty associated account -> unreachable
+                (_, _, _) => unreachable!(),
+            };
 
-                let tx_return = finish_tx(config, &res, false).await?;
+            if let Some(res) = maybe_res {
+                let tx_return = finish_tx(config, &res?, false).await?;
 
                 results.push(match tx_return {
                     TransactionReturnData::CliSignature(signature) => {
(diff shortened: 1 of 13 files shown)


token/cli/src/main.rs
```diff
@@ -1896,45 +1886,28 @@ async fn command_required_transfer_memos(
...
-                &existing_extensions,
-            )?);
+            token
+                .reallocate(
+                    &token_account_address,
+                    &owner,
+                    &[ExtensionType::MemoTransfer],
+                    &bulk_signers,
+                )
+                .await?;
```
2501babe, line 1896: im undecided about this. theres a few approaches:
* reallocate as its own transaction, as i do here. this happens never more than once per account, so it feels fine
* always include a realloc instruction in the enable/disable memos methods in the token client. this is... tolerable... because process_reallocate bails early if its unnecessary, but i dont really like it
* fetch the account in the enable/disable memos methods in the token client. this is the most reasonable alternative, tho i dont strongly prefer it, because a) we cant do this in sign-only mode and the client shouldnt gate its modes, b) very few client methods perform fetches, and usually it will be unnecessary
* fuck up the interface, either by passing a realloc bool or adding a second method or something. this is dumb i dont like it

token/cli/src/main.rs
```diff
@@ -1712,127 +1666,78 @@ async fn command_gc(
...
-                close_empty_associated_accounts,
-                total_balance > 0,
-            ) {
-                (true, _, true) => continue, // don't ever close associated token account with amount
-                (true, false, _) => continue, // don't close associated token account if close_empty_associated_accounts isn't set
-                (true, true, false) => println_display(
-                    config,
-                    format!("Closing Account {}", associated_token_account),
-                ),
-                _ => {}
```
2501babe, line 1767: this is painfully confusing, my series of if statements is more verbose but i think easier to read

token/cli/src/main.rs
```diff
@@ -1433,28 +1403,21 @@ async fn command_close(
     recipient: Pubkey,
     bulk_signers: BulkSigners,
 ) -> CommandResult {
-    let mint_pubkey = if !config.sign_only {
```
2501babe, line 1436: this is a breaking change but i think its a merited one given how much nicer it makes the client function for close_account. tbh tho the more i interact with sign_only the more i want to restrict it to like, mint burn transfer approve revoke. something like that

token/cli/src/main.rs
```diff
@@ -964,80 +917,54 @@ async fn command_transfer(
...
             transfer_balance,
-            mint_info.decimals,
-        )?);
-    }
-    if let Some(text) = memo {
-        instructions.push(spl_memo::build_memo(
-            text.as_bytes(),
-            &[&config.fee_payer.pubkey()],
-        ));
-    }
```
2501babe, line 1030: the old version pushed the memo _after_ the transfer instruction, but the client (and, consequently, my version) pushes it _before_. based on this in the token22 processor:
```
        if memo_required(&destination_account) {
            check_previous_sibling_instruction_is_memo()?;
        }
```

the previous implementation was in error

token/cli/src/main.rs
```diff
@@ -819,7 +764,7 @@ async fn validate_mint(config: &Config<'_>, token: Pubkey) -> Result<Pubkey, Err
 #[allow(clippy::too_many_arguments)]
 async fn command_transfer(
```
2501babe, line 765: i commented `comman_transfer` heavily because it was quite a bit of work to follow exactly what all its branches do, if its too much i can remove them tho

token/cli/src/main.rs
```diff
@@ -1712,127 +1666,78 @@ async fn command_gc(
...
-                    &token,
-                    &associated_token_account,
-                    &owner,
-                    &config.multisigner_pubkeys,
-                    amount,
-                    decimals,
-                )?);
-            }
-            // Close the account if config.owner is able to
-            if close_authority == owner {
```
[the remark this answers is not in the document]
2501babe: neeeevermind, _im_ the one who misunderstood. i was reading the implementation for something unrelated and realized that the token program actually refused the owner as an authority if close authority is set (which strikes me as a bit strange honestly, since the default state of None _doesnt_ mean there is no authority)

2501babe:
the most recent push fixes the comment thread i marked resolved re: close authority in gc. unfortunately the behavior is somewhat more restricted than before. previously, it would attempt to transfer tokens out of accounts with separate close authority, but leave them open. now, it leaves them untouched

i drafted the logic for this and it made me somewhat queasy, so rather than wandering into the weeds and implementing it, ill wait for your review
```
            match (close_authority == owner, is_associated, amount == 0) { 
                // owner authority, associated or auxiliary, empty -> close
                // owner authority, auxiliary, nonempty -> empty and close
                // separate authority, auxiliary, nonempty -> transfer
                // separate authority, associated or auxiliary, empty -> print warning
                // anything else, including a nonempty associated account -> unreachable
            }
```

because, see, i _could_ do this. but what i might like _better_ is to have it revoke the close authority and close the damn account. personally i find it very strange that setting close authority makes the owner unable to close the account, because having no close authority doesnt make the account uncloseable. but, on the other hand, this makes the logic even more complicated. and thats not even getting into the thought i had to gate authority revocation behind another cli flag...

anyway, the options are:
* leave gc as-is, ignore noncloseable accounts entirely
* impl the logic in the code block, replicate the previous transfer behavior
* remove close authority and close the account, either unconditionally or behind a flag, either as its own transaction or by modifying the client more

i think i want to do 2 and let the user mess around with close authority on their own time, i dont think its that commonly used a feature

2501babe:
nevermind, writing the previous comment was enough to convince me i was right, so i implemented 2. i hope when youre back this saga is more amusing than confusing

token/client/src/token.rs
```diff
@@ -672,20 +704,37 @@ where
...
+            if *destination != self.get_associated_token_address(&recipient) {
+                return Err(TokenError::AccountInvalidAccount);
+            }
+
+            instructions.push(create_associated_token_account_idempotent(
+                &self.payer.pubkey(),
+                &recipient,
+                &self.pubkey,
+                &self.program_id,
+            ));
```
2501babe, line 733: gah, while digging through token project issues, i found something that indicated i might not be allowed to do this, because the version of the associated program on mainnet-beta lacks this instruction?

the rationale for the idempotent version here is that transfer is one of the special few functions that `must` support offline mode, so this makes it more likely to "do the right thing." eg, a user can specify they want to fund the recipient just in case, and the transaction will still work if the account exists

i also create transactions in the new version of gc that depend on this behavior because its convenient to not have to track associated account existence perfectly, but i could work around that if necessary

2501babe:
re: close authority once again, today i learned an account owner cant even remove close authority from it! this feels wrong. nevertheless it means 3 was impossible anyway so 2 is the correct flow

token/cli/src/main.rs
joncinque: It's a tough one... I think you chose the best approach

joncinque (commented):
Looks great! Most of my comments are pretty small. Really glad to see this whole transition done

token/cli/src/main.rs
2501babe: aaa sorry, i forgot to revert this when i reverted my (as it turned out misguided) `close_account` changes

token/cli/src/main.rs
```diff
@@ -834,33 +779,52 @@ async fn command_transfer(
...
+            mint_info.decimals
+        )
+        .into());
+    }
+
+    // decimals determines whether transfer_checked is used or not
+    let decimals = if use_unchecked_instruction {
+        None
+    } else if mint_decimals.is_some() {
+        mint_decimals
```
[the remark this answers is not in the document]
2501babe: the branch above is "if in online mode and decimals were provided, ensure it is correct." in offline mode, it defaults to 9 and could be incorrect. this three-way branch does unchecked if told to, uses the provided decimals if provided (online: verified correct but optional, offline: assumed correct but mandatory), and uses the mint decimals otherwise (online: correct by definition, offline: always 9, but this branch is never hit)

the behavior would be identical but maybe it would be clearer if i branched on `sign_only` instead

token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -154,7 +154,7 @@ impl ConfidentialTokenAccountMeta {
             .unwrap();
 
         token
-            .enable_required_transfer_memos(&token_account, owner)
+            .enable_required_transfer_memos(&token_account, &owner.pubkey(), &vec![owner])
```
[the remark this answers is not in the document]
2501babe: gah, i started doing this reflexively after hitting errors of this form in other places:
```
error[E0277]: the trait bound `[Arc<dyn solana_sdk::signature::Signer>; 1]: solana_sdk::signers::Signers` is not satisfied
   --> token/client/src/token.rs:402:31
    |
402 |             .try_partial_sign(&[self.payer.clone()], latest_blockhash)
    |              ---------------- ^^^^^^^^^^^^^^^^^^^^^ the trait `solana_sdk::signers::Signers` is not implemented for `[Arc<dyn solana_sdk::signature::Signer>; 1]`
    |              |
    |              required by a bound introduced by this call
    |
```
but it turns out by and large the compiler accepts the array without complaint in most other contexts. ill revert this change here, and open a pr later on top of #3669 that fixes all the other places i did this...

token/cli/src/main.rs
```diff
@@ -1732,107 +1650,98 @@ async fn command_gc(
...
-                config.output_format.formatted_string(&signature)
-            }
-            TransactionReturnData::CliSignOnlyData(sign_only_data) => {
-                config.output_format.formatted_string(&sign_only_data)
-            }
-        };
-        result += "\n";
-    }
-    Ok(result)
+    Ok(results.join(""))
```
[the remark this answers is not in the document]
2501babe: theres already two newlines at the end of each `formatted_string()` output, so this just makes it not three
```
./gc_test.sh
Fetching token accounts associated with program TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb
Processing token: 8eXp7DA7fVCoFLyLLHh43vRVwL9oNZ2tBKVUqRwr4yyB
Processing token: D7XK3136qJ6LcZwSqmWRxUo8Ni8MSWwT4m9sa3RiJdMk
Processing token: zGYxHm8gL9fwKprzxfFKiCT2yg1RWautXjULHSRyuta

Signature: 4LsRppqgQR77imqZH8Ares143R9g6NxbRdFdMtLonGhAc3qTAeMPcXbjFtzWMSiLUSC5QCQrEF6VuJnG9Ra91Xth

Signature: 2KZJDUw22cz8Zn1u9TgVx8NRMExhmP41SjGshwnLGcusKTWeYeEHZAbyUnpEXr3XJfnk5KbdzeVWrTdKyfaZ3ee1

Signature: 2MjCKAsaBfXm7cdty1ewyjwbQWXE145bT8AzCsdagVwbi3YSQHkstBphZfxuPECvyaDbvJPeoEaroqPiHUvfBW64

Signature: 5gatgFyApEY2MhM9KejyK269sXGL3q4iWtQBwSpqEiuZUs8A5pytnahC6b4jBPBkX6FMGWmUZVhgRHvJGptiZ7Lu

```

token/cli/src/main.rs
```diff
@@ -834,33 +779,52 @@ async fn command_transfer(
...
-    config.check_account(&sender, Some(token)).await?;
+
+    // in any sign_only block, we can safely unwrap this
+    let maybe_sender_state = match token.get_account_info(&sender).await {
+        Ok(a) => Some(a),
+        Err(_) if config.sign_only => None,
+        Err(e) => {
+            return Err(format!("Error: failed to fetch sender account {}: {}", sender, e).into())
+        }
+    };
```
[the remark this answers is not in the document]
2501babe: youre right, i orginally wrote this expecting to use it more than once, but didnt after refactoring

token/cli/src/main.rs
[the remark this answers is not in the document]
2501babe: actually i think ill just comment this better, checking `is_some()` makes it more robust to the callsite or cli interface getting refactored and passing `None` in in other cases

token/cli/src/main.rs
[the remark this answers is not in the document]
joncinque: Ah cool, thanks for clarifying, retracting this

2501babe:
alright i think i addressed everything! lmk if theres anything i missed or anything else

joncinque (approved):
Awesome, thanks for the great work!

PR #3649, opened 2022-09-26
token-cli: dont push fee_payer to bulk_signers

### LINKED ISSUE #3509: token-cli: bulk_signers has too much stuff

2501babe:
unless im extremely stupid (entirely possible) this is literally all we need to do. authority is always pushed onto the vec in the relevant match arm of `process_command`, totally irrespective of anything to do with `fee_payer`. that means the `fee_payer` push is completely pointless because fees are always paid by `payer` on `Token`

closes #3509

### DIFF
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -161,9 +161,6 @@ impl<'a> Config<'a> {
                 eprintln!("error: {}", e);
                 exit(1);
             });
-        if !bulk_signers.contains(&fee_payer) {
-            bulk_signers.push(fee_payer.clone());
-        }
 
         let verbose = matches.is_present("verbose");
         let output_format = matches


2501babe:
actually this needs to be rebased after #3629 lands because it depends on everything using the client

joncinque (approved):
Yep, this looks right to me, nice!  At this point, the fee payer is only used internally in the token client, as it should be, so we never have a reason to include it in bulk_signers

PR #3652, opened 2022-09-27
Update README.md

Bhaney44:
I cleaned up exposed URLs in the README.md files. The goal was to make the was to use a consistent link scheme across the document to present a more professional and neat appearance.


joncinque (commented):
Just a little typo to clarify, then we can get this in

joncinque (approved):
Thanks for your contribution!

PR #3661, opened 2022-10-03
SignerType::Ephemeral

guibescos:
The goal of this PR is allowing each proposal to create one account using `system_program::create_account`. To use `system_program::create_account`, the account that gets created needs to sign. To enable this we add an extra PDA seeded by `[b"proposal_extra_account", proposal_id]`. If this PDA is in the instructions of the proposal, spl-governance will sign on behalf of this PDA.
The main motivation of making this PR is that some Solana programs require you to create accounts and giving them to the program before interacting with the program. A notable example is wormhole where sending a message requires you to provide an account as a signer where the wormhole message will be written.
This should allow spl-governance to send 1 wormhole message per proposal, making cross-chain governance possible.


SebastianBor:
> The goal of this PR is allowing each proposal to create one account using `system_program::create_account`. 

This is a great solution to the problem of accounts creation in proposal instructions but I think allowing to create only a single account per `ProposalTransaction` is too restrictive and we can extend it to any number of accounts.

In order to minimise the changes and make it backward compatible we could do the following:

1. Change `AccountMetaData.is_signer` from `bool` to `signer`:`SignerType` enum where the enum would be defined as follows:
```
SignerType {
   None,
   Static,
   Ephemeral
}
```
 https://github.com/solana-labs/solana-program-library/blob/944bb968f3a42af3f6606bb91bba90ade2e54792/governance/program/src/state/proposal_transaction.rs#L44 
With the enum we would have:

- `None` - the account is not a singer and the value is backward compatible with `false` 
- `Static` - the account is normal signer and the value is backward compatible with `true`
- `Ephemeral` - the account is ephemeral signer and is only used during `ProposalTransaction` execution. It would be a PDA with the seeds: [`'ephemeral-signer'`,`proposal_transaction_pubkey`,`account_seq_number_le_bytes`] where `account_seq_number_le_bytes` would be `u16` consecutive number of the given ephemeral signer account within the `ProposalTransaction` account list across all the instructions in the transaction.

2. When executing `ProposalTransaction` instructions add the signer seeds for ephemeral signers if the given instruction needs them
3. It won't be necessary to define `ProposalExtraAccount` because we would never create the ephemeral signer accounts https://github.com/solana-labs/solana-program-library/blob/dd290427bbe0782542a40cdc8bb70fccfd139c89/governance/program/src/state/proposal_extra_account.rs#L14

Please let me know what you think

guibescos:
I'm revisiting this PR after applying the feedback.
Ephemeral signers are a feature for proposals to be able to request extra PDAs from the governance contract that will sign the transaction execution.
The user workflow is :
- At `insert_transaction`, the user declares the account as `signer: SignerType::Ephemeral`. The pubkey of this account does not matter.
- Each `SignerType::Ephemeral` will be replaced by a unique PDA seeded by `proposal_transaction_pubkey` and its rank among the transactions' accounts with `SignerType::Ephemeral`
- When calling `execute_transaction` the ephemeral signers will sign the transaction.

PR #3669, opened 2022-10-05
remove decimals parameter from token client methods

### LINKED ISSUE #3648: token-client: remove `decimals` arg from all functions
i was thinking about this when refactoring `transfer`, it makes sense to me that `Token` should carry `decimals: Option<u8>` because it is, in fact, a property of the token. functions that require `decimals` can return a new `TokenError` variant if its not there

(ping jon next week, this requires editing a *lot* of calls in tests, so i dont want to do it on spec)

2501babe:
this properly belongs on the object. as always, this interface change excludes the confidential functions, which can be stabilized when confidential transfers go in the cli

~rebase on master when #3482 lands~

closes #3648

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -320,11 +320,13 @@ pub fn signers_of(
 fn token_client_from_config(
     config: &Config<'_>,
     token_pubkey: &Pubkey,
+    decimals: Option<u8>,
 ) -> Token<ProgramRpcClientSendTransaction> {
     let token = Token::new(
         config.program_client.clone(),
         &config.program_id,
         token_pubkey,
+        decimals,
         config.fee_payer.clone(),
     );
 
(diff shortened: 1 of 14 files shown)


token/cli/src/main.rs
```diff
@@ -1671,15 +1671,16 @@ async fn command_gc(
     for (token_pubkey, accounts) in accounts_by_token.into_iter() {
         println_display(config, format!("Processing token: {}", token_pubkey));
 
-        let token = token_client_from_config(config, &token_pubkey);
+        let token =
+            token_client_from_config(config, &token_pubkey, accounts.values().next().map(|a| a.1));
```
joncinque, line 1675: nit: Rather than pulling out the decimals here, how about making the key for the outer hashmap `(Pubkey, u8)` for the mint pubkey and the decimals.  Then you'll have the decimals available in line 1671 as:

```
for ((token_pubkey, decimals), accounts) in accounts_by_token.into_iter() {
```

And you won't need to store the decimals in the value for each account in the inner hashmap. What do you think?

joncinque (approved):
Just some nits and one little fixup, but nothing that should stop this from going in. Looks good!

PR #3681, opened 2022-10-09
token-2022/confidential-transfer: check pubkey and ciphertext assocaited with proof data to token accounts

samkim-crypto:
Some of the components that are associated with the zk instruction data are not checked to the data components associated with accounts, which could lead to some vulnerabilities. This PR adds these checks.


samkim-crypto:
Subsumed by PR #3768, #3867

PR #3685, opened 2022-10-10
[token-js]: new overload of confirm tx in examples

### LINKED ISSUE #3571: token-js: Avoid using deprecated `confirmTransaction` in examples
#### Problem

The `confirmTransaction` method was recently overloaded and the old method signature using just a signature was deprecated.

#### Solution

Follow the change at  #3567 and make those same changes.

atharmohammad:
Closes #3571


joncinque (approved):
Thanks for the PR! It's too bad that `requestAirdrop` doesn't give a blockhash for the airdrop transaction, so this is probably our best bet

PR #3693, opened 2022-10-11
update solana to 1.14.4

### LINKED ISSUE #3687: compat issues for cli release 2.1.0
ok so i created a 2.1.0 tag, and prior to pushing it to cargo, did a test run with `scripts/cargo-install-all.sh` from the monorepo on 1.10. this is the script that is used to build the binary tarball hosted on the site. unfortunately i got this:
```
+ /home/hana/work/solana/solana/cargo install --locked spl-token-cli --root /tmp/lol --path /home/hana/work/solana/solana-program-library/token/cli
+ exec cargo +1.59.0 install --locked spl-token-cli --root /tmp/lol --path /home/hana/work/solana/solana-program-library/token/cli
error: failed to load manifest for workspace member `/home/hana/work/solana/solana-program-library/token-upgrade/cli`

Caused by:
  failed to load manifest for dependency `spl-token-client`

Caused by:
  failed to parse manifest at `/home/hana/work/solana/solana-program-library/token/client/Cargo.toml`

Caused by:
  namespaced features with the `dep:` prefix are only allowed on the nightly channel and requires the `-Z namespaced-features` flag on the command-line
```

caused by the addition of this:
```
[features]
default = ["display"]
display = ["dep:solana-cli-output"]
```

because this feature is nightly in 1.59

i guess the question is, do we want to mess around with the monorepo build stuff to enforce our desired rust version? since i know we have the `rust-toolchain.toml` file, which i didnt know was being ignored. or can we un feature-lock the dependency, or accomplish the intent here through some other syntax, or something?

cc @joncinque

2501babe:
this all builds locally, normal and sbf. letting ci catch any deprecations or test issues

testing local build against monorepo 1.1{3,4}.x install script, then test and pr monorepo to hardcode 1.10.x at 2.0.17, then i push another version bump here in this pr, and once both get merged i can proceed with a release

closes #3687


2501babe:
alright. ive confirmed by hand that this builds with the 1.14.x script, and confirmed that adding `--version 2.0.17` to the 1.10.x and 1.13.x scripts allows them to build as well

joncinque (approved):
Looks great, thanks! Maybe we should consider adding workspace-wide dependencies to make this process easier, ie https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

2501babe:
hmm... maybe! looking over this feature, it seems like individual packages opt in to the dependencies so that the ones they dont need dont get pulled into builds, which is good

tbh the process of updating the solana dependencies takes literally one second with the script tho. if i had to make this process more streamlined and safer, it would if if ci could ensure that a token-cli version bump (and _only_ that; we dont want this to run every commit, it takes _forever_) correctly builds with the script from whatever the current draft monorepo releases are. this feels... nightmarish to do tho, because it relies on an external repo _and_ like some bullshit on the github website (since git cant actually track this concept) so im not sure if its really possible (especially since how ci works is a total mystery to me)

joncinque:
We checkout the whole solana monorepo for the 2xtx token build, so it wouldn't be too too bad if we wanted to do it... but yeah, reading versions could be really crappy.

Unless I'm missing another step, we essentially need to make sure that it builds under a few versions of Rust, right?  We could add just a `cargo check` step to token-cli with a multiple versions of Rust.  Would that do it?

2501babe:
well the twotx ci flow is a hack that goes away once the feature is live afaik

hmm yea i think the problem could simplify to building on multiple rusts. in order of descending robustness to changes in the monorepo build process, the solutions are:
* `scripts/cargo-install-all.sh` runs successfully for all draft solana releases, when forced to use a local version of the unreleased token-cli
* `"$cargo" $maybeRustVersion install --locked spl-token-cli --root "$installDir" --path /path/to/cli` runs successfully for all draft solana release versions of `cargo`, a script in the root, not the binary
* `cargo check --locked` runs successfully for all versions of rust listed in `ci/rust-version.sh` in all draft solana releases

and then below this you get into hardcoding rust versions we might care about, which become stale and need to be kept current. hmm

maybe im overthinking it. i guess if you were gonna forget to update the rust versions in some file we use to track what the monorepo rust versions should be, youd also forget the run the monorepo install script anyway, so its not like its _more_ brittle

well, hm, honestly any solution except for actually running the monorepo install script, im gonna run the script anyway, so it doesnt save me anything. and the question is moot at the moment since we are all on 1.60 having a good time in 1.14 and previous versions are now locked on 2.0.17. so. gonna drop this for now

if theres a good way to detect what the draft releases for the monorepo are without liek scraping the website tho that might be something to consider in the future

PR #3710, opened 2022-10-13
fix stake-pool/program/tests/update_validator_list_balance test

HaoranYi:
With partitioned stake rewards, the rewards are no longer credited on the start block of the epoch. Therefore, the assert in the test that checks for rewards at the start block of the epoch is broken. 
TODO Update the test once we have the rpc method ready to query the reward interval.

### DIFF
--- a/stake-pool/program/tests/update_validator_list_balance.rs
+++ b/stake-pool/program/tests/update_validator_list_balance.rs
@@ -209,7 +209,11 @@ async fn success() {
         &stake_pool_accounts.validator_list.pubkey(),
     )
     .await;
-    assert!(new_lamports > initial_lamports);
+
+    // With partition rewards, the reward is paid after reward interval
+    // So the lamports and credits_observed won't change at the start of the new epoch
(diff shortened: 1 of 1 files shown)


joncinque:
We have end-to-end staking tests that use a test validator with short epochs of 32 slots, and then we wait for activations to happen and check that everything works as expected.

It's a small use-case, but how will partitioned rewards work with networks that have short epochs?

HaoranYi:
Yes. The intervals are adaptive to the numbers of slots. Calc interval = number of slots in epoch / 10.min(100). Credit interval = num of slots in epoch /20.min(50)

HaoranYi:
I at you on the code for it in the pull request.

joncinque:
> Yes. The intervals are adaptive to the numbers of slots. Calc interval = number of slots in epoch / 10.min(100). Credit interval = num of slots in epoch /20.min(50)

Ok great, perfect!

> I at you on the code for it in the pull request.

Yes please

PR #3712, opened 2022-10-13
token-2022: implement CpiGuard

### LINKED ISSUE #3694: PROPOSAL: `CpiGuard` Extension
# CpiGuard

## tldr

cpi guard is an extension that prevents certain token actions inside cpi, to protect users from implicitly signing for actions they cant see, hidden in third-party programs

full summary of changes:
* changes to existing token-2022 processors:
  - `transfer`: if extension enabled and in cpi, authority *must* be delegate, or abort
  - `burn`: if extension enabled and in cpi, authority *must* be delegate, or abort
  - `close_account`: if extension enabled and in cpi, destination *must* be owner, or abort [IN FLUX: we may further limit to native mint accounts]
  - `approve`: if extension enabled and in cpi, abort
  - `set_authority`:
    * if `Account` and extension enabled and in cpi *and* authority to change is close authority *and* intent is to remove it, remove it
    * otherwise, if `Account` and extension enabled and in cpi, abort
    * if `Account` and authority to change is owner and extension enabled, abort *regardless of cpi* (this is to prevent moving a `CpiGuard` account to a program)
* new `CpiGuard` instruction and processor, modeled after `RequiredTransferMemos` interface. changing `CpiGuard` status in cpi is *not allowed*
* program tests for all invariants listed above
* token-client support for `CpiGuard`
* token-cli support for `CpiGuard` (todo after extension merged)
* token-js support for `CpiGuard` (todo after extension merged)
* documentation in `token-2022/extensions.mdx` for `CpiGuard` (todo after extension merged)

this is an extension that can be enabled or disabled by users at-will. it does not break backwards compat. rather, it simply automates a choice to systematically decline to sign certain types of transactions

## background

interacting with a dapp as a user means signing transactions that are constructed for you by its frontend code. there are three fundamental ways for a dapp to transfer funds from the user to the dapp (or, equivalently, burn them):
* insert a transfer instruction in the transaction, which the user signs for
* insert an approve instruction in the transaction, which the user signs for, and perform a cpi transfer under program authority
* require the token program and a user signer for an opaque program instruction, and perform a cpi transfer under user authority

the first two are safe, in that the user can see exactly what is being done, with zero ambiguity. the third is quite dangerous. passing in a wallet signature allows the program to perform any action the user could, without any visibility into what it may be doing. there have been some attempts at workarounds, for instance phantom may simulate the transaction and attempt to warn about balance changes. but, fundamentally, this is intractable [jocular citation of turing 1936]

this also doesnt cover the approve and set authority instructions, which can grant authority to transfer all user funds without immediately effecting a balance change. wallets could also warn if the various authorities change after simulation, but again, this isnt perfect

there are two ways to make this much safer that _dont_ have to grapple with undecidability:
* wallets should warn whenever a wallet signature is made available to an opaque (non-system, non-token) instruction, and users should be educated to treat the request for a signature on such an instruction as highly suspect
* the token program should have a means to prevent cpi calls on user authority from taking place, forcing opaque programs to directly ask for such authority in a user-visible manner

this proposal concerns the second

## proposal

add a new `Account` extension to Token2022: `CpiGuard`

this would allow a user to _opt in_ to backwards-compat-breaking behavior, rejecting certain kinds of cpi actions involving their token account:
* `transfer` via cpi must go through a delegate
* `burn` via cpi must go through a delegate
* `close_account` via cpi must go through a close authority
* `approve` via cpi is banned
* `set_authority` via cpi is banned
* disabling cpi guard via cpi is banned

this change is _non-breaking_ for users who do not enable it, and users are free to enable or disable it as they please. the purpose of this is to protect users from accidentally granting a program permission to perform actions with their signature, if they personally choose not to allow such behavior

it also is part of an attempt to nudge program authors toward designing with "nothing up their sleeve," creating a ux pitfall that encourages them to use user-authed token instructions rather than user-authed cpi calls. especially with address maps live and double-size transactions in the pipeline, program authors should have sufficient space in their transactions to do things the right way

this change does _not_ impact the ability of programs to auth their _own_ accounts, or accounts of other programs, via cpi. they should simply not enable the extension

## thoughts concerns etc

i waffled on whether to ban `close_account` entirely or not. i decided against it because, while close authority does behave in a surprising manner, the user is free to grant it to a program if they really want to

we might want to ban changing cpi guard status in cpi entirely, because if a program enabled it for their own account, they would not be able to disable it. there is also a very stupid edge case where a user could transfer ownership of a cpi-locked account to a program, which then wouldnt be able to close it. honestly i realized _while writing this proposal_ that we needed to ban removing cpi guard in cpi or else a program could trivially bypass it

the tricky bit here is we cannot detect whether an account is a pda or an eoa without an ed25519 syscall, and requiring this overcomplicates a relatively simple extension. perhaps we unconditionally ban changing cpi guard status inside cpi, and we unconditionally ban transferring ownership of an account unless cpi guard is disabled. this feels right but i havent fully thought through all the edge cases

i considered making this extension trimodal, with a third "ban token actions via cpi entirely" state. but this seems superfluous. the point is not to prevent action from being taken, but to force developers to expose those actions to the user

its worth noting that this may seem like it overlaps the functionality of `ImmutableOwner`, but it doesnt. we prevent a program from changing an account owner via cpi, but the program is still free to surface a `set_authority` instruction for owner, and a user is free to sign for it

the owner/authority distinction is not perfect, if the user were for whatever reason to set themself as their own delegate or close authority (assuming this isnt blocked the processor, i havent checked). we could check that the supposed authority is different from owner. but we could also just say this is dumb

## alternate names

my original name was `SafeCpi`, which i immediately discarded in favor of `SaferCpi` so as to disclaim all liability for the consequences of my actions. `CpiLock` was another candidate. `TokenGuard` or `NoSignaturePassthrough` or some other things were also considered. i think `CpiGuard` sounds best, but im open to suggestions

## fin

please let me know questions/comments/words of encouragment/stern rebukes and chastisements, and feel free to tag anyone who you think might also have thoughts or should also have visibility on this. ideally i get this in before we freeze Token2022, but, i guess we will see

cc @joncinque @CriesofCarrots @mvines

2501babe:
closes #3694


token/program-2022/src/error.rs
```diff
@@ -156,9 +155,17 @@ pub enum TokenError {
...
+    // 40
     /// The deposit amount for the confidential extension exceeds the maximum limit
     #[error("Deposit amount exceeds maximum limit")]
     MaximumDepositAmountExceeded,
+    /// Cpi Guard is enabled, and a program tried to take a protected action via cpi
+    #[error("Cpi Guard is enabled, and a program tried to take a protected action via cpi")]
+    CpiGuardEnabled,
+    /// Cpi Guard status cannot be changed in cpi
+    #[error("Cpi Guard status cannot be changed in cpi")]
+    CpiGuardSettingsLocked,
```
2501babe, line 168: i might replace the first error here with a different error for every single disallowed action. but im not sure if thats "too much"

2501babe:
well the first pass on the processor and instruction is basically done. i copied everything from `RequiredTransferMemos` almost exactly, partly because it does exactly what i need, and partly because im unsure about how closely i need to adhere to a house style for mainnet programs

2501babe:
alright the hardest part of this entire thing honestly was setting up the ability to test calls to token22 through a cpi intermediary and that is dooone. and i have correct tests for the "cpi guard cant be changed in cpi" bullet point. everything works. next up: add the guards into the token instructions!

2501babe:
with fresh eyes on this after the weekend, im not sure about my cpi testing strategy. i wrote this like, each testable action gets one passthrough program instruction that performs the required action, obviating the need to use the data array for anything or implement serialization logic. but now im wondering if it wouldnt be better if the passthrough took an arbitrary serialized instruction, parsed, and invoked it, obviating the need for multiple instructions on the passthrough?

2501babe:
@joncinque the extension is done, if you wanted to take a peek. the only interesting file is `token/program-2022/src/processor.rs`; the extension itself is largely copy-pasted from required memo transfers, and all the other files are test infra, most of which ill be summarily deleting in favor of your instruction wrapper

also ty for telling me about `get_stack_height()`! it made coding this shockingly simple compared to what i thought id have to do

token/program-2022-test/tests/cpi_guard.rs
```diff
@@ -0,0 +1,265 @@
...
+    std::sync::Arc,
+};
+
+// set up a bank and bank client with spl token 2022 and the test cpi caller
+// also creates a token with no extensions and inits two token accounts
+async fn make_context() -> (TestContext, Pubkey) {
+    if std::env::var("BPF_OUT_DIR").is_err() && std::env::var("SBF_OUT_DIR").is_err() {
+        panic!("CpiGuard tests MUST be invoked with `cargo test-sbf`, NOT `cargo test --feature test-sbf`. \
+                In a non-BPF context, `get_stack_height()` always returns 0, and all tests WILL fail.");
+    }
```
joncinque, line 26: You inspired this: https://github.com/solana-labs/solana/pull/28451

token/program-2022-test/tests/cpi_guard.rs
```diff
@@ -0,0 +1,265 @@
...
+                &cpi_caller_id,
+                &spl_token_2022::id(),
+                &alice.pubkey(),
+                &alice.pubkey(),
+            )
+            .unwrap()],
+            &[&alice],
+        )
+        .await
+        .unwrap_err();
```
joncinque, line 101: nit: can you compare this to the expected error? ie

```
let error = token.process_ixs(...).await.unwrap_err();
assert_eq!(error, MyExpectedError);
```

token/program-2022-test/tests/cpi_guard.rs
```diff
@@ -0,0 +1,265 @@
...
```
joncinque, line 132: nit: same here

token/program-2022/src/error.rs
```diff
@@ -156,9 +155,36 @@ pub enum TokenError {
...
+
+    // 40
     /// The deposit amount for the confidential extension exceeds the maximum limit
     #[error("Deposit amount exceeds maximum limit")]
     MaximumDepositAmountExceeded,
+    /// CPI Guard cannot be enabled or disabled in CPI
+    #[error("CPI Guard cannot be enabled or disabled in CPI")]
+    CpiGuardSettingsLocked,
+    /// CPI Guard is enabled, and a program attempted to transfer user funds without using a delegate
+    #[error("CPI Guard is enabled, and a program attempted to transfer user funds without using a delegate")]
```
joncinque, line 167: nit: in all of these error messages, to be totally clear, can you add "from within a CPI" at the end? ie
```suggestion
    #[error("CPI Guard is enabled, and a program attempted to transfer user funds without using a delegate from within a CPI")]
```
Or some other language to that effect

token/program-2022/src/extension/cpi_guard/instruction.rs
```diff
@@ -0,0 +1,100 @@
...
+    ///
+    ///   * Multisignature authority
+    ///   0. `[writable]` The account to update.
+    ///   1. `[]` The account's multisignature owner.
+    ///   2. ..2+M `[signer]` M signer accounts.
+    ///
+    Enable,
+    /// Allow all token operations to happen via CPI as normal.
+    ///
+    /// Fails if the account does not have the extension present.
```
joncinque, line 39: The implementation does not follow this comment, since `disable_cpi_guard` also initializes the extension.  I don't have a strong opinion either way, but I think it's nicer to initialize it even in the case of `disable_cpi_guard`, which is what memo transfer does too.  We'll have to update that other comment too

token/program-2022/src/processor.rs
```diff
@@ -549,6 +564,14 @@ impl Processor {
                         return Err(TokenError::ImmutableOwner.into());
                     }
 
+                    if let Ok(cpi_guard) = account.get_extension::<CpiGuard>() {
+                        if cpi_guard.lock_cpi.into() && in_cpi() {
+                            return Err(TokenError::CpiGuardSetAuthorityBlocked.into());
+                        } else if cpi_guard.lock_cpi.into() {
+                            return Err(TokenError::CpiGuardOwnerChangeBlocked.into());
+                        }
```
joncinque, line 572: nit: I may be being dense, but these two errors seem redundant, and it would be simpler to only surface the `CpiGuardOwnerChangeBlocked` error

token/program-2022/src/processor.rs
```diff
@@ -889,6 +927,15 @@ impl Processor {
                 .base
                 .is_owned_by_system_program_or_incinerator()
             {
+                if let Ok(cpi_guard) = source_account.get_extension::<CpiGuard>() {
+                    if cpi_guard.lock_cpi.into()
+                        && in_cpi()
+                        && !cmp_pubkeys(destination_account_info.key, &source_account.base.owner)
+                    {
+                        return Err(TokenError::CpiGuardCloseAccountBlocked.into());
+                    }
```
joncinque, line 936: I think we'll need to relax this check if `source_account.base.close_authority` is not None.  Otherwise, we can get this kinda hilarious "attack":

* I create an account for you, set my PDA as close authority
* I set ownership of the account to you
* you add CPI guard
* I can never close your account through my program

Does that make sense?

joncinque (commented):
The logic looks great! Just a few little questions and nits

token/program-2022/src/processor.rs
2501babe: its an error either way yea, but the intent is that `CpiGuardSetAuthorityBlocked` is meant to be more "severe" than `CpiGuardOwnerChangeBlocked`. the former indicates that a program is trying to do something fishy, which the user may want to be wary of. the latter can only be triggered by an instruction that was directly visible to them in a transaction they signed, so it probably indicates they just forgot

i recognize im doing a bit of [this](https://xkcd.com/2501/) expecting users to interpret the messages correctly, but... 

in my head there are two types of users of cpi guard. one is extremely savvy, knows what all this shit means, and wants the extension for an additional safety check, or for philosophical reasons. they would care about the distinction and may enable or disable guard when needed. theres also extremely... unsavvy users, who _know_ theyre unsavvy, and they turn the extension on and never turn it off. theyre scared of getting rugged in a way they cant understand or predict, and this gives them a bit of peace of mind that theres a class of attacks they arent susceptible to. to them, all the errors are meaningless anyway

incidentally the current setup is these two errors in the `Owner` arm and just the first error in the `CloseAuthority` arm. the fallthrough idea might make more sense the other way i had it, with a check _before_ the match "IF cpi guard on AND in cpi AND NOT(authority to change is close AND new close is null)", just the owner change check in the `Owner` arm, and nothing in the `CloseAuthority` arm. i changed it to this because it seemed more readable and consistently put cpi checks after owner validation

token/program-2022/src/processor.rs
2501babe: the program could still close the account, theyd just have to direct the lamports to me and not them

although it does raise the interesting question of, is that how this should work? since the original intent of close authority was to allow programs to reclaim lamports from user accounts the program funded...

the broader issue here is adversarial: cpi guard is meant to give _more_ control to users, whereas close authority takes control _away_ from users. fundamentally these two features are in tension

since rent exemption lamps are of negligible value, perhaps the check should look like "IF cpi guard enabled AND in cpi AND mint is native THEN lamports must go to owner"? i had excessive concern about close authority a few weeks ago since i hadnt learned until refactoring `command_gc` that it cant be used to steal tokens. now i kind of regard it as just... obnoxious, but with a very bad edge case for native mint

i think my ideal solution is:
* further limit this check to native mint accounts
* improve `ImmutableOwner` to also ban setting close authority (separate pr)
* hopefully someday deprecate it

token/program-2022/src/processor.rs
joncinque: Gotcha, that makes sense. It's true that it's clearer to fail it for two different reasons in different contexts, and if people ever decide to catch the errors, they'll be able to understand better.  I'll retract this then, thanks!

token/program-2022/src/processor.rs
joncinque: The close authority was specifically added for a project that was creating loads and loads of non-native accounts for their users, but with the ability to reclaim the lamports if they never use them. On the flipside, it's a real edge case, since there probably aren't any projects that are closing user accounts through a program.

If they really wanted to reclaim the account, they could change their program to set the authority to a hot wallet and then do it normally.  I'll retract this one too, seems like too much of an edge case, and there are ways around it if needed.

2501babe:
im now based on the instruction padder and using it for tests. ill address the rest of the fixmes and write tests for the remaining fns tomorrow. also since im removing the caller thing iwrote the commit history is fucked so i will be squash merging this (ive become very fond of rebase merging...)

2501babe:
@joncinque just to clarify when you have a chance:
* was verdict on close account to leave it as it is, or further restrict to native mint?
* would you be open to a pr to prevent setting close authority on an account with `ImmutableOwner`?

joncinque:
> was verdict on close account to leave it as it is, or further restrict to native mint?

Sorry, I was unclear on this -- I think it's good to leave it as it is

> would you be open to a pr to prevent setting close authority on an account with ImmutableOwner?

I'm probably missing something here, since I don't totally understand the rationale.  If an account has an immutable owner, an outside close authority doesn't threaten or change the immutable ownership of the account. You still can't set a different owner. Or is the concept that the close authority exerts some form of "ownership" over the account?

2501babe:
yea i guess something like "granting a close authority is tantamount to partial alienation of ownership" but if that doesnt make sense im not like, strongly in favor

joncinque:
In that case, my vote is to leave it out for now

2501babe:
alright tests are done, final fixes tomorrow and then we should be good

Cargo.toml
```diff
@@ -39,6 +39,7 @@ members = [
   "token/program-2022",
   "token/program-2022-test",
   "token/client",
+  "token/program-2022-test/cpi-caller",
```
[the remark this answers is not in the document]
2501babe: im assuming this is meant to be a comment on my adding my caller program to `build.rs`, which i assumed was necessary (and added your instruction padder to it to replace.) will try removing it and see if things still work but if you meant something else lmk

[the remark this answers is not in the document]
2501babe: lol jk i guess without it in build.rs, ci doesnt build it at all

2501babe:
ok i think this is done. note again i will be squashing this hstory lol

token/program-2022-test/tests/cpi_guard.rs
```diff
@@ -0,0 +1,769 @@
...
+
+    let mut program_test = ProgramTest::new(
+        "spl_token_2022",
+        spl_token_2022::id(),
+        processor!(SplToken2022Processor::process),
+    );
+
+    program_test.add_program(
+        "spl_instruction_padding",
+        instruction_padding_id,
```
joncinque, line 51: up to you, but if you want to make this simpler and not require passing back the pubkey id as the result of this function, you can just use `spl_instruction_padding::id()` everywhere. Also, I've deployed it to all networks

token/program-2022-test/tests/cpi_guard.rs
```diff
@@ -0,0 +1,769 @@
...
+        .unwrap();
+
+    (test_context, instruction_padding_id)
+}
+
+fn client_error(token_error: TokenError) -> TokenClientError {
+    TokenClientError::Client(Box::new(TransportError::TransactionError(
+        TransactionError::InstructionError(0, InstructionError::Custom(token_error as u32)),
+    )))
+}
```
joncinque, line 89: this is a nice helper

token/program-2022-test/tests/cpi_guard.rs
```diff
@@ -0,0 +1,769 @@
...
+            // all other actions succeed outside cpi with guard
+            (_, true, false) => result.unwrap(),
+            // removing a close authority succeeds in cpi with guard
+            (SetAuthTest::RemoveCloseAuth, true, true) => result.unwrap(),
+            // changing owner, adding close, or changing close all fail in cpi with guard
+            (_, true, true) => assert_eq!(
+                result.unwrap_err(),
+                client_error(TokenError::CpiGuardSetAuthorityBlocked)
+            ),
+        }
```
joncinque, line 767: this is a very slick way of checking all cases

joncinque (dismissed):
great work! only micro things to do with as you please

token/program-2022/src/processor.rs
```diff
@@ -829,13 +859,21 @@ impl Processor {
...
-                    authority_info,
-                    authority_info_data_len,
-                    account_info_iter.as_slice(),
-                )?,
+                _ => {
+                    if let Ok(cpi_guard) = source_account.get_extension::<CpiGuard>() {
+                        if cpi_guard.lock_cpi.into() && in_cpi() {
+                            return Err(TokenError::CpiGuardBurnBlocked.into());
+                        }
+                    }
```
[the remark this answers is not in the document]
2501babe: oops, ty i thought i reversed all those

t-nelson:
looks like this [broke](https://buildkite.com/solana-labs/solana/builds/84350#01840d61-575b-4fa9-af38-18ec5bb2b3c9/494-2086) the monorepo ci downstream-builds job somehow.  it's complaining about the shared object not being available, despite a shiny green "compiling" above and no suspicious log lines in between. :shrug: 

gonna revert in the meantime.  feel free to put it back once resolved

PR #3721, opened 2022-10-17
Governance: MultiChoice type Weighted

ochaloup:
#### Summary
Implementation of `Weighted` multi choice vote type. The implementation makes the `MultiChoice` `VoteType` to be of either `Approval` type (existing behaviour) or `Weighted` that makes possible to vote with percentage of the full vote weight.

####  Backward compatibility notes
The change is not backward compatible for multi choice proposals. However the option has never been exposed in the UI and hence hasn't been used. In case there is an unknown instance of spl-gov which uses it it would only affect historical proposals which couldn't be loaded

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2987,16 +2987,28 @@ checksum = "624a8340c38c1b80fd549087862da4ba43e08858af025b236e509b6649fc13d5"
 
 [[package]]
 name = "openssl"
-version = "0.10.38"
+version = "0.10.48"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c7ae222234c30df141154f159066c5093ff73b63204dcda7121eb082fc56a95"
+checksum = "518915b97df115dd36109bfa429a48b8f737bd05508cf9588977b599648926d2"
 dependencies = [
  "bitflags",
  "cfg-if 1.0.0",
  "foreign-types",
  "libc",
  "once_cell",
+ "openssl-macros",
  "openssl-sys",
+]
+
+[[package]]
+name = "openssl-macros"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b501e44f11665960c7e7fcf062c7d96a14ade4aa98116c004b2e37b5be7d736c"
+dependencies = [
+ "proc-macro2 1.0.50",
+ "quote 1.0.21",
+ "syn 1.0.107",
 ]
 
 [[package]]
@@ -3016,9 +3028,9 @@ dependencies = [
 
 [[package]]
 name = "openssl-sys"
-version = "0.9.72"
+version = "0.9.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7e46109c383602735fa0a2e48dd2b7c892b048e1bf69e5c3b1d804b7d9c203cb"
+checksum = "666416d899cf077260dac8698d60a60b435a46d57e82acb1be3d0dad87284e5b"
 dependencies = [
  "autocfg",
  "cc",
--- a/governance/program/src/state/proposal.rs
+++ b/governance/program/src/state/proposal.rs
@@ -115,7 +115,7 @@ pub enum VoteType {
 #[derive(Clone, Debug, PartialEq, Eq, BorshDeserialize, BorshSerialize, BorshSchema)]
 pub enum MultiChoiceType {
     /// Multiple options can be approved with full weight allocated to each approved option
-    Approval,
+    FullWeight,
 
     /// Multiple options can be approved with weight allocated proportionally to the percentage of the total weight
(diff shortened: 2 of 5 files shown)


ochaloup:
The PR was updated to not mixing changes of the new `VoteType` and removing limit of the max permitted options for the proposal.

governance/program/src/state/proposal.rs
```diff
@@ -102,6 +106,19 @@ pub enum VoteType {
     },
 }
 
+/// Type of MultiChoice.
+#[derive(Clone, Debug, PartialEq, Eq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub enum MultiChoiceType {
+    /// Normal multi choice type that requires decision for only one option when voting.
```
SebastianBor, line 112: It's not for just one option when voting. A user can approve multiple options and as a result of voting on an option the user's full weight is allocated to each approved option

ochaloup: Thanks. I changed the explanation comment and added a simple cargo test.

governance/program/src/state/proposal.rs
```diff
@@ -209,7 +226,7 @@ pub struct ProposalV2 {
 impl AccountMaxSize for ProposalV2 {
     fn get_max_size(&self) -> Option<usize> {
         let options_size: usize = self.options.iter().map(|o| o.label.len() + 19).sum();
-        Some(self.name.len() + self.description_link.len() + options_size + 295)
+        Some(self.name.len() + self.description_link.len() + options_size + 296)
```
ochaloup, line 229: @SebastianBor I would like to check with you if adding the byte for the size that comes of adding the `enum` field within the proposal structure is correct here?

ochaloup (commented):
@SebastianBor I adjusted the PR and verified the math as discussed.
I hope the PR would be fine for full review round.

governance/program/src/state/proposal.rs
SebastianBor: Yes, it's because it increases the maximum possible size

governance/program/src/state/proposal.rs
```diff
@@ -87,6 +87,15 @@ pub enum VoteType {
     /// Ex. voters are given 5 options, can choose up to 3 (max_voter_options)
     /// and only 1 (max_winning_options) option can win and be executed
     MultiChoice {
+        /// Type of MultiChoice
+        #[allow(dead_code)]
+        choice_type: MultiChoiceType,
+
+        /// The min number of options a voter mast choose, the default is 1
```
ochaloup, line 94:
```suggestion
        /// The min number of options a voter must choose, the default is 1
```

joncinque (commented):
Really great work overall! Just a few small things to iron out, then this will be ready to go

Cargo.lock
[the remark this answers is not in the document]
SebastianBor: This commit was merged from master to fix the build on this branched. I did full sync with the master to resolve conflicts

governance/program/src/state/proposal.rs
```diff
@@ -102,6 +111,17 @@ pub enum VoteType {
     },
 }
 
+/// Type of MultiChoice.
+#[derive(Clone, Debug, PartialEq, Eq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub enum MultiChoiceType {
+    /// Multiple options can be approved with full weight allocated to each approved option
+    Approval,
```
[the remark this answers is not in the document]
SebastianBor: Renamed to `FullWeight`

governance/program/src/state/proposal.rs
```diff
@@ -818,42 +840,78 @@ impl ProposalV2 {
...
+                                choice_type: MultiChoiceType::Weighted,
+                                min_voter_options: _,
+                                max_voter_options: _,
+                                max_winning_options: _,
+                            } => {
+                                // Calculate the total percentage for all choices for weighted choice vote
+                                // The total must add up to exactly 100%
+                                total_choice_weight_percentage = total_choice_weight_percentage
+                                    .checked_add(choice.weight_percentage)
+                                    .unwrap();
```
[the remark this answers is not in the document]
SebastianBor: Done

governance/program/src/state/proposal.rs
```diff
@@ -1088,13 +1146,15 @@ pub fn assert_valid_proposal_options(
     }
 
     if let VoteType::MultiChoice {
+        choice_type: _,
+        min_voter_options: _,
```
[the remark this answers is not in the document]
SebastianBor: Added the check and updated comments for the fields. In the current version they are only placeholders for future implementations and we only validate the passed values match what the code implicitly does
- `min_voter_options == 1`  -  a voter must choose at least one option 
- `max_voter_options == options.len()` - a voter can choose any number of options 
- `max_winning_options == options.len()` - all options can win

PR #3725, opened 2022-10-18
token-2022: Add PermanentDelegate extension

### LINKED ISSUE #2762: Allow a program permanent token authority even after transfer
Problem: You have a game where a person can buy a piece of property and receive an NFT that allows the person to use the property in-game while they hold the NFT.  However, the game rules require that the property always be up for auction.  At the end of the auction, the on-chain program transfers the NFT from the first player to the auction winner no matter what.  It is against the rules for any player to prevent the auction from happening, but it is totally fine for the player to give the NFT to another player while the auction is happening.  Ideally, whoever owns the NFT at any given time should be able to see it in their wallet, transfer it to another wallet, or also sell it on a secondary auction house outside the game.  In other words, it should just behave like a normal NFT token but the game has the ability to make transfers.

Simply setting a contract PDA as an authority does not work, since the player can revoke authority on their associated token account anytime and prevent the contract from transferring the token at the end of the auction.  Also, if the player transfers the token to another wallet, authority does not transfer.  

An on-chain program cannot enforce this without doing the following:  
The contract mints the NFT, sets the token owner to the player and immediately freezes it.  The token will at least show up in the players wallet now, and the contract can maintain control of the NFT.  BUT, the wallet UX for the token is immediately degraded.  If the user tries to transfer the NFT to another wallet, they'll see an error and be very confused.  Anytime the chain program needs to transfer the NFT, it has to first unfreeze the token, do the transfer, and then immediately refreeze the token which is weird and wasteful. 

To allow transfers at all, the contract itself would have to implement a `Transfer()` function that wraps the token program with an unfreeze(), transfer(), freeze() and have a screen in the game for doing transfers.  This is going to be very confusing for a player and basically means the game has it's own wallet-like UX.  It also creates a walled garden for these tokens that defeats the spirit of web3.    Lastly, with freezing, a third party auction house will not be able work with the token.

Alternatively, doing some type of multsig might work but sounds like a bad idea and even worse UX.  I've seen it also discouraged from being used in other github issues.

Since token 2022 is adding things like transfer fees, I figured I'd put this on the radar too.  Having a sliding scale of delegation and program permissions for tokens could have many other applications as well.  For example, a financial service that allows transaction roll-back for fraud prevention, pre-transfer review by an oracle, limiting token transfers to a whitelist of allowed holding wallets, or allow game items to be consumed / burned when an event happens in game. 

Of course, there is the risk of a malicious service using this mechanism as a rug-pull, but wallets can provide the user with information on the token permissions so they can make informed choices about the token based on mint metadata.

**TLDR;**  allow a program to have unrevocable authority over a token that survives a transfer.

### LINKED ISSUE #2823: DelegatedFreezeAccount Instruction: Prevent user from removing delegate
We have been thinking long and hard about utilizing freeze_authority + delegate to enable temporary and conditional ownership of tokens. (https://cardinal.so) So the user claims the token and it in same instruction delegates it and gets their token-account frozen so it can be revoked on demand.

We have been building a protocol for doing this for new tokens and were planning to try to enable this functionality for existing NFTs (whose freeze_authority is held by metaplex-token-metadata)

Our proposal there was going to be this: freeze/unfreeze instructions that proxy instruction through to spl_token_program IF the signer is the delegate of the single token in the token_account that they are trying to freeze. 
(WIP https://github.com/cardinal-labs/metaplex-program-library/commit/efa533c544628864a8f409c00cfb6c1fbdbf6dbb)

In other words, if someone delegated you the NFT in a token account, you can freeze that token account. This works for NFTs afaict and we were just realizing that it maybe could apply to all token_accounts (if ALL tokens in the account are delegated to you).

Essentially this opens up a world of possibilities around conditional ownership, rentals, time-revoking tokens etc. where you as the holder can claim it but agree by signing away delegate rights that this other program can manage / revoke it from you.

Is this something that makes sense to just add into spl-token program? I can't think of any reasons this would cause issues. The only interesting thing is that if I delegate tokens to someone, they could always freeze my account forever, but they also could always just take the tokens so I think thats okay.

One thing I just thought of was that if this is your ATA, outright freeze might cause you to not be able to accept new tokens. This is a non-issue with NFTs of course but may be a concern with regular tokens potentially. Maybe it should enforce non-ATA if it were to work with FTs.

@joncinque

joncinque:
#### Problem

NFT projects want to implement a Harberger Tax on the tokens, which requires the ability for an external authority to take over the tokens if needed.  A few other projects have expressed the desire for something similar, so it's time to add it in finally.

#### Solution

Add the ability for mints to specify a "permanent delegate" during initialization, that can always transfer or burn tokens, similar to a normal delegate, but with an unlimited amount, and that can't be revoked.

Also, adds a warning during account initialization that the tokens can be seized, which will probably be useful for people hit by the Harberger Tax.

cc @aeyakovenko 

Fixes #2762 Fixes #2823

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -43,7 +43,7 @@ use spl_token_2022::{
         memo_transfer::MemoTransfer,
         mint_close_authority::MintCloseAuthority,
         transfer_fee::{TransferFeeAmount, TransferFeeConfig},
-        ExtensionType, StateWithExtensionsOwned,
+        BaseStateWithExtensions, ExtensionType, StateWithExtensionsOwned,
     },
     instruction::*,
     state::{Account, AccountState, Mint},
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -744,6 +744,7 @@ async fn command_authorize(
         AuthorityType::TransferFeeConfig => "transfer fee authority",
         AuthorityType::WithheldWithdraw => "withdraw withheld authority",
         AuthorityType::InterestRate => "interest rate authority",
+        AuthorityType::PermanentDelegate => "permanent delegate",
     };
 
     let (mint_pubkey, previous_authority) = if !config.sign_only {
@@ -781,6 +782,7 @@ async fn command_authorize(
                         Err(format!("Mint `{}` is not interest-bearing", account))
                     }
                 }
+                AuthorityType::PermanentDelegate => unimplemented!(),
             }?;
 
             Ok((account, previous_authority))
@@ -813,7 +815,8 @@ async fn command_authorize(
                 | AuthorityType::CloseMint
                 | AuthorityType::TransferFeeConfig
                 | AuthorityType::WithheldWithdraw
-                | AuthorityType::InterestRate => Err(format!(
+                | AuthorityType::InterestRate
+                | AuthorityType::PermanentDelegate => Err(format!(
                     "Authority type `{}` not supported for SPL Token accounts",
                     auth_str
                 )),
@@ -3662,6 +3665,7 @@ async fn process_command<'a>(
                 "transfer-fee-config" => AuthorityType::TransferFeeConfig,
                 "withheld-withdraw" => AuthorityType::WithheldWithdraw,
                 "interest-rate" => AuthorityType::InterestRate,
+                "permanent-delegate" => AuthorityType::PermanentDelegate,
                 _ => unreachable!(),
             };
 
--- a/token/program-2022/src/extension/permanent_delegate.rs
+++ b/token/program-2022/src/extension/permanent_delegate.rs
@@ -20,7 +20,7 @@ impl Extension for PermanentDelegate {
 
 /// Attempts to get the permanent delegate from the TLV data, returning None
 /// if the extension is not found
-pub fn maybe_get_permanent_delegate<S: BaseState, BSE: BaseStateWithExtensions<S>>(
+pub fn get_permanent_delegate<S: BaseState, BSE: BaseStateWithExtensions<S>>(
     state: &BSE,
 ) -> Option<Pubkey> {
     state
--- a/token/program-2022/src/extension/permanent_delegate.rs
+++ b/token/program-2022/src/extension/permanent_delegate.rs
@@ -1,0 +1,18 @@
+use {
+    crate::{
+        extension::{Extension, ExtensionType},
+        pod::*,
+    },
+    bytemuck::{Pod, Zeroable},
+};
+
+/// Permanent delegate extension data for mints.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct PermanentDelegate {
+    /// Optional authority to close the mint
+    pub delegate: OptionalNonZeroPubkey,
+}
+impl Extension for PermanentDelegate {
+    const TYPE: ExtensionType = ExtensionType::PermanentDelegate;
+}
--- a/token/program-2022/src/extension/permanent_delegate.rs
+++ b/token/program-2022/src/extension/permanent_delegate.rs
@@ -1,9 +1,10 @@
 use {
     crate::{
-        extension::{Extension, ExtensionType},
+        extension::{BaseState, BaseStateWithExtensions, Extension, ExtensionType},
         pod::*,
     },
     bytemuck::{Pod, Zeroable},
+    solana_program::pubkey::Pubkey,
 };
 
 /// Permanent delegate extension data for mints.
@@ -15,4 +16,15 @@ pub struct PermanentDelegate {
 }
 impl Extension for PermanentDelegate {
     const TYPE: ExtensionType = ExtensionType::PermanentDelegate;
+}
+
+/// Attempts to get the permanent delegate from the TLV data, returning None
+/// if the extension is not found
+pub fn maybe_get_permanent_delegate<S: BaseState, BSE: BaseStateWithExtensions<S>>(
+    state: &BSE,
+) -> Option<Pubkey> {
+    state
+        .get_extension::<PermanentDelegate>()
+        .ok()
+        .and_then(|e| Option::<Pubkey>::from(e.delegate))
 }
--- a/token/program-2022/src/extension/permanent_delegate.rs
+++ b/token/program-2022/src/extension/permanent_delegate.rs
@@ -10,7 +10,7 @@ use {
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
 pub struct PermanentDelegate {
-    /// Optional authority to close the mint
+    /// Optional permanent delegate for transferring or burning tokens
     pub delegate: OptionalNonZeroPubkey,
 }
 impl Extension for PermanentDelegate {
--- a/token/program-2022/src/instruction.rs
+++ b/token/program-2022/src/instruction.rs
@@ -622,10 +622,13 @@ pub enum TokenInstruction<'a> {
     /// Accounts expected by this instruction:
     ///
     ///   0. `[writable]` The mint to initialize.
+    ///
+    /// Data expected by this instruction:
+    ///   Pubkey for the permanent delegate
+    ///
     InitializePermanentDelegate {
-        /// Authority that may sign for `Transfer` instructions on any account
-        #[cfg_attr(feature = "serde-traits", serde(with = "coption_fromstr"))]
-        delegate: COption<Pubkey>,
+        /// Authority that may sign for `Transfer`s and `Burn`s on any account
+        delegate: Pubkey,
     },
 }
 impl<'a> TokenInstruction<'a> {
@@ -759,7 +762,7 @@ impl<'a> TokenInstruction<'a> {
             33 => Self::InterestBearingMintExtension,
             34 => Self::CpiGuardExtension,
             35 => {
-                let (delegate, _rest) = Self::unpack_pubkey_option(rest)?;
+                let (delegate, _rest) = Self::unpack_pubkey(rest)?;
                 Self::InitializePermanentDelegate { delegate }
             }
             _ => return Err(TokenError::InvalidInstruction.into()),
@@ -919,7 +922,7 @@ impl<'a> TokenInstruction<'a> {
             }
             &Self::InitializePermanentDelegate { ref delegate } => {
                 buf.push(35);
-                Self::pack_pubkey_option(delegate, &mut buf);
+                buf.extend_from_slice(delegate.as_ref());
             }
         };
         buf
@@ -1786,14 +1789,16 @@ pub fn initialize_non_transferable_mint(
 pub fn initialize_permanent_delegate(
     token_program_id: &Pubkey,
     mint_pubkey: &Pubkey,
-    delegate: Option<&Pubkey>,
+    delegate: &Pubkey,
 ) -> Result<Instruction, ProgramError> {
     check_program_account(token_program_id)?;
-    let delegate = delegate.cloned().into();
     Ok(Instruction {
         program_id: *token_program_id,
         accounts: vec![AccountMeta::new(*mint_pubkey, false)],
-        data: TokenInstruction::InitializePermanentDelegate { delegate }.pack(),
+        data: TokenInstruction::InitializePermanentDelegate {
+            delegate: *delegate,
+        }
+        .pack(),
     })
 }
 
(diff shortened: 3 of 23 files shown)


token/program-2022/src/instruction.rs
```diff
@@ -735,6 +751,10 @@ impl<'a> TokenInstruction<'a> {
             31 => Self::CreateNativeMint,
             32 => Self::InitializeNonTransferableMint,
             33 => Self::InterestBearingMintExtension,
+            35 => {
```
joncinque, line 754: this is `35` out of deference to #3712

token/cli/src/main.rs
```diff
@@ -614,6 +614,7 @@ async fn command_authorize(
         AuthorityType::TransferFeeConfig => "transfer fee authority",
         AuthorityType::WithheldWithdraw => "withdraw withheld authority",
         AuthorityType::InterestRate => "interest rate authority",
+        AuthorityType::PermanentDelegate => "permanent delegate",
```
jordansexton, line 617: Should this be
```suggestion
        AuthorityType::PermanentDelegate => "permanent delegate authority",
```

2501babe: imo no, existing convention on token objects is `freeze_authority` `close_authority` etc, but jsut plain `delegate`. in my head, if a "delegate authority" existed, it would be something that can modify the delegate

kevinfaveri:
Brainstorming here the impacts of this on the NFT world (if we have a better place to discuss and voice opinions, send me):

Seems like an amazing way to solve things on chain. Someone could develop a monitor of royalties that automatically rollback a trade that happened without paying royalties (did not pay royalties > transfer the nft back to the old owner > relist the nft for the same price it was > return the money paid to the buyer [except marketplace fees if they exist, of course])

This way, you don't need to rely on a single centralized smart contract like some solutions on the market right now. The trade off tho is that this would allow ruggers to burn the project to the ground (literally); get away NFTs you own and list all of them when preparing to rug. 

So a point to consider, I guess. Not that rugged NFTs are worth much, but today people: are able to still sell that reduce the 100% loss to like 50% loss in most cases; the project can be derrugged; etc. 

Also, a bit concerned about the implications of this going against the "web3 ethos"  and the implications for Solana.

TLDR: I think it is a good addition. Way more corporate tho and less descentralised imo; but practical, as collections who hold the IP of it would be able to forcefully enforce their IP or power over the collection in any situation. In a world where people are used to partial ownership (don't think this is the case right now, but think this is where we are headed in all blockchains... You can see with USDC for example on ETH).

2501babe:
>Someone could develop a monitor of royalties that automatically rollback a trade that happened without paying royalties (did not pay royalties > transfer the nft back to the old owner > relist the nft for the same price it was > return the money paid to the buyer [except marketplace fees if they exist, of course])
the last step isnt possible. the nft mint would have the power to move the nft, but not any token they dont control the mint to

i guess it would serve as a strong deterrant tho, if you know that buying an nft without paying the royalty means a program might steal it back from you and not return your money. but, isnt the point of the transfer fee extension supposed to be that working "around" a royalty system is programmatically impossible?

token/program-2022/src/instruction.rs
```diff
@@ -605,6 +605,23 @@ pub enum TokenInstruction<'a> {
...
+    /// `InitializeMint`.
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
+    /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
+    /// then space required for this extension, plus any others.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint to initialize.
+    InitializePermanentDelegate {
```
2501babe, line 620: this comment should include the data expected

token/program-2022/src/extension/permanent_delegate.rs
```diff
@@ -0,0 +1,18 @@
...
+    },
+    bytemuck::{Pod, Zeroable},
+};
+
+/// Permanent delegate extension data for mints.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct PermanentDelegate {
+    /// Optional authority to close the mint
+    pub delegate: OptionalNonZeroPubkey,
```
2501babe, line 14: this comment needs to be updated

token/program-2022/src/instruction.rs
```diff
@@ -1744,6 +1773,21 @@ pub fn initialize_non_transferable_mint(
...
+) -> Result<Instruction, ProgramError> {
+    check_program_account(token_program_id)?;
+    let delegate = delegate.cloned().into();
+    Ok(Instruction {
+        program_id: *token_program_id,
+        accounts: vec![AccountMeta::new(*mint_pubkey, false)],
+        data: TokenInstruction::InitializePermanentDelegate { delegate }.pack(),
+    })
+}
+
```
2501babe, line 1790: is there a reason not to error here if they pass in `None`?

2501babe, line 1790: is there any reason to allow the user to init this as `None`? since theres no functional difference between not having the extension at all. (if we end up rejecting this, should also change `process_initialize_permanent_delegate` to reject it)

token/program-2022/src/processor.rs
```diff
@@ -298,24 +308,32 @@ impl Processor {
...
             } else {
                 0
-            }
+            };
+
+            let maybe_permanent_delegate = mint
+                .get_extension::<PermanentDelegate>()
+                .map(|e| Option::<Pubkey>::from(e.delegate))
+                .ok()
+                .flatten();
```
2501babe, line 323: im not requesting this change but ptsd from haskell compells me to note that the above (fmap and join) is equivalent to
```
            let maybe_permanent_delegate = mint
                .get_extension::<PermanentDelegate>()
                .ok()   
                .and_then(|e| Option::<Pubkey>::from(e.delegate)));
```

this (bind)

2501babe:
aside from the few extremely minor things above, this is approved from me! very straightforward overall

kevinfaveri:
> > Someone could develop a monitor of royalties that automatically rollback a trade that happened without paying royalties (did not pay royalties > transfer the nft back to the old owner > relist the nft for the same price it was > return the money paid to the buyer [except marketplace fees if they exist, of course])
> 
> the last step isnt possible. the nft mint would have the power to move the nft, but not any token they dont control the mint to
> 
> i guess it would serve as a strong deterrant tho, if you know that buying an nft without paying the royalty means a program might steal it back from you and not return your money. but, isnt the point of the transfer fee extension supposed to be that working "around" a royalty system is programmatically impossible?
> 
> (the tax mechanic however is a unique usecase that makes sense to me)
> 
> > Also, a bit concerned about the implications of this going against the "web3 ethos" and the implications for Solana.
> 
> tbh i had the same reaction initially, but i guess it isnt really any more egregious than freeze authority. "i can take away your thing but might give it back if you do a thing" is functionally equivalent to "i can freeze your thing but might unfreeze it if you do a thing." i would feel uncomfortable if such a feature became ubiquitous on normal token mints (i dont know anything about nft community ethos to judge sentiment there) but i hope that user preference against such a thing when it doesnt provide value serves as a counterweight

-> For the last step... Yeah, it would not be possible as it is two different parties (marketplace/collection owner). And about the transfer fee extension... Yeah it would work for cryptocurrencies, but for NFTs that only have a single master edition (1 supply) how would it work? Because from docs (https://spl.solana.com/token-2022/extensions#transfer-fees) feels like it would need to get the fees as the token it is transferring, no? Might work with AMM pools tho

-> About the web3 ethos, all else... This is really tricky. I feel like we need blockchain 101 education for the NFT community (I mean, not so long ago, they were really surprise about the fact that 99% of NFTs on Solana are mutable --- Metashield episode). I fear the same reaction if this is the biz model that gets traction (gets used by 99% of the collections). Anyway, I truly see this all as inevitable and see things as you see: we probably moving to an world with multiple types of NFTs assets

2501babe:
oh yea sorry youre right, for some reason i had it in my head that transfer fees were assessed in native sol

token/program-2022/src/extension/permanent_delegate.rs
joncinque: Man, I really thought I caught all of those.  thanks!

token/program-2022/src/instruction.rs
joncinque: That's a good point. I was thinking for maximum flexibility to do this, but it's kind of a silly case to support. I'll make it mandatory everywhere

token/program-2022/src/processor.rs
joncinque: That's probably better than using a `flatten()`, I always feel a bit icky when I do it

2501babe:
new thought because of #3785, should permanent delegates be able take from frozen accounts? in cases where one person controls all the authorities, its more an ergonomic than functional concern, since they can unfreeze/transfer/freeze in one transaction. but for cases where they may be controlled by separate entities (i can imagine companies that split them up for user trust/security reasons), we need to make a judgement call on which authority ultimately has "supremacy"

joncinque:
It's a good point, and I considered it during the implementation.  Overall, I think it's better to keep it as is and keep each authority separated and clear, but I could be wrong.

joncinque:
This is ready to go from my side.  Anyone else want to take a look at this -- maybe @CriesofCarrots or @samkim-crypto ?

CriesofCarrots:
> This is ready to go from my side. Anyone else want to take a look at this -- maybe @CriesofCarrots or @samkim-crypto ?

Sure, I'll review tomorrow!

token/program-2022/src/processor.rs
```diff
@@ -299,24 +309,31 @@ impl Processor {
...
             // Transfer fee amount extension exists on the account, but no mint
             // was provided to calculate the fee, abort
-            if source_account
+            let fee = if source_account
                 .get_extension_mut::<TransferFeeAmount>()
                 .is_ok()
             {
                 return Err(TokenError::MintRequiredForTransfer.into());
             } else {
                 0
```
samkim-crypto, line 334: minor: to save an assignment, can we simplify by directly returning `(0, None)` in the else?

samkim-crypto (commented):
Looks good to me!

CriesofCarrots (commented):
Looks good! Just one code pattern question.

token/program-2022/src/processor.rs
```diff
@@ -832,13 +871,27 @@ impl Processor {
                 return Err(TokenError::MintDecimalsMismatch.into());
             }
         }
+        let maybe_permanent_delegate = mint
+            .get_extension::<PermanentDelegate>()
+            .map(|e| Option::<Pubkey>::from(e.delegate))
+            .ok()
+            .flatten();
```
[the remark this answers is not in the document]
2501babe: oops, this is because i suggested `and_then` in the first instance in an offhand "its cool these are logically equivalent" way rather than a "you should make this change" way and so didnt flag the second instance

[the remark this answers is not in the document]
joncinque: So, I did the DRY bit, which ended up being pretty gnarly, since one call is on `StateWithExtensions` and the other is on `StateWithExtensionsMut`, which required a trait, which I then needed to use everywhere.  Let me know what you think!

token/program-2022/src/extension/permanent_delegate.rs
```diff
@@ -16,3 +17,14 @@ pub struct PermanentDelegate {
 impl Extension for PermanentDelegate {
     const TYPE: ExtensionType = ExtensionType::PermanentDelegate;
 }
+
+/// Attempts to get the permanent delegate from the TLV data, returning None
+/// if the extension is not found
+pub fn maybe_get_permanent_delegate<S: BaseState, BSE: BaseStateWithExtensions<S>>(
```
CriesofCarrots, line 23: Bikeshedding: either `maybe_permanent_delegate` or `get_permanent_delegate` would seem more in line with precedent, at least in the monorepo where I'm more familiar. Looks like we don't really have other methods with `maybe` prefixes in SPL repo.

token/program-2022/src/processor.rs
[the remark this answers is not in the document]
CriesofCarrots: Sorry that was so much plumbing, but I actually like it quite a lot! I feel like I should have thought of those State implementation changes a while ago...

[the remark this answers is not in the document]
joncinque: Yeah, it ended up being really neat at the base level :smile:

joncinque:
I've tested the monorepo CI to make sure nothing breaks, so merging this in!

PR #3729, opened 2022-10-20
Governance: removing limit for number of options to be part of a proposal

ochaloup:
The current implementation defines an hard limit of max 10 options being part of a proposal. This removes the limit and let the processing to push to whatever number.
In current state it's possible to propose 10s of options for the proposal otherwise an copute limit is hit. When the coputation limit is increased with the `ComputeBudget` instruction it was possible to work with 200 of options. That limits would be good to be investigated further.

### DIFF
--- a/governance/program/src/processor/mod.rs
+++ b/governance/program/src/processor/mod.rs
@@ -97,15 +97,17 @@ pub fn process_instruction(
         vote_type,
         options,
         use_deny_option,
+        proposal_seed,
     } = &instruction
     {
         // Do not iterate through options
-        msg!("GOVERNANCE-INSTRUCTION: CreateProposal {{name: {:?}, description_link: {:?}, vote_type: {:?}, use_deny_option: {:?}, number of options: {} }}",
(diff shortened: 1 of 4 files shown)


ochaloup:
Related weighted choice vote type https://github.com/solana-labs/solana-program-library/pull/3721

governance/program/tests/program_test/mod.rs
```diff
@@ -2011,7 +2014,11 @@ impl GovernanceProgramTest {
 
         self.bench
             .process_transaction(
-                &[create_proposal_transaction],
+                &[
+                    ComputeBudgetInstruction::set_compute_unit_limit(1_000_000u32),
```
SebastianBor, line 2018: Let's increase the budget only when it's needed and not every time we invoke `CreateProposal` instruction, it could be generalised at the bench level by adding `with_compute_unit_limit()`which would set a flag indicating next `process_transaction` call should include the budget instructions 

And then please add a test which creates proposal with number of options requiring the budget

ochaloup: @SebastianBor Thanks for your feedback.
I added the method as you mentioned to `bench` and provided the test.
Please, review if I haven't missed something.

PR #3735, opened 2022-10-21
Account Compression: Rename Wrapper to Noop

ngundotra:
* Rename `Wrapper` -> `SplNoop`
* Rename Account pubkey from `log_wrapper` to `noop`

### DIFF
--- a/account-compression/programs/account-compression/Cargo.toml
+++ b/account-compression/programs/account-compression/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "spl-account-compression"
-version = "0.1.3"
+version = "0.1.4"
 description = "Solana Program Library Account Compression Program"
 authors = ["Solana Maintainers <maintainers@solana.foundation>"]
 repository = "https://github.com/solana-labs/solana-program-library"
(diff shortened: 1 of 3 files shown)


account-compression/programs/account-compression/src/data_wrapper/mod.rs
```diff
@@ -11,17 +11,17 @@ use crate::events::{AccountCompressionEvent, ApplicationDataEvent, ApplicationDa
 use anchor_lang::{prelude::*, solana_program::program::invoke};
 
 #[derive(Clone)]
-pub struct Wrapper;
+pub struct SplNoop;
```
joncinque, line 14: nit: the prefix `Spl` isn't used anywhere else, so for consistency this could be called `Noop`. Also, you can rename the directory to be consistent, perhaps as `noop`

account-compression/sdk/tests/accounts/concurrentMerkleTreeAccount.test.ts
```diff
@@ -0,0 +1,149 @@
+import {
```
joncinque, line 1: The tests look good, but can you add them in a separate PR? It's confusing to see new tests in a PR about a rename

joncinque (commented):
Just a couple of little things, then this can get merged in

joncinque (commented):
Looks good!

PR #3738, opened 2022-10-22
`spl-token` cli - use default config file

jkbpvsc:
## Problem
The `spl-token` uses the argument `-C` to specify the cli config file, if the argument isn't provided a default config file is created, using  the default signer `~/.config/solana/id.json`.
This behavior is inconsistent with the main `solana` cli, where in case the cli config file isn't explicitly passed, the cli will use the default cli config file at `~/.config/solana/cli/config.yml`

## Changes
These changes consolidate the default cli config file loading behavior between the `spl-token` and  `solana`.
If the file is not explicitly specified with `-C`, the default cli config at `~/.config/solana/cli/config.yml` if it exists.

Main [solana cli for reference](https://github.com/solana-labs/solana/blob/master/cli/src/clap_app.rs#L17)

### DIFF
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -112,6 +112,8 @@ impl<'a> Config<'a> {
                 eprintln!("error: Could not find config file `{}`", config_file);
                 exit(1);
             })
+        } else if let Some(config_file) = &*solana_cli_config::CONFIG_FILE {
+            solana_cli_config::Config::load(config_file).unwrap_or_default()
         } else {
             solana_cli_config::Config::default()
         };
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -21,7 +21,6 @@ use solana_clap_utils::{
     offline::{self, *},
     ArgConstant,
 };
-use solana_cli_config::CONFIG_FILE;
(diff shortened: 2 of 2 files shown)


2501babe:
this already takes place here: https://github.com/solana-labs/solana-program-library/blob/master/token/cli/src/config.rs#L56-L65

jkbpvsc:
Ah I see.
However the same fallback is missing in the `new_with_clients_and_ws_url` where the `cli_config` gets loaded for the `Config` struct.

https://github.com/solana-labs/solana-program-library/blob/master/token/cli/src/config.rs#L115

`Config::new()` calls `Config::new_with_clients_and_ws_url()` which is when the `Config` gets really initialized, the `cli_config` gets loaded and stored.

2501babe:
ooh yes youre right

2501babe:
(ci failure is unrelated to this, im in the process of attempting to deal with it)

2501babe:
oki if you would be so kind as to rebase on master to make ci happy this should be good to go

jkbpvsc:
Merged

2501babe:
ty for your contribution!

PR #3744, opened 2022-10-24
Governance: Vote change in a single transaction

SebastianBor:
#### Summary
In order to support `ChangeVote` instruction semantics it must be possible to call `RelinquishVote` and `CastVote` in a single transaction. 

#### Implementation Details
`dispose_account` was changed to reallocate account data to 0 and assign it back to `SystemProgram`. This change allows to reallocate `VoteRecord` account in `CastVote` within a single transaction after it was previously disposed by `RelinquishVote`


joncinque (approved):
Nice and simple, looks good!

PR #3748, opened 2022-10-25
ci: Update twotx.patch to apply properly

joncinque:
#### Problem

The twoxtx step is failing due to an outdated patch, see https://github.com/solana-labs/solana-program-library/actions/runs/3316735829/jobs/5478868796 for an example.

#### Solution

Update the patch!


2501babe:
jinx #3747 (ill close mine lol)

joncinque:
Ah my bad, we could have used yours too! Thanks for looking into it too :heart:

PR #3755, opened 2022-10-25
Add memo to transfer()

mikemaccana:
In Solana's native token documentation, there is frequently examples of using memos. Eg: [Solana Cookbook](https://solanacookbook.com/references/basic-transactions.html#how-to-add-a-memo-to-a-transaction) or [Quicknode documentation](https://www.quicknode.com/guides/solana-development/how-to-use-the-solana-memo-program) - memos provide a handy way to see the reason for a transactions, eg comparing the same transaction in Phantom versus [Portal](https://getportal.app/) (using the memo app):

image

However `transfer()`, SPL's high level tool, doesn't allow memos on Token transactions. 

This PR adds an optional `memo` for SPL transactions and an associated test. If memo is `undefined` or `null` or an empty string, no memo will be made. The change is compatible with the existing function signature of `transfer()` as it only adds options and does not change the order.

### DIFF
--- a/token/js/src/actions/transfer.ts
+++ b/token/js/src/actions/transfer.ts
@@ -1,6 +1,7 @@
-import type { ConfirmOptions, Connection, PublicKey, Signer, TransactionSignature } from '@solana/web3.js';
-import { sendAndConfirmTransaction, Transaction } from '@solana/web3.js';
-import { TOKEN_PROGRAM_ID } from '../constants.js';
+import type { ConfirmOptions, Connection, Signer, TransactionSignature } from '@solana/web3.js';
+import { PublicKey } from '@solana/web3.js';
+import { sendAndConfirmTransaction, Transaction, TransactionInstruction } from '@solana/web3.js';
+import { TOKEN_PROGRAM_ID, MEMO_PROGRAM_ID } from '../constants.js';
 import { createTransferInstruction } from '../instructions/transfer.js';
 import { getSigners } from './internal.js';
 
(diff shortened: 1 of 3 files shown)


token/js/src/actions/transfer.ts
```diff
@@ -28,13 +31,27 @@ export async function transfer(
...
 
+    // Add an (optional) note describing the transaction.
+    // Don't bother adding if memo is an empty string
+    if (memo?.length && memoProgramId) {
+        await transaction.add(
+            new TransactionInstruction({
+                keys: [{ pubkey: source, isSigner: true, isWritable: true }],
+                data: Buffer.from(memo, 'utf-8'),
+                programId: new PublicKey(memoProgramId),
+            })
```
joncinque, line 52: Rather than redefining this here, how about using the instruction builder and constant from `@solana/spl-memo` https://github.com/solana-labs/solana-program-library/blob/master/memo/js/src/index.ts ?

token/js/src/actions/transfer.ts
```diff
@@ -28,13 +31,27 @@ export async function transfer(
     amount: number | bigint,
     multiSigners: Signer[] = [],
     confirmOptions?: ConfirmOptions,
-    programId = TOKEN_PROGRAM_ID
+    programId = TOKEN_PROGRAM_ID,
+    memoProgramId = MEMO_PROGRAM_ID,
+    memo?: string
```
joncinque, line 36: Instead of adding these as more complicated / undefined variables, let's make this clearer through a new `actions` file called `transferWithMemo`.  There, you can make the memo mandatory

joncinque (commented):
Thanks for your contribution! For more clarity and easier maintenance, we can refactor this into a new function and reuse `@solana/spl-memo` (more clarity in the actual review). Let me know if you have any questions!

PR #3762, opened 2022-10-26
token-2022: Return errors for malformed extension data

### LINKED ISSUE #3695: [token-2022] Ok returned for malformed extension data (TOB-STK-1)
#### Description
If the Account TLV data is malformed in the TLV record data or is truncated (i.e. the Account data length is less than the start offset plus length of the TLV data), the function returns Ok.

https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/src/extension/mod.rs#L127

#### Recommendations
Short term, return an error if the TLV data is corrupt. This ensures that the Token Program does not continue processing should the provided account's extension data be corrupt.

joncinque:
#### Problem

If extension data is malformed, we can return some false positives in `get_extension_types`.

#### Solution

Check step by step if it's possible to deserialize the TLV data properly, and return an error if not.

Fixes #3695

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -129,20 +129,29 @@ fn get_extension_types(tlv_data: &[u8]) -> Result<Vec<ExtensionType>, ProgramErr
     let mut start_index = 0;
     while start_index < tlv_data.len() {
         let tlv_indices = get_tlv_indices(start_index);
-        if tlv_data.len() < tlv_indices.value_start {
-            return Ok(extension_types);
+        if tlv_data.len() < tlv_indices.length_start {
+            // not enough bytes to store the type, malformed
+            return Err(ProgramError::InvalidAccountData);
         }
         let extension_type =
             ExtensionType::try_from(&tlv_data[tlv_indices.type_start..tlv_indices.length_start])?;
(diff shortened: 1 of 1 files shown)


samkim-crypto (approved):
All looks clear to me!

token/program-2022/src/extension/mod.rs
```diff
@@ -1578,12 +1608,21 @@ mod test {
...
-        let mut buffer = vec![0; BASE_ACCOUNT_LENGTH + 2];
+        // ok since there are two bytes for the type, which is `Uninitialized`
+        let mut buffer = vec![0; BASE_ACCOUNT_LENGTH + 3];
         let mut state = StateWithExtensionsMut::<Mint>::unpack_uninitialized(&mut buffer).unwrap();
         let err = state.get_extension_mut::<MintCloseAuthority>().unwrap_err();
         assert_eq!(err, ProgramError::InvalidAccountData);
 
         assert_eq!(state.get_extension_types().unwrap(), vec![]);
+
+        // malformed since there aren't too bytes for the type
```
CriesofCarrots, line 1619: “too bytes”

samkim-crypto: I should have caught this when I was reviewing. Sorry about that!

Issue #3771, opened 2022-10-27
token-2022: cpiguard docs/js/cli


2501babe:
now that the cpi guard extension is merged, we need to:
* ~support enable/disable from the cli~ (done)
* ~and js client~
* ~add the extension to `spl-token display` (along with permanent delegate)~
* document the extension in `docs/src/token-2022/extensions.mdx`

2501babe:
i landed the necessary account decoder changes in monorepo, so `spl-token display` is now as simple as an extra print line, once we have a new 1.14 release to base on

staccDOTsol:
hannah why r u here

PR #3775, opened 2022-10-28
token-cli: support cpi guard enable/disable

2501babe:
part of #3771. will do js and docs in separate prs. displaying in the cli requires me to pr the solana account decoder

edit: aaand the solana account decoder pr would require a new token-2022 release, so thats not happening for a bit...

tentative plan is after permanent delegate is merged too, we figure out if/when the confidential stuff will be in a good state for a release, then we release token program/client/cli all together. and then i can update the account decoder for both extensions, then update the cli to display them and release that again just itself


joncinque (approved):
Looks great, thanks!

PR #3785, opened 2022-10-30
token-cli: Support DefaultAccountState extension

### LINKED ISSUE #3404: token-cli: Support the `DefaultAccountState` extension
#### Problem

Token-2022 supports setting a default state on all new accounts, but the CLI doesn't expose the functionality.

#### Solution

Support all operations for this extension, which means adding flags / commands to:

* initialize a mint with a default account state using a `--default-account-state frozen|initialized` flag on `create-mint`
* update the default account state using a new command: `update-default-account-state <MINT_ADDRESS> frozen|initialized` with an optional `--freeze-authority <SIGNER>` arg

joncinque:
#### Problem

The default account state extension exists, but isn't supported in the CLI.

#### Solution

Add support for the extension through a flag during `create-token` called `default-frozen-accounts` and a new command `update-default-account-state`.

I don't feel strongly about the flag names, so happy to bikeshed.  We could make it `--default-account-state initialized` or `--default-account-state frozen` as an alternative to be totally flexible during mint initialization, but that seemed more confusing, since people might feel compelled to use `--default-account-state initialized` as a "sensible default", when really they don't want the extension at all.

Fixes #3404

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -37,12 +37,12 @@ use solana_sdk::{
 use spl_associated_token_account::get_associated_token_address_with_program_id;
 use spl_token_2022::{
     extension::{
-        cpi_guard::CpiGuard, interest_bearing_mint::InterestBearingConfig,
-        memo_transfer::MemoTransfer, mint_close_authority::MintCloseAuthority, ExtensionType,
-        StateWithExtensionsOwned,
+        cpi_guard::CpiGuard, default_account_state::DefaultAccountState,
+        interest_bearing_mint::InterestBearingConfig, memo_transfer::MemoTransfer,
+        mint_close_authority::MintCloseAuthority, ExtensionType, StateWithExtensionsOwned,
     },
     instruction::*,
-    state::{Account, Mint},
+    state::{Account, AccountState, Mint},
 };
 use spl_token_client::{
     client::{ProgramRpcClientSendTransaction, RpcClientResponse},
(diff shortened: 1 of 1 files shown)


2501babe:
this all looks good to me! i agree designing the flag this way is better, so users dont feel like they "need" to specify initialized

i wonder if we should automatically create the extension in `AccountState::Initialized` if the user passes in `--enable-freeze` however? to give mints with a freeze authority the choice to enable default frozen later if desired

also i think i might have an issue in token-2022 unrelated to this pr but need to test it first, it doesnt block merging this tho

CriesofCarrots:
> Maybe we go with the `--default-account-state` flag then?

Yeah, I think this is probably the best, most flexible approach. Can make that parameter require `--enable-freeze`, and also use the help text to explain how `--default-account-state initialized` is different from no parameter?

joncinque:
I changed to the `default-account-state` flag with some additional info. Let me know thoughts on the language / impl!

token/cli/src/main.rs
```diff
@@ -417,6 +419,13 @@ async fn command_create_token(
         })
     }
 
+    if let Some(state) = default_account_state {
+        if !enable_freeze {
+            return Err("Token requires a freeze authority to default to frozen accounts".into());
```
CriesofCarrots, line 424: Seems like this is superfluous with `.requires("enable_freeze")` set? Should we assert/expect?

CriesofCarrots (commented):
That parameter ooks good to me! I didn't do a super close review of the whole, though. Let me know if you want me to :)

token/cli/src/main.rs
joncinque: I was thinking in case we call this function in a test incorrectly, but we never do that, so changed it to an assert. thanks!

2501babe:
lgtm! the explanation of the difference between initialized and not using the the extension in the help message will go a long way to solving possible confusion

PR #3756, opened 2022-10-25
revert CpiGuard reversion

2501babe:
to save you the pain of actually reading any of this:
```
hana@laptop:cli$ git show master
commit e69c7706fb5b1c651b086853d5cb1d688958083a (upstream/master, upstream/HEAD, master)
Author: J <jakob.povsic@gmail.com>
Date:   Wed Oct 26 05:41:37 2022 +0800

    `spl-token` cli - use default config file (#3738)
    
    * add missing load from CONFIG_FILE default path

diff --git a/token/cli/src/config.rs b/token/cli/src/config.rs
index ef6ee30f..b6cfff1c 100644
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -112,6 +112,8 @@ impl<'a> Config<'a> {
                 eprintln!("error: Could not find config file `{}`", config_file);
                 exit(1);
             })
+        } else if let Some(config_file) = &*solana_cli_config::CONFIG_FILE {
+            solana_cli_config::Config::load(config_file).unwrap_or_default()
         } else {
             solana_cli_config::Config::default()
         };
```

```
hana@laptop:cli$ git diff 20221013_cpiguard..20221025_cpiguard_antireunvert
diff --git a/token/cli/src/config.rs b/token/cli/src/config.rs
index ef6ee30f..b6cfff1c 100644
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -112,6 +112,8 @@ impl<'a> Config<'a> {
                 eprintln!("error: Could not find config file `{}`", config_file);
                 exit(1);
             })
+        } else if let Some(config_file) = &*solana_cli_config::CONFIG_FILE {
+            solana_cli_config::Config::load(config_file).unwrap_or_default()
         } else {
             solana_cli_config::Config::default()
         };
```

PR #3791, opened 2022-11-01
Fix CpiGuard typo, but actually DRY processor methods

CriesofCarrots:
I was reading #3756 to understand what the new extension does, and stumbled on a typo.
I also noticed an opportunity to DRY the processor code, and then realized I introduced the verbose pattern myself in RequiredMemoTransfers. Wdyt about this tidying?

### DIFF
--- a/token/program-2022/src/extension/cpi_guard/processor.rs
+++ b/token/program-2022/src/extension/cpi_guard/processor.rs
@@ -48,7 +48,7 @@ fn process_enable_cpi_guard(program_id: &Pubkey, accounts: &[AccountInfo]) -> Pr
     Ok(())
 }
 
-fn process_diasble_cpi_guard(program_id: &Pubkey, accounts: &[AccountInfo]) -> ProgramResult {
+fn process_disable_cpi_guard(program_id: &Pubkey, accounts: &[AccountInfo]) -> ProgramResult {
     let account_info_iter = &mut accounts.iter();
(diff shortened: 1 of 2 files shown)


2501babe:
100% in favor, i copied the existing structure just because i thought it was done that way on purpose for style reasons

token/program-2022/src/extension/cpi_guard/processor.rs
```diff
@@ -88,11 +63,11 @@ pub(crate) fn process_instruction(
     match decode_instruction_type(input)? {
         CpiGuardInstruction::Enable => {
             msg!("CpiGuardInstruction::Enable");
-            process_enable_cpi_guard(program_id, accounts)
+            process_toggle_cpi_guard(program_id, accounts, true /* enable */)
         }
         CpiGuardInstruction::Disable => {
             msg!("CpiGuardInstruction::Disable");
-            process_diasble_cpi_guard(program_id, accounts)
```
2501babe, line 95: also thank you for catching this!

2501babe (approved):

CriesofCarrots:
> i thought it was done that way on purpose for style reasons

I figured! Originally, only `enable` initialized the extension, so it seemed clearer for the methods to be separate. When I updated the code to also initialize on `disable`, I neglected to clean up after myself (to wit: that incorrect code comment that you caught and fixed)

PR #3792, opened 2022-11-01
token-cli: Support transfer fee extension

### LINKED ISSUE #3408: token-cli: Support the `TransferFee` extension
#### Problem

Token-2022 has the functionality to include a transfer fee on all transfers of a given mint, but the CLI doesn't support it.

#### Solution

Add support for all parts of the extension, which means adding flags / commands to:

* during `create-mint`, allow for a flag: `--transfer-fee <FEE_CONFIG_AUTHORITY> <WITHDRAW_WITTHELD_AUTHORITY> <FEE_BASIS_POINTS> <MAXIMUM_FEE>`
* new flag during `transfer` to specify a fee: `--expected-fee <FEE_AMOUNT>`  note that this flag is incompatible with `--use-unchecked-instruction`!
* new command: `withdraw-withheld-tokens <DESTINATION_ADDRESS> <MINT_OR_ADDRESS> [<MINT_OR_ADDRESS> ...]` this can accept a list of mints or accounts and create the appropriate withdraw withheld instruction
* automatically add a `HarvestWithheldTokensToMint` instruction during `close` if the account has any withheld tokens
* new command: `set-transfer-fee <FEE_BASIS_POINTS> <MAXIMUM_FEE>` with an optional `--rate-authority <SIGNER>` argument

joncinque:
#### Problem

The transfer fee extension has existed in token-2022 for awhile, but it's not usable in the CLI yet.

#### Solution

It looks like a doozy, but each commit goes in order to implement each of the points in #3408.

Note: this one is *not* urgent, just needed to be done, so I started on it, and kept going. Let me know if you prefer that I break it up.

Fixes #3408

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -39,7 +39,8 @@ use spl_token_2022::{
     extension::{
         cpi_guard::CpiGuard, default_account_state::DefaultAccountState,
         interest_bearing_mint::InterestBearingConfig, memo_transfer::MemoTransfer,
-        mint_close_authority::MintCloseAuthority, ExtensionType, StateWithExtensionsOwned,
+        mint_close_authority::MintCloseAuthority, transfer_fee::TransferFeeAmount,
+        ExtensionType, StateWithExtensionsOwned,
     },
     instruction::*,
     state::{Account, AccountState, Mint},
@@ -135,6 +136,7 @@ pub enum CommandName {
     EnableCpiGuard,
     DisableCpiGuard,
     UpdateDefaultAccountState,
+    WithdrawWithheldTokens,
 }
 impl fmt::Display for CommandName {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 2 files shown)


token/cli/src/main.rs
```diff
@@ -876,6 +877,9 @@ async fn command_transfer(
         maybe_transfer_balance.unwrap()
     };
 
+    let maybe_fee =
+        ui_fee.map(|ui_amount| spl_token::ui_amount_to_amount(ui_amount, mint_info.decimals));
+
```
2501babe, line 964: do you think the fee calculations ever run into precision issues in practice? we convert here from a user-provided float string to a u64, and in the program do a somewhat tricky calculation with integer division and ceil, and `process_transfer` expects the values to match exactly...

im also not sure who in practice is going to be calculating out these values to pass into the cli. perhaps the user should pass in how many bps they expect the fee calculation to be, or they could pass a maximum amount and we calculate to ensure the actual is below that?

or the existing design may be fine, im just thinking out loud

token/cli/src/main.rs
```diff
@@ -1033,6 +1037,18 @@ async fn command_transfer(
...
                 &sender_owner,
                 transfer_balance,
+                maybe_fee,
+                &bulk_signers,
+            )
+            .await?
+    } else if let Some(fee) = maybe_fee {
+        token
+            .transfer_with_fee(
+                &sender,
```
2501babe, line 1129: im not a fan of the fee being a fallback here, if the user passes `--fund-recipient --expected-fee 1` and the fee is 1000000, the transfer will go through. easy solution is to make the flags mutually exclude each other, more annoying solution is to use `calculate_fee` to sanity check they arent being rugged

2501babe:
dont really have any other comments, this all looks good to me. breaking it up into commits like this is appreciated!

token/cli/src/main.rs
joncinque: Your point is definitely valid. It could be that this is very rarely used, but at least it provides some additional optional protection.

Potentially as a safeguard later, we can add a `maximum-expected-fee` parameter to be more loose with checks.  Either way, I think there's value in both.

joncinque:
Merging on red for unrelated failures,  #3813 will fix the twoxtx build, and #3810 will fix the governance build

PR #3795, opened 2022-11-02
token-cli: Add non-transferable mint creation

joncinque:
#### Problem

The non-transferable mint extension exists, but isn't usable in the CLI

#### Solution

Add the ability to create a non-transferable token mint

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -386,6 +386,7 @@ async fn command_create_token(
     authority: Pubkey,
     enable_freeze: bool,
     enable_close: bool,
+    enable_non_transferable: bool,
     memo: Option<String>,
     rate_bps: Option<i16>,
     bulk_signers: Vec<Arc<dyn Signer>>,
@@ -415,6 +416,10 @@ async fn command_create_token(
             rate_authority: Some(authority),
             rate: rate_bps,
         })
+    }
+
+    if enable_non_transferable {
+        extensions.push(ExtensionInitializationParams::NonTransferable);
     }
 
     if let Some(text) = memo {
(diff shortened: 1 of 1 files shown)


token/cli/src/main.rs
```diff
@@ -2160,6 +2165,14 @@ fn app<'a, 'b>(
                             Rate authority defaults to the mint authority."
                         ),
                 )
+                .arg(
+                    Arg::with_name("enable_non_transferable")
+                        .long("enable-non-transferable")
+                        .takes_value(false)
```
2501babe, line 2171:
```suggestion
                        .long("enable-non-transferable")
                        .alias("enable-nontransferable")
                        .takes_value(false)
```

i have ptsd from cargo doing `--no-run` but `--nocapture`

2501babe, line 2171:
```suggestion
                        .alias("enable-nontransferable")
```

i have ptsd from cargo doing `--no-run` but `--nocapture`

token/cli/src/main.rs
```diff
@@ -2160,6 +2165,14 @@ fn app<'a, 'b>(
                             Rate authority defaults to the mint authority."
                         ),
                 )
+                .arg(
+                    Arg::with_name("enable_non_transferable")
+                        .long("enable-non-transferable")
```
2501babe, line 2170:
```suggestion
                        .long("enable-non-transferable")
                        .alias("enable-nontransferable")
```

i have ptsd from cargo doing `--no-run` but `--nocapture`

2501babe:
other than one polite request, this is approved from me!

token/cli/src/main.rs
joncinque: ooh yeah good call, `--nocapture` gets me every time

PR #3798, opened 2022-11-04
[confidential-extension] Temporarily disable fees

samkim-crypto:
Temporarily disabling fees to make breaking updates to the zk-token-sdk.


samkim-crypto:
Merging on red for unrelated failure

PR #3800, opened 2022-11-04
token-cli: Specify that `default-account-state` takes a value

joncinque:
#### Problem

The default account state extension isn't working because the `default-account-state` flag doesn't say to take a value.

#### Solution

Specify `takes_value`


joncinque:
Merging on red for unrelated failure

PR #3812, opened 2022-11-11
token-swap: Fix fee inclusion during single token withdraw

joncinque:
#### Problem

The fee inclusion is backwards during the single withdrawal calculation.  We're already taking the fee into account during the calculations at https://github.com/solana-labs/solana-program-library/blob/de73b277f3a49a1bf90c639975fc696f233ae117/token-swap/program/src/curve/base.rs#L106 and https://github.com/solana-labs/solana-program-library/blob/de73b277f3a49a1bf90c639975fc696f233ae117/token-swap/program/src/curve/base.rs#L134.

However, you'll notice that the second one is incorrect.  Since we're calculating the exact amount of *pool tokens* required to give the *source* amount during withdrawal, we actually need to *add* the fee to the source amount rather than subtract.  By doing it this way, we were essentially discounting the tokens on single-sided withdrawals.

#### Solution

To do this right, we need an inverse fee calculation to figure out how many source tokens we're really going to withdraw.  With that, we finally get the correct results.


token-swap/program/src/curve/fees.rs
```diff
@@ -61,6 +61,29 @@ pub fn calculate_fee(
...
+    fee_denominator: u128,
+) -> Option<u128> {
+    if fee_numerator == 0 {
+        Some(post_fee_amount)
+    } else if fee_numerator == fee_denominator || post_fee_amount == 0 {
+        Some(0)
+    } else {
+        let numerator = post_fee_amount.checked_mul(fee_denominator)?;
+        let denominator = fee_denominator.checked_sub(fee_numerator)?;
+        ceil_div(numerator, denominator)
```
kklas, line 83: shouldn't the inverse be `post_fee_amount * fee_denominator / fee_numerator` ? This seems to say `post_fee_amount * fee_denominator / (fee_denominator - fee_numerator)`

kklas:
> While I was at it, I noticed that the fee calc only took into account the trading fee, and not the owner fee, so I added that too.

The `owner_withdraw_fee` is also not respected.

token-swap/program/src/curve/base.rs
```diff
@@ -393,4 +396,61 @@ mod tests {
...
+                swap_destination_amount,
+                TradeDirection::AtoB,
+                &fees,
+            )
+            .unwrap();
+
+        // deposit just A, get pool tokens
+        let new_pool_tokens = swap_curve
+            .deposit_single_token_type(
+                results.source_amount_swapped,
```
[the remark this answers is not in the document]
joncinque: `source_amount` is not correct, because you might actually swap that whole amount.

The calculation for withdrawal works backwards, so you provide how many tokens you want to withdraw, and it tells you how many pool tokens you'll need for that. So this test does exactly what you said:

1. swap A -> B as the control
2. deposit amount A swapped, get pool tokens
3. withdraw B tokens, make sure that the amount of pool tokens required matches what you got from the deposit

token-swap/program/src/curve/fees.rs
joncinque: the term `inverse` may be confusing. Essentially, given the *output*, we want to know what the *input* was. The typical fee calculation is:

```
output = input * (1 - numerator / denominator)
```

so if we only have the output, we need to do:

```
input = output / (1 - numerator / denominator)
```

kklas:
Thanks for the explanation. The names of the functions and the call stack are super confusing but this looks correct to me now at a glance.

PR #3818, opened 2022-11-15
Governance: Cleanup drive space before executing build

SebastianBor:
#### Summary 

The CI build started failing to compile spl-governance with linker crashing due to too many integrations tests. 
The root cause of the issue was the build running out of drive space and the solution removes unnecessary binaries from  the ubuntu image as per https://github.com/actions/virtual-environments/issues/2840


joncinque (approved):
Nice find, and thanks for fighting through this

PR #3820, opened 2022-11-16
managed-token: add approve/revoke/wrap methods

nothing0012:
This PR adds the following entrypoints

- **Approve**: So that we can add a delegate of the managed-token; Note that even if it has a delegate, the approve and future transfer are still gated by the upstream_authority.
- **Revoke**: So that we can remove any delegate of the managed-token
- **Wrap** (convert existing mint to managed-token, if mint_authority and freeze_authority are signers)

The intended use case is for upstream_authority to be a PDA of some higher level program, and at the same time it adds the native features of approve/revoke, and also with a conversion entrypoint called 'wrap'.

Following up with the managed-token prototype https://github.com/solana-labs/solana-program-library/pull/3573.


managed-token/program/src/instruction.rs
```diff
@@ -70,6 +70,30 @@ pub enum ManagedTokenInstruction {
...
+    #[account(4, name = "freeze_authority")]
+    #[account(5, name = "token_program", desc = "Token program")]
+    Revoke,
+
+    #[account(0, writable, name = "mint")]
+    #[account(1, signer, name = "mint_authority")]
+    #[account(2, signer, name = "freeze_authority")]
+    #[account(3, name = "upstream_authority")]
+    #[account(4, name = "token_program", desc = "Token program")]
+    Wrap,
```
joncinque, line 96: nit: the term `Wrap` typically implies a temporary move that can be undone (unwrapped), and this is giving full ownership of the mint to the program without any way to reverse.

How about something like `MigrateToProgramManagement` or `Migrate` to make it clearer that it's going away from the current manager?

managed-token/program/src/lib.rs
```diff
@@ -220,3 +237,59 @@ pub fn process_close(accounts: &[AccountInfo]) -> ProgramResult {
...
+        token_program,
+    )?;
+    set_authority(
+        mint,
+        &authority,
+        AuthorityType::FreezeAccount,
+        freeze_authority,
+        token_program,
+    )
+}
```
joncinque, line 295: In general, I'm bearish about instructions that can be done outside of a program.  For example, this can be easily accomplished with a client library or CLI.

On the flip side, this is meant to be a "I don't want to deal with things, just do it for me" kind of program, so it seems like a better fit for this kind of instruction.

managed-token/program/tests/test.rs
```diff
@@ -141,3 +172,116 @@ async fn test_spl_managed_token_basic() {
...
+        1,
+    )
+    .unwrap();
+    process_transaction(
+        lwc,
+        vec![create_eve_ata_ix, successful_transfer_ix],
+        vec![&bob, &authority],
+    )
+    .await
+    .unwrap();
```
joncinque, line 249: Could you check to make sure that the transfer was successful and that the delegate was set?

managed-token/program/tests/test.rs
```diff
@@ -141,3 +172,116 @@ async fn test_spl_managed_token_basic() {
...
+    let wrap_ix = create_wrap_instruction(
+        &mint_key,
+        &old_authority.pubkey(),
+        &old_authority.pubkey(),
+        &new_authority.pubkey(),
+    )
+    .unwrap();
+    process_transaction(lwc, vec![wrap_ix], vec![&old_authority])
+        .await
+        .unwrap();
```
joncinque, line 286: Could you check that the mint has the proper authorities now?

joncinque (commented):
Apologies for the lateness on this -- great work! A few little things, then this should be good to go.  Let me know what you think.  Thanks for your contribution

managed-token/program/src/lib.rs
nothing0012: agreed, can remove it for now to minimize the change.

managed-token/program/tests/test.rs
nothing0012: removed the wrap entrypoint

joncinque:
Great, thanks for your contribution!

nothing0012:
@joncinque thank you! do you know when will the foundation deploy the program? previously didn't see the program in devnet or mainnet

joncinque:
@nothing0012 there's no intention of having a foundation deployment of this program right now, so it's meant to be a reference implementation for people who want to manage tokens easily.

This may change if people use the program as a unified layer for token royalties, but those talks are still in very early stages. If you need this soon, feel free to deploy your own version

Issue #3772, opened 2022-10-28
governance: unify final proposal state as `Completed` for proposal with no transaction inserted


ochaloup:
The governance proposal may be "locked" in state `Succeeded` and won't be moved to `Completed` if one does not insert any transaction to be executed while want to use deny options.

When having a proposal withoug the transactions to be executed while creating the proposal with `use_deny_option = false` then the proposal finishes in `Completed` state after `FinalizeVote` ix. 
When there is the `use_deny_option = true` then the proposal ends in `Succeeded` state. As there is no transaction to execute it's unclear to call `ExecuteTransaction` wish some `null` data (if it's even possible).
It should be clear and easy to move such a proposal to final state `Completed`.

The proposal finalize vote use deny vs. not use deny:
https://github.com/solana-labs/solana-program-library/blob/aaaf537be3555267a35e08130f911d0960cb8d2c/governance/program/src/state/proposal.rs#L415
https://github.com/solana-labs/solana-program-library/blob/aaaf537be3555267a35e08130f911d0960cb8d2c/governance/program/src/state/proposal.rs#L395

Current behaviour of `Succeeded` to `Completed` state change:

* `useDenyOption == false`; no transaction can be executed and the `FinalizeVote` switches to end state `ProposalState::Completed`
* `useDenyOption == true` and no transactions; the `FinalizeVote` switches to `ProposalState::Succeeded` and it's not possible (or unclear how) to move to `Completed`
* `useDenyOption == true` + transactions; the `FinalizeVote` switches to `ProposalState::Succeeded` and then `ExecuteTransaction` will run stored transactions and at the end moved to `ProposalState::Completed`

-----

As per discussion with @SebastianBor, see at
https://discord.com/channels/910194960941338677/910566058740568094/1035192118349668392
It could be a solution to add a new instruction `CompleteProposal`. If conditions are met then it would transition a proposal to a `Completed` state. It would work in the following use cases
1) It would allow to transition past proposal which are stuck in Succeeded state 
2) It would allow to transition non executable proposals (ExecutingWithErrors)  into Completed state after let's say 1 year of being in that state to make sure they are non actionable any longer

PR #3826, opened 2022-11-18
Governance: complete proposal instruction

ochaloup:
#### Summary

Make it possible to transition off-chain/manually executable `Proposal` from `Succeeded` to `Completed` state.

### DIFF
--- a/governance/program/tests/process_complete_proposal.rs
+++ b/governance/program/tests/process_complete_proposal.rs
@@ -175,3 +175,55 @@ async fn test_complete_proposal_with_completed_state_transaction_exists_error() 
 
     assert_eq!(err, GovernanceError::InvalidStateToCompleteProposal.into());
 }
+
+#[tokio::test]
+async fn test_complete_proposal_with_owner_or_delegate_must_sign_error() {
+    // Arrange
+    let mut governance_test = GovernanceProgramTest::start_new().await;
+
+    let realm_cookie = governance_test.with_realm().await;
+    let governed_account_cookie = governance_test.with_governed_account().await;
+
+    let mut token_owner_record_cookie = governance_test
+        .with_community_token_deposit(&realm_cookie)
+        .await
+        .unwrap();
+
+    let mut governance_cookie = governance_test
+        .with_governance(
+            &realm_cookie,
+            &governed_account_cookie,
+            &token_owner_record_cookie,
+        )
+        .await
+        .unwrap();
+
+    let mut proposal_cookie = governance_test
+        .with_signed_off_proposal(&token_owner_record_cookie, &mut governance_cookie)
+        .await
+        .unwrap();
+
+    governance_test
+        .with_cast_yes_no_vote(&proposal_cookie, &token_owner_record_cookie, YesNoVote::Yes)
+        .await
+        .unwrap();
+
+    // Try to maliciously sign using different owner signature
+    let token_owner_record_cookie2 = governance_test
+        .with_council_token_deposit(&realm_cookie)
+        .await
+        .unwrap();
+    token_owner_record_cookie.token_owner = token_owner_record_cookie2.token_owner;
+
+    // Act
+    let err = governance_test
+        .complete_proposal(&mut proposal_cookie, &token_owner_record_cookie)
+        .await
+        .err()
+        .unwrap();
+
+    assert_eq!(
+        err,
+        GovernanceError::GoverningTokenOwnerOrDelegateMustSign.into()
+    );
+}
--- a/governance/program/tests/program_test/mod.rs
+++ b/governance/program/tests/program_test/mod.rs
@@ -3152,33 +3152,20 @@ impl GovernanceProgramTest {
         proposal_cookie: &mut ProposalCookie,
         token_owner_record_cookie: &TokenOwnerRecordCookie,
     ) -> Result<(), ProgramError> {
-        self.complete_proposal_using_instruction(
-            proposal_cookie,
-            token_owner_record_cookie,
-            NopOverride,
-        )
-        .await
-    }
+        let complete_proposal_authority = token_owner_record_cookie.get_governance_authority();
 
-    #[allow(dead_code)]
-    pub async fn complete_proposal_using_instruction<F: Fn(&mut Instruction)>(
-        &mut self,
-        proposal_cookie: &mut ProposalCookie,
-        token_owner_record_cookie: &TokenOwnerRecordCookie,
-        instruction_override: F,
-    ) -> Result<(), ProgramError> {
-        let complete_authority = token_owner_record_cookie.get_governance_authority();
-
-        let mut complete_proposal_ix = complete_proposal(
+        let complete_proposal_ix = complete_proposal(
             &self.program_id,
             &proposal_cookie.address,
             &proposal_cookie.account.token_owner_record,
-            &complete_authority.pubkey(),
+            &complete_proposal_authority.pubkey(),
         );
-        instruction_override(&mut complete_proposal_ix);
 
         self.bench
-            .process_transaction(&[complete_proposal_ix], Some(&[complete_authority]))
+            .process_transaction(
+                &[complete_proposal_ix],
+                Some(&[complete_proposal_authority]),
+            )
             .await?;
 
         Ok(())


governance/program/tests/process_complete_proposal.rs
```diff
@@ -0,0 +1,74 @@
...
+    let mut governance_cookie = governance_test
+        .with_governance(
+            &realm_cookie,
+            &governed_account_cookie,
+            &token_owner_record_cookie,
+        )
+        .await
+        .unwrap();
+
+    let mut proposal_cookie = governance_test
```
SebastianBor, line 32: nit: There is a helper function `with_signed_off_proposal` which combines `Create`, `AddSignatory` and `SingOff` instructions

governance/program/tests/process_complete_proposal.rs
```diff
@@ -0,0 +1,74 @@
...
+
+mod program_test;
+
+use solana_program_test::tokio;
+
+use program_test::*;
+use spl_governance::state::enums::ProposalState;
+
+#[tokio::test]
+async fn test_execute_complete_proposal_without_transaction() {
```
SebastianBor, line 11: nit: Since it's the basic use case for the instruction it should be named `test_complete_proposal`. This way it tells you which test to troubleshoot first in case of any problems with the instruction

governance/program/src/processor/process_complete_proposal.rs
```diff
@@ -0,0 +1,54 @@
...
+    if proposal_data.state == ProposalState::ExecutingWithErrors
+        && proposal_data.voting_completed_at.is_some()
+        && proposal_data.voting_completed_at.unwrap()
+            + TIME_FOR_COMPLETE_PROPOSAL_WITH_ERRORS_SECONDS
+            < clock.unix_timestamp
+    {
+        proposal_data.closed_at = Some(clock.unix_timestamp);
+        proposal_data.state = ProposalState::Completed;
+    }
+
```
SebastianBor, line 51: nit: It's not going to change semantics in the current implementation but let's throw an error if the `Proposal` can't be completed (plus test)

governance/program/tests/program_test/mod.rs
```diff
@@ -3143,4 +3143,42 @@ impl GovernanceProgramTest {
 
         Ok(max_voter_weight_record_cookie)
     }
+
+    #[allow(dead_code)]
+    pub async fn with_complete_proposal(
```
SebastianBor, line 3148: nit: The function doesn't return any cookie (it's imperative and doesn't create any account state) and hence should be named `complete_proposal`

governance/program/tests/program_test/mod.rs
```diff
@@ -3143,4 +3143,42 @@ impl GovernanceProgramTest {
...
+        self.with_complete_proposal_using_instruction_impl(
+            governance_cookie,
+            proposal_cookie,
+            NopOverride,
+        )
+        .await
+    }
+
+    #[allow(dead_code)]
+    pub async fn with_complete_proposal_using_instruction_impl<F: Fn(&mut Instruction)>(
```
SebastianBor, line 3162: nit: There is no need for the `_impl` suffix which is used for internal implementations

governance/program/tests/program_test/mod.rs
```diff
@@ -3143,4 +3143,42 @@ impl GovernanceProgramTest {
...
+            &governance_cookie.address,
+            &proposal_cookie.address,
+        );
+        instruction_override(&mut complete_proposal_ix);
+
+        self.bench
+            .process_transaction(&[complete_proposal_ix], None)
+            .await?;
+
+        let proposal_data = self.get_proposal_account(&proposal_cookie.address).await;
```
SebastianBor, line 3179: Do not use actual state in the test sdk functions. The idea is that they should return what is expected and then you can compare, and assert, with the actual state in the test

SebastianBor (commented):
Mostly looks good and I only added nits regarding naming and semantic conventions used in the project.

Let's also scope this to just completed `Succeeded` proposals and hence also fix the state transition in `FinaliseVote`

ochaloup:
@SebastianBor thank you for the points, I hope I addressed them (expecting the result code could be squashed by github here thus I added a separate commit into this PR)

governance/program/src/processor/process_complete_proposal.rs
```diff
@@ -0,0 +1,54 @@
...
+            .options
+            .iter()
+            .all(|o| o.transactions_count == 0)
+    {
+        proposal_data.closed_at = Some(clock.unix_timestamp);
+        proposal_data.state = ProposalState::Completed;
+    }
+
+    // proposal execution still finishing with errors and time elapsed -> Completed
+    if proposal_data.state == ProposalState::ExecutingWithErrors
```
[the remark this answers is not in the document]
ochaloup: I would create a separate PR when this one is completed.

SebastianBor (approved):
Looks good

governance/program/src/instruction.rs
```diff
@@ -1660,3 +1666,21 @@ pub fn refund_proposal_deposit(
         data: instruction.try_to_vec().unwrap(),
     }
 }
+
+/// Complete proposal when proposal is stuck in Succeeded
+#[allow(clippy::too_many_arguments)]
```
joncinque, line 1671: nit: this can be removed, 2 arguments is definitely not too many :smile:

governance/program/src/instruction.rs
```diff
@@ -523,6 +523,12 @@ pub enum GovernanceInstruction {
     ///   1. `[writable]` ProposalDeposit account. PDA seeds: ['proposal-deposit', proposal, deposit payer]
     ///   2. `[writable]` Proposal deposit payer (beneficiary) account
     RefundProposalDeposit {},
+
+    /// Verify the state and move the Proposal to Completed.
+    /// Permission-less operation whoever could call.
+    ///
+    ///   0. `[writable]` Proposal account.
+    CompleteProposal {},
```
joncinque, line 531: Nit: can you add more comments about when this is used? For example, specifying "this instruction only works for `Succeeded` proposals that have no instructions".

And perhaps the name can be changed to `CompleteEmptySuccessfulProposal` or some other name that describes the intention more.

You're probably more aware of the terms used in the space, so I'll defer to you!

joncinque (commented):
The logic looks great, just a question about the terms for this

governance/program/src/instruction.rs
ochaloup: @joncinque thanks. I elaborated a bit more on function of the instruction in comment.
I tend to use the generic `CompleteProposal` name but I would like to ask @SebastianBor  for opinion on this.
This instruction should  (or could) have two parts. This PR resolves only one of them. I created a separate issue to make it clear (my wrong not doing so before).
The one that this PR resolves is
https://github.com/solana-labs/solana-program-library/issues/3772
The other (a potentially future PR) is:
https://github.com/solana-labs/solana-program-library/issues/4047

joncinque: Ah that's great, thanks for clarifying!  Since all of the other comments are in the enum description, could you move your comments on `complete_proposal` up to here? And if it were me, I'd rename it to `CompleteStuckProposal`, but I'll let you figure that out :smile:

joncinque (dismissed):

governance/program/src/instruction.rs
ochaloup: Thank you @joncinque .

I fixed the place where the documentation comment resides at + I renamed the instruction to `CompleteStuckProposal` as you suggested.
If you find time to check it would be great.

joncinque (approved):
Looks good to me! I'll let @SebastianBor merge it if it looks good to him

SebastianBor:
After further review of the logic I've made changes to the semantics  of  `CompleteProposal` instruction because it's not abnormal situation (it's not stuck) to have an instruction-less and executable Proposal in the `Succeeded` state.

It's a valid scenario where we have a binding proposal (as opposed to opinion only) but the actions of the proposal are either off-chain or have to be executed manually. In such a scenario `CompleteProposal` allows the owner of the `Proposal` to flag it as `Completed` once the external actions are taken. 

@joncinque please give it another look given the changes in the semantics

PR #3828, opened 2022-11-18
Governance: Remove stats only fields to make room for more Config values

SebastianBor:
#### Summary

Remove stats only field `voting_proposal_count` from `Governance` account to make room for more `RealmConfig` values without breaking binary layout compatibility with previous versions.

### DIFF
--- a/governance/program/src/processor/process_sign_off_proposal.rs
+++ b/governance/program/src/processor/process_sign_off_proposal.rs
@@ -29,7 +29,7 @@ pub fn process_sign_off_proposal(program_id: &Pubkey, accounts: &[AccountInfo]) 
 
     let mut realm_data = get_realm_data(program_id, realm_info)?;
 
-    let mut governance_data =
+    let _governance_data =
         get_governance_data_for_realm(program_id, governance_info, realm_info.key)?;
 
     let mut proposal_data =
@@ -88,12 +88,6 @@ pub fn process_sign_off_proposal(program_id: &Pubkey, accounts: &[AccountInfo]) 
 
     realm_data.voting_proposal_count = realm_data.voting_proposal_count.checked_add(1).unwrap();
     realm_data.serialize(&mut *realm_info.data.borrow_mut())?;
-
-    governance_data.voting_proposal_count = governance_data
-        .voting_proposal_count
-        .checked_add(1)
-        .unwrap();
-    governance_data.serialize(&mut *governance_info.data.borrow_mut())?;
 
     Ok(())
 }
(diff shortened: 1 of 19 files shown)


governance/program/src/processor/process_sign_off_proposal.rs
```diff
@@ -29,7 +29,7 @@ pub fn process_sign_off_proposal(program_id: &Pubkey, accounts: &[AccountInfo])
 
     let mut realm_data = get_realm_data(program_id, realm_info)?;
 
-    let mut governance_data =
+    let _governance_data =
```
mschneider, line 32: why even load when it's never read?

governance/program/src/state/legacy.rs
```diff
@@ -125,14 +125,6 @@ pub struct GovernanceV1 {
...
     /// Governance config
     pub config: GovernanceConfig,
-
-    /// Reserved space for future versions
-    pub reserved: [u8; 6],
-
-    /// The number of proposals in voting state in the Governance
-    /// Note: This is field introduced in V2 but it took space from reserved
-    /// and we have preserve it for V1 serialization roundtrip
-    pub voting_proposal_count: u16,
```
joncinque, line 135: Why are you removing the reserve entirely too?

governance/program/src/processor/process_sign_off_proposal.rs
SebastianBor: It's still needed to enforce the `Realm` -> `Governance` -> `Proposal` relationship. It could be replaced by PDA check but the `Governance`  account is going to be needed in future versions once we support mandatory signatories.
I added comments to explain it 87d15129a07bcbc773ae7a43157df2ac683e4ca0

governance/program/src/state/legacy.rs
SebastianBor: It was a bug that it was still there because all the reserved space was moved to `GovernanceConfig` in previous PRs for `V3`. It escaped me before and this is why I added the account size test as a guard https://github.com/solana-labs/solana-program-library/blob/e0dd05fa5caf929e8910346ac4bb6a0841ddbfdf/governance/program/src/state/governance.rs#L557

I should find a better way to test for those legacy accounts translations, but it might not be needed any longer because now we can migrate all the accounts on the fly to the latest version using `reallocs`

mschneider: it would be great to have basic unit tests for v1->v2 transitions for all concerned account types

PR #3808, opened 2022-11-10
Governance: Extend Veto vote period by min_transaction_hold_up_time

SebastianBor:

PR #3830, opened 2022-11-18
Governance: voting_cool_off_time

SebastianBor:
#### Summary

Introduce voting cool of time to add an extra time when `Community` and/or `Council` can disapprove or veto proposals.

#### Implementation Details
- All existing Governances with `max_voting_time`  > `10h` default to `1h` cool of time. The setting can be further adjusted or removed via config change proposal. 
- Cool of time doesn't apply when a vote ends early with `Early` or `Strict` tipping rules.

### DIFF
--- a/governance/chat/program/tests/program_test/mod.rs
+++ b/governance/chat/program/tests/program_test/mod.rs
@@ -189,7 +189,7 @@ impl GovernanceChatProgramTest {
             community_vote_threshold: VoteThreshold::YesVotePercentage(60),
             min_community_weight_to_create_proposal: 5,
             min_transaction_hold_up_time: 10,
-            base_voting_time: 10,
+            voting_base_time: 10,
             community_vote_tipping: spl_governance::state::enums::VoteTipping::Strict,
             council_vote_threshold: VoteThreshold::YesVotePercentage(10),
             council_veto_vote_threshold: VoteThreshold::YesVotePercentage(50),
--- a/governance/chat/program/tests/program_test/mod.rs
+++ b/governance/chat/program/tests/program_test/mod.rs
@@ -189,7 +189,7 @@ impl GovernanceChatProgramTest {
             community_vote_threshold: VoteThreshold::YesVotePercentage(60),
             min_community_weight_to_create_proposal: 5,
             min_transaction_hold_up_time: 10,
-            max_voting_time: 10,
+            base_voting_time: 10,
             community_vote_tipping: spl_governance::state::enums::VoteTipping::Strict,
             council_vote_threshold: VoteThreshold::YesVotePercentage(10),
             council_veto_vote_threshold: VoteThreshold::YesVotePercentage(50),
--- a/governance/chat/program/tests/program_test/mod.rs
+++ b/governance/chat/program/tests/program_test/mod.rs
@@ -196,7 +196,8 @@ impl GovernanceChatProgramTest {
             council_veto_vote_threshold: VoteThreshold::YesVotePercentage(50),
             council_vote_tipping: spl_governance::state::enums::VoteTipping::Strict,
             community_veto_vote_threshold: VoteThreshold::YesVotePercentage(55),
-            reserved: [0; 5],
+            reserved: 0,
+            voting_cool_off_time: 1,
         };
 
         let token_owner_record_address = get_token_owner_record_address(
--- a/governance/chat/program/tests/program_test/mod.rs
+++ b/governance/chat/program/tests/program_test/mod.rs
@@ -186,18 +186,18 @@ impl GovernanceChatProgramTest {
         let governed_account_address = Pubkey::new_unique();
 
         let governance_config = GovernanceConfig {
+            community_vote_threshold: VoteThreshold::YesVotePercentage(60),
             min_community_weight_to_create_proposal: 5,
-            min_council_weight_to_create_proposal: 2,
             min_transaction_hold_up_time: 10,
             max_voting_time: 10,
-            community_vote_threshold: VoteThreshold::YesVotePercentage(60),
             community_vote_tipping: spl_governance::state::enums::VoteTipping::Strict,
             council_vote_threshold: VoteThreshold::YesVotePercentage(10),
             council_veto_vote_threshold: VoteThreshold::YesVotePercentage(50),
+            min_council_weight_to_create_proposal: 2,
             council_vote_tipping: spl_governance::state::enums::VoteTipping::Strict,
             community_veto_vote_threshold: VoteThreshold::YesVotePercentage(55),
-            reserved: 0,
             voting_cool_off_time: 1,
+            reserved: 0,
         };
 
         let token_owner_record_address = get_token_owner_record_address(
--- a/governance/program/src/state/governance.rs
+++ b/governance/program/src/state/governance.rs
@@ -36,7 +36,7 @@ pub struct GovernanceConfig {
     /// The base voting time in seconds for proposal to be open for voting
     /// Voting is unrestricted during the base voting time and any vote types can be cast
     /// The base voting time can be extend by optional cool off time when only negative votes (Veto and Deny) are allowed
-    pub base_voting_time: u32,
+    pub voting_base_time: u32,
 
     /// Conditions under which a Community vote will complete early
     pub community_vote_tipping: VoteTipping,
@@ -520,7 +520,7 @@ mod test {
             community_vote_threshold: VoteThreshold::YesVotePercentage(60),
             min_community_weight_to_create_proposal: 5,
             min_transaction_hold_up_time: 10,
-            base_voting_time: 5,
+            voting_base_time: 5,
             community_vote_tipping: VoteTipping::Strict,
             council_vote_threshold: VoteThreshold::YesVotePercentage(60),
             council_veto_vote_threshold: VoteThreshold::YesVotePercentage(50),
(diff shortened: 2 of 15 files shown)


governance/chat/program/tests/program_test/mod.rs
```diff
@@ -196,7 +196,8 @@ impl GovernanceChatProgramTest {
             council_veto_vote_threshold: VoteThreshold::YesVotePercentage(50),
             council_vote_tipping: spl_governance::state::enums::VoteTipping::Strict,
             community_veto_vote_threshold: VoteThreshold::YesVotePercentage(55),
-            reserved: [0; 5],
+            reserved: 0,
```
mschneider, line 199: it's a bit confusing that reserved comes before voting_cool_off_time

SebastianBor: Reordered

governance/program/src/state/proposal.rs
```diff
@@ -274,11 +275,25 @@ impl ProposalV2 {
             return Err(GovernanceError::ProposalVotingTimeExpired.into());
         }
 
-        Ok(())
+        match vote {
+            // Once in the voting cool off time approving votes are no longer accepted
+            // Abstain is considered as positive vote because when attendance quorum is used it can tip the scales
+            Vote::Approve(_) | Vote::Abstain => {
+                if self.expected_vote_end_time(config) - (config.voting_cool_off_time as i64)
```
mschneider, line 282: it would be easier to understand if max_vote_time would not include voting_cool_off_time. this can cause issues with configuration.

governance/program/tests/use_veto_vote.rs
```diff
@@ -962,3 +962,69 @@ async fn test_veto_vote_with_community_max_voter_weight_addin_and_veto_not_tippe
 
     assert_eq!(proposal_account.state, ProposalState::Voting);
 }
+
+#[tokio::test]
```
mschneider, line 966: missing tests for full coverage:
cast electoral fails in cool off
relinquish succeeds in cool off
finalize fails in cool off

governance/program/src/state/proposal.rs
SebastianBor: I think the only question here is  is whether 1) `voting_cool_off_time` should be added to the configured `max_voting_time` or 2) it should be defined as a period within it and the rest of the logic stays the same and code changes are comparable.

No matter how we define it the following must still hold: `voting_time` = `unrestricted_voting_time` + `cool_off_time` because during the `cool_off_time` all the voting invariants are preserved and we only restrict what type of votes can be cast:
- Relinquishing vote means withdrawing your vote while once voting ends it only releases your lockup
- The above implies withdrawing your vote is allowed while once voting ends it's not
- Casting votes, even if restricted, is allowed while it's not once voting ends
- Finalising is not allowed 
- Proposal cancellation is still allowed while it's not once voting ends
- Proposal still counts as active and consumes your active proposal quota 
- Proposal can't be executed before voting ends 

In terms of code changes, if we decide to add `voting_cool_off_time` to `max_voting_time`, the only change would be in `expected_vote_end_time` and the rest would stay more or less the same.
```
    pub fn expected_vote_end_time(&self, config: &GovernanceConfig) -> UnixTimestamp {
        self.voting_at
            .unwrap()
            .checked_add(config.max_voting_time + config.voting_cool_off_time as i64)
            .unwrap()
    }
```

Given the above I think keeping `max_voting_time` as is and defining `voting_cool_off_time` as a period within it seems more natural and easier to think about from user perspective. If you want the vote to last for 3 days then it does last for 3 days but voting is restricted towards the end of the voting time.

In terms of the infra changes the bot would have to be changed anyway because it should notify before `voting_cool_off_time` starts in case you want to `Approve` and then again before it ends in case you want `Deny` it.

This change adds `1h` `voting_cool_off_time` to all existing `Governances` with `max_voting_time` > `10h` because i think it's a good default and I'm planning to make the same in the UI but it doesn't restrict you from creating new Governances without it. 

I agree it can confuse users and some in flight proposals could fail if users were planning to vote in the last minute, but I think it's still better than asking all DAOs to change their config which could create an avalanche of config changes proposals. If the proposals which would fail had DAO support then they could be recreated and voted again anyway. The only issue I can think about would be time critical proposals which could cause problems.  

Alternatively we could change `voting_cool_off_time` for existing `Governances` to something even smaller, like 10m maybe or keep it unchanged? What do you think?

As an additional thought, I'm still thinking about extending the `Veto` power into the `hold_up_time`  https://github.com/solana-labs/solana-program-library/pull/3808 (now reverted) and then conceptually `voting_cool_off_time` is within `max_voting_time` and `hold_up_time` is after it.

governance/program/tests/use_veto_vote.rs
SebastianBor: The first 2 tests where there before and I added the 3rd 
1) cast electoral fails in cool off  https://github.com/solana-labs/solana-program-library/blob/ef697fedddda09c1202fe864fc18e31b9d88c4fb/governance/program/tests/process_cast_vote.rs#L1555
2) relinquish succeeds in cool off - it tests you can relinquish your `Yes` vote and then vote `No`  https://github.com/solana-labs/solana-program-library/blob/ef697fedddda09c1202fe864fc18e31b9d88c4fb/governance/program/tests/process_relinquish_vote.rs#L603
3) finalize fails in cool off https://github.com/solana-labs/solana-program-library/blob/1e61b65c68c87cc32b15409274575d7ed52b5ca7/governance/program/tests/process_finalize_vote.rs#L447

governance/program/src/state/governance.rs
```diff
@@ -309,8 +313,17 @@ pub fn get_governance_data(
...
-        governance_data.config.reserved = [0; 5];
+        // Reset voting_cool_off_time and reserved space previously used for voting_proposal_count
+
+        // Default voting_cool_off_time to 1h for max_voting_time >= 10h
+        governance_data.config.voting_cool_off_time =
+            if governance_data.config.max_voting_time >= 36000 {
+                3600
+            } else {
+                0
+            };
```
joncinque, line 324: Sorry, I still don't understand the reason for making a default change to all existing governances.  What's the problem with an avalanche of governances trying to make this change? I would be more worried about my proposal's voting time changing all of a sudden

governance/program/src/state/proposal.rs
joncinque: For what it's worth, I agree with Max. Having the cool-off as a segment within the voting time is more confusing than creating a separate segment for the cool-off period. And that way, you don't have to worry about the cool off time being greater than the total vote time.  If you're configuring a governance, and you want to extend just cool-off by 1 hour, you need to make *two* changes, one to the cool off period, and another to the voting time.  That doesn't seem intuitive to me.

joncinque (commented):
R+ after addressing Max's points

governance/program/src/state/proposal.rs
mschneider: I would prefer if hold-up would not change semantics and stay purely a time delay as it was before, we just have another state from now on and you should probably update the flow chart, so it's easy to understand for new devs.

draft -> signing off -> voting (vote, veto, cancel) -> cool-off (vote-down, veto, cancel) -> hold-up -> execute

governance/program/src/state/governance.rs
mschneider: mainly operational, a lot of DAO's would require support from labs in order to upgrade their realm

governance/program/src/state/proposal.rs
SebastianBor: I've changed the math to add `cool_off_time` to the configured `voting_time`,  but I also had to do some further renames for it all to make sense.

- `config.max_voting_time` renamed to `config.base_voting_time` - The configured time can't be called `max_voting_time` any longer because max voting time now equals to configured `voting_time` + `cool_off_time` end hence the `base_voting_time` name

- `proposal.vote_end_time()` renamed to `proposal.max_voting_end_time()`and it includes both `base_voting_time` and `cool_off_time`

- `proposal.has_vote_time_ended()` renamed to `proposal.has_max_voting_time_ended()`

governance/program/src/state/governance.rs
SebastianBor: My thinking was that it was a recommended setting and really all DAOs should use it to increase their security and without the default override most of them would never make the change. 

However given the potential confusion I backed out and removed the default and would increase awareness among the DAOs instead.

governance/program/src/state/proposal.rs
mschneider: i'm still in favor of making cool-off a first class phase and always address it directly, but if you need to manage it below the voting state, i would make a couple of changes to simplify the readability of the code:
`voting_total_time = config.voting_base_time + config.voting_cool_off_time`

governance/program/src/state/proposal.rs
```diff
@@ -264,35 +264,54 @@ impl ProposalV2 {
...
+        if self.has_max_voting_time_ended(config, current_unix_timestamp) {
             return Err(GovernanceError::ProposalVotingTimeExpired.into());
         }
 
-        Ok(())
+        match vote {
+            Vote::Approve(_) | Vote::Abstain => {
+                // Once the base voting time passes and we are in the voting cool off time approving votes are no longer accepted
+                // Abstain is considered as positive vote because when attendance quorum is used it can tip the scales
+                if self.base_voting_end_time(config) < current_unix_timestamp {
```
mschneider, line 282: nitpick: could use another `has_ended` fn

governance/program/src/state/proposal.rs
SebastianBor: I adjusted all the names related to voting time to make them consistent

- `voting_base_time`
- `voting_base_time_end`
- `has_voting_base_time_ended`
- `voting_cool_off_time`
- `voting_max_time_end`
- `has_voting_max_time_ended`

Note: 
- `voting_total_time` is called `voting_max_time` because it's the configured `max` voting time and not necessarily a `total` voting time  because voting can last less time based on the vote tipping config  
- We never explicitly check for end of `cool_off_time` and hence there is no need for `voting_cool_off_time_end` and `has_voting_cool_off_ended` functions

governance/program/src/state/proposal.rs
SebastianBor: Done: `has_voting_base_time_ended`

Issue #3776, opened 2022-10-28
token-2022: release


2501babe:
after #3725 lands, we should do releases for the token-2022 program, client, and cli, all together. need to coordinate to make sure the confidential transfer stuff is in an acceptable state for this. this is a prereq to updating the account decoder to support the cpi guard and permanent delegate extensions, which is a prereq to supporting them in `spl-token display`

2501babe:
tyera found a bug in token-cli [at least 2.0.16 and 2.0.17] that breaks closing accounts, which i inadvertently fixed in the conversion to token-client. so id like to proceed with a new cli release next week

cc @joncinque @samkim-crypto, would you say the token-2022 program is in a state where we could release now? perhaps after merging the permanent delegate feature? are there other issues/prs to close out first? id be happy to help with any ofc

the situation is monorepo crates that we depend on depend on an old token-2022, creating an unresolvable cycle. my goal is to:
* bump/release token-2022 plus the version spl depends on
* pr monorepo to bump the token-2022 version it depends on
* bump/release token-client and token-cli

joncinque:
It makes sense to do a release after merging the permanent delegate extension. We'll be having one more audit for the CpiGuard + permanent delegate, which likely means we'll need to do one more release after that.

2501babe:
im guessing this covers the "dependent crates" mentioned in #3807, and we can release token-upgrade-cli along with this too

joncinque:
Yes please, that would be perfect

2501babe:
token-2022, token-client, and token-cli are out. first and third have corresponding github releases and tags, second does not, in keeping with past practice (i presume we will change that with 1.0.0)

leaving this issue open for the upgrade program/cli (need to deal with the name issue) and also upgrading monorepo (so we can add the new extensions to the acocunt decoder)

2501babe:
alright i handled the ownership handover for token-upgrade and token-upgrade-cli and cut initial releases for both. also monorepo is on the new version of token-2022, so cli improvements that involve the account decoder can be done once the next 1.14 is out. the release process is over. the galaxy is at peace

PR #3831, opened 2022-11-19
token-2022: bump version to 0.5

2501babe:
as described in other issues, im going to start the (somewhat complicated) release process, which in total will look like:
* bump token22. cargo release
* change monorepo version of token22 to match
* bump versions of token cli, token client, and token upgrade cli. cargo release 

this is just for the first bullet. please lmk if theres any changes you want to get into token22 before this

part of #3776

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5814,7 +5814,7 @@ dependencies = [
  "num-traits",
  "solana-program",
  "spl-token 3.5.0",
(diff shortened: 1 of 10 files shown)


samkim-crypto (approved):
Thanks! Looks good to me

joncinque (approved):

PR #3838, opened 2022-11-21
bump token-cli to 2.2 and token-client to 0.3

2501babe:
i removed the patch version locking since really we shouldnt need it internal to the repo


2501babe:
oh note that due to a previous release attempt, monorepo version 1.10 and 1.13 are locked on cli 2.0.17, so this doesnt affect them. test built successfully on 1.14.7 and 1.15.0

joncinque (approved):

PR #3839, opened 2022-11-21
stake-pool: Add tolerance for stake accounts at minimum

joncinque:
#### Problem

As the value of stake pool tokens increase, it could become impossible to reduce a stake account to its minimum value due to rounding.

#### Solution

This is a very tricky problem, but to give at least some cushion, round up to the lamports-per-pool-token value.

While working on this, I refactored the `BigVec` search to go through a closure rather than a function pointer, and it reduced compute unit consumption so much that the max pool size has increased by 50%!

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -844,10 +844,9 @@ impl Processor {
         if header.max_validators == validator_list.len() {
             return Err(ProgramError::AccountDataTooSmall);
         }
-        let maybe_validator_stake_info = validator_list.find::<ValidatorStakeInfo>(
-            validator_vote_info.key.as_ref(),
-            ValidatorStakeInfo::memcmp_pubkey,
-        );
+        let maybe_validator_stake_info = validator_list.find::<ValidatorStakeInfo, _>(|x| {
+            ValidatorStakeInfo::memcmp_pubkey(x, validator_vote_info.key)
+        });
         if maybe_validator_stake_info.is_some() {
             return Err(StakePoolError::ValidatorAlreadyAdded.into());
         }
@@ -994,10 +993,9 @@ impl Processor {
 
         let (meta, stake) = get_stake_state(stake_account_info)?;
         let vote_account_address = stake.delegation.voter_pubkey;
-        let maybe_validator_stake_info = validator_list.find_mut::<ValidatorStakeInfo>(
-            vote_account_address.as_ref(),
-            ValidatorStakeInfo::memcmp_pubkey,
-        );
+        let maybe_validator_stake_info = validator_list.find_mut::<ValidatorStakeInfo, _>(|x| {
+            ValidatorStakeInfo::memcmp_pubkey(x, &vote_account_address)
+        });
         if maybe_validator_stake_info.is_none() {
             msg!(
                 "Vote account {} not found in stake pool",
@@ -1154,10 +1152,9 @@ impl Processor {
         let (meta, stake) = get_stake_state(validator_stake_account_info)?;
         let vote_account_address = stake.delegation.voter_pubkey;
 
-        let maybe_validator_stake_info = validator_list.find_mut::<ValidatorStakeInfo>(
-            vote_account_address.as_ref(),
-            ValidatorStakeInfo::memcmp_pubkey,
-        );
+        let maybe_validator_stake_info = validator_list.find_mut::<ValidatorStakeInfo, _>(|x| {
+            ValidatorStakeInfo::memcmp_pubkey(x, &vote_account_address)
+        });
         if maybe_validator_stake_info.is_none() {
             msg!(
                 "Vote account {} not found in stake pool",
@@ -1316,10 +1313,9 @@ impl Processor {
 
         let vote_account_address = validator_vote_account_info.key;
 
-        let maybe_validator_stake_info = validator_list.find_mut::<ValidatorStakeInfo>(
-            vote_account_address.as_ref(),
-            ValidatorStakeInfo::memcmp_pubkey,
-        );
+        let maybe_validator_stake_info = validator_list.find_mut::<ValidatorStakeInfo, _>(|x| {
+            ValidatorStakeInfo::memcmp_pubkey(x, vote_account_address)
+        });
         if maybe_validator_stake_info.is_none() {
             msg!(
                 "Vote account {} not found in stake pool",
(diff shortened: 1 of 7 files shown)


bennofs (approved):

bennofs:
While doing the final review on this, I found one more case that is not handled yet:

- there is an active, non-preferred validator with much stake
- the preferred validator has only slightly more stake than min (less than lamports per pool token more)

Then the code forces you to withdraw from the preferred validator, which is impossible since even burning 1 token is too much to withdraw from that validator. All the other withdraw options are also not available because we still have the other, non-preferred validator forcing us to withdraw from an active validator.

bennofs:
Here is a test case which fails and demonstrates the issue:

```rust
#[tokio::test]
async fn success_with_small_preferred_withdraw() {
    let (
        mut context,
        stake_pool_accounts,
        validator_stake,
        deposit_info,
        user_transfer_authority,
        user_stake_recipient,
        tokens_to_burn,
    ) = setup_for_withdraw(spl_token::id()).await;

    // make pool tokens very valuable, so it isn't possible to exactly get down to the minimum
    transfer(
        &mut context.banks_client,
        &context.payer,
        &context.last_blockhash,
        &stake_pool_accounts.reserve_stake.pubkey(),
        deposit_info.stake_lamports * 5, // each pool token is worth more than one lamport
    )
    .await;
    stake_pool_accounts
        .update_all(
            &mut context.banks_client,
            &context.payer,
            &context.last_blockhash,
            &[validator_stake.vote.pubkey()],
            false,
        )
        .await;

    let preferred_validator = simple_add_validator_to_pool(
        &mut context.banks_client,
        &context.payer,
        &context.last_blockhash,
        &stake_pool_accounts,
        None,
    )
    .await;

    stake_pool_accounts
        .set_preferred_validator(
            &mut context.banks_client,
            &context.payer,
            &context.last_blockhash,
            instruction::PreferredValidatorType::Withdraw,
            Some(preferred_validator.vote.pubkey()),
        )
        .await;

    println!("{}",  get_account(&mut context.banks_client, &preferred_validator.stake_account).await.lamports);


    // add a tiny bit of stake, less than lamports per pool token to preferred validator
    let preferred_deposit = simple_deposit_stake(
        &mut context.banks_client,
        &context.payer,
        &context.last_blockhash,
        &stake_pool_accounts,
        &preferred_validator,
        2_000_000_000 + 1,
    )
    .await
    .unwrap();

    println!("{}",  get_account(&mut context.banks_client, &preferred_validator.stake_account).await.lamports);
    let rent = context.banks_client.get_rent().await.unwrap();
    let rent_exempt = rent.minimum_balance(std::mem::size_of::<stake::state::StakeState>());
    // decrease a little stake, not all
    let error = stake_pool_accounts
        .decrease_validator_stake(
            &mut context.banks_client,
            &context.payer,
            &context.last_blockhash,
            &preferred_validator.stake_account,
            &preferred_validator.transient_stake_account,
            2_000_000_000 + rent_exempt,
            preferred_validator.transient_stake_seed,
        )
        .await;
    assert!(error.is_none());   
    println!("{}",  get_account(&mut context.banks_client, &preferred_validator.stake_account).await.lamports);

    // warp forward to deactivation
    let first_normal_slot = context.genesis_config().epoch_schedule.first_normal_slot;
    let slots_per_epoch = context.genesis_config().epoch_schedule.slots_per_epoch;
    context
        .warp_to_slot(first_normal_slot + slots_per_epoch)
        .unwrap();

    // update to merge deactivated stake into reserve
    stake_pool_accounts
        .update_all(
            &mut context.banks_client,
            &context.payer,
            &context.last_blockhash,
            &[validator_stake.vote.pubkey(), preferred_validator.vote.pubkey()],
            false,
        )
        .await;

    // withdraw from preferred fails
    let new_authority = Pubkey::new_unique();
    let error = stake_pool_accounts
        .withdraw_stake(
            &mut context.banks_client,
            &context.payer,
            &context.last_blockhash,
            &user_stake_recipient.pubkey(),
            &user_transfer_authority,
            &deposit_info.pool_account.pubkey(),
            &preferred_validator.stake_account,
            &new_authority,
            1,
        )
        .await;
    assert!(error.is_some());

    // preferred is empty, withdrawing from non-preferred works
    let new_authority = Pubkey::new_unique();
    let error = stake_pool_accounts
        .withdraw_stake(
            &mut context.banks_client,
            &context.payer,
            &context.last_blockhash,
            &user_stake_recipient.pubkey(),
            &user_transfer_authority,
            &deposit_info.pool_account.pubkey(),
            &validator_stake.stake_account,
            &new_authority,
            tokens_to_burn / 6,
        )
        .await;
    assert!(error.is_none());
}
```

stake-pool/program/src/processor.rs
```diff
@@ -2548,11 +2550,37 @@ impl Processor {
...
+                    // For example, if the pool token is extremely valuable, it might only
+                    // be possible to reduce the stake account to a minimum of
+                    // `stake_rent + minimum_delegation + lamports_per_pool_token - 1`.
+                    //
+                    // After that, the minimum amount of pool tokens to get to this amount
+                    // may actually be worth
+                    // `stake_rent + minimum_delegation + lamports_per_pool_token * 2 - 2`.
+                    // We give an extra grace on this check of two lamports, which should be
+                    // reasonable. At worst, it just means that a withdrawer is losing out
+                    // on two lamports.
```
bennofs, line 2568: I am curious, I think there is something I am missing here. We are comparing against the actual lamports in the stake account here, so a tolerance of +lamports_per_token should always be enough. 

For example, let's assume we only reduce the stake account to minimum_stake_lamports + (lamports_per_token - 1). But then `split_from_lamports = minimum_stake_lamports +(lamports_per_token -  1)` already, and we can always find a token burn amount so that we are within +lamports_per_token of that. No need to double, as far as I understand.

bennofs (changes requested):

stake-pool/program/src/processor.rs
joncinque: You're right, I just wrote the test poorly during my first pass. No need for the limit, thanks for taking the time to notice this!

joncinque:
> While doing the final review on this, I found one more case that is not handled yet:

Thanks for noticing this and providing the test case, I've put in the test and the fix.

With that, everything's been addressed from your most recent review.

bennofs (approved):
Looks good to me now

PR #3846, opened 2022-11-25
Governance: add options in chunks (new instruction AddProposalOptions)

ochaloup:
Making possible to add options not only within the single instruction at time of proposal creation (`CreateProposal`) while making possible to add options in similar way as transactions are inserted before sign-off.

This PR consists of possibility to create a proposal without any option and adding new instruction `AddProposalOptions` that resizes the account size to fit the newly added options within.

### DIFF
--- a/governance/program/src/instruction.rs
+++ b/governance/program/src/instruction.rs
@@ -523,6 +523,19 @@ pub enum GovernanceInstruction {
     ///   1. `[writable]` ProposalDeposit account. PDA seeds: ['proposal-deposit', proposal, deposit payer]
     ///   2. `[writable]` Proposal deposit payer (beneficiary) account
     RefundProposalDeposit {},
+
+    /// Insert Options to proposal
+    ///
+    ///   0. `[writable]` Proposal account.
+    ///   1. `[]` TokenOwnerRecord account of the Proposal owner
+    ///   2. `[signer]` Governance Authority (Token Owner or Governance Delegate)
+    ///   3. `[writable, signer]` Payer
+    ///   4. `[]` System
+    InsertProposalOptions {
+        /// Options to be added into proposal account
+        #[allow(dead_code)]
+        options: Vec<String>,
+    },
 }
 
 /// Creates CreateRealm instruction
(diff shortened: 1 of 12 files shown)


governance/program/src/instruction.rs
```diff
@@ -204,6 +204,13 @@ pub enum GovernanceInstruction {
         #[allow(dead_code)]
         /// Unique seed for the Proposal PDA
         proposal_seed: Pubkey,
+
+        #[allow(dead_code)]
+        /// Indicates whether space for the proposal account will be taken from data provided through the call
+        /// (when None) or the space for the account will be pre-defined by some amount
+        /// Prefetching the space can be used for adding options into the proposal in separate instruction when in draft state
+        /// Value 0 means no prefetching to happen.
+        prefetch_space: u64,
```
SebastianBor, line 213: Since `solana` runtime now supports resizing of the accounts I think it'll be better to resize the account as we add new options instead of allocating it upfront

SebastianBor (commented):
I'm wondering if this actually solves the problem. 

- What is the maximum number of options we can add this way before we start hitting compute limits when deserialising the account using `Borsh` as it is done now? 
- What is the limit and cost of `VoteRecord` which stores the vote results as an array for all options? Would we also have to implement voting using chunks? Or should we use a sparse array instead?


I'm still debating with myself whether we should try to implement gauges on top of the existing proposal implementation or we are stretching it too much and adding too much complexity and maybe instead we should implement a new proposal type. 

Do you have all the requirements for gauges is some doc I could use to evaluate the possible solutions?

ochaloup:
I did the update of PR. Can you please @SebastianBor review?

* I removed the `prefetch_space` and the size of `proposal` account is resized at time of inserting the option in as chunk
* as the resize of the account is limited by 10KB in Solana it's needed to be done in chunks either way. When tested the in our use case when one option label has about 50 characters limit of 10KB of placing all data at once on creation comes ~95 of options
* The current voting limits may limit us as well. I think it would be good enough to have the sparse array voting. Do you think this would require a new instruction or it could be implemented as part of the the `process_cast_vote` and adding a new type of `Vote` under `enum (like `VoteChoiceIndexed`)? Or how to do the best way?

governance/program/src/processor/process_insert_proposal_options.rs
```diff
@@ -0,0 +1,75 @@
...
+            vote_weight: 0,
+            vote_result: OptionVoteResult::None,
+            transactions_executed_count: 0,
+            transactions_count: 0,
+            transactions_next_index: 0,
+        };
+        proposal_data.options.push(po);
+    }
+
+    let proposal_new_data_size = if let Some(proposal_data_size) = proposal_data.get_max_size() {
```
SebastianBor, line 57: Do we need this check? If we add options it's always going to be a bigger size, no?

governance/program/src/state/proposal.rs
```diff
@@ -1078,13 +1093,22 @@ pub fn get_proposal_address<'a>(
...
+pub fn assert_valid_proposal_options(options: &[String]) -> Result<(), ProgramError> {
+    if options.iter().any(|o| o.is_empty()) {
+        return Err(GovernanceError::InvalidProposalOptions.into());
+    }
+
+    Ok(())
+}
+
+/// Assert options to assign proposal are valid for Proposal vote_type
+pub fn assert_valid_proposal_options_on_signoff(
```
SebastianBor, line 1106: is this still needed if we check the options in every `AddOptions` instruction?

governance/program/src/processor/process_insert_proposal_options.rs
ochaloup: The thing could be that no option is provided in the instruction data. It should be probably rather reported as an error when that occurs. I changed it to return error, could be that way?

governance/program/src/state/proposal.rs
ochaloup: The check is split to two parts now.
There is a check if option description is empty for `AddProposalOptions`  ix and then a check if the list of options are empty during sign off. To me it makes sense to have that separation but we can change it.

ochaloup:
@SebastianBor Thanks. I updated the PR based on the review points.

PR #3847, opened 2022-11-25
Governance: attendance quorum implementation + survey type proposals

ochaloup:
The PR implements `AttendanceQuorum` a new `VoteThreshold` type + providing updates on survey type proposals (proposals that permit no deny votes and contain no transaction).

### DIFF
--- a/governance/program/src/processor/process_set_governance_config.rs
+++ b/governance/program/src/processor/process_set_governance_config.rs
@@ -1,15 +1,16 @@
 //! Program state processor
 
+use crate::{
+    error::GovernanceError,
+    state::governance::{assert_is_valid_governance_config, get_governance_data, GovernanceConfig},
+};
 use solana_program::{
     account_info::{next_account_info, AccountInfo},
     entrypoint::ProgramResult,
     pubkey::Pubkey,
 };
-
-use crate::{
-    error::GovernanceError,
-    state::governance::{assert_is_valid_governance_config, get_governance_data, GovernanceConfig},
-};
+use solana_program::{rent::Rent, sysvar::Sysvar};
+use spl_governance_tools::account::{extend_account_size, AccountMaxSize};
 
 /// Processes SetGovernanceConfig instruction
 pub fn process_set_governance_config(
@@ -20,6 +21,8 @@ pub fn process_set_governance_config(
     let account_info_iter = &mut accounts.iter();
 
     let governance_info = next_account_info(account_info_iter)?; // 0
+    let payer_info = next_account_info(account_info_iter)?; // 1
+    let system_info = next_account_info(account_info_iter)?; // 2
 
     // Only governance PDA via a proposal can authorize change to its own config
     if !governance_info.is_signer {
@@ -35,6 +38,14 @@ pub fn process_set_governance_config(
     // For example changing approval quorum could accidentally make proposals to succeed which would otherwise be defeated
 
     governance_data.config = config;
+
+    let data_size = governance_data
+        .get_max_size()
+        .ok_or(GovernanceError::CannotCalculateSizeOfGovernanceData)?;
+    if governance_info.data.borrow_mut().len() < data_size {
+        let rent = Rent::get()?;
+        extend_account_size(governance_info, payer_info, data_size, &rent, system_info)?;
+    }
 
     governance_data.serialize(&mut *governance_info.data.borrow_mut())?;
 
--- a/governance/program/src/state/enums.rs
+++ b/governance/program/src/state/enums.rs
@@ -143,7 +143,7 @@ impl Default for ProposalState {
 
 /// The type of the vote threshold used to resolve a vote on a Proposal
 ///
-/// Note: In the current version only YesVotePercentage and Disabled thresholds are supported
+/// Note: In the current version only YesVotePercentage, AttendanceQuorum and Disabled thresholds are supported
 #[derive(Clone, Debug, PartialEq, Eq, BorshDeserialize, BorshSerialize, BorshSchema)]
 pub enum VoteThreshold {
     /// Voting threshold of Yes votes in % required to tip the vote (Approval Quorum)
@@ -152,11 +152,19 @@ pub enum VoteThreshold {
     /// In other words a '+1 vote' tie breaker is always required to have a successful vote
     YesVotePercentage(u8),
 
-    /// The minimum number of votes in % out of the entire pool of governance tokens eligible to vote
-    /// which must be cast for the vote to be valid
-    /// Once the quorum is achieved a simple majority (50%+1) of Yes votes is required for the vote to succeed
-    /// Note: Quorum is not implemented in the current version
-    QuorumPercentage(u8),
+    /// The minimum number of votes in basis points (1/10000) out of the entire pool of governance tokens eligible to vote
+    /// which must be cast for the proposal to be considered valid.
+    /// Once the quorum is achieved then pass level (in %) has to be casted for an option to succeed,
+    /// if defined to 50% it's a simple majority (50%+1) of Yes votes for the option to succeed
+    AttendanceQuorum {
+        #[allow(dead_code)]
+        /// The minimum number of votes in basis points (1/10000) out of the entire pool of governance tokens eligible to vote
+        /// which must be cast for the proposal being considered as successful
+        threshold: u16,
+        #[allow(dead_code)]
+        /// The minimum number of Yes votes in % required to be voted on an option for an option being successful
+        pass_level: u8,
+    },
 
     /// Disabled vote threshold indicates the given voting population (community or council) is not allowed to vote
     /// on proposals for the given Governance
--- a/governance/program/src/state/enums.rs
+++ b/governance/program/src/state/enums.rs
@@ -1,3 +1,4 @@
+#![allow(deprecated)]
 //! State enumerations
 
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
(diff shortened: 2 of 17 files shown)


governance/program/src/state/enums.rs
```diff
@@ -155,12 +155,17 @@ pub enum VoteThreshold {
...
     QuorumPercentage(u8),
 
     /// Disabled vote threshold indicates the given voting population (community or council) is not allowed to vote
     /// on proposals for the given Governance
     Disabled,
+
+    /// The minimum number of votes in % out of the entire pool of governance tokens eligible to vote
+    /// which must be cast for the proposal being considered as successful
+    /// Once the quorum is achieved all options are considered as successful
+    QuorumPercentageAllSucceed(u8),
```
SebastianBor, line 168: This should not be defined here because it would make it possible to very easily create malicious proposals, deceive community and trick it into approving some harmful instructions.  
 
In general a proposal with this type of voting semantics  cannot be used to execute instructions per option. It can only be used as a poll to ask for opinions or preferences across the given options.

Because of the above this should be defined as a vote type for a proposal and not as governance rules threshold and the governance rules quorum would be then used to determine whether the poll is valid or not (ex. minimum required attendance).
For such a proposal we can allocate a single, proposal level executable instruction (as opposed to an instruction per option)  and then make ti possible to execute it if the poll is valid.

SebastianBor (commented):
- For the attendance quorum we also have to implement the correct behaviour for tipping rules (`Early`, `Strict`, `Disabled`) 
- Let's do one thing at a time and do the `QuorumPercentageAllSucceed` semantics in a separate PE

governance/program/src/state/enums.rs
```diff
@@ -155,12 +155,17 @@ pub enum VoteThreshold {
     /// The minimum number of votes in % out of the entire pool of governance tokens eligible to vote
     /// which must be cast for the vote to be valid
     /// Once the quorum is achieved a simple majority (50%+1) of Yes votes is required for the vote to succeed
-    /// Note: Quorum is not implemented in the current version
+    // TODO: if quorum is achieved how it's considered about veto and deny votes?
     QuorumPercentage(u8),
```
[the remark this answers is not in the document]
ochaloup: Oh, `AttendanceQuorum(5000, 0)` does not work as `all succeeded` because of `deny_weigth` https://github.com/solana-labs/solana-program-library/blob/master/governance/program/src/state/proposal.rs#L403

ochaloup:
@SebastianBor I did a change here but it's only partial. I would like ask you for help on how to proceed right.

* `QuorumPercentageAllSucceed` was removed from this PR
* wondering what's the best way to define `AttendanceQuorum` enum
* I haven't got how is meant with `SetGovernanceConfig`. Is it that the `AccountMaxSize` for the `GovernanceV2` and `reserved_v2` cannot be changed? (I proposed that initially now)

ochaloup:
@SebastianBor I updated the PR based on points defined above in comments.
I would like to get a feedback from you to understand what paths are considered fine and what should be adjusted.

governance/program/src/state/proposal.rs
```diff
@@ -102,6 +102,15 @@ pub enum VoteType {
     },
 }
 
+/// Proposal type of definition the Proposal finalize vote processing
+#[derive(Clone, Debug, PartialEq, Eq, BorshDeserialize, BorshSerialize, BorshSchema)]
+pub enum ProposalType {
+    /// Standard proposal that verifies the defined threholds
+    VoteThreshold,
+    /// Threshold is not considered on finalizing votes, all options succeeds automatically
+    NoThresholdAllSucceed,
```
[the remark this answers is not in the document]
ochaloup: @SebastianBor I see.

* please, can you elaborate what's are the key points that are required to be adjusted here (probably such kind of proposal may not permit transactions?)
* is the way how the `Proposal` struct is changed fine from your perspective or you consider that could be done better in code style?

governance/program/src/processor/process_set_governance_config.rs
```diff
@@ -36,6 +40,13 @@ pub fn process_set_governance_config(
 
     governance_data.config = config;
 
+    // TODO: serializing with Borsh here is not a great idea
+    let data_size = governance_data.try_to_vec().unwrap().len();
```
SebastianBor, line 44: You can use `get_max_size()` to get the required size

governance/program/src/state/enums.rs
```diff
@@ -152,11 +152,11 @@ pub enum VoteThreshold {
...
-    /// The minimum number of votes in % out of the entire pool of governance tokens eligible to vote
-    /// which must be cast for the vote to be valid
-    /// Once the quorum is achieved a simple majority (50%+1) of Yes votes is required for the vote to succeed
-    /// Note: Quorum is not implemented in the current version
-    QuorumPercentage(u8),
+    /// The minimum number of votes in basis points (1/10000) out of the entire pool of governance tokens eligible to vote
+    /// which must be cast for the vote to be valid.
+    /// Once the quorum is achieved then pass level (in %) is taken into account,
+    /// if defined to 50% it's a simple majority (50%+1) of Yes votes for the vote to succeed
+    AttendanceQuorum(u16, u8),
```
SebastianBor, line 159: nit: Please use explicit names for the fields 
`AttendanceQuorum { threshold: u16, passLevel: u8 }`

governance/program/src/state/governance.rs
```diff
@@ -104,7 +104,7 @@ pub struct GovernanceV2 {
     /// Reserved space for versions v2 and onwards
     /// Note 1: V1 accounts must be resized before using this space
     /// Note 2: The reserved space should be used from the end to also allow the config to grow if needed
-    pub reserved_v2: Reserved120,
+    pub reserved_v2: Reserved112,
```
SebastianBor, line 107: There is no need to take this reserved space because we reallocate the account to `max_size` in `set_governance_config`

governance/program/src/state/proposal.rs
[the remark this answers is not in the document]
SebastianBor: I think we don't need this `ProposalType`  enum at all because a proposal without `Deny` vote is already a survey only `Proposal` and can't have executable instructions.

For survey only proposals we should only consider the whole `Proposal` as `Succeeded` or `Failed` and the individual options should just stay with `OptionVoteResult`  as `None` because there is no `Succeeded `/`Failed` condition for them.
The `Succeeded`/`Failed` status of the whole proposal should be determined based on whether the relevant quorum was achieved and hence making the survey valid. For example if enough members participated with the `AttendanceQuorum` in the survey then we consider it as `Succeeded`.

governance/program/src/state/proposal.rs
```diff
@@ -583,6 +647,11 @@ impl ProposalV2 {
         Ok(max_voter_weight)
     }
 
+    /// Makes sum of weights of all votes that have been voted at proposal until now
+    pub fn resolve_all_votes_weight(&self) -> u64 {
```
SebastianBor, line 651: This is incorrect, because in case where users can place their vote weight on multiple options the same weight would be calculated multiple times.
You have to introduce a field on `Proposal` account to track the total weight of approving votes. You can use the space from `reserved: [u8; 64]` to introduce the field

governance/program/src/state/proposal.rs
```diff
@@ -622,9 +691,21 @@ impl ProposalV2 {
         vote_threshold: &VoteThreshold,
         vote_kind: &VoteKind,
     ) -> Option<ProposalState> {
-        let min_vote_threshold_weight =
+        let mut min_vote_threshold_weight =
             get_min_vote_threshold_weight(vote_threshold, max_voter_weight).unwrap();
 
+        if let VoteThreshold::AttendanceQuorum(_, pass_level) = vote_threshold {
```
SebastianBor, line 697: For attendance quorum threshold you have to add all approving votes  weight plus deny vote weight with the following logic

If `approve_weight` + `deny_weight` <  `attendance_threshold` then `Proposal` failed 
Evaluate each option individually
if `option_weight` >= `passLevel` and `option_weight` > `deny_weight` then option `Succeeded`

SebastianBor (commented):
I reviewed the main changes but it's still needed to look in more details into the vote results calculations and ensure the `voteTipping` rules are applied correctly for `SingleChoice` vote

ochaloup (commented):
I made changes discussed in the comments while the PR contains 2 commits now.
The first one adds `AttendanceQuorum`, the second one adjusts survey like proposal when `deny_weight` is not configured.

I checked the code and math. To me it works fine but I hope that I haven't missed some of the requirements.
(while I could need a help to make some better naming)

governance/program/src/state/proposal.rs
```diff
@@ -194,7 +194,11 @@ pub struct ProposalV2 {
     pub vote_threshold: Option<VoteThreshold>,
 
     /// Reserved space for future versions
-    pub reserved: [u8; 64],
```
ochaloup, line 197: Thinking if it could be done line `[u32; 6] + [u8;8]`. I'm not sure what's best practice about this.

governance/program/src/state/enums.rs
```diff
@@ -152,11 +152,19 @@ pub enum VoteThreshold {
...
-    /// The minimum number of votes in % out of the entire pool of governance tokens eligible to vote
-    /// which must be cast for the vote to be valid
-    /// Once the quorum is achieved a simple majority (50%+1) of Yes votes is required for the vote to succeed
-    /// Note: Quorum is not implemented in the current version
-    QuorumPercentage(u8),
+    /// The minimum number of votes in basis points (1/10000) out of the entire pool of governance tokens eligible to vote
+    /// which must be cast for the proposal to be considered valid.
+    /// Once the quorum is achieved then pass level (in %) has to be casted for an option to succeed,
+    /// if defined to 50% it's a simple majority (50%+1) of Yes votes for the option to succeed
+    AttendanceQuorum {
```
mschneider, line 159: better to only add a new enum type, and mark the old one as deprecated this way it's easier to reason about backwards compatibility

governance/program/src/state/proposal.rs
mschneider: the way you did it, is fine

governance/program/src/state/enums.rs
mschneider: also: why allow dead_code? the code is used no?

governance/program/src/state/enums.rs
```diff
@@ -152,11 +152,19 @@ pub enum VoteThreshold {
...
+        #[allow(dead_code)]
+        /// The minimum number of votes in basis points (1/10000) out of the entire pool of governance tokens eligible to vote
+        /// which must be cast for the proposal being considered as successful
+        threshold: u16,
+        #[allow(dead_code)]
+        /// The minimum number of Yes votes in % required to be voted on an option for an option being successful
+        pass_level: u8,
```
mschneider, line 166: yes votes in % of what? entire pool of governance tokens? please update comment.

also "threshold" & "pass_level" could be more descriptive,
maybe just reuse previous terms "quorum_bps", "yes_vote_percentage"

ochaloup: Correct, % of all governance tokens. Thanks for noting that. I tried to put it more descriptive.

I quite like the recommendation of used terms `quorum_bps` and `yes_vote_percentage`.

ochaloup (commented):
@mschneider Thank you. I like your points. I updated the PR based on your suggestions. Interested in points from @SebastianBor around this.

PR #3849, opened 2022-11-29
ci: Fix 2xtx build

joncinque:
#### Problem

As reported on Stack Exchange and select users, the 2x-tx build isn't working.  See https://solana.stackexchange.com/questions/4449/solana-build-error-after-twoxtx-setup-sh-patch-with-2x-transaction-size for more info about the error.

#### Solution

Fix the const assert checks in the patch.

cc @askibin


joncinque:
Merging on red since this passed the important check: Token Pull Request / cargo-test-sbf-twoxtx

PR #3853, opened 2022-11-29
stake-pool: Refresh blockhash more often on update tests

joncinque:
#### Problem

The stake pool tests still fail CI randomly.  This may be due to a known issue of transactions short-circuiting if they have the same signature, e.g. doing the same transfer twice with the same blockhash.  This might be causing some tests to timeout in CI.

#### Solution

Refresh the blockhash more often in the `update_validator_list_balance*` tests.


joncinque:
One day, this will be totally stable.  Until then, we fix CI errors!

PR #3854, opened 2022-11-29
stake-pool: Make transaction signatures unique in withdraw test

joncinque:
#### Problem

There are still new failures in the stake pool tests. The most recent one https://github.com/solana-labs/solana-program-library/actions/runs/3576868159/jobs/6015187724 is likely due to running the same transaction signature twice.

#### Solution

Make the transaction different to force a new signature.


joncinque:
Maybe this will finally do it!

PR #3856, opened 2022-11-30
stake-pool: Support stake redelegation

### LINKED ISSUE #3749: stake-pool: Integrate stake redelegation
#### Problem

The stake program added the redelegate instruction in https://github.com/solana-labs/solana/pull/26294, which easily moves stake from one validator to another without losing an epoch of rewards due to the flow of deactivate -> wait -> delegate -> wait.

Stake pools are extremely important users of stake accounts, and we want most stake to go through a stake pool at some point.  When stake pools rebalance, they lose out on rewards, impacting their performance, so many just don't do it, or only move a minimal amount to keep their financial performance high.

#### Solution

Add a `redelegate` instruction to the stake pool program, which moves a certain number of lamports from one validator to another.  Underneath, this instruction will:

* split lamports from the main stake account into the redelegation stake account, a PDA given by some seed (also part of the instruction), where the seed will be stored in the `ValidatorStakeInfo` entry for that validator in the pool
* redelegate from the redelegation stake account to an ephemeral stake account, a PDA given by some other seed (honestly, it doesn't matter because it will be immediately deleted in the next step)
* immediately merge the ephemeral stake account into the transient stake account for the validator

This model allows for multiple redelegations *into* the same validator in one epoch.  The redelegated stake account needs to stay around to be reabsorbed into the pool, so there's only one redelegation possible *from* a validator per epoch. However, this means that it is *not* possible to decrease stake from a validator *and also* redelegate to it.  That would be a weird situation though.

During the `update_validator_list_balance` instruction, instead of passing in two accounts per validator (the main stake account and the transient stake account), we'll also pass in the redelegation stake account, so we'll update in groups of three.  If the redelegation stake account exists and is deactivated, the reserve stake account receives the rent-exempt lamports.

Since there's enough changes happening in https://github.com/solana-labs/solana-program-library/pull/3714, let's land that first before tackling this.

cc @mvines

joncinque:
#### Problem

https://github.com/solana-labs/solana/pull/26294 added the ability to redelegate stake in one go, without waiting for cooldowns and warmups, allowing delegators to move stake much more easily.

While normal delegators will certainly make good use of it, stake pools will truly benefit from it, allowing managers to maintain high performance on their pool.

#### Solution

Add the ability to redelegate on stake pools!

The implementation aims to keep this simple, reusing the transient stake accounts used for normally rebalancing pool.  The instruction processor roughly does:

* split from source account into transient
* redelegate from transient account into ephemeral account -- this account is only used for the duration of the transaction, and uses a PDA to avoid the additional signature cost on a normal ephemeral account. To avoid griefing possibilities, you can pass in any u64 seed for it.
* if the destination validator has a transient account, try to merge the ephemeral account into it -- this means you can redelegate *into* one validator multiple times per epoch, which is pretty neat. This encourages easy movement into good validators.
* if the destination validator has no transient account, split the entirety of the ephemeral account into the new transient account

Fixes #3749

### DIFF
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -501,7 +501,7 @@ pub enum StakePoolInstruction {
     ///  5. `[w]` Source transient stake account to receive split and be redelegated
     ///  6. `[w]` Uninitialized ephemeral stake account to receive redelegation
     ///  7. `[w]` Destination transient stake account to receive ephemeral stake by merge
-    ///  8. `[w]` Destination stake account to receive transient stake after activation
+    ///  8. `[]` Destination stake account to receive transient stake after activation
     ///  9. `[]` Destination validator vote account
     /// 10. `[]` Clock sysvar
     /// 11. `[]` Stake History sysvar
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -473,6 +473,58 @@ pub enum StakePoolInstruction {
         /// seed used to create ephemeral account.
         ephemeral_stake_seed: u64,
     },
+
+    /// (Staker only) Redelegate active stake on a validator, eventually moving it to another
+    ///
+    /// Internally, this instruction splits a validator stake account into its
+    /// corresponding transient stake account, redelegates it to an ephemeral stake
+    /// account, then merges that stake into the destination transient stake account.
+    ///
+    /// In order to rebalance the pool without taking custody, the staker needs
+    /// a way of reducing the stake on a stake account. This instruction splits
+    /// some amount of stake, up to the total activated stake, from the canonical
+    /// validator stake account, into its "transient" stake account.
+    ///
+    /// The instruction only succeeds if the source transient stake account and
+    /// ephemeral stake account do not exist.
+    ///
+    /// The amount of lamports to move must be at least twice rent-exemption
+    /// plus the minimum delegation amount. Rent-exemption is required for the
+    /// source transient stake account, and rent-exemption plus minimum delegation
+    /// is required for the destination ephemeral stake account.
+    ///
+    ///  0. `[]` Stake pool
+    ///  1. `[s]` Stake pool staker
+    ///  2. `[]` Stake pool withdraw authority
+    ///  3. `[w]` Validator list
+    ///  4. `[w]` Source canonical stake account to split from
+    ///  5. `[w]` Source transient stake account to receive split and be redelegated
+    ///  6. `[w]` Uninitialized ephemeral stake account to receive redelegation
+    ///  7. `[w]` Destination transient stake account to receive ephemeral stake by merge
+    ///  8. `[w]` Destination stake account to receive transient stake after activation
+    ///  9. `[]` Destination validator vote account
+    /// 10. `[]` Clock sysvar
+    /// 11. `[]` Rent sysvar
+    /// 12. `[]` Stake History sysvar
+    /// 13. `[]` Stake Config sysvar
+    /// 14. `[]` System program
+    /// 15. `[]` Stake program
+    Redelegate {
+        /// Amount of lamports to redelegate
+        #[allow(dead_code)] // but it's not
+        lamports: u64,
+        /// Seed used to create source transient stake account
+        #[allow(dead_code)] // but it's not
+        source_transient_stake_seed: u64,
+        /// Seed used to create destination ephemeral account.
+        #[allow(dead_code)] // but it's not
+        ephemeral_stake_seed: u64,
+        /// Seed used to create destination transient stake account. If there is
+        /// already transient stake, this must match the current seed, otherwise
+        /// it can be anything
+        #[allow(dead_code)] // but it's not
+        destination_transient_stake_seed: u64,
+    },
 }
 
 /// Creates an 'initialize' instruction.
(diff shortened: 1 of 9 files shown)


mvines:
@2501babe - have you had a chance to learn about the stake pool program yet? 
If no, this could be a nice place to start asking question and learning.
If yes, please help review to get more eyes on this.

@joncinque - do you have a line to any ecosystem teams using the SPL stake-pool program? if possible it'd be nice to involve engineers from those teams for review here too

stake-pool/program/src/instruction.rs
```diff
@@ -421,6 +421,58 @@ pub enum StakePoolInstruction {
...
+    ///  2. `[]` Stake pool withdraw authority
+    ///  3. `[w]` Validator list
+    ///  4. `[w]` Source canonical stake account to split from
+    ///  5. `[w]` Source transient stake account to receive split and be redelegated
+    ///  6. `[w]` Uninitialized ephemeral stake account to receive redelegation
+    ///  7. `[w]` Destination transient stake account to receive ephemeral stake by merge
+    ///  8. `[w]` Destination stake account to receive transient stake after activation
+    ///  9. `[]` Destination validator vote account
+    /// 10. `[]` Clock sysvar
+    /// 11. `[]` Rent sysvar
```
mvines, line 455: sysvar `get()` can be used for these instead of passing in the addresses right?

stake-pool/program/src/instruction.rs
```diff
@@ -421,6 +421,58 @@ pub enum StakePoolInstruction {
...
+        lamports: u64,
+        /// Seed used to create source transient stake account
+        #[allow(dead_code)] // but it's not
+        source_transient_stake_seed: u64,
+        /// Seed used to create destination ephemeral account.
+        #[allow(dead_code)] // but it's not
+        ephemeral_stake_seed: u64,
+        /// Seed used to create destination transient stake account. If there is
+        /// already transient stake, this must match the current seed, otherwise
+        /// it can be anything
```
mvines, line 472: Should we recommend the client generate a random u64 value?
`otherwise it can be anything` will cause every good lazy programmer to hard code 0

stake-pool/program/src/instruction.rs
joncinque: Ah, good call. The clock is required for stake merge, so just the rent can be removed

stake-pool/program/src/instruction.rs
joncinque: They can be reused, similar to transient stake accounts used for increase / decrease. `0` is fine for the most part, but the stake bots just increment every time to be safe.

stake-pool/program/src/instruction.rs
mvines: Ah dang. Yeah we'd need to add a new stake program instruction

joncinque:
> do you have a line to any ecosystem teams using the SPL stake-pool program? if possible it'd be nice to involve engineers from those teams for review here too

@mvines yep, I've involved Marinade so far, and will involve Socean too. Otherwise, this will also get audited

CMEONE:
> do you have a line to any ecosystem teams using the SPL stake-pool program? if possible it'd be nice to involve engineers from those teams for review here too
@mvines Engineers from BlazeStake and Jito will also review the code!

@AlexanderRay does JPool want to take a look at this PR as well? I'm happy to work together again with you on adding and testing more JS bindings as well :)

stake-pool/program/src/instruction.rs
```diff
@@ -421,6 +421,57 @@ pub enum StakePoolInstruction {
...
+    ///
+    ///  0. `[]` Stake pool
+    ///  1. `[s]` Stake pool staker
+    ///  2. `[]` Stake pool withdraw authority
+    ///  3. `[w]` Validator list
+    ///  4. `[w]` Source canonical stake account to split from
+    ///  5. `[w]` Source transient stake account to receive split and be redelegated
+    ///  6. `[w]` Uninitialized ephemeral stake account to receive redelegation
+    ///  7. `[w]` Destination transient stake account to receive ephemeral stake by merge
+    ///  8. `[w]` Destination stake account to receive transient stake after activation
```
2501babe, line 452:
```suggestion
    ///  8. `[]` Destination stake account to receive transient stake after activation
```
`AccountMeta::new_readonly(*destination_validator_stake, false),`

2501babe:
going to finish reviewing this on monday, i stumbled on `stake_allow_zero_undelegated_amount` when looking at the stake program code and went down a rabbithole because of its possible relevance to the single-pool

stake-pool/program/src/processor.rs
```diff
@@ -1444,6 +1568,316 @@ impl Processor {
...
+                &vote_account_address,
+                NonZeroU32::new(validator_stake_info.validator_seed_suffix),
+            )?;
+            if validator_stake_info.transient_stake_lamports > 0 {
+                return Err(StakePoolError::TransientAccountInUse.into());
+            }
+            if validator_stake_info.status != StakeStatus::Active {
+                msg!("Validator is marked for removal and no longer allows increases");
+                return Err(StakePoolError::ValidatorNotFound.into());
+            }
```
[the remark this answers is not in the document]
2501babe: oh, is this in reference to the increase to the ephemeral account instead?

stake-pool/program/tests/redelegate.rs
```diff
@@ -0,0 +1,1129 @@
...
+        stake_pool_accounts,
+        source_validator_stake,
+        destination_validator_stake,
+        minimum_redelegate_lamports,
+        slot,
+    )
+}
+
+#[tokio::test]
+async fn success() {
```
2501babe, line 122: it would be nice if the test names were prefixed with `redelegate_` so that `cargo test-sbf redelegate` does this expected thing

2501babe:
i cant find anything substantive to criticize. also i made a point of tracing where all the account infos are validated and that all looks solid to me

also, for my own enlightenment... i was initially confused why we didnt reclaim the rent from the source transient account at the end of `process_redelegate`, and used one of the tests to confirm it actually didnt work (because redelegate marks the source as deactivated), and tracked down the `StakeError::MergeTransientStake`. this is because the source transient account still retains the effective stake until the next epoch, at which point it moves to the destination transient account, right? at which point, `UpdateValidatorListBalance` will merge the destination transient acocunt into the destination main account, and sweep the source transient account into the reserve

stake-pool/program/src/processor.rs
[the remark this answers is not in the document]
joncinque: Nice spot, this was definitely some bad copy-paste. I'll change it to "no longer allows redelegation"

stake-pool/program/tests/redelegate.rs
joncinque: ah, I Just do `cargo test-sbf --test redelegate` to scope the runs. Since you can scope that way, I think it's redundant to add `redelegate_` again in the test name. Is that ok?

joncinque:
It's a confusing side-effect of the redelegate implementation, which takes advantage of the duplication of info between the `Delegation` struct on the stake account and its lamports. All of the lamports (minus rent) are moved from the source to the destination, but the source is still delegated and deactivating, and retains the delegated amount info. This way, the source gets properly credited with epoch rewards.

The destination *also* gets the effective stake amount, so the stake is actually double counted, but the lamports are only in one place. I spent a long time making sure that it wasn't exploitable, and I could only find weird corner cases. If you want to spend some time looking at the redelegate implementation, more eyes are always welcome!

And yeah, as you noticed, when the source transient stake deactivates, then the reserve can consume its rent and rewards.

stake-pool/program/tests/redelegate.rs
2501babe: ooh til

joncinque:
Merging to provide an easier patch for audits. Feel free to provide post-merge comments!

Issue #3858, opened 2022-11-30
Allow for multiple delegates per Token Account


jamesrp13:
**Problem:** At present, there can only be one delegate per Token Account. This means that any delegation to a program that will perform ongoing or future actions on behalf of a user will likely be interrupted by users interacting with a separate program that also requires delegation. 

**Solution:** My preferred solution would be to add this to the Token Program itself, but that seems unlikely for the program that is live given that it would require changes to account structures. The next best solution is probably a separate program (like the ATA program) that becomes the new standard through which delegation is performed. This program would handle delegate approvals and would itself be the single delegate for all token accounts moving forward. It would then be responsible for routing delegation functions properly.

**Use cases:** While we certainly should be smart about how we use delegation, there are plenty of cases where delegation is what provides for the best end user experience. The primary example I can think of is for subscription payments. Without delegation, this can only happen via some sort of escrow pattern or making users manually transfer funds at each subscription interval. That would suck.

Idk what else we should put in here, but let's at least get the conversation started.

jamesrp13:
It appears @mfrager has made something along these lines already: https://github.com/atellix/token-delegate

Curious what others think of it

mfrager:
If people want to use this program as well, I will make it immutable for security purposes. Any feedback is very welcome! Thanks.

jamesrp13:
Thanks @mfrager! It may be a good interim step to having a more formal multi-delegate pattern integrated into the token program or at a minimum made part of the SPL

dankelleher:
Isn't delegating to a PDA owned by a 1-of-n multisig program a more composable solution? This program would not need to be specific to SPL-Token, but would handle any comparable use case.

I'm thinking along the lines of the [cryptid](https://github.com/identity-com/cryptid) model:

1. define a set of linked addresses (your multi-delegates)
2. set the token account delegate to an account abstraction PDA based on that set
3. transactions from one of the delegates would go through the account abstraction program to check the delegate is in the list

Upsides:

- More composable - the delegation logic would be encapsulated in the account abstraction
- More powerful - you could add rules governing delegates to the account abstraction program (e.g. delegate A may spend up to X tokens)
- More flexible - use the same account abstraction on other programs

Downsides:
- Tx compute/size overhead of piping through the account abstraction program
- More complexity on the client side

mfrager:
I’m not sure the previous reply understands exactly the problem to be solved.

[...]

PR #3859, opened 2022-12-01
Multi delegate program draft

Arrowana:
Added some thoughts into a proposal and wrote a draft of the program process_instruction

it is loose but i don't want to over constrain with checks until the design looks sound.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6134,6 +6134,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "spl-multi-delegate"
+version = "0.0.1"
+dependencies = [
+ "assert_matches",
+ "borsh",
+ "num-derive",
+ "num-traits",
+ "solana-program",
+ "spl-token 3.5.0",
+ "thiserror",
+]
+
+[[package]]
 name = "spl-name-service"
 version = "0.2.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,8 @@ members = [
   "utils/cgen",
   "utils/test-client",
   "token-lending/flash_loan_receiver",
+  "token-multi-delegate/program",
+  #"token-multi-delegate/program-test",
 ]
 exclude = [
   "farms/farm-client",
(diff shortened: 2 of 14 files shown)


jamesrp13:
@Arrowana thanks for putting this together. My concern about multi-delegation happening in a separate program from the token program is the attack vector (or just inconvenience) of undercutting existing delegates by reassigning the underlying single delegate on a token account. 

E.g. a user has 5 delegates approved for a USDC account, then they submit a transaction to another program with the intent of adding it as a 6th delegate in the same USDC account. What they don't realize is that this program doesn't use the same multi-delegate program and instead sets the delegate on the token account itself, thereby removing the connection with the multi-delegate program and effectively removing the first 5 delegates.

I personally think the addition of something to the Token Program itself to avoid this possibility is worth not putting the onus onto the end user to know that there are various approaches to token delegation and choosing programs according to which they want.

That being said, I'm curious how this fits into what you've designed, if you have suggestions for mitigating the problem I mentioned above, etc.

Arrowana:
Having it in spl was one way to share a way of doing it for everyone.

There are definitely some issues that could be eliminated if it was directly in the token program:
- Re-approving PDA "reapproves" all the previous delegates
- Approving something else once shortcircuits the multi delegate
- Looks like another token transfer with the delegate signing without extra accounts

But i don't think the trivial relationship between delegate and a token transfer (same for owner or delegate) has to/can go away so it feels like those issues will remain.

The token transfer will need the multi delegate account and the underlying delegate, or were you thinking on storing all the delegates under the token account itself?

jamesrp13:
I don't have any issue storing the multi-delegate in a separate program. The only thing I'm not loving about having a separate program is how easy it is to short-circuit the multi-delegate (E.g. user simply interacting with a program that adds itself as the token program delegate directly rather than using the multi-delegate program). 

I would absolutely love to find a solution for that.

jamesrp13:
Noting that issue #3858 has some discussion going on about this

nickgarfield:
Ah, I didn't realize this PR was already underway. We've outlined a data model in https://github.com/solana-labs/solana-program-library/issues/3858#issuecomment-1334782657 that could fit into token22 with minimal changes to the existing interfaces. Would love to get your feedback on it @Arrowana! 

As mentioned in the comments above, the main concern with pushing the multi-delegate implementation into a secondary program is that the authority can be short-circuited by other programs. Ideally, approvals should be a non-rivalrous. The token22 program seems like an ideal opportunity to solve this problem at the root level and remove the artificial scarcity currently limiting users to one approved delegate per token account. This would massively unblock payments developers, finally allowing them to re-create the experiences users have come to expect from traditional fiat services.

PR #3875, opened 2022-12-06
Expose more types from CMT

austbot:
Indexing CMT is made difficult without additional types.
Consider when needing to ensure we have not missed a tree while indexing. We will need an account data source.
* Geyser
* RPC Via GPA

If we get the raw account data and deserialize it into a `ConcurrentMerkleTreeHeader` we will not be able to introspect the data version or event the account type.

This pr exposes new types and bumps the version. Please consider this contribution as it could be a blocker to the additional layer of merkle tree backfilling needed for Metaplex Read API.

I also add a `get_creation_slot` on the Main header struct as a convenience feature as I'm relatively sure that information will be present in all data versions.

### DIFF
--- a/account-compression/Cargo.lock
+++ b/account-compression/Cargo.lock
@@ -1200,7 +1200,7 @@ dependencies = [
 
 [[package]]
 name = "spl-account-compression"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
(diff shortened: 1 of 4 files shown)


ngundotra:
Looks great. Thanks Austin!

PR #3879, opened 2022-12-06
Governance: assertion for proposal final state

ochaloup:
Adding assertion for check the state of the proposal is final.

### DIFF
--- a/governance/program/src/state/proposal.rs
+++ b/governance/program/src/state/proposal.rs
@@ -259,6 +259,22 @@ impl ProposalV2 {
         Ok(())
     }
 
+    /// Checks the Proposal was finalized (no more state transition will happen)
+    pub fn assert_is_finalized(&self) -> Result<(), ProgramError> {
+        match self.state {
+            ProposalState::Completed
+            | ProposalState::Cancelled
+            | ProposalState::Defeated
+            | ProposalState::Vetoed => Ok(()),
+            ProposalState::Executing
+            | ProposalState::ExecutingWithErrors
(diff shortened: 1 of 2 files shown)


governance/program/src/state/proposal.rs
```diff
@@ -259,6 +259,22 @@ impl ProposalV2 {
         Ok(())
     }
 
+    /// Checks the Proposal was finalized (no more state transition will happen)
+    pub fn assert_is_finalized(&self) -> Result<(), ProgramError> {
```
SebastianBor, line 263: nit: `Finalised` can be confusing because `FinaliseVote` instruction is used to transition a `Proposal` in `Voting` state to either `Defeated` or `Succeeded`.
We could be more explicit about it and name the function `assert_is_final_state()` instead.

SebastianBor (commented):
The function is not used anywhere in the code. Is the intention to use it externally as an api call?

ochaloup:
Yes, that was meant to be called from outside from the contract we use to call the SPL GOV. But I found we need rather to check only the `Completed` state. 

I'm going to close this PR. Thanks for the point about the `final_state` naming.

ochaloup:
Sorry for going back and forth. But realized it's reasonable for me to have such check in "API". I renamed the method. If it's ok for you to have it here it would be good for me.

ochaloup:
@SebastianBor I created the PR of the same meaning once againhttps://github.com/solana-labs/solana-program-library/pull/3964

PR #3884, opened 2022-12-09
token-swap: Fix owner fee calculation on swap

joncinque:
#### Problem

The token-swap owner fee calculation on trades does *not* mint enough pool tokens, since it figures out how many tokens to mint using `withdraw_single_token_type_exact_amount_out`. This is incorrect, since it takes into account how many pool tokens *currently* exist, and not how many *will* exist after minting.

#### Solution

No solution yet in this PR, just a failing test that shows the issue.  E.g. if the owner swap fee is worth 5 A tokens, right after a swap, if the owner tries to withdraw the pool tokens they received, they will get *less* than 5 A tokens.

cc @kklas

### DIFF
--- a/token-swap/program/src/curve/base.rs
+++ b/token-swap/program/src/curve/base.rs
@@ -567,4 +567,67 @@ mod test {
             );
         }
     }
+
+    #[test]
+    fn owner_fee_calc() {
+        // setup
+        let pool_supply: u128 = 1_000_000_000;
+        let swap_source_amount: u128 = 2_000_000_000;
+        let swap_destination_amount: u128 = 50_000_000_000;
+        let source_amount: u128 = 10_000_000;
+        let fees = Fees {
+            owner_trade_fee_numerator: 1,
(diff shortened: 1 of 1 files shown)


kklas:
I'm highly confident that [`deposit_single_token_type`](https://github.com/solana-labs/solana-program-library/blob/master/token-swap/program/src/curve/base.rs#L106)  should be used instead (with fees omitted).

At least for the CP curve, it calls into [`this`](https://github.com/solana-labs/solana-program-library/blob/5d16d5b926cd677d40c3f207906085dd8f1a6aa5/token-swap/program/src/curve/constant_product.rs#L95-L100) function which correctly implements the single sided deposit formula from the Balancer whitepaper:
image

Using this formula will give a correct amount of LP tokens that correspond to the input token amount deposited. If the LP receiver then does a single sided withdrawal using these LP tokens, it will get the token amount which correctly corresponds to that same amount (but slightly less because on single sided withdrawal it also pays the trade fees on half of the amount which is expected).

PR #3888, opened 2022-12-09
Open Discussion For Permissioned Transfer Extension

jpbogle:
POC
this is purely for the sake of discussion right now and just put some code there to show what I mean before continuing

**Problem**
Given lots of arguments over interfaces on solana not existing, there are many mutually exclusive solutions for restricting transfer of tokens (examples below)

https://github.com/solana-labs/solana-program-library/tree/master/managed-token
https://github.com/magiceden-oss/open_creator_protocol
https://github.com/magiceden-oss/community-managed-token
https://github.com/cardinal-labs/cardinal-creator-standard

They all require a new account being passed into to every program and wallet or dapp to call a new program for transfer. The problem is the interface gets pushed to the client and its a big problem to have to update clients to get access to a new implementation for example. 

**Suggestion**
One type of “interface” could be instead using a transfer authority in token22 such that it has a defined program and list of accounts it can CPI to on transfer that can either error or not. This way all transfers can still go through token-program first and then cpi to another program to handle checks

This program can define the accounts it wants for the given mint and then configure itself to fail based on data it has in those accounts.

**Discussion**
1. Is this something labs would accept
2. Extension would ideally be re-sizeable with realloc rather than fixed size for the list of additional_accounts but this is not considered at all here for sake of discussion and left with fixed size
3. Naming of TransferAuthority kind of conflicts with the other authorities which are all change-able via the “SetAuthority” Instruction so maybe a different name is better ---> Renamed to PermissionedTransfer
3. This extensions pushes the signers back in the accounts iterator because it requires additional account_infos. Those could come AFTER signers if desired but it seemed okay to go before signers but i didnt think too closely yet
4. The unlucky part about this approach is that transfer right now does not always take in Mint account, it seems `TokenInstruction::Transfer` does not include the mint necessarily and `TokenInstruction::TransferChecked` does
       - `TokenInstruction::Transfer` is deprecated - is this planning to be removed entirely? The reason this extension would also need extension state on the account to indicate the mint is a required account on transfer - Similar to how `TransferFeeAmount` extension works it looks to just break on the non checked transfer
5. Of course since this is just for discussion there is no tests or client code right now

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -13,6 +13,7 @@ use {
             mint_close_authority::MintCloseAuthority,
             non_transferable::NonTransferable,
             permanent_delegate::PermanentDelegate,
+            transfer_authority::{TransferAuthorityAccount, TransferAuthorityMint},
             transfer_fee::{TransferFeeAmount, TransferFeeConfig},
         },
         pod::*,
@@ -53,6 +54,8 @@ pub mod non_transferable;
 pub mod permanent_delegate;
 /// Utility to reallocate token accounts
 pub mod reallocate;
+/// Transfer authority extension
+pub mod transfer_authority;
 /// Transfer Fee extension
 pub mod transfer_fee;
 
@@ -641,6 +644,10 @@ pub enum ExtensionType {
     CpiGuard,
     /// Includes an optional permanent delegate
     PermanentDelegate,
+    /// Config for transfer authority on mint
+    TransferAuthorityMint,
+    /// Config for transfer authority on account
+    TransferAuthorityAccount,
     /// Padding extension used to make an account exactly Multisig::LEN, used for testing
     #[cfg(test)]
     AccountPaddingTest = u16::MAX - 1,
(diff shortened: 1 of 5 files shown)


joncinque:
After chatting around to get to know the problem a bit more, a solution like this could be really useful. We could make it just have the program id, and by convention the caller must also pass some PDA, and then every other account in the instruction gets passed too.

The token program CPIs into the specified program id with all of the accounts marked as readonly and with some predefined instruction data. We just need to define some interface for that instruction data.

jpbogle:
@joncinque ya it could be a PDA derived at the same address in the specified program - but if there is an array of additional accounts that get passed all read_only then new ones can be added, specificaly like InstructionSysvar is one that some may need

and ya for instruction data was thinking maybe just a `u8` like `1` in instruction data - i left it empty array for now

another option to make it work with anchor interface is hard coding the instruction discriminator for like hash("transfer_restriction_extension") to make it easier for an anchor program

joncinque:
> another option to make it work with anchor interface is hard coding the instruction discriminator for like hash("transfer_restriction_extension") to make it easier for an anchor program

Yeah that would be perfect

> perhaps the transfer amount could be included as well - which would allow someone to do something based on the amount being transferred

Definitely, that makes sense too

jpbogle:
added anchor discriminator computed as `hash("global:permissioned_token_transfer") and the amount to instruction data -- I believe any anchor program that implements `permissioned_token_transfer(amount: u64)` should work correctly and non-anchor programs can just look for match on those first 8 bytes

joncinque:
With all of the permissioned-transfer work, now called the "transfer hook interface" and implemented in #4147 #4151 and #4165, I think this can be closed

PR #3890, opened 2022-12-09
token-manager-registry: ABI for performing managed token transfers

joncinque:
#### Problem

With many different token metadata / royalty standards, it becomes very difficult for marketplaces to figure out how to move tokens. #3888 does a great job at summarizing the issue.

#### Solution

This is a POC of how a unified system could work between all metadata standards.  TLDR

* centralized registry says which program manages the mint
* manager program implements `GetTransferAccounts` and `UnifiedTransfer` with the precise ABI. In this example, `GetTransferAccounts` is instruction data of just `[8]`, and `UnifiedTransfer` is `[9, <amount_as_little_endian_u64>]`
* marketplace uses registry + `GetTransferAccounts` + `UnifiedTransfer` to move any token

There exists a centralized (yay) token manager registry, where the mint authority registers that their mint is managed by a certain program. This simply writes the program id of the managing program. So if someone uses this with spl-managed-token for mint A, the `PDA(mint A, token manager registry)` will have the pubkey for spl-managed-token.

A marketplace wants to transfer an A token.

First they fetch the managing program id by looking at the data stored at `PDA(mint A, token manager registry)`, and they see that it's spl-managed-token.

Next, they simulate a transaction with `GetTransferAccounts` on spl-managed-token, providing the mint and one more account given by `PDA(mint, spl-managed-token)`. The return data on this transaction gives the account metas required *on top* of the source, mint, destination, and owner.

Last, they call `UnifiedTransfer` with the source, mint, destination, owner, amount, and additional account metas.

You can see the whole flow at `success_with_unified_transfer`. It's a little goofy, but I think it conveys the core ideas. All of this is totally open to more changes, your feedback is very appreciated!

cc @jnwng @jpbogle please tag everyone who might be interested

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6399,6 +6399,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "spl-token-manager-registry"
+version = "0.1.0"
+dependencies = [
+ "solana-program",
+ "solana-program-test",
+ "solana-sdk",
+ "spl-token-2022 0.5.0",
+]
+
+[[package]]
 name = "spl-token-swap"
 version = "3.0.0"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6086,6 +6086,7 @@ dependencies = [
  "solana-sdk",
  "spl-associated-token-account 1.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "spl-token 3.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "spl-token-manager-registry",
  "thiserror",
  "tokio",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ members = [
   "stateless-asks/program",
   "token-lending/cli",
   "token-lending/program",
+  "token-manager-registry/program",
   "token-swap/program",
   "token-swap/program/fuzz",
   "token-upgrade/cli",
--- a/managed-token/program/Cargo.toml
+++ b/managed-token/program/Cargo.toml
@@ -26,6 +26,7 @@ solana-program = "1.14.10"
 shank = "^0.0.5"
 spl-token = { version = "3.5.0", features = ["no-entrypoint"] }
 spl-associated-token-account = { version = "1.1.1", features = [ "no-entrypoint", ] }
+spl-token-manager-registry = { path = "../../token-manager-registry/program", version = "0.1", features = [ "no-entrypoint", ] }
 thiserror = "^1.0.24"
 borsh = "0.9.3"
 
--- a/managed-token/program/src/accounts.rs
+++ b/managed-token/program/src/accounts.rs
@@ -9,6 +9,7 @@ pub struct InitializeMint<'a, 'info> {
     pub mint: &'a AccountInfo<'info>,
     pub payer: &'a AccountInfo<'info>,
     pub upstream_authority: &'a AccountInfo<'info>,
+    pub unified_transfer: &'a AccountInfo<'info>,
     pub system_program: &'a AccountInfo<'info>,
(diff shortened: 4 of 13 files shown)


aeyakovenko:
instead of dynamic dispatch, why doesn't the dapp just coerse the program into one that supports the same interface as Token?  It just needs to registry to tell the caller that the callee supports the token instructions.

nothing0012:
i think we might need more than just transfer interface, example is escrowless listings, "approve" and some higher order of "freeze" are nice to have to have as part of the unified interface.

jpbogle:
@joncinque I think i understand this helps makes managed-token into somewhat of an interface - so to summarize you register the token in the registry, and then any program can implement `GetTransferAccounts` and `UnifiedTransfer`? and this example provides an implementation of that in the spl-managed-token program? In this example, `unified_transfer` requires a signature from `upstream_authority` so if you want that to be a PDA it has to call - because wallet would just view `get_transfer_accounts` as opaque, the upstream authority doesnt have to be a signer necessarily right if it wasnt a signer in return data

If i understand this correctly the flow wallets that want to support "managed" tokens, AND non-managed tokenKeg tokens AND non-managed token22 tokens would be

1. check the registry program (hard coded address) at PDA(mint_id) for data

2. If data is present, deserialize the programID from that data, simulate getTransferAccounts on that programID to see the return data and then call `unifiedTransfer` on that data
    - return data includes token program in the accounts so this works with any token program already as well since the return data can check the token program it should

3. If data is NOT present
   - look at the owner of the token account to see if its owned by TokenKeg or Token22
   -  call "transfer" (or transfer_checked) on the "owner" program of the token account (will work on either tokenKeg or token22)

@aeyakovenko 
> instead of dynamic dispatch, why doesn't the dapp just coerse the program into one that supports the same interface as Token? It just needs to registry to tell the caller that the callee supports the token instructions.

I think the reason something like `GetTransferAccounts` that tells client additional accounts to pass it is required because otherwise if its just coerced, it means the only accounts that can be used are the same ones in token program so it cant really do much different than existing token program instructions

joncinque:
> i think we might need more than just transfer interface, example is escrowless listings, "approve" and some higher order of "freeze" are nice to have to have as part of the unified interface.

@nothing0012 I'm not sure I understand what you're suggesting. `approve` and `freeze` work directly on the tokens, right? in that case, shouldn't they just use the tokenkeg / token22 instructions? Or are you imagining some other use case where these instructions get routed through another program?

jpbogle:
>@nothing0012 I'm not sure I understand what you're suggesting. approve and freeze work directly on the tokens, right? in that case, shouldn't they just use the tokenkeg / token22 instructions? Or are you imagining some other use case where these instructions get routed through another program?

@joncinque `approve` and `freeze` don't work directly on the tokens because the tokens are frozen (in order to restrict transfer) so thats why to actually approve, it has to unfreeze, approve, refreeze

jpbogle:
>It looks like you understand the whole thing perfectly! Given what I'm proposing, how does it work for Cardinal's use cases? Keeping in mind that this is just a proposal, so everything is changeable. We can add more accounts and instructions as needed.

This should work ya. I am curious what the comparisons or thoughts of an approach like this one are https://github.com/solana-labs/solana-program-library/pull/3888

Just because this approach with consistently frozen tokens renders token-program essentially useless, but the goal is really to just restrict transfer, in token22 extension approach, `approve` and `revoke` would still work as is. Another big difference is that the entrypoints for wallets would always be token22 and require no additional network calls. 

With the registry approach here, the entrypoint isnt the token program and in theory the token program could never be called at all i think by the implementing program so not sure if thats considered a pro vs a con?

joncinque:
Closing in favor of the transfer hook interface implemented in #4147

PR #3895, opened 2022-12-13
stake-pool: allow removing validator below minimum

2501babe:
some random things i noted when reviewing the program as a whole. the comment changes are obvious enough. the change from `!=` to `>` is to cope with the possibility that `MINIMUM_DELEGATION_SOL` rises to put a poorly resourced validator below it, which otherwise would require the pool to stake _more_ to it to be able to remove it


joncinque (approved):
Thanks for noticing these!

PR #3899, opened 2022-12-13
clippy: Remove `useless_conversion`

joncinque:
#### Problem

When making the banks client error more expressive, we had to add a bunch of `useless_conversion` allows in the code to keep the downstream projects build intact in #2657

#### Solution

Now that the change is rolled out, we can remove all of the `#[allow(...)]`s


joncinque:
Merging on red for an unrelated JS test failure

PR #3904, opened 2022-12-14
Metadata pointer extension

jpbogle:
An example for an extension that points to metadata address with a schema type

**Summary**
Right now, it is just "assumed" metadata exists in the token-metadata program and it means that address gets hard-coded, but this extension would allow wallet / dapps to look for the specific metadata address that could exist in any program

The idea here is this would create a strong relationship between mint and metadata address specifically

**Solution**
This suggestion would add an extension that can be used on a `mint` account to specify a specific metadata address to use. This could live under any program and use any pre-defined schema for the data

**Schemas**
The PR also includes a single byte for the schema type. This can be used to indicate what kind of data exists at that address

Open to suggestions for how schemas should work, but i was thinking schema: 0 would deserialize as metaplex metadata and then other people can use the same schema, and new schemas can be registered somewhere else, separate program probably, that registers how to deserialize the data at that metadata address

cc @nothing0012 @joncinque

### DIFF
--- a/token/program-2022/src/extension/mint_metadata.rs
+++ b/token/program-2022/src/extension/mint_metadata.rs
@@ -1,0 +1,19 @@
+use solana_program::pubkey::Pubkey;
+
+use {
+    crate::extension::{Extension, ExtensionType},
+    bytemuck::{Pod, Zeroable},
+};
+
+/// Mint metadata extension data for mints.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct MintMetadata {
+    /// Schema of metadata
+    pub schema: u8,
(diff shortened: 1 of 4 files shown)


joncinque:
In general, we've never included info that has no purpose on the protocol. A metadata account has no impact on how the token protocol operates, and would be better served with a different program. I'm happy to be contradicted if there are good reasons though!

jpbogle:
@joncinque ya it makes sense - just when thinking about how to create different kinds of on-chain metadata that wallets can use like an interface I didn't see many other options

if multiple existed derived in the same way from the mint, the wallet / dapp would still need to know where to look unless there was some registry of metadatas with a priority order

joncinque:
This has been added in https://github.com/solana-labs/solana-program-library/pull/4549, closing this PR

PR #3909, opened 2022-12-14
Decrease MAX_POOL_SIZE in huge_pool test to make it work with rust 1.65

dmakarov:
The new rust compiler and standard library generate slightly increased number of SBF instructions, which triggers failure of the stake-pool/program/tests/huge_pool.rs test due to running over the compute budget limit. I would like to decrease the maximum size of the pool to make the test pass when compiled by the new toolchain.

### DIFF
--- a/stake-pool/program/tests/huge_pool.rs
+++ b/stake-pool/program/tests/huge_pool.rs
@@ -25,7 +25,7 @@ use {
 // the test require so many helper accounts.
 // 20k is also a very safe number for the current upper bound of the network.
 const MAX_POOL_SIZE_WITH_REQUESTED_COMPUTE_UNITS: u32 = 20_000;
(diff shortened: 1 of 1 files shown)


dmakarov:
@joncinque Jon, is it ok to merge this?

dmakarov:
> Ah well, if there's no other way, that's fine! It's weird though, it doesn't impact `MAX_POOL_SIZE_WITH_REQUESTED_COMPUTE_UNITS`?

I haven't looked at `MAX_POOL_SIZE_WITH_REQUESTED_COMPUTE_UNITS` because the only test that fails is `cleanup_all` and I thought that it is affected by `MAX_POOL_SIZE` only.

joncinque:
Ah ok, that's fine then, thanks!

PR #3910, opened 2022-12-14
fix: stake-pool-js: add browser CJS build

jordansexton:
stake-pool-js links its browser CJS output to an ESM build. This is likely to cause issues for a browser build that expects CJS. This PR adds an additional CJS build for browser output, and simplifies the Rollup config a little.

### DIFF
--- a/stake-pool/js/rollup.config.js
+++ b/stake-pool/js/rollup.config.js
@@ -31,7 +31,69 @@ function generateConfig(configType, format) {
     },
   };
 
-  if (configType !== 'browser') {
+  if (browser) {
+    if (format === 'iife') {
+      config.external = ['http', 'https'];
+
+      config.output = [
+        {
+          file: 'dist/index.iife.js',
+          format: 'iife',
+          name: 'solanaStakePool',
+          sourcemap: true,
+        },
+        {
+          file: 'dist/index.iife.min.js',
+          format: 'iife',
+          name: 'solanaStakePool',
+          sourcemap: true,
+          plugins: [terser({ mangle: false, compress: false })],
+        },
+      ];
+    }
+    else {
+      config.output = [
+        {
+          file: 'dist/index.browser.cjs.js',
+          format: 'cjs',
+          sourcemap: true,
+        },
+        {
+          file: 'dist/index.browser.esm.js',
+          format: 'es',
+          sourcemap: true,
+        },
+      ];
+
+      // Prevent dependencies from being bundled
+      config.external = [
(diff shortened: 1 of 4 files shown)


stake-pool/js/rollup.config.js
```diff
@@ -31,95 +31,85 @@ function generateConfig(configType, format) {
...
+        '@solana/spl-token',
+        '@solana/web3.js',
+        'bn.js',
+        'buffer',
+      ];
+    }
+
+    // TODO: Find a workaround to avoid resolving the following JSON file:
+    // `node_modules/secp256k1/node_modules/elliptic/package.json`
+    config.plugins.push(json());
```
jordansexton, line 81: ESM and CJS build _should_ be able to use identical config for everything except the `format`.

stake-pool/js/rollup.config.js
```diff
@@ -31,95 +31,85 @@ function generateConfig(configType, format) {
...
+    ];
   }
 
   return config;
 }
 
 export default [
   generateConfig('node'),
-  generateConfig('browser', 'esm'),
+  generateConfig('browser'),
```
jordansexton, line 113: We will get both CJS and ESM outputs from this config, same as the Node config.

jordansexton (commented):
Just left some clarifying comments on the intent here, since the refactor of conditionals created a bunch of churn.

joncinque (approved):
It was tough, but I think I followed all of the code paths -- looks great!

joncinque:
Looks like you'll also need to prettier on it

jordansexton:
Sounds good, I'll run and inspect the build locally as well.

jordansexton:
Found at least one relevant issue -- the files in the npm package @ https://www.npmjs.com/package/@solana/spl-stake-pool?activeTab=explore are significantly out of sync with respect to what's on Github, even prior to this PR. The build outputs are different than what's described here. There's only one build output for Node.js, which is an ESM file, but the published package.json isn't marked as as `"type": "module"`. This will cause problems for many projects importing this file which may expect it to be CJS. Nevertheless, this PR will fix it once published.

jordansexton:
Unrelated to this fix, but this package still depends on `"@solana/spl-token": "^0.1.8"`. We might want to update it, and check the other dependencies while we're at it.

jordansexton:
Published in @solana/spl-stake-pool@0.6.5

joncinque:
Thanks for taking care of all this!

PR #3920, opened 2022-12-16
token-js: rename ExtensionType.InterestBearingMint

2501babe:
consider this a suggestion to accept or reject. this is the only variant on the `ExtensionType` enum that doesnt match the name used in token22. because the js enum is supposed to correspond 1:1 with the program enum (to the extent that the order must match too, because the js enum values are compared to the type tags stored in `tlv_data` onchain) i feel they should be identical

admittedly the name is worse. as an alternative maybe we should rename `InterestBearingConfig` instead. but either way the js should be subservient to the rust imo

### DIFF
--- a/token/js/src/extensions/extensionType.ts
+++ b/token/js/src/extensions/extensionType.ts
@@ -24,7 +24,7 @@ export enum ExtensionType {
     ImmutableOwner,
     MemoTransfer,
     NonTransferable,
-    InterestBearingMint,
+    InterestBearingConfig,
     CpiGuard,
     PermanentDelegate,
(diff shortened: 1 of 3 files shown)


2501babe:
ive been getting
```
downloading v1.14.10 installer
Error: error decoding response body: invalid type: map, expected a sequence at line 1 column 0
Error: Process completed with exit code 1.
```

here and there for the past few days, i think its just transient network issues, itll pass once it lets me rerun it

joncinque (approved):
Makes sense to me, thanks for fixing this!

PR #3925, opened 2022-12-16
stake-pool: Add `DecreaseAdditionalValidatorStake` instruction

joncinque:
#### Problem

It's possible to do multiple increases in a stake pool with #3924, but this is also possible with decreases!

#### Solution

There's nothing technically stopping two deactivating stakes from merging, so introduce a new instruction on the stake pool program which allows for more decreases in an epoch.

The normal instruction splits into the transient account and then deactivates it, so this instruction:

* splits into an ephemeral stake account
* if the transient account exists: merges the ephemeral stake account into the transient stake account
* if the transient account doesn't exist: splits from ephemeral into transient, so pools can transition to always using this new instruction

The implementation is extremely similar to #3924, and this instruction is less important, so be sure to check this out afterwards.

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -1124,6 +1124,7 @@ impl Processor {
         accounts: &[AccountInfo],
         lamports: u64,
         transient_stake_seed: u64,
+        maybe_ephemeral_stake_seed: Option<u64>,
     ) -> ProgramResult {
         let account_info_iter = &mut accounts.iter();
         let stake_pool_info = next_account_info(account_info_iter)?;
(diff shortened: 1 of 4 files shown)


stake-pool/program/src/processor.rs
```diff
@@ -1218,38 +1246,127 @@ impl Processor {
...
         }
 
-        create_stake_account(
-            transient_stake_account_info.clone(),
-            transient_stake_account_signer_seeds,
-            system_program_info.clone(),
-        )?;
+        let maybe_split_from_account_info =
+            if let Some(ephemeral_stake_seed) = maybe_ephemeral_stake_seed {
+                let ephemeral_stake_account_info = maybe_ephemeral_stake_account_info.unwrap();
```
2501babe, line 1251: presume you want to port trents suggestion here also

2501babe:
lgtm!

stake-pool/program/src/processor.rs
joncinque: Yeah, I'll merge the other one first, rebase, then make that change, thanks for spotting

joncinque:
Merging this to give auditors a clean commit hash to look at -- post-merge reviews are always welcome!

PR #3937, opened 2022-12-20
token-cli: support CpiGuard and PermanentDelegate

2501babe:
heres what the warning looks like, happy to debate wording:

perm-del

part of #3771

### DIFF
--- a/token/cli/src/output.rs
+++ b/token/cli/src/output.rs
@@ -191,6 +191,8 @@ pub(crate) struct CliTokenAccount {
     pub(crate) is_associated: bool,
     #[serde(flatten)]
     pub(crate) account: UiTokenAccount,
+    #[serde(skip_serializing)]
+    pub(crate) has_permanent_delegate: bool,
 }
 
 impl QuietDisplay for CliTokenAccount {}
@@ -255,6 +257,22 @@ impl fmt::Display for CliTokenAccount {
         if !self.is_associated {
             writeln!(f)?;
             writeln!(f, "* Please run `spl-token gc` to clean up Aux accounts")?;
+        }
+
+        if self.has_permanent_delegate {
+            writeln!(f)?;
+            writeln!(
+                f,
+                "* {} ",
+                style("This token has a permanent delegate!").bold()
+            )?;
+            writeln!(
+                f,
+                "  This means the mint may withdraw {} funds from this account at {} time.",
+                style("all").bold(),
+                style("any").bold(),
+            )?;
+            writeln!(f, "  If this was not adequately disclosed to you, you may be dealing with a malicious mint.")?;
         }
 
         Ok(())
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1763,11 +1763,23 @@ async fn command_address(
 async fn command_display(config: &Config<'_>, address: Pubkey) -> CommandResult {
     let account_data = config.get_account_checked(&address).await?;
 
-    let decimals = if let Some(mint_address) = get_token_account_mint(&account_data.data) {
-        Some(config.get_mint_info(&mint_address, None).await?.decimals)
-    } else {
-        None
-    };
+    let (decimals, has_permanent_delegate) =
+        if let Some(mint_address) = get_token_account_mint(&account_data.data) {
+            let mint_account = config.get_account_checked(&mint_address).await?;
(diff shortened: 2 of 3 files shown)


token/cli/src/output.rs
```diff
@@ -257,6 +259,20 @@ impl fmt::Display for CliTokenAccount {
...
+            writeln!(
+                f,
+                "* {} ",
+                style("This account's mint has a permanent delegate!").bold()
+            )?;
+            writeln!(f, "  This means the mint owner or their agent may withdraw {} funds from this account at {} time.",
+            style("all").bold(),
+            style("any").bold(),
+            )?;
+            writeln!(f, "  If this was not adequatly disclosed to you, you may be dealing with a malicious mint.")?;
```
joncinque, line 273:
```suggestion
            writeln!(f, "  If this was not adequately disclosed to you, you may be dealing with a malicious token.")?;
```

token/cli/src/output.rs
```diff
@@ -257,6 +259,20 @@ impl fmt::Display for CliTokenAccount {
             writeln!(f, "* Please run `spl-token gc` to clean up Aux accounts")?;
         }
 
+        if self.has_permanent_delegate {
+            writeln!(f)?;
+            writeln!(
+                f,
+                "* {} ",
+                style("This account's mint has a permanent delegate!").bold()
```
joncinque, line 267: nit: maybe keep it simpler?
```suggestion
                style("This token mint has a permanent delegate!").bold()
```

joncinque (commented):
Looks good! Not approving so we can get someone else to look at the language

2501babe:
@CriesofCarrots would you like to bikeshed the language re: what `spl-token display` should say about permanent delegates, when checking an account? current version is:
```
This token mint has a permanent delegate!
This means the mint owner or their agent may withdraw all funds from this account at any time.
If this was not adequately disclosed to you, you may be dealing with a malicious mint.
```

token/cli/src/output.rs
```diff
@@ -257,6 +259,20 @@ impl fmt::Display for CliTokenAccount {
             writeln!(f, "* Please run `spl-token gc` to clean up Aux accounts")?;
         }
 
+        if self.has_permanent_delegate {
+            writeln!(f)?;
+            writeln!(
+                f,
+                "* {} ",
+                style("This token mint has a permanent delegate!").bold()
```
CriesofCarrots, line 267:
```suggestion
                style("This token has a permanent delegate!").bold()
```

I know Jon commented on this line before, but I got confused because it reads like the account being displayed is a mint, and it isn't. Maybe go even simpler?

token/cli/src/output.rs
```diff
@@ -257,6 +259,20 @@ impl fmt::Display for CliTokenAccount {
...
+            )?;
+            writeln!(f, "  This means the mint owner or their agent may withdraw {} funds from this account at {} time.",
```
CriesofCarrots, line 269: I got a little hung up on what "mint owner or their agent" means. Would this be too obscure/complex?

```suggestion
            writeln!(f, "  This means the mint's permanent delegate authority may withdraw {} funds from this account at {} time.",
```

CriesofCarrots (commented):
Why yes, I would love to bikeshed, ty :)
Couple thoughts, not strongly held.

token/cli/src/output.rs
2501babe: perhaps "This means the person who controls the mint may..."?

token/cli/src/output.rs
2501babe: yea i like this best actually

token/cli/src/output.rs
CriesofCarrots: Well, it might not be a single person, right? And it might not be the same authority that controls minting or freezing or closing.
But I'm not sure if greater precision and accuracy is worth it.

2501babe: yea my intended audience is nontechnicals and im trying to communicate "if you put a lot of money in this and dont know what this thing means then someone is trying to steal from you." i imagine permanent delegates are going to be used for rugs unless there is a lot of signposting about what its for (this is going to be on wallets, mostly, unfortunately)

maybe "This means the mint may withdraw"?

CriesofCarrots: > maybe "This means the mint may withdraw"?

I like this best of the options so far. :shipit: ?

CriesofCarrots (approved):
Wfm!

PR #3938, opened 2022-12-20
[confidential-extension] Restrict `InitializeMint` and `UpdateMint` instructions

samkim-crypto:
#### Problem
Currently, the `InitializeMint` instruction allows the initiator to submit the entire `ConfidentialTransferMint` state, allowing it to instantiate `withheld_amount` to any ciphertext. This allows the initiator to potentially create new tokens confidentially without affecting the mint.

The `UpdateMint` instruction allows the authority to change the `withdraw_withheld_authority_encryption_pubkey` and `withheld_amount`. As in `InitializeMint`, the authority could potentially overwrite the withheld amount and create or burn tokens privately without being detected.

#### Summary of changes
This PR restricts the two instructions above. For `InitializeMint`, the initial `withheld_amount` is always set to be the zeroed ciphertext. For `UpdateMint`, modify it so that it does not affect `withdraw_withheld_authority_encryption_pubkey` and `withheld_amount`.

### DIFF
--- a/token/program-2022/src/extension/confidential_transfer/instruction.rs
+++ b/token/program-2022/src/extension/confidential_transfer/instruction.rs
@@ -32,12 +32,15 @@ pub enum ConfidentialTransferInstruction {
     /// The instruction fails if the `TokenInstruction::InitializeMint` instruction has already
     /// executed for the mint.
     ///
+    /// Note that the `withdraw_withheld_authority_encryption_pubkey` cannot be updated after it is
+    /// initialized.
+    ///
     /// Accounts expected by this instruction:
     ///
     ///   0. `[writable]` The SPL Token mint.
     ///
     /// Data expected by this instruction:
-    ///   `ConfidentialTransferMint`
+    ///   `InitializeMintData`
     ///
     InitializeMint,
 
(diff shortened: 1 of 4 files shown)


token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -510,44 +543,71 @@ pub struct WithdrawWithheldTokensFromAccountsData {
...
-    ct_mint: &ConfidentialTransferMint,
+    authority: Option<&Pubkey>,
+    auto_approve_new_accounts: bool,
+    auditor_encryption_pubkey: &EncryptionPubkey,
+    withdraw_withheld_authority_encryption_pubkey: &EncryptionPubkey,
 ) -> Result<Instruction, ProgramError> {
     check_program_account(token_program_id)?;
     let accounts = vec![AccountMeta::new(*mint, false)];
+
+    // TODO: use `OptionalNonZeroPubkey`
```
joncinque, line 558: It's too late for this, but we should avoid new `TODO`s in the code, since they typically never get done. Can you add an issue tracking this?

token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -419,8 +425,35 @@ pub enum ConfidentialTransferInstruction {
...
+/// Data expected by `ConfidentialTransferInstruction::UpdateMint`
+#[derive(Clone, Copy, Debug, PartialEq, Pod, Zeroable)]
+#[repr(C)]
+pub struct UpdateMintData {
+    /// Determines if newly configured accounts must be approved by the `authority` before they may
+    /// be used by the user.
+    pub auto_approve_new_accounts: PodBool,
+    /// New authority to decode any transfer amount in a confidential transfer.
+    pub auditor_encryption_pubkey: EncryptionPubkey,
+}
```
joncinque, line 453: How can we set `withdraw_withheld_authority_encryption_pubkey`? Is the idea that it's only possible when there are no withheld fees, which can never be guaranteed?

joncinque (commented):
Looks clear for the most part, just a few little points

token/program-2022/src/extension/confidential_transfer/instruction.rs
samkim-crypto: Got it. The issue was created #3941.

token/program-2022/src/extension/confidential_transfer/instruction.rs
samkim-crypto: Correct, the `withdraw_withheld_authority_encryption_pubkey` can only safely be updated when there are no (confidential extension) accounts with non-zero withheld fees. I created issue #3942 discussing this issue.

PR #3955, opened 2023-01-03
Feat: SNS Realloc Name Account

stegaBOB:
Pretty straightforward PR. This PR adds the ability to realloc the SNS name account. 

Todo: 
 - [x] JS bindings
 - [ ] JS Tests


stegaBOB:
The lack of JS tests here is a bit concerning. I think I'll spend some time at some point before this is ready to go to just write a few to make sure my changes function as they should.

name-service/program/src/instruction.rs
```diff
@@ -89,6 +89,20 @@ pub enum NameRegistryInstruction {
     ///   2. `[writeable]` Refund account
     ///
     Delete,
+
+    /// Realloc the data of a name record.
+    ///
+    /// The space change cannot be more than `MAX_PERMITTED_DATA_LENGTH` different than
```
joncinque, line 95: nit for clarity, since you can reduce by any amount
```suggestion
    /// The space change cannot be more than `MAX_PERMITTED_DATA_LENGTH` greater than
```

name-service/program/src/instruction.rs
```diff
@@ -89,6 +89,20 @@ pub enum NameRegistryInstruction {
...
+    /// current `space`.
+    ///
+    /// Accounts expected by this instruction:
+    ///   0. `[writeable]` Name record to be reallocated
+    ///   1. `[signer]` Account owner
+    ///
+    Realloc {
+        /// Number of bytes of memory to allocate in addition to the `NameRecordHeader`.
```
joncinque, line 103: nit so we're sure it's the new size, and not the additional amount
```suggestion
        /// New total number of bytes in addition to the `NameRecordHeader`.
```

Also, let's add a warning that there are no checks on the data currently in the name record, so data may be truncated

name-service/program/src/instruction.rs
```diff
@@ -89,6 +89,20 @@ pub enum NameRegistryInstruction {
...
     Delete,
+
+    /// Realloc the data of a name record.
+    ///
+    /// The space change cannot be more than `MAX_PERMITTED_DATA_LENGTH` different than
+    /// current `space`.
+    ///
+    /// Accounts expected by this instruction:
+    ///   0. `[writeable]` Name record to be reallocated
+    ///   1. `[signer]` Account owner
```
joncinque, line 100: nit: could you update this comment to the actual accounts provided? and also specify that the payer account may be refunded

name-service/program/src/processor.rs
```diff
@@ -1,3 +1,6 @@
+use solana_program::rent::Rent;
+use solana_program::sysvar::Sysvar;
+use std::cmp::Ordering;
```
joncinque, line 3: nit: can you add these `use`s into the main block with the rest?

name-service/program/src/processor.rs
```diff
@@ -239,6 +242,57 @@ impl Processor {
...
+        let payer_account = next_account_info(accounts_iter)?;
+        let name_account = next_account_info(accounts_iter)?;
+        let name_owner = next_account_info(accounts_iter)?;
+
+        let name_record_header = NameRecordHeader::unpack_from_slice(&name_account.data.borrow())?;
+
+        // Verifications
+        if !name_owner.is_signer || name_record_header.owner != *name_owner.key {
+            msg!("The given name owner is incorrect or not a signer.");
+            return Err(ProgramError::InvalidArgument);
```
joncinque, line 257: nit: maybe we can use a slightly clearer error message?
```suggestion
            return Err(ProgramError::MissingRequiredSignature);
```

name-service/program/tests/functional.rs
```diff
@@ -10,6 +10,7 @@ use solana_sdk::{
     transaction::Transaction,
     transport::TransportError,
 };
+use spl_name_service::instruction::realloc;
```
joncinque, line 13: nit: can you fold this `use` into the statement right below?

name-service/program/tests/functional.rs
```diff
@@ -175,6 +176,58 @@ async fn test_name_service() {
...
+    sign_send_instruction(
+        &mut ctx,
+        update_instruction.clone(),
+        vec![&sol_subdomains_class],
+    )
+    .await
+    .unwrap_err();
+
+    #[allow(clippy::integer_arithmetic)]
+    let new_space = space * 2;
```
joncinque, line 199: nit: you can just the checked version here 
```suggestion
    let new_space = space.checked_mul(2).unwrap();
```

name-service/program/tests/functional.rs
```diff
@@ -175,6 +176,58 @@ async fn test_name_service() {
...
+    };
+
+    sign_send_instruction(&mut ctx, realloc_instruction(new_space), vec![])
+        .await
+        .unwrap();
+
+    // warping slot to resend update ix without dropping duplicate txn
+    #[allow(clippy::integer_arithmetic)]
+    let new_slot = ctx.banks_client.get_root_slot().await.unwrap() + 32;
+    ctx.warp_to_slot(new_slot).expect("warp failed");
```
joncinque, line 219: Rather than warping, you can simply fetch a new blockhash using `ctx.banks_client.get_new_latest_blockhash(&ctx.last_blockhash)` and then setting that onto the context. Or make a different instruction

name-service/js/src/bindings.ts
```diff
@@ -217,3 +218,49 @@ export async function deleteNameRegistry(
...
+ */
+export async function reallocNameAccount(
+  connection: Connection,
+  name: string,
+  space: number,
+  payerKey: PublicKey,
+  nameClass?: PublicKey,
+  nameParent?: PublicKey
+): Promise<TransactionInstruction> {
+  const hashed_name = await getHashedName(name);
```
joncinque, line 241: nit: i know this predates your PR, but in JS, let's stick with camelCase
```suggestion
  const hashedName = await getHashedName(name);
```

joncinque (commented):
Really great work! I only have some tiny nits, the implementation is solid

name-service/program/src/instruction.rs
stegaBOB: Ah that's good to know! Makes sense for what realloc is actually doing under the hood, but for whatever reason I had it in my head that it went both ways.

name-service/program/tests/functional.rs
stegaBOB: I think the reason I've been against doing that in the past is because last I checked its actually just waiting for a new block + blockhash to happen. On a large scale this can slow stuff down quite a bit. For this one instance though you're totally right. I'll change that.

name-service/program/src/processor.rs
stegaBOB: This was another instance of just copy/paste from the delete ix. Should I update those errors as well/? Doing that could technically be considered a breaking change, although I highly doubt it would break anything.

name-service/js/src/bindings.ts
stegaBOB: Good catch! Seeing JS in snake_case hurts my eyes! Switching back and forth from TS and rust really messes with the mind :rofl:

name-service/program/src/processor.rs
joncinque: Ah I see. Let's definitely keep the existing errors as they are, and then we can keep things as you've done them to be consistent. I'll retract this

name-service/js/src/bindings.ts
joncinque: That can be done in follow-up work, no need to do it here

joncinque:
do you intend to add the JS tests in this PR or with later work? Also, it looks like the `hashed_name` -> `hashedName` change wasn't applied everywhere, since my suggestion only changed the line where it's declared, and not where it's used.

stegaBOB:
Whoops! I'll fix that. As for the TS tests, I'd prefer not to write them at all haha. If you think it's fine to merge in as it is now without tests, then by all means I guess? I don't have the bandwidth right now to write TS tests from scratch so it might take little while if the tests were to be in this PR.

joncinque:
That's fine, I'll merge this as is. Do you mind adding an issue for the JS tests?

stegaBOB:
Awesome! And yep, heres the issue: https://github.com/solana-labs/solana-program-library/issues/4004

PR #3964, opened 2023-01-09
Governance: assertion for proposal final state

ochaloup:
Recreated PR (https://github.com/solana-labs/solana-program-library/pull/3879) for having an API method to check the final state of the proposal. Expected to be called from outside (from other contract).

### DIFF
--- a/governance/program/src/state/proposal.rs
+++ b/governance/program/src/state/proposal.rs
@@ -259,6 +259,23 @@ impl ProposalV2 {
         Ok(())
     }
 
+    /// Checks the Proposal was finalized (no more state transition will happen)
+    pub fn assert_is_final_state(&self) -> Result<(), ProgramError> {
+        match self.state {
+            ProposalState::Completed
+            | ProposalState::Cancelled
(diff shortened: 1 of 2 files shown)


governance/program/src/state/proposal.rs
```diff
@@ -259,6 +259,22 @@ impl ProposalV2 {
...
+            ProposalState::Completed
+            | ProposalState::Cancelled
+            | ProposalState::Defeated
+            | ProposalState::Vetoed => Ok(()),
+            ProposalState::Executing
+            | ProposalState::ExecutingWithErrors
+            | ProposalState::SigningOff
+            | ProposalState::Voting
+            | ProposalState::Draft
+            | ProposalState::Succeeded => Err(GovernanceError::InvalidStateNotFinal.into()),
```
SebastianBor, line 274: nit: Please add a comment here that in the current version due to the state transition bug non executable proposals will be stuck in `Succeeded` state

SebastianBor (approved):

PR #3965, opened 2023-01-09
example: Token transfer with PDA

### LINKED ISSUE #829: example: Provide ProgramTest example with spl-token CPI
Program developers will want to write end-to-end tests using other programs, which is not totally possible with unit tests.  `solana-program-test` is extremely useful, so we should have a small example of how to use it.  A few things to consider:

* an example of adding another program's `processor` function (e.g. spl_token)
* an example using a builtin (e.g. stake)

Also, there's an error that I ran into while doing this:
`error: the `#[global_allocator]` in solana_measure conflicts with global allocator in: spl_stake_pool`

so an extra comment or an FAQ point about that error and how to resolve it would be useful as well, e.g. adding
`#![cfg(all(target_arch = "bpf", not(feature = "no-entrypoint")))]` in entrypoint.rs to avoid defining `global_allocator` twice.

tag @jackcmay

joncinque:
#### Problem

There's no program-test example that uses spl-token, and I want to demo one.

#### Solution

Add a simple example "giver" program that simply transfers tokens from a PDA-owned account.

Fixes  #829

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5905,6 +5905,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "spl-example-transfer-tokens"
+version = "1.0.0"
+dependencies = [
+ "solana-program",
+ "solana-program-test",
+ "solana-sdk",
+ "spl-token 3.5.0",
+]
+
+[[package]]
 name = "spl-feature-proposal"
 version = "1.0.0"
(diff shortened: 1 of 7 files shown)


walla40:
$ cd <program directory>
$ cargo build-bpf --dump

PR #3966, opened 2023-01-09
ci: Add exception for tokio in audit job

joncinque:
#### Problem

There was a security vulnerability issued on tokio 1.14.  https://github.com/solana-labs/solana/pull/29585 added an exception for this advisory.

#### Solution

Do the same thing in SPL, and add the exception.


joncinque:
Audit passed, so merging this in right away!

PR #3967, opened 2023-01-10
token-js: Allow PublicKey[] | Signer[] instead of Signer[] when possible

mPaella:
Its unnecessary to pass the `Signer` when it's `secretKey` is never used. This PR changes multiple functions to allow for `Signer[] | PublicKey[]` instead of just `Signer[]`

If you would prefer me to use function overloading let me know and I can change it


token/js/src/extensions/cpiGuard/instructions.ts
```diff
@@ -70,7 +70,11 @@ function createCpiGuardInstruction(
     const keys = [{ pubkey: account, isSigner: false, isWritable: true }];
     keys.push({ pubkey: authority, isSigner: !multiSigners.length, isWritable: false });
     for (const signer of multiSigners) {
-        keys.push({ pubkey: signer.publicKey, isSigner: true, isWritable: false });
+        keys.push({
+            pubkey: signer instanceof PublicKey ? signer : signer.publicKey,
+            isSigner: true,
+            isWritable: false,
+        });
     }
```
joncinque, line 78: This predates your PR, but could all of the instances that *don't* call `addSigners` call that instead? so this would become:

```
const keys = addSigners(
        [
            { pubkey: account, isSigner: false, isWritable: true },
        ],
        authority,
        multiSigners
);
```

joncinque (commented):
Great work! Just a comment that'll make this a bit DRY-er by using `addSigners` some more, and after that I think this is good to go.

token/js/src/instructions/initializeMultisig.ts
```diff
@@ -43,7 +44,11 @@ export function createInitializeMultisigInstruction(
         { pubkey: SYSVAR_RENT_PUBKEY, isSigner: false, isWritable: false },
     ];
     for (const signer of signers) {
-        keys.push({ pubkey: signer, isSigner: false, isWritable: false });
+        keys.push({
```
mPaella, line 47: Using `addSigners` here would result in a different keys array, which makes me think these keys aren't correct...

token/js/src/extensions/cpiGuard/instructions.ts
mPaella: Updated!

token/js/src/instructions/initializeMultisig.ts
joncinque: Correct, you can't make the change here

token/js/src/instructions/internal.ts
```diff
@@ -1,11 +1,20 @@
-import type { AccountMeta, PublicKey, Signer } from '@solana/web3.js';
+import type { AccountMeta, Signer } from '@solana/web3.js';
+import { PublicKey } from '@solana/web3.js';
 
 /** @internal */
-export function addSigners(keys: AccountMeta[], ownerOrAuthority: PublicKey, multiSigners: Signer[]): AccountMeta[] {
+export function addSigners(
+    keys: AccountMeta[],
+    ownerOrAuthority: PublicKey,
+    multiSigners: Signer[] | PublicKey[]
```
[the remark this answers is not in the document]
mPaella: as well as changing the other method signatures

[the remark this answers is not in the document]
joncinque: I think `Signer[] | PublicKey[]` looks nicer, but that's entirely personal. Probably `(Signer | PublicKey)[]` is more correct though, so we can go with that

joncinque (approved):
Thanks so much for your contribution!

mPaella:
> Thanks so much for your contribution!

Would love to release this soon, if possible 🙏

joncinque:
@mPaella done! https://www.npmjs.com/package/@solana/spl-token

PR #3974, opened 2023-01-11
single-validator stake pool program

2501babe:
this is obviously a very rough sketch, but just to give an idea of what i had in mind re: "what if the 1lamp minimum stays and we did a new single-validator stake pool program instead of an automated manager on top of the existing one"

welcome to the stake birdbath

```
hana@laptop:stake-pool$ wc program/src/*
   366   1222  12000 program/src/big_vec.rs
    26     59    711 program/src/entrypoint.rs
   154    530   5592 program/src/error.rs
  1736   5652  64508 program/src/instruction.rs
   175    552   5369 program/src/lib.rs
  3950   9342 167792 program/src/processor.rs
  1329   4483  49937 program/src/state.rs
  7736  21840 305909 total
hana@laptop:stake-pool$ wc birdbath/src/*
   26    59   711 birdbath/src/entrypoint.rs
  154   530  5592 birdbath/src/error.rs
  261   648  8991 birdbath/src/instruction.rs
   45   135  1352 birdbath/src/lib.rs
  974  2817 38164 birdbath/src/processor.rs
 1460  4189 54810 total
```

when i said "reduced attack surface" this is basically what i meant. we can get rid of ~80% of the code, including all the tricky calculations, fees, multiple stake accounts, different validators, all kinds of stuff just gets deleted. bigvec and all state go away (although i might stick the validator pubkey on the signing authority just to make the code "feel" safer)

right now this only has five instructions: initialize, deposit, withdraw, create metaplex metadata, update metaplex metadata. i realized while doing my deletions that the fourth instruction can go away too, and we can just stick that in initialize

a possible fifth (nee sixth) instruction would be a mechanism for a user to deposit sol, to give a "click to stake" ux where they dont need to run two transactions. i had a few ideas of how to do this, but the common theme is they create a stake account, delegate it, and assign its authorities to the program. these transient stake accounts could be swept into the main pool account permissionlessly, and tokens would be minted to the user when this happens, so no fees and no reserve are required

obvi no need to "review" just curious what you think. lines 395-720 of processor.rs are the interesting bits

part of #4077

### DIFF
--- a/stake-pool/single-pool/Cargo.toml
+++ b/stake-pool/single-pool/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "spl-single-validator-pool"
+version = "1.0.0"
+description = "Solana Program Library Single-Validator Stake Pool"
+authors = ["Solana Labs Maintainers <maintainers@solanalabs.com>"]
+repository = "https://github.com/solana-labs/solana-program-library"
+license = "Apache-2.0"
+edition = "2021"
+
+[features]
+no-entrypoint = []
+test-sbf = []
+
+[dependencies]
+arrayref = "0.3.6"
+borsh = "0.9"
+mpl-token-metadata =  { version = "1.7.0", features = [ "no-entrypoint" ] }
+num-derive = "0.3"
+num-traits = "0.2"
+num_enum = "0.5.9"
+solana-program = "1.14.12"
+spl-token = { version = "3.5", path = "../../token/program", features = [ "no-entrypoint" ] }
+spl-associated-token-account = { version = "1.1", path = "../../associated-token-account/program", features = [ "no-entrypoint" ] }
+thiserror = "1.0"
+
+[lib]
+crate-type = ["cdylib", "lib"]
--- a/stake-pool/single-pool/src/instruction.rs
+++ b/stake-pool/single-pool/src/instruction.rs
@@ -1,0 +1,386 @@
+//! Instruction types
+
+#![allow(clippy::too_many_arguments)]
+
+use {
+    crate::{
+        find_default_deposit_account_address, find_pool_authority_address, find_pool_mint_address,
+        find_pool_stake_address, USER_STAKE_SEED,
+    },
+    borsh::{BorshDeserialize, BorshSerialize},
+    mpl_token_metadata::pda::find_metadata_account,
+    solana_program::{
+        instruction::{AccountMeta, Instruction},
+        program_pack::Pack,
+        pubkey::Pubkey,
+        rent::Rent,
+        stake, system_instruction, system_program, sysvar,
+    },
+};
+
+/// Instructions supported by the SinglePool program.
+#[repr(C)]
+#[derive(Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize)]
+pub enum SinglePoolInstruction {
+    ///   Initialize the mint and stake account for a new single-validator pool.
+    ///   The pool stake account must contain the rent-exempt minimum plus the minimum delegation.
+    ///   No tokens will be minted: to deposit more, use `Deposit` after `InitializeStake`.
+    ///
+    ///   0. `[]` Validator vote account
+    ///   1. `[w]` Pool stake account
+    ///   2. `[]` Pool authority
+    ///   3. `[w]` Pool token mint
+    ///   4. `[]` Rent sysvar
+    ///   5. `[]` Clock sysvar
+    ///   6. `[]` Stake history sysvar
+    ///   7. `[]` Stake config sysvar
+    ///   8. `[]` System program
+    ///   9. `[]` Token program
+    ///  10. `[]` Stake program
+    InitializePool,
+
+    ///   Deposit stake into the pool.  The output is a "pool" token representing fractional
+    ///   ownership of the pool stake. Inputs are converted to the current ratio.
+    ///
+    ///   0. `[w]` Pool stake account
+    ///   1. `[]` Pool authority
+    ///   2. `[w]` Pool token mint
+    ///   3. `[w]` User stake account to join to the pool
+    ///   4. `[w]` User account to receive pool tokens
+    ///   5. `[w]` User account to receive lamports
+    ///   6. `[]` Clock sysvar
+    ///   7. `[]` Stake history sysvar
+    ///   8. `[]` Token program
+    ///   9. `[]` Stake program
+    DepositStake {
+        /// Validator vote account address
+        vote_account_address: Pubkey,
+    },
+
+    ///   Redeem tokens issued by this pool for stake at the current ratio.
+    ///
+    ///   0. `[w]` Pool stake account
+    ///   1. `[]` Pool authority
+    ///   2. `[w]` Pool token mint
+    ///   3. `[w]` User stake account to receive stake at
+    ///   4. `[w]` User account to take pool tokens from
+    ///   5. `[]` Clock sysvar
+    ///   6. `[]` Token program
+    ///   7. `[]` Stake program
+    WithdrawStake {
+        /// Validator vote account address
+        vote_account_address: Pubkey,
+        /// User authority for the new stake account
+        user_stake_authority: Pubkey,
+        /// Amount of tokens to redeem for stake
+        token_amount: u64,
+    },
+
+    ///   Create token metadata for the stake-pool token in the metaplex-token program.
+    ///   Step three of the permissionless three-stage initialization flow.
+    ///   Note this instruction is not necessary for the pool to operate, to ensure we cannot
+    ///   be broken by upstream.
+    ///
+    ///   0. `[]` Pool authority
+    ///   1. `[]` Pool token mint
+    ///   2. `[s, w]` Payer for creation of token metadata account
+    ///   3. `[w]` Token metadata account
+    ///   4. `[]` Metadata program id
+    ///   5. `[]` System program id
+    CreateTokenMetadata {
+        /// Validator vote account address
+        vote_account_address: Pubkey,
+    },
+
+    ///   Update token metadata for the stake-pool token in the metaplex-token program.
+    ///
+    ///   0. `[]` Validator vote account
+    ///   1. `[]` Pool authority
+    ///   2. `[s]` Vote account authorized withdrawer
+    ///   3. `[w]` Token metadata account
+    ///   4. `[]` Metadata program id
+    UpdateTokenMetadata {
+        /// Token name
+        name: String,
+        /// Token symbol e.g. stkSOL
+        symbol: String,
+        /// URI of the uploaded metadata of the spl-token
+        uri: String,
+    },
+}
+
+/// Creates all necessary instructions to initialize the stake pool.
+pub fn initialize(
+    program_id: &Pubkey,
+    vote_account: &Pubkey,
+    payer: &Pubkey,
+    rent: &Rent,
+    minimum_delegation: u64,
+) -> Vec<Instruction> {
+    let stake_address = find_pool_stake_address(program_id, vote_account);
+    let stake_space = std::mem::size_of::<stake::state::StakeState>();
+    let stake_rent_plus_one = rent
+        .minimum_balance(stake_space)
+        .saturating_add(minimum_delegation);
+
+    let mint_address = find_pool_mint_address(program_id, vote_account);
+    let mint_rent = rent.minimum_balance(spl_token::state::Mint::LEN);
+
+    vec![
+        system_instruction::transfer(payer, &stake_address, stake_rent_plus_one),
+        system_instruction::transfer(payer, &mint_address, mint_rent),
+        initialize_pool(program_id, vote_account),
+        create_token_metadata(program_id, vote_account, payer),
+    ]
+}
+
+/// Creates an `InitializePool` instruction.
+pub fn initialize_pool(program_id: &Pubkey, vote_account: &Pubkey) -> Instruction {
+    let mint_address = find_pool_mint_address(program_id, vote_account);
+
+    let data = SinglePoolInstruction::InitializePool.try_to_vec().unwrap();
+    let accounts = vec![
+        AccountMeta::new_readonly(*vote_account, false),
+        AccountMeta::new(find_pool_stake_address(program_id, vote_account), false),
+        AccountMeta::new_readonly(find_pool_authority_address(program_id, vote_account), false),
+        AccountMeta::new(mint_address, false),
+        AccountMeta::new_readonly(sysvar::rent::id(), false),
+        AccountMeta::new_readonly(sysvar::clock::id(), false),
+        AccountMeta::new_readonly(sysvar::stake_history::id(), false),
+        AccountMeta::new_readonly(stake::config::id(), false),
+        AccountMeta::new_readonly(system_program::id(), false),
+        AccountMeta::new_readonly(spl_token::id(), false),
+        AccountMeta::new_readonly(stake::program::id(), false),
+    ];
+
+    Instruction {
+        program_id: *program_id,
+        accounts,
+        data,
+    }
+}
+
+/// Creates all necessary instructions to deposit stake.
+pub fn deposit(
+    program_id: &Pubkey,
+    vote_account: &Pubkey,
+    user_stake_account: &Pubkey,
+    user_token_account: &Pubkey,
+    user_lamport_account: &Pubkey,
+    user_withdraw_authority: &Pubkey,
+) -> Vec<Instruction> {
+    let pool_authority = find_pool_authority_address(program_id, vote_account);
+
+    vec![
+        stake::instruction::authorize(
+            user_stake_account,
+            user_withdraw_authority,
+            &pool_authority,
+            stake::state::StakeAuthorize::Staker,
+            None,
+        ),
+        stake::instruction::authorize(
+            user_stake_account,
+            user_withdraw_authority,
+            &pool_authority,
+            stake::state::StakeAuthorize::Withdrawer,
+            None,
+        ),
+        deposit_stake(
+            program_id,
+            vote_account,
+            user_stake_account,
+            user_token_account,
+            user_lamport_account,
+        ),
+    ]
+}
+
+/// Creates a `DepositStake` instruction.
+pub fn deposit_stake(
+    program_id: &Pubkey,
+    vote_account: &Pubkey,
+    user_stake_account: &Pubkey,
+    user_token_account: &Pubkey,
+    user_lamport_account: &Pubkey,
+) -> Instruction {
+    let data = SinglePoolInstruction::DepositStake {
+        vote_account_address: *vote_account,
+    }
+    .try_to_vec()
+    .unwrap();
+
+    let accounts = vec![
+        AccountMeta::new(find_pool_stake_address(program_id, vote_account), false),
+        AccountMeta::new_readonly(find_pool_authority_address(program_id, vote_account), false),
+        AccountMeta::new(find_pool_mint_address(program_id, vote_account), false),
+        AccountMeta::new(*user_stake_account, false),
+        AccountMeta::new(*user_token_account, false),
+        AccountMeta::new(*user_lamport_account, false),
+        AccountMeta::new_readonly(sysvar::clock::id(), false),
+        AccountMeta::new_readonly(sysvar::stake_history::id(), false),
+        AccountMeta::new_readonly(spl_token::id(), false),
+        AccountMeta::new_readonly(stake::program::id(), false),
+    ];
+
+    Instruction {
+        program_id: *program_id,
+        accounts,
+        data,
+    }
+}
+
+/// Creates all necessary instructions to withdraw stake into a given stake account.
+/// If a new stake account is required, the user should first include `system_instruction::create_account`
+/// with account size `std::mem::size_of::<stake::state::StakeState>()` and owner `stake::program::id()`.
+pub fn withdraw(
+    program_id: &Pubkey,
+    vote_account: &Pubkey,
+    user_stake_account: &Pubkey,
+    user_stake_authority: &Pubkey,
+    user_token_account: &Pubkey,
+    user_token_authority: &Pubkey,
+    token_amount: u64,
+) -> Vec<Instruction> {
+    let pool_authority = find_pool_authority_address(program_id, vote_account);
+
+    vec![
+        spl_token::instruction::approve(
+            &spl_token::id(),
+            user_token_account,
+            &pool_authority,
+            user_token_authority,
+            &[],
+            token_amount,
+        )
+        .unwrap(),
+        withdraw_stake(
+            program_id,
+            vote_account,
+            user_stake_account,
+            user_stake_authority,
+            user_token_account,
+            token_amount,
+        ),
+    ]
+}
+
+/// Creates a `WithdrawStake` instruction.
+pub fn withdraw_stake(
+    program_id: &Pubkey,
+    vote_account: &Pubkey,
+    user_stake_account: &Pubkey,
+    user_stake_authority: &Pubkey,
+    user_token_account: &Pubkey,
+    token_amount: u64,
+) -> Instruction {
+    let data = SinglePoolInstruction::WithdrawStake {
+        vote_account_address: *vote_account,
+        user_stake_authority: *user_stake_authority,
+        token_amount,
+    }
+    .try_to_vec()
+    .unwrap();
+
+    let accounts = vec![
+        AccountMeta::new(find_pool_stake_address(program_id, vote_account), false),
+        AccountMeta::new_readonly(find_pool_authority_address(program_id, vote_account), false),
+        AccountMeta::new(find_pool_mint_address(program_id, vote_account), false),
+        AccountMeta::new(*user_stake_account, false),
+        AccountMeta::new(*user_token_account, false),
+        AccountMeta::new_readonly(sysvar::clock::id(), false),
+        AccountMeta::new_readonly(spl_token::id(), false),
+        AccountMeta::new_readonly(stake::program::id(), false),
+    ];
+
+    Instruction {
+        program_id: *program_id,
+        accounts,
+        data,
+    }
+}
+
+/// Creates necessary instructions to create and delegate a new stake account to a given validator.
+/// Uses a fixed address for each wallet and vote account combination to make it easier to find for deposits.
+/// This is an optional helper function; deposits can come from any owned stake account without lockup.
+pub fn create_and_delegate_user_stake(
+    vote_account: &Pubkey,
+    user_wallet: &Pubkey,
+    lamports: u64,
+) -> Vec<Instruction> {
+    stake::instruction::create_account_with_seed_and_delegate_stake(
+        user_wallet,
+        &find_default_deposit_account_address(vote_account, user_wallet),
+        vote_account,
+        USER_STAKE_SEED,
+        vote_account,
+        &stake::state::Authorized {
+            staker: *user_wallet,
+            withdrawer: *user_wallet,
+        },
+        &stake::state::Lockup::default(),
+        lamports,
+    )
+}
+
+/// Creates a `CreateTokenMetadata` instruction.
+pub fn create_token_metadata(
+    program_id: &Pubkey,
+    vote_account: &Pubkey,
+    payer: &Pubkey,
+) -> Instruction {
+    let pool_authority = find_pool_authority_address(program_id, vote_account);
+    let pool_mint = find_pool_mint_address(program_id, vote_account);
+    let (token_metadata, _) = find_metadata_account(&pool_mint);
+    let data = SinglePoolInstruction::CreateTokenMetadata {
+        vote_account_address: *vote_account,
+    }
+    .try_to_vec()
+    .unwrap();
+
+    let accounts = vec![
+        AccountMeta::new_readonly(pool_authority, false),
+        AccountMeta::new_readonly(pool_mint, false),
+        AccountMeta::new(*payer, true),
+        AccountMeta::new(token_metadata, false),
+        AccountMeta::new_readonly(mpl_token_metadata::id(), false),
+        AccountMeta::new_readonly(system_program::id(), false),
+    ];
+
+    Instruction {
+        program_id: *program_id,
+        accounts,
+        data,
+    }
+}
+
+/// Creates an `UpdateTokenMetadata` instruction.
+pub fn update_token_metadata(
+    program_id: &Pubkey,
+    vote_account: &Pubkey,
+    authorized_withdrawer: &Pubkey,
+    name: String,
+    symbol: String,
+    uri: String,
+) -> Instruction {
+    let pool_authority = find_pool_authority_address(program_id, vote_account);
+    let pool_mint = find_pool_mint_address(program_id, vote_account);
+    let (token_metadata, _) = find_metadata_account(&pool_mint);
+    let data = SinglePoolInstruction::UpdateTokenMetadata { name, symbol, uri }
+        .try_to_vec()
+        .unwrap();
+
+    let accounts = vec![
+        AccountMeta::new_readonly(*vote_account, false),
+        AccountMeta::new_readonly(pool_authority, false),
+        AccountMeta::new_readonly(*authorized_withdrawer, true),
+        AccountMeta::new(token_metadata, false),
+        AccountMeta::new_readonly(mpl_token_metadata::id(), false),
+    ];
+
+    Instruction {
+        program_id: *program_id,
+        accounts,
+        data,
+    }
+}
--- a/stake-pool/single-pool/src/lib.rs
+++ b/stake-pool/single-pool/src/lib.rs
@@ -1,0 +1,82 @@
+#![deny(missing_docs)]
+
+//! A program for liquid staking with a single validator
+
+pub mod error;
+pub mod instruction;
+pub mod processor;
+
+#[cfg(not(feature = "no-entrypoint"))]
+pub mod entrypoint;
+
+// export current sdk types for downstream users building with a different sdk version
+pub use solana_program;
+use solana_program::pubkey::Pubkey;
+
+// XXX TODO FIXME change this
+// (XXX ask how do we as a company handle privkeys for our onchain programs?)
+solana_program::declare_id!("3cqnsMsT6LE96pxv7GR4di5rLqHDZZbR3FbeSUeRLFqY");
+
+const POOL_STAKE_PREFIX: &[u8] = b"stake";
+const POOL_AUTHORITY_PREFIX: &[u8] = b"authority";
+const POOL_MINT_PREFIX: &[u8] = b"mint";
+
+const MINT_DECIMALS: u8 = 9;
+
+// authorized withdrawer starts immediately after the enum tag
+const VOTE_STATE_START: usize = 4;
+const VOTE_STATE_END: usize = 36;
+
+const USER_STAKE_SEED: &str = "single-pool-user-stake";
+
+fn find_address_and_bump(
+    program_id: &Pubkey,
+    vote_account_address: &Pubkey,
+    prefix: &[u8],
+) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&[prefix, vote_account_address.as_ref()], program_id)
+}
+
+fn find_pool_stake_address_and_bump(
+    program_id: &Pubkey,
+    vote_account_address: &Pubkey,
+) -> (Pubkey, u8) {
+    find_address_and_bump(program_id, vote_account_address, POOL_STAKE_PREFIX)
+}
+
+fn find_pool_authority_address_and_bump(
+    program_id: &Pubkey,
+    vote_account_address: &Pubkey,
+) -> (Pubkey, u8) {
+    find_address_and_bump(program_id, vote_account_address, POOL_AUTHORITY_PREFIX)
+}
+
+fn find_pool_mint_address_and_bump(
+    program_id: &Pubkey,
+    vote_account_address: &Pubkey,
+) -> (Pubkey, u8) {
+    find_address_and_bump(program_id, vote_account_address, POOL_MINT_PREFIX)
+}
+
+/// Find the canonical stake account address for a given vote account.
+pub fn find_pool_stake_address(program_id: &Pubkey, vote_account_address: &Pubkey) -> Pubkey {
+    find_pool_stake_address_and_bump(program_id, vote_account_address).0
+}
+
+/// Find the canonical authority address for a given vote account.
+pub fn find_pool_authority_address(program_id: &Pubkey, vote_account_address: &Pubkey) -> Pubkey {
+    find_pool_authority_address_and_bump(program_id, vote_account_address).0
+}
+
+/// Find the canonical token mint address for a given vote account.
+pub fn find_pool_mint_address(program_id: &Pubkey, vote_account_address: &Pubkey) -> Pubkey {
+    find_pool_mint_address_and_bump(program_id, vote_account_address).0
+}
+
+/// Find the address of the default intermediate account that holds activating user stake before deposit.
+pub fn find_default_deposit_account_address(
+    vote_account_address: &Pubkey,
+    user_wallet_address: &Pubkey,
+) -> Pubkey {
+    Pubkey::create_with_seed(vote_account_address, USER_STAKE_SEED, user_wallet_address).unwrap()
+}
--- a/stake-pool/single-pool/src/processor.rs
+++ b/stake-pool/single-pool/src/processor.rs
@@ -1,0 +1,1031 @@
+//! program state processor
+
+use {
+    crate::{
+        error::SinglePoolError, instruction::SinglePoolInstruction, MINT_DECIMALS,
+        POOL_AUTHORITY_PREFIX, POOL_MINT_PREFIX, POOL_STAKE_PREFIX, VOTE_STATE_END,
+        VOTE_STATE_START,
+    },
+    borsh::BorshDeserialize,
+    mpl_token_metadata::{
+        instruction::{create_metadata_accounts_v3, update_metadata_accounts_v2},
+        pda::find_metadata_account,
+        state::DataV2,
+    },
+    solana_program::{
+        account_info::{next_account_info, AccountInfo},
+        borsh::try_from_slice_unchecked,
+        entrypoint::ProgramResult,
+        msg,
+        native_token::LAMPORTS_PER_SOL,
+        program::invoke_signed,
+        program_error::ProgramError,
+        program_pack::Pack,
+        pubkey::Pubkey,
+        rent::Rent,
+        stake::{
+            self,
+            state::{Meta, Stake, StakeState},
+        },
+        stake_history::Epoch,
+        system_instruction, system_program,
+        sysvar::{clock::Clock, Sysvar},
+        vote::program as vote_program,
+    },
+    spl_token::state::Mint,
+};
+
+/// Calculate pool tokens to mint, given outstanding token supply, pool active stake, and deposit active stake
+fn calculate_deposit_amount(
+    pre_token_supply: u64,
+    pre_pool_stake: u64,
+    user_stake_to_deposit: u64,
+) -> Option<u64> {
+    if pre_pool_stake == 0 || pre_token_supply == 0 {
+        Some(user_stake_to_deposit)
+    } else {
+        u64::try_from(
+            (user_stake_to_deposit as u128)
+                .checked_mul(pre_token_supply as u128)?
+                .checked_div(pre_pool_stake as u128)?,
+        )
+        .ok()
+    }
+}
+
+/// Calculate pool stake to return, given outstanding token supply, pool active stake, and tokens to redeem
+fn calculate_withdraw_amount(
+    pre_token_supply: u64,
+    pre_pool_stake: u64,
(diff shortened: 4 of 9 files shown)


2501babe:
ok i was gonna keep dev in a new branch but i guess theres no real point. initialize works and deposit nearly works, ive managed to keep basically the same interface as i had originally but just a few fewer necessary accounts. nothing terribly exciting, will continue after i catch up on code reviews

2501babe:
alright, initialize/deposit/withdraw all seem to basically work. next steps are:
* i would like to fold create metadata into initialize
* fix up error.rs, which is unchanged from the multistake pool code. many variants can go away, some new ones need to be added
* port over tests in a more disciplined manner, the existing stuff is just sanity checks while i got the processor actually working. maybe theres a way we can abstract the test code to run the relevant subset of tests against both tho, without having to copy-paste... will think about this

stake-pool/birdbath/src/processor.rs
```diff
@@ -0,0 +1,1100 @@
...
+
+        let mint_seeds = &[
+            POOL_MINT_PREFIX,
+            validator_vote_info.key.as_ref(),
+            &[mint_bump_seed],
+        ];
+        let mint_signers = &[&mint_seeds[..]];
+
+        // change to Rent::get() if i get rid of the invokes that require the AccountInfo
+        let rent = &Rent::from_account_info(rent_info)?;
```
joncinque, line 504: We'll need a version of `initialize` on the stake program that doesn't require the rent account... but that can be added!

stake-pool/birdbath/src/processor.rs
```diff
@@ -0,0 +1,1100 @@
...
+                validator_vote_info.clone(),
+                clock_info.clone(),
+                stake_history_info.clone(),
+                stake_config_info.clone(),
+                pool_authority_info.clone(),
+            ],
+            authority_signers,
+        )?;
+
+        // could mint the token here if we wanted, either to user or incinerator
```
joncinque, line 595: It would be economically sound to give pool tokens for the stake account minimum stake, at the very least, since they'll be giving rent-exemption to make the mint and stake account.

stake-pool/birdbath/src/processor.rs
```diff
@@ -0,0 +1,1100 @@
...
+
+        Ok(())
+    }
+
+    // XXX ok cool next up ummm
+    // the other two functions are extremely simplified version of their namesakes
+    // for deposit we literally only need to call stake_merge (or a hana version), not authorize
+    // because the user can et both authorities to ours rather than going through deposit authority
```
joncinque, line 603: This does simplify things a lot! And since you check the address of the destination, I think you're totally safe. We'll need to make sure it doesn't open up any attacks from people trying to merge accounts owned by the pool, but I don't think it's possible

stake-pool/birdbath/src/processor.rs
```diff
@@ -0,0 +1,1100 @@
...
+            vote_account_address,
+            pool_stake_info.clone(),
+            pool_authority_info.clone(),
+            bump_seed,
+            user_lamport_account_info.clone(),
+            clock_info.clone(),
+            stake_history_info.clone(),
+            stake_program_info.clone(),
+            leftover_rent,
+        )?;
```
joncinque, line 745: This is extremely touchy logic -- in this case, by just sending the lamports back, it's possible for someone to take their rent.

During `merge`, the absorbed account gets set to `Uninitialized`, essentially zero'ed out, so there's no more data about who's authorized on the account. When the lamports get sent back, it's just an uninitialized stake with lamports, so anyone can call `initialize` on it to take it back. For them to keep their stake, we need to initialize it again after sending the lamports.

stake-pool/birdbath/src/processor.rs
```diff
@@ -0,0 +1,1100 @@
...
+        }
+
+        check_system_program(system_program_info.key)?;
+        check_rent_sysvar(rent_sysvar_info.key)?;
+        check_account_owner(payer_info, &system_program::id())?;
+        check_mpl_metadata_program(mpl_token_metadata_program_info.key)?;
+
+        /* XXX HANA commenting out pool/manager validation because i want to delete the stakepool struct
+         * the way this would work is we have a sensible default for each token that says what it is
+         * and probably allow the owner of the validator to give it a more colorful description
```
joncinque, line 868: From what I've heard, the validators do want that control

joncinque (commented):
I just took a quick pass at it. Looks like it's definitely on the right track! I think we can definitely refactor the tests to work for both programs. You'll need to move all of the stake pool tests into a new crate like `token/program-2022-test`, use `test_case` to run certain tests for both programs, and have `StakePoolAccounts` also support pool stakes. There might be some other way, but that was my first thought.

stake-pool/birdbath/src/processor.rs
2501babe: this sends it back to the user wallet, not the uninitialized stake account. just in case, i made sure this actually works:
```
HANA lamps before staking: 999997955720
     lamps after staking: 998495657840 (1502297880 less than before, 2297880 excluding stake)
     lamps after deposit: 998497935720 (2277880 more than before)
```

joncinque: Oops, I misread the variables, my bad!

stake-pool/birdbath/src/processor.rs
2501babe: yea theres definitely test cases id like to cover of users trying to do tricky shit but... im pretty sure this is all safe, because ultimately every vote account gets one and only one canonical stake account, and every stake account has a different authority

stake-pool/birdbath/src/processor.rs
2501babe: oh i only take 1 lamport from the initializer and assume theyll deposit more normally. i kinda like the edge case protection that theres always one lamport unspoken for by any tokens in circulation

stake-pool/birdbath/src/processor.rs
2501babe: as a style thing, or is there a technical reason? if we want a version that doesnt take rent, id rather just convert `initialize` back to the old style (do the work in instructions then validate in `initialize`). i like this way because its safer but theres no need to have two functions regardless

2501babe:
thank you for the close look!

2501babe:
had trouble reworking tests to be generic over both pool... i was trying to do it bottom-up by modifying existing code in-place, but im starting over and trying the other way, from the top down starting with new structs/traits and porting over code as it can fit

also... naming. the two ideas i had so far are:
* umbrella term: "stake pool", existing program: "multi-pool", new program: "single-pool"
* umbrella: "stake intermediary", existing: "stake pool", new: "stake passthrough"
* umbrella: "stake pool", existing: "single-validator pool", new: "multi-validator pool"

i dont really like either especially much. one of the annoying things is anything which emphasizes the "poolness" of the existing program or the "singularness" of the new one is misleading, because theyre both pooling _stake_, its just one is also pooling _validators_. really the difference is one is a mux and one is a mux+demux

maybe "managed" and "permissionless" pools... dunno. something to keep thinking about

2501babe:
btw @joncinque did you have an opinion on whether i should change my initialize instruction to function like the existing stake pool (do work in transaction-level instructions and validate in program instruction) or if the way i impled it (do work in program instruction to eliminate necessity of validation) is preferable?

joncinque:
> umbrella term: "stake pool", existing program: "multi-pool", new program: "single-pool"

I like the simplicity of these, but that's also because I've had stake pools on the brain for too long. But maybe the official term is "multi-validator pool" and "single-validator pool" and we can shorten to "single" and "multi" as we like. I haven't been able to come up with anything better either...

> btw @joncinque did you have an opinion on whether i should change my initialize instruction to function like the existing stake pool (do work in transaction-level instructions and validate in program instruction) or if the way i impled it (do work in program instruction to eliminate necessity of validation) is preferable?

Typically it's best to do things at the transaction-level, but the mint and stake account for single pools are on PDAs, so we can't do `create_account` on them outside. Are you specifically referring to the lamport transfer portion for rent-exemption? For those, most programs take in a signing payer to fund, like associated-token-account and address-lookup-tables, but to be honest, I think it's better practice to fund the accounts at the top-level of the transaction and not pass in a payer, ie. your transaction is `[system_instruction::transfer, system_instruction::transfer, single_pool::initialize]`. What do you think?

And ok, I'm warming up to "single pool"

2501babe:
experimenting with converting an existing pool test to be a multi-only case in the new framework, looks pretty good ngl

```diff
24,33c25,27
< async fn setup() -> (BanksClient, Keypair, Hash, StakePoolAccounts, Keypair) {
<     let (mut banks_client, payer, recent_blockhash) = program_test().start().await;
<     let stake_pool_accounts = StakePoolAccounts::default();
<     stake_pool_accounts
<         .initialize_stake_pool(
<             &mut banks_client,
<             &payer,
<             &recent_blockhash,
<             MINIMUM_RESERVE_LAMPORTS,
<         )
---
> async fn setup(env: &Env) -> (BanksClient, Keypair, Hash, Keypair) {
>     let (mut banks_client, payer, recent_blockhash) = env.program_test().start().await;
>     env.initialize(&mut banks_client, &payer, &recent_blockhash)
39,45c33
<     (
<         banks_client,
<         payer,
<         recent_blockhash,
<         stake_pool_accounts,
<         new_staker,
<     )
---
>     (banks_client, payer, recent_blockhash, new_staker)
47a36,37
> #[test_case(EnvBuilder::MultiPoolTokenkeg.env() ; "multi-pool tokenkeg")]
> #[test_case(EnvBuilder::MultiPoolToken22.env() ; "multi-pool token22")]
49,51c39,41
< async fn success_set_staker_as_manager() {
<     let (mut banks_client, payer, recent_blockhash, stake_pool_accounts, new_staker) =
<         setup().await;
---
> async fn success_set_staker_as_manager(env: Env) {
>     if let Env::MultiPool(ref stake_pool_accounts) = env {
>         let (mut banks_client, payer, recent_blockhash, new_staker) = setup(&env).await;
65c55,56
<     let stake_pool = get_account(&mut banks_client, &stake_pool_accounts.stake_pool.pubkey()).await;
---
>         let stake_pool =
>             get_account(&mut banks_client, &stake_pool_accounts.stake_pool.pubkey()).await;
69a61
>     }
```

2501babe:
the file naming scheme like `multi_only-set_staker.rs` annoys me but it is a (ideally temporary) cope for the fact that no amount of `tests/main.rs`, `tests/multi/main.rs`, `tests/multi/mod.rs` and so on bullshitting i tried resulted in something that compiled and ran all tests lol, the crate/package/module/compilation unit voodoo of rust still continues to elude me sometimes

stake-pool/stake-pool-test/tests/helpers/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+                    .await
+            }
+        }
+    }
+}
+
+// XXX ok im confused about parametrization again
+// * if i do an Accounts trait, and non-generic Env, i impl initialize on the trait
+//   env only exists to produce the accounts
+//   but because i dont return a concrete type i need to work in separate branch arms in tests?
```
joncinque, line 182: That depends. You can either use `dyn MyTraitType`, or an enum-with-value that does some form of `enum_dispatch` in its implementation, which is what `Env` is doing now. Since we don't need to worry about performance, it might simpler to use `Box<dyn MyTraitType>` (or Arc).

I'm probably missing some bits, but here's a potential approach. You define a trait with all of the functionality required for single pools only, ie. deposit, withdraw, initialize, create metadata, update metadata, along with functions to give you the important addresses (mint, stake account, etc). You have two concrete types, `SinglePool` and `MultiPool`, which each implement that minimal interface.

For tests on both single and multipool, you run everything through the trait. Since you have functions that give you the addresses, you should (?) be able to check through that.

For tests on just one or the other, you can directly return the concrete type from your builder, or expose `unwrap_multi()` / `unwrap_single()` functions to pull out the concrete type, and just work on that. That way you don't need to wrap tests in an `if let Env::(...)`

joncinque (commented):
I like the direction this is going in!

2501babe:
alright status of this is... i need `VoteState::deserialize` to get a pubkey off the vote account to match with a signer for `update_metadata` so that the owner of the validator can change mpl metadata. this requires monorepo v1.15. unfortunatly a) there is more work to do (#4057) to update spl for v1.15, and b) `VoteState::deserialize` in v1.15 would blow the stack if called, and presently all importers of `solana_program` get an error to this effect from `cargo build-sbf`, so that should be fixed sooner rather than later aside from this things dependency on that

so my current order of operations is:
1) ~make a v1.15 branch to rebase on~
2) implement `update_metadata` against v1.14 with a FIXME for the pubkey check
3) handroll a `VoteState::deserialize` impl against monorepo v1.15 that doesnt use more than 4kb of stack (:rotating_light: ive written byte parsers in c before but never in rust so this will be an "educational experience" ie take me a bit of time probably) but NOT merge it yet
4) continue v1.15 fixups in #4057 
5) rebase this on the v1.15 branch

at which point ill ask you to review and see if we are ready for an audit of the single-pool code. then i go back to monorepo to test the deserialize properly (this is the same deserialize that is used by validators, so it needs to be Correct, ill ask for help with fuzzing and benchmarking here), and then can go back to porting the rest of the stake pool tests to `stake-pool-test/`

joncinque:
> authorized_withdrawer is at a fixed offset, before the first vec, so parsing isnt strictly necessary... im not sure what validation would be needed other than asserting the vote account is owned by the proper program tho

I would say go with that first, so that the program can be completed and audited first. This program doesn't need to hinge on the vote state being correct. You could even decide to leave off `update_metadata` in v1, since there's no economic impact.

2501babe:
ok metadata functions work now, i think i just need to:
* redo errors.rs for all the new and removed items
* move initialize funding out into system transfers per your suggestion
* go through it again to see if theres anything to clean up that i missed...

2501babe:
ok i squashed my history and undid my premature `stake-pool/program/` rename because it made fast-fowards hell. rebased on master and updated to the latest mpl, also did all the errors cleanup

minor question, you suggested upthread that the payer for `initialize` could transfer the rent lamps into the accounts as independent instructions. do you have a preferred interface for the instruction function in this case? ie, is it acceptable to make `instruction::initialize` async to have it get the rent values itself, or should the caller do this and pass them in as `u64`s?

2501babe:
think i just need to...

before audit:
* ~cover the `VoteState0_23_5` case i skipped~
* ~redo errors.rs for all the new and removed items~
* move initialize funding out into system transfers per your suggestion
* fix token calculations (and ideally unit test them)
* go through it again to see if theres anything to clean up that i missed...

during audit:
* refactor test setup to use a trait for `Env`
* port rest of multipool tests
* fix ci (ie, get these new tests in ci, but also figure out why ci isnt running on this pr anymore...)

also need to decide on naming lol. have not really been consistent on whether its "single stake pool" or "stake single pool", assuming we want to call it that at all

stake-pool/single-pool/src/processor.rs
```diff
@@ -33,42 +33,32 @@ use {
...
-            (deposit_lamports as u128)
-                .checked_mul(token_supply as u128)?
-                .checked_div(validator_lamports as u128)?,
+            (user_stake_to_deposit as u128)
+                .checked_mul(pre_token_supply as u128)?
+                .checked_div(pre_pool_stake as u128)?,
         )
         .ok()
     }
 }
```
2501babe, line 52: on your next pass, please pay special attention to this and `calculate_withdraw_amount` because i have changed how they work from the multi-pool version

the comment illustrates a couple weird cases where token calculations break down: when no tokens have yet been minted, and when the amount of rent-exemption lamports in the stake account overwhelms the amount of stake in it

i think for this program it actually makes sense to do all calculations based on stake, rather than lamports, because:
* in single-pool, user rent returns to the user, whereas in multi-pool, user rent moves to the reserve. thus deposited stake and deposited lamports should be identical
* in single-pool, the program stake account can never be closed, so the rent deposit tendered at pool initialization is effectively forfeit, whereas in multi-pool, rent may be reclaimed with validators are wound down, so that value is still "live" for the pool

stake-pool/single-pool/src/processor.rs
```diff
@@ -721,13 +712,8 @@ impl Processor {
...
 
-        // XXX TODO FIXME use of lamports_added is VERY WRONG
-        // we mint tokens based on the total lamports that flow THROUGH our stake account
-        // including the rent we return!! meaning someone can steal funds
-        // we actually want to use *stake_added*. and probably (as noted at the fn def) the pre stake too
-        let new_pool_tokens =
-            calculate_deposit_amount(token_supply, pre_validator_lamports, lamports_added)
-                .ok_or(SinglePoolError::UnexpectedMathError)?;
+        let new_pool_tokens = calculate_deposit_amount(token_supply, pre_pool_stake, stake_added)
+            .ok_or(SinglePoolError::UnexpectedMathError)?;
```
2501babe, line 716: re: the above also note this nightmare bug stemming from my naively copying the lamports calculations

stake-pool/single-pool/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+// XXX obviously i hate this and if anyone has a better idea...
+// XXX this is only pub for testing purposes because the struct is in tests. not sure if theres a better way
+//    4 (enum)
+// + 32 (node_pubkey)
+// + 32 (authorized_voter)
+// +  8 (authorized_voter_epoch)
+// + (32 + 8 * 3) * 32 + 8 (prior_voters)
+// = 1876
+/// doc
+pub const LEGACY_VOTE_STATE_OFFSET: usize = 1876;
```
2501babe, line 36: also someone who is familiar with `VoteStateVersions` should sanity check this, im assuming that `VoteState0_23_5` (and `VoteState` but thats more obvious) onchain is _always_ wrapped in the enum, ie that the four byte offset is _always_ there. in other words, that `VoteStateVersions` was planned from there very beginning as future-proofing, before there were multiple versions

the reason i assume this has to be the case is `VoteState::deserialize` is written to parse both versions and upconvert the old one, and there is no way it could work if the enum tag wasnt in the account data for the old version

2501babe:
incidentally after all the work ive done since the original poc weve only net added a hundred lines, still under 20% of multipool
```
hana@laptop:single-pool$ wc src/*
   26    59   713 src/entrypoint.rs
   67   249  2424 src/error.rs
  374  1096 12861 src/instruction.rs
   68   282  2365 src/lib.rs
 1079  2624 39213 src/processor.rs
 1614  4310 57576 total
```

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+        // stake lamports added, as a stake difference
+        let stake_added = post_pool_stake
+            .checked_sub(pre_pool_stake)
+            .ok_or(SinglePoolError::ArithmeticUnderflow)?;
+
+        // excess rent, as all lamports less stake
+        let leftover_rent = lamports_added
+            .checked_sub(stake_added)
+            .ok_or(SinglePoolError::ArithmeticUnderflow)?;
```
joncinque, line 667: This is the safest behavior, but I want to walk through an alternative.

What if we have `leftover_rent = post_pool_lamports - post_pool_stake - stake_meta.stake_rent - minimum_delegation`? With that logic, if someone just transfers lamports into the stake pool expecting pool tokens, the following depositor can reclaim those lamports.

Most importantly, this means that the stake account never carries around extra lamports, and if someone messes up, they can always reclaim their funds by depositing immediately after. Or, more likely, someone else will. Otherwise, it's possible to just have unused lamports bricked in these accounts. What do you think?

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+        // sanity check: the user stake account is empty
+        if user_stake_info.lamports() != 0 {
+            return Err(SinglePoolError::UnexpectedMathError.into());
+        }
+
+        // we add initial lamports to supply to make the math work without minting tokens to incinerator
+        let token_supply = {
+            let pool_mint_data = pool_mint_info.try_borrow_data()?;
+            let pool_mint = Mint::unpack_from_slice(&pool_mint_data)?;
+            pool_mint.supply.saturating_add(INITIAL_LAMPORTS)
```
joncinque, line 688: If the pool account may start with additional lamports, or if the minimum delegation amount is raised after pool creation, does this logic break? I think it's probably ok to use 1 SOL or the minimum delegation floor, but I haven't thought fully through the implications.

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+        let bump_seed = check_pool_authority_address(
+            program_id,
+            vote_account_address,
+            pool_authority_info.key,
+        )?;
+        check_pool_mint_address(program_id, vote_account_address, pool_mint_info.key)?;
+        check_token_program(token_program_info.key)?;
+        check_stake_program(stake_program_info.key)?;
+
+        // we enforce that our stake is active so depositing/withdrawing during the warmup epoch is impossible
```
joncinque, line 620: I'm wondering if we need to enforce that... are there any issues with moving stake during the warmup epoch?

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+        let pre_pool_stake = get_active_stake_state(pool_stake_info, clock.epoch)?
+            .delegation
+            .stake;
+        msg!("Stake pre split {}", pre_pool_stake);
+
+        // we add initial lamports to supply to make the math work without minting tokens to incinerator
+        let token_supply = {
+            let pool_mint_data = pool_mint_info.try_borrow_data()?;
+            let pool_mint = Mint::unpack_from_slice(&pool_mint_data)?;
+            pool_mint.supply.saturating_add(INITIAL_LAMPORTS)
```
joncinque, line 765: Same point regarding the minimum to add here

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+        let post_pool_stake = get_active_stake_state(pool_stake_info, clock.epoch)?
+            .delegation
+            .stake;
+        msg!("Stake post split {}", post_pool_stake);
+
+        Ok(())
+    }
+
+    // this is an optional, but typical, part of initialization
+    // we separate it from the initialize instruction so that upstream cannot break us
```
joncinque, line 817: Very smart, good point!

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+        check_account_owner(payer_info, &system_program::id())?;
+        check_mpl_metadata_program(mpl_token_metadata_program_info.key)?;
+        check_mpl_metadata_account_address(metadata_info.key, pool_mint_info.key)?;
+
+        if !payer_info.is_signer {
+            msg!("Payer did not sign metadata creation");
+            return Err(SinglePoolError::SignatureMissing.into());
+        }
+
+        // checking the mint exists confirms pool is initialized
```
joncinque, line 848: I love all of these excessive checks

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+        {
+            let pool_mint_data = pool_mint_info.try_borrow_data()?;
+            let _ = Mint::unpack_from_slice(&pool_mint_data)?;
+        }
+
+        let vote_address_str = vote_account_address.to_string();
+        let token_name = format!(
+            "SPL S-Val Token ({}...{})",
```
joncinque, line 856: nit: how about this exactly-32-characters name?
```suggestion
            "SPL Single StkPool ({}...{})",
```

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+
+        let new_metadata_instruction = create_metadata_accounts_v3(
+            *mpl_token_metadata_program_info.key,
+            *metadata_info.key,
+            *pool_mint_info.key,
+            *pool_authority_info.key,
+            *payer_info.key,
+            *pool_authority_info.key,
+            token_name,
+            "".to_string(),
```
joncinque, line 869: nit: In 10 chars, we could do `format!("{}..{}", &vote_address_str[0..4],
            &vote_address_str[vote_address_str.len().saturating_sub(4)..vote_address_str.len()],)` -- what do you think?

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+        let state_variant = vote_account_data
+            .get(..VOTE_STATE_START)
+            .and_then(|s| s.try_into().ok())
+            .ok_or(SinglePoolError::UnparseableVoteAccount)?;
+
+        let (withdrawer_start, withdrawer_end) = match u32::from_le_bytes(state_variant) {
+            0 => (LEGACY_VOTE_STATE_START, LEGACY_VOTE_STATE_END),
+            1 => (VOTE_STATE_START, VOTE_STATE_END),
+            _ => return Err(SinglePoolError::UnparseableVoteAccount.into()),
+        };
```
joncinque, line 944: The vote program only has vote accounts, so your logic is good. And I don't think there are any more 0.23.5 structs, or if they do exist, they're effectively dead anyway.

I know @t-nelson has been talking about a new vote account version for some time... since it's early in the struct, can we guarantee that the withdrawer key will always be at the same place and make this logic work for new enum variants?

joncinque (commented):
This is looking really close! A few open questions for us to resolve based on offline chats, but this should be good to go soon. Note that I didn't look at any of the tests, but if you point me to the files to focus on, I can take a look on my next pass

stake-pool/single-pool/src/processor.rs
t-nelson: we're trying to avoid pulling in the vote program as a dependency here?

withdraw auth is one, one-byte field in front of a variably-length field :sweat_smile:

2501babe: `VoteState::deserialize` cant be used because of stack size violations, it got moved into `solana_program` but had to be ifdeffed out of bpf builds (see https://github.com/solana-labs/solana/issues/30188). this is a workaround until/unless i write a non-serde parser for it

stake-pool/single-pool/src/instruction.rs
```diff
@@ -0,0 +1,374 @@
...
+        /// URI of the uploaded metadata of the spl-token
+        uri: String,
+    },
+}
+
+// XXX we can bikeshed names of single-instruction vs "batteries included" helper functions
+// but i like making the latter the implicit default... instruction::initialize_instruction looks stupid tho idk
+// maybe that could be instruction::initialize but this could be instructions::initialize?
+// i dont want to just arbitrarily define a new convention tho
+/// Creates all necessary instructions to initialize the pool.
```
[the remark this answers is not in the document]
2501babe: ended up resolving this as a side effect of splitting `Initialize` into `InitializeMint` and `InitializeStake` to accomodate creating a token account to mint into

stake-pool/single-pool/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+// XXX TODO FIXME change this
+// (XXX ask how do we as a company handle privkeys for our onchain programs?)
+solana_program::declare_id!("3cqnsMsT6LE96pxv7GR4di5rLqHDZZbR3FbeSUeRLFqY");
+
+const POOL_STAKE_PREFIX: &[u8] = b"stake";
+const POOL_AUTHORITY_PREFIX: &[u8] = b"authority";
+const POOL_MINT_PREFIX: &[u8] = b"mint";
+
+const MINT_DECIMALS: u8 = 9;
+const INITIAL_LAMPORTS: u64 = 1;
```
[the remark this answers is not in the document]
2501babe: is there a reason to floor it like that, rather than just using the program response unconditionally?

stake-pool/single-pool/src/instruction.rs
```diff
@@ -0,0 +1,374 @@
...
+            vote_account,
+            user_stake_account,
+            user_token_account,
+            user_lamport_account,
+        ),
+    ]
+}
+
+/// Creates a `DepositStake` instruction.
+pub fn deposit_stake_instruction(
```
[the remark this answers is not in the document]
2501babe: gonna try `deposit` for the magic and `deposit_stake` for the mundane

stake-pool/single-pool/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+// authorized withdrawer starts at:
+//    4 (enum tag)
+// + 32 (node_pubkey)
+// + 32 (authorized_voter)
+// +  8 (authorized_voter_epoch)
+// + (32 + 8 * 3) * 32 + 8 (prior_voters)
+// = 1876
+const LEGACY_VOTE_STATE_START: usize = 1876;
+const LEGACY_VOTE_STATE_END: usize = 1908;
```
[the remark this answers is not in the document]
2501babe: it was an amusing exercise regardless!

stake-pool/single-pool/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[dependencies]
+arrayref = "0.3.6"
+borsh = "0.9"
+mpl-token-metadata =  { version = "1.7.0", features = [ "no-entrypoint" ] }
+num-derive = "0.3"
+num-traits = "0.2"
+num_enum = "0.5.9"
+serde = "1.0.130"
+serde_derive = "1.0.103"
```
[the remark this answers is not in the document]
2501babe: huh, i actually installed `cargo udeps` to try to catch things to cut, but it missed these... maybe it didnt care because theyre pulled in by other things, so these lines add version constraints but dont actually pull in new packages. it did catch that i didnt need `spl-math`

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+fn get_active_stake_state(
+    stake_account_info: &AccountInfo,
+    current_epoch: Epoch,
+) -> Result<stake::state::Stake, ProgramError> {
+    let stake_state =
+        try_from_slice_unchecked::<stake::state::StakeState>(&stake_account_info.data.borrow())?;
+    match stake_state {
+        stake::state::StakeState::Stake(_, stake)
+            if stake.delegation.activation_epoch < current_epoch
+                && stake.delegation.deactivation_epoch == Epoch::MAX =>
```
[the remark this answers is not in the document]
2501babe: i added it to assert that the pool stake account is active before anyone can use deposit/withdraw... but i guess other than abundance of caution, there isnt any kind of attack i can think of that this prevents from happening. maybe its pointless

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+                token_program_info.key,
+                pool_mint_info.key,
+                pool_authority_info.key,
+                None,
+                MINT_DECIMALS,
+            )?,
+            &[
+                pool_mint_info.clone(),
+                rent_info.clone(),
+                system_program_info.clone(),
```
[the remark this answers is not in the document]
2501babe: oh wow i didnt know that about invoke, id always just been copying the calls id seen elsewhere :sweat_smile:

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+            &ix,
+            &[burn_account, mint, authority, token_program],
+            signers,
+        )
+    }
+
+    #[inline(never)] // needed due to stack size violation
+    fn process_initialize(program_id: &Pubkey, accounts: &[AccountInfo]) -> ProgramResult {
+        let account_info_iter = &mut accounts.iter();
+        let validator_vote_info = next_account_info(account_info_iter)?;
```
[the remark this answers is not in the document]
2501babe: ill be updating this to check it, since its now actually necessary since i split `Initialize` into two stages

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1079 @@
...
+            POOL_MINT_PREFIX,
+            validator_vote_info.key.as_ref(),
+            &[mint_bump_seed],
+        ];
+        let mint_signers = &[&mint_seeds[..]];
+
+        // create the pool mint. user has already transferred in rent
+        let mint_space = spl_token::state::Mint::LEN;
+        let mint_rent = rent.minimum_balance(mint_space);
+        if pool_mint_info.lamports() != mint_rent {
```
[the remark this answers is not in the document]
2501babe: oh, good catch!

stake-pool/single-pool/src/processor.rs
2501babe: we use the true supply now that i mint the initialization tokens as part of supporting minimum delegation

stake-pool/single-pool/src/processor.rs
2501babe: yea im gonna think on this again but i believe youre right. im pretty sure i added the "active" qualifier to the function because i didnt _need_ the extra functionality, but i dont believe it does anything bad to have it

2501babe:
ok everything i resolved or replied to is either done or noted in comments to handle... also i support minimum delegation and removed the legacy vote code. will handle the rest tomorrow

also also note to self that i might get rid of all the `Self::` helpers. i copy-pasted them over from the multipool code, but because we do a lot less here, all of them except mint are only ever used once, so they just burn compute units for nothing

stake-pool/single-pool/src/instruction.rs
```diff
@@ -0,0 +1,420 @@
...
+    InitializeMint,
+
+    ///   Initialize the stake account for a new single-validator pool.
+    ///   Step two of the permissionless three-stage initialization flow.
+    ///
+    ///   0. `[]` Validator vote account
+    ///   1. `[w]` Pool stake account
+    ///   2. `[]` Pool authority
+    ///   3. `[w]` Pool token mint
+    ///   4. `[w]` User associated token account to receive pool tokens
```
joncinque, line 40: nit: It doesn't need to be an associated token account -- any will do!

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1078 @@
...
+        let mint_signers = &[&mint_seeds[..]];
+
+        // create the pool mint. user has already transferred in rent
+        let mint_space = spl_token::state::Mint::LEN;
+        let mint_rent = rent.minimum_balance(mint_space);
+
+        // TODO remove this after confirming for myself that spl-token doesnt allow creating rentpigs
+        if pool_mint_info.lamports() < mint_rent {
+            return Err(SinglePoolError::WrongRentAmount.into());
+        }
```
joncinque, line 485: token checks that it's rent-exempt https://github.com/solana-labs/solana-program-library/blob/5359039e15460c2f2130d2f6a1450339ae849a56/token/program/src/processor.rs#L48

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1078 @@
...
+                stake_config_info.clone(),
+                pool_authority_info.clone(),
+            ],
+            authority_signers,
+        )?;
+
+        // mint tokens to the user corresponding to their stake deposit
+        // note that this is safe from economic attacks despite the stake being in an activating state
+        // because we block anyone else from depositing until it has activated
+        // XXX actually do we give a shit about the activating stake?
```
joncinque, line 617: I would say it's probably fine. Since there's no way to deactivate the stake, the worst is some confusion when someone creates a pool and can't deposit their activated stake immediately. You can add a helpful error message in deposit if someone tries to do that though

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1078 @@
...
+        // TODO reread this part of the stake program (or just test for ourselves actually...) that uh
+        // no weird shit like "merge into activating and then split" can implicitly deactivate...
+        Self::token_mint_to(
+            validator_vote_info.key,
+            token_program_info.clone(),
+            pool_mint_info.clone(),
+            user_token_account_info.clone(),
+            pool_authority_info.clone(),
+            authority_bump_seed,
+            minimum_delegation,
```
joncinque, line 628: I don't think this should mint even `minimum_delegation`, since that amount is untouchable by depositors. It can do `pool_stake_info.lamports() - stake_rent_plus_initial` however, since those lamports are usable.

Otherwise, someone could have pool tokens that they can never withdraw because of the minimums

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1078 @@
...
+
+        // split stake into a blank stake account the user has created for this purpose
+        Self::stake_split(
+            vote_account_address,
+            pool_stake_info.clone(),
+            pool_authority_info.clone(),
+            bump_seed,
+            withdraw_stake,
+            user_stake_info.clone(),
+        )?;
```
joncinque, line 838: This shouldn't happen, but as a sanity check, can you check that `withdraw_stake != pool_stake_info.lamports()`? Otherwise it's possible to take control of the pool's stake account by deleting it and resurrecting it in the same transaction

stake-pool/single-pool/src/processor.rs
```diff
@@ -0,0 +1,1078 @@
...
+        // we add initial lamports to supply to make the math work without minting tokens to incinerator
+        let token_supply = {
+            let pool_mint_data = pool_mint_info.try_borrow_data()?;
+            let pool_mint = Mint::unpack_from_slice(&pool_mint_data)?;
+            pool_mint.supply
+        };
+
+        // withdraw amount is determined off stake just like deposit amount
+        let withdraw_stake = calculate_withdraw_amount(token_supply, pre_pool_stake, token_amount)
+            .ok_or(SinglePoolError::UnexpectedMathError)?;
```
joncinque, line 813: `pre_pool_stake` needs to take into account the untouchable minimum delegation amount which can't be withdrawn from the account

joncinque:
> also also note to self that i might get rid of all the Self:: helpers. i copy-pasted them over from the multipool code, but because we do a lot less here, all of them except mint are only ever used once, so they just burn compute units for nothing

Up to you, but it could be nicer to keep them factored out -- most likely, the compiler is inlining them anyway, so I doubt they're burning much more compute. If you do try it, I'd be interested to know the impact on compute usage!

stake-pool/single-pool/src/processor.rs
2501babe: i kinda have been going back and forth on this. its not really that "those" tokens cant withdraw "their" stake, but rather that this part of the cost of initializing the pool is socialized among all depositors. but otoh 1sol isnt a lot to pay to make the pool, but minting the tokens slightly disincentivizes early deposits...

i think im going to not mint, but instead of messing with token supply, i realized i can just introduce "available stake" and the math should work the same

stake-pool/single-pool/src/processor.rs
2501babe: `"SPL Single Stake Pool {}..{}"`?

stake-pool/single-pool/src/processor.rs
2501babe: hmm, we could do that. maybe trim two chars off the latter slice and prefix `s-`? or drop the dots/use ? i like the idea of some kind of recognizable prefix...

actually maybe instead of `"{}..{}"` with four and four, we do `"{}"` with first eight. here and above. the collision resistance doesnt change any and it gives us two free chars

stake-pool/single-pool/src/processor.rs
2501babe: and then maybe up here `"SPL Single Pool {}"` with as many bytes as we can fit

2501babe:
ok this is almost done, current items are:
* very minor error cleanup
* finish bikeshedding mpl metadata
* extra lamport hoovering
* check out the helper stuff

2501babe:
i think the last things for me to do here are:
* unmerge the tests into a separate pr for continuing work
* rebase on master
* flatten history

but ill leave those things alone for now so as not to break the "what changed since last time" diffs for you

2501babe:
ive been thinking maybe it makes more sense to just have a `tests/` directory for each program rather than do the env framework stuff i was doing. there really arent that many overlapping test conditions, and the main stuff thats duplicated is just helper functions, not anything essential to test (which is all in the program code)

stake-pool/single-pool/src/processor.rs
```diff
@@ -891,11 +915,8 @@ impl Processor {
         }
 
         let vote_address_str = vote_account_address.to_string();
-        let token_name = format!(
-            "SPL S-Val Token ({}...{})",
-            &vote_address_str[0..4],
-            &vote_address_str[vote_address_str.len().saturating_sub(4)..vote_address_str.len()],
-        );
+        let token_name = format!("SPL Single Pool {}", &vote_address_str[0..15]);
+        let token_symbol = format!("s-{}", &vote_address_str[0..7]);
```
joncinque, line 919: micro-nit: how about the `st` prefix, like with `stETH` or `stSOL`?
```suggestion
        let token_symbol = format!("st{}", &vote_address_str[0..7]);
```

joncinque (commented):
This is really close! Mostly bikeshedding and two actual questions

joncinque:
> but ill leave those things alone for now so as not to break the "what changed since last time" diffs for you

You have no idea how much I appreciate that

> ive been thinking maybe it makes more sense to just have a tests/ directory for each program rather than do the env framework stuff i was doing. there really arent that many overlapping test conditions, and the main stuff thats duplicated is just helper functions, not anything essential to test (which is all in the program code)

Really sorry about that, I was hoping that it would help you with writing tests, since the multi stake pool is pretty well-tested. I hope that you didn't spend too much time on it! Feel free to copy-paste whatever you need.

Since the program is so close, we can merge this in after the last round of nits, get the audit started, and the tests can come while the audit is happening.

stake-pool/single-pool/src/processor.rs
```diff
@@ -611,22 +628,23 @@ impl Processor {
...
-            validator_vote_info.key,
-            token_program_info.clone(),
-            pool_mint_info.clone(),
-            user_token_account_info.clone(),
-            pool_authority_info.clone(),
-            authority_bump_seed,
-            minimum_delegation,
-        )?;
+        // mint tokens if they overpaid the minimum amount
+        // XXX we could do away with this and cut two accounts from the instruction but idk this is more polite
```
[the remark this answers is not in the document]
2501babe: well, now that weve gone full circle (dont mint for any stake -> mint for all stake -> only mint for excess stake -> dont mint) i can combine `InitializeMint` and `InitializeStake` again :sweat_smile:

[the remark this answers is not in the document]
2501babe: also i can solve the "short names for helpers long names for instructions but initialize is backwards" thing from before by giving the instruction a longer name haha

stake-pool/single-pool/src/processor.rs
```diff
@@ -690,34 +712,33 @@ impl Processor {
...
             .ok_or(SinglePoolError::ArithmeticOverflow)?;
 
-        // excess rent, as all lamports less stake
-        let leftover_rent = lamports_added
-            .checked_sub(stake_added)
+        // we calculate absolute rather than relative to deposit amount to allow claiming lamports mistakenly transferred in
+        let excess_lamports = post_pool_lamports
+            .checked_sub(post_pool_stake)
+            .and_then(|amount| amount.checked_sub(pool_stake_meta.rent_exempt_reserve))
+            .and_then(|amount| amount.checked_sub(minimum_delegation))
```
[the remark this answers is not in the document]
2501babe: ah yea thats a simpler equivalent

stake-pool/single-pool/src/processor.rs
```diff
@@ -690,34 +712,33 @@ impl Processor {
...
-        if stake_added != pre_user_stake {
+        // sanity check: we have not somehow gone below the minimum
+        if post_pool_stake < minimum_delegation {
             return Err(SinglePoolError::UnexpectedMathError.into());
         }
 
-        // sanity check: our calculated excess rent matches the user prior rent
-        if leftover_rent != user_unstaked_lamports {
+        // sanity check: our calculated new stake matches the user prior stake
+        if stake_added != pre_user_stake {
```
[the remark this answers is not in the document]
2501babe: omg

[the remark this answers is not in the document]
2501babe: but the `excess_lamports` calculation is still safe, right? because the user rent gets added to `pool_stake_state.delegation.stake`?

ah, but because of this we need to guard the withdraw call in case its zero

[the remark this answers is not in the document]
2501babe: wait but that also means calculating tokens off `stake_added` is wrong, and we need to use the difference between post and pre

2501babe:
ok i pushed what should be the final fixes, ill remove tests and squash merge pending your final approval!

stake-pool/single-pool/src/processor.rs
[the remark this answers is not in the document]
joncinque: Yeah sorry about that... considering that you get back extra lamports during deposit, it seemed better to merge the instructions once again

stake-pool/single-pool/src/processor.rs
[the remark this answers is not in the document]
joncinque: That's all correct

joncinque (dismissed):
R+ once the tests are removed and CI is green -- great job and thanks for your patience!

PR #3976, opened 2023-01-16
Governance: Allow token owner to revoke their own membership

SebastianBor:
#### Summary

Make it possible for the owner of governing `Membership` token to voluntarily revoke its own memebership


joncinque (approved):
Simple and clean, looks good!

PR #3977, opened 2023-01-17
Account Compression: Change VerifyLeaf to avoid modifying canopy

ngundotra:
Right now this instruction fails when the canopy is not fully initialized, because the `merkleTree` account is passed as read only.

I changed this instruction to be read-only so now the instruction works as intended

### DIFF
--- a/account-compression/programs/account-compression/src/canopy.rs
+++ b/account-compression/programs/account-compression/src/canopy.rs
@@ -19,12 +19,12 @@
 use crate::error::AccountCompressionError;
 use crate::events::ChangeLogEvent;
 use anchor_lang::prelude::*;
-use bytemuck::cast_slice_mut;
+use bytemuck::{cast_slice, cast_slice_mut};
 use spl_concurrent_merkle_tree::node::{empty_node_cached, Node, EMPTY};
 use std::mem::size_of;
 
 #[inline(always)]
-pub fn check_canopy_bytes(canopy_bytes: &mut [u8]) -> Result<()> {
+pub fn check_canopy_bytes(canopy_bytes: &[u8]) -> Result<()> {
     if canopy_bytes.len() % size_of::<Node>() != 0 {
         msg!(
             "Canopy byte length {} is not a multiple of {}",
@@ -38,7 +38,7 @@ pub fn check_canopy_bytes(canopy_bytes: &mut [u8]) -> Result<()> {
 }
 
 #[inline(always)]
-fn get_cached_path_length(canopy: &mut [Node], max_depth: u32) -> Result<u32> {
+fn get_cached_path_length(canopy: &[Node], max_depth: u32) -> Result<u32> {
     // The offset of 2 is applied because the canopy is a full binary tree without the root node
     // Size: (2^n - 2) -> Size + 2 must be a power of 2
     let closest_power_of_2 = (canopy.len() + 2) as u32;
@@ -89,7 +89,7 @@ pub fn update_canopy(
 }
 
 pub fn fill_in_proof_from_canopy(
-    canopy_bytes: &mut [u8],
+    canopy_bytes: &[u8],
     max_depth: u32,
     index: u32,
     proof: &mut Vec<Node>,
@@ -97,7 +97,7 @@ pub fn fill_in_proof_from_canopy(
     // 30 is hard coded as it is the current max depth that SPL Compression supports
     let mut empty_node_cache = Box::new([EMPTY; 30]);
     check_canopy_bytes(canopy_bytes)?;
-    let canopy = cast_slice_mut::<u8, Node>(canopy_bytes);
+    let canopy = cast_slice::<u8, Node>(canopy_bytes);
     let path_len = get_cached_path_length(canopy, max_depth)?;
 
     // We want to compute the node index (w.r.t. the canopy) where the current path
@@ -115,7 +115,6 @@ pub fn fill_in_proof_from_canopy(
         if canopy[cached_idx] == EMPTY {
             let level = max_depth - (31 - node_idx.leading_zeros());
             let empty_node = empty_node_cached::<30>(level, &mut empty_node_cache);
-            canopy[cached_idx] = empty_node;
             inferred_nodes.push(empty_node);
         } else {
             inferred_nodes.push(canopy[cached_idx]);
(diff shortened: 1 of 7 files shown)


account-compression/programs/account-compression/src/canopy.rs
```diff
@@ -115,7 +115,6 @@ pub fn fill_in_proof_from_canopy(
         if canopy[cached_idx] == EMPTY {
             let level = max_depth - (31 - node_idx.leading_zeros());
             let empty_node = empty_node_cached::<30>(level, &mut empty_node_cache);
-            canopy[cached_idx] = empty_node;
```
joncinque, line 118: Is this OK to remove since it's just zero initializing? Is there a potential issue if an account isn't zero-initialized? E.g. re-using an account

account-compression/programs/account-compression/src/macros.rs
```diff
@@ -0,0 +1,106 @@
+/// This macro applies functions on a ConcurrentMerkleT:ee and emits leaf information
+/// needed to sync the merkle tree state with off-chain indexers.
+#[macro_export]
+macro_rules! merkle_tree_depth_size_apply_fn {
```
joncinque, line 4: nit: how about making this private too, ie 
```suggestion
macro_rules! _merkle_tree_depth_size_apply_fn {
```

account-compression/programs/account-compression/src/macros.rs
```diff
@@ -0,0 +1,106 @@
...
+    ($max_depth:literal, $max_size:literal, $id:ident, $bytes:ident, $func:ident, true, $($arg:tt)*)
```
joncinque, line 5: nit take-it-or-leave-it: rather than specifying `true` and `false` for mutability and matching on that, how about a private enum like
```
enum TreeLoad {
    Immutable,
    Mutable,
}
```

account-compression/sdk/.eslintrc.js
```diff
@@ -38,7 +38,7 @@ module.exports = {
     "no-unused-vars": "off",
     quotes: [
       "error",
-      "single",
+      "double",
```
joncinque, line 41: I know it's annoying, but I'd prefer that you do this in a separate PR since it's totally unrelated and adds a lot of churn to this PR

joncinque (commented):
I'm not sure if I picked up all the changes on the TS side because of the reformatting, but overall looks good! Just some nits

account-compression/programs/account-compression/src/macros.rs
ngundotra: AFAIK macros can't do runtime expansions like matching over an enum. True/false for mutability is really just SDK ergonomics.

account-compression/programs/account-compression/src/canopy.rs
ngundotra: Yes it's okay to remove. There is a separate method `update_canopy` that sets the values of the canopy over time. 

I think this call is meant to save compute, but it comes at the cost of requiring the entire `merkleTree` account to be passed as mutable in the `verifyLeaf` instruction.

account-compression/programs/account-compression/src/macros.rs
ngundotra: Updated!

account-compression/sdk/.eslintrc.js
ngundotra: Fixed!

account-compression/programs/account-compression/src/macros.rs
joncinque: You're doing a compile-time match, not a runtime match. So you can do:

```
enum TreeLoad {
      Immutable,
      Mutable,
  }

  macro_rules! apply_fn {
      ($func:ident, TreeLoad::Immutable, $($arg:tt)*) => {
          println!("immutable");
          $func($($arg)*)
      };
      ($func:ident, TreeLoad::Mutable, $($arg:tt)*) => {
          println!("mutable");
          $func($($arg)*)
      };
  }

  fn do_print(word: &str) {
      println!("Hello, {word}!");
  }

  fn main() {
      apply_fn!(do_print, TreeLoad::Immutable, "immutable hi");
      apply_fn!(do_print, TreeLoad::Mutable, "mutable hi");
  }
```

Although weirdly enough, it says that `TreeLoad` is never used, so :shrug:

ngundotra: Awesome, thanks! Updated with the Enum

ngundotra:
@joncinque Let me know if looks good to go. Simplified PR + addressed feedback

account-compression/programs/account-compression/src/macros.rs
```diff
@@ -0,0 +1,110 @@
...
+#[macro_export]
+macro_rules! merkle_tree_apply_fn {
+    ($header:ident, $id:ident, $bytes:ident, $func:ident, $($arg:tt)*) => {
+        _merkle_tree_apply_fn!($header, $id, $bytes, $func, TreeLoad::Mutable, $($arg)*)
+    };
+}
+
+/// This applies a given function on a read-only ConcurrentMerkleTree
+#[macro_export]
+macro_rules! merkle_tree_apply_fn_immutable {
```
joncinque, line 106: nit: sorry, just realizing now, since it's more idiomatic to specify when something is mutable rather than immutable, so this can be  `merkle_tree_apply_fn` and the other one becomes `merkle_tree_apply_fn_mut`. Not sure if you have outside users depending on this

joncinque (approved):
Just one more nit, nothing stopping this though, looks good!

account-compression/programs/account-compression/src/macros.rs
ngundotra: Not sure if I have outside users consuming this merkle tree method either to be honest.

Would it be better to just tag `merkle_tree_apply_fn` as `#[deprecated]` and instead rename to `merkle_tree_apply_fn_mutable` and `merkle_tree_apply_fn_immutable` ?

joncinque: https://crates.io/crates/spl-account-compression/reverse_dependencies -- you can always talk with metaplex? Honestly, not a big deal either way, but if you move to version 0.2, you can break, according to semver

ngundotra: I checked their code - no usage of the macros. 

Changed names to match your suggestion. Thanks 🙏

PR #3978, opened 2023-01-17
Support `create` and `update` token metadata in stake-pool python bindings

### LINKED ISSUE #3370: stake-pool-py: Support add / update metadata for pool token
#### Problem

#3335 added new instructions to the stake pool, to add and update token metadata, but this is not supported in python.

#### Solution

Add support for these instructions in the python bindings.

gnapoli23:
**Problem**
#3335 added new instructions (`create_metadata` and `update_metadata`) to the stake-pool, while these are not supported in python.

**Solution**
Add python bindings `create_token_metadata` and `update_token_metadata` in stake-pool, with their relative tests.

Fixes #3370


stake-pool/py/stake_pool/constants.py
```diff
@@ -17,6 +17,9 @@
 MINIMUM_ACTIVE_STAKE: int = MINIMUM_DELEGATION
 """Minimum active delegated staked required in a stake account"""
 
+METADATA_PROGRAM_ID: PublicKey = PublicKey("metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s")
+"""Public key that identifies the SPL Token Metadata program."""
```
joncinque, line 21: nit: this is the metaplex metadata program
```suggestion
"""Public key that identifies the Metaplex Token Metadata program."""
```

stake-pool/py/stake_pool/instructions.py
```diff
@@ -496,6 +542,12 @@ class InstructionType(IntEnum):
     "seed" / Int32ul
 )
 
+TOKEN_METADATA_LAYOUT = Struct(
+    "name" / Bytes(32),
+    "symbol" / Bytes(32),
+    "uri" / Bytes(32)
+)
```
joncinque, line 549: You'll need to be careful with these. They're encoded as `String` in the interface, which in Borsh means they're actually like vectors, meaning 4 bytes to encode the length, followed by the utf8-encoded bytes.  You can find more info at the borsh spec: https://borsh.io/ -> "Specification"

stake-pool/py/stake_pool/instructions.py
```diff
@@ -917,3 +971,63 @@ def decrease_validator_stake(params: DecreaseValidatorStakeParams) -> Transactio
...
+        AccountMeta(pubkey=params.stake_pool, is_signer=False, is_writable=False),
+        AccountMeta(pubkey=params.manager, is_signer=True, is_writable=False),
+        AccountMeta(pubkey=withdraw_authority, is_signer=False, is_writable=False),
+        AccountMeta(pubkey=params.pool_mint, is_signer=False, is_writable=False),
+        AccountMeta(pubkey=params.payer, is_signer=True, is_writable=True),
+        AccountMeta(pubkey=token_metadata, is_signer=False, is_writable=True),
+        AccountMeta(pubkey=METADATA_PROGRAM_ID, is_signer=False, is_writable=False),
+        AccountMeta(pubkey=SYS_PROGRAM_ID, is_signer=False, is_writable=False),
+        AccountMeta(pubkey=SYSVAR_RENT_PUBKEY, is_signer=False, is_writable=False),
+    ]
```
joncinque, line 991: nit: I know it's not the easiest usage pattern, but to make this consistent with the other instruction creators, can you provide the ability to specify all accounts?

joncinque (commented):
This is on the right track! You'll need to be careful on the instruction encoding, but other than that, all good

stake-pool/py/stake_pool/constants.py
gnapoli23: Yes sure, C/P mistake 😅

stake-pool/py/stake_pool/instructions.py
gnapoli23: Thanks, I was actually in doubts about that so I put `Bytes` as a "placeholder" to further updates. Thanks for the clarification.

gnapoli23: @joncinque I gave a look at borsh specification, does this implementation match for `String` type?
```
TOKEN_METADATA_LAYOUT = Struct(
    "name" / Prefixed(Int32ul, GreedyString("utf8")),
    "symbol" / Prefixed(Int32ul, GreedyString("utf8")),
    "uri" / Prefixed(Int32ul, GreedyString("utf8"))
)
```
Here you can find the doc of `contruct` module for [Prefixed](https://construct.readthedocs.io/en/latest/api/core.html#construct.core.Pre) and [GreedyString](https://construct.readthedocs.io/en/latest/api/core.html#construct.core.GreedyString).
`Prefixed` contains a field describing the length (so it parses this filed and then reads this amount of bytes), while `GreedyString` is essentially a `GreedyBytes` that encodes bytes with unicode byte encoding.

joncinque: Yep, that looks perfect!

gnapoli23:
@joncinque
Updated as discussed. For tests, can you give me more details? I take a look at the one related to `deposit_sol` as you suggested, but unfortunately i didn't got what is the logic flow to follow regarding new `create` and `update` functions.
Is there any Rust test about this fns that has to be considered as reference test?
More over, I have implement also `stake-pool`'s py async actions for these new functions, right? Seems there's nothing present yet about.

joncinque:
The idea is to test that the new instruction bindings work, and that you can create and update token metadata through the stake pool program. The `deposit_sol` test just showed the overall flow of what one of those python tests looks like. 

You can see the metadata rust tests at https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/tests/create_pool_token_metadata.rs and https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/tests/update_pool_token_metadata.rs.

Feel free to add `actions` for these too, it'll make testing easier.

And finally, you'll need to add the token metadata program to the test validator when it's booted for testing, so be sure to change this in conftest.py:

```
@pytest.fixture(scope="session")
def solana_test_validator():
    old_cwd = os.getcwd()
    newpath = tempfile.mkdtemp()
    os.chdir(newpath)
    validator = Popen([
        "solana-test-validator",
        "--reset", "--quiet",
        "--bpf-program", "SPoo1Ku8WFXoNDMHPsrGSTSG1Y47rzgn41SLUNakuHy",
        f"{old_cwd}/../../target/deploy/spl_stake_pool.so",
        "--bpf-program", "metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s",
        f"{old_cwd}/../program/tests/fixtures/mpl_token_metadata.so",
        "--slots-per-epoch", str(NUM_SLOTS_PER_EPOCH),
    ],)
    yield
    validator.kill()
    os.chdir(old_cwd)
    shutil.rmtree(newpath)
```

gnapoli23:
@joncinque I added actions for `create_token_metadata` and `update_token_metdata`, can you give a look? I'll tests in the maintime :)

stake-pool/py/stake_pool/actions.py
```diff
@@ -589,3 +590,56 @@ async def decrease_validator_stake(
...
+            program_id=STAKE_POOL_PROGRAM_ID,
+            stake_pool=stake_pool_address,
+            manager=stake_pool.manager,
+            pool_mint=stake_pool.pool_mint,
+            payer=payer.public_key,
+            name=name,
+            symbol=symbol,
+            uri=uri,
+            withdraw_authority=withdraw_authority,
+            token_metadata=METADATA_PROGRAM_ID,
```
joncinque, line 615: This should be the actual metadata account, and not the metadata program id.

stake-pool/py/stake_pool/actions.py
```diff
@@ -589,3 +590,56 @@ async def decrease_validator_stake(
...
+                                name: str, symbol: str, uri: str):
+    resp = await client.get_account_info(stake_pool_address, commitment=Confirmed)
+    data = resp['result']['value']['data']
+    stake_pool = StakePool.decode(data[0], data[1])
+
+    (withdraw_authority, _seed) = find_withdraw_authority_program_address(STAKE_POOL_PROGRAM_ID, stake_pool_address)
+
+    txn = Transaction()
+    txn.add(
+        sp.CreateTokenMetadataParams(
```
joncinque, line 605: Maybe I'm missing something, but this just gives a tuple, and not an instruction. You'll need to call `create_metadata_account` with `sp.CreateTokenMetadataParams` as a parameter.

stake-pool/py/stake_pool/instructions.py
```diff
@@ -917,3 +983,77 @@ def decrease_validator_stake(params: DecreaseValidatorStakeParams) -> Transactio
...
+    """Creates an instruction to create metadata using the mpl token metadata program for the pool token."""
+    if params.withdraw_authority:
+        withdraw_authority = params.withdraw_authority
+    else:
+        (withdraw_authority, _seed) = find_withdraw_authority_program_address(params.program_id, params.stake_pool)
+
+    if params.token_metadata:
+        token_metadata = params.token_metadata
+    else:
+        (token_metadata, _seed) = find_metadata_account(params.pool_mint)
```
joncinque, line 998: To avoid the earlier footgun, let's keep this logic as dumb as possible and put *all* parameters in `CreateTokenMetadataParams`, including the rent sysvar, system program id, and metadata program id, and *not* make any of them optional.  

Essentially, let's make this the same as all of the other instruction creators, and then the `action` can fill in the right parameters

joncinque (commented):
> I'm not so sure about the signers in send_transaction, ' cause i saw that in Rust code there's also the stake pool manager as signer, but in python it is actually a PublicKey and not a Keypair (as expected by the function itself).

The `action` is a simplified wrapper that assumes the `payer` is actually the manager, and there is no analogue to the Rust side, since it's also signing and sending the transaction, whereas the Rust version is just creating the instruction.

joncinque:
Btw, looking through your PR made me realize that we're behind in the stake pool program, so I've updated to the newest bindings at #4012 -- you'll just need to remove the sysvar rent account during "create metadata".

stake-pool/py/stake_pool/actions.py
gnapoli23: Yep sorry, missed it

gnapoli23:
@joncinque Thanks, I did some updates following you guidelines, hope they're correct. I also added a simple test, can you check if it's ok?

stake-pool/py/stake_pool/actions.py
```diff
@@ -589,3 +589,58 @@ async def decrease_validator_stake(
     signers = [payer, staker] if payer != staker else [payer]
     await client.send_transaction(
         txn, *signers, opts=TxOpts(skip_confirmation=False, preflight_commitment=Confirmed))
+
+
+async def create_token_metadata(client: AsyncClient, payer: Keypair, stake_pool_program_id: PublicKey,
+                                stake_pool_address: PublicKey, token_program_id: PublicKey,
+                                withdraw_authority: PublicKey, name: str, symbol: str, uri: str):
```
joncinque, line 596: nit: For the actions, let's keep it simpler, same as you're using it in the test, and in this function you can derive the missing parameters:
```suggestion
async def create_token_metadata(client: AsyncClient, payer: Keypair,
                                stake_pool_address: PublicKey,
                                name: str, symbol: str, uri: str):
```

joncinque (commented):
It's looking close! Once the create test works, then you just need an update test, then this should be good to go!

stake-pool/py/tests/test_create_token_metadata.py
```diff
@@ -0,0 +1,30 @@
...
+    symbol = "SYM"
+    uri = "test_uri"
+    await create_token_metadata(async_client, payer, stake_pool_address, name, symbol, uri)
+
+    (metadata_program_address, _seed) = find_metadata_account(stake_pool.pool_mint)
+    resp = await async_client.get_account_info(metadata_program_address, commitment=Confirmed)
+    data = resp['result']['value']['data']
+    assert data[0] == name
+    assert data[1] == symbol
+    assert data[2] == uri
```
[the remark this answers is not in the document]
gnapoli23: Yes I saw that data in Rust code are deserialized in the Metadata struct, I fell something was definitely missing. I did some change at the test and reverted the last commit on actions.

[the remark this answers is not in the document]
gnapoli23: Yes I saw that data in Rust code are deserialized in the Metadata struct, I felt something was definitely missing. I did some change at the test and reverted the last commit on actions.

gnapoli23:
Hi @joncinque , sorry to bother...did you read my last comment? Thanks.

stake-pool/py/stake_pool/actions.py
```diff
@@ -589,3 +589,60 @@ async def decrease_validator_stake(
...
+                program_id=STAKE_POOL_PROGRAM_ID,
+                stake_pool=stake_pool_address,
+                manager=stake_pool.manager,
+                pool_mint=stake_pool.pool_mint,
+                payer=payer.public_key,
+                name=name,
+                symbol=symbol,
+                uri=uri,
+                withdraw_authority=withdraw_authority,
+                token_metadata=TOKEN_PROGRAM_ID,
```
joncinque, line 615: `token_metadata` should be the result of `find_metadata_account(stake_pool.pool_mint)`, and then you still need to add the system program id and token metadata program id

stake-pool/py/stake_pool/actions.py
```diff
@@ -589,3 +589,60 @@ async def decrease_validator_stake(
...
+            sp.UpdateTokenMetadataParams(
+                program_id=STAKE_POOL_PROGRAM_ID,
+                stake_pool=stake_pool_address,
+                manager=stake_pool.manager,
+                pool_mint=stake_pool.pool_mint,
+                name=name,
+                symbol=symbol,
+                uri=uri,
+                withdraw_authority=withdraw_authority,
+                token_metadata=TOKEN_PROGRAM_ID,
```
joncinque, line 643: Same here, you'll need `find_metadata_account(stake_pool.pool_mint)` and the metadata program id

joncinque (commented):
Sorry, I thought you were still working on it. Once you fix the issues, get the create test working, and add an update test, this should be good to go!

gnapoli23:
@joncinque Hi, I updated `actions.py` and the `solana_test_validator` setup function as suggested, but unfortunately I'm not able to run test somehow, it keeps on telling me that:
`ERROR    solanaweb3.rpc.httprpc.HTTPClient:async_http.py:35 Health check failed with error: All connection attempts failed`

Giving you more info, I'm working in a VirtualBox Ubuntu machine, don' t know if it's related to this.
I also tried to run the `setup-test-validato.sh` script to see if something changes, it creates the validators pool (indeed creates also all related files under `keys` folder, but still gives the same error).

Hope this can be relevant info, sorry to bother you :)

joncinque:
It looks like there was a bug in your test, since it's trying to decode the wrong part of the response.  Be sure to check the errors reported when you run `python3 -m pytest`!

Here's a patch that will fix your test and make everything work:
```
-    raw_data = resp['result']
-    assert name == str(raw_data[69:101], "utf-8")
-    assert symbol == str(raw_data[105:115], "utf-8")
-    assert uri == str(raw_data[119:319], "utf-8")
+    data = resp['result']['value']['data']
+    raw_data = decode_byte_string(data[0], data[1])
+    assert name == str(raw_data[69:101], "utf-8")[:len(name)]
+    assert symbol == str(raw_data[105:115], "utf-8")[:len(symbol)]
+    assert uri == str(raw_data[119:319], "utf-8")[:len(uri)]
```

After that, you just need an update test, then this will be good to go!

gnapoli23:
@joncinque Hi, rebased, fixed the create test and added the update one :)

joncinque:
Thanks for all of your patience and for integrating all of my suggestions -- this is finally good to go!

PR #3979, opened 2023-01-18
stake-pool: Wait at least two epoch boundaries to set fee

joncinque:
#### Problem

If a malicious manager wants to hike a fee, they can still do so right at the end of an epoch and have it take into account immediately at the start of the next epoch, giving users no time to get their funds out if they want.

#### Solution

Force a wait of at least two epoch boundaries before applying a new fee. This solution needs to be backwards compatible with existing stake pools, which use `Option<Fee>`. In Borsh, this is encoded as:

```
if x.is_some() {
  repr(1 as u8)
  repr(x.unwrap() as ident)
} else {
  repr(0 as u8)
}
```

This introduces a new enum `FutureEpoch`, which encodes `0` as `None` (same as option), `1` as "wait one epoch" (same as option), and then `2` as "wait two epochs" (the new state).

The biggest issue with this approach is that it requires calling "update" in two different epochs before applying the fee, rather than having the fee apply at a particular future epoch. Meaning, you can do "set fee", wait two epochs, then call "update", and you'll only move forward one epoch. You must call "set fee", wait at least one epoch boundary, call "update", wait at least one more epoch boundary, and call "update" again. Thankfully, at worst, it just takes longer for the new fee to kick in, which is acceptable (to me).


stake-pool/program/src/state.rs
```diff
@@ -793,6 +793,62 @@ impl ValidatorListHeader {
...
+        }
+    }
+
+    /// Get the value if it's ready, which is only at `One` epoch remaining
+    pub fn get(&self) -> Option<&T> {
+        match self {
+            Self::None | Self::Two(_) => None,
+            Self::One(v) => Some(v),
+        }
+    }
```
2501babe, line 841: this is cute

2501babe (approved):

stake-pool/program/src/state.rs
joncinque: I certainly had fun writing it :smile:

PR #3980, opened 2023-01-18
stake-pool: Add slippage to all deposit and withdraw ixs

joncinque:
#### Problem

When depositing and withdrawing from stake pools, it's impossible for a user to guarantee how much they'll receive.

#### Solution

Add `*WithSlippage` variants to all deposit and withdraw instructions, which guarantee a minimum output, similar to an AMM.


joncinque:
Yeah exactly, someone can watch for transactions depositing into the pool and put in an instruction to increase the deposit fee immediately. In your case, I think you don't need it since there's no fees to consider

2501babe:
im imagining a situation where someone does a reasonably large deposit so i frontrun their deposit with my own and backrun a withdrawal

2501babe (approved):

PR #3985, opened 2023-01-19
Account Compression: bump rust version

ngundotra:
* New macros for applying functions to merkle trees
* Fixed issue with verifyLeaf instruction on merkle trees with canopy


charlestonXdao (commented):
good

Issue #4006, opened 2023-01-30
ci: Remove uses of `Pubkey::new`


joncinque:
#### Problem

The monorepo has deprecated `Pubkey::new` in https://github.com/solana-labs/solana/pull/29805, and SPL still uses `Pubkey::new` in a few places.

#### Solution

Once 1.14.14 exists, remove all uses of `Pubkey::new` in favor of `Pubkey::try_from` and undo the workaround from #4007

nahnicholas:
Ty

joncinque:
This was done as part of  #4592

PR #4007, opened 2023-01-30
ci: Update twoxtx build to ignore warnings

joncinque:
#### Problem

The twoxtx build is currently failing because the monorepo has deprecated `Pubkey::new`, SPL still uses it, and CI builds all tests with `-D warnings`.

#### Solution

There's no release of 1.14 with the `Pubkey::try_from` implementation on `&[u8]` just yet, so we can't fix this correctly until that's ready.  Keep it simple for now and just build the token-2022 tests directly, without `-D warnings`.

#4006 will keep track of the later work

### DIFF
--- a/.github/workflows/pull-request-token.yml
+++ b/.github/workflows/pull-request-token.yml
@@ -86,7 +86,7 @@ jobs:
 
       - name: Set env vars
         run: |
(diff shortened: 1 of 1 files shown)


samkim-crypto (dismissed):
Makes sense to me!

nahnicholas (commented):
Ty

PR #4009, opened 2023-01-31
spl-uip: Universal IDL Program

ngundotra:
Universal IDL Program to manage a registry for program IDL's irrespective of whether they used Anchor.

Currently, programs not written in Anchor are unable to compose with the rest of the ecosystem because they lack the Anchor instructions to modify their own IDL (which is a program account). 

Having the Universal IDL Program managed by SPL can fix this immediate issue.

Additionally, this PR addresses two other problems with the ecosystem:
1 - Fracturing IDL standards (see Anchor IDL vs Shank IDL)
2 - Frozen programs lack IDL registry 

Problem 1: we can define a standard or set of standards on how information should be written/read on-chain & let ecosystem battle it out. But we can make it easier for ecosystem to flourish by providing a rock-solid foundation for at least WHERE the IDL is stored on-chain.

Problem 2: SPL programs don't have IDLs, as well as a handful of other frozen programs. With a bit of centralized admin authority on Solana Labs' part, we can manage & define anchor IDLs for these frozen programs in the same locations as upgradeable programs' IDLs.

Description below

# Universal IDL Program

This is a program that allows program authorities to upload and maintain Interface Description Language files for their programs. 

### IDL Authority
For upgradeable programs, the IDL can only be created by the upgrade authority.

For frozen programs, a meta authority must create a PDA setting an arbitrary keypair as an IDL authority for that program via execution of the `declare_frozen_authority` instruction.

### IDL Upgrades

We want the IDL process to be useable with SPL Governance voting with minimal difficulty. 
Therefore, we require the `idl`'s associated `buffer` account to be written to first.
Then the `idl`'s `authority` can `resize` the idl buffer to match the buffer length, and then
perform `set_buffer` to copy the buffer bytes into the IDL.

This makes it possible to execute an `upgrade` instruction for IDLs in a single transaction.
Minimizing instruction data needed to perform upgrades makes this action easier to perform in Governance votes.

### IDL Parsing and Verification

This program makes no assumptions about the formatting, serialization or structure of the
data stored in the IDL.

We expect clients retreiving IDLs to parse and interpret IDL data however they wish.

The provided CLI uses Zlib compression to encode & decode utf-8 bytes as IDL data (same as Anchor).
We perform 0 additional verification of the data.


Henry-E:
Thoughts, questions, comments

- Security wise we've settled on: only the program upgrade authority should be allowed to specify the IDL for a given program? I personally like this approach but discussed it more below.
- Anchor made some recent changes to the IDL program like adding closing, resizing and zero copy (for really big IDLs)
- Hopefully you got rid of the previous way of deriving the program address for the IDL and replaced it with a regular PDA?
- There were some checks missing in the original IDL program, you need to make sure that you can't accidentally pass in the temporary upgrade buffer where the actual IDL account is expected and vice versa.

Honestly, a program like this is probably feature complete once it has all the basic features that were present in the original IDL program + these new features like closing (though the IDL can always be reinit so long as the program authority is still active), resizing accounts + zero copy (so the program doesn't run out of compute when doing 10kb updates). If the program was audited you could probably kill the program upgrade authority after uploading 🤷 .

The only downside I thought about was that people might be annoyed at having to rely on the program upgrade authority to upload IDLs, since that's usually behind a multisig or some kind of council vote. But at the same time, IDL changes really shouldn't happen unless the program is itself being updated, so they can probably be packaged together. I would be curious to hear from people like drift or mango how they would feel about it since they are the biggest IDL power users (and their IDLs are huge).

universal-idl-program/program/src/instructions/create_idl/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+        ),
+        &[ctx.payer.clone(), ctx.idl_account.clone()],
+        &[&idl_seeds
+            .iter()
+            .map(|s| s.as_slice())
+            .collect::<Vec<&[u8]>>()],
+    )?;
+
+    let mut idl: Idl = Idl::new();
+    idl.authority = *ctx.program_authority.key;
```
jarry-xiao, line 203: What would happen if the program_authority (on the actual program) were to change?

ngundotra: IDL authority stays the same

jarry-xiao: This feels potentially a little weird, but I think it's probably okay as long as any downstream clients know how to validate idl.auth == program.auth

universal-idl-program/program/src/instructions/create_idl/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+                bincode::deserialize(&program_data.data.borrow())
+                    .map_err(|_| ProgramError::InvalidAccountData)?;
+
+            Ok(Self::UpgradeableAuthority {
+                upgradeable_program: program,
+                program_data: program_data_state,
+            })
+        } else if is_frozen_program(program) {
+            let frozen_authority_account = next_account_info(accounts_iter)?;
+            assert_frozen_authority_seeds(program.key, frozen_authority_account.key)?;
```
jarry-xiao, line 105: I think this function is semantically difficult to reason about. I know it's checking something important, but it's unclear what it's doing.

If `frozen_authority_account` is a PDA and this is performing verification, I think it would be apt to either add a comment or rename the function `assert_correct_frozen_authority_address` or something of that nature.

universal-idl-program/program/src/account_checks.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+pub fn is_correct_account_type(data: &[u8], discriminator: [u8; 8]) -> bool {
+    data[..8] == discriminator
+}
+
+pub fn assert_program_authority_in_allowlist(account: &AccountInfo, name: &str) -> ProgramResult {
+    let allowlist_frozen_authorities: Vec<&str> =
+        ["6tgR1upn2bsMdiprpfUAWmoniEJ8E1XVF8f9gLmWqyTS"].to_vec();
```
jarry-xiao, line 146: Interesting approach:

List of admins that are able to be meta_authorities to support IDLs for frozen programs

universal-idl-program/program/src/instructions/create_idl/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+    program_authority: Pubkey,
+    idl_account: Pubkey,
+    program: Pubkey,
+    program_data_or_frozen_auth: Pubkey,
+) -> Result<Instruction, ProgramError> {
+    Ok(Instruction {
+        program_id,
+        accounts: vec![
+            AccountMeta::new(payer, true),
+            AccountMeta::new_readonly(program_authority, true),
```
jarry-xiao, line 47: How would this possible sign in the case that program_authority is a PDA of the current program?

jarry-xiao:
Currently don't see a way to:
1) crank buffers to upload IDL's
2) do anything for frozen authority

Will probably need to give this a closer look to fully understand

ngundotra:
Will set as `Ready for Review` once the following have been completed

1. Add tests:
  i.  Drift IDLs
  ii. Deploying IDL with SPL Governance
  
2. Address Jon Cinque's offline feedback:

Add `buffer_authority` that is only kp with ability to write to `buffer` so that goverance idl instructions are only `idl` `resize`s and `set_buffer`

3. Address Henry E's feedback:

Zero copy the IDL bytes from buffer to IDL to avoid breaking compute

universal-idl-program/program/src/instructions/create_idl/mod.rs
Henry-E: It would probably be better to just load up the program account and verify the upgrade authority everytime. Then it's stuck having the same security profile as the program itself.

universal-idl-program/program/src/instructions/create_idl/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+        &[&idl_seeds
+            .iter()
+            .map(|s| s.as_slice())
+            .collect::<Vec<&[u8]>>()],
+    )?;
+
+    let mut idl: Idl = Idl::new();
+    idl.authority = *ctx.program_authority.key;
+    idl.slot = Clock::get()?.slot;
+    idl.save(ctx.idl_account)?;
```
jarry-xiao, line 205: I think this is potentially suspect due to the memory limits (depending on how borsh deals with Vec<u8>). As long as there is not an intermediate serialized representation on the stack I think this call is safe, otherwise you might blow through the 4KB limit

universal-idl-program/program/src/instructions/set_buffer/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+    let buffer: Idl = Idl::from_account_info(ctx.buffer_account)?;
+
+    assert_address(
+        &idl.authority,
+        ctx.authority.key,
+        "Expected authority to be the same as the one in the idl account",
+    )?;
+
+    // Copy buffer data into idl account
+    idl.data = buffer.data.clone();
```
jarry-xiao, line 96: You might want to use a syscall for this

universal-idl-program/program/src/instructions/extend/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+};
+use borsh::{BorshDeserialize, BorshSerialize};
+use lazy_format::lazy_format;
+use solana_program::{account_info::next_account_info, pubkey::Pubkey, system_program};
+use solana_program::{account_info::AccountInfo, program_error::ProgramError};
+use solana_program::{entrypoint::ProgramResult, msg};
+use solana_program::{instruction::AccountMeta, rent::Rent};
+use solana_program::{instruction::Instruction, sysvar::Sysvar};
+
+pub fn extend(
```
jarry-xiao, line 17: This instruction name really confused me, I think `write_buffer` is a lot more sensible

jarry-xiao:
Question: Why not just upload IDLs to buffers directly with the BPF loader program. There's no actual way to check that the executable is valid (that's purely on a CLI level). Programs can decide to put the buffer address in their program `security.txt` and explorers can use that to parse the data

universal-idl-program/program/src/instructions/extend/mod.rs
ngundotra: it's not writing to the buffer however, it's extending.

How about `extend_buffer_data` ?

ngundotra:
Having a link from program ID to IDL address is only 1/2 of the problem. The other 1/2 of the problem is exposing instructions that allow program developers to modify their IDL. 

Additional features of having IDL managed by this UIP:
* Easier to deploy IDLs via SPL Gov votes
* We can annotate frozen programs with IDLs
* Programs can be smaller since they don't need Anchor IDL ix code inside them

jarry-xiao:
> Having a link from program ID to IDL address is only 1/2 of the problem. The other 1/2 of the problem is exposing instructions that allow program developers to modify their IDL.

Sounds like UIP should primarily function as an authority layer which makes sense. What the reason for reinventing the wheel on uploading bytes step? I feel like a cleaner design would be to allow users to upload buffers using BPF loader (fork the CLI and remove literally 1 check). Each Idl account holds a pointer to a buffer uploaded through BPF loader that contains the IDL data.

Main argument here is you can potentially reuse a good amount of infra

ngundotra:
Okay that is a really good point. That's a great idea.

However, we should always be able to fetch the IDL (or resolve its absence) in a *single* network request, and avoid fetching large amounts of data where possible. If we put the IDL in BPF loader bytes, then there are 3 network calls needed to fetch the IDL.

1. fetch program account
2. fetch IDL
  a. upgradeable program:
    i. fetch its program data account
    ii. read program data bytes & derive IDL address & fetch IDL account 
  b. frozen program:
   i. read program bytes & derive IDL address
   
It is a strong preference of this design that IDLs should be resolved in at most 1 network call. This preference allows us to build a lightweight transaction composition standard based on CPI descriptions in IDLs.

universal-idl-program/program/src/instructions/create_idl/mod.rs
ngundotra: You would have to have program instructions on the PDA's program that allow CPI to this program

jarry-xiao: I thought `program_authority` could be the `frozen_authority` derived from the current program.

Might have misread the first time around

2501babe:
im gonna try to find time to review this as well if i can, but i had a driveby question (having now read the linked rfc, but none of the pr code yet): is the anchor idl grammer sufficient to handle vanilla programs that may conditionally call `next_account_info` depending on their arguments?

jarry-xiao:
> im gonna try to find time to review this as well if i can, but i had a driveby question (having now read the linked rfc, but none of the pr code yet): is the anchor idl grammer sufficient to handle vanilla programs that may conditionally call `next_account_info` depending on their arguments?

It definitely is not, because to my knowledge it can only process static account lists as opposed to dynamic/arbitrary account trees. FWIW I think the static list case is probably fine for 95% of programs and the only reason to do anything besides that is to have optional inputs (which I think anchor now supports)

2501babe:
ahh ok so the goal is to cover the median program and not necessarily all of them

although, i gather these are meant to be hand-curated rather than generated, so i guess people can just throw together "whatever works." i think you could present multiple pseudo-instructions to cover branching logic in the processor

and yea static list is fine for almost if not all usecases, the problem is whether people use them anyway... ive def seen some that use branching to allow adding features while preserving backwards compat, optional accounts so the list isnt crowded by unused things, optional accounts that depend on args, etc

ngundotra:
Based on feedback in the #spl channel of the discord [here](https://discord.com/channels/428295358100013066/774014770402689065/1071170817968316416), I will be closing this PR and opening it back up as an SRC next week.

PR #4013, opened 2023-02-01
name-service-js: add tests

### LINKED ISSUE #4004: Missing TS Integration tests for Name Service JS bindings
There are no tests for the TS library for Solana Name Service program. 

Solution: write tests

atharmohammad:
fixes #4004

### DIFF
--- a/name-service/js/package.json
+++ b/name-service/js/package.json
@@ -23,26 +23,36 @@
     "prepublish": "tsc",
     "lint": "yarn pretty && eslint --max-warnings 0 'src/*.ts'",
     "lint:fix": "yarn pretty:fix && eslint 'src/*.ts' --fix",
-    "pretty": "prettier --check 'src/*.ts'",
-    "pretty:fix": "prettier --write 'src/*.ts'",
-    "doc": "yarn typedoc src/index.ts"
+    "pretty": "prettier --check '{src/*.ts,test/*/*.ts}'",
+    "pretty:fix": "prettier --write '{src/*.ts,test/*/*.ts}'",
+    "doc": "yarn typedoc src/index.ts",
+    "test": "yarn test:unit && yarn test:e2e",
+    "test:unit": "mocha test/unit",
+    "test:e2e": "start-server-and-test 'solana-test-validator --bpf-program namesLPneVptA9Z5rqUDD9tMTWEJwofgaYwp8cawRkX ../../target/deploy/spl_name_service.so --reset --quiet' http://localhost:8899/health 'mocha test/e2e'"
   },
   "prettier": {
     "singleQuote": true
   },
   "devDependencies": {
     "@tsconfig/recommended": "^1.0.1",
+    "@types/chai": "^4.3.4",
+    "@types/chai-as-promised": "^7.1.5",
+    "@types/mocha": "^10.0.1",
     "@types/node": "^14.14.20",
     "@typescript-eslint/eslint-plugin": "^4.0.1",
     "@typescript-eslint/parser": "^4.0.1",
     "babel-eslint": "^10.1.0",
+    "chai": "^4.3.7",
+    "chai-as-promised": "^7.1.1",
     "eslint": "^7.8.0",
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-eslint-comments": "^3.2.0",
     "eslint-plugin-functional": "^3.0.2",
     "eslint-plugin-import": "^2.22.0",
+    "mocha": "^10.2.0",
     "prettier": "^2.2.1",
-    "ts-node": "^9.1.1",
+    "start-server-and-test": "^1.15.3",
+    "ts-node": "^10.9.1",
     "typedoc": "^0.22.11",
     "typescript": "^4.1.3"
   },
@@ -51,5 +61,12 @@
     "@solana/web3.js": "^1.11.0",
     "bn.js": "^5.1.3",
     "borsh": "^0.4.0"
+  },
+  "mocha": {
+    "require": [
+      "ts-node/register"
+    ],
+    "recursive": true,
+    "extension": "ts"
   }
 }
--- a/name-service/js/test/e2e/index.test.ts
+++ b/name-service/js/test/e2e/index.test.ts
@@ -1,0 +1,129 @@
+import {
+  Connection,
+  Keypair,
+  LAMPORTS_PER_SOL,
+  PublicKey,
+  sendAndConfirmTransaction,
+  Transaction,
+} from '@solana/web3.js';
+import chai, { expect } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+
+import {
+  createNameRegistry,
+  deleteNameRegistry,
+  getHashedName,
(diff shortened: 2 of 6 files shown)


name-service/js/package.json
```diff
@@ -23,25 +23,35 @@
     "prepublish": "tsc",
     "lint": "yarn pretty && eslint --max-warnings 0 'src/*.ts'",
     "lint:fix": "yarn pretty:fix && eslint 'src/*.ts' --fix",
-    "pretty": "prettier --check 'src/*.ts'",
-    "pretty:fix": "prettier --write 'src/*.ts'",
-    "doc": "yarn typedoc src/index.ts"
+    "pretty": "prettier --check '{src/*.ts,test/*/*.ts}'",
+    "pretty:fix": "prettier --write '{src/*.ts,test/*/*.ts}'",
+    "doc": "yarn typedoc src/index.ts",
+    "test": "npm run test:unit && npm run test:e2e",
```
joncinque, line 29: nit for consistency:
```suggestion
    "test": "yarn test:unit && yarn test:e2e",
```

name-service/js/package.json
```diff
@@ -51,5 +61,16 @@
...
+  "mocha": {
+    "require": [
+      "ts-node/register"
+    ],
+    "recursive": true,
+    "extension": [
+      "ts",
+      "js",
+      "tsx"
+    ]
```
joncinque, line 74: nit: Let's keep this simple since everything is in ts
```suggestion
    "extension": [
      "ts",
    ]
```

name-service/js/test/e2e/index.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  before(async () => {
+    connection = new Connection(url, 'confirmed');
+    payer = Keypair.generate();
+    const airdropSignature = await connection.requestAirdrop(
+      payer.publicKey,
+      4 * LAMPORTS_PER_SOL
+    );
+    await connection.confirmTransaction(airdropSignature, 'confirmed');
+    name = '.sol';
+    nameKey = await getNameKey(name);
```
joncinque, line 41: nit: It looks like these are always the same, so should we just keep them as consts declared higher up?

name-service/js/test/e2e/index.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      owner.publicKey,
+      lamports
+    );
+    const tx = new Transaction().add(inst);
+    await sendAndConfirmTransaction(connection, tx, [payer]);
+    const nameAccount = await NameRegistryState.retrieve(connection, nameKey);
+    nameAccount.owner.equals(owner.publicKey);
+    expect(nameAccount.data?.length).to.eql(10);
+  });
+  it('Update Name Registery', async () => {
```
joncinque, line 63: Could you refactor all these tests so they don't depend on each other? Right now, for example, this test assumes that `name` has an existing registry, so instead could you create a new registry for each test as needed?  You could have it in the `before` clause

name-service/js/test/unit/index.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      new Numberu64(10)
+    );
+
+    expect(instruction.keys).to.have.length(6);
+    instruction.keys[0].pubkey.equals(SystemProgram.programId);
+    instruction.keys[1].pubkey.equals(payerKey);
+    instruction.keys[2].pubkey.equals(nameAccountKey);
+    instruction.keys[3].pubkey.equals(nameOwnerKey);
+    instruction.keys[4].pubkey.equals(new PublicKey(Buffer.alloc(32)));
+    instruction.keys[5].pubkey.equals(new PublicKey(Buffer.alloc(32)));
```
joncinque, line 48: For completeness, can you add a test that uses nameClassKey / nameParent / nameParentOwner and checks that they're properly filled in here?

name-service/js/yarn.lock
```diff
@@ -4,21 +4,21 @@
 
 "@babel/code-frame@7.12.11":
   version "7.12.11"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.12.11.tgz#f4ad435aa263db935b8f10f2c552d23fb716a63f"
+  resolved "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.11.tgz"
```
joncinque, line 7: Could you regenerate the yarn file using yarn? Looks like everything got changed to registry.npmjs.org rather than registry.yarnpkg.com

joncinque (commented):
This is really great work, thanks for your contribution! Just a few bits, then this should be good to go.

Can you also add a line to `ci/js-test-name-service.sh` to do `yarn test`? This way we can use these right away in CI! https://github.com/solana-labs/solana-program-library/blob/master/ci/js-test-name-service.sh is the file

name-service/js/package.json
```diff
@@ -23,26 +23,36 @@
...
     "eslint": "^7.8.0",
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-eslint-comments": "^3.2.0",
     "eslint-plugin-functional": "^3.0.2",
     "eslint-plugin-import": "^2.22.0",
+    "mocha": "^10.2.0",
     "prettier": "^2.2.1",
-    "ts-node": "^9.1.1",
+    "start-server-and-test": "^1.15.3",
+    "ts-node": "^10.9.1",
```
atharmohammad, line 55: have to update ts-node since it was giving error when running mocha with yarn. don't particularly know why but error was ``` Debug Failure. False expression: Non-string value passed to `ts.resolveTypeReferenceDirective` ```. Found on stackoverflow it might be due to ts-node version mismatch.

atharmohammad:
@joncinque how to build  name service program before js-test. Make cargo-test-sbf run before js-test ?

joncinque (commented):
The changes look good! Regarding the built program available to CI, you can see how other tests are doing it. Take a look at `pull-request-token-swap.yml` as an example. You'll see that it uploads the program at the end of the `cargo-test-sbf` step, and then it downloads the program during the `js-test` portion. You'll need to add steps to do the same thing in `pull-request-name-service.yml`

name-service/js/test/e2e/index.test.ts
```diff
@@ -0,0 +1,129 @@
...
+} from '../../src';
+
+chai.use(chaiAsPromised);
+const url = 'http://localhost:8899';
+
+describe('Name Service Program', () => {
+  let connection: Connection;
+  let payer: Keypair;
+  let owner: Keypair;
+  let nameKey: PublicKey;
```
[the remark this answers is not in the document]
atharmohammad: do you mean describe block ? in mocha making asynchronous describe is not supported. 
https://github.com/mochajs/mocha/issues/2975#issuecomment-326162912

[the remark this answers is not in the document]
joncinque: Ah crap, might be another reason to eventually move to jest :sweat_smile: ok no worries then, thanks for the info!

joncinque (approved):

PR #4015, opened 2023-02-01
stake-pool: Revert huge pool number to 3k

joncinque:
#### Problem

The monorepo downstream build is failing due to the bumped up huge pool number, see https://buildkite.com/solana-labs/solana/builds/89326

#### Solution

Bump it back down to 3k, where it was before


joncinque:
Merging right away to unblock CI

PR #4017, opened 2023-02-02
token-cli: Remove multisig owner as signer

joncinque:
#### Problem

As reported by a user, the multisig signer functionality is currently not working, since it's including the multisig address itself as a signer for the transaction.

#### Solution

Don't include the owner as a signer if we're in a multisig situation. The main changes are the replacements of `!bulk_signers.contains(&authority_signer)` with `config.multisigner_pubkeys.is_empty()` and then always filtering out the authority in the token client. I think that's correct, but please correct me if I'm wrong!

### DIFF
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -98,6 +98,26 @@ impl<'a> Config<'a> {
         .await
     }
 
+    fn extract_multisig_signers(
+        matches: &ArgMatches<'_>,
+        wallet_manager: &mut Option<Arc<RemoteWalletManager>>,
+        bulk_signers: &mut Vec<Arc<dyn Signer>>,
+        multisigner_ids: &'a mut Vec<Pubkey>,
+    ) -> Vec<&'a Pubkey> {
+        let multisig_signers = signers_of(matches, MULTISIG_SIGNER_ARG.name, wallet_manager)
+            .unwrap_or_else(|e| {
+                eprintln!("error: {}", e);
+                exit(1);
+            });
+        if let Some(mut multisig_signers) = multisig_signers {
+            multisig_signers.sort_by(|(_, lp), (_, rp)| lp.cmp(rp));
+            let (signers, pubkeys): (Vec<_>, Vec<_>) = multisig_signers.into_iter().unzip();
+            bulk_signers.extend(signers);
+            multisigner_ids.extend(pubkeys);
+        }
+        multisigner_ids.iter().collect::<Vec<_>>()
+    }
+
     pub(crate) async fn new_with_clients_and_ws_url(
         matches: &ArgMatches<'_>,
         wallet_manager: &mut Option<Arc<RemoteWalletManager>>,
@@ -117,18 +137,8 @@ impl<'a> Config<'a> {
         } else {
             solana_cli_config::Config::default()
         };
-        let multisig_signers = signers_of(matches, MULTISIG_SIGNER_ARG.name, wallet_manager)
-            .unwrap_or_else(|e| {
-                eprintln!("error: {}", e);
-                exit(1);
-            });
-        if let Some(mut multisig_signers) = multisig_signers {
-            multisig_signers.sort_by(|(_, lp), (_, rp)| lp.cmp(rp));
-            let (signers, pubkeys): (Vec<_>, Vec<_>) = multisig_signers.into_iter().unzip();
-            bulk_signers.extend(signers);
-            multisigner_ids.extend(pubkeys);
-        }
-        let multisigner_pubkeys = multisigner_ids.iter().collect::<Vec<_>>();
+        let multisigner_pubkeys =
+            Self::extract_multisig_signers(matches, wallet_manager, bulk_signers, multisigner_ids);
 
         let config = SignerFromPathConfig {
             allow_null_signer: !multisigner_pubkeys.is_empty(),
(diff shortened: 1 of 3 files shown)


token/cli/src/config.rs
```diff
@@ -117,18 +137,8 @@ impl<'a> Config<'a> {
...
-            });
-        if let Some(mut multisig_signers) = multisig_signers {
-            multisig_signers.sort_by(|(_, lp), (_, rp)| lp.cmp(rp));
-            let (signers, pubkeys): (Vec<_>, Vec<_>) = multisig_signers.into_iter().unzip();
-            bulk_signers.extend(signers);
-            multisigner_ids.extend(pubkeys);
-        }
-        let multisigner_pubkeys = multisigner_ids.iter().collect::<Vec<_>>();
+        let multisigner_pubkeys =
+            Self::extract_multisig_signers(matches, wallet_manager, bulk_signers, multisigner_ids);
```
joncinque, line 141: This just a little movement that I thought would be needed to test, and then it wasn't, but it was a neat little refactor so I kept it

token/client/src/token.rs
```diff
@@ -388,13 +388,12 @@ where
...
         &self,
         authority: &'b Pubkey,
-        signing_pubkeys: &'a Vec<Pubkey>,
+        signing_pubkeys: &'a [Pubkey],
     ) -> Vec<&'a Pubkey> {
-        if signing_pubkeys.as_ref() == [*authority] {
-            vec![]
-        } else {
-            signing_pubkeys.iter().collect::<Vec<_>>()
-        }
```
joncinque, line 397: I can't remember why we decided to special-case this, but it seems like we should always filter the authority from the multisig pubkeys, right? This breaks a self-owned multisig, ie a multisig where the address is also a participant. I'm not even sure if that's supported in the program to be honest.

joncinque: Never mind, I figured it out and reverted it! If it's self-owned, we don't want to remove `authority` from this list. But also, we need to work out nonces. If there's a nonce authority in the bulk signers, we need to make sure it's properly omitted here. Gonna think about this one

joncinque:
So my solution for the nonce authority is to hold onto it as a separate signer, and not include it in the bulk signers. This way, "bulk signers" really means "token authority signers", which is either the owner, or the multisig signers.

I've also added an offline signing test with a nonce to be sure we don't break this in the future, and properly mark the absent signers.

Let me know what you think! The nonce authority thing on the token client is a breaking change, but I think it should be fine.

token/client/src/token.rs
2501babe: i dont remember exactly how this interacted with nonce mode, but (iirc, that is, i know this is true but i dont know if there are other special cases im forgetting), the idea of this was...

originally, basically every signer got pushed into `bulk_signers`, which meant every token instruction received every transaction signer as a possible multisig signer. to deal with this, first we stopped pushing duplicates into the array. then we changed `payer` on `Token` to be a signer rather than a pubkey, so that we could exclude it from `bulk_signers` too. i _think_ at this point that meant the vec could only contain:
* one authority for a single-signer instruction
* multiple authorities for a multi-signer instruction
* one authority for the instruction plus the nonce authority in nonce mode

and distinguishing the second and third cases was not possible inside the token client, so.... ok yes i think this is all correct and covers the last case that i didnt

the detail of how all this came about is in #3509 and #3649 fwiw

2501babe:
ah and then i scroll down 1/3 of a screen and see you arrived at this conclusion already :sweat_smile:

token/cli/src/main.rs
```diff
@@ -3712,8 +3713,8 @@ async fn process_command<'a>(
 
             let (authority_signer, authority) =
                 config.signer_or_default(arg_matches, "authority", &mut wallet_manager);
-            if !bulk_signers.contains(&authority_signer) {
-                bulk_signers.push(authority_signer);
+            if config.multisigner_pubkeys.is_empty() {
+                push_signer_with_dedup(authority_signer, &mut bulk_signers);
```
2501babe, line 3717: this is basically just a santity check right? if `multisigner_pubkeys` has a length then its already in the array anyway so dont bother to check?

token/client/src/token.rs
```diff
@@ -388,9 +391,9 @@ where
     fn get_multisig_signers<'a, 'b>(
         &self,
         authority: &'b Pubkey,
-        signing_pubkeys: &'a Vec<Pubkey>,
+        signing_pubkeys: &'a [Pubkey],
```
2501babe, line 394: also ty for cleaning things like this up, i definitely used vecs in a few places when i started because i was fighting trait impls or the borrowchecker

2501babe (approved):

token/cli/src/main.rs
joncinque: Oh yeah that's true. Since fee payer and nonce authority are stored separately, we don't actually need to do the dedup check. I'll remove it and just do a normal push.

token/client/src/token.rs
joncinque: You can actually thank clippy for that! For some reason, it got triggered once there was a change in the implementation.

joncinque:
> ah and then i scroll down 1/3 of a screen and see you arrived at this conclusion already

It always makes me feel better when someone else independently arrives at the same result :smile:

token/cli/src/main.rs
joncinque: Oh nvm, I misunderstood your point. This `multisigner_pubkeys` check is crucial! If `multisigner_pubkeys` has a length, then we're in "multisig mode", so we don't want the authority to sign, since that's actually the multisig address, and that never signs. If it's added, then we'll create a transaction where it's a signer, which is incorrect.

But in "non-multisig mode", we need the authority to sign, so we must add it here. And just to be totally safe, I'll keep this as is with the dedup, just in case there's some bizarre edge case that we haven't considered.

2501babe: ohhh right, gotcha

PR #4025, opened 2023-02-03
[token-2022, confidential-extension] Add confidential transfer authority type

samkim-crypto:
#### Problem
In token-2022, there is currently an inconsistency in the way authorities associated with extensions are updated. In the confidential transfer (CT) extension, the instruction `ConfidentialTransferInstruction::UpdateMint` must be used to update the mint authority. However, in the rest of token-2022, `TokenInstruction::SetAuthority` instruction is used to update the authorities. 

#### Summary of Changes
- Extend the `AuthorityType` to include `AuthorityType::ConfidentialTransfer` and update the `SetAuthority` instruction to support confidential transfer authority
- Limit the `ConfidentialTransferInstruction::UpdateMint` to only update the mint parameters, not the authority.

##### Additional Safety when Updating the Confidential Transfer Authority
In addition to fixing the inconsistency, separating the ability to (1) update the CT mint authority and (2), update the rest of the CT mint parameters into two separate instructions can add an extra layer of safety for users when updating the mint.

If the authority is updated to None and the `ConfidentialTransferMint.auto_approve_new_accounts` is updated to false, then no new accounts can ever be approved for confidential transfers. Currently, a user can update both of these fields using the single `UpdateMint` instruction. With the changes above, the user must submit two different instructions to update the fields.

### DIFF
--- a/token/program-2022/src/instruction.rs
+++ b/token/program-2022/src/instruction.rs
@@ -1009,7 +1009,7 @@ pub enum AuthorityType {
     PermanentDelegate,
     /// Authority to update confidential transfer mint and aprove accounts for confidential
     /// transfers
-    ConfidentialTransfer,
+    ConfidentialTransferMint,
 }
 
 impl AuthorityType {
@@ -1024,7 +1024,7 @@ impl AuthorityType {
             AuthorityType::CloseMint => 6,
             AuthorityType::InterestRate => 7,
             AuthorityType::PermanentDelegate => 8,
-            AuthorityType::ConfidentialTransfer => 9,
+            AuthorityType::ConfidentialTransferMint => 9,
         }
     }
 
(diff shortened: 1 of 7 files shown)


nahnicholas (commented):
Ty

nahnicholas (commented):
Ty @nahnicholas

nahnicholas:
Ty

token/program-2022/src/instruction.rs
```diff
@@ -1007,6 +1007,9 @@ pub enum AuthorityType {
     InterestRate,
     /// Authority to transfer or burn any tokens for a mint
     PermanentDelegate,
+    /// Authority to update confidential transfer mint and aprove accounts for confidential
+    /// transfers
+    ConfidentialTransfer,
```
joncinque, line 1012: nit: can we call this `ConfidentialTransferMint`? Otherwise I could see some confusion from users trying to set their `ConfidentialTransfer` authority and thinking that account can perform confidential transfers on their behalf

joncinque (dismissed):
Other than the nit, looks good!

token/program-2022/src/instruction.rs
samkim-crypto: Yep, good call!

PR #4039, opened 2023-02-08
token-2022: Bump program, cli, and client for release

joncinque:
#### Problem

We've had a few changes to token-2022 since 0.5.0, including confidential transfer instruction changes, a fix to non-transferable tokens, and a CLI fix.

The CLI fix is the main driver for this, since it would be good to get a release out with the fix.

#### Solution

Bump token-cli, token-client, and token-2022 for a new release. Let me know if you have any additional bits you want to sneak in before cutting!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6398,7 +6398,7 @@ dependencies = [
 
 [[package]]
 name = "spl-token-cli"
-version = "2.2.0"
+version = "2.3.0"
 dependencies = [
  "assert_cmd",
  "clap 2.34.0",
@@ -6432,7 +6432,7 @@ dependencies = [
 
 [[package]]
 name = "spl-token-client"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "async-trait",
  "solana-cli-output",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6517,7 +6517,7 @@ dependencies = [
 
 [[package]]
 name = "spl-token-upgrade"
-version = "0.1.0"
+version = "0.1.1"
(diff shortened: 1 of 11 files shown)


2501babe (changes requested):
>We've had a few changes to token-2022 since 0.5.0, including CPI guard, permanent delegate, ...

these are actually in 0.5. it sounds like there may be new things that necessitate a token program release regardless (and maybe cli, possibly not client?), just flagging this to be sure

https://github.com/solana-labs/solana-program-library/releases/tag/token-2022-v0.5.0
https://github.com/solana-labs/solana-program-library/releases/tag/token-cli-v2.2.0

i didnt make tags or github pages for these because there havent been any in the past yet, but i also cut releases of token-client, token-upgrade, and token-upgrade-cli around the same commit

joncinque:
> these are actually in 0.5. it sounds like there may be new things that necessitate a token program release regardless (and maybe cli, possibly not client?), just flagging this to be sure

Ah sorry, yes you're right. I simply tried to do `cargo publish --dry-run` on the token-client after bumping, and it didn't work because we needed to publish a new token-2022.

Looking closer, the actual changes are in the confidential transfer extension.  I've updated the PR description to reflect that

joncinque:
> i didnt make tags or github pages for these because there havent been any in the past yet, but i also cut releases of token-client, token-upgrade, and token-upgrade-cli around the same commit

That's totally fine. I've also bumped token-upgrade and its CLI, thanks for flagging

2501babe (approved):
cool!

samkim-crypto (approved):

PR #4040, opened 2023-02-08
[token-2022] Create proof-program feature

samkim-crypto:
#### Problem
There is a PR in progress https://github.com/solana-labs/solana/pull/29996 that updates the proof verification program so that it break token-2022 confidential transfer. This results in downstream CI tests to fail.

#### Summary of Changes
Temporarily create a `proof-program` feature in the token-2022 program to disable the components that uses the proof program. Once https://github.com/solana-labs/solana/pull/29996 is merged and solana-program is updated, either make the `proof-program` feature as default or remove the feature altogether.

### DIFF
--- a/token/client/Cargo.toml
+++ b/token/client/Cargo.toml
@@ -24,3 +24,4 @@ thiserror = "1.0"
 [features]
 default = ["display"]
 display = ["dep:solana-cli-output"]
+proof-program = []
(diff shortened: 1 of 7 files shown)


joncinque:
Do you mind if I merge https://github.com/solana-labs/solana-program-library/pull/4039 before this? Then we can get a clean release out before all of the zk changes

samkim-crypto:
Yes, definitely! Let's release a clean version first.

PR #4048, opened 2023-02-09
Account Compression TS SDK: add `MerkleTree` static functions

ngundotra:
Additions:
- Easier to use constructor method `sparseMerkleTreeFromLeaves` which by default creates sparse MerkleTree leaves for you
- Easier to use `verify` & `hashProof` methods (now static)

Fix:
- Move `js-sha3` to dependencies, instead of `devDependencies`

### DIFF
--- a/account-compression/sdk/src/merkle-tree/index.ts
+++ b/account-compression/sdk/src/merkle-tree/index.ts
@@ -1,6 +1,6 @@
-import pkg from "js-sha3";
-import * as Collections from "typescript-collections";
-import { PublicKey } from "@solana/web3.js";
+import pkg from 'js-sha3';
+import * as Collections from 'typescript-collections';
+import { PublicKey } from '@solana/web3.js';
 const { keccak_256 } = pkg;
 
 let CACHE_EMPTY_NODE = new Map<number, Buffer>();
@@ -98,7 +98,7 @@ export class MerkleTree {
     let node = this.leaves[leafIndex];
 
     let height = 0;
-    while (typeof node.parent !== "undefined") {
+    while (typeof node.parent !== 'undefined') {
       if (minimizeProofHeight && height >= treeHeight) {
         break;
       }
@@ -113,7 +113,7 @@ export class MerkleTree {
         if (!hashed.equals(parent.node)) {
           console.log(hashed);
           console.log(parent.node);
-          throw new Error("Invariant broken when hashing left node");
+          throw new Error('Invariant broken when hashing left node');
         }
       } else {
         proof.push(parent.left!);
@@ -122,7 +122,7 @@ export class MerkleTree {
         if (!hashed.equals(parent.node)) {
           console.log(hashed);
           console.log(parent.node);
-          throw new Error("Invariant broken when hashing right node");
+          throw new Error('Invariant broken when hashing right node');
         }
       }
       node = parent;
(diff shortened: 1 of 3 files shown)


account-compression/sdk/src/merkle-tree/index.ts
```diff
@@ -53,6 +59,30 @@ export class MerkleTree {
...
+   *
+   * @param leaves leaves of the tree
+   * @param depth number of levels in the tree
+   * @returns MerkleTree
+   */
+  static sparseMerkleTreeFromLeaves(
+    leaves: Buffer[],
+    depth: number
+  ): MerkleTree {
+    const _leaves: Buffer[] = [];
```
joncinque, line 75: Do you need this at all? Or do you mean to pass this in when creating the tree to avoid modifying the user's `leaves` array? That would mean you'll need to push the empty buffers to this too

account-compression/sdk/src/merkle-tree/index.ts
```diff
@@ -1,6 +1,6 @@
-import pkg from 'js-sha3';
-import * as Collections from 'typescript-collections';
-import { PublicKey } from '@solana/web3.js';
+import pkg from "js-sha3";
+import * as Collections from "typescript-collections";
+import { PublicKey } from "@solana/web3.js";
```
joncinque, line 3: nit: can you revert the `'` -> `"` changes? the rest of the repo uses `'`

account-compression/sdk/src/merkle-tree/index.ts
```diff
@@ -143,11 +172,22 @@ export class MerkleTree {
       }
       if (verbose) console.log(`node ${i} ${new PublicKey(node).toString()}`);
     }
+    return node;
+  }
+
+  static verify(
+    root: string,
+    merkleTreeProof: MerkleTreeProof,
+    verbose = false
```
joncinque, line 181: nit: for consistency how about
```suggestion
    verbose: boolean = false
```

account-compression/sdk/src/merkle-tree/index.ts
```diff
@@ -143,11 +172,22 @@ export class MerkleTree {
...
+  ): boolean {
+    const node = MerkleTree.hashProof(merkleTreeProof, verbose);
     const rehashed = new PublicKey(node).toString();
     const received = new PublicKey(root).toString();
     if (verbose) console.log(`hashed ${rehashed} got ${received}`);
```
joncinque, line 186: nit: can you move this log to after your error? otherwise you'll get almost the same log twice for incorrect proofs in verbose mode

joncinque (commented):
A couple of questions, and can you also add a test for it?

ngundotra:
@jnwng `MerkleTree.sparseMerkleTreeFromLeaves` should be used to create trees instead of the normal constructor. The normal constructor is exposed for testing purposes, and for those that are familiar with the MerkleTree data structure.

account-compression/sdk/src/merkle-tree/index.ts
ngundotra: Meant to pass `_leaves` to ctor in order to leave user's array untouched. Thanks for catching!

account-compression/sdk/src/merkle-tree/index.ts
ngundotra: Fixed! Thanks

account-compression/sdk/src/merkle-tree/index.ts
ngundotra: Fixed!

account-compression/sdk/src/merkle-tree/index.ts
ngundotra: Fixed!

joncinque (approved):
Looks great!

PR #4055, opened 2023-02-15
README: Add program and crate information

joncinque:
#### Problem

We have a lot of crates and packages in SPL, but no way to see them all from the top-level.

#### Solution

Add info in the README


ngundotra:
There's also `spl-account-compression` and `@solana/spl-account-compression`!

joncinque:
Ah darn, yes! Thanks for noticing @ngundotra :pray: is [3345beb](https://github.com/solana-labs/solana-program-library/pull/4055/commits/3345beb315abe64edeecfde1a1146c5d4ca61f8a) good?

ngundotra:
Yup looks good! Thanks for including `spl-noop` & `spl-concurrent-merkle-tree` as well.

Yeah, docs are here: https://solana-labs.github.io/solana-program-library/account-compression/sdk/docs/

2501babe (dismissed):
incidentally i verified all the urls here resolve except the crates.io ones because ig cloudfront doesnt like curl

PR #4056, opened 2023-02-16
README: Update with latest token-2022 audit

joncinque:
#### Problem

A token-2022 audit report was recently added to the audits repo, but the spl readme is still referencing the previous one

#### Solution

Update the README with the newest audit


samkim-crypto (approved):
Looks good to me!

PR #4057, opened 2023-02-16
update monorepo version to v1.15.2

2501babe:
will add a second commit to clean `Pubkey::new` warnings. doing this as a separate pr so i can more easily rebase the single-pool branch. doing this at all because i need `solana_program::vote::state`, a new addition. making this pr now because i ran into an issue id like guidance on...


2501babe:
ok so ok im a bit out of my depth @joncinque 
```
hana@laptop:program$ pwd
/home/hana/work/solana/spl/token/program
hana@laptop:program$ cargo build-sbf
Error: Function _ZN14solana_program4vote5state9VoteState11deserialize17h1c6cbdac89c96039E Stack offset of 6344 exceeded max offset of 4096 by 2248 bytes, please minimize large stack variables
Error: Function _ZN229_$LT$solana_program..vote..state..vote_state_0_23_5.._..$LT$impl$u20$serde..de..Deserialize$u20$for$u20$solana_program..vote..state..vote_state_0_23_5..VoteState0_23_5$GT$..deserialize..__Visitor$u20$as$u20$serde..de..Visitor$GT$9visit_seq17hb16cc6a4176678f8E Stack offset of 5752 exceeded max offset of 4096 by 1656 bytes, please minimize large stack variables
```

i updated my spl branch to monorepo v1.15.2, used some of the new vote state reexports it provides in single-pool, and ran `cargo build-sbf` and got something like this. thinking i was doing something wrong, i tried boxing some things, but still got the errors. gradually commented out all the new code and still got the errors

eventually i got here: new branch off spl master, updated to 1.15, `rm -rf targets/`, error from building the token program. so... im 95% sure this is coming from upstream

im gonna do what i can to debug it. the next braindead thing ill try is sprinkling `#[inline(never)]` on likely looking monorepo functions in blind imitation of the stake pool program. it looks like its complaining because the deserialize function is allocated on the stack but im not sure why i would do that. im also... suspicious of the serde deserialize being in the mangled function signature? since everywhere ive seen deserialize the vote account stuff has explicitly used bincode... not really sure

but wanted to see if youve seen this kind of error before and had any insight on it. and also maybe a "yep thats a monorepo issue" so i feel less dumb opening a monorepo ticket without being super confident its actually a monorepo issue :sweat_smile:

2501babe:
incidentally im kind of baffled why this isnt in 1.14 because as far as i can tell it was added back in august https://github.com/solana-labs/solana/pull/26993

CriesofCarrots:
I've seen that error before. I have a vague recollection that it was due to a stale sbf cache, and that cargo clean resolved it. Been meaning to repro, since [others have been hitting this](https://github.com/solana-labs/solana/issues/30188). If you figure out the solution before I get back around to it, can you please post on that linked issue?

CriesofCarrots:
> incidentally not really important but im kind of baffled why this isnt in 1.14 because as far as i can tell it was added back in august [solana-labs/solana#26993](https://github.com/solana-labs/solana/pull/26993)

That's because we cut v1.14 (aka v1.11) off master in June 😝

joncinque:
So unfortunately you're out of luck with `Error: Function _ZN14solana_program4vote5state9VoteState11deserialize17h1c6cbdac89c96039E Stack offset of 6344 exceeded max offset of 4096 by 2248 bytes, please minimize large stack variables`

It roughly means that you can't call `VoteState::deserialize` in your program, since the function requires 6344 in stack space, when you only get a fixed 4096 bytes of stack. `#[inline(never)]` won't help you in this case, since the function is just too darn big. You could maybe write a custom deserializer, since the struct is 3731 bytes, but that seems like overkill.

To get around this, you can do the hacky thing and just read off the bytes from the right location.

2501babe:
ah yea i reached the same conclusion by just deleting code until i found the problem lol. continu(ed/ing) the convo in the issue tyera linked

this is an issue that affects everything that pulls in solana_program whether they use the type or not, so looks like im writing a deserialization function

2501babe:
closing this for now since i no longer need 1.15 for the single pool

PR #4059, opened 2023-02-23
Add GetSeq to CMT

austbot:
This pr adds a getSeq method to the zero copy CMT struct needed for indexing.


ngundotra:
Need to run `cargo clippy --fix --allow-dirty` or just checkout CI format workflow

PR #4063, opened 2023-02-24
Fix twoxtx setup vis-a-vis workspace inheritance

CriesofCarrots:
The cargo-test-sbf-twoxtx CI job seems broken since the monorepo landed workspace inheritance.
SPL is on v1.60, and workspace inheritance was added in v1.64. So the twotx job probably needs to inherit the monorepo's rust toolchain and use the monorepo's `cargo-test-sbf`

Also, the SPL workspace needs to exclude the nested twoxtx-solana monorepo.

Needs rebase on #4064

### DIFF
--- a/token/twoxtx-setup.sh
+++ b/token/twoxtx-setup.sh
@@ -25,4 +25,11 @@ if [[ ! -f twoxtx-solana/.twoxtx-patched ]]; then
 fi
 
 ../patch.crates-io.sh twoxtx-solana
+repo="token/twoxtx-solana"
+if sed -n '/exclude = \[/,/\]/p' ../Cargo.toml | grep -q "$repo"; then
+  echo "$repo is already excluded"
+else
+  sed -i'' ../Cargo.toml -e "/exclude/a \ \ \"$repo\","
+fi
+
 exit 0
--- a/.github/workflows/pull-request-token.yml
+++ b/.github/workflows/pull-request-token.yml
@@ -114,7 +114,7 @@ jobs:
       - name: Build and test token-2022 twoxtx (TEMPORARY)
         run: |
           ./token/twoxtx-setup.sh
-          cargo +${{ env.RUST_STABLE_VERSION }} test-sbf --manifest-path token/program-2022-test/Cargo.toml -- --nocapture
+          ./token/twoxtx-solana/cargo-test-sbf --manifest-path token/program-2022-test/Cargo.toml -- --nocapture
 
   js-test:
     runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


CriesofCarrots:
Curious. This works for me locally when I run patch.crates-io.sh, the toolchain file cp, and `cargo-test-bpf` all pointing at my existing monorepo (tip of master, with twotx.patch)

2501babe:
oh ive been running into the workspace issue today too (#4064 adds the toolchain file to `patch.crates-io.sh` for this very reason hehe)

it may be that the ci binaries are not tracking the required solana version? on my local machine i got bit by this because `cargo --version` was dictated by `rust-toolchain.toml` (1.16 -> 1.67) but `cargo build-sbf --version` was dictated by whichever version of the solana binaries i had happened to have installed (1.14 -> 1.62)

CriesofCarrots:
Thanks, @2501babe . @samkim-crypto and I figured we need to use the `cargo-test-bpf` from the twoxtx monorepo, and also add that monorepo to the SPL workspace exclusions. I'm just working on coding it up properly.
I'll remove the rust-toolchain.toml copy, though, since you're handling that within `patch.crates-io.sh`

CriesofCarrots:
Sorry for the premature ping, reviewers. However, @joncinque , do you have any clever suggestions?
The issue is that the `program-test-2022` build script uses `cargo build-sbf`, but needs to be using `token/twoxtx-solana/cargo-build-bpf` instead. Should we just build the 2 necessary programs manually for the twoxtx CI job?

token/twoxtx-setup.sh
```diff
@@ -25,4 +25,11 @@ if [[ ! -f twoxtx-solana/.twoxtx-patched ]]; then
 fi
 
 ../patch.crates-io.sh twoxtx-solana
+repo="token/twoxtx-solana"
+if sed -n '/exclude = \[/,/\]/p' ../Cargo.toml | grep -q "$repo"; then
+  echo "$repo is already excluded"
+else
+  sed -i'' ../Cargo.toml -e "/exclude/a \ \ \"$repo/\","
```
joncinque, line 32: nitty nit: is the slash required? I don't see it used for any of the other workspace items
```suggestion
  sed -i'' ../Cargo.toml -e "/exclude/a \ \ \"$repo\","
```

joncinque (commented):
> Should we just build the 2 necessary programs manually for the twoxtx CI job?

That seems like the most reasonable approach, assuming the build script is smart enough to not try to rebuild the programs. I can't come up with anything better

token/twoxtx-setup.sh
CriesofCarrots: Nope, looks like it's not necessary. I wonder if all the other lines could be condensed to just `farms`...

samkim-crypto (approved):

joncinque (approved):
Looks great, thanks for fixing!

PR #4064, opened 2023-02-24
track rust version when using local monorepo

2501babe:
as it stands, if you run `./patch.crates-io.sh /path/to/solana`, it updates you to the monorepo crates, but leaves you on the spl version of rust. this causes breakage when monorepo uses features that are stable on their version but unstable on ours

monorepo uses `rust-toolchain.toml` now so the fix is simple! and correct, since we always track the monorepo rust version when we formally update

### DIFF
--- a/patch.crates-io.sh
+++ b/patch.crates-io.sh
@@ -24,6 +24,8 @@ solana_dir=$(cd "$solana_dir" && pwd)
 cd "$(dirname "$0")"
 
 source "$solana_dir"/scripts/read-cargo-variable.sh
+cp "$solana_dir"/rust-toolchain.toml .
+
 # get version from Cargo.toml first. if it is empty, get it from other places.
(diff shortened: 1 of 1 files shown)


joncinque (dismissed):
Makes sense -- thanks for fixing!

patch.crates-io.sh
```diff
@@ -20,6 +20,8 @@ fi
 
 set -e
 
+cp "$solana_dir"/rust-toolchain.toml .
```
CriesofCarrots, line 23: This actually doesn't work for me when I run `./token/twoxtx-setup.sh` locally (although it was okay in CI, for some reason: https://github.com/solana-labs/solana-program-library/actions/runs/4268055882/jobs/7430253561)

Can we move below L26?

2501babe: aa yes ty!

CriesofCarrots (approved):

PR #4065, opened 2023-02-26
fix typo in confidential token docs

gr8den:


samkim-crypto:
Thanks for the catch!

PR #4067, opened 2023-02-27
ci: Only copy monorepo toolchain file if it exists

joncinque:
#### Problem

The patch crates io script tries to use the version of Rust from the monorepo by copying the rust-toolchain.toml file, but this file doesn't exist for releases prior to 1.15.

#### Solution

Only copy the toolchain file if it exists. Since this is blocking the monorepo on 1.14 currently, I'll merge this right away.


steviez (commented):
LGTM

PR #4068, opened 2023-02-28
docs: Update ATA derivation example to use `findProgramAddressSync`

quellen-sol:
**Problem**

The async `findProgramAddress` function is deprecated in favor of `findProgramAddressSync`, and the docs don't currently reflect that.

**Changes**

Update docs to show newcomer developers that this is the method that should be used when deriving ATAs and PDAs in general.


joncinque (approved):
Thanks for noticing this! Since the actual function in the JS library is called `getAssociatedTokenAddressSync`, there's no confusion with anything in the library, so we can keep the name as `findAssociatedTokenAddress` in this test function

Issue #3065, opened 2022-04-05
token-cli: Support confidential transfers


ghost:
I could not find anything in docs. Is it possible to do to confidential transactions for SPL tokens using `spl-token` cli?

Assuming this was added in solana-program-library: https://github.com/solana-labs/solana-program-library/tree/master/token/program-2022/src/extension/confidential_transfer

```
solana --version
solana-cli 1.10.6 (src:8d5c7b7d; feat:3235626988)

spl-token --version
spl-token-cli 2.0.15
```

cryptowalletr:
> I could not find anything in docs. Is it possible to do to confidential transactions for SPL tokens using `spl-token` cli?
> 
> 
> 
> Assuming this was added in solana-program-library: https://github.com/solana-labs/solana-program-library/tree/master/token/program-2022/src/extension/confidential_transfer
> 
> 
> 
> ```
> 
> solana --version
> 
> solana-cli 1.10.6 (src:8d5c7b7d; feat:3235626988)
> 
> 
> 
> spl-token --version
> 
> spl-token-cli 2.0.15
> 
> ```
> 
> 

Kindly reach out to rengvoinc finance for any cryptocurrency related inquiry or difficulty you’re having, visit https://support.rengvoinc.com/welcome-to-rengvoinc-wallet-support/ get swift response and help on how to solve the issue and any other related problem you are having.

1440000bytes:
> The CLI currently doesn't support this, but once Solana 1.11 lands, then we can land #3071 and add support using the `spl-token-client` crate

@joncinque last release for solana was 1.11 and #3071 is merged so can we use confidential transactions now?

joncinque:
I believe we still need to upgrade this repo to use 1.11, and we're holding off until 1.11 is branched off from master.  After that, this should definitely be unblocked.  The work will still need to be done to support the transfers, and we'll have to double the size of transactions on the network.

So there's still a few bits to be done before this is totally supported!

PR #4080, opened 2023-03-10
[token-2022, token-cli] Update CreateMint and Authorize to support confidential extension

samkim-crypto:
#### Problem
The token-cli does not yet support the confidential extension (https://github.com/solana-labs/solana-program-library/issues/3065).

Most of the confidential extension instructions would need support for ElGamal keys in the monorepo clap-utils. However, `CreateMint` and `Authorize` commands should still be able to be supported without these changes.

#### Summary of Changes
Add support for confidential extension to the `CreateMint` and `Authorize` commands. The rest of the confidential extension instructions will be supported in a separate PR once clap-utils is updated.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -834,7 +834,20 @@ async fn command_authorize(
                         ))
                     }
                 }
-                AuthorityType::ConfidentialTransferMint => unimplemented!(),
+                AuthorityType::ConfidentialTransferMint => {
+                    if let Ok(confidential_transfer_mint) =
+                        mint.get_extension::<ConfidentialTransferMint>()
+                    {
+                        Ok(COption::<Pubkey>::from(
+                            confidential_transfer_mint.authority,
+                        ))
+                    } else {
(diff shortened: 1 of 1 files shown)


2501babe (approved):

2501babe:
lgtm! also i pushed a commit so that `spl-token display` isnt broken, now that the cli can create mints with this extension

i thought a bit about the correct ux for this... i feel like confidential transfer fits in fine with `create-token` and `authorize` but some of the extension-specific commands should probably be behind a subcommand. id have to review what all the confidential commands actually are to have stronger design opinions for the future work here

samkim-crypto:
Great! I missed updating the display. Thanks for the update! I'll merge it in as soon the as the CI passes.

samkim-crypto:
Yeah, I think the right ux for the confidential extension commands is definitely something that we need to think about. We would definintely need separate subcommands for most of the confidential extension commands, but whether they should be grouped under single subcommand or not is a good question.

PR #4085, opened 2023-03-13
remove extra `can use` on account-compression docs

trankhacvy:
Remove duplicate `can use`.


joncinque (approved):

PR #4090, opened 2023-03-16
stake pool: dont pass unnecessary program infos

2501babe:
carrying over the love from #3974


joncinque (approved):
Thanks for cleaning this up!

PR #4092, opened 2023-03-20
More direct delegate test

Arrowana:
- Previous test seems to use owner but large amount, wrong code path?
- assertion that trying to use more than delegated amount with sufficient amount fails
- assertion that using someone else than delegate or owner after approving delegate fails
- trailing OwnerMismatch test for delegate being wiped when all delegated amount has been used remains, but unrelated to above necessary tests

### DIFF
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -2564,6 +2564,50 @@ mod tests {
         )
         .unwrap();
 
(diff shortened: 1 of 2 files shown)


joncinque (approved):
Sorry for the slow response -- these are definitely better, thanks for noticing and fixing!

PR #4094, opened 2023-03-23
[token-2022, token-cli] Enable migration of native SOL from an account to wSOL ATA 

Dzonixy:
Taking into account concerns noted in issue 2742 (https://github.com/solana-labs/solana-program-library/issues/2742) as well as the big impact of the allowing native SOL transfer to an token program owned PDA, I got an idea to try and solve the issue by including new instruction which will allow "rescue" of native SOL that are stuck on an account without violating the authority principle. Proof of concept PR https://github.com/solana-labs/solana-program-library/pull/2741 (https://github.com/solana-labs/solana-program-library/pull/2741) mentioned in the issue was used as a basis.

Problem
Native SOL transferred to a account by mistake are stuck and can't be transferred anywhere. This could have happened to anyone following for example the official Solana multisig docs and transfered lamports to a Multisig address, especially to the teams existing before solutions like Squads or Snowflake emerged.

Solution
The idea was to enable instruction which needs to be signed by an owner(s) of the account in order to transfer Native SOL (minus rent exemption) to the WrappedSol Associated Token Account owned by the signer(s). When Native SOLs are transferred to that ATA, owner(s) can handle WrappedSol tokens as intended. This way, the funds are still governed by the same authority, and more importantly the Native SOLs are not stuck anymore. This instruction cannot be invoked only for a Native SOL token account, which preserves all current logic.

Summary of Changes:

Adds an instruction RecoverLamports to both token programs, which moves native SOL (excluding rent) from a token program owned account to a previously initialised WrappedSOL ATA.
Adds an recover-lamports command to CLI which takes in the source account address and requires neccessary signatures .


joncinque:
Thanks for resubmitting this. As I mentioned in your other PR, we can definitely include this in token-2022, but not token. 

We can have this work for any type of account: mint, token account, and multisig, as `WithdrawExcessLamports`. Then you don't need to worry about wrapped SOL at all, and just move the excess lamports to some destination provided by the user.

For authorization, you can use the mint authority for mints, the owner for token accounts, and the multisig keys for multisigs.

What do you think?

Dzonixy:
Thanks, for the review. I agree about the authorization. I left the wrapped SOL as a requirement because it leaves the Lamports within the token programs and gives the authority the standard way to handle them, but if you think that's not necessary I'll remove the requirement and make it so that an arbitrary account can be passed. Do you see anything else that needs changing apart from removing this ix from token?

joncinque:
> I left the wrapped SOL as a requirement because it leaves the Lamports within the token programs and gives the authority the standard way to handle them, but if you think that's not necessary I'll remove the requirement and make it so that an arbitrary account can be passed

I don't see any particular reason to restrict it, since it's just lamports in the end. It'll make the command more useful.

> Do you see anything else that needs changing apart from removing this ix from token?

Conceptually, I think we're aligned -- I'll take a closer look once the changes from my previous comment are integrated.

Dzonixy:
@joncinque I removed changes made to token program and i renamed the instruction. Can you take a closer look and tell me what else can i do ti improve the PR?

token/client/src/token.rs
```diff
@@ -2429,4 +2429,27 @@ where
         )
         .await
     }
+
+    pub async fn withdraw_excess_lamports<S: Signers>(
+        &self,
+        source: &Pubkey,
+        authority: &Pubkey,
+        destination: &Pubkey,
```
joncinque, line 2437: nit for consistency with `transfer`:
```suggestion
        source: &Pubkey,
        destination: &Pubkey,
        authority: &Pubkey,
```

token/program-2022/src/processor.rs
```diff
@@ -1292,6 +1292,83 @@ impl Processor {
...
+        program_id: &Pubkey,
+        accounts: &[AccountInfo],
+    ) -> ProgramResult {
+        let account_info_iter = &mut accounts.iter();
+
+        let destination_info = next_account_info(account_info_iter)?;
+        let source_info = next_account_info(account_info_iter)?;
+        let authority_info = next_account_info(account_info_iter)?;
+
+        let source_data = source_info.data.borrow();
```
joncinque, line 1308: Since `validate_owner`  also borrows the data for `authority_info`, in the case of a multisig, you may get a double-borrow which would panic.

Could you add tests to successfully recover lamports from each a mint, multisig, and token account? You should also add tests for self-owned accounts, ie a token account or mint that is owned by itself. The multisig case should be covered, thankfully

token/program-2022/src/instruction.rs
```diff
@@ -1858,6 +1871,31 @@ pub(crate) fn encode_instruction<T: Into<u8>, D: Pod>(
...
 
+/// Creates a `WithdrawExcessLamports` Instruction
+pub fn withdraw_excess_lamports(
+    token_program_id: &Pubkey,
+    source_account: &Pubkey,
+    authority: &Pubkey,
+    destination_account: &Pubkey,
+    signers: Vec<&Pubkey>,
+) -> Result<Instruction, ProgramError> {
+    let mut accounts = vec![
```
joncinque, line 1882: Since this is only supported in token-2022, can you add a check, similar to other instructions like

```
check_program_account(token_program_id)?;
```

token/cli/src/main.rs
```diff
@@ -3586,6 +3624,41 @@ fn app<'a, 'b>(
...
+                        .required(true)
+                        .help("Specify the address of the account to recover lamports from"),
+                )
+                .arg(
+                    Arg::with_name("destination_account")
+                        .validator(is_valid_pubkey)
+                        .value_name("REFUND_ACCOUNT_ADDRESS")
+                        .takes_value(true)
+                        .required(true)
+                        .help("Specify the address of the account to recover lamports to"),
```
joncinque, line 3644: nit:
```suggestion
                        .help("Specify the address of the account to send lamports to"),
```

token/cli/src/main.rs
```diff
@@ -3586,6 +3624,41 @@ fn app<'a, 'b>(
                 )
                 .arg(mint_decimals_arg())
         )
+        .subcommand(
+            SubCommand::with_name(CommandName::WithdrawExcessLamports.into())
+                .about("Withdraw lamports from a Token Program owned account")
+                .arg(
+                    Arg::with_name("source_account")
+                        .validator(is_valid_pubkey)
+                        .value_name("ACCOUNT")
```
joncinque, line 3633: nit since you're clear with the other one:
```suggestion
                        .value_name("SOURCE_ACCOUNT_ADDRESS")
```

token/cli/src/main.rs
```diff
@@ -3586,6 +3624,41 @@ fn app<'a, 'b>(
                 )
                 .arg(mint_decimals_arg())
         )
+        .subcommand(
+            SubCommand::with_name(CommandName::WithdrawExcessLamports.into())
+                .about("Withdraw lamports from a Token Program owned account")
+                .arg(
+                    Arg::with_name("source_account")
```
joncinque, line 3631: nit for consistency with transfer
```suggestion
                    Arg::with_name("from")
```

token/cli/src/main.rs
```diff
@@ -3586,6 +3624,41 @@ fn app<'a, 'b>(
...
+                        .validator(is_valid_pubkey)
+                        .value_name("ACCOUNT")
+                        .takes_value(true)
+                        .required(true)
+                        .help("Specify the address of the account to recover lamports from"),
+                )
+                .arg(
+                    Arg::with_name("destination_account")
```
joncinque, line 3639: nit for consistency with transfer
```suggestion
                    Arg::with_name("recipient")
```

token/cli/src/main.rs
```diff
@@ -3586,6 +3624,41 @@ fn app<'a, 'b>(
...
+                        .value_name("KEYPAIR")
+                        .validator(is_valid_signer)
+                        .takes_value(true)
+                        .help(
+                            "Specify the keypair for the wallet which owns both the source \
+                             and the destination accounts. \
+                             This may be a keypair file or the ASK keyword. \
+                             Defaults to the client keypair."
+                        ),
+                )
```
joncinque, line 3658: with `owner_address_arg` I don't think you need this too

token/cli/src/main.rs
```diff
@@ -4293,6 +4366,22 @@ async fn process_command<'a>(
             )
             .await
         }
+        (CommandName::WithdrawExcessLamports, arg_matches) => {
+            let (signer, authority) =
+                config.signer_or_default(arg_matches, "owner", &mut wallet_manager);
+            push_signer_with_dedup(signer, &mut bulk_signers);
```
joncinque, line 4372: You'll need to only do this if it's *not* a multisig, ie
```suggestion
        if config.multisigner_pubkeys.is_empty() {
            push_signer_with_dedup(signer, &mut bulk_signers);
        }
```
Otherwise, it'll try to sign with the multisig address

token/cli/src/main.rs
```diff
@@ -2059,6 +2060,43 @@ async fn command_sync_native(config: &Config<'_>, native_account_address: Pubkey
...
 }
 
+async fn command_withdraw_excess_lamports(
+    config: &Config<'_>,
+    source_account: Pubkey,
+    destination_account: Pubkey,
+    authority: Pubkey,
+    bulk_signers: Vec<Arc<dyn Signer>>,
+) -> CommandResult {
+    let token = native_token_client_from_config(config)?;
```
joncinque, line 2070: Rather than using the native client, which can be confusing, let's explicitly use `Pubkey::default()` for the mint address, ie:
```suggestion
    // default is safe here because withdraw_excess_lamports doesn't use it
    let token = token_client_from_config(config, &Pubkey::default(), None)?;
```

token/cli/src/main.rs
```diff
@@ -2059,6 +2060,43 @@ async fn command_sync_native(config: &Config<'_>, native_account_address: Pubkey
...
+    println_display(
+        config,
+        format!(
+            "Recovering lamports from Account: {} to Account: {}",
```
joncinque, line 2074: nit
```suggestion
            "Withdrawing excess lamports\n  Sender: {}\n  Destination: {}",
```

token/cli/src/main.rs
```diff
@@ -7040,4 +7129,209 @@ mod tests {
...
+        .unwrap();
+
+        assert_eq!(
+            excess_lamports,
+            config
+                .rpc_client
+                .get_balance(&receiver.pubkey())
+                .await
+                .unwrap()
+        );
```
joncinque, line 7234: Since you're testing three different cases, you can make three separate tests for them. This seems like a nice break to start the second test

token/cli/src/main.rs
```diff
@@ -7040,4 +7129,209 @@ mod tests {
...
+        assert_eq!(
+            2 * excess_lamports,
+            config
+                .rpc_client
+                .get_balance(&receiver.pubkey())
+                .await
+                .unwrap()
+        );
+
+        let token_account =
```
joncinque, line 7294: And you can start the third test here

joncinque (commented):
Really great work! It's mostly cleanup nits, some more general token-2022 support, some program-level tests to complement the CLI tests, and then this should be very close

joncinque:
oh also, you'll need to run `clippy fmt --all` to format everything properly, and `cargo +nightly-2022-04-01 clippy -Zunstable-options --all-targets --features test-sbf -- --deny=clippy::integer_arithmetic` to make sure all the lints pass

joncinque (commented):
Confirming that all of the comments marked resolved from the first review have been correctly resolved, please ping me when the last few points are ready!

Dzonixy:
@joncinque, my on-chain unit test don't pass because of the `Rent` sysvar. Any advice on how to handle this? Seems like there is no need to pass it in, but there is no test in the program that doesn't pass in `Rent` and uses `Rent::get()?`. Should i do it like in the rest of the and make the rent account optional but only test for when it's sent?

joncinque:
> my on-chain unit test don't pass because of the Rent sysvar. Any advice on how to handle this?

Unless you're testing just a tiny portion, it should be available since the syscall stub is implemented: https://github.com/solana-labs/solana-program-library/blob/139ed61e84f027d62e6a179e56f2af38d990a2f6/token/program-2022/src/processor.rs#L1602

You may need to set the syscall stubs like at https://github.com/solana-labs/solana-program-library/blob/139ed61e84f027d62e6a179e56f2af38d990a2f6/token/program-2022/src/processor.rs#L1618-L1625

Can you push the failing test to your branch? Then it'll be much easier to diagnose and come up with a better solution

joncinque:
@Dzonixy how are things going on this? I'd like to integrate it within the next week, so if you have some time to keep going, I'd really appreciate it. If not, I can take over the branch and get the last bits done.

Dzonixy:
@joncinque i'm busy at the moment, but i'll try to finish it by the end of the week. I'll let you know whether i made it or not.

joncinque:
@Dzonixy sounds good, and yes please let me know either way. This is really close, so it would be a shame to leave it out!

yogevyuval:
@Dzonixy Any news regarding this PR? will be much appreciated

Dzonixy:
@joncinque I think i did it. The setting of the syscall stubs did the trick. I managed to write the test for Multisig, i'll do two more for Account and Mint by the end of the day.

Dzonixy:
@joncinque it's done. Let me know if i should change it in some way.

token/program-2022/src/processor.rs
```diff
@@ -7602,4 +7677,249 @@ mod tests {
             )
         );
     }
+
+    #[test]
+    #[serial]
```
joncinque, line 7682: nit: I see why you were running into the issue now -- `do_process_instruction_dups` doesn't set the syscalls. The best solution would be to have `do_process_instruction` call into `do_process_instruction_dups`, and set the syscalls once in `do_process_instruction_dups`, but we can do that in follow-up

token/program-2022/src/instruction.rs
```diff
@@ -1892,6 +1905,33 @@ pub(crate) fn encode_instruction<T: Into<u8>, D: Pod>(
...
+    destination_account: &Pubkey,
+    authority: &Pubkey,
+    signers: &[&Pubkey],
+) -> Result<Instruction, ProgramError> {
+    check_program_account(token_program_id)?;
+
+    let mut accounts = vec![
+        AccountMeta::new(*source_account, false),
+        AccountMeta::new(*destination_account, false),
+        AccountMeta::new(*authority, signers.is_empty()),
```
joncinque, line 1921: This should be 
```suggestion
        AccountMeta::new_readonly(*authority, signers.is_empty()),
```

token/program-2022/src/processor.rs
```diff
@@ -1371,6 +1371,77 @@ impl Processor {
...
+            }
+        } else if source_data.len() == Multisig::LEN {
+            Self::validate_owner(
+                program_id,
+                source_info.key,
+                authority_info,
+                authority_info.data_len(),
+                account_info_iter.as_slice(),
+            )?;
+        }
```
joncinque, line 1423: Out of an absolute abundance of caution can you add an `else` case that errors? ie
```suggestion
        } else {
            return Err(TokenError::InvalidState.into());
        }
```

token/program-2022/src/processor.rs
```diff
@@ -1371,6 +1371,77 @@ impl Processor {
...
+                program_id,
+                &account.base.owner,
+                authority_info,
+                authority_info.data_len(),
+                account_info_iter.as_slice(),
+            )?;
+        } else if let Ok(mint) = StateWithExtensions::<Mint>::unpack(&source_data) {
+            if &mint.base.mint_authority.expect("No mint authority") != authority_info.key {
+                return Err(TokenError::AuthorityTypeNotSupported.into());
+            }
```
joncinque, line 1403: Can you remove this? The check is being done right after

token/program-2022/src/processor.rs
```diff
@@ -7602,4 +7677,249 @@ mod tests {
...
+
+        let account_info: AccountInfo = (&account_key, true, &mut account_account).into();
+
+        do_process_instruction_dups(
+            withdraw_excess_lamports(
+                &program_id,
+                &account_key,
+                &destination_key,
+                &owner_key,
+                &[&owner_key],
```
joncinque, line 7842: nit: since this isn't owned by a multisig, this param should be 
```suggestion
                &[],
```

token/program-2022/src/processor.rs
```diff
@@ -7602,4 +7677,249 @@ mod tests {
...
+        let owner_info = AccountInfo::new(
+            &owner_key,
+            true,
+            false,
+            &mut owner_lamports,
+            &mut owner_data,
+            &system_program_id,
+            false,
+            Epoch::default(),
+        );
```
joncinque, line 7797: To test the case that I was worried about, could you have `account_account` owned by `account_key` rather than by `owner_key`? You'll probably need to use `process_instruction_dups` for `initialize_account` too

token/program-2022/src/processor.rs
```diff
@@ -7602,4 +7677,249 @@ mod tests {
...
+
+        let account_info: AccountInfo = (&mint_key, true, &mut mint_account).into();
+
+        do_process_instruction_dups(
+            withdraw_excess_lamports(
+                &program_id,
+                &mint_key,
+                &destination_key,
+                &owner_key,
+                &[&owner_key],
```
joncinque, line 7912: nit: same here
```suggestion
                &[],
```

token/program-2022/src/processor.rs
```diff
@@ -7602,4 +7677,249 @@ mod tests {
...
+        let owner_info = AccountInfo::new(
+            &owner_key,
+            true,
+            false,
+            &mut owner_lamports,
+            &mut owner_data,
+            &system_program_id,
+            false,
+            Epoch::default(),
+        );
```
joncinque, line 7875: and same here, can you have `mint_account` have `mint_key` as its mint authority?

joncinque (commented):
We're extremely close! Just a few last bits to clean up and some improvements for the program-level tests

token/program-2022/src/processor.rs
Dzonixy: Sure, i'll get on it.

Dzonixy: Could you take a look? No matter what i put in as the authority the process execute successfully... I can't see why

joncinque: You're passing in the correct `AccountInfo`, which is all that really matters in this weird test environment, and not the accounts in the instruction. Here's the patch:

```
diff --git a/token/program-2022/src/processor.rs b/token/program-2022/src/processor.rs
index bdc4556d..49d69954 100644
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -7780,20 +7780,6 @@ mod tests {
         );

         let system_program_id = system_program::id();
-        // let mut owner_lamports = 0;
-        // let mut owner_data = vec![];
-        let owner_key = Pubkey::new_unique();
-        // let mut owner_account = SolanaAccount::new(0, 0, &system_program::id());
-        // let owner_info = AccountInfo::new(
-        //     &owner_key,
-        //     true,
-        //     false,
-        //     &mut owner_lamports,
-        //     &mut owner_data,
-        //     &system_program_id,
-        //     false,
-        //     Epoch::default(),
-        // );

         let mut destination_lamports = 0;
         let mut destination_data = vec![];
@@ -7814,7 +7800,7 @@ mod tests {

         let mut rent_sysvar = rent_sysvar();
         do_process_instruction(
-            initialize_mint(&program_id, &mint_key, &owner_key, None, 2).unwrap(),
+            initialize_mint(&program_id, &mint_key, &mint_key, None, 2).unwrap(),
             vec![&mut mint_account, &mut rent_sysvar],
         )
         .unwrap();
@@ -7864,20 +7850,7 @@ mod tests {
         let excess_lamports = 4_000_000_000_000;

         let program_id = crate::id();
-        let mut owner_lamports = 0;
-        let mut owner_data = vec![];
         let system_program_id = system_program::id();
-        let owner_key = Pubkey::new_unique();
-        let owner_info = AccountInfo::new(
-            &owner_key,
-            true,
-            false,
-            &mut owner_lamports,
-            &mut owner_data,
-            &system_program_id,
-            false,
-            Epoch::default(),
-        );

         let mut destination_lamports = 0;
         let mut destination_data = vec![];
@@ -7901,7 +7874,7 @@ mod tests {
         let mut rent_sysvar = rent_sysvar();

         do_process_instruction(
-            initialize_mint(&program_id, &mint_key, &owner_key, None, 2).unwrap(),
+            initialize_mint(&program_id, &mint_key, &mint_key, None, 2).unwrap(),
             vec![&mut mint_account, &mut rent_sysvar],
         )
         .unwrap();
@@ -7909,12 +7882,12 @@ mod tests {
         let account_info: AccountInfo = (&mint_key, true, &mut mint_account).into();

         do_process_instruction_dups(
-            withdraw_excess_lamports(&program_id, &mint_key, &destination_key, &program_id, &[])
+            withdraw_excess_lamports(&program_id, &mint_key, &destination_key, &mint_key, &[])
                 .unwrap(),
             vec![
                 account_info.clone(),
                 destination_info.clone(),
-                owner_info.clone(),
+                account_info.clone(),
             ],
         )
         .unwrap();
```

If you pass in something other than `account_info` in that last changed line, it'll error properly

Dzonixy: I see now, thanks. It's been some time since i last looked at this, so i forgot the details.

joncinque (approved):
This is finally ready to merge! Thanks so much for your effort and persistence, it's really appreciated

rugged1111:
> This is finally ready to merge! Thanks so much for your effort and persistence, it's really appreciated

Will this work on a token that has been on market since 2024? I've sent sol straight to the this token's contract address, H1aN3vcvB68eaFPbMkoAss3vnfi4AhP5C2dpnrZzdBc7 must still find a way how to get in touch with them. I wanted to buy their memecoin not to send them my sol as a gift.

PR #4105, opened 2023-03-31
permissioned-transfer: Add basic example program for checking permissions

joncinque:
#### Problem

We want to add permissioned transfers to token-2022, in a similar vein as #3888. The slight difference is that we'll only store the program id in the mint in token-2022, and the rest of the logic is done by the program.

#### Solution

Outline the interface that should be implemented by other devs, along with a simple example program.

Here's the full list of things to do:

- [x] make permissioned-transfer program
- [x] add end-to-end tests
- [x] improve `ValidationPubkeys` to be any length (currently hardcoded to 3 to work simply with bytemuck)
- [x] add on-chain instruction creator which reads `ValidationPubkeys` and pulls out accounts from an `AccountInfo` iterator
- [x] add hook to token-2022 to store the program id and call `validate` if a permissioned token program is specified
- [ ] add on-chain instruction creator which reads a `Mint` and pulls out all additional accounts from an `AccountInfo` iterator
- [ ] add to CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6268,6 +6268,7 @@ dependencies = [
 name = "spl-permissioned-transfer"
 version = "0.1.0"
 dependencies = [
+ "arrayref",
  "bytemuck",
  "num-derive",
  "num-traits",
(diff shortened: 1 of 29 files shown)


joncinque:
Closing since this has been implemented through a bunch of other PRs

Issue #4077, opened 2023-03-09
single-validator stake pools


2501babe:
tldr we are writing a stripped-down stake pool program for the single-validator case that enables liquid staking with no fees and 100% capital efficiency. the pr comments are a mess because it was originally a poc so im going to track work to-be done here

present status: code is done, pending review from jon, then we discuss/make changes and send into preliminary audit

work (no blockers):
* fuzz deposit/withdraw calculations, including how they hold up post-rewards
* port over all `program/tests/` code to `stake-pool-tests/tests/`
* write a `build.rs` plus whatever yaml files (if needed?) to get the new stuff running on ci
* update cli to work with both programs
* update js to work with both programs

work (pending discussion or upstream changes):
* assign this a vanity address. i dont know what the process is for how we manage program keys, size mainnet program buffers (but we can realloc program accounts now maybe?), pay fees, etc
* click-to-stake instruction(s). this probably depends on the stake program adding multistake accounts. the original idea i had for this was: the user creates and delegates a new stake account and sets withdrawer to the pool authority and staker to their wallet or token account or something. (or we could just alloc a special account. maybe a linked list.) when it activates, anyone can permissionlessly fix `Authorized`, sweep stake into the pool, and mint tokens to the user. perhaps the user fronts a bounty to incentivize next-epoch merges. with multistake accounts, i believe the flow is much the same, except we track the user via a special account, and the stake is batched in the multistake intake. minting tokens after activation is a design necessity to obviate the need for fees (but again we may have users pay a fee to the minter for timeliness)
* slippage. we may want to take this as an argument to `deposit` and `withdraw`. we are not exposed to the "manager frontruns a fee change" attack vector the multi-pool is, but we are exposed to sandwich attacks. in theory with access to stake ≫ pool value, an mever could intercept ~all user deposits. we should at least debate whether this is purely theoretical or an actual plausible attack vector

questions / comments / slander / invective:
* is anyone assigned to the multistake impl? should i be working on this?
* we will need to do some light rework to accomodate multistake accounts. ideally we want every pool to have a multistake and no other stake account. if multistake lands before we deploy, just edit the code to use them and audit again. but if multistake will land _after_, we want to design a clean upgrade path before _we_ deploy, so that the pools can transition without leaving behind legacy cruft

mvines:
> but if multistake will land after, we want to design a clean upgrade path before we deploy, so that the pools can transition without leaving behind legacy cruft

[...]

PR #4117, opened 2023-04-13
single-validator stake pool tests

2501babe:
part of #4077


2501babe:
sorry for the big line count but its 100% tests so nothing too serious

stake-pool/single-pool/tests/helpers/mod.rs
```diff
@@ -0,0 +1,423 @@
...
+            got, expected
+        );
+    };
+
+    // this silly thing is because we can guarantee From<T> has a Debug for T
+    // but TryFrom<T> produces Result<T, E> and E may not have Debug. so we cant call unwrap
+    // also we use TryFrom because we have to go `instruction error-> program error`
+    // because StakeError impls the former but not the latter...
+    // and that conversion is merely surjective........
+    // infomercial lady: "if only there were a better way!"
```
joncinque, line 411: Oof yeah, it's awful. If we need to add `Debug` to any sdk types to make this easier, I'm super game

stake-pool/single-pool/tests/helpers/token.rs
```diff
@@ -0,0 +1,54 @@
...
+        program_pack::Pack,
+        pubkey::Pubkey,
+        signature::{Keypair, Signer},
+        transaction::Transaction,
+    },
+    spl_associated_token_account as atoken,
+    spl_token::state::{Account, Mint},
+};
+
+pub async fn create_ata(
```
joncinque, line 19: Not needed for this PR, but `TokenClient` was designed to avoid files like this, so whatever we need to eventually get rid of this and use `TokenClient` would be nice

stake-pool/single-pool/tests/helpers/token.rs
```diff
@@ -0,0 +1,54 @@
...
+    banks_client: &mut BanksClient,
+    payer: &Keypair,
+    owner: &Pubkey,
+    recent_blockhash: &Hash,
+    pool_mint: &Pubkey,
+) {
+    #[allow(deprecated)]
+    let instruction = atoken::create_associated_token_account(&payer.pubkey(), owner, pool_mint);
```
joncinque, line 27: nit: If we don't use the new stuff, who will? :smiling_face_with_tear: 
```suggestion
    let instruction = atoken::instruction::create_associated_token_account(&payer.pubkey(), owner, pool_mint, &spl_token::id());
```

stake-pool/single-pool/tests/initialize.rs
```diff
@@ -0,0 +1,54 @@
...
+        minimum_delegation,
+    );
+    let message = Message::new(&instructions, Some(&context.payer.pubkey()));
+    let transaction = Transaction::new(&[&context.payer], message, context.last_blockhash);
+
+    context
+        .banks_client
+        .process_transaction(transaction)
+        .await
+        .unwrap_err();
```
joncinque, line 53: nit: Can you check the error here?

stake-pool/single-pool/tests/create_pool_token_metadata.rs
```diff
@@ -0,0 +1,59 @@
...
+        &context.payer.pubkey(),
+    );
+    let message = Message::new(&[instruction], Some(&context.payer.pubkey()));
+    let transaction = Transaction::new(&[&context.payer], message, context.last_blockhash);
+
+    context
+        .banks_client
+        .process_transaction(transaction)
+        .await
+        .unwrap_err();
```
joncinque, line 58: Can you check the error here?

stake-pool/single-pool/tests/accounts.rs
```diff
@@ -0,0 +1,177 @@
...
+        let mut instructions = instructions.clone();
+        let instruction_account = &mut instructions[i].accounts[j];
+
+        // wallet address can be arbitrary
+        if instruction_account.pubkey == accounts.alice.pubkey() {
+            continue;
+        }
+
+        let prev_pubkey = instruction_account.pubkey;
+        instruction_account.pubkey = Keypair::new().pubkey();
```
joncinque, line 150: nit: Can use `Pubkey::new_unique()` here too
```suggestion
        instruction_account.pubkey = Pubkey::new_unique();
```

stake-pool/single-pool/tests/accounts.rs
```diff
@@ -0,0 +1,177 @@
...
+            check_error(e, SinglePoolError::InvalidPoolAuthority)
+        } else if prev_pubkey == accounts.mint {
+            check_error(e, SinglePoolError::InvalidPoolMint)
+        } else if [system_program::id(), spl_token::id(), stake::program::id()]
+            .contains(&prev_pubkey)
+        {
+            check_error(e, ProgramError::IncorrectProgramId)
+        }
+
+        // TODO explicitly check clock/rent/stake history/stake config? we just let the stake program do it
```
joncinque, line 175: If you can trust the program you call into, and your program doesn't use the account explicitly, my take is to not check the accounts

stake-pool/single-pool/tests/deposit.rs
```diff
@@ -0,0 +1,453 @@
...
+            &mut context.banks_client,
+            &context.payer,
+            &context.last_blockhash,
+            &accounts.stake_account,
+            extra_lamports,
+        )
+        .await;
+    }
+
+    let mut fees = USER_STARTING_LAMPORTS - wallet_lamports_after_stake - stake_lamports;
```
joncinque, line 96: If you don't want to worry about fees, how about setting `context.payer` as the fee payer?

stake-pool/single-pool/tests/deposit.rs
```diff
@@ -0,0 +1,453 @@
...
+        &stake_account.pubkey(),
+        &token_account,
+        &context.payer.pubkey(),
+        &context.payer.pubkey(),
+    );
+    let message = Message::new(&instructions, Some(&context.payer.pubkey()));
+    let transaction = Transaction::new(&[&context.payer], message, context.last_blockhash);
+
+    // this gives a random borsh error upon attempting to deserialize the mint
+    // TODO perhaps in a separate pr, wrap this for all processors with a more helpful error
```
joncinque, line 328: That actually makes sense -- to save on compute, you're deserializing the stake account with Borsh in the program. Since the stake account is uninitialized, you'll get that Borsh error

joncinque (commented):
This is really great work! Mostly small things, but you've covered all the bases, and looks like these tests give very complete coverage

stake-pool/single-pool/src/processor.rs
```diff
@@ -1029,3 +1029,323 @@ impl Processor {
...
+                        // tokens per supply increased with stake per total
+                        if prev_total_stake > 0 {
+                            assert_eq!(
+                                format!("{:.6}", pool.share(user) - prev_share),
+                                format!(
+                                    "{:.6}",
+                                    pool.stake(user) as f64 / pool.total_stake as f64
+                                        - prev_stake as f64 / prev_total_stake as f64
+                                )
+                            );
```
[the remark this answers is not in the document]
2501babe: because to my great shame i got an error like this
```
error[E0283]: type annotations needed for `Uniform<{float}>`
    --> stake-pool/single-pool/src/processor.rs:1193:24
     |
1193 |         let op_range = Uniform::from(if with_rewards { 0.0..1.0 } else { 0.0..0.65 });
     |             --------   ^^^^^^^ cannot infer type for type `{float}`
     |             |
     |             consider giving `op_range` the explicit type `Uniform<{float}>`, where the type parameter `{float}` is specified
     |
     = note: multiple `impl`s satisfying `{float}: SampleUniform` found in the `rand` crate:
             - impl SampleUniform for f32;
             - impl SampleUniform for f64;
note: required by a bound in `Uniform`
    --> /home/hana/.cargo/registry/src/github.com-1ecc6299db9ec823/rand-0.8.5/src/distributions/uniform.rs:179:23
     |
179  | pub struct Uniform<X: SampleUniform>(X::Sampler);
     |                       ^^^^^^^^^^^^^ required by this bound in `Uniform`
help: consider specifying the type argument in the function call
     |
1193 |         let op_range = Uniform::<X>::from(if with_rewards { 0.0..1.0 } else { 0.0..0.65 });
     |                               +++++
```

and figured i was doing something wrong with the macro, without realizing it was just that i was passing an int for a float...

stake-pool/single-pool/tests/helpers/mod.rs
2501babe: unless there is A Better Way, its not that simple... we simply dont have access to `E` in the context of the type signature because its not in the signature of `TryFrom<T>`. `TryFrom` itself would need to restrict its `E` to only those which impl `Debug`

instead i think we want to `impl From<StakeError> for ProgramError` but im not sure its worth touching the stake program to make some peripheral tests nicer... although maybe it would help someone else someday

stake-pool/single-pool/tests/helpers/token.rs
2501babe: this is my first time thinking about using it outside the cli, and thinking as a user, the main ergonomic hurdle is constructing the `ProgramClient`. i want a function like `build_from_mint_address` or something that i just pass in... i guess my rpc or banks client and the mint address and it makes the program client, fetches the mint, fills out the decimals and program id. just a first idea

stake-pool/single-pool/tests/deposit.rs
2501babe: yea i just mean if im a user trying to deposit into a pool that doesnt exist i would see borsh garbage but would rather see "someone needs to initialize this pool"

stake-pool/single-pool/tests/helpers/token.rs
joncinque: That makes sense. We should consider adding `ProgramClient` into one of the monorepo crates so that the pattern is easier to follow for other program clients, along with some nice helpers like you're suggesting.

joncinque (approved):
Looks great!

PR #4121, opened 2023-04-14
ci: Fix twoxtx build by properly patching crates

joncinque:
#### Problem

The twoxtx build is failing to properly patch the local checkout of the monorepo because of workspace dependencies.

#### Solution

It's a bit of a kludge, but just copy over the workspace dependencies and package info from the top-level Cargo.toml, and rewrite the paths. Since it only does something different if `[workspace.package]` and `[workspace.dependencies]` are defined, it's meant to work with the downstream build on older and newer versions of the monrepo.

### DIFF
--- a/patch.crates-io.sh
+++ b/patch.crates-io.sh
@@ -9,6 +9,30 @@ if [[ -z $solana_dir ]]; then
   exit 1
 fi
 
+copyCargoSection() {
+  declare section="$1"
+  declare source_cargo_toml="$2"
+  declare destination_cargo_toml="$3"
+
+  if grep -q "\[$section\]" "$destination_cargo_toml"; then
+    echo "$destination_cargo_toml already has $section"
+  else
+    copy_package_info=false
+    while read line;
+    do
+      if [[ ("$line" == "" || "$line" = \[*) && "$copy_package_info" == "true" ]]; then
+        copy_package_info=false
+      fi
+      if [[ "$line" == "[$section]" ]]; then
+        copy_package_info=true
+      fi
+      if [[ "$copy_package_info" == "true" ]]; then
+        echo "$line" >> "$destination_cargo_toml"
+      fi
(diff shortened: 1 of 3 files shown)


yihau:
I think maybe this one can be the stopgap.
btw, I can't reproduce the same issue on my small test project. They just work without importing those workspace stuff 🤔 I'm a little bit confused but will dig into it later!

yihau:
oh. I think I reproduced the issue successfully. we don't need to copy `[workspace.package]` and `[workspace.dependencies]` if  we can do either
1) clone twoxtx-solana outside solana-program-library folder
2) add `exclude` to root Cargo.toml for twoxtx-solana

joncinque:
If you just run the commands from the twoxtx build from the root of SPL, it'll show up, ie:

```
          ./token/twoxtx-setup.sh
          ./token/twoxtx-solana/cargo-build-sbf --manifest-path token/program-2022/Cargo.toml
```

joncinque:
> oh. I think I reproduced the issue successfully. we don't need to copy [workspace.package] and [workspace.dependencies] if we can do either ...
 
> clone twoxtx-solana outside solana-program-library folder

Ah very cool! This does the trick

> add exclude to root Cargo.toml for twoxtx-solana

Sorry, I don't understand, how would this look exactly?

yihau:
oh! I mean we can add a exclude rule in the Cargo.toml. it looks like 

```diff
[workspace]
members = [
  "associated-token-account/program",
  "associated-token-account/program-test",
  "binary-option/program",
  (omit ... )
]

+exclude = [
+  "<solana-program-library folder>/token/twoxtx-solana",
+]

[profile.dev]
split-debuginfo = "unpacked"

```

joncinque:
That's so much better :laughing: thanks so much! Closing this in favor of #4125

PR #4122, opened 2023-04-14
token-cli: Always use supplied blockhash for nonce txs

joncinque:
#### Problem

The supplied `--blockhash` arg is ignored when online and finally sending the transaction, making it impossible to work with nonces.

#### Solution

Plumb the nonce blockhash down and use it.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -2352,11 +2352,17 @@ impl offline::ArgsConfig for SignOnlyNeedsFullMintSpec {
     fn sign_only_arg<'a, 'b>(&self, arg: Arg<'a, 'b>) -> Arg<'a, 'b> {
         arg.requires_all(&[MINT_ADDRESS_ARG.name, MINT_DECIMALS_ARG.name])
     }
+    fn signer_arg<'a, 'b>(&self, arg: Arg<'a, 'b>) -> Arg<'a, 'b> {
+        arg.requires_all(&[MINT_ADDRESS_ARG.name, MINT_DECIMALS_ARG.name])
+    }
 }
 
 struct SignOnlyNeedsMintDecimals {}
 impl offline::ArgsConfig for SignOnlyNeedsMintDecimals {
     fn sign_only_arg<'a, 'b>(&self, arg: Arg<'a, 'b>) -> Arg<'a, 'b> {
+        arg.requires_all(&[MINT_DECIMALS_ARG.name])
+    }
+    fn signer_arg<'a, 'b>(&self, arg: Arg<'a, 'b>) -> Arg<'a, 'b> {
         arg.requires_all(&[MINT_DECIMALS_ARG.name])
     }
 }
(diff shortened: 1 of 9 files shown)


token/client/src/token.rs
```diff
@@ -430,11 +440,12 @@ where
 
         instructions.extend_from_slice(token_instructions);
 
-        let latest_blockhash = self
-            .client
-            .get_latest_blockhash()
-            .await
-            .map_err(TokenError::Client)?;
+        let latest_blockhash = self.nonce_blockhash.unwrap_or(
```
t-nelson, line 443: could make this change in the if-let below, so we consistently do the option 3-tuple thing everywhere, gross as it is :wink:

joncinque: Yeah sure, I didn't want to duplicate all of the other places where it's used below, but you're right that it would be clearer

token/cli/src/main.rs
```diff
@@ -193,8 +193,6 @@ pub fn mint_decimals_arg<'a, 'b>() -> Arg<'a, 'b> {
         .takes_value(true)
         .value_name("MINT_DECIMALS")
         .validator(is_mint_decimals)
-        .requires(SIGN_ONLY_ARG.name)
```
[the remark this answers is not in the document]
joncinque: This is currently done through `offline::ArgsConfig` for the appropriate commands. This will relax the restriction for online mode though, so that you can still specify the decimals and all that when finally sending the transaction, which makes for a nicer UX if you ask me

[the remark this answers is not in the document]
t-nelson: yeah was mainly hoping to disallow doing the final sign+submit invocation of an offline transaction without the identical cli args used for the sign-only invocation(s)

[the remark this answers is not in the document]
joncinque: That's what we're doing here by removing the sign only requirement from these other args, so that submission can keep using those args too

token/cli/src/main.rs
```diff
@@ -2346,27 +2352,39 @@ impl offline::ArgsConfig for SignOnlyNeedsFullMintSpec {
     fn sign_only_arg<'a, 'b>(&self, arg: Arg<'a, 'b>) -> Arg<'a, 'b> {
         arg.requires_all(&[MINT_ADDRESS_ARG.name, MINT_DECIMALS_ARG.name])
     }
+    fn signer_arg<'a, 'b>(&self, arg: Arg<'a, 'b>) -> Arg<'a, 'b> {
+        arg.requires_all(&[MINT_ADDRESS_ARG.name, MINT_DECIMALS_ARG.name])
```
t-nelson, line 2356: iirc, `requires_all()` discards existing, rather than appending.  so i think we need to explicitly redeclare `BLOCKHASH_ARG.name` here

t-nelson: oh n/m i'm full of shit, as usual.

https://docs.rs/clap/2.34.0/src/clap/args/arg.rs.html#1784-1787

joncinque: Looks like it adds according to the impl: https://docs.rs/clap/2.34.0/src/clap/args/arg.rs.html#1783

t-nelson (approved):
:shipit:

joncinque:
The failing checks are unrelated -- merging on red!

PR #4123, opened 2023-04-16
[compression] expose macros for invoking getter/setters on const generic CMT

ngundotra:
bumps crate version to 0.1.9

Previously exported macros were unusable bc inner macros were private. Now fixed.


joncinque (approved):
Interesting, I didn't know that, but it makes sense. Thanks!

PR #4125, opened 2023-04-17
ci: Exclude solana dir during patch to avoid workspace issues

joncinque:
#### Problem

The token twoxtx build is failing because of workspace dependencies in the monorepo.

#### Solution

Exclude the local solana directory when patching crates to avoid two conflicting workspaces. This is a much better version of #4121 thanks to @yihau


yihau (approved):
🪖

PR #4126, opened 2023-04-17
token-cli: error handling fixes

jbiseda:
#### Problem
In several places in the cli `Result`s are unwrapped without error handling.

#### Summary of changes
Check `Result`s for `Err`.

### DIFF
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -307,7 +307,8 @@ impl<'a> Config<'a> {
         override_name: &str,
         wallet_manager: &mut Option<Arc<RemoteWalletManager>>,
     ) -> Result<Pubkey, Error> {
-        let token = pubkey_of_signer(arg_matches, "token", wallet_manager).unwrap();
+        let token = pubkey_of_signer(arg_matches, "token", wallet_manager)
+            .map_err(|e| -> Error { e.to_string().into() })?;
         self.associated_token_address_for_token_or_override(
             arg_matches,
             override_name,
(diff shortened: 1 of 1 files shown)


token/cli/src/config.rs
```diff
@@ -355,8 +362,11 @@ impl<'a> Config<'a> {
         address_name: &str,
         wallet_manager: &mut Option<Arc<RemoteWalletManager>>,
     ) -> Result<Pubkey, Error> {
-        if let Some(address) = pubkey_of_signer(arg_matches, address_name, wallet_manager).unwrap()
-        {
+        let address = match pubkey_of_signer(arg_matches, address_name, wallet_manager) {
+            Ok(address) => address,
+            Err(e) => return Err(e.to_string().into()),
+        };
```
joncinque, line 368: nit: how about bubbling this in one line by mapping the error?
```suggestion
        let address = pubkey_of_signer(arg_matches, address_name, wallet_manager).map_err(|e| e.to_string().into())?;
```

token/cli/src/config.rs
```diff
@@ -326,13 +330,16 @@ impl<'a> Config<'a> {
         wallet_manager: &mut Option<Arc<RemoteWalletManager>>,
         token: Option<Pubkey>,
     ) -> Result<Pubkey, Error> {
-        if let Some(address) = pubkey_of_signer(arg_matches, override_name, wallet_manager).unwrap()
-        {
+        let address = match pubkey_of_signer(arg_matches, override_name, wallet_manager) {
+            Ok(address) => address,
+            Err(e) => return Err(e.to_string().into()),
+        };
```
joncinque, line 336: nit: how about mapping the error?
```suggestion
        let address = pubkey_of_signer(arg_matches, override_name, wallet_manager).map_err(|e| e.to_string().into())?;
```

token/cli/src/config.rs
```diff
@@ -307,14 +307,18 @@ impl<'a> Config<'a> {
...
     ) -> Result<Pubkey, Error> {
-        let token = pubkey_of_signer(arg_matches, "token", wallet_manager).unwrap();
-        self.associated_token_address_for_token_or_override(
-            arg_matches,
-            override_name,
-            wallet_manager,
-            token,
-        )
-        .await
+        match pubkey_of_signer(arg_matches, "token", wallet_manager) {
```
joncinque, line 310: nit: we can keep this roughly the same by mapping the error, ie
```suggestion
        let token = pubkey_of_signer(arg_matches, "token", wallet_manager).map_err(|e| e.to_string().into())?;
```

joncinque (commented):
Thanks for noticing these and contributing some fixes! Just a nit to map the error rather than matching

token/cli/src/config.rs
jbiseda: good suggestion, updated all instances with `map_err`.

jbiseda:
> Thanks for noticing these and contributing some fixes! Just a nit to map the error rather than matching

no problem, just happened to stumble across this

joncinque (approved):
The twoxtx job is getting fixed in #4125 so feel free to merge once CI passes

PR #4127, opened 2023-04-17
docs: Add token-2022 one-pager

joncinque:
#### Problem

Although the GitHub project information lives at https://github.com/orgs/solana-labs/projects/34/views/1, that's not really consumable by a general audience.

#### Solution

Add a token-2022 one-pager. While I was at it, I added the NCC audit, hope that's ok!

### DIFF
--- a/docs/src/token-2022/status.md
+++ b/docs/src/token-2022/status.md
@@ -1,0 +1,69 @@
+---
+title: Project Status
+---
+
+The Token-2022 program is still under audit and not meant for full production use.
+In the meantime, all clusters have the latest program deployed **for testing and
+development purposes ONLY**.
+
+## Timeline
+
+Here is the general program timeline and rough ETAs:
+
+| Issue | ETA |
+| --- | --- |
+| Finish program | end of May |
+| Final audit | end of June |
+| Mainnet recommendation | middle of July |
+| Freeze program | Q1 2024 |
+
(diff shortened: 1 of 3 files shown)


docs/src/token-2022/status.md
```diff
@@ -0,0 +1,69 @@
...
+| Finish program | end of May |
+| Final audit | end of June |
+| Mainnet recommendation | middle of July |
+| Freeze program | Q1 2024 |
+
+More information: https://github.com/orgs/solana-labs/projects/34
+
+## Remaining items
+
+### Zero-knowledge proof split
```
mvines, line 24: The zk support also requires the cluster be running 1.14 + enabled feature gates, worth calling that out in this section too?

joncinque: I had it and then removed it to avoid putting too much, but since you've called it out, I think that means it's needed :smile: any ETA for 1.14 that I can add?

mvines: For this doc I think it's best to just call out the dependency on 1.14 without a specific eta, easier to maintain going forward.

but for an 1.14 ETA, soon? Would love to see mainnet on 1.14 by the end of May. There's no real blocker anymore, just perception issues

joncinque: Sounds good, done!

mvines (approved):

PR #4128, opened 2023-04-18
fix: only add exclude rule when solana root is under spl root

yihau:
I forgot we also use `patch.crates-io.sh` in downstream project test. We should add the exclude rule only when the solana root is under spl root I think 😢

### DIFF
--- a/patch.crates-io.sh
+++ b/patch.crates-io.sh
@@ -22,6 +22,7 @@ set -e
 
 solana_dir=$(cd "$solana_dir" && pwd)
 cd "$(dirname "$0")"
+project_root=$(pwd)
 
 source "$solana_dir"/scripts/read-cargo-variable.sh
 
@@ -32,8 +33,7 @@ if [[ -f "$toolchain_file" ]]; then
 fi
 
 # only add exclude rule when solana root is under spl root
-spl_project_root=$(git rev-parse --show-toplevel)
-if echo "$solana_dir" | grep "^$spl_project_root" > /dev/null; then
+if echo "$solana_dir" | grep "^$project_root" > /dev/null; then
   echo "Excluding $solana_dir from workspace"
   echo
   for crate in "${workspace_crates[@]}"; do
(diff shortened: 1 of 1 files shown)


patch.crates-io.sh
```diff
@@ -31,17 +31,21 @@ if [[ -f "$toolchain_file" ]]; then
...
 
-echo "Excluding $solana_dir from workspace"
-echo
-for crate in "${workspace_crates[@]}"; do
-  if grep -q "exclude.*$solana_dir" "$crate"; then
-    echo "$crate is already patched"
-  else
-    cat >> $crate << EXCLUDE
+# only add exclude rule when solana root is under spl root
+spl_project_root=$(git rev-parse --show-toplevel)
```
joncinque, line 35: We're using `$(dirname "$0")` higher up, why not just store that directly and grep against that?
```suggestion
project_root=$(dirname "$0")
```
This way, this script can work with other repos, since it's not meant to be spl specific

joncinque (commented):
Thanks for fixing this!

patch.crates-io.sh
yihau: oh! cuz my idea is to compare two full path to know whether solana root is under spl root. if I use`$(dirname "$0")`, I think it will only give me `.` and I couldn't use the same logic 🤔

joncinque: Ah sorry, I meant `pwd` after we cd into `$(dirname "$0")`

yihau: ah. sorry. I think I have a misunderstand with the situation. I think we can use the pwd!

yihau: fixed it in [d02e563](https://github.com/solana-labs/solana-program-library/pull/4128/commits/d02e563431b597929f974f72fecb88a75971c081)

joncinque (approved):
Yes exactly, thanks! I was right about to put up the same exact PR :sweat_smile:

yihau:
yeah! sorry for the misleading 😢

joncinque:
> yeah! sorry for the misleading

No worries, I was misleading too

PR #4134, opened 2023-04-20
docs: Add permanent delegate extension and example

joncinque:
#### Problem

The permanent delegate extension exists and is supported in the CLI and JS lib, but it's not in the docs.

#### Solution

Add the docs and examples.

### DIFF
--- a/docs/src/token-2022/extensions.mdx
+++ b/docs/src/token-2022/extensions.mdx
@@ -1202,3 +1202,96 @@ Signature: 5DQs6hzkfGq3uotESuVwF7MGeMawwfQcm1e9RHaUeVySDV6xpUzYhzdb6ygqJfsEZqewg
 
   </TabItem>
 </Tabs>
+
+### Permanent Delegate
+
+With Token-2022, it's possible to specify a permanent account delegate for a
+mint. This authority has unlimited delegate privileges over any account for that
+mint, meaning that it can burn or transfer any amount of tokens.
+
+While this feature certainly has room for abuse, it has many important real-world
+use cases.
+
+In some jurisdictions, a stablecoin issuer must be able to seize assets from
+sanctioned entities. Through the permanent delegate, the stablecoin issuer can
+transfer or burn tokens from accounts owned by sanctioned entities.
+
+It's also possible to implement a [Harberger Tax](http://www.harbergertax.com/)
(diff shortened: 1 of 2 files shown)


docs/src/token-2022/extensions.mdx
```diff
@@ -1202,3 +1202,96 @@ Signature: 5DQs6hzkfGq3uotESuVwF7MGeMawwfQcm1e9RHaUeVySDV6xpUzYhzdb6ygqJfsEZqewg
...
+token. After a sale, the permanent delegate can move the NFT from the owner to
+the buyer if the previous owner doesn't pay the tax.
+
+#### Example: Create a mint with a permanent delegate
+
+<Tabs className="unique-tabs" groupId="language-selection">
+  <TabItem value="cli" label="CLI" default>
+
+```console
+$ spl-token --program-id TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb create-token --enable-permanent-delegate
```
mvines, line 1230: What authority is set as a the permanent delegate in this case?

joncinque: it defaults to the mint authority, and can be changed with an `authorize` call. I've added another example command -- what do you think?

mvines: In the JS case, the auth is explicitly set upon creation with:
```js
  createInitializePermanentDelegateInstruction(mint, permanentDelegate.publicKey, TOKEN_2022_PROGRAM_ID),
```

Does `spl-token create-token` not expose that ability? The extra `authorize` example is great, seems like ideally the cli and JS token creation examples would behave identically though for easy of understanding by a new reader

joncinque: The new args were modeled off the `--enable-freeze` CLI arg, which always uses the mint authority. We can change all the similar CLI args, ie. `--enable-close` and `--enable-permanent-delegate` to take a pubkey authority, but that would make them inconsistent with `--enable-freeze`.

mvines: ok, np. seems ok, I was just asking,

mvines (approved):

joncinque:
Docs and token-js tests passed, which were the only affected parts in this PR, so merging this on red

PR #4138, opened 2023-04-25
single pool: fix create_and_delegate_user_stake

2501babe:
note this does not impact any (possible? not clear if one is ongoing) audits since this is a bug just in the helper function, not the program


joncinque (approved):
Looks good!

PR #4142, opened 2023-04-26
[compression] removed direct export of esm module

nickfrosty:
Removed the direct export of the ESM module since common js can be imported in esm.

Also removed the extraneous `tsc` command in the JS `build` script

Related: https://github.com/metaplex-foundation/js/pull/480#issuecomment-1520127040


ngundotra:
@nickfrosty can u bump package version? LGTM 👍

nickfrosty:
@ngundotra versioned bumped!

PR #4146, opened 2023-04-28
tlv-account-resolution: Add state interface library for additional accounts

joncinque:
#### Problem

Following on the work from #4137, we want a permissioned-transfer extension in token-2022 done through an interface.

#### Solution

Splitting off some work from #4105, this adds a library defining a generic "state interface" for writing the additional required accounts for an instruction.

cc @austbot


libraries/account-resolution/README.md
```diff
@@ -0,0 +1,130 @@
...
+For example, a `Transferable` interface, along with a `transfer` instruction,
+also requires implementations to expose a `get_additional_accounts_for_transfer`
+instruction.
+
+In the program implementation, this instruction writes the additional accounts
+into return data, making it easy for on-chain and off-chain clients to consume.
+
+See this
+[relevant sRFC](https://forum.solana.com/t/srfc-00010-additional-accounts-request-transfer-spec/122)
+for more information.
```
ngundotra, line 86: I think this should be retitled `Comparison with Other methods` or something, since `get_additional_accounts_for_transfer` is not directly related. Unless I'm just missing something

libraries/account-resolution/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+This can be confusing. Typically, a type implements `TlvDiscriminator`, so that
+the type can be written into TLV data. In this case, `ExtraAccountMetas` is
+generic over `TlvDiscriminator`, meaning that a program can write many different instances of
+`ExtraAccountMetas` into one account, using different `TlvDiscriminator`s.
+
+Also, it's reusing an instruction discriminator as a TLV discriminator. For example,
+if the `transfer` instruction has a discriminator of `[1, 2, 3, 4, 5, 6, 7, 8]`,
+then the account uses a TLV discriminator of `[1, 2, 3, 4, 5, 6, 7, 8]` to denote
+where the additional account metas are stored.
```
ngundotra, line 127: This is really cool. Looking forward to seeing examples of managing this state on-chain.

ngundotra (changes requested):
Missing from this spec / library is an indication of how the `account-resolution` library is meant to work. ie - is there a specific PDA / derivation path for finding the account that has this instruction data?

Also, since this is mostly a TLV library + account info helpers, maybe a better name is `tlv-account-resolution`? Technically another spec could come and be `borsh-account-resolution`, which seems like the only difference would be that no other types could be defined alongside the AccountMetas.

Otherwise this is sick, clean code! Love the zero-copy stuff

libraries/account-resolution/src/state.rs
```diff
@@ -0,0 +1,447 @@
...
+            .saturating_add(PodSlice::<PodAccountMeta>::size_of(num_items)?))
+    }
+
+    /// Add the additional account metas to an existing instruction
+    pub fn add_to_instruction<T: TlvDiscriminator>(
+        instruction: &mut Instruction,
+        data: &[u8],
+    ) -> Result<(), ProgramError> {
+        let state = TlvStateBorrowed::unpack(data)?;
+        let bytes = state.get_bytes::<T>()?;
```
buffalojoec, line 126: Clean, so clean

buffalojoec (approved):

libraries/account-resolution/README.md
joncinque: Yeah sure that's fine. I wanted to point out that there's really two main ways of resolving accounts that we all came up with -- either off-chain by reading account data, or on-chain via simulation. `get_additional_accounts_for_transfer` is meant to be part of the example of how you would do it using simulation. Either way, I'll move it to a separate section.

joncinque:
> Also, since this is mostly a TLV library + account info helpers, maybe a better name is tlv-account-resolution? 

Yeah for sure, maybe even clearer as just `spl-tlv-extra-accounts`? It does some account resolution, but maybe it's unclear since it's just storing state, what do you think?

> Technically another spec could come and be borsh-account-resolution, which seems like the only difference would be that no other types could be defined alongside the AccountMetas.

Actually, the data format here is exactly the same as a borsh vector! So this already has borsh support by design, just doesn't explicitly use it

joncinque:
> I think the idea is to decouple the library from the on-chain implementation (?). So, to me it makes sense for that derivation to live within the interface code.

Yeah that's exactly it. It's meant to be a *generic* state interface, to be used by *other* interfaces

libraries/account-resolution/src/state.rs
```diff
@@ -0,0 +1,447 @@
...
+///     &remaining_account_infos,
+/// );
+/// ```
+pub struct ExtraAccountMetas;
+impl ExtraAccountMetas {
+    /// Initialize pod slice data for the given instruction and account infos
+    /// into the given buffer
+    pub fn init<'a, T: TlvDiscriminator, M: 'a>(
+        data: &mut [u8],
+        account_infos: &'a [M],
```
austbot, line 81: maybe change param name to something general

libraries/account-resolution/src/state.rs
austbot: its clean and ill put this feedback on the tlv but i think we can auto convert it to the correct type see https://gist.github.com/austbot/40fabf36a3e7bcef8f755900157d54ff#file-current-rs-L154

this auto type conversion may be a good optional feature to simplify the tlv for other devs, not needed in this case probably

austbot: https://gist.github.com/austbot/40fabf36a3e7bcef8f755900157d54ff#file-current-rs-L353

austbot (commented):
this is great building off the other work

libraries/account-resolution/src/state.rs
joncinque: Yeah sure, do you have a suggestion for a name? As you noticed, I couldn't come up with anything better :sweat_smile: maybe `account_metas`? or `into_account_metas`? or `source_account_metas`?

libraries/account-resolution/src/state.rs
```diff
@@ -0,0 +1,447 @@
...
+        account_infos: &[AccountInfo<'a>],
+    ) -> Result<(), ProgramError> {
+        let state = TlvStateBorrowed::unpack(data)?;
+        let bytes = state.get_bytes::<T>()?;
+        let extra_account_metas = PodSlice::<PodAccountMeta>::unpack(bytes)?;
+        cpi_instruction
+            .accounts
+            .extend(extra_account_metas.data().iter().map(Into::into));
+
+        for account_meta in extra_account_metas.data().iter().map(AccountMeta::from) {
```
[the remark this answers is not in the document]
joncinque: It's a good point... I was aiming to avoid the additional allocation required for the map. Maybe we could eventually expose another version that uses a map internally? That way you can go with "allocation-free-but-maybe-loads-of-compute-units" or "allocs-but-fewer-compute-units". What do you think?

libraries/account-resolution/src/state.rs
joncinque: In this case we can't do it automatically unfortunately because the `PodSlice` is unsized. I spent awhile trying to come up with a slick way to do zero-copy serde for unsized types, but the lifetimes are really tricky to manage. In the end, TLV v1 keeps things super simple, but if we come up with a way to make it work in the TLV library, I'm super down to integrate it!

libraries/account-resolution/src/state.rs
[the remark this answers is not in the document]
austbot: you could store in a diff type than account meta that doest duplicate the pubkey an then consume it with into orsomething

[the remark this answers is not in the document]
joncinque: Oh of course, we can get this down to one conversion by pushing into the vector, ie:

```
        for account_meta in extra_account_metas.data().iter().map(AccountMeta::from) {
            let account_info = account_infos
                .iter()
                .find(|&x| *x.key == account_meta.pubkey)
                .ok_or(AccountResolutionError::IncorrectAccount)?
                .clone();
            cpi_account_infos.push(account_info);
            cpi_instruction.accounts.push(account_meta);
        }
```

Thanks!

joncinque:
Merging this in since there hasn't been any additional feedback. Any additional post-merge reviews are always appreciated <3

PR #4147, opened 2023-04-28
transfer-hook-interface: Add interface defining a hook called during token-2022 transfer

joncinque:
#### Problem

We want to add a hook callable by token-2022 during transfers.

#### Solution

Add a program interface and sample program to do that! This builds on #4146, so you only need to look at the last commit. There were a couple more required functions in the account resolution library, so I've added those with this PR.

As far as interfaces are concerned, this is the last step! Once this interface is defined, token-2022 just needs to call into it if a particular extension is configured.

Note: this has been called "permissioned-transfer" for awhile, but looking at it wholly, it really allows for any sort of program to be called. This could even be used to assess royalties if accounts are configured to be writable! Token-2022 will just need to downgrade signer from any accounts already specified in the transfer to be safe.

cc @austbot (for some reason, you don't appear on the autocomplete list for reviewers just yet)

Part of #4105


token/transfer-hook-interface/src/instruction.rs
```diff
@@ -0,0 +1,206 @@
...
+    ///   1. `[]` Token mint
+    ///   2. `[]` Destination account
+    ///   3. `[]` Source account's owner/delegate
+    ///   4. `[]` Validation account
+    ///   5..5+M `[]` `M` additional accounts, written in validation account data
+    ///
+    Execute {
+        /// Amount of tokens to transfer
+        amount: u64,
+    },
```
ngundotra, line 32: Having an interface that only gets access to token accounts seems weird, since we may want to ~steal~ charge royalties in sol against the owner wallet in the hook program. That feels like a pretty natural usage

token/transfer-hook-interface/src/instruction.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+/// TLV instruction type only used to define the discriminator. The actual data
+/// is entirely managed by `ExtraAccountMetas`, and it is the only data contained
+/// by this type.
+pub struct ExecuteInstruction;
+impl TlvDiscriminator for ExecuteInstruction {
+    const TLV_DISCRIMINATOR: Discriminator = Discriminator::new(EXECUTE_DISCRIMINATOR);
+}
+/// First 8 bytes of `hash::hashv(&["spl-transfer-hook-interface:execute"])`
+const EXECUTE_DISCRIMINATOR: [u8; Discriminator::LENGTH] = [105, 37, 101, 197, 75, 251, 102, 26];
```
ngundotra, line 54: 🔥

token/transfer-hook-interface/src/instruction.rs
```diff
@@ -0,0 +1,206 @@
...
+// annoying, but needed to perform a match on the value
+const EXECUTE_DISCRIMINATOR_SLICE: &[u8] = &EXECUTE_DISCRIMINATOR;
+/// First 8 bytes of `hash::hashv(&["spl-transfer-hook-interface:initialize-extra-account-metas"])`
+const INITIALIZE_EXTRA_ACCOUNT_METAS_DISCRIMINATOR: &[u8] = &[43, 34, 13, 49, 167, 88, 235, 235];
```
ngundotra, line 58: I doubled checked both of these hashes here: https://emn178.github.io/online-tools/sha256.html

token/transfer-hook-interface/src/instruction.rs
```diff
@@ -0,0 +1,206 @@
...
+        validate_state_pubkey,
+        amount,
+    );
+    instruction.accounts.extend_from_slice(additional_accounts);
+    instruction
+}
+
+/// Creates an `Execute` instruction, without the additional accounts
+#[allow(clippy::too_many_arguments)]
+pub fn execute(
```
ngundotra, line 125: Can this be renamed to `create_execute_instruction`? `execute` sounds like its making the `invoke` call

token/transfer-hook-interface/src/inline_spl_token.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Extract the mint authority from the account bytes
+pub fn get_mint_authority(account_data: &[u8]) -> Result<COption<Pubkey>, ProgramError> {
+    const MINT_SIZE: usize = 82;
+    if account_data.len() < MINT_SIZE {
+        Err(ProgramError::InvalidAccountData)
+    } else {
+        let mint_authority = array_ref![account_data, 0, 36];
+        unpack_coption_key(mint_authority)
+    }
+}
```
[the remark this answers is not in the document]
ngundotra: oh i see, the SPL pattern is `instruction::<ix-name>`. This is fine

token/transfer-hook-interface/src/instruction.rs
```diff
@@ -0,0 +1,206 @@
...
+#[cfg(test)]
+mod test {
+    use {super::*, crate::NAMESPACE, solana_program::hash};
+
+    #[test]
+    fn validate_packing() {
+        let amount = 111_111_111;
+        let check = TransferHookInstruction::Execute { amount };
+        let packed = check.pack();
+        let preimage = hash::hashv(&[format!("{NAMESPACE}:execute").as_bytes()]);
```
ngundotra, line 183: nit: Can you add a comment above this line that says "Please use EXECUTE_DISCRIMINATOR in your program, the following is just for test purposes"

I learned how to write programs on Solana using SPL Token as a reference, so I want to make sure that for fellow travelers, there are some helpful crumbs for this new SPL ix data pattern

token/transfer-hook-interface/src/instruction.rs
```diff
@@ -0,0 +1,206 @@
...
+        let unpacked = TransferHookInstruction::unpack(&expect).unwrap();
+        assert_eq!(unpacked, check);
+    }
+
+    #[test]
+    fn initialize_validation_pubkeys_packing() {
+        let check = TransferHookInstruction::InitializeExtraAccountMetas;
+        let packed = check.pack();
+        let preimage =
+            hash::hashv(&[format!("{NAMESPACE}:initialize-extra-account-metas").as_bytes()]);
```
ngundotra, line 198: nit: Ideally same here as with the previous test 🙏

token/transfer-hook-interface/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+// Export current sdk types for downstream users building with a different sdk version
+pub use solana_program;
+use solana_program::pubkey::Pubkey;
+
+/// Namespace for all programs implementing transfer-hook
+pub const NAMESPACE: &str = "spl-transfer-hook-interface";
+
+/// Size for discriminator in account and instruction data
+pub const DISCRIMINATOR_LENGTH: usize = 8;
```
ngundotra, line 29: Seeing this makes me feel uncomfortable with how TLV Discriminators work, since it seems like a footgun to have to always define the size of discriminators. But for now, this is sick!

token/transfer-hook-interface/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+pub(crate) fn collect_extra_account_metas_signer_seeds<'a>(
+    mint: &'a Pubkey,
+    bump_seed: &'a [u8],
+) -> [&'a [u8]; 3] {
+    [EXTRA_ACCOUNT_METAS_SEED, mint.as_ref(), bump_seed]
+}
+
+solana_program::declare_id!("pERmRFhRmg9JaJdsocrUnLLigHXrwWTxBu2SafwK2cd");
```
ngundotra, line 57: Does this need a `#[cfg(not(feature="no-entrypoint"))]`?

token/transfer-hook-interface/src/offchain.rs
```diff
@@ -0,0 +1,64 @@
...
+/// ```rust,ignore
+/// use solana_client::nonblocking::rpc_client::RpcClient;
+/// use solana_sdk::pubkey::Pubkey;
+///
+/// let program_id = Pubkey::new_unique();
+/// let mint = Pubkey::new_unique();
+/// let client = RpcClient::new_mock("succeeds".to_string());
+///
+/// let extra_account_metas = get_extra_account_metas(
+///     |address| self.client.get_account(&address),
```
ngundotra, line 29: The point is that off-chain clients can infer their own PDAs, right?

ngundotra: In a generic library, this makes sense. But in this library, I think this is too general, and a bit confusing. 
 
I think here it should only take `mint`.

IMO it can be spun out onto a library ontop of tlv in new lib like `tlv-account-resolution-keyed` ¯\_(ツ)_/¯ just spitballing here.

token/transfer-hook-interface/README.md
```diff
@@ -0,0 +1,106 @@
+## Transfer-Hook Interface
+
+### Example program
+
+Here is an example program that only implements the required "execute" instruction,
+assuming that the proper account data is already written to the appropriate 
+program-derived address defined by the interface.
```
ngundotra, line 7: If this is just an example program, shouldn't this be in a separate repo under the `solana-developers/` org?

token/transfer-hook-interface/src/instruction.rs
ngundotra: Wait shouldn't this discriminator & the init discriminator be put into the `spl-token-2022` library?

ngundotra (dismissed):
Code looks great.

1. I think this may need to be rebased on top of the TLV PR, since currently it shows all the tlv-account-resolution stuff being added again.

2. It looks like `transfer-hook-interface` is an example program defining the spec that programs need to adhere to for token22 extension. If this is the case, then the IX discriminators & `namespace` constants should be exported under `spl-token-2022` library, and imported in the example program. Also, maybe it should be in a separate org, since it's an example program.

token/transfer-hook-interface/src/instruction.rs
joncinque: The interface states it gets the transfer accounts PLUS all of the additional accounts specified, so it doesn't just get the transfer accounts. That's what the `    ///   5..5+M `[]` `M` additional accounts, written in validation account data` line is about. Maybe it needs to be clearer?

token/transfer-hook-interface/src/instruction.rs
joncinque: it's not particularly consistent since none of the Rust crates follow the `create_<ix_name>_instruction`. Maybe we should rename `execute` to `perform_transfer_hook` or something?

token/transfer-hook-interface/src/instruction.rs
joncinque: Sure, if that helps. Maybe a comment above the declaration `EXECUTE_DISCRIMINATOR` would be helpful too then?

token/transfer-hook-interface/src/lib.rs
joncinque: ah good call, I should just remove it entirely

token/transfer-hook-interface/src/offchain.rs
joncinque: The point of this is to decouple it from any particular client, so it works with rpc-client and banks-client, which is how the token-client works. So at its core this only takes the mint and program id, which are the only required bits to fetch the info. In the implementation, it does the PDA derivation itself. It felt simpler to keep it in this library with a feature flag rather than spinning out a separate crate just for one file.

token/transfer-hook-interface/README.md
joncinque: It defines the interface as well, which is used directly in token-2022. The example program is really just to help with integration testing, but this could be the right time to define how interfaces are declared here. How do you imagine packaging an interface? Does it just provide the state + instructions? Or does it also come with an example program for testing?

joncinque:
> I think this may need to be rebased on top of the TLV PR, since currently it shows all the tlv-account-resolution stuff being added again.

Yeah, I pointed this out in the PR description: "This builds on https://github.com/solana-labs/solana-program-library/pull/4146, so you only need to look at the last commit." once #4146 lands I'll definitely rebase!

> It looks like transfer-hook-interface is an example program defining the spec that programs need to adhere to for token22 extension. If this is the case, then the IX discriminators & namespace constants should be exported under spl-token-2022 library, and imported in the example program. Also, maybe it should be in a separate org, since it's an example program.

So you want to define the interface within token-2022, ie. declaring a separate interface within a program? And then host an example implementation in a separate organization? Sorry, that seems really confusing! It seems more natural to me to have a program (with its crate) calling another program whose code lives in its own crate, but maybe I'm thinking about this wrong in the context of interfaces.

token/transfer-hook-interface/src/instruction.rs
joncinque: Did both

token/transfer-hook-interface/src/inline_spl_token.rs
```diff
@@ -0,0 +1,27 @@
...
+    match *tag {
+        [0, 0, 0, 0] => Ok(COption::None),
+        [1, 0, 0, 0] => Ok(COption::Some(Pubkey::new_from_array(*body))),
+        _ => Err(ProgramError::InvalidAccountData),
+    }
+}
+
+/// Extract the mint authority from the account bytes
+pub fn get_mint_authority(account_data: &[u8]) -> Result<COption<Pubkey>, ProgramError> {
+    const MINT_SIZE: usize = 82;
```
buffalojoec, line 20: Why not `use spl_token_2022::state::Mint::LEN as MINT_SIZE` ?

buffalojoec: Assuming the circular dep thing?

token/transfer-hook-interface/src/instruction.rs
buffalojoec: This might not be related but the way you've written the comments reminds me of the nomenclature for passing multisig accounts to a token instruction, ie:
```rust
    ///
    ///   * Single owner/delegate
    ///   0. `[writable]` The source account.
    ///   1. `[writable]` The destination account.
    ///   2. `[signer]` The source account's owner/delegate.
    ///
    ///   * Multisignature owner/delegate
    ///   0. `[writable]` The source account.
    ///   1. `[writable]` The destination account.
    ///   2. `[]` The source account's multisignature owner/delegate.
    ///   3. ..3+M `[signer]` M signer accounts.
    Transfer {
        /// The amount of tokens to transfer.
        amount: u64,
    },
```

Have we considered how both of these variably-sized additional accounts might work together?

buffalojoec: Starting to think they would just be included, but would have to be signers, no?

joncinque: The idea is to keep it simple for multisigs, and just provide the authority *without* signers. The token program won't provide any existing of the token accounts as signer or writable to be safe

token/transfer-hook-interface/src/inline_spl_token.rs
joncinque: Yeah exactly

token/transfer-hook-interface/src/instruction.rs
buffalojoec: Maybe instead rename the CPI helper from `execute<'a>` to `invoke_hook<'a>` or `hook_invoke<'a>`?

buffalojoec: Agreed with Noah though it's a bit confusing since the functions return `Instruction`, so might I suggest `build_execute_instruction` or something like that?

buffalojoec: Or `create_` hah (Noah's comment)

buffalojoec:
> > It looks like transfer-hook-interface is an example program defining the spec that programs need to adhere to for token22 extension. If this is the case, then the IX discriminators & namespace constants should be exported under spl-token-2022 library, and imported in the example program. Also, maybe it should be in a separate org, since it's an example program.
> 
> So you want to define the interface within token-2022, ie. declaring a separate interface within a program? And then host an example implementation in a separate organization? Sorry, that seems really confusing! It seems more natural to me to have a program (with its crate) calling another program whose code lives in its own crate, but maybe I'm thinking about this wrong in the context of interfaces.

I think the idea of having a closely-tied example implementation is cool, but it could get confusing if you're reading the Rust docs on this and wondering what in the world is the `example_processor` module.

If you want it all in the same source on GitHub maybe feature-gate the example code from the crate? Or split off an examples folder?

To Noah's point an example program does feel like docs, cookbook reference, or [program example](https://github.com/solana-developers/program-examples).

token/transfer-hook-interface/src/instruction.rs
ngundotra: can they be separate? that's kind of confusing

token/transfer-hook-interface/src/instruction.rs
ngundotra: Thank you!

token/transfer-hook-interface/src/offchain.rs
ngundotra: Checks out, thanks!

token/transfer-hook-interface/README.md
ngundotra: Yeah I see what you're saying. Maybe just move the entrypoint & processor to `tests/` folder so its more clear.

What you're saying about testing makes sense. It's confusing because the lib is intertwined with the test program.

ngundotra (commented):
Can the `example_*` items be moved to `tests/`? Otherwise this looks ready to ship IMO

joncinque:
> Can the example_* items be moved to tests/? Otherwise this looks ready to ship IMO

If the processor is in `tests`, then it isn't available to anyone else who wants to use the interface in their program and test calling into something.

> If you want it all in the same source on GitHub maybe feature-gate the example code from the crate? Or split off an examples folder?

Feature-gating seems kinda weird too, since then you'll have to import the crate once in your program with no features, and then a second time in your `dev-dependencies` with the feature on.

Could do an `examples` folder, which would be something like `src/examples/processor.rs`? Unless it makes more sense to have a separate crate with just the example program, ie `spl-transfer-hook-example`? I'm happy to bikeshed this since I want to do the right thing here as a model for other interfaces.

token/transfer-hook-interface/src/instruction.rs
joncinque: Everywhere else in programs, it's `crate::instruction::instruction_name`, which I've always liked, since it's very clear and avoids repetition.

It sounds like `Execute` is probably a bad name. Maybe `HandleTransfer` would be better? Or `ProcessTransfer`? Or `CallProcedure`? Or `HandleEvent`?

token/transfer-hook-interface/src/instruction.rs
joncinque: Sorry, what's "they" in this case? This interface and token-2022? They are!

joncinque:
Never mind, I've seen the light. There's a new crate `spl-transfer-hook-example`, which just does the example program. It makes the interface much much neater.

buffalojoec:
Left one more comment on instruction naming convention @joncinque but this looks good to me

libraries/type-length-value/src/discriminator.rs
```diff
@@ -34,6 +34,11 @@ impl AsRef<[u8]> for Discriminator {
         &self.0[..]
     }
 }
+impl AsRef<[u8; Discriminator::LENGTH]> for Discriminator {
+    fn as_ref(&self) -> &[u8; Discriminator::LENGTH] {
+        &self.0
+    }
+}
```
ngundotra, line 41: That's pretty fire

token/transfer-hook-example/src/processor.rs
```diff
@@ -0,0 +1,139 @@
...
+        TransferHookInstruction::Execute { amount } => {
+            msg!("Instruction: Execute");
+            process_execute(program_id, accounts, amount)
+        }
+        TransferHookInstruction::InitializeExtraAccountMetas => {
+            msg!("Instruction: InitializeExtraAccountMetas");
+            process_initialize_extra_account_metas(program_id, accounts)
+        }
+    }
+}
```
ngundotra, line 139: Gorgeous!

token/transfer-hook-interface/src/invoke.rs
```diff
@@ -0,0 +1,51 @@
...
+        authority_info,
+        validation_info.clone(),
+    ];
+    ExtraAccountMetas::add_to_cpi_instruction::<instruction::ExecuteInstruction>(
+        &mut cpi_instruction,
+        &mut cpi_account_infos,
+        &validation_info.try_borrow_data()?,
+        additional_accounts,
+    )?;
+    invoke(&cpi_instruction, &cpi_account_infos)
```
ngundotra, line 50: does this need to be `invoke_signed`?

token/transfer-hook-interface/src/instruction.rs
ngundotra: Fire thank you!

token/transfer-hook-interface/src/instruction.rs
ngundotra: I think `Execute` is okay, especially since it's actual ix-bytes are hashed with the prefix of the interface name.

👍 lgtm

ngundotra (approved):
LGTM, just checking the `invoke` vs `invoke_signed` in `invoke.rs`

buffalojoec (approved):

token/transfer-hook-interface/src/invoke.rs
joncinque: Nope! No PDAs expected to sign as part of the interface. But it's definitely an idea to keep in mind for other interfaces which may require a PDA and a signature for that PDA. The IDL may need to encode that information somehow.

joncinque:
Thanks for all the feedback! I think I'll stick with `Execute` in the end to avoid the double "transfer" in something like `spl_transfer_hook_interface::instruction::handle_transfer`. This can certainly change in the future though

PR #4148, opened 2023-04-29
Fix ESM type exports in spl-token-swap

gh-888:
When importing @solana/spl-token-swap as an ESM module, the typescript compiler is unable to resolve the types, we need to export them from package.json.

### DIFF
--- a/token-swap/js/package.json
+++ b/token-swap/js/package.json
@@ -21,7 +21,8 @@
   "sideEffects": false,
   "exports": {
       "import": "./dist/esm/index.js",
-      "require": "./dist/cjs/index.js"
+      "require": "./dist/cjs/index.js",
+      "types": "./dist/types/index.d.ts"
   },
   "files": [
     "dist",


gh-888:
I've created a dummy repo on where we can repro the error: https://github.com/gh-888/spl-token-swap-esm-ts

joncinque (commented):
Hi there, thanks for your contribution! Are you sure that this works as intended? Looking at the documentation at https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing -- it specifies to always put "types" first, and it creates a new object for `"import"` and `"require"`. Same with https://nodejs.org/api/packages.html#community-conditions-definitions

gh-888:
@joncinque Thanks for taking a look, yep I believe this works as intended. It also follows similar changes that were done for the token program, see #3508

The [typescript documentation](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing) mentions this:

> With TypeScript’s original Node support, it would look for a "main" field, and then look for declaration files that corresponded to that entry. For example, if "main" pointed to ./lib/index.js, TypeScript would look for a file called ./lib/index.d.ts. A package author could override this by specifying a separate field called "types" (e.g. "types": "./types/index.d.ts").
> 
> The new support works similarly with [import conditions](https://nodejs.org/api/packages.html). By default, TypeScript overlays the same rules with import conditions - if you write an import from an ES module, it will look up the import field, and from a CommonJS module, it will look at the require field. If it finds them, it will look for a corresponding declaration file. If you need to point to a different location for your type declarations, you can add a "types" import condition.

I believe we wouldn't need this change if `index.d.ts` was in `dist/esm`, but since the package supports both cjs and esm, there's no choice than putting it in a different location, hence the need to specify it in `exports`.

Feel free to try in my repo (the one that I link above). By default, `tsc` will error, but if we manually edit `node_modules/@solana/spl-token-swap/package.json` with the proposed change, then it's all good.

Hope this clarifies :)

joncinque:
> I believe we wouldn't need this change if index.d.ts was in dist/esm, but since the package supports both cjs and esm, there's no choice than putting it in a different location, hence the need to specify it in exports.

Ah gotcha, thanks for the explanation!

> Feel free to try in my repo (the one that I link above). By default, tsc will error, but if we manually edit node_modules/@solana/spl-token-swap/package.json with the proposed change, then it's all good.

And thanks for providing the example, it made things a lot clearer. I can confirm that the change fixes the token-swap import error

joncinque (approved):

joncinque:
Published version 0.2.2 -- thanks for your help!

PR #4149, opened 2023-05-01
js: move `types` condition to the front

Andarist:
I moved `types` condition to the front. `package.json#exports` are **order-sensitive** - they are always matched from the top to the bottom. When a match is found then it should be used and no further matching should occur. 

Right now, the current setup works in TypeScript but it's considered a bug and it should not be relied upon, see the thread and the comment [here](https://github.com/microsoft/TypeScript/issues/50762#issuecomment-1528318260). For that reason, I would like to fix all popular packages that misconfigured their `exports` this way so the bug can be fixed in TypeScript.

⚠️ this PR focuses solely on fixing "🐛 Used fallback condition" problem but the "👺 Masquerading as ESM" and "⚠️ ESM (dynamic import only)" remain here. You can check the reported errors [here](https://arethetypeswrong.github.io/?p=%40solana%2Fspl-token%400.3.7)

### DIFF
--- a/account-compression/sdk/package.json
+++ b/account-compression/sdk/package.json
@@ -26,9 +26,9 @@
   "types": "./dist/types/index.d.ts",
   "exports": {
     ".": {
+      "types": "./dist/types/index.d.ts",
       "require": "./dist/cjs/index.js",
-      "import": "./dist/cjs/index.js",
-      "types": "./dist/types/index.d.ts"
+      "import": "./dist/cjs/index.js"
     },
     "./idl/spl_account_compression.json": "./idl/spl_account_compression.json"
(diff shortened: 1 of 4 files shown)


joncinque (approved):
Thanks for fixing this everywhere!

Andarist:
@joncinque is there any rough ETA for the next release that would include this fix?

joncinque:
These are typically published ad-hoc, so whenever you need it :sweat_smile: I can publish memo, token, and token-swap today, but I'm not sure if there are other considerations for account compression. Which package(s) do you need?

joncinque:
All right, they're done!

PR #4164, opened 2023-05-04
transfer hook manager

buffalojoec:
This PR draft introduces the concept of a `TransferHookManager` helper struct, which applies all the checks @joncinque demonstrated in his example program in PR #4151 .

Marking this a draft for now to gauge interest in the concept and overall flow.

Note: I based this on @joncinque 's branch for PR #4151 so it's going to show way more changes than that.

---

Developers who wish to create a validation account for the `TransferHook` extension can use this simple one-liner to write the validation data to their validation account. 

They can also use another simple one-liner to validate the validation account data.

This will likely need to be modified to support more extensions, and perhaps may not even need to be isolated to just `TransferHook`, but I wanted to share to get feedback!

---

Key files:
* https://github.com/buffalojoec/solana-program-library/blob/master/token/transfer-hook-interface/src/manager.rs
* https://github.com/buffalojoec/solana-program-library/blob/master/token/transfer-hook-example/src/processor.rs


token/libraries/concurrent-merkle-tree/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
```
austbot, line 1: why is this moved here?

buffalojoec: I have this compared with `master` but I based it off Jon's recent PR, so I might need to change the base branch

PR #4165, opened 2023-05-05
transfer-hook: Add top-level tests and helpers

joncinque:
#### Problem

Even though token-2022 properly supports the transfer hook interface with #4151, we still can't do the cherry-on-top: a generic swap of one hooked token for another.

#### Solution

Add more onchain and offchain helpers to support the full flow, and a test showcasing the whole thing. It's really neat, you just need to use the on-chain helper with all of the known accounts, and it does the rest for you. Each layer only knows what it needs to get to the next layer.

The relevant test is in `transfer_hook.rs:success_transfers_using_onchain_helper()`. It was a nightmare to debug when one of the helpers forgot one account :upside_down_face: 

NOTE: This is done on top of #4151, so you only need to look at the last commit to eventually see what's coming.

cc @austbot


token/program-2022/src/offchain.rs
```diff
@@ -0,0 +1,58 @@
...
+    },
+    solana_program::{instruction::AccountMeta, program_error::ProgramError, pubkey::Pubkey},
+    spl_transfer_hook_interface::offchain::get_extra_account_metas,
+    std::future::Future,
+};
+
+/// Offchain helper to get all additional required account metas for a checked transfer
+///
+/// To be client-agnostic and to avoid pulling in the full solana-sdk, this
+/// simply takes a function that will return its data as `Future<Vec<u8>>` for
```
buffalojoec, line 17: what is this, JavaScript? 😂

I dig, didn't know you could do this in Rust tbh

token/transfer-hook-interface/src/onchain.rs
```diff
@@ -49,3 +52,40 @@ pub fn execute<'a>(
...
+        .ok_or(TransferHookError::IncorrectAccount)?;
+
+    ExtraAccountMetas::add_to_cpi_instruction::<instruction::ExecuteInstruction>(
+        cpi_instruction,
+        cpi_account_infos,
+        &validation_info.try_borrow_data()?,
+        additional_accounts,
+    )?;
+    cpi_account_infos.push(validation_info.clone());
+    cpi_account_infos.push(program_info.clone());
```
buffalojoec, line 83: The ordering of accounts here jumps out at me a bit. In the `instruction.rs` of `transfer-hook-interface` you have:
```text
///   0. `[]` Source account
///   1. `[]` Token mint
///   2. `[]` Destination account
///   3. `[]` Source account's owner/delegate
///   4. `[]` Validation account
///   5..5+M `[]` `M` additional accounts, written in validation account data
```

ngundotra:
The swap example looks really nice! no issues pop out to me, just a lot of dev tooling to be done in near future 😅 

LGTM!

ngundotra (dismissed):

token/transfer-hook-interface/src/onchain.rs
joncinque: What's cool is that it doesn't matter! The helpers pull them out in the right order, and in a few other places it's also using this ordering. It's true that it might look unclear though, so I'll fix up the ordering everywhere

buffalojoec: Or just outline with a comment that the helper muscles that! Good stuff!

buffalojoec (approved):

PR #4189, opened 2023-05-05
spl-program-error macros

buffalojoec:
This PR introduces a library for derive macros & a procedural macro attribute that can make implementing traits for Solana Program error much easier.

---

Macros for implementing error-based traits on enums.

- `#[derive(IntoProgramError)]`: automatically derives the trait `From<Self> for solana_program::program_error::ProgramError`.
- `#[derive(DecodeError)]`: automatically derives the trait `solana_program::decode_error::DecodeError<T>`.
- `#[derive(PrintProgramError)]`: automatically derives the trait `solana_program::program_error::PrintProgramError`.
- `#[spl_program_error]`: Automatically derives all below traits:
  - `Clone`
  - `Debug`
  - `Eq`
  - `DecodeError`
  - `IntoProgramError`
  - `PrintProgramError`
  - `thiserror::Error`
  - `num_derive::FromPrimitive`
  - `PartialEq`


libraries/program-error-derive/README.md
```diff
@@ -0,0 +1,166 @@
...
+                ::solana_program::log::sol_log("Mint has no mint authority")
+            }
+            ExampleError::IncorrectMintAuthority => {
+                ::solana_program::log::sol_log(
+                    "Incorrect mint authority has signed the instruction",
+                )
+            }
+        }
+    }
+}
```
joncinque, line 165: It's these last three that are the most annoying to remember to add. Rather than just one macro that adds everything, how about several macros for each of these?

It looks like `impl From<ExampleError> for solana_program::program_error::ProgramError` requires `num_derive::FromPrimitive` to do that `e as u32` part.

And it looks like `impl solana_program::program_error::PrintProgramError` requires `thiserror::Error` and `DecodeError`.

So maybe there's four separate macros, with an option to do the full thing. ie.

```
#[print_program_error]
#[decode_error]
#[into_program_error]
#[Derive(...)]
pub enum MyError { }
```

but then you have `#[spl_program_error]` which just does all of them. That way you can do whatever you wish.  What do you think?

Side note: we might even consider adding them directly to https://github.com/solana-labs/solana/tree/master/sdk/macro as separate macros.

joncinque (commented):
Great work! This seems pretty close already, just a few bits to sort out

buffalojoec:
Resolved all comments, thanks @joncinque !

Let's sync on the idea of PR'ing this into the Solana SDK once you've had another look.

libraries/program-error/derive/src/lib.rs
```diff
@@ -0,0 +1,45 @@
+//! Crate defining a procedural macro for building Solana program errors
+
+#![allow(clippy::integer_arithmetic)]
```
joncinque, line 3: nit: Is this `allow` required? If yes, could you add a comment explaining why?

libraries/program-error/tests/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+            Into::<ProgramError>::into(bench::ExampleError::IncorrectMintAuthority),
+            Into::<ProgramError>::into(print::ExampleError::IncorrectMintAuthority),
+        );
+        // `DecodeError<T>`
+        assert_eq!(
+            <bench::ExampleError as DecodeError<bench::ExampleError>>::type_of(),
+            <bench::ExampleError as DecodeError<print::ExampleError>>::type_of(),
+        );
+        // `PrintProgramError`
+        // (!) Not sure how better to test this yet - thoughts?
```
joncinque, line 66: There *is* a way, but it's a little annoying.

Since these are calling `msg!` off-chain, that'll default to `println!` based on the default program stubs at https://github.com/solana-labs/solana/blob/0495051a673b0599385891bb2b5caaf650eab879/sdk/program/src/program_stubs.rs#L27

If you want to test this properly, you can implement your own type which implements `SyscallStubs` for `sol_log` which writes the string to some mutable static variable, and then read out that string. That's how we used to test programs back before program-test!

The token program does something similar to this for return data, so you should be able to copy that: https://github.com/solana-labs/solana-program-library/blob/f1620caeaaf739f21ef788e7a19662b8e62c2009/token/program/src/processor.rs#L1055

libraries/program-error/tests/print.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[error("Mint has no mint authority")]
+    MintHasNoMintAuthority,
+    /// Incorrect mint authority has signed the instruction
+    #[error("Incorrect mint authority has signed the instruction")]
+    IncorrectMintAuthority,
+}
+
+/// Tests that all macros compile
+#[test]
+fn _test() {
```
joncinque, line 28: nit: may as well just call this `test_macros_compile`

libraries/program-error/tests/into.rs
```diff
@@ -0,0 +1,22 @@
...
+    #[error("Mint has no mint authority")]
+    MintHasNoMintAuthority,
+    /// Incorrect mint authority has signed the instruction
+    #[error("Incorrect mint authority has signed the instruction")]
+    IncorrectMintAuthority,
+}
+
+/// Tests that all macros compile
+#[test]
+fn _test() {
```
joncinque, line 20: nit: same here, may as well just call this `test_macros_compile`. side note: I'm not a fan of underscore prefacing when it's not necessary

libraries/program-error/tests/decode.rs
```diff
@@ -0,0 +1,29 @@
...
+    #[error("Mint has no mint authority")]
+    MintHasNoMintAuthority,
+    /// Incorrect mint authority has signed the instruction
+    #[error("Incorrect mint authority has signed the instruction")]
+    IncorrectMintAuthority,
+}
+
+/// Tests that all macros compile
+#[test]
+fn _test() {
```
joncinque, line 27: nit: same here, may as well just call this `test_macros_compile`

libraries/program-error/derive/src/macro_impl.rs
```diff
@@ -1,54 +1,69 @@
...
-    let ppe_match_arms = variants.iter().map(|variant| {
-        let variant_ident = &variant.ident;
-        let error_msg = get_error_message(variant)
-            .unwrap_or_else(|| String::from("Unknown custom program error"));
-        quote! {
-            #ident::#variant_ident => {
-                solana_program::msg!(#error_msg)
+impl MacroType {
+    /// Generates the corresponding tokens based on variant selection
+    pub fn generate_tokens(&self, item_enum: ItemEnum) -> proc_macro2::TokenStream {
```
joncinque, line 16: Slick, I like it!

joncinque (commented):
Looks really great, just a few last little nits -- nice job!

joncinque:
As for moving it to `solana-sdk`, I'm remembering discussions about how the crate is too big in general, so maybe we can keep this in spl. If you wanted to add a docs page later for https://spl.solana.com, that might help for discoverability

libraries/program-error/derive/src/lib.rs
buffalojoec: Nope apparently not, not sure why I even had it in there in the first place....

buffalojoec:
> As for moving it to `solana-sdk`, I'm remembering discussions about how the crate is too big in general, so maybe we can keep this in spl. If you wanted to add a docs page later for https://spl.solana.com, that might help for discoverability

Should we add a tab for "Libraries" and get some docs on TLV as well?

joncinque:
> Should we add a tab for "Libraries" and get some docs on TLV as well?

Yeah that'd be good

libraries/program-error/derive/src/lib.rs
buffalojoec: It's required in the derive sub-crate, that was why !

libraries/program-error/tests/mod.rs
buffalojoec: Working on this - last piece

buffalojoec:
Added the required tests for `PrintProgramError`.

Circling back to the `#![allow(clippy::integer_arithmetic)]` inside the `derive/src/lib.rs`, when I remove it, clippy barfs on the following:
```rust
error: integer arithmetic detected
  --> libraries/program-error/derive/src/macro_impl.rs:66:5
   |
66 | /     quote! {
67 | |         impl solana_program::program_error::PrintProgramError for #ident {
68 | |             fn print<E>(&self)
69 | |             where
...  |
80 | |         }
81 | |     }
   | |_____^
   |
   = note: requested on the command line with `-D clippy::integer-arithmetic`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#integer_arithmetic
   = note: this error originates in the macro `$crate::quote_token_with_context` (in Nightly builds, run with -Z macro-backtrace for more info)
```

As the error states, it looks like the code underlying `quote_token_with_context` uses integer arithmetic. If you'd rather I don't include the clippy omission I can see about using a different method

joncinque:
> As the error states, it looks like the code underlying quote_token_with_context uses integer arithmetic. If you'd rather I don't include the clippy omission I can see about using a different method

Ah ok, don't worry about it then -- can you just add a comment explaining this?

libraries/program-error/tests/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+            <bench::ExampleError as DecodeError<spl::ExampleError>>::type_of(),
+        );
+        // `PrintProgramError`
+        set_expected_data("Mint has no mint authority".as_bytes().to_vec());
+        PrintProgramError::print::<spl::ExampleError>(&spl::ExampleError::MintHasNoMintAuthority);
+        set_expected_data(
+            "Incorrect mint authority has signed the instruction"
+                .as_bytes()
+                .to_vec(),
+        );
```
joncinque, line 133: Looks like this was missing the check
```suggestion
        set_expected_data(
            "Incorrect mint authority has signed the instruction"
                .as_bytes()
                .to_vec(),
        );
        PrintProgramError::print::<print::ExampleError>(
            &print::ExampleError::IncorrectMintAuthority,
        );
```

libraries/program-error/tests/mod.rs
```diff
@@ -0,0 +1,135 @@
...
+        // `DecodeError<T>`
+        assert_eq!(
+            <bench::ExampleError as DecodeError<bench::ExampleError>>::type_of(),
+            <bench::ExampleError as DecodeError<decode::ExampleError>>::type_of(),
+        );
+    }
+    // `#[derive(PrintProgramError)]`
+    #[test]
+    fn test_derive_print_program_error() {
```
joncinque, line 67: To avoid segfaulting or other potential false negatives, this test and the following should be executed `#[serial]`ly

joncinque (commented):
Just about there! Along with adding the comment about the `#[allow(...)]`

libraries/program-error/tests/mod.rs
```diff
@@ -32,6 +33,7 @@ mod tests {
 
     // `#[derive(IntoProgramError)]`
     #[test]
+    #[serial]
```
joncinque, line 36: Since this one doesn't use the expected data, it should be fine to run in parallel, right?

libraries/program-error/tests/mod.rs
```diff
@@ -46,6 +48,7 @@ mod tests {
 
     // `#[derive(DecodeError)]`
     #[test]
+    #[serial]
```
joncinque, line 51: Same here

libraries/program-error/derive/src/lib.rs
```diff
@@ -1,5 +1,12 @@
 //! Crate defining a procedural macro for building Solana program errors
 
+// Required to include `#[allow(clippy::integer_arithmetic)]`
+// below since the tokens generated by `quote!` in the implementation
+// for `MacroType::PrintProgramError` and `MacroType::SplProgramError`
+// trigger the lint upstream through `quote_token_with_context` within the
+// `quote` crate
+//
+// Culprate is `macro_impl.rs:66`
```
joncinque, line 9: nit: culprit

joncinque (commented):
Looks like there are some build errors due to a lockfile error. Can you rebase your branch?

libraries/program-error/tests/mod.rs
buffalojoec: yeah true, sounds good, done!

buffalojoec:
> Looks like there are some build errors due to a lockfile error. Can you rebase your branch?

Rebased, but same problem. Any idea why it's adding it twice? I've just added it as a dev dependency like you're doing in a few `token` crates

joncinque:
The weird thing is that the base of your Cargo.lock looks totally different than what's actually in the repo. For example, if you look at https://github.com/solana-labs/solana-program-library/blob/master/Cargo.lock -- there's only one "serial_test" at 2.0.0, whereas yours has 0.5.1 and 0.8.0.

Try copying the lockfile by hand from master, and then running your tests. That should bring it into a better state

buffalojoec:
> The weird thing is that the base of your Cargo.lock looks totally different than what's actually in the repo. For example, if you look at https://github.com/solana-labs/solana-program-library/blob/master/Cargo.lock -- there's only one "serial_test" at 2.0.0, whereas yours has 0.5.1 and 0.8.0.
> 
> Try copying the lockfile by hand from master, and then running your tests. That should bring it into a ~better~ correct state

Lol I did this earlier but it felt.. wrong. It works though

buffalojoec:
I just figured it out - the Rust Analyzer VS Code extension 😭

PR #4216, opened 2023-05-09
associated-token-account: Refresh blockhash more on flakey tests

joncinque:
#### Problem

The ATA tests have been flaky, potentially because there's not enough blockhash refreshing going on.#4214 

#### Solution

I can't reproduce locally, but refreshing the blockhash more often has helped the stake pool tests in the past, so do it more often in the ATA tests too.

I've also removed some unnecessary keypair generations, which have been a problem in the past too.


joncinque:
Only the ATA tests count, and those passed, so merging on red

PR #4217, opened 2023-05-09
token-lending: Remove unused genesis code and dependencies

joncinque:
#### Problem

While looking at #4208, I noticed that serde_yaml is only used in one place, the token-lending genesis test helpers. The genesis helpers are completely unused at this point, and dead code is a maintenance burden.

#### Solution

Remove the genesis helpers and the additional crates that it pulls in. If these are still used though, I'm happy to close the PR -- let me know!


jordansexton (approved):
Yeah, makes sense to remove now that this isn't used. I don't think I ever needed this when working on the project either.

PR #4229, opened 2023-05-09
docs: Scrub all references to token-list

joncinque:
#### Problem

https://github.com/solana-labs/token-list has been archived for awhile, but our docs still recommend to use it.

#### Solution

Change to a link to Metaplex docs. Also, why was it referenced in the name service JS bindings? The world may never know

### DIFF
--- a/docs/src/token.mdx
+++ b/docs/src/token.mdx
@@ -1819,11 +1819,12 @@ The sender's wallet must not require that the recipient's main wallet address
 hold a balance before allowing the transfer.
 
 ### Registry for token details
+
 At the moment there exist two solutions for Token Mint registries:
 
-* hard coded addresses in the wallet or dapp
-* [spl-token-registry](https://www.npmjs.com/package/@solana/spl-token-registry)
-package, maintained at [https://github.com/solana-labs/token-list](https://github.com/solana-labs/token-list)
+* Hard coded addresses in the wallet or dapp
+* Metaplex Token Metadata. Learn more at the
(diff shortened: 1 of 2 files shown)


docs/src/token.mdx
```diff
@@ -1819,11 +1819,12 @@ The sender's wallet must not require that the recipient's main wallet address
...
 ### Registry for token details
+
 At the moment there exist two solutions for Token Mint registries:
 
-* hard coded addresses in the wallet or dapp
-* [spl-token-registry](https://www.npmjs.com/package/@solana/spl-token-registry)
-package, maintained at [https://github.com/solana-labs/token-list](https://github.com/solana-labs/token-list)
+* Hard coded addresses in the wallet or dapp
+* Metaplex Token Metadata. Learn more at the
+[Token Metadata Documentation](https://docs.metaplex.com/programs/token-metadata/)
```
jordansexton, line 1827: Do we want to link to https://github.com/solana-labs/token-list still from here? It has links and instructions to use some tools to create a token with metadata that might be more helpful than just Metaplex docs for people who aren't familiar with the history of the token list.

jordansexton (approved):
LGTM

docs/src/token.mdx
joncinque: That's a good point... ok, i'll put it back with an EOL warning. Even months later, many people have linked the token-list repo in questions without reading the warning at the top

joncinque:
The docs CI passed, which is all that matters -- merging during CI!

PR #4261, opened 2023-05-11
ci: Pin cargo-audit to earlier version

joncinque:
#### Problem

The audit job is failing because the newest version of cargo-audit requires Rust version 1.63.

#### Solution

Until SPL moves to Solana 1.16, we're stuck with Rust 1.60, so temporarily pin the version of cargo-audit


joncinque:
Audit passed, which is all that matters -- merging early!

PR #4355, opened 2023-05-19
Adding PDA support to TLV Account Resolution library

buffalojoec:
This PR adds support for PDAs derived by some set of seeds but whose address may not be known (or possible to know) at the time of creating the validation account for the interface program.

---
🔍 TL/DR:
TLV Account Resolution now supports PDA seeds as a configuration instead of a fixed address. However, this means we have to carry the seeds around and pass them into account resolution functions, which is not desirable and looks awful.

Hoping to get some ideas/thoughts on this, which is why this bad boy is a draft.

---
Currently, using the TLV Account Resolution library only allows you to enforce required accounts for validation that have a fixed address at the time of creating the validation account. This PR aims to solve this problem.

This PR adds:
* A `RequiredAccount` enum to wrap both types of accounts:
    * Accounts with fixed addresses (currently packed into the validation account as `AccountMeta`)
    * Accounts that have program-derived addresses from some set of seeds
* `AccountMetaPda` struct for managing account information with seeds instead of an address
* `seeds.rs` module for handling types of seeds and their corresponding configurations/derivations
* Addition of discriminator to `PodAccountMeta`
* Handling of seeds in any functions adding accounts to an instruction, ie. `add_to_vec(..)`

---
Notable snippets:
* [Passing seeds while adding accounts to an instruction](https://github.com/buffalojoec/solana-program-library-jonc/blob/joec-tlv-pdas/libraries/tlv-account-resolution/src/state.rs#L653-L670)
* [Evaluation of seeds against their required data types](https://github.com/buffalojoec/solana-program-library-jonc/blob/joec-tlv-pdas/libraries/tlv-account-resolution/src/seeds.rs#L141-L150)
* [De-escalation with seeds](https://github.com/buffalojoec/solana-program-library-jonc/blob/joec-tlv-pdas/libraries/tlv-account-resolution/src/state.rs#L303-L338)

---
Drawbacks:
* [May have to pass seeds in program instruction](https://github.com/buffalojoec/solana-program-library-jonc/blob/joec-tlv-pdas/token/program-2022/src/processor.rs#L268)
* [Have to pass seeds in off-chain helper](https://github.com/buffalojoec/solana-program-library-jonc/blob/joec-tlv-pdas/token/client/src/token.rs#L808)

---
Some notes:
* Right now, the configuration for seeds (as represented in the buffer) is a `[u8; 32]` - in order to align with the size of a `Pubkey`, however we can possibly modify this, since we don't need 32 seeds!
* Much of the original functionality remains in tact. Specifically, for those who do not need to use PDAs in their validation account, they can use the original function laid down by @joncinque - ie. `init_with_account_metas`.
* The only change anyone not using PDAs would notice is the additional function parameter (`Option<Vec<SeedConfig>>`), where you can simply provide a `None`.
* ⚠️ Checks will not pass until I've added the seeds argument everywhere necessary - this includes client, CLI, tests, and examples. We should land on how these args should look for each use case first.


libraries/tlv-account-resolution/README.md
```diff
@@ -39,21 +39,112 @@ ExtraAccountMetas::init_with_account_metas::<MyInstruction>(&mut buffer, &extra_
...
+        Pubkey::new_unique(),
+        Pubkey::new_unique(),
+        Pubkey::new_unique(),
+    )),
+];
+
+// Then you provide the seeds as you add accounts to the instruction (off-chain)
+let program_id = Pubkey::new_unique();
+let mut instruction = Instruction::new_with_bytes(program_id, &[], vec![]);
+ExtraAccountMetas::add_to_instruction::<MyInstruction>(&program_id, &mut instruction, &buffer, Some(provided_seeds)).unwrap();
```
joncinque, line 126: We can potentially simplify this tremendously by *enforcing* that PDAs must be derivable by:

* the account pubkeys already in the instruction
* the literal account pubkeys in the ExtraAccountMetas
* the instruction data bytes already there
* any literal data stored in its entry

This way, we never need to pass `provided_seeds`, and the Seeds types become:

```
enum Seeds {
    InstructionPubkey(u8), // index in the instruction
    ExtraAccountPubkey(u8), // index in `ExtraAccountMetas`
    InstructionBytes(u8, u8), // a range to use from the account bytes
    Literal(Vec<u8>), // not actually a vec, but some number of encoded bytes
}
```

Naturally, you'll need to do something more clever for the literal, maybe you can simplify this by making different enum variants for different sizes, ie

```
enum Seeds {
    ... everything else ...
    LiteralByte([u8; 1]),
    LiteralTwoBytes([u8; 2]),
    LiteralThreeBytes([u8; 3]),
   ... etc, probably up to 16 or so?
}
```

What do you think?

libraries/tlv-account-resolution/src/seeds.rs
```diff
@@ -0,0 +1,526 @@
...
+/// As you can see from the implementations below, we are
+/// currently supporting up to 10 seeds - which is probably
+/// more than necessary
+pub trait ProvidedSeeds {
+    /// A vector of the type of seeds used to build the config
+    fn seed_types(&self) -> Vec<Seed>;
+    /// A vector of bytes for use in `Pubkey::find_program_address`
+    fn byte_vectors(&self) -> Vec<Vec<u8>>;
+}
+impl<S0> ProvidedSeeds for (S0,)
```
joncinque, line 182: These are very neat

joncinque (commented):
This is really great work! My main question is: where do the seeds come from? For example, if I'm a client, how do I figure the additional seeds that I might have to pass in? If we strongly constrain where the seeds come from, as in my comment, I think we'll simplify a lot while keeping it as expressive as anyone would need.

And sorry that it took me so long to get to this :disappointed:

libraries/tlv-account-resolution/README.md
buffalojoec: Yes, I agree! This was the direction I realized this needed to go once I finally worked my way out to the client-side calls and token program(s). 

With the approach you're suggesting, we can implement account resolution exactly the way anchor's SDK does it, where you almost don't even notice it's happening.

This will be my next step - and that will bring this out of draft phase if we still want this supported

buffalojoec: In fact, your enum closely resembles the `seeds` in both Anchor's and Shank's IDLs

buffalojoec:
> This is really great work! My main question is: where do the seeds come from? For example, if I'm a client, how do I figure the additional seeds that I might have to pass in? If we strongly constrain where the seeds come from, as in my comment, I think we'll simplify a lot while keeping it as expressive as anyone would need.
> 
> And sorry that it took me so long to get to this 😞

Yep, this was the exact snag. I mentioned this a little bit in my one reply to your in-line comment but essentially I got all of the implementation, bytemuck, and what not in there and then started poking around the idea of passing the seeds in, but it started to very quickly look horrendous.

So essentially I'd like to go with automatic resolution as you suggested in that msg 💪🏼

libraries/tlv-account-resolution/README.md
joncinque: Sweet, I think this is definitely worth adding.

buffalojoec: Copy that, will circle back on this one

buffalojoec:
Closing in favor of new PR coming

PR #4403, opened 2023-05-26
[WIP] Request for Comment, Add Back the Init with root

austbot:
Teams have been asking for the set with root functionality. Previously this was thought to be a DOS vector for indexers, but that was only in the NFT case. Set With root will allow other use cases of account compression that have their own indexing systems to set a root and initial canopy so that they dont need to do a 'shadow' indexing strategy of making small modifications to each node to "bump" the indexer. In these scenarios the trees may be quite large, hence I have provided a way to set a proof in a buffer and init the tree with that buffer.

It doesnt currently function but there is enough code for you to get the idea. 
Optionally let them send proofs from an account to filla really big trees initial canopy thereby reducing the nodes needed for further proving.

Let me know if this is a direction you are okay with and if so I will finish the work

### DIFF
--- a/account-compression/Cargo.lock
+++ b/account-compression/Cargo.lock
@@ -24,23 +24,23 @@ dependencies = [
 
 [[package]]
 name = "anchor-attribute-access-control"
-version = "0.26.0"
+version = "0.27.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


ngundotra:
Hey Austin, thanks for writing this up. Can you explain the algorithm via an sRFC? I'd like to keep track of algorithmic stuff separately from the code review

PR #4433, opened 2023-05-31
docs: Delay token-2022 status dates by one month as we finish

joncinque:
#### Problem

The actual token-2022 program completion will likely take a couple more weeks.

#### Solution

Bump out the dates in the status page. Let me know if these dates look ok @samkim-crypto !

### DIFF
--- a/docs/src/token-2022/status.md
+++ b/docs/src/token-2022/status.md
@@ -12,18 +12,17 @@ Here is the general program timeline and rough ETAs:
 
 | Issue | ETA |
 | --- | --- |
-| Finish program | end of June |
-| Final audit | end of July |
-| Mainnet recommendation | middle of August |
-| Freeze program | Q1 2024 |
+| Code-complete & final audit | Summer 2023 |
+| Mainnet recommendation | Fall 2023 (depends on v1.16) |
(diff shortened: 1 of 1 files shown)


docs/src/token-2022/status.md
```diff
@@ -12,9 +12,9 @@ Here is the general program timeline and rough ETAs:
 
 | Issue | ETA |
 | --- | --- |
-| Finish program | end of May |
-| Final audit | end of June |
-| Mainnet recommendation | middle of July |
+| Finish program | end of June |
+| Final audit | end of July |
+| Mainnet recommendation | middle of August |
```
mvines, line 17: I'd push this out further as it current depends on adoption of v1.16 and feature gate activations.

Actually explicitly calling the the v1.16 dep and the corresponding feature gates here would be super, that gives people more vis into the real release constraints and a handy excuse for why the estimate date might be pushed out later :)

joncinque: Ah I see. My initial thinking was that we can officially recommend it on mainnet before confidential transfers are available, but it seems like that'll be too confusing.

Thanks for the fly-by review!

samkim-crypto (approved):

PR #4435, opened 2023-06-01
docs: Be really clear that token-2022 is still under audit

joncinque:
#### Problem

Only the token-2022 status page is very clear that the program is still under audit, but people may miss that information.

#### Solution

Add the information to the top-level token-2022 docs page, the program's README, and the repo SECURITY page.


mvines (approved):
Thanks for doing this so quickly

joncinque:
Merging on red, just docs changes

PR #4456, opened 2023-06-02
fix cargo-audit task

2501babe:
the audit task has a dependency resolution problem that i will now attempt to debug

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -99,7 +99,7 @@ jobs:
         uses: actions-rs/install@v0.1
         with:
           crate: cargo-audit
-          version: 0.17.5 # TODO use latest once repo uses Rust v1.63+
+          version: 0.16.0 # TODO use latest once repo uses Rust v1.63+
 
       - name: Run Cargo Audit
         run: ./ci/do-audit.sh


2501babe:
we are in fact using 1.60.0 as we should be. it looks like grabbing my last success, it only succeeds when something else happens to have already installed it...? not sure how that happens

```
with:
    crate: cargo-audit
    version: latest
    use-tool-cache: false
    use-cache: true
  env:
    RUST_STABLE: 1.60.0
Tool cache is disabled in the Action inputs, skipping it
Falling back to the `cargo install` command
Installing "cargo-audit = latest"
  /home/runner/.cargo/bin/cargo install cargo-audit
      Updating crates.io index
   Downloading crates ...
    Downloaded cargo-audit v0.17.[5](https://github.com/solana-labs/solana-program-library/actions/runs/4796466598/jobs/8532258964#step:5:5)
       Ignored package `cargo-audit v0.1[7](https://github.com/solana-labs/solana-program-library/actions/runs/4796466598/jobs/8532258964#step:5:7).5` is already installed, use --force to override
```

in other words maybe stateful magic (idk) hid this, nornally it fails without a lockfile on 1.60, i guess play whackamole locking other versions of things...

2501babe:
downgrading to cargo-audit 0.16 lol, seems to have no adverse effects

joncinque (approved):
Thanks for fixing this!

PR #4464, opened 2023-06-05
Add link to doc: instruction.rs

skaunov:
`InitializeMint` contains important note(s), so it's better to facilitate peeking into it from `InitializeMint2`


joncinque:
There's an issue with the token CI currently, so merging this on red since it's just a docs change

joncinque (approved):

PR #4470, opened 2023-06-05
ci: Split up token job to avoid using all disk space

joncinque:
#### Problem

The token CI job is failing due to no space left on device, ie https://github.com/solana-labs/solana-program-library/actions/runs/5164629307/jobs/9303653512?pr=4450

#### Solution

Move the ATA and transfer hook example builds to separate steps. The transfer hook tests have no other dependencies, so it's easy. The ATA tests currently use the BPF version of token-2022, but the tests are fine with the fake native version too. Finally, the JS tests require the ATA program, so it's uploaded / downloaded separately.

### DIFF
--- a/.github/workflows/pull-request-token.yml
+++ b/.github/workflows/pull-request-token.yml
@@ -61,12 +61,6 @@ jobs:
       - name: Build and test token
         run: ./ci/cargo-test-sbf.sh token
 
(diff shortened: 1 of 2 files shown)


joncinque:
Shipping it! We might be able to also split out the "serde" build, but we can do that in another PR

PR #4471, opened 2023-06-05
Patch ntapi for windows build with rust 1.69+

ryoqun:
the solana monorepo's tarballs bundle `spl-token-cli` for convenience. and that fancy repo uses still shiny rust v1.69 (im eyeing for 1.70 soon). and `ntapi` (one of monorepo dep hell) is pulled into this spl repo. and it isn't buildable anymore with the rust when targeted to the Windows os.

it's currently worked around with this yet another sinful hack...: https://github.com/solana-labs/solana/pull/31970#discussion_r1218039494 

~so, this pr needs to be shipped rather promptly after merge to be included in spl-token-cli v2.4.1....~ (EDIT: the monorepo-side cli-arg based patch for cargo install spl-token-cli seem to a better approach.)

links:
- https://github.com/solana-labs/solana/pull/31961
- https://github.com/solana-labs/solana/pull/31970
- https://github.com/solana-labs/solana/pull/31971
- https://github.com/solana-labs/solana/pull/31981

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2864,8 +2864,7 @@ dependencies = [
 [[package]]
 name = "ntapi"
 version = "0.3.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f6bb902e437b6d86e03cce10a7e2af662292c5dfef23b65899ea3ac9354ad44"
+source = "git+https://github.com/solana-labs/ntapi?rev=5980bbab2e0501a8100eb88c12222d664ccb3a0a#5980bbab2e0501a8100eb88c12222d664ccb3a0a"
 dependencies = [
  "winapi 0.3.9",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2863,8 +2863,8 @@ dependencies = [
 
 [[package]]
 name = "ntapi"
-version = "0.3.6"
-source = "git+https://github.com/solana-labs/ntapi?rev=5980bbab2e0501a8100eb88c12222d664ccb3a0a#5980bbab2e0501a8100eb88c12222d664ccb3a0a"
+version = "0.3.7"
+source = "git+https://github.com/solana-labs/ntapi?rev=97ede981a1777883ff86d142b75024b023f04fad#97ede981a1777883ff86d142b75024b023f04fad"
 dependencies = [
  "winapi 0.3.9",
 ]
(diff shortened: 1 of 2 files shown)


yihau:
I think the root problem is to use rust 1.69+ to build but 
solana-program-library is still using 1.60.0
https://github.com/solana-labs/solana-program-library/blob/f3207f9ad753e6a1cf6cae603d7470f77439d600/rust-toolchain.toml#L2

not sure should we patch this one here. maybe we can do these things instead 🤔 
1. build/release spl bins in solana-program-library repo scope
2. if we want to bundle any of them in monorepo release, we only download those bins instead of using monorepo's rust version to rebuild them again.

ryoqun:
> maybe we can do these things instead thinking

yeah, that works. but then

> not sure should we patch this one here.

spl repo will/should use rust 1.69 eventually. and if the tokio upgrade (the old `ntapi` blocker) don't happen by the time, this patching is needed still. also, bp-ing the tokio upgrade to v1.16 won't be a viable option as well due to the size and nature of it just for restoring the windows build.....

joncinque (commented):
We can definitely merge this as is, but since we do `cargo install` for spl-token-cli in the monorepo, I'm not sure if this change will have the effect you're hoping.

`cargo publish` and `cargo install` only uses what's in crates.io typically, so windows users will still have the problem. On the flipside, if they build from source with this patch, then they'll be good.

I don't know of a good way to fix the `cargo install` issue, but we can at least merge this for people building from source.

Edit: Oh, and please correct me if I'm wrong!

ryoqun:
> `cargo publish` and `cargo install` only uses what's in crates.io

oooh, yeah, i worried this concern. i was about to verify this. now, i confirmed this is true... hmm

will think a bit for better alternatives...

ryoqun:
> will think a bit for better alternatives...

hey getting back to this with my own answer: https://github.com/solana-labs/solana/pull/31981#discussion_r1218783475

> we can at least merge this for people building from source.

now i think let's do this for this use case (maybe windows devs) and just for consistency. so i just pushed this to align with https://github.com/solana-labs/solana/pull/31981: https://github.com/solana-labs/solana-program-library/pull/4471/commits/307ae75ff72e6570c6ab784674ab4d380fa8296b

joncinque (approved):
Ok great, works for me, thanks for looking into this!

PR #4477, opened 2023-06-05
ci: Separate token-2022 serde test step

joncinque:
#### Problem

The token `cargo-test-sbf` is the slowest part of the token CI.

#### Solution

The token-2022 serde test step has no dependencies on any other part of the token CI test step, so separate it out!


joncinque:
Next we have to tackle the CLI tests, probably through some custom test harness, maybe integrating with https://github.com/LukasKalbertodt/libtest-mimic or some similar tool. This is a good step though!

PR #4496, opened 2023-06-07
Handle existing patch table..

ryoqun:
#4471 added `[patch.crates-io]` table. that broke ci script detection a bit.

so, make the script detection robust...


patch.crates-io.sh
```diff
@@ -88,11 +88,14 @@ done
 echo "Patching in $solana_ver from $solana_dir"
 echo
 for crate in "${workspace_crates[@]}"; do
-  if grep -q '\[patch.crates-io\]' "$crate"; then
+  if grep -q "# The following entries are auto-generated by $0" "$crate"; then
     echo "$crate is already patched"
   else
+    if ! grep -q '\[patch.crates-io\]' "$crate"; then
```
ryoqun, line 94: i know this is super inefficient for invoking `grep` many time but this is simplest solution and who cares for this tiny script? :)

ryoqun:
sorry i skipped reviews to restore solana v1.14 monorepo ci asap. and this change is small and only relevant to ci...

CriesofCarrots:
Ha, grepping for comment. I guess that works! Thanks for fixing

PR #4510, opened 2023-06-08
single pool: cli

2501babe:


2501babe:
note to self to update ci scripts so this triggers single pool and vice versa

stake-pool/single-pool-cli/src/cli.rs
```diff
@@ -0,0 +1,399 @@
...
+    /// The pool to create default MPL token metadata for
+    #[clap(short, long = "pool", value_parser = Pubkey::from_str)]
+    pub pool_address: Option<Pubkey>,
+
+    /// The vote account corresponding to the pool to create metadata for
+    #[clap(long = "vote-account", value_parser = Pubkey::from_str)]
+    pub vote_account_address: Option<Pubkey>,
+}
+
+// XXX i didnt include uri because it looks useless for fungible tokens
```
joncinque, line 206: Isn't that where they get the token image from? Have I been misunderstanding how token metadata works this whole time?

stake-pool/single-pool-cli/src/cli.rs
```diff
@@ -0,0 +1,399 @@
...
+
+pub fn is_valid_token_name(s: &str) -> Result<(), String> {
+    if s.len() > 32 {
+        Err("Maximum token name length is 32 characters".to_string())
+    } else {
+        Ok(())
+    }
+}
+
+pub fn is_valid_token_uri(s: &str) -> Result<(), String> {
```
joncinque, line 282: nit:
```suggestion
pub fn is_valid_token_symbol(s: &str) -> Result<(), String> {
```

stake-pool/single-pool-cli/src/cli.rs
```diff
@@ -0,0 +1,399 @@
...
+
+// this plus with_signer(s) is because solana clap v3 utils requires ArgMatches
+// which means its impossible to parse keypairs in value_parser
+// instead, we take the input into a string wrapper from the cli
+// and then once the first pass is over, we do a second manual pass converting to signer wrappers
+#[derive(Clone, Debug)]
+pub enum SignerArg {
+    Source(String),
+    Signer(Arc<dyn Signer>),
+}
```
joncinque, line 308: I believe this is a longstanding issue with why the conversion to clap v3 will be so difficult. Maybe @samkim-crypto can weigh in since he's been doing work in that area

stake-pool/single-pool-cli/src/cli.rs
```diff
@@ -0,0 +1,399 @@
...
+
+pub fn signer_from_arg(
+    signer_arg: Option<SignerArg>,
+    default_signer: &Arc<dyn Signer>,
+) -> Arc<dyn Signer> {
+    match signer_arg {
+        Some(SignerArg::Signer(signer)) => signer,
+        _ => default_signer.clone(),
+    }
+}
```
joncinque, line 334: Looking later, I think I understand this behavior more. Just to be sure though, is the default only meant to be used in the case of `None`? If so, can you return an error if it's `Some(SignerArg::Source(_))`?

stake-pool/single-pool-cli/src/cli.rs
```diff
@@ -0,0 +1,399 @@
...
+            } else {
+                return Err(format!("Cannot parse signer {} / {}", name, path).into());
+            };
+            Some(SignerArg::Signer(Arc::from(signer)))
+        }
+        a => a,
+    })
+}
+
+// TODO replace with monorepo OutputFormat once it has Clone
```
joncinque, line 394: Bummer

stake-pool/single-pool-cli/src/output.rs
```diff
@@ -0,0 +1,279 @@
...
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct StakePoolOutput {
+    #[serde_as(as = "DisplayFromStr")]
+    pub pool_address: Pubkey,
+    #[serde_as(as = "DisplayFromStr")]
+    pub vote_account_address: Pubkey,
+    pub available_stake: u64,
+    pub tokens_outstanding: u64,
```
joncinque, line 97: nit for clarity? or maybe `token_supply`?
```suggestion
    pub tokens_minted: u64,
```

stake-pool/single-pool-cli/src/output.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+}
+
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DepositOutput {
+    #[serde_as(as = "DisplayFromStr")]
+    pub pool_address: Pubkey,
+    pub token_amount: u64, // XXX should we stringify because js number...
```
joncinque, line 188: Eh, maybe save it for later. They shouldn't be using this from JS, right?

stake-pool/single-pool-cli/src/quarantine.rs
```diff
@@ -0,0 +1,197 @@
...
+        .program_client
+        .get_account(sysvar::rent::id())
+        .await?
+        .unwrap();
+    let rent = bincode::deserialize::<Rent>(&rent_data.data)?;
+
+    Ok(rent)
+}
+
+pub async fn get_minimum_delegation(config: &Config) -> Result<u64, Error> {
```
joncinque, line 38: Ah, this is kinda annoying... but I think you're right, this is the best option with a `ProgramClient`. Much better than adding to that trait

stake-pool/single-pool-cli/src/quarantine.rs
```diff
@@ -0,0 +1,197 @@
...
+        .program_client
+        .get_account(*stake_account_address)
+        .await?
+    {
+        match bincode::deserialize::<StakeState>(&stake_account.data)? {
+            StakeState::Stake(meta, stake) => Ok(Some((meta, stake))),
+            StakeState::Initialized(_) => {
+                Err(format!("Stake account {} is undelegated", stake_account_address).into())
+            }
+            _ => unimplemented!(),
```
joncinque, line 74: Just in case there are zombie accounts, it'd be good to also handle `StakeState::Uninitialized`

stake-pool/single-pool-cli/src/quarantine.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    } else {
+        Ok(None)
+    }
+}
+
+pub async fn blank_stake_instruction(
```
joncinque, line 81: nit:
```suggestion
pub async fn create_uninitialized_stake_account_instruction(
```

stake-pool/single-pool-cli/tests/test.rs
```diff
@@ -0,0 +1,284 @@
...
+    },
+    spl_single_validator_pool::{self as single_pool},
+    spl_token_client::client::{ProgramClient, ProgramRpcClient, ProgramRpcClientSendTransaction},
+    std::{path::PathBuf, process::Command, str::FromStr, sync::Arc},
+    tempfile::NamedTempFile,
+};
+
+// XXX TODO FIXME uhh few things... need to ensure binary is built
+// also i dont know how to advance an epoch on TestValidator?
+// i guess its just running the fill binary? so maybe thats actually impossible?
```
joncinque, line 28: The way we do this now is by setting a very short epoch, ie 32 slots, then wait until the epoch rolls over. Here's the example from the stake bot tests:

* short epochs: https://github.com/solana-foundation/stake-o-matic/blob/9e90e1c86f4ed82a5e6cba617000f49ca54fe762/bot/src/stake_pool_v0.rs#L937-L943
* waiting https://github.com/solana-foundation/stake-o-matic/blob/9e90e1c86f4ed82a5e6cba617000f49ca54fe762/bot/src/rpc_client_utils.rs#L517

stake-pool/single-pool-cli/src/main.rs
```diff
@@ -0,0 +1,710 @@
...
+mod cli;
+use cli::*;
+
+mod output;
+use output::*;
+
+mod quarantine;
+
+#[tokio::main]
+async fn main() -> Result<(), Error> {
```
joncinque, line 37: Wow, this is extremely nice, so short and clear

stake-pool/single-pool-cli/src/main.rs
```diff
@@ -0,0 +1,710 @@
...
+        token.get_associated_token_address(&owner.pubkey())
+    };
+
+    let previous_token_amount = token
+        .get_account_info(&token_account_address)
+        .await?
+        .base
+        .amount;
+
+    // TODO flag to skip the authorize calls and just deposit_stake?
```
joncinque, line 265: eh if you want, i'd make them hidden though to be safe

stake-pool/single-pool-cli/src/main.rs
```diff
@@ -0,0 +1,710 @@
...
+
+    if token_amount > token_account.base.amount {
+        return Err(format!(
+            "Withdraw amount {} exceeds tokens in account ({})",
+            token_amount, token_account.base.amount
+        )
+        .into());
+    }
+
+    // note a delegate authority is not allowed here because we must authorize the pool authority
```
joncinque, line 373: Thanks for clarifying this, I had forgotten the behavior

stake-pool/single-pool-cli/src/main.rs
```diff
@@ -0,0 +1,710 @@
...
+        &stake_authority_address,
+        &token_account_address,
+        &token_authority.pubkey(),
+        token_amount,
+    ));
+
+    let transaction = Transaction::new_signed_with_payer(
+        &instructions,
+        Some(&payer.pubkey()),
+        &vec![payer, token_authority, Arc::new(stake_account)], // TODO dedupe
```
joncinque, line 402: nit: Is it possible to avoid the `Arc` and pass everything with `as_ref()` or `&` ie
```suggestion
        &vec![&payer, &token_authority, &stake_account], // TODO dedupe
```

stake-pool/single-pool-cli/src/main.rs
```diff
@@ -0,0 +1,710 @@
...
+
+// create token metadata
+async fn command_create_metadata(
+    config: &Config,
+    command_config: CreateMetadataCli,
+) -> CommandResult {
+    let payer = config.fee_payer()?;
+
+    // first get the pool address
+    // i dont check metadata because i dont want to get entangled with mpl
```
joncinque, line 435: :pray:

joncinque (commented):
Fly-by first review! It looks really great, most of my comments are nits and minor things. Just to double-check, this does work with ledgers, correct?

stake-pool/single-pool-cli/src/cli.rs
```diff
@@ -0,0 +1,399 @@
...
+
+    /// The vote account corresponding to the pool to deposit into. Optional when stake account or pool is provided
+    #[clap(long = "vote-account", value_parser = Pubkey::from_str)]
+    pub vote_account_address: Option<Pubkey>,
+
+    /// Signing authority on the stake account to be deposited. Defaults to the client keypair
+    #[clap(long = "stake-authority", id = "STAKE_WITHDRAW_AUTHORITY_KEYPAIR", validator = |s| is_valid_signer(s))]
+    pub stake_withdraw_authority: Option<SignerArg>,
+
+    /// The token account to mint to. Defaults to the client keypair's associated token account
```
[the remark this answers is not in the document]
2501babe: it means the default signer, aka whatever is defined in the config. fee payer defaults to this also, but can be overridden, but doesnt affect anything else. the token cli behaves the same

stake-pool/single-pool-cli/src/cli.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub vote_account_address: Pubkey,
+
+    /// Do not create MPL metadata for the pool mint
+    #[clap(long)]
+    pub skip_metadata: bool,
+}
+
+#[derive(Clone, Debug, Args)]
+#[clap(group(ArgGroup::new("stake-source").required(true).args(&["stake-account-address", "default-stake-account"])))]
+#[clap(group(ArgGroup::new("pool-source").args(&["pool-address", "vote-account-address"])))]
```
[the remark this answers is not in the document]
2501babe: not sure theres a way to make it a constant but i put it in a function

stake-pool/single-pool-cli/src/cli.rs
2501babe: oh i have no idea actually. i was reading mpl docs and they just say stuff like "some external URI that can be anything" or "URI pointing to JSON representing the asset." i could add it back as an optional param

stake-pool/single-pool-cli/src/main.rs
2501babe: best i can do is `&vec![&*payer, &*token_authority, &stake_account]` but im not sure thats much better

2501babe:
ok everything marked resolved is done... i think im mostly finished with this, just need to respond to your other comments, and i think i need to change the pubkeys in args to do something similar to `SignerArgs` after looking at the clap utils source code and realizing theres other things people might want to do than supply pubkeys which im unaware of

it should work on ledgers in theory, i pass `wallet_manager` through the right places, but i have a dev device coming so i can test soon

stake-pool/single-pool-cli/src/cli.rs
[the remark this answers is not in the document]
joncinque: That works too!

stake-pool/single-pool-cli/src/cli.rs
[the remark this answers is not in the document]
joncinque: Ah gotcha, that makes sense, thanks!

stake-pool/single-pool-cli/src/main.rs
joncinque: How about `&[payer.as_ref(), token_authority.as_ref(), &stake_account]`? If that doesn't work, we can keep yours

joncinque:
All the changes you've made look good, and your plan sounds good. This is extremely close, just ping me when you want a final check

stake-pool/single-pool-cli/src/cli.rs
buffalojoec: Most of the time it's a pic or a video yep! But technically can be any string, as long as whatever client/UI is rendering it knows what to expecting

stake-pool/single-pool-cli/src/quarantine.rs
2501babe: thanks to doing a code review i learned today theres an rpc call for this in 1.14 so im gonna use that instead. this is gonna be abstracted when i write a stake program client anyway

joncinque: Ah my bad, I forgot that you had an explicit rpc client. I thought everything was going through `ProgramClient`.  In that case, definitely use it!

stake-pool/single-pool-cli/src/cli.rs
2501babe: looking at this again i think the way forward is writing a reduced functionality version that eliminates the four uses of `ArgMatches`. i was gonna kick the can but while writing the issue i decided its actually easy after all lol

it will be in 1.17 when i pr it so the `SignerArg` stuff needs to stay for now but this is definitely temporary

https://github.com/solana-labs/solana/issues/32421

2501babe: this turns out to be nontrivial and maybe not possible, because `Signer` cant require `Send + Sync` because of the c wrapper library necessary for ledger support

joncinque: We might be able to find a way around that by wrapping in `Arc`s or `AtomicPtr`s or something... which type is causing the issue? Is it `RemoteWalletType`?

2501babe: `RemoteKeypair`, have more detail in https://github.com/solana-labs/solana/issues/32421

2501babe:
@joncinque think im done this for now

joncinque (approved):
Great job! I really liked how you did the test env too, gonna keep that in mind for future CLI testing

2501babe:
yea im happy w how the test stuff turned out, "how do we integration test cli binaries" has been an annoyance to me ever since the bug with misisng config file bricking spl-token. next time i mess with this i wanna try sticking test validator in a static global mutex oncecell and see if we can get these running fast

2501babe:
im baffled by this error... i rebased on master to solve merge conflicts on Cargo.lock, and now ci cant build the single pool cli because it claims the multi pool cli has a hard 1.16.3 dependency... but it doesnt, theres no 1.16.3 in any Cargo.toml or Cargo.lock, and all the flows work fine for me on my own machine

2501babe:
well this is a plausible culprit but i need to figure out how its getting that version since the one specified in all these files is 1.16.1...

```
Run ./ci/install-build-deps.sh
  ./ci/install-build-deps.sh
  ./ci/install-program-deps.sh
  echo "$HOME/.local/share/solana/install/active_release/bin" >> $GITHUB_PATH
  shell: /usr/bin/bash -e {0}
  env:
    RUST_STABLE: 1.69.0
    SOLANA_VERSION: v1.16.[3](https://github.com/solana-labs/solana-program-library/actions/runs/5538387718/jobs/10108271659?pr=4510#step:8:3)
```

2501babe:
oh, it looks like we upgraded to 1.16.3 half an hour ago lol so i guess it messed up ci for me, should be fine after another rebase

PR #4544, opened 2023-06-14
Derive macro for TLV discriminators

buffalojoec:
This PR introduces a derive macro for automatically generating the tokens to:
- Implement the `TlvDiscriminator` trait for a struct or enum
- Establish constants for the discriminator bytes and slice

RFC:
- Naming convention of the derive macro: `SplTlv`
- Should we continue to house the constants outside of any trait implementation? I understand this was likely due to the fact it must be manually set (ie. `[u8; 8]`) and can't be derived as a constant. But with a macro, we can do this. But it will require some re-writes of existing constant value handling

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,7 +189,7 @@ version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "726535892e8eae7e70657b4c8ea93d26b8553afb1ce617caee529ef96d7dee6c"
 dependencies = [
- "proc-macro2 1.0.56",
+ "proc-macro2 1.0.60",
  "quote 1.0.26",
  "syn 1.0.107",
  "synstructure",
@@ -201,7 +201,7 @@ version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2777730b2039ac0f95f093556e61b6d26cebed5393ca6f152717777cec3a42ed"
 dependencies = [
- "proc-macro2 1.0.56",
+ "proc-macro2 1.0.60",
  "quote 1.0.26",
  "syn 1.0.107",
 ]
@@ -265,7 +265,7 @@ version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "10f203db73a71dfa2fb6dd22763990fa26f3d2625a6da2da900d23b87d26be27"
 dependencies = [
- "proc-macro2 1.0.56",
+ "proc-macro2 1.0.60",
  "quote 1.0.26",
  "syn 1.0.107",
 ]
@@ -276,7 +276,7 @@ version = "0.1.68"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9ccdd8f2a161be9bd5c023df56f1b2a0bd1d83872ae53b71a84a12c9bf6e842"
 dependencies = [
- "proc-macro2 1.0.56",
+ "proc-macro2 1.0.60",
  "quote 1.0.26",
  "syn 2.0.15",
 ]
@@ -423,7 +423,7 @@ dependencies = [
  "lazy_static",
  "lazycell",
  "peeking_take_while",
- "proc-macro2 1.0.56",
+ "proc-macro2 1.0.60",
  "quote 1.0.26",
  "regex",
  "rustc-hash",
@@ -539,7 +539,7 @@ dependencies = [
  "borsh-derive-internal",
  "borsh-schema-derive-internal",
  "proc-macro-crate 0.1.5",
- "proc-macro2 1.0.56",
+ "proc-macro2 1.0.60",
  "syn 1.0.107",
 ]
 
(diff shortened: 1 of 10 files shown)


joncinque (commented):
I thought about this for a bit, and I realized that I got tripped up between the instruction and TLV discriminators, because the transfer hook interface is a weird case.

In most cases, the interface instructions don't need to implement TlvDiscriminator, since they don't typically belong in TLV structures! It's only in the case of account resolution that we'll want that, since the account resolution library uses the instruction discriminator as the TLV discriminator. In most cases the two are independent.

With that said, here's what I'm thinking:

* move just the discriminator and the rest of this into its own crate for declaring interface instructions. The TLV crate will have its own discriminator type, in order to make the separation as clear as possible.
* the transfer hook interface uses this crate *and also* the TLV crate, but just for the execute instruction
* the token-metadata interface only uses the new crate for defining the instructions

With the word discriminator slice stuff, I had to do it because it's not possible to match on a const function, ie you can't do:

```
match instruction {
    MyInstruction.as_slice() => {}
}
```

And traits do not allow for const functions. However! There's a neat `as_slice()` function which does exactly what we need https://doc.rust-lang.org/std/primitive.array.html#method.as_slice

So I put in a helper to show how this could work without the ugly slice trick at #4547

Let me know how this sounds!

buffalojoec:
> With that said, here's what I'm thinking:
> 
> * move just the discriminator and the rest of this into its own crate for declaring interface instructions. The TLV crate will have its own discriminator type, in order to make the separation as clear as possible.
> * the transfer hook interface uses this crate _and also_ the TLV crate, but just for the execute instruction
> * the token-metadata interface only uses the new crate for defining the instructions

Ok, so we have two kinds of discriminators:
- TLV discriminators
- Interface instruction discriminators

And you want the macro to be specifically for **interface instructions** rather than TLV entries?

It's worth mentioning that technically these macros would be super similar, as they're both declaring an 8-byte discriminator for something. One is for TLV and the other is for an instruction, but maybe a generic "discriminator" derive macro could work for both?

joncinque:
> One is for TLV and the other is for an instruction, but maybe a generic "discriminator" derive macro could work for both?

Maybe we should just keep it simple for now, and have one discriminator type that's reused by the TLV library. I came up with other ways that included multiple derive macros, but it all seemed way too confusing.

> And just for clarity, in the case of Token2022's transfer hook, it would only use the generic "discriminator" derive for it's TLV entry, not the instruction

The transfer-hook is still a weird case, since it uses the same 8 bytes as the instruction discriminator AND the TLV. The concept is that the account resolution struct can store the additional accounts for any number of instructions, so it should do some sort of conversion. But in the end, keeping just one discriminator type seems to make this as simple as possible.

What do you think?

buffalojoec:
> > One is for TLV and the other is for an instruction, but maybe a generic "discriminator" derive macro could work for both?
> 
> Maybe we should just keep it simple for now, and have one discriminator type that's reused by the TLV library. I came up with other ways that included multiple derive macros, but it all seemed way too confusing.
> 
> > And just for clarity, in the case of Token2022's transfer hook, it would only use the generic "discriminator" derive for it's TLV entry, not the instruction
> 
> The transfer-hook is still a weird case, since it uses the same 8 bytes as the instruction discriminator AND the TLV. The concept is that the account resolution struct can store the additional accounts for any number of instructions, so it should do some sort of conversion. But in the end, keeping just one discriminator type seems to make this as simple as possible.
> 
> What do you think?

Macro land can suck you in pretty fast :) but yeah, I think that makes sense. Let's just do a simple library that turns a string into an 8-byte discriminator. You basically specify the namespace as so:

```rust
#[derive(SplDiscriminator)]
#[namespace("global:my_object")]
pub struct MyObject {}
```

and get:

```rust
/* Auto-generated */
impl SplDiscriminator for MyObject {
   const DISCRIMINATOR: Discriminator = Discriminator::new([..]) // macro-generated bytes
}
```

and then if you're also using TLV, you could just manually write:

```rust
impl TlvDiscriminator for MyObject {
    const TLV_DISCRIMINATOR: Discriminator = Self::DISCRIMINATOR;
}
```

So, for someone using TLV they'd do:

```rust
#[derive(SplDiscriminator)]
#[namespace("global:my_instruction")]
pub struct MyObject {}

impl TlvDiscriminator for MyObject {
    const TLV_DISCRIMINATOR: Discriminator = Self::DISCRIMINATOR;
}
```

The main benefit here is for non-TLV users it's a handy way to build a discriminator deterministically from some string literal, and for TLV users it's eliminating the need to hard-code the bytes, even if you do still have to write the trait implementation.

I see the `SplDiscriminator` being a tiny standalone crate, which TLV would then depend on.

wdyt?

joncinque:
That makes sense to me! Just one little nit, @ngundotra mentioned the potential issue with using the variable name `DISCRIMINATOR` at https://github.com/solana-labs/solana-program-library/pull/4137#pullrequestreview-1402152936 -- so maybe we say `SPL_DISCRIMINATOR` to be safe?

buffalojoec:
> That makes sense to me! Just one little nit, @ngundotra mentioned the potential issue with using the variable name `DISCRIMINATOR` at [#4137 (review)](https://github.com/solana-labs/solana-program-library/pull/4137#pullrequestreview-1402152936) -- so maybe we say `SPL_DISCRIMINATOR` to be safe?

sg, will submit a new PR for this lib. Can close this one 💪🏼

PR #4546, opened 2023-06-14
token-metadata: Implement example program

joncinque:
#### Problem

#4514 introduced an interface for token-metadata programs, but there's no program that implements the interface.

#### Solution

Add that program! I ended up tweaking some bits of the interface while implementing this. Most of the PR is made up of tests, and I found the processor implementation to look quite neat with the right helpers. Fingers crossed that the token-2022 implementation will also go smoothly...

cc @austbot

### DIFF
--- a/libraries/type-length-value/src/state.rs
+++ b/libraries/type-length-value/src/state.rs
@@ -411,7 +411,7 @@ impl<'a> TlvState for TlvStateMut<'a> {
 /// Packs a borsh-serializable value into an existing TLV space, reallocating
 /// the account and TLV as needed to accommodate for any change in space
 #[cfg(feature = "borsh")]
-pub fn realloc_and_borsh_serialize<V: TlvDiscriminator + borsh::BorshSerialize>(
+pub fn realloc_and_borsh_serialize<V: SplDiscriminate + borsh::BorshSerialize>(
     account_info: &solana_program::account_info::AccountInfo,
     value: &V,
 ) -> Result<(), ProgramError> {
@@ -421,7 +421,7 @@ pub fn realloc_and_borsh_serialize<V: TlvDiscriminator + borsh::BorshSerialize>(
             type_start: _,
             length_start,
             value_start,
-        } = get_indices(&data, V::TLV_DISCRIMINATOR, false)?;
+        } = get_indices(&data, V::SPL_DISCRIMINATOR, false)?;
         usize::try_from(*pod_from_bytes::<Length>(&data[length_start..value_start])?)?
     };
     let new_length = solana_program::borsh::get_instance_packed_len(&value)?;
--- a/libraries/type-length-value/src/state.rs
+++ b/libraries/type-length-value/src/state.rs
@@ -408,6 +408,54 @@ impl<'a> TlvState for TlvStateMut<'a> {
     }
 }
 
+/// Packs a borsh-serializable value into an existing TLV space, reallocating
+/// the account and TLV as needed to accommodate for any change in space
+#[cfg(feature = "borsh")]
+pub fn realloc_and_borsh_serialize<V: TlvDiscriminator + borsh::BorshSerialize>(
+    account_info: &solana_program::account_info::AccountInfo,
+    value: &V,
+) -> Result<(), ProgramError> {
+    let previous_length = {
+        let data = account_info.try_borrow_data()?;
+        let TlvIndices {
+            type_start: _,
+            length_start,
+            value_start,
+        } = get_indices(&data, V::TLV_DISCRIMINATOR, false)?;
+        usize::try_from(*pod_from_bytes::<Length>(&data[length_start..value_start])?)?
+    };
+    let new_length = solana_program::borsh::get_instance_packed_len(&value)?;
+    let previous_account_size = account_info.try_data_len()?;
+    if previous_length < new_length {
+        // size increased, so realloc the account, then the TLV entry, then write data
+        let additional_bytes = new_length
+            .checked_sub(previous_length)
+            .ok_or(ProgramError::AccountDataTooSmall)?;
+        account_info.realloc(previous_account_size.saturating_add(additional_bytes), true)?;
+        let mut buffer = account_info.try_borrow_mut_data()?;
+        let mut state = TlvStateMut::unpack(&mut buffer)?;
+        state.realloc::<V>(new_length)?;
+        state.borsh_serialize(value)?;
+    } else {
+        // do it backwards otherwise, write the state, realloc TLV, then the account
+        let mut buffer = account_info.try_borrow_mut_data()?;
+        let mut state = TlvStateMut::unpack(&mut buffer)?;
+        state.borsh_serialize(value)?;
+        let removed_bytes = previous_length
+            .checked_sub(new_length)
+            .ok_or(ProgramError::AccountDataTooSmall)?;
+        if removed_bytes > 0 {
+            // we decreased the size, so need to realloc the TLV, then the account
+            state.realloc::<V>(new_length)?;
+            // this is probably fine, but be safe and avoid invalidating references
+            drop(state);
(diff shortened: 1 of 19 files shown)


libraries/type-length-value/src/state.rs
```diff
@@ -408,6 +408,54 @@ impl<'a> TlvState for TlvStateMut<'a> {
     }
 }
 
+/// Packs a borsh-serializable value into an existing TLV space, reallocating
+/// the account and TLV as needed to accommodate for any change in space
+#[cfg(feature = "borsh")]
+pub fn realloc_and_borsh_serialize<V: TlvDiscriminator + borsh::BorshSerialize>(
+    account_info: &solana_program::account_info::AccountInfo,
```
buffalojoec, line 415: Should this not just operate over a slice of bytes, rather than an `AccountInfo`? 
All of the other methods in the lib don't use any structs from `solana-program`, which allows this library to be used generically over any buffers

I see you're offering some ease of use for realloc, but that may not belong in this library

libraries/type-length-value/src/state.rs
```diff
@@ -408,6 +408,54 @@ impl<'a> TlvState for TlvStateMut<'a> {
...
+    if previous_length < new_length {
+        // size increased, so realloc the account, then the TLV entry, then write data
+        let additional_bytes = new_length
+            .checked_sub(previous_length)
+            .ok_or(ProgramError::AccountDataTooSmall)?;
+        account_info.realloc(previous_account_size.saturating_add(additional_bytes), true)?;
+        let mut buffer = account_info.try_borrow_mut_data()?;
+        let mut state = TlvStateMut::unpack(&mut buffer)?;
+        state.realloc::<V>(new_length)?;
+        state.borsh_serialize(value)?;
```
buffalojoec, line 438: You're going to need to transfer lamports in or out based on the new size (increased size will error if no new lamports are added for rent)

Seems like you're putting the onus on the developer to take this step, which also bleeds into my point in my other comment about this maybe not being necessary to do with `AccountInfo` directly

token-metadata/example/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+#![allow(clippy::integer_arithmetic)]
+#![deny(missing_docs)]
+#![cfg_attr(not(test), forbid(unsafe_code))]
+
+pub mod processor;
+
+#[cfg(not(feature = "no-entrypoint"))]
+mod entrypoint;
+
+// Export current sdk types for downstream users building with a different sdk version
```
buffalojoec, line 12: Is this necessary for an example program?

token-metadata/example/src/processor.rs
```diff
@@ -0,0 +1,220 @@
...
+    data: Initialize,
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+
+    let metadata_info = next_account_info(account_info_iter)?;
+    let update_authority_info = next_account_info(account_info_iter)?;
+    let mint_info = next_account_info(account_info_iter)?;
+    let mint_authority_info = next_account_info(account_info_iter)?;
+
+    // scope the mint authority check, in case the mint is in the same account!
```
buffalojoec, line 55: Asking for clarification: Mint authority and mint in the same account you mean?

token-metadata/example/src/processor.rs
```diff
@@ -0,0 +1,220 @@
...
+        update_authority,
+        mint: *mint_info.key,
+        ..Default::default()
+    };
+    let instance_size = get_instance_packed_len(&token_metadata)?;
+
+    // allocate a TLV entry for the space and write it in
+    let mut buffer = metadata_info.try_borrow_mut_data()?;
+    let mut state = TlvStateMut::unpack(&mut buffer)?;
+    state.alloc::<TokenMetadata>(instance_size)?;
```
buffalojoec, line 86: This is slick 🔥

token-metadata/example/src/processor.rs
```diff
@@ -0,0 +1,220 @@
...
+    let mut token_metadata = {
+        let buffer = metadata_info.try_borrow_data()?;
+        let state = TlvStateBorrowed::unpack(&buffer)?;
+        state.borsh_deserialize::<TokenMetadata>()?
+    };
+
+    check_update_authority(update_authority_info, &token_metadata.update_authority)?;
+
+    // Update the field
+    token_metadata.update(data.field, data.value);
```
buffalojoec, line 113: 🔥 🔥

token-metadata/example/src/processor.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    // Update / realloc the account
+    realloc_and_borsh_serialize(metadata_info, &token_metadata)?;
```
buffalojoec, line 116: Curious how this doesn't also require a transfer for additional rent? Maybe I'm missing something

token-metadata/example/tests/program_test.rs
```diff
@@ -0,0 +1,167 @@
...
+    value: String,
+) {
+    let rent = context.banks_client.get_rent().await.unwrap();
+    let old_space = token_metadata.tlv_size_of().unwrap();
+    let old_rent_lamports = rent.minimum_balance(old_space);
+
+    token_metadata.update(field.clone(), value.clone());
+
+    let new_space = token_metadata.tlv_size_of().unwrap();
+    let new_rent_lamports = rent.minimum_balance(new_space);
```
buffalojoec, line 141: Ahhhh, there it is!

I think this is less intuitive to do on client-side, but willing to hear thoughts.

I've typically handled realloc additional rent within the program itself, which I guess someone could still do, but curious if it's something we want to include in the example program

buffalojoec (commented):
lgtm Jon, just had a couple nits and my main feedback was around handling reallocation with transferring of additional rent. 

You're doing it client-side, I made a beef about doing it program-side, want to hear yer thoughts

libraries/type-length-value/src/state.rs
joncinque: I see this as a common pattern for any program that wants to keep its accounts at the minimum required size. Most of this library assumes fixed-size types defined through `Pod`, so this is definitely a different situation. You're correct -- even the borsh examples just operate on slices of bytes.

I originally had this in the token-metadata program, but since the order of operations is tricky ie. do I realloc TLV first or the underlying account? And you don't want to get this wrong, or you can potentially clobber your TLV and render your account unusable.

Since this is generic for any borsh types, we can provide a working implementation for any program to use. Do you have any ideas of where it could go?

libraries/type-length-value/src/state.rs
joncinque: That's correct, I don't want to include payers anywhere in interfaces, since that provides easy attack opportunities for programs that implement the interface.

For best use, a program should implement some `withdraw_excess_lamports` interface for the right authority to claim additional lamports, and the tools need to transfer the additional required lamports as a separate instruction, as demonstrated in the tests.

If you have a better idea of where this could go, I'm all ears! I definitely took the easy way out here. We can add a `borsh.rs` or `program.rs` file with borsh / solana program-specific utilities?

token-metadata/example/src/processor.rs
joncinque: yep! people do wacky things

token-metadata/example/tests/program_test.rs
buffalojoec: Yep, that's fair. Interfaces forcing me to abandon habits.

> A program should never require a payer and system program

This is good to know

libraries/type-length-value/src/state.rs
buffalojoec: I'll see if I can come up with a better home for borsh/`AccountInfo` stuff, but for now I think it's cool to merge. I can address later in another PR

buffalojoec (dismissed):

PR #4547, opened 2023-06-15
TLV: Add const `as_slice()` function for easier matching

joncinque:
#### Problem

As noted in #4544, there's some ugliness around matching discriminators, since you can only match on const slices.

#### Solution

Thanks to the const `as_slice()` function on arrays at https://doc.rust-lang.org/std/primitive.array.html#method.as_slice -- we can make this way neater!


buffalojoec (approved):
Squeaky clean 🧼

PR #4549, opened 2023-06-15
token-2022: Add metadata pointer extension

joncinque:
#### Problem

As the world of token-metadata potentially fractures between many different programs, it will be impossible to know which metadata address is correct.

#### Solution

Store the correct metadata address directly in a mint extension. A wallet just needs to fetch the mint, read out the metadata address, and then fetch that address.

Note that there is *NO* functionality with this address, and it's just storing two addresses -- the update authority, and the metadata account address.

Most of this is copy-pasted from the transfer hook program id management, and it's meant to be extremely simple.

cc @austbot and @jnwng

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -43,6 +43,7 @@ use spl_token_2022::{
         default_account_state::DefaultAccountState,
         interest_bearing_mint::InterestBearingConfig,
         memo_transfer::MemoTransfer,
+        metadata_pointer::MetadataPointer,
         mint_close_authority::MintCloseAuthority,
         permanent_delegate::PermanentDelegate,
         transfer_fee::{TransferFeeAmount, TransferFeeConfig},
@@ -783,6 +784,7 @@ async fn command_authorize(
         AuthorityType::ConfidentialTransferFeeConfig => {
             "confidential transfer fee config authority"
         }
+        AuthorityType::MetadataPointer => "metadata pointer authority",
     };
 
     let (mint_pubkey, previous_authority) = if !config.sign_only {
@@ -884,6 +886,16 @@ async fn command_authorize(
                         ))
                     }
                 }
+                AuthorityType::MetadataPointer => {
+                    if let Ok(extension) = mint.get_extension::<MetadataPointer>() {
+                        Ok(COption::<Pubkey>::from(extension.authority))
+                    } else {
+                        Err(format!(
+                            "Mint `{}` does not support a metadata pointer",
+                            account
+                        ))
+                    }
+                }
             }?;
 
             Ok((account, previous_authority))
@@ -920,7 +932,8 @@ async fn command_authorize(
                 | AuthorityType::PermanentDelegate
                 | AuthorityType::ConfidentialTransferMint
                 | AuthorityType::TransferHookProgramId
-                | AuthorityType::ConfidentialTransferFeeConfig => Err(format!(
+                | AuthorityType::ConfidentialTransferFeeConfig
+                | AuthorityType::MetadataPointer => Err(format!(
                     "Authority type `{}` not supported for SPL Token accounts",
                     auth_str
                 )),
@@ -3884,6 +3897,8 @@ async fn process_command<'a>(
                 "permanent-delegate" => AuthorityType::PermanentDelegate,
                 "confidential-transfer-mint" => AuthorityType::ConfidentialTransferMint,
                 "transfer-hook-program-id" => AuthorityType::TransferHookProgramId,
+                "confidential-transfer-fee" => AuthorityType::ConfidentialTransferFeeConfig,
+                "metadata-pointer" => AuthorityType::MetadataPointer,
                 _ => unreachable!(),
             };
 
--- a/token/program-2022/src/extension/metadata_pointer/instruction.rs
+++ b/token/program-2022/src/extension/metadata_pointer/instruction.rs
@@ -1,0 +1,122 @@
+use {
+    crate::{
+        check_program_account,
+        instruction::{encode_instruction, TokenInstruction},
+        pod::OptionalNonZeroPubkey,
+    },
+    bytemuck::{Pod, Zeroable},
+    num_enum::{IntoPrimitive, TryFromPrimitive},
+    solana_program::{
+        instruction::{AccountMeta, Instruction},
+        program_error::ProgramError,
+        pubkey::Pubkey,
+    },
+    std::convert::TryInto,
+};
+
+/// Metadata pointer extension instructions
+#[derive(Clone, Copy, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum MetadataPointerInstruction {
+    /// Initialize a new mint with a metadata pointer
+    ///
+    /// Fails if the mint has already been initialized, so must be called before
+    /// `InitializeMint`.
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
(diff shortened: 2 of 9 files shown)


token/program-2022/src/extension/metadata_pointer/instruction.rs
```diff
@@ -0,0 +1,122 @@
...
+    Ok(encode_instruction(
+        token_program_id,
+        accounts,
+        TokenInstruction::MetadataPointerExtension,
+        MetadataPointerInstruction::Update,
+        &UpdateInstructionData {
+            metadata_address: metadata_address.try_into()?,
+        },
+    ))
+}
```
ngundotra, line 122: Looks good!

token/program-2022/src/extension/metadata_pointer/instruction.rs
```diff
@@ -0,0 +1,122 @@
...
+    ///   * Multisignature authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[]` The mint's metadata pointer authority.
+    ///   2. ..2+M `[signer]` M signer accounts.
+    ///
+    /// Data expected by this instruction:
+    ///   `crate::extension::metadata_pointer::instruction::UpdateInstructionData`
+    ///
+    Update,
+}
```
ngundotra, line 56: I think we _may_ end-up accidentally kneecapping collections via this interface, but otherwise looks good!

We can always ship a new interface for Metadata forks with collection mgmt

cc @austbot

token/cli/src/main.rs
```diff
@@ -3884,6 +3897,8 @@ async fn process_command<'a>(
                 "permanent-delegate" => AuthorityType::PermanentDelegate,
                 "confidential-transfer-mint" => AuthorityType::ConfidentialTransferMint,
                 "transfer-hook-program-id" => AuthorityType::TransferHookProgramId,
+                "confidential-transfer-fee" => AuthorityType::ConfidentialTransferFeeConfig,
```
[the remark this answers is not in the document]
joncinque: Correct, doing a solid for @samkim-crypto

token/program-2022/src/extension/metadata_pointer/instruction.rs
joncinque: Sorry, I don't know too much about collections -- how would this affect them? Are you thinking of a situation in which some mints in a collection get their metadata changed so they're no longer part of a collection?

joncinque:
> looks like there is nothing stopping me from pointing to the same metadata as someone else this is bad ideally the bytes are in the mint account not external

That's correct, which is why the metadata account must store the mint address that it's describing. When they both point to each other, then you know it's legit. and putting the metadata bytes in the account is the next PR, don't worry. This gives flexibility for any other metadata program to work too

token/program-2022/src/extension/metadata_pointer/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+pub mod processor;
+
+/// Metadata pointer extension data for mints.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct MetadataPointer {
+    /// Authority that can set the metadata address
+    pub authority: OptionalNonZeroPubkey,
+    /// Account address that holds the metadata
+    pub metadata_address: OptionalNonZeroPubkey,
```
buffalojoec, line 21: Assuming this is optional so you can nullify it later?
Would that mean if you initialize a metadata pointer and then null it later, you're still stuck with the 65 bytes?

token/program-2022/src/extension/metadata_pointer/instruction.rs
buffalojoec: Personally I think it's worth fleshing out your concern @ngundotra - I don't like the sound of two metadata interfaces right out of the gate, only one of which is forkable

ngundotra: Metaplex's Token Metadata has the ability to set & verify collections of NFTs (grouping ID + some collection level metadata) but this current Metadata interface doesn't have any setup for that.

I think it's fine to push this out + then see what people naturally do to create collections/grouping on top.

We might get positively surprised シ

joncinque: I dug through the Metaplex code, and here's how it currently works:

* there's a collection parent mint, with its own metadata
* child mints that are part of this collection have a pointer in their metadata to the parent *mint* address
* child mints are verified by checking that the parent mint's metadata update authority signs a transaction

Depending on what we want, there's a few ways to go about it.

We can add a "collection" interface, which does the following:

* init / update collection parent -- sets in an authority to an account
* add / remove child -- requires collection parent to sign, writes or unsets the collection parent address in the child

You just need two structs: `CollectionParent` and `CollectionChild`.

```
struct CollectionParent {
    authority: Pubkey,
    /// any other fields that you might want, maybe all of the children pubkeys
}

struct CollectionChild {
    parent: Pubkey
}
```

Note that there's *nothing* about tokens or metadata in here. A program that implements just the token metadata interface can tack on these TLV structures for parent and child, and it'll work very similarly to metaplex. Rather than writing the parent mint address though, it'll write the parent metadata address.

BUT, and here's where this gets cool: token-2022 can implement this interface too. And in that case, everything will point to mints.

What do you think?

buffalojoec: So you'd just add the collection pointer into the metadata (or Token2022's mint) the same way you're kind of doing the metadata pointer rn?

> child mints are verified by checking that the parent mint's metadata update authority signs a transaction

And the metadata interface still has an update authority, so this should still be possible for those implementing the collection interface, so that's good.

I think the tricky part here is that MPL Token Metadata owns all the PDA accounts housing metadata/collection data, so there's more flexibility across data.
More specifically, MPL-TM can verify collection members like so:
- You create metadata and point at the collection, saying you are a part of the collection (unverified)
- You send an instruction to MPL-TM (`VerifyCollection`), and it validates the update authority's signature and **changes the data field in the metadata to verified** - see [Metadata](https://docs.rs/mpl-token-metadata/latest/mpl_token_metadata/state/struct.Metadata.html) and [Collection](https://docs.rs/mpl-token-metadata/latest/mpl_token_metadata/state/struct.Collection.html) (from collection field in Metadata)

So you can kind of think of MPL-TM as a program that implements _both_ the `token-metadata` and `token-collection`(or whatever) interfaces.
With that in mind, it's completely possible to replicate MPL-TM collection-metadata mapping **if your program implements both interfaces**. 
However, if you are using one program's collection standard and another program's metadata standard, then what?

joncinque: Ah sorry I said this part wrong:

> child mints are verified by checking that the parent mint's metadata update authority signs a transaction

Should be the parent mint's "collection authority", and that's it, since there's no overlap between metadata and collections with this interface. The two concepts are *totally* distinct.

>  More specifically, MPL-TM can verify collection members like so:

We don't need to worry about verified and unverified, since you must sign with the parent authority in order to write the collection pubkey in the child. Every mint is immediately verified.

Unless there's a workflow where someone wants to add a mint to a collection as unverified. Or where unverified token mints are important. We can add a `verified` bool in the Child struct if needed though, no problem!

>  However, if you are using one program's collection standard and another program's metadata standard, then what?

Then you're a metadata program like metaplex's token-metadata, right? Maybe I'm missing something more profound to this question, sorry.

The whole point is that you can have just a metadata program, you can have just a collection program, and you can have both -- it's all interfaces!

buffalojoec (approved):

PR #4555, opened 2023-06-16
Fix downstream dependencies for `spl-program-error`

buffalojoec:
This PR fixes the requirement for crates that wish to use `spl-program-error` to have to import `thiserror`, `num-derive`, and `num-traits`.

Note: This will NOT require these crates to be present in the Cargo.toml file to use the **procedural macro attribute** `#[spl_program_error]`, however, if you decide to use any of the **derive macros** you'll need to import these crates.

### DIFF
--- a/libraries/program-error/derive/src/macro_impl.rs
+++ b/libraries/program-error/derive/src/macro_impl.rs
@@ -104,6 +104,7 @@ pub fn spl_program_error(input: ItemEnum) -> proc_macro2::TokenStream {
     let print_program_error = print_program_error(ident, variants);
     quote! {
         #[derive(Clone, Debug, Eq, thiserror::Error, num_derive::FromPrimitive, PartialEq)]
+        #[num_traits = "num_traits"]
         #input
 
         #into_program_error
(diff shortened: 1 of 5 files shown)


libraries/program-error/derive/src/macro_impl.rs
```diff
@@ -104,6 +104,7 @@ pub fn spl_program_error(input: ItemEnum) -> proc_macro2::TokenStream {
     let print_program_error = print_program_error(ident, variants);
     quote! {
         #[derive(Clone, Debug, Eq, thiserror::Error, num_derive::FromPrimitive, PartialEq)]
+        #[num_traits = "num_traits"]
```
joncinque, line 107: Ah very clever, this was the bit that I missed, nice job!

joncinque (approved):
perfect!

libraries/program-error/derive/src/macro_impl.rs
buffalojoec: Turns out I just had to read their docs 😭

buffalojoec:
Ran a rebase from `master`, looks like didn't really even need to

joncinque (approved):

PR #4556, opened 2023-06-16
`spl discriminator` crate

buffalojoec:
This PR resolves conversations from #4544 

I've introduced a small crate for just discriminators - `spl-discriminator` - and any other crates previously using the `Discriminator` struct from the `type-length-value` now depend on this crate.

`spl-discriminator` houses the `Discriminator` struct, a new trait `SplDiscriminator` for constants, and the `SplDiscriminator` derive macro.

The derive macro simply generates the trait implementation for `SplDiscriminator` from some string literal.

RFC:
- I chose to demonstrate how we could make the `TlvDiscriminator` trait restricted to only types that implement `SplDiscriminator`, however this is just a spec - I can roll this back out, but seems like maybe a good thing to do.
- I also at one point made the `spl-discriminator` crate available downstream from `spl-type-length-value`, but left it out for now. If so desired, we could set these crates up so that anyone using TLV (`spl-type-length-value`) would not need to import both crates.


buffalojoec:
debugging CI

libraries/type-length-value/src/state.rs
```diff
@@ -2,16 +2,25 @@
...
 use {
     crate::{
-        discriminator::{Discriminator, TlvDiscriminator},
         error::TlvError,
         length::Length,
         pod::{pod_from_bytes, pod_from_bytes_mut},
     },
     bytemuck::Pod,
     solana_program::program_error::ProgramError,
+    spl_discriminator::{Discriminator, SplDiscriminator},
```
joncinque, line 11: Seeing these two imports together confuses me a bit. Like, what's the difference between `Discriminator` and `SplDiscriminator`? I have to read back and remember which is the type and which is the trait. Maybe we can rename `SplDiscriminator` to something like `SplTypeDiscriminator` since it's meant to discriminate between concrete types. What do you think?

token-metadata/interface/src/instruction.rs
```diff
@@ -200,23 +169,23 @@ impl TokenMetadataInstruction {
         }
         let (discriminator, rest) = input.split_at(Discriminator::LENGTH);
         Ok(match discriminator {
-            INITIALIZE_DISCRIMINATOR_SLICE => {
+            Initialize::TLV_DISCRIMINATOR_SLICE => {
```
joncinque, line 172: Oh man, this is gorgeous

token-metadata/interface/src/instruction.rs
```diff
@@ -34,80 +36,47 @@ pub struct Initialize {
...
 }
-impl TlvDiscriminator for Initialize {
-    /// Please use this discriminator in your program when matching
-    const TLV_DISCRIMINATOR: Discriminator = Discriminator::new(INITIALIZE_DISCRIMINATOR);
-}
-/// First 8 bytes of `hash::hashv(&["spl_token_metadata_interface:initialize_account"])`
-const INITIALIZE_DISCRIMINATOR: [u8; Discriminator::LENGTH] = [210, 225, 30, 162, 88, 184, 77, 141];
-// annoying, but needed to perform a match on the value
-const INITIALIZE_DISCRIMINATOR_SLICE: &[u8] = &INITIALIZE_DISCRIMINATOR;
+impl TlvDiscriminator for Initialize {}
```
joncinque, line 39: We can just totally remove the `impl TlvDiscriminator for ...` bits since we know they're not necessary, unless you prefer that I do it in a separate PR

joncinque (commented):
Really great work! I mostly have naming nits / bikeshedding

libraries/discriminator/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+### The `SplDiscriminator` Derive Macro
+
+The `SplDiscriminator` derive macro is a particularly useful tool for those who wish to derive their 8-byte discriminator from a particular string literal. Typically, you would have to run a hash function against the string literal, then copy the first 8 bytes, and then hard-code those bytes into a statement like the one above.
+
+Instead, you can simply annotate a struct or enum with `SplDiscriminator` and provide a **namespace** via the `discriminator_namespace` attribute, and the macro will automatically derive the 8-byte discriminator for you!
+
+```rust
+#[derive(SplDiscriminator)] // Implements `SplDiscriminator` for your struct/enum using your declared string literal namespace
+#[discriminator_namespace("some_discriminator_namespace")]
```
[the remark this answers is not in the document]
buffalojoec: OK gotcha, `namespace` is a bit confusing. Let's go with `discriminator_hash_input` since it's clearly describing what it's for

token-metadata/interface/src/instruction.rs
buffalojoec: So just completely eliminate TLV-specific discriminators, and use only the `SPL_DISCRIMINATOR`?

libraries/discriminator/src/discriminator.rs
```diff
@@ -1,17 +1,14 @@
...
-pub trait TlvDiscriminator {
-    /// Associated value type discriminator, checked at the start of TLV entries
-    const TLV_DISCRIMINATOR: Discriminator;
+/// A trait for managing 8-byte discriminators in a slab of bytes
+pub trait SplDiscriminator {
+    /// The 8-byte discriminator as a `[u8; 8]`
+    const SPL_DISCRIMINATOR: Discriminator;
+    /// The 8-byte discriminator as a slice (`&[u8]`)
+    const SPL_DISCRIMINATOR_SLICE: &'static [u8] = Self::SPL_DISCRIMINATOR.as_slice();
 }
```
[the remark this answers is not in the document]
buffalojoec: I'll give this a rip

libraries/discriminator/src/discriminator.rs
```diff
@@ -1,17 +1,14 @@
...
-//! Type-Length-Value structure. Since the word "type" is reserved in Rust,
-//! we use the term "Discriminator" and "Type" interchangeably.
+//! The traits and types used to create a discriminator for a type
 
-use {
-    bytemuck::{Pod, Zeroable},
-    solana_program::program_error::ProgramError,
-};
+use bytemuck::{Pod, Zeroable};
+use solana_program::program_error::ProgramError;
```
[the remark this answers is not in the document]
buffalojoec: Am I doing something wrong, or does the `toolchain.toml` restrict you from using these configs with 1.60 stable?

libraries/discriminator/derive/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+
+#![deny(missing_docs)]
+#![cfg_attr(not(test), forbid(unsafe_code))]
+
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use quote::ToTokens;
+use spl_discriminator_syn::SplDiscriminatorBuilder;
+use syn::parse_macro_input;
```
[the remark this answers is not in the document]
buffalojoec: I had a feeling this was coming soon lol

libraries/discriminator/src/discriminator.rs
```diff
@@ -1,17 +1,14 @@
...
-    const TLV_DISCRIMINATOR: Discriminator;
+/// A trait for managing 8-byte discriminators in a slab of bytes
+pub trait SplDiscriminator {
+    /// The 8-byte discriminator as a `[u8; 8]`
+    const SPL_DISCRIMINATOR: Discriminator;
+    /// The 8-byte discriminator as a slice (`&[u8]`)
+    const SPL_DISCRIMINATOR_SLICE: &'static [u8] = Self::SPL_DISCRIMINATOR.as_slice();
 }
 
 /// Discriminator used as the type in the TLV structure
```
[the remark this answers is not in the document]
buffalojoec: I'm down for this, but that would mean we'd also just need a config arg for the macro to tell it how many bytes to use, right?

libraries/type-length-value/src/state.rs
buffalojoec: I guess this seems off-putting because of the way we're using the traits combined with `TlvDiscriminator`. Ideally, if you just want a discriminator and don't care about the traits, you can just use the `Discriminator` struct on its own. Which directly relates to your other comment of providing a `new_with_hash_input` function.

The tricky part there is it's not going to be a constant function, so the macro will still have to do the hashing as it currently is, however that still doesn't require a `solana-program` import

libraries/discriminator/src/discriminator.rs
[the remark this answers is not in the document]
buffalojoec: Looked into this a little, but marking resolved since you mentioned a new PR

libraries/type-length-value/src/state.rs
buffalojoec: Thoughts on `HasSplDiscriminator`? or just `HasDiscriminator`?

buffalojoec:
The only hold up here (right now) is I am flummoxed by some dependency issues with `build-sbf` or `test-sbf`.

It seems like there's some issue with either `proc-macro` or `syn` -related dependencies when it comes to `rustc 1.60` so I'm currently debugging that & losing [more] hair

Screenshot 2023-06-20 at 1 33 22 PM

libraries/discriminator/src/discriminator.rs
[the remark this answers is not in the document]
joncinque: ah yeah, you need nightly to use it. So in the monorepo, you do `./cargo nightly fmt --all` from the top-level

libraries/type-length-value/src/state.rs
joncinque: I looked up for any naming conventions for Rust traits, and nothing really came up except for https://github.com/rust-lang/api-guidelines/discussions/28 which would lean toward `Discriminate` as the imperative form, which is pretty hilarious.

And yet, I can't come up with anything better than `SplDiscriminate`. I'm not a fan of `Has*` or `With*` since it's not really canonical anywhere else. Or we can go with `SplDiscriminates`, similar to https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html

On another side, maybe we can rename `Discriminator` to `ArrayDiscriminator` to make it clearer that it's a concrete type?

buffalojoec: I like `ArrayDiscriminator` yep.
I wanted to make a joke about `*Discriminates` but the more I say/look at it the more it kinda makes sense, but it still does have some obvious connotations

buffalojoec: I think I'm going with `SplDiscriminates`

buffalojoec:
> The only hold up here (right now) is I am flummoxed by some dependency issues with `build-sbf` or `test-sbf`.

Looks like you can't have `proc-macro2` as a direct dependency in a crate that isn't a `proc-macro` type if you're building to `cdylib` (I think?)

CI problem solved.

token-metadata/interface/src/instruction.rs
buffalojoec: Done

libraries/discriminator/src/discriminator.rs
```diff
@@ -0,0 +1,72 @@
+//! The traits and types used to create a discriminator for a type
+
+use {
+    bytemuck::{Pod, Zeroable},
+    solana_program::{hash, program_error::ProgramError},
+};
+
+/// A trait for managing 8-byte discriminators in a slab of bytes
+pub trait SplDiscriminates {
```
joncinque, line 9: nit: sorry, I'm leaning more towards `SplDiscriminate` in the end. I can't find any examples in the rust std lib using a non-imperative verb.

libraries/discriminator/src/discriminator.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+    /// Creates a new `ArrayDiscriminator` from some hash input string literal
+    pub fn new_with_hash_input(hash_input: &str) -> Self {
+        let hash_bytes = hash::hashv(&[hash_input.as_bytes()]).to_bytes();
+        let mut discriminator_bytes = [0u8; 8];
+        discriminator_bytes.copy_from_slice(&hash_bytes[..8]);
+        Self(discriminator_bytes)
+    }
+    /// Returns the discriminator length using the common `len()` method
+    pub fn len(&self) -> usize {
```
joncinque, line 41: Maybe I'm missing something, but why is this needed? It's an array, so it doesn't really have a length

joncinque (commented):
Just two last bits then we should be good to go!

libraries/discriminator/src/discriminator.rs
buffalojoec: Just for the common nomenclature of calling `len()` on it instead of `ArrayDiscriminator::LENGTH`, and we chatted about opening this up for variable discriminator lengths, so this sets the stage for that.

If you think it's unnecessary I can remove

libraries/discriminator/src/discriminator.rs
buffalojoec: 😭

joncinque: Please feel free to convince me otherwise!

libraries/discriminator/src/discriminator.rs
buffalojoec: Yeah sure! Assuming `len()` is more common pattern for things with variable size that can grow/shrink, like vectors? 
Our "variable length" discriminator is still fixed once initialized

buffalojoec: Done deal, removed

joncinque (approved):
Great job, thanks for dealing with the back and forth!

PR #4557, opened 2023-06-16
Upgrade solana to 1.16.1

samkim-crypto:
- Upgrading solana to version 1.16.1
- Upgrading rust version to 1.69.0
- Updating cargo audit to the latest version 1.17.6
- Temporarily disabling build warnings in the ci due to deprecation of `Pubkey::new`

PR #4558, opened 2023-06-16
Upgrade rust version to 1.69.0 and cargo audit to 1.17.6

samkim-crypto:
- Update rust version to 1.69.0 that the mono uses
- Update `cargo audit` version to the latest version 1.17.6


samkim-crypto:
Seems like rust version upgrade has to be made in the same PR as the solana version upgrade to prevent the CI from failing. Closing this PR and replacing it with https://github.com/solana-labs/solana-program-library/pull/4557.

PR #4582, opened 2023-06-20
token-2022: Sync native balance during realloc

joncinque:
#### Problem

As kindly pointed out by @ryoqun, when reallocating a native account, we don't update the rent amount stored in `is_native`, meaning that the rent exempt reserve falls out of sync.

#### Solution

During realloc, also update the rent-exempt reserve in native accounts.

### DIFF
--- a/token/program-2022-test/tests/reallocate.rs
+++ b/token/program-2022-test/tests/reallocate.rs
@@ -196,16 +196,18 @@ async fn reallocate_without_current_extension_knowledge() {
     );
 }
 
-#[test_case(&[ExtensionType::CpiGuard], true, true ; "transfer lamports and sync with extension")]
-#[test_case(&[ExtensionType::CpiGuard], true, false ; "transfer lamports only with extension")]
-#[test_case(&[ExtensionType::CpiGuard], false, false ; "do not transfer, with extension")]
-#[test_case(&[], true, true ; "transfer lamports and sync without extension")]
-#[test_case(&[], true, false ; "transfer lamports only without extension")]
-#[test_case(&[], false, false ; "do not transfer, without extension")]
+#[test_case(&[ExtensionType::CpiGuard], 1_000_000_000, true ; "transfer more than new rent and sync")]
+#[test_case(&[ExtensionType::CpiGuard], 1_000_000_000, false ; "transfer more than new rent")]
+#[test_case(&[ExtensionType::CpiGuard], 1, true ; "transfer less than new rent and sync")]
+#[test_case(&[ExtensionType::CpiGuard], 1, false ; "transfer less than new rent")]
+#[test_case(&[ExtensionType::CpiGuard], 0, false ; "no transfer with extension")]
+#[test_case(&[], 1_000_000_000, true ; "transfer lamports and sync without extension")]
+#[test_case(&[], 1_000_000_000, false ; "transfer lamports without extension")]
+#[test_case(&[], 0, false ; "no transfer without extension")]
 #[tokio::test]
 async fn reallocate_updates_native_rent_exemption(
     extensions: &[ExtensionType],
-    transfer_lamports: bool,
+    transfer_lamports: u64,
     sync_native: bool,
 ) {
     let mut context = TestContext::new().await;
@@ -221,13 +223,12 @@ async fn reallocate_updates_native_rent_exemption(
     let alice_account = alice_account.pubkey();
 
     // transfer more lamports
-    let transfer_amount = 1_000_000_000;
-    if transfer_lamports {
+    if transfer_lamports > 0 {
         let mut context = context.lock().await;
         let instructions = vec![system_instruction::transfer(
             &context.payer.pubkey(),
             &alice_account,
-            transfer_amount,
+            transfer_lamports,
         )];
         let tx = Transaction::new_signed_with_payer(
             &instructions,
@@ -245,11 +246,7 @@ async fn reallocate_updates_native_rent_exemption(
     if sync_native {
         token.sync_native(&alice_account).await.unwrap();
         let account_info = token.get_account_info(&alice_account).await.unwrap();
-        if transfer_lamports {
-            assert_eq!(account_info.base.amount, transfer_amount);
-        } else {
-            assert_eq!(account_info.base.amount, 0);
-        }
+        assert_eq!(account_info.base.amount, transfer_lamports);
     }
 
     let token_account = token.get_account_info(&alice_account).await.unwrap();
--- a/token/program-2022-test/tests/reallocate.rs
+++ b/token/program-2022-test/tests/reallocate.rs
@@ -196,11 +196,18 @@ async fn reallocate_without_current_extension_knowledge() {
     );
 }
 
-#[test_case(true, true ; "transfer lamports and sync")]
-#[test_case(true, false ; "transfer lamports only")]
-#[test_case(false, false ; "do not transfer")]
+#[test_case(&[ExtensionType::CpiGuard], true, true ; "transfer lamports and sync with extension")]
+#[test_case(&[ExtensionType::CpiGuard], true, false ; "transfer lamports only with extension")]
+#[test_case(&[ExtensionType::CpiGuard], false, false ; "do not transfer, with extension")]
+#[test_case(&[], true, true ; "transfer lamports and sync without extension")]
+#[test_case(&[], true, false ; "transfer lamports only without extension")]
+#[test_case(&[], false, false ; "do not transfer, without extension")]
 #[tokio::test]
-async fn reallocate_syncs_native(transfer_lamports: bool, sync_native: bool) {
+async fn reallocate_updates_native_rent_exemption(
+    extensions: &[ExtensionType],
+    transfer_lamports: bool,
+    sync_native: bool,
+) {
     let mut context = TestContext::new().await;
     context.init_token_with_native_mint().await.unwrap();
     let TokenContext { token, alice, .. } = context.token_context.unwrap();
@@ -251,19 +258,14 @@ async fn reallocate_syncs_native(transfer_lamports: bool, sync_native: bool) {
 
     // reallocate resizes account to accommodate new extension
     token
-        .reallocate(
-            &alice_account,
-            &alice.pubkey(),
-            &[ExtensionType::CpiGuard],
-            &[&alice],
-        )
+        .reallocate(&alice_account, &alice.pubkey(), extensions, &[&alice])
         .await
         .unwrap();
 
     let account = token.get_account(alice_account).await.unwrap();
     assert_eq!(
         account.data.len(),
-        ExtensionType::get_account_len::<Account>(&[ExtensionType::CpiGuard])
+        ExtensionType::get_account_len::<Account>(extensions)
     );
     let expected_rent_exempt_reserve = {
         let mut context = context.lock().await;
@@ -273,24 +275,13 @@ async fn reallocate_syncs_native(transfer_lamports: bool, sync_native: bool) {
     let token_account = token.get_account_info(&alice_account).await.unwrap();
     let post_amount = token_account.base.amount;
     let post_rent_exempt_reserve = token_account.base.is_native.unwrap();
-    if transfer_lamports && !sync_native {
-        let rent_diff = expected_rent_exempt_reserve
-            .checked_sub(pre_rent_exempt_reserve)
-            .unwrap();
-        // if we didn't sync and transferred lamports, the extra required rent
-        // exemption lamports are taken from the unsynced lamports
-        assert_eq!(
-            post_amount,
-            pre_amount
-                .checked_add(transfer_amount)
-                .and_then(|x| x.checked_sub(rent_diff))
-                .unwrap()
-        );
-    } else {
-        // amount of lamports should be totally unchanged otherwise
-        assert_eq!(pre_amount, post_amount);
-    }
+    // amount of lamports should be totally unchanged
+    assert_eq!(pre_amount, post_amount);
     // but rent exempt reserve should change
     assert_eq!(post_rent_exempt_reserve, expected_rent_exempt_reserve);
-    assert!(pre_rent_exempt_reserve < post_rent_exempt_reserve);
+    if extensions.is_empty() {
+        assert_eq!(pre_rent_exempt_reserve, post_rent_exempt_reserve);
+    } else {
+        assert!(pre_rent_exempt_reserve < post_rent_exempt_reserve);
+    }
 }
--- a/token/program-2022-test/tests/reallocate.rs
+++ b/token/program-2022-test/tests/reallocate.rs
@@ -5,12 +5,19 @@ use {
     program_test::{TestContext, TokenContext},
     solana_program_test::tokio,
     solana_sdk::{
-        instruction::InstructionError, program_option::COption, pubkey::Pubkey, signature::Signer,
-        signer::keypair::Keypair, transaction::TransactionError, transport::TransportError,
+        instruction::InstructionError,
+        program_option::COption,
+        pubkey::Pubkey,
+        signature::Signer,
+        signer::keypair::Keypair,
+        system_instruction,
+        transaction::{Transaction, TransactionError},
+        transport::TransportError,
     },
     spl_token_2022::{error::TokenError, extension::ExtensionType, state::Account},
     spl_token_client::token::{ExtensionInitializationParams, TokenError as TokenClientError},
     std::convert::TryInto,
+    test_case::test_case,
 };
 
 #[tokio::test]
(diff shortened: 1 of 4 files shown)


token/program-2022/src/extension/reallocate.rs
```diff
@@ -83,8 +84,23 @@ pub fn process_reallocate(
...
 
     // unpack to set account_type, if needed
-    let mut token_account = token_account_info.data.borrow_mut();
-    set_account_type::<Account>(&mut token_account)?;
+    let mut token_account_data = token_account_info.data.borrow_mut();
+    set_account_type::<Account>(&mut token_account_data)?;
+
+    // sync the rent exempt reserve for native accounts
+    let mut token_account = StateWithExtensionsMut::<Account>::unpack(&mut token_account_data)?;
+    if token_account.base.is_native.is_some() {
```
ryoqun, line 92: nit: seems `token_account.base.is_native()` is more idiomatic across codebase.

token/program-2022-test/tests/reallocate.rs
```diff
@@ -188,3 +188,52 @@ async fn reallocate_without_current_extension_knowledge() {
...
+    let alice_account = Keypair::new();
+    token
+        .create_auxiliary_token_account(&alice_account, &alice.pubkey())
+        .await
+        .unwrap();
+    let alice_account = alice_account.pubkey();
+
+    let token_account = token.get_account_info(&alice_account).await.unwrap();
+    let pre_amount = token_account.base.amount;
+    let pre_is_native = token_account.base.is_native.unwrap();
```
CriesofCarrots, line 208:
```suggestion
    let pre_rent_exempt_reserve = token_account.base.is_native.unwrap();
```

token/program-2022-test/tests/reallocate.rs
```diff
@@ -188,3 +188,52 @@ async fn reallocate_without_current_extension_knowledge() {
...
+        ExtensionType::get_account_len::<Account>(&[ExtensionType::CpiGuard])
+    );
+    let rent_exempt_reserve = {
+        let mut context = context.lock().await;
+        let rent = context.banks_client.get_rent().await.unwrap();
+        rent.minimum_balance(account.data.len())
+    };
+    let token_account = token.get_account_info(&alice_account).await.unwrap();
+    let post_amount = token_account.base.amount;
+    let post_is_native = token_account.base.is_native.unwrap();
```
CriesofCarrots, line 233:
```suggestion
    let post_rent_exempt_reserve = token_account.base.is_native.unwrap();
```

token/program-2022-test/tests/reallocate.rs
```diff
@@ -188,3 +188,52 @@ async fn reallocate_without_current_extension_knowledge() {
...
+        )
+        .await
+        .unwrap();
+
+    let account = token.get_account(alice_account).await.unwrap();
+    assert_eq!(
+        account.data.len(),
+        ExtensionType::get_account_len::<Account>(&[ExtensionType::CpiGuard])
+    );
+    let rent_exempt_reserve = {
```
CriesofCarrots, line 226:
```suggestion
    let expected_rent_exempt_reserve = {
```

token/program-2022-test/tests/reallocate.rs
```diff
@@ -188,3 +188,52 @@ async fn reallocate_without_current_extension_knowledge() {
...
+    // reallocate resizes account to accommodate new extension
+    token
+        .reallocate(
+            &alice_account,
+            &alice.pubkey(),
+            &[ExtensionType::CpiGuard],
+            &[&alice],
+        )
+        .await
+        .unwrap();
```
CriesofCarrots, line 219: A case added to this test where `token_account.base.amount > 0` always fails on this unwrap.

token/program-2022/src/extension/reallocate.rs
```diff
@@ -83,8 +84,23 @@ pub fn process_reallocate(
...
+        let new_amount = token_account_info
+            .lamports()
+            .checked_sub(new_minimum_balance)
+            .ok_or(TokenError::Overflow)?;
+        if new_amount < token_account.base.amount {
```
CriesofCarrots, line 97: Currently, we only progress past this check if the native account is "empty", ie. only holds rent-exempt reserve and base amount = 0. I assume that is not your intent, is it?

To fix, I think we need to primarily edit L76, but I can't suggest there. Something like:
```
    let lamports_to_be_subtracted = if let Some(amount) = native_amount {
        token_account_info.lamports().saturating_sub(amount)
    } else {
        token_account_info.lamports()
    };
    let lamports_diff = new_minimum_balance.saturating_sub(lamports_to_be_subtracted);
```
where `native_amount` is returned in the block on L37 so we don't have to unpack the account again.
Something like:
```
    let (mut current_extension_types, native_amount) = {
        let token_account = token_account_info.data.borrow();
        let account = StateWithExtensions::<Account>::unpack(&token_account)?;
        Processor::validate_owner(
            program_id,
            &account.base.owner,
            authority_info,
            authority_info_data_len,
            account_info_iter.as_slice(),
        )?;
        let native_amount = if account.base.is_native() {
            Some(account.base.amount)
        } else {
            None
        };
        (account.get_extension_types()?, native_amount)
    };
```

token/program-2022/src/extension/reallocate.rs
CriesofCarrots: +1

CriesofCarrots (changes requested):
Unfortunately, this only works for accounts where `base.amount == 0`, which I assume is not the intent.

(Also, oof and thank you for spotting this, @ryoqun !)

token/program-2022-test/tests/reallocate.rs
CriesofCarrots: (Sorry, several of these are just nits to make the test more readable. Take 'em or leave 'em)

token/program-2022/src/extension/reallocate.rs
```diff
@@ -83,8 +84,23 @@ pub fn process_reallocate(
...
+            return Err(TokenError::InvalidState.into());
+        }
```
ryoqun, line 99: hmm, i think this cause unwanted error for this edge case:

1. prepare native token account without extensions (account data size: 165) with `token_account_info.lamports() == 2_081_040` (= rent exempt minimum for account data size: 171). so `token_account.base.amount == 41_760` (`2_081_040` - `2_039_280`).
2. call `reallocate(..., &[ExtensionType::CpiGuard], ...)`, this reallocates to be +6 more bytes: account data size: 171
3. so this `if` evaluates to `if 0 < 41_760` and this triggers error. i think this instead succeed.

token/program-2022/src/extension/reallocate.rs
joncinque: Ah right of course! Thanks for noticing this issue. In my head I thought it would just transfer the right amount of lamports in, which is totally wrong.

This seems like the most secure approach. We could also use the value stored in `is_native` when calculating the amount of lamports to transfer, but I think we'll be safer by not trusting that always.

token/program-2022/src/extension/reallocate.rs
joncinque: Yeah that's true, I was thinking they should do a "sync native" beforehand, but there's really no reason to

joncinque:
I integrated all the fixes that you requested with more complete test coverage. Thanks for your careful reviews!

token/program-2022-test/tests/reallocate.rs
joncinque: Added the test case separately

token/program-2022/src/extension/reallocate.rs
CriesofCarrots: So one thing that seems a little weird to me now is that as-written, the `reallocate_syncs_native::transfer_lamports_only` magically syncs the additional lamports so that post `base.amount = 999958240`. But if you call the same transfer+reallocate with no extensions (empty slice), there is no sync and the post `base.amount` is still 0.

Maybe this is the behavior we want, but how about a test case that makes this clear?
...I'm realizing we should have had tests that cover realloc with empty slice in the first place.

It would also be nice to have test cases that cover transfers smaller than the diff amount needed for the extension.

joncinque: That's a good point -- we can skip the sync native bit, and only update the rent exempt reserve. What do you think?

joncinque: I got convinced that was better, and added some empty tests too.

token/program-2022/src/extension/reallocate.rs
```diff
@@ -72,19 +74,42 @@ pub fn process_reallocate(
...
-    set_account_type::<Account>(&mut token_account)?;
+    let mut token_account_data = token_account_info.data.borrow_mut();
+    set_account_type::<Account>(&mut token_account_data)?;
+
+    // sync the rent exempt reserve for native accounts
+    let mut token_account = StateWithExtensionsMut::<Account>::unpack(&mut token_account_data)?;
+    if token_account.base.is_native() {
+        // sanity check that there are enough lamports to cover the token amount
+        // and the rent exempt reserve
+        let native_amount = native_amount.ok_or(TokenError::InvalidState)?;
```
CriesofCarrots, line 103: These 2 lines are technically redundant, since we've already checked `is_native()` to construct `native_amount`.
Could do:
```suggestion
    if let Some(native_amount) = native_amount {
        // sanity check that there are enough lamports to cover the token amount
        // and the rent exempt reserve
```
I feel ambivalent about it.

token/program-2022-test/tests/reallocate.rs
```diff
@@ -188,3 +195,93 @@ async fn reallocate_without_current_extension_knowledge() {
...
+
+    let alice_account = Keypair::new();
+    token
+        .create_auxiliary_token_account(&alice_account, &alice.pubkey())
+        .await
+        .unwrap();
+    let alice_account = alice_account.pubkey();
+
+    // transfer more lamports
+    let transfer_amount = 1_000_000_000;
```
CriesofCarrots, line 224: What did you think about ensuring the `ExtensionType::CpiGuard` cases still work when 0 < this amount < 41760?

CriesofCarrots (commented):
> I got convinced that was better

Great, I'm happy with this behavior too! Last thoughts from me

token/program-2022/src/extension/reallocate.rs
joncinque: Ah that is neater, thanks! I just didn't do an unwrap to avoid a conversation with the auditor

token/program-2022-test/tests/reallocate.rs
joncinque: oh yes, sorry, i'll add that too

joncinque:
Ok this should finally be ready! Thanks for your patience

CriesofCarrots (approved):
lgtm! Thanks for all the work on these tests :)

CriesofCarrots:
...dependent on passing CI, of course :)

token/program-2022/src/extension/reallocate.rs
```diff
@@ -72,19 +74,41 @@ pub fn process_reallocate(
...
+    let mut token_account_data = token_account_info.data.borrow_mut();
+    set_account_type::<Account>(&mut token_account_data)?;
+
+    // sync the rent exempt reserve for native accounts
+    let mut token_account = StateWithExtensionsMut::<Account>::unpack(&mut token_account_data)?;
+    if let Some(native_amount) = native_amount {
+        // sanity check that there are enough lamports to cover the token amount
+        // and the rent exempt reserve
+        let lamports_in_account_data = native_amount
+            .checked_add(new_minimum_balance)
```
ryoqun, line 104: nit: `new_minimum_balance.checked_add(native_amount)` for consistency with L82's `.saturating_sub(...)`.

token/program-2022/src/extension/reallocate.rs
```diff
@@ -33,7 +34,7 @@ pub fn process_reallocate(
     let authority_info_data_len = authority_info.data_len();
 
     // check that account is the right type and validate owner
-    let mut current_extension_types = {
+    let (mut current_extension_types, native_amount) = {
```
ryoqun, line 37: naming: how about `native_token_amount`? it's a bit verbose but less ambiguous (imo) and follows widely-used terminology of `token_amount` by further qualifying it.

token/program-2022/src/extension/reallocate.rs
```diff
@@ -72,19 +74,41 @@ pub fn process_reallocate(
...
+    }
 
     // unpack to set account_type, if needed
-    let mut token_account = token_account_info.data.borrow_mut();
-    set_account_type::<Account>(&mut token_account)?;
+    let mut token_account_data = token_account_info.data.borrow_mut();
+    set_account_type::<Account>(&mut token_account_data)?;
+
+    // sync the rent exempt reserve for native accounts
+    let mut token_account = StateWithExtensionsMut::<Account>::unpack(&mut token_account_data)?;
```
ryoqun, line 99: better of putting inside the following `if` to save CU? Otherwise, this extra `.unpack()` CU is imposed for all of mints (not alone native mint)

token/program-2022/src/extension/reallocate.rs
joncinque: The `saturating_sub` makes sense because you could have too many lamports in your account, but I did a `checked_add` here in the weird case where someone overallocates a huge account and also has a lot of lamports, but you're right, it's fine (and simpler) to saturate.

token/program-2022/src/extension/reallocate.rs
joncinque: ah good call, thanks

token/program-2022/src/extension/reallocate.rs
joncinque: Sure, done

token/program-2022/src/extension/reallocate.rs
```diff
@@ -72,19 +74,41 @@ pub fn process_reallocate(
     // if additional lamports needed to remain rent-exempt, transfer them
     let rent = Rent::get()?;
     let new_minimum_balance = rent.minimum_balance(needed_account_len);
```
ryoqun, line 76: naming: `let new_rent_exempt_reserve =` (let's be consistent with corresponding code in `_process_initialize_account()`).

token/program-2022/src/extension/reallocate.rs
```diff
@@ -72,19 +74,41 @@ pub fn process_reallocate(
...
-    set_account_type::<Account>(&mut token_account)?;
+    let mut token_account_data = token_account_info.data.borrow_mut();
+    set_account_type::<Account>(&mut token_account_data)?;
+
+    // sync the rent exempt reserve for native accounts
+    let mut token_account = StateWithExtensionsMut::<Account>::unpack(&mut token_account_data)?;
+    if let Some(native_amount) = native_amount {
+        // sanity check that there are enough lamports to cover the token amount
+        // and the rent exempt reserve
+        let lamports_in_account_data = native_amount
```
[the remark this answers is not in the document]
joncinque: I was trying to communicate "these are the lamports that the account thinks it has" but "minimum_lamports" is definitely clearer, thanks

token/program-2022/src/extension/reallocate.rs
joncinque: done

token/program-2022/src/extension/reallocate.rs
joncinque: Ah ok, that makes way more sense! I went back to checked and changed the order

token/program-2022/src/extension/reallocate.rs
ryoqun: yeah, i think so. actually syncing is kind of _transfer_ and we would need to account for existence of CpiGuard extension in very broad sense. by not syncing, we're avoiding altogether this semantic war. :)

token/program-2022/src/extension/reallocate.rs
```diff
@@ -71,20 +73,42 @@ pub fn process_reallocate(
...
 
-    // unpack to set account_type, if needed
-    let mut token_account = token_account_info.data.borrow_mut();
-    set_account_type::<Account>(&mut token_account)?;
+    let current_lamport_reserve = token_account_info
+        .lamports()
+        .checked_sub(native_token_amount.unwrap_or(0))
+        .ok_or(TokenError::Overflow)?;
+    let lamports_diff = new_rent_exempt_reserve.saturating_sub(current_lamport_reserve);
+    if lamports_diff > 0 {
```
ryoqun, line 83: after some long thinking, i concluded conditionally omitting `invoke(transfer)` is safe to save CUs. this allows anything to be passed in as `payer_info` and `system_program_info` if reallocated account is funded adequately, while not documented so.

That safeness is based on the fact callers shouldn't rely on signedness of `payer` indirectly via internal workings of `reallocate` in this specific IX use case. In contract, `system_program::transfer` do check full-validation even for the 0-lamport transfer case.

ryoqun (approved):
lgtm

thanks for being patient with bunch of review back and forth.

in the end, i'm very happy that we settled on very sensible semantics around reallocating native token and added bunch of tests on this revisit of this ix.

token/program-2022/src/extension/reallocate.rs
joncinque: That was my thinking too, I'm glad that you agree!

PR #4583, opened 2023-06-21
feat: StateWithExtensionsOwned impl Clone for easier client usage

Arrowana:


joncinque (approved):
Makes sense, thanks for your contribution! Gonna merge on red since the audit fix has been done, and the relevant CI tests passed

PR #4589, opened 2023-06-21
single-pool: Inline and remove MPL dependency

joncinque:
#### Problem

As noted in #4588, because of circular dependencies between SPL and Metaplex Token Metadata, the SPL can't upgrade to Solana 1.16.

#### Solution

This goes one of the lazy routes, and directly inlines the bits needed for metadata support in the single validator stake pool program. The inline mpl file is exactly the same as in #4588 , since I couldn't think of a better place to put it.


2501babe:
lgtm but if you wanted to rebase on #4588 to eliminate the inline file as discussed thatd be preferred

2501babe (dismissed):

joncinque:
Done, rebased and symlinked, which also allows us to completely remove the mpl crates.

PR #4594, opened 2023-06-22
token-js: created calculateFees function

dhruvdabhi101:
Problem : 
The rust side provides a nifty way to calculate a transfer fee using `calculate_fee` , but wallet for JS Clients cannot use it.
Solution: 
Added `calculateFee` function for JS Client.


token/js/src/extensions/transferFee/state.ts
```diff
@@ -52,6 +65,22 @@ export const TransferFeeConfigLayout = struct<TransferFeeConfig>([
...
+export function getEpochFee(transferFeeConfig: TransferFeeConfig, epoch: bigint): TransferFee {
+    if (epoch >= transferFeeConfig.newerTransferFee.epoch) {
+        return transferFeeConfig.newerTransferFee;
+    } else {
+        return transferFeeConfig.olderTransferFee;
+    }
+}
+
+/** Calculate the fee for the given epoch and input amount */
+export function calculateEpochFee(transferFeeConfig: TransferFeeConfig, preFeeAmount: bigint): bigint {
```
joncinque, line 78: Be sure to pass the epoch in here!

joncinque (commented):
The guts of it look good! Please add a new unit test file for that in https://github.com/solana-labs/solana-program-library/tree/master/token/js/test/unit

token/js/src/extensions/transferFee/state.ts
dhruvdabhi101: yeah. Thanks for pointing that out.

token/js/src/extensions/transferFee/state.ts
```diff
@@ -52,6 +65,21 @@ export const TransferFeeConfigLayout = struct<TransferFeeConfig>([
...
+export function getEpochFee(transferFeeConfig: TransferFeeConfig, epoch: bigint): TransferFee {
+    if (epoch >= transferFeeConfig.newerTransferFee.epoch) {
+        return transferFeeConfig.newerTransferFee;
+    } else {
+        return transferFeeConfig.olderTransferFee;
+    }
+}
+
+/** Calculate the fee for the given epoch and input amount */
+export function calculateEpochFee(transferFeeConfig: TransferFeeConfig, preFeeAmount: bigint, epoch: bigint): bigint {
```
joncinque, line 78: nit for consistency with the rust side, let's keep the parameters in the same order
```suggestion
export function calculateEpochFee(transferFeeConfig: TransferFeeConfig, epoch: bigint, preFeeAmount: bigint): bigint {
```

token/js/src/extensions/transferFee/state.ts
```diff
@@ -41,6 +41,19 @@ export function transferFeeLayout(property?: string): Layout<TransferFee> {
...
 }
 
+/** Calculate the transfer fee */
+export function calculateFee(transferFee: TransferFee, preFeeAmount: bigint): bigint {
+    const transferFeeBasisPoints = transferFee.transferFeeBasisPoints;
+    if (transferFeeBasisPoints === 0 || preFeeAmount === BigInt(0)) {
+        return BigInt(0);
+    } else {
+        const numerator = preFeeAmount * BigInt(transferFeeBasisPoints);
+        const rawFee = numerator / ONE_IN_BASIS_POINTS;
```
joncinque, line 51: Sorry, I'm just now realizing that this is incorrect. We do a "ceiling" division in the rust side, so that it always gives the "ceiling" result, and not the floor like this code is doing. To get the ceiling, you can do:
```suggestion
        const rawFee = (numerator + ONE_IN_BASIS_POINTS - 1) / ONE_IN_BASIS_POINTS;
```

See https://github.com/solana-labs/solana-program-library/blob/2a2b2a763b0847ebe6a028016b14d04d95bf655d/token/program-2022/src/extension/transfer_fee/mod.rs#L38-L62 for more info

Note that this change will modify the results from your tests

token/js/test/unit/transferfee.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    it('should calculate the fee correctly when preFeeAmount is non-zero', () => {
+        const transferFee = {
+            epoch: 1n,
+            maximumFee: 100n,
+            transferFeeBasisPoints: 50,
+        };
+        const preFeeAmount = 500n;
+        const fee = calculateFee(transferFee, preFeeAmount);
+        expect(fee).to.eql(2n);
```
joncinque, line 36: This should become `3n` with the math change I outlined earlier

joncinque (commented):
Just a couple more comments, but this looking very close!

dhruvdabhi101:
> Just a couple more comments, but this looking very close!

yeah, i'm sorry I am new to the open source so i sometimes make mistakes.

joncinque (commented):
Just one last question, then this should be good to go!

Thanks for being so responsive, and don't worry about being new to open-source, I really appreciate your time and effort!

token/js/test/unit/transferfee.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        const fee = calculateEpochFee(transferFeeConfig, epoch, preFeeAmount);
+        expect(fee).to.eql(2n);
+    });
+
+    it('should calculate the fee correctly for newerTransferFee', () => {
+        const preFeeAmount = 300n;
+        const epoch = 2n;
+        const fee = calculateEpochFee(transferFeeConfig, epoch, preFeeAmount);
+        expect(fee).to.eql(3n);
+    });
```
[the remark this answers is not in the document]
dhruvdabhi101: I will remove them in next commit just now

joncinque (approved):
This is good to go, thanks for your contribution and your patience!

PR #4592, opened 2023-06-22
Update Solana to 1.16.1 and Rust to 1.69

joncinque:
#### Problem

The 1.16 Solana crates are out, but SPL is still on 1.14 and needs the new functionality.

#### Solution

Update the:

* rust version to 1.69
* nightly to 2023-04-19
* solana tools to 1.16.1
* solana crates to 1.16.1
* borsh to 0.10

And fix:

* new clippy warnings
* deprecated warnings in the new solana crates

This may take some time to usher through CI, so marking as a draft

PR #4595, opened 2023-06-23
single-pool: support new 1.16 vote account

### LINKED ISSUE #4593: single-pool: Consider relaxing the vote account discriminator check
#### Problem

The single-validator stake pool program uses the withdraw authority on the vote account to validate updates to the token metadata.

This works fine currently, but the vote state is getting an update in 1.16: https://github.com/solana-labs/solana/blob/f62d2318863f6385bcbee0691edc87ff643a4108/sdk/program/src/vote/state/vote_state_versions.rs#L7, which causes the deserialization to fail.

This was discovered while upgrading to 1.16 in #4592, and fixed there by also checking for the number "2" at https://github.com/solana-labs/solana-program-library/blob/3bb3b907dfd82207f71a740259f9fc39be1d85d1/stake-pool/single-pool/src/processor.rs#L227

This could be an issue going forward if any new vote state versions are added, causing the single pool program to fail.

#### Solution

Maybe we can relax the check and pass for any number greater than 0. I'm not convinced this is the best way, which is why I created an issue.

cc @2501babe

2501babe:
closes #4593

### DIFF
--- a/stake-pool/single-pool/src/processor.rs
+++ b/stake-pool/single-pool/src/processor.rs
@@ -224,7 +224,7 @@ fn check_vote_account(vote_account_info: &AccountInfo) -> Result<(), ProgramErro
         .ok_or(SinglePoolError::UnparseableVoteAccount)?;
 
     match u32::from_le_bytes(state_variant) {
-        1 => Ok(()),
+        1 | 2 => Ok(()),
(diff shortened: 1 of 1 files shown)


joncinque (approved):
You didn't ask for a review, but I'd like to merge this to unblock #4592, so here's a fly-by approval!

PR #4601, opened 2023-06-26
`spl-discriminator` generic arg support

buffalojoec:
This PR adds `borsh` AND generic support for the `#[derive(SplDiscriminate)]` macro.

You can now apply this macro to structs/enums with:
- One lifetime specifier: `MyItem<'a>`
- Multiple lifetime specifiers: `MyItem<'a, 'b>`
- Generic type arguments: `MyItem<T>`
- Multiple generic type arguments: `MyItem<U, V>`
- Generic types with trait constraints (where clauses): `MyItem<T> where T: Copy`
- *Any combination of the above

I've also implemented `BorshDeserialize` and `BorshSerialize` for the `ArrayDiscriminator` struct, so one can now use this array of 8 bytes in any types that use `BorshDeserialize` or `BorshSerialize`.

Lastly, added a couple conversions for `ArrayDiscriminator` to go _into_ a `[u8; 8]` and _into_ a `u64`

### DIFF
--- a/libraries/discriminator/Cargo.toml
+++ b/libraries/discriminator/Cargo.toml
@@ -7,7 +7,11 @@ repository = "https://github.com/solana-labs/solana-program-library"
 license = "Apache-2.0"
 edition = "2021"
 
+[features]
+borsh = ["dep:borsh"]
+
 [dependencies]
+borsh = { version = "0.10", optional = true }
 bytemuck = { version = "1.13.1", features = ["derive"] }
 solana-program = "1.16.1"
 spl-discriminator-derive = { version = "0.1.0", path = "./derive" }
--- a/libraries/discriminator/src/discriminator.rs
+++ b/libraries/discriminator/src/discriminator.rs
@@ -14,6 +14,10 @@ pub trait SplDiscriminate {
 }
 
 /// Array Discriminator type
+#[cfg_attr(
+    feature = "borsh",
+    derive(borsh::BorshSerialize, borsh::BorshDeserialize)
+)]
 #[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
 #[repr(transparent)]
 pub struct ArrayDiscriminator([u8; ArrayDiscriminator::LENGTH]);
@@ -64,5 +68,15 @@ impl TryFrom<&[u8]> for ArrayDiscriminator {
         <[u8; Self::LENGTH]>::try_from(a)
             .map(Self::from)
             .map_err(|_| ProgramError::InvalidAccountData)
+    }
+}
+impl From<ArrayDiscriminator> for [u8; 8] {
+    fn from(from: ArrayDiscriminator) -> Self {
+        from.0
+    }
+}
+impl From<ArrayDiscriminator> for u64 {
+    fn from(from: ArrayDiscriminator) -> Self {
+        u64::from_le_bytes(from.0)
     }
 }
--- a/libraries/discriminator/src/lib.rs
+++ b/libraries/discriminator/src/lib.rs
@@ -21,7 +21,7 @@ mod tests {
 
     #[allow(dead_code)]
     #[derive(SplDiscriminate)]
-    #[discriminator_hash_input("some_discriminator_hash_input")]
+    #[discriminator_hash_input("my_first_instruction")]
     pub struct MyInstruction1 {
         arg1: String,
         arg2: u8,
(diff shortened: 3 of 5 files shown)


libraries/discriminator/src/lib.rs
```diff
@@ -21,34 +21,81 @@ mod tests {
...
+        U: Clone + Copy,
+        V: Clone + Copy,
+    {
+        data1: &'c [U],
+        data2: &'c [V],
     }
 
     fn assert_discriminator<T: spl_discriminator::discriminator::SplDiscriminate>(
         hash_input: &str,
+        case_num: u8,
```
joncinque, line 84: nit: this `case_num` will be very brittle to maintain -- why not just print `hash_input` on error?

libraries/discriminator/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = "Apache-2.0"
 edition = "2021"
 
 [dependencies]
+borsh = "0.10"
```
joncinque, line 11: In general, borsh adds a lot of bloat in the final binary, so we should avoid deriving it unnecessarily.

Even though features can be hard to reason about, let's featurize the borsh bits for now, similar to `type-length-value`.

libraries/discriminator/src/lib.rs
```diff
@@ -21,34 +21,81 @@ mod tests {
...
-        Three,
+    #[discriminator_hash_input("global:my_instruction_with_lifetime")]
+    pub struct MyInstruction3<'a> {
+        data: &'a [u8],
+    }
+
+    #[allow(dead_code)]
+    #[derive(SplDiscriminate)]
+    #[discriminator_hash_input("global:my_instruction_with_one_generic")]
+    pub struct MyInstruction4<T> {
```
joncinque, line 49: Is there a particular use case for a generic? I worry that it could cause more confusion than anything else. If I'm defining an instruction or struct in an interface, I probably want the type to be well-defined, otherwise people could deserialize or interpret it incorrectly.

On the flipside, since it isn't a huge lift to add, I'm thinking we can include it and see how people deal with it.

joncinque (commented):
Looks great for the most part! Just a few questions

libraries/discriminator/src/lib.rs
buffalojoec: Yeah, I felt like it was more important to cover our bases where someone might need a discriminator for some use case where their struct has a generic field, and all of a sudden the macro fails bc they have the arg.
Keep in mind - even though the discriminators we're using here are going to be primarily for interfaces - we are trying to de-couple this lib from just that particular use case

joncinque: It kinda goes against the point of a discriminator though. A discriminator is meant to be a unique tag for a type. With the ability to define it on a generic type, `MyInstruction<u8>` and `MyInstruction<String>` have the same discriminator, when they are different concrete types. What can any code do with that? But there's probably some other use case that I'm missing.

libraries/discriminator/src/discriminator.rs
```diff
@@ -79,3 +76,26 @@ impl From<ArrayDiscriminator> for u64 {
...
+impl borsh::ser::BorshSerialize for ArrayDiscriminator
+where [u8; 8]: borsh::ser::BorshSerialize
+{
+    fn serialize<W: borsh::maybestd::io::Write>(
+        &self,
+        writer: &mut W,
+    ) -> ::core::result::Result<(), borsh::maybestd::io::Error> {
+        borsh::BorshSerialize::serialize(&self.0, writer)
+    }
+}
```
joncinque, line 101: It'll be easier to do `#[cfg_attr(feature = "borsh", derive(BorshSerialize, BorshDeserialize))]` rather than deriving these by hand

https://stackoverflow.com/questions/42046327/conditionally-derive-based-on-feature-flag

buffalojoec: oh nice! Didn't know you could do this

libraries/discriminator/src/lib.rs
buffalojoec: So you think by default we shouldn't support generic types, then? Only lifetimes?
One use case that comes to mind for me is something like a TLV type with one generic field.

Maybe you're packing a buffer with TLV types and want to be able to flop between a few different types of data for one field, ie:

```
MyTlvData<T> {
   name: String,
   data: T,
}
```

Imagine even moreso if the type `T` is _another_ TLV struct (and can flop between different ones for nested TLV)

I see your point, but in my head the above struct is still the same data, especially if it's trait-scoped pretty well.

Ultimately my thought was leave it up to the user to decide if a generic should be used based on your uniqueness arguments

buffalojoec: Maybe I'm thinking they shouldn't have unique discriminators and sounds like you think they def should (I'm not sure who is right lol).

It sounds like for the following example - where I'm going to create a slab of bytes with N number of TLV entries for type `Vault` - you'd rather someone use different types altogether (ie. different discriminators)

```
#[spl_discrim("one_vaulted_token")]
struct Vault<V> where V: SplToken {
   authority: Pubkey,
   token: V,
}

#[derive(SplToken)]
struct Usdc {}

#[derive(SplToken)]
struct Bonk {}

...
let first_tlv_entry = Usdc::new();
let second_tlv_entry = Bonk::new();
tlv_state.pack(first_tlv_entry);
tlv_state.pack(second_tlv_entry);
...
```

You could also go with nested TLV:

```
#[spl_discrim("one_vaulted_token")]
struct Vault<V> where V: SplDiscriminate {
   authority: Pubkey,
   token: V,
}

#[spl_discrim("vaulted_usdc")]
struct Usdc {}

#[spl_discrim("vaulted_bonk")]
struct Bonk {}

...
let first_tlv_entry = Usdc::new();
let second_tlv_entry = Bonk::new();
tlv_state.pack(first_tlv_entry);
tlv_state.pack(second_tlv_entry);
...
```

joncinque: Yeah I guess someone could do this. The crucial bit to note here is that these examples don't work unless we allow for multiple entries of the same type.

joncinque (approved):
Looks good!

PR #4602, opened 2023-06-26
Expanding TLV lib

buffalojoec:
This PR seeks to make progress toward building out the TLV library (`type-length-value`) to be more robust and cover a wider range of generic use cases for TLV.
- Moves `PodSlice` and associated types/functions into `type-length-value`, rather than `tlv-account-resolution`
- Cuts unnecessary dependencies and trait impls in favor of `spl-program-error`

The `PodSlice` types I felt are extremely useful and might be useful to other TLV use cases - beyond just account resolution. It's my belief that they should be included in the standard library. Open to feedback!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6782,15 +6782,12 @@ name = "spl-tlv-account-resolution"
 version = "0.1.0"
 dependencies = [
  "bytemuck",
- "num-derive",
- "num-traits",
- "num_enum 0.6.1",
  "solana-program",
  "solana-program-test",
  "solana-sdk",
  "spl-discriminator",
+ "spl-program-error",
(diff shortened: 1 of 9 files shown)


joncinque (approved):
I'm assuming that everything is just copy-pasted, so I didn't take a deep look. But it looks good! We might just want to move all of the pod types into their own library at some point...

buffalojoec:
> I'm assuming that everything is just copy-pasted, so I didn't take a deep look. But it looks good! We might just want to move all of the pod types into their own library at some point...

Yes, copy-pasta'd !

Agreed a `Pod` lib might be useful/appropriate in the future but I vote waiting until we can identify more use cases ie. zero-copy stuff

PR #4617, opened 2023-06-27
docs: Update node version and docusaurus

joncinque:
#### Problem

The docs are pretty far behind, which is why dependabot bumped a bunch of stuff in #4598. Unfortunately, even though that all succeeded locally, it failed deployment to vercel because it was still using node 14 over there.

#### Solution

Redo the bumps, but also force Node 18 here. Vercel is already using the new version


joncinque:
The docs job passed, which is the most important part -- merging this in!

PR #4623, opened 2023-06-27
bump/token-2022

buffalojoec:
- `spl-token-2022`: 1.0.0
- `spl-transfer-hook-interface`: 0.1.0
- `spl-transfer-hook-example`: 0.1.0

The two `transfer-hook` crates were 0.0.1 on crates.io, so we don't actually need to bump them, just publish them.

### DIFF
--- a/associated-token-account/program-test/Cargo.toml
+++ b/associated-token-account/program-test/Cargo.toml
@@ -16,4 +16,4 @@ solana-program-test = "1.16.1"
 solana-sdk = "1.16.1"
 spl-associated-token-account = { version = "1.1", path = "../program", features = ["no-entrypoint"] }
 spl-token = { version = "4.0", path = "../../token/program", features = ["no-entrypoint"] }
-spl-token-2022 = { version = "0.6", path = "../../token/program-2022", features = ["no-entrypoint"] }
+spl-token-2022 = { version = "1.0", path = "../../token/program-2022", features = ["no-entrypoint"] }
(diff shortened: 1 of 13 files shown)


associated-token-account/program-test/Cargo.toml
```diff
@@ -16,4 +16,4 @@ solana-program-test = "1.16.1"
 solana-sdk = "1.16.1"
 spl-associated-token-account = { version = "1.1", path = "../program", features = ["no-entrypoint"] }
 spl-token = { version = "4.0", path = "../../token/program", features = ["no-entrypoint"] }
-spl-token-2022 = { version = "0.6", path = "../../token/program-2022", features = ["no-entrypoint"] }
+spl-token-2022 = { version = "1.0", path = "../../token/program-2022", features = ["no-entrypoint"] }
```
joncinque, line 19: before 1.0, minor versions function like major versions, so we can bump this 0.7 -- we'll go to 1.0 for the first official recommended mainnet release

joncinque (commented):
This looks like it just does token-2022 and not the transfer-hook, is that what you meant to do?

associated-token-account/program-test/Cargo.toml
buffalojoec: Copy, going with 0.7

joncinque (approved):
Ah sorry I misunderstood. I thought you were bumping the transfer hook crates in this PR, so I got confused. All good!

PR #4636, opened 2023-06-27
feat: cli subcommand for token metadata address extension

### LINKED ISSUE #4551: token-cli: Support metadata pointer extension
#### Problem

With https://github.com/solana-labs/solana-program-library/pull/4549, token-2022 can also store a pointer to a metadata account, but this isn't exposed in the CLI except for setting authorities.

#### Solution

Add the functionality

jim4067:
Resolves #4551

image

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -7458,7 +7458,7 @@ mod tests {
     #[serial]
     async fn metadata_pointer() {
         let (test_validator, payer) = new_validator_for_test().await;
-        let program_id = &spl_token_2022::id();
+        let program_id = spl_token_2022::id();
         let config = test_config_with_default_signer(&test_validator, &payer, &program_id);
         let metadata_address = Pubkey::new_unique();
 
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -7485,8 +7485,8 @@ mod tests {
         let mint_state = StateWithExtensionsOwned::<Mint>::unpack(account.data).unwrap();
 
         let extension = mint_state.get_extension::<MetadataPointer>().unwrap();
-
         assert!(mint_state.get_extension::<MetadataPointer>().is_ok());
+        
         assert_eq!(
             extension.metadata_address,
             Some(metadata_address).try_into().unwrap()
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -7467,6 +7467,7 @@ mod tests {
             &payer,
             &[
                 "spl-token",
+                CommandName::CreateToken.into(),
                 "--program-id",
                 &program_id.to_string(),
                 "--metadata-address",
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -7461,43 +7461,10 @@ mod tests {
         let config =
             test_config_with_default_signer(&test_validator, &payer, &spl_token_2022::id());
         let program_id = &spl_token_2022::id();
-        let token_keypair = Keypair::new();
         let token_path = NamedTempFile::new().unwrap();
-        let token_pubkey = token_keypair.pubkey();
-        let bulk_signers: Vec<Arc<dyn Signer>> =
-            vec![Arc::new(clone_keypair(&payer)), Arc::new(token_keypair)];
         let metadata_address = Pubkey::new_unique();
 
-        command_create_token(
-            &config,
-            TEST_DECIMALS,
-            token_pubkey,
-            payer.pubkey(),
-            false,
-            true,
-            false,
-            false,
-            None,
-            Some(metadata_address),
-            None,
-            None,
-            None,
-            None,
-            bulk_signers,
-        )
-        .await
-        .unwrap();
-
-        let account = config.rpc_client.get_account(&token_pubkey).await.unwrap();
-        let test_mint = StateWithExtensionsOwned::<Mint>::unpack(account.data).unwrap();
-        let extension = test_mint.get_extension::<MetadataPointer>().unwrap();
-        assert!(test_mint.get_extension::<MetadataPointer>().is_ok());
-        assert_eq!(
-            extension.metadata_address,
-            Some(metadata_address).try_into().unwrap()
-        );
-
-        process_test_command(
+        let result = process_test_command(
             &config,
             &payer,
             &[
@@ -7507,10 +7474,22 @@ mod tests {
                 "--program-id",
                 &program_id.to_string(),
                 "--metadata-address",
-                metadata_address.to_string().as_str()
+                metadata_address.to_string().as_str(),
             ],
         )
-        .await
-        .unwrap();
+        .await;
+
+        let value: serde_json::Value = serde_json::from_str(&result.unwrap()).unwrap();
+        let mint = Pubkey::from_str(value["commandOutput"]["address"].as_str().unwrap()).unwrap();
+        let account = config.rpc_client.get_account(&mint).await.unwrap();
+        let mint_state = StateWithExtensionsOwned::<Mint>::unpack(account.data).unwrap();
+
+        let extension = mint_state.get_extension::<MetadataPointer>().unwrap();
+
+        assert!(mint_state.get_extension::<MetadataPointer>().is_ok());
+        assert_eq!(
+            extension.metadata_address,
+            Some(metadata_address).try_into().unwrap()
+        );
     }
 }
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -2622,8 +2622,8 @@ fn app<'a, 'b>(
                         ),
                 )
                 .arg(
-                    Arg::with_name("metadata-address")
-                        .long("metadata_address")
+                    Arg::with_name("metadata_address")
+                        .long("metadata-address")
                         .value_name("METADATA_ADDRESS")
                         .takes_value(true)
                         .help(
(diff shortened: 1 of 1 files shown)


token/cli/src/main.rs
```diff
@@ -2613,6 +2621,15 @@ fn app<'a, 'b>(
                             Rate authority defaults to the mint authority."
                         ),
                 )
+                .arg(
+                    Arg::with_name("metadata-address")
```
joncinque, line 2625: Otherwise it won't ever get picked up!
```suggestion
                    Arg::with_name("metadata_address")
```

token/cli/src/main.rs
```diff
@@ -3768,8 +3785,10 @@ async fn process_command<'a>(
             let decimals = value_t_or_exit!(arg_matches, "decimals", u8);
             let mint_authority =
                 config.pubkey_or_default(arg_matches, "mint_authority", &mut wallet_manager)?;
-            let memo = value_t!(arg_matches, "memo", String).ok();
+            let memo: Option<String> = value_t!(arg_matches, "memo", String).ok();
```
joncinque, line 3788: nit: Can you revert this change?

token/cli/src/main.rs
```diff
@@ -3768,8 +3785,10 @@ async fn process_command<'a>(
...
             let rate_bps = value_t!(arg_matches, "interest_rate", i16).ok();
+            let metadata_address: Option<Pubkey> =
+                value_t!(arg_matches, "metadata_address", Pubkey).ok();
```
joncinque, line 3791: nit for consistency
```suggestion
            let metadata_address = value_t!(arg_matches, "metadata_address", Pubkey).ok();
```

joncinque (commented):
Can you also add a test for this? You just need to create the token with a metadata address and then check that the mint has that extension and metadata address.

As it stands, this doesn't work unfortunately, but with a test, we'll be sure!

token/cli/src/main.rs
jim4067: a push with the changes removed should do the trick, I think

joncinque:
Let's keep all the conversation here rather than in two places. To build the programs, you need to go into `token/program-2022` and run `cargo build-sbf`, and then go into `token/program` and run `cargo build-sbf`, and then go into `associated-token-account/program` and run `cargo build-sbf`, then you can run `cargo test` in the CLI directory

token/cli/src/main.rs
```diff
@@ -7425,4 +7453,64 @@ mod tests {
...
+            None,
+            Some(metadata_address),
+            None,
+            None,
+            None,
+            None,
+            bulk_signers,
+        )
+        .await
+        .unwrap();
```
joncinque, line 7489: Rather than using `command_create_token`, which doesn't exercise the command line parsing, can you execute using `process_test_command`, ie:

```
process_test_command(
    &config, &payer,
    &["spl-token", CommandName::CreateToken.into(), "--metadata-address", metadata_address.to_string()]
).await;
```

You're doing this later in the test, so can you remove this and move that call up here? You'll just need to extract the mint info, just like in the `create_token_default` test

jim4067: Is it okay to omit the `--program_id` arguments?

joncinque: Ah, no, definitely not

jim4067:
having trouble testing. 
Getting this error when trying to build `program-2022` - `error: package `ark-serialize v0.4.2` cannot be built because it requires rustc 1.63 or newer, while the currently active rustc version is 1.62.0-dev`


I'm thinking it has something to do with the MSRV for these crates but I can't find them in the `cargo.toml`

joncinque:
The rust toolchain is defined at the `rust-toolchain.toml` file https://github.com/solana-labs/solana-program-library/blob/master/rust-toolchain.toml

Otherwise, it looks like you're using an older version of the SBF platform tools. Be sure to run `solana-install init 1.16.1` to match up with the SPL repo

token/cli/src/main.rs
```diff
@@ -7507,10 +7474,22 @@ mod tests {
...
+        .await;
+
+        let value: serde_json::Value = serde_json::from_str(&result.unwrap()).unwrap();
+        let mint = Pubkey::from_str(value["commandOutput"]["address"].as_str().unwrap()).unwrap();
+        let account = config.rpc_client.get_account(&mint).await.unwrap();
+        let mint_state = StateWithExtensionsOwned::<Mint>::unpack(account.data).unwrap();
+
+        let extension = mint_state.get_extension::<MetadataPointer>().unwrap();
+
+        assert!(mint_state.get_extension::<MetadataPointer>().is_ok());
```
joncinque, line 7489: nit: you can remove this line since you already unwrapped on the same call one line before

joncinque (commented):
We're looking very close!

jim4067:
let me know when it's ready, so that I can squash the commits.

token/cli/src/main.rs
joncinque: Sorry, I meant this exact line, not the whitespace that was there before

token/cli/src/main.rs
```diff
@@ -7425,4 +7453,43 @@ mod tests {
...
         );
     }
+
+    #[tokio::test]
+    #[serial]
+    async fn metadata_pointer() {
+        let (test_validator, payer) = new_validator_for_test().await;
+        let config =
+            test_config_with_default_signer(&test_validator, &payer, &spl_token_2022::id());
+        let program_id = &spl_token_2022::id();
```
joncinque, line 7463: nit: you can reuse the same variable, no need to redeclare it elsewhere!
```suggestion
        let program_id = &spl_token_2022::id();
        let config =
            test_config_with_default_signer(&test_validator, &payer, &program_id);
```

token/cli/src/main.rs
```diff
@@ -2613,6 +2621,15 @@ fn app<'a, 'b>(
                             Rate authority defaults to the mint authority."
                         ),
                 )
+                .arg(
+                    Arg::with_name("metadata_address")
+                        .long("metadata-address")
+                        .value_name("METADATA_ADDRESS")
```
joncinque, line 2627: nit: no need to show `--metadata-address METADATA_ADDRESS` in the help, so this can be simplified
```suggestion
                        .value_name("ADDRESS")
```

token/cli/src/main.rs
```diff
@@ -2613,6 +2621,15 @@ fn app<'a, 'b>(
...
+                        .takes_value(true)
+                        .help(
+                            "Specify metadata address to use."
```
joncinque, line 2630: nit: let's make this a bit clearer
```suggestion
                            "Specify address that stores token metadata."
```

token/cli/src/main.rs
```diff
@@ -7425,4 +7453,43 @@ mod tests {
...
     }
+
+    #[tokio::test]
+    #[serial]
+    async fn metadata_pointer() {
+        let (test_validator, payer) = new_validator_for_test().await;
+        let config =
+            test_config_with_default_signer(&test_validator, &payer, &spl_token_2022::id());
+        let program_id = &spl_token_2022::id();
+        let token_path = NamedTempFile::new().unwrap();
```
joncinque, line 7464: Is this needed? Only tests that need the file to be written somewhere use it. Otherwise, the CLI will generate a new keypair, and you can pull it out from `value["commandOutput"]["address"]`, as you're already doing

token/cli/src/main.rs
```diff
@@ -7425,4 +7453,43 @@ mod tests {
...
+            &config,
+            &payer,
+            &[
+                "spl-token",
+                CommandName::CreateToken.into(),
+                token_path.path().to_str().unwrap(),
+                "--program-id",
+                &program_id.to_string(),
+                "--metadata-address",
+                metadata_address.to_string().as_str(),
```
joncinque, line 7477: nit for consistency with `program_id`
```suggestion
                &metadata_address.to_string(),
```

joncinque (commented):
Sorry, since my last suggestion wasn't totally applied, I found a few other nits to clean up. All the functionality is correct though!

joncinque:
> let me know when it's ready, so that I can squash the commits.

Don't worry, we have a "Squash and merge" policy by default, so feel free to add as many commit as you need

joncinque:
Be sure to run `cargo fmt --all` too before committing

token/cli/src/main.rs
jim4067: Here we expect a string slice, that's why the `metadata_address.to_string().as_str()`

joncinque (commented):
Sorry, one last bit!

token/cli/src/main.rs
```diff
@@ -7425,4 +7453,39 @@ mod tests {
                 .unwrap()
         );
     }
+
+    #[tokio::test]
+    #[serial]
+    async fn metadata_pointer() {
+        let (test_validator, payer) = new_validator_for_test().await;
+        let program_id = &spl_token_2022::id();
+        let config = test_config_with_default_signer(&test_validator, &payer, &program_id);
```
joncinque, line 7462: Clippy's complaining about this, you can just do
```suggestion
        let program_id = spl_token_2022::id();
        let config = test_config_with_default_signer(&test_validator, &payer, &program_id);
```

joncinque (approved):
Thanks so much for your contribution!

PR #4646, opened 2023-06-28
token-2022: Make `Extension::get_account_len` fallible

joncinque:
#### Problem

#### Solution

This is going to be done in a few steps for ease of review. This step just makes `ExtensionType::get_account_len` fallible, so that if you put in `ExtensionType::Metadata`, you get an error.

Thoughts on renaming the function to make that clearer? I went the lazy route and kept it the same name, but I thought about `get_static_account_len` or `get_account_len_with_sized`, but the added words didn't seem worth it.

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -469,7 +469,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
             value_start,
         } = get_extension_indices::<V>(self.tlv_data, false)?;
 
-        if self.tlv_data[type_start..].len() < V::TYPE.get_tlv_len()? {
+        if self.tlv_data[type_start..].len() < V::TYPE.try_get_tlv_len()? {
             return Err(ProgramError::InvalidAccountData);
         }
         let length = pod_from_bytes::<Length>(&self.tlv_data[length_start..value_start])?;
@@ -499,7 +499,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
             value_start,
         } = get_extension_indices::<V>(self.tlv_data, true)?;
 
-        if self.tlv_data[type_start..].len() < V::TYPE.get_tlv_len()? {
+        if self.tlv_data[type_start..].len() < V::TYPE.try_get_tlv_len()? {
             return Err(ProgramError::InvalidAccountData);
         }
         let extension_type = ExtensionType::try_from(&self.tlv_data[type_start..length_start])?;
@@ -670,14 +670,15 @@ pub enum ExtensionType {
     ConfidentialTransferFeeAmount,
     /// Mint contains a pointer to another account (or the same account) that holds metadata
     MetadataPointer,
-    /// Mint contains token metadata
-    Metadata,
+    /// Test unsized mint extension
+    #[cfg(test)]
+    UnsizedMintTest = u16::MAX - 2,
     /// Padding extension used to make an account exactly Multisig::LEN, used for testing
     #[cfg(test)]
-    AccountPaddingTest = u16::MAX - 1,
+    AccountPaddingTest,
     /// Padding extension used to make a mint exactly Multisig::LEN, used for testing
     #[cfg(test)]
-    MintPaddingTest = u16::MAX,
+    MintPaddingTest,
 }
 impl TryFrom<&[u8]> for ExtensionType {
     type Error = ProgramError;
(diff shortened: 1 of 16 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -670,6 +670,8 @@ pub enum ExtensionType {
     ConfidentialTransferFeeAmount,
     /// Mint contains a pointer to another account (or the same account) that holds metadata
     MetadataPointer,
+    /// Mint contains token metadata
+    Metadata,
```
CriesofCarrots, line 674: Seems like slipping this in this PR at least justifies a separate commit. Maybe unintended?

token/program-2022/src/extension/mod.rs
```diff
@@ -693,79 +695,87 @@ impl From<ExtensionType> for [u8; 2] {
...
         // dedupe extensions
         let mut extensions = vec![];
         for extension_type in extension_types {
             if !extensions.contains(&extension_type) {
                 extensions.push(extension_type);
             }
         }
-        let tlv_len: usize = extensions.iter().map(|e| e.get_tlv_len()).sum();
-        if tlv_len
+        let tlv_len: Result<usize, _> = extensions.iter().map(|e| e.get_tlv_len()).sum();
```
CriesofCarrots, line 756: Sneaky ;)

token/program-2022/src/extension/mod.rs
```diff
@@ -693,79 +695,87 @@ impl From<ExtensionType> for [u8; 2] {
...
-                pod_get_packed_len::<ConfidentialTransferFeeConfig>()
+                Ok(pod_get_packed_len::<ConfidentialTransferFeeConfig>())
             }
             ExtensionType::ConfidentialTransferFeeAmount => {
-                pod_get_packed_len::<ConfidentialTransferFeeAmount>()
+                Ok(pod_get_packed_len::<ConfidentialTransferFeeAmount>())
             }
-            ExtensionType::MetadataPointer => pod_get_packed_len::<MetadataPointer>(),
+            ExtensionType::MetadataPointer => Ok(pod_get_packed_len::<MetadataPointer>()),
+            ExtensionType::Metadata => Err(ProgramError::InvalidArgument),
```
CriesofCarrots, line 731: What do you think about adding a method along the lines of `get_account_type` that is something like:
```
pub fn sized(&self) -> bool {
    match self {
        ExtensionType::Metadata => false,
        _ => true, // or list extensions explicitly to force people adding new extensions to handle
    }
}
```

And then the method here becomes:
```
pub fn get_type_len(&self) -> Result<usize, ProgramError> {
    if !self.sized() {
        return Err(ProgramError::InvalidArgument);
    }
    Ok(match self {
        ...
        ExtensionType::Metadata => unreachable!(),
    })
}
```
I feel like this would make it slightly more obvious what's going on with extensions like Metadata. Flipside, it is kind of bloaty. Thoughts?

CriesofCarrots (commented):
> Thoughts on renaming the function to make that clearer? I went the lazy route and kept it the same name, but I thought about get_static_account_len or get_account_len_with_sized, but the added words didn't seem worth it.

The rusty name for fallible methods might be something like `try_get_account_len` (`try_get_type_len`, etc). Wdyt? I know this doesn't express the kinds of extensions that will fail. My suggestion about a `ExtensionType::sized()` method is intended to make that more obvious.

token/program-2022/src/extension/mod.rs
joncinque: I wanted to have an error case to test, but I guess I could do a different one through a `#[cfg(test)]`

token/program-2022/src/extension/mod.rs
joncinque: it look even nicer in later PRs ;-)

token/program-2022/src/extension/mod.rs
```diff
@@ -693,79 +695,87 @@ impl From<ExtensionType> for [u8; 2] {
...
         {
             tlv_len.saturating_add(size_of::<ExtensionType>())
         } else {
             tlv_len
-        }
+        };
+        Ok(tlv_len)
     }
 
     /// Get the required account data length for the given ExtensionTypes
```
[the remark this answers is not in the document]
joncinque: will do!

token/program-2022/src/extension/mod.rs
CriesofCarrots: Or you could just break into different commits ;)

joncinque:
> The rusty name for fallible methods might be something like try_get_account_len (try_get_type_len, etc).

Ooh yeah that's much better. I was fixated on the "this only works if the types are sized" bit, but `try_` already helps a lot

token/program-2022/src/extension/mod.rs
joncinque: I'm down for it, feels a lot like "code as documentation", so even though it isn't necessary, it clarifies what's going on

token/program-2022/src/extension/mod.rs
joncinque: The `#[cfg(test)]` unsized variant seemed clearer, and it'll be much better for later

joncinque:
All your feedback should be addressed

token/program-2022/src/extension/mod.rs
```diff
@@ -670,12 +670,15 @@ pub enum ExtensionType {
...
+    #[cfg(test)]
+    UnsizedMintTest = u16::MAX - 2,
     /// Padding extension used to make an account exactly Multisig::LEN, used for testing
     #[cfg(test)]
-    AccountPaddingTest = u16::MAX - 1,
+    AccountPaddingTest,
     /// Padding extension used to make a mint exactly Multisig::LEN, used for testing
     #[cfg(test)]
-    MintPaddingTest = u16::MAX,
+    MintPaddingTest,
```
CriesofCarrots, line 681: Is there a reason to remove the variant index assignments that I missed?

CriesofCarrots (approved):
One nit/question, but lgtm!

token/program-2022/src/extension/mod.rs
joncinque: just to avoid the verbosity of setting at each one, and instead just setting it at the lowest test one, trusting the enum to assign the correct numbers. Should I revert?

CriesofCarrots: It's probably fine; I don't feel strongly. Feel free to merge as-is :)

PR #4647, opened 2023-06-28
token-2022: Refactor for unsized extension support

joncinque:
#### Problem

#### Solution

This is just a refactor, moving some functions and parameters around, and adding some more return info from other functions. No functional changes are contained here for ease of review in the next bit, which starts to add new functionality, after #4646 lands too.

Roughly, the steps after that go:

* add `alloc` in the public interface, for allocating bytes to an extension
* add `UnsizedExtension` trait, which specifies that an extension can use `alloc` and `realloc`. This is for compile-time safety, to keep using the in-place `get_extension` for all other extensions
* add `realloc`, for an existing extension (ripping off https://github.com/solana-labs/solana-program-library/blob/ed8818c53438b32f96a77f86752e98db02a764ef/libraries/type-length-value/src/state.rs#L358)
* add helpers for fetching the sizes of all TLV entries, along with calculating a new size for a realloc
* add `realloc` for the whole account (ripping off https://github.com/solana-labs/solana-program-library/blob/ed8818c53438b32f96a77f86752e98db02a764ef/libraries/type-length-value/src/state.rs#L414)

Once that's all in place, it'll be a breeze to implement the metadata instructions!

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -96,6 +96,23 @@ fn get_tlv_indices(type_start: usize) -> TlvIndices {
     }
 }
 
+/// Helper function to tack on the size of an extension bytes if an account with
+/// extensions is exactly the size of a multisig
+const fn adjust_len_for_multisig(account_len: usize) -> usize {
+    if account_len == Multisig::LEN {
+        account_len.saturating_add(size_of::<ExtensionType>())
+    } else {
+        account_len
+    }
+}
+
+/// Helper function to calculate exactly how many bytes a value will take up
+const fn add_type_and_length_to_len(value_len: usize) -> usize {
+    value_len
+        .saturating_add(size_of::<ExtensionType>())
+        .saturating_add(pod_get_packed_len::<Length>())
+}
+
 /// Helper struct for returning the indices of the type, length, and value in
 /// a TLV entry
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


CriesofCarrots (commented):
I'm going to wait until rebase to do a fine-toothed review, but as someone who isn't yet familiar with the path to metadata, I'm having a little trouble grokking everything here and ensuring there's not behavior change.
Could you split up the various changes in this PR into separate commits?
For instance, potential commits:
- moving the multisig adjustment logic from `get_total_tlv_len()` to `get_account_len()`
- replacing multisig adjustment logic with helper fn
- changing `get_extension_types_and_allocated_len` return type (not immediately clear why)
- etc

joncinque:
Of course, no problem! The part that I skipped over is that once we have access to the raw bytes, and can alloc / realloc unsized extensions, then we can use any not-in-place serde (read: borsh) much more easily.

For initialization, the flow goes:
* have an instance of the unsized extension
* alloc the exact number of bytes that it needs
* write the unsized instance into those bytes

And during update, the floes goes:
* have a new instance of the unsized extension
* get the old size from the buffer
* get the new size from the instance
* realloc the TLV slot and underlying `AccountInfo`
* write the unsized instance into the bytes

I'll split up the commits more clearly to work up to that

joncinque:
Closing in favor of #4656

PR #4649, opened 2023-06-28
[token-2022] Update confidential transfer configure account for Solana 1.16.1

samkim-crypto:


samkim-crypto:
Subsumed by https://github.com/solana-labs/solana-program-library/pull/4663 and https://github.com/solana-labs/solana-program-library/pull/4664.

PR #4656, opened 2023-06-29
token-2022: Allow alloc and realloc of unsized extensions

joncinque:
#### Problem

This is a more complete and well-separated (I hope) version of #4647.

#### Solution

Step by step, it roughly goes:

* expose the raw extension bytes (the refactor part)
* add a new trait for unsized extensions
* add new alloc and realloc functions for the unsized extensions

Everthing from ba1d352b4873a6b86c0f3323600505be82f2bffe to 57e93d13274388a0c064453c257e3d903131e4aa is refactoring, and after that you'll see the new functionality.

I can separate this into multiple PRs if needed, but I hope the progression makes more sense here!

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -294,6 +294,26 @@ fn get_extension_bytes<S: BaseState, V: Extension>(tlv_data: &[u8]) -> Result<&[
     Ok(&tlv_data[value_start..value_end])
 }
 
+fn get_extension_bytes_mut<S: BaseState, V: Extension>(
+    tlv_data: &mut [u8],
+) -> Result<&mut [u8], ProgramError> {
+    if V::TYPE.get_account_type() != S::ACCOUNT_TYPE {
+        return Err(ProgramError::InvalidAccountData);
+    }
+    let TlvIndices {
+        type_start,
+        length_start,
+        value_start,
+    } = get_extension_indices::<V>(tlv_data, false)?;
+
+    if tlv_data[type_start..].len() < add_type_and_length_to_len(V::TYPE.try_get_type_len()?) {
+        return Err(ProgramError::InvalidAccountData);
+    }
+    let length = pod_from_bytes::<Length>(&tlv_data[length_start..value_start])?;
+    let value_end = value_start.saturating_add(usize::from(*length));
+    Ok(&mut tlv_data[value_start..value_end])
+}
+
 /// Trait for base state with extension
 pub trait BaseStateWithExtensions<S: BaseState> {
     /// Get the buffer containing all extension data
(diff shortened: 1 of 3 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -142,7 +160,12 @@ fn get_extension_indices<V: Extension>(
     Err(ProgramError::InvalidAccountData)
 }
 
-fn get_extension_types(tlv_data: &[u8]) -> Result<Vec<ExtensionType>, ProgramError> {
+/// Fetches the extension types written to the buffer, along with the total
+/// number of bytes used by the TLV entries, including the type and length of
+/// each extension.
```
CriesofCarrots, line 165: The punctuation makes this a little ambiguous; one way to read it is: "Fetches the extension types written to the buffer, including the type and length of each extension."

The best rewrite I've come up with so far is below, but I'm not totally happy with it. Got any ideas?

```suggestion
/// Fetches the extension types written to the buffer, along with the total
/// number of bytes used by the TLV entries, each of which includes the type and length of
/// the extension.
```

token/program-2022/src/extension/mod.rs
```diff
@@ -159,7 +159,12 @@ fn get_extension_indices<V: Extension>(
...
+fn get_extension_types_and_len(
+    tlv_data: &[u8],
+) -> Result<(Vec<ExtensionType>, usize), ProgramError> {
```
CriesofCarrots, line 168: I know the return-type tuple only has 2 elements, but I think it might be nicer as a struct with named fields.

CriesofCarrots (commented):
Thank you so much for splitting up the commits! Significantly easier to review.
Just a couple places I got hung up on docs, and one return-type suggestion.

token/program-2022/src/extension/mod.rs
```diff
@@ -96,6 +97,23 @@ fn get_tlv_indices(type_start: usize) -> TlvIndices {
...
+/// extensions is exactly the size of a multisig
+const fn adjust_len_for_multisig(account_len: usize) -> usize {
+    if account_len == Multisig::LEN {
+        account_len.saturating_add(size_of::<ExtensionType>())
+    } else {
+        account_len
+    }
+}
+
+/// Helper function to calculate exactly how many bytes a value will take up
```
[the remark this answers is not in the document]
joncinque: Ah I see what you mean, maybe just "given its length"?

token/program-2022/src/extension/mod.rs
```diff
@@ -736,7 +736,7 @@ impl ExtensionType {
     /// Get the data length of the type associated with the enum
     ///
     /// Fails if the extension type is unsized
-    pub fn try_get_type_len(&self) -> Result<usize, ProgramError> {
+    fn try_get_type_len(&self) -> Result<usize, ProgramError> {
```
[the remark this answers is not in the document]
joncinque: I had a few branches that made the change, and thought it would be better to have it done separately

token/program-2022/src/extension/mod.rs
joncinque: When adding the new type, I added this info more clearly into its field comment. Let me know how it looks!

token/program-2022/src/extension/mod.rs
joncinque: No problem, I think it neatly resolves the problems with the unclear function comment

joncinque:

token/program-2022/src/extension/mod.rs
CriesofCarrots: Much clearer, thank you for taking the trouble!

CriesofCarrots (approved):
lgtm!

PR #4658, opened 2023-06-29
[compression] added function to locate ChangeLogEvent from transaction

nickfrosty:
Added a helper function named `getAllChangeLogEventV1FromTransaction` to locate and return an array of ChangeLogEventV1 from a [TransactionResponse](https://solana-labs.github.io/solana-web3.js/types/TransactionResponse.html) or [VersionedTransactionResponse](https://solana-labs.github.io/solana-web3.js/types/VersionedTransactionResponse.html) 

Among other things, this will enable people to more easily locate the leaf index that was utilized during an interaction with the tree.

### DIFF
--- a/account-compression/sdk/src/events/index.ts
+++ b/account-compression/sdk/src/events/index.ts
@@ -1,11 +1,15 @@
-import { bs58 } from '@project-serum/anchor/dist/cjs/utils/bytes';
-import BN from 'bn.js';
+import { bs58 } from "@project-serum/anchor/dist/cjs/utils/bytes";
+import BN from "bn.js";
 
-import { ChangeLogEventV1 } from '../types';
-import { accountCompressionEventBeet } from '../generated/types/AccountCompressionEvent';
-import { ApplicationDataEvent, ChangeLogEventV1 as CLV1 } from '../generated';
-import { PublicKey, TransactionResponse } from '@solana/web3.js';
-import { SPL_NOOP_PROGRAM_ID } from '../constants';
+import { ChangeLogEventV1 } from "../types";
+import { accountCompressionEventBeet } from "../generated/types/AccountCompressionEvent";
+import { ApplicationDataEvent, ChangeLogEventV1 as CLV1 } from "../generated";
+import {
+  PublicKey,
+  TransactionResponse,
+  VersionedTransactionResponse,
+} from "@solana/web3.js";
+import { SPL_NOOP_PROGRAM_ID } from "../constants";
 
 /**
  * Helper method for indexing a {@link ConcurrentMerkleTree}
@@ -17,7 +21,7 @@ export function deserializeChangeLogEventV1(data: Buffer): ChangeLogEventV1 {
     .toFixedFromData(data, 0)
     .read(data, 0);
 
-  if (event.__kind == 'ChangeLog' && event.fields[0].__kind == 'V1') {
+  if (event.__kind == "ChangeLog" && event.fields[0].__kind == "V1") {
     const changeLogV1: CLV1 = event.fields[0].fields[0];
     return {
       treeId: changeLogV1.id,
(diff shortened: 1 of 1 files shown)


account-compression/sdk/src/events/index.ts
```diff
@@ -46,3 +49,70 @@ export function deserializeApplicationDataEvent(
...
+  if (!txResponse) throw Error("No txResponse provided");
+
+  // find the correct index of the `programId` instruction
+  const relevantIndex =
+  txResponse.transaction.message.compiledInstructions.findIndex(
+    (instruction) => {
+      return (
+        txResponse?.transaction.message.staticAccountKeys[
+          instruction.programIdIndex
+        ].toBase58() === programId.toBase58()
```
ngundotra, line 75: Make sure to support transactions that use look up tables
```typescript
let accounts = staticAccountKeys + readOnly + writable;
if (accounts[instruction.programIdIndex] == programId.toBase58()) {
  ...
}
```

ngundotra (changes requested):
Thanks for this Nick! Feedback on the account lookups.

Also, can you write this to support multiple ChangeLogs being found?

account-compression/sdk/src/events/index.ts
nickfrosty: This is now added into the PR

nickfrosty:
I have added support for versioned transactions, addresses from LUTs, and returning an array of changelog events.
I also ran prettier on this file along, so the diff is a bit bigger than before

account-compression/sdk/src/events/index.ts
```diff
@@ -39,10 +46,79 @@ export function deserializeApplicationDataEvent(
...
     default:
-      throw Error('Unable to decode buffer as ApplicationDataEvent');
+      throw Error("Unable to decode buffer as ApplicationDataEvent");
   }
 }
+
+/**
+ * Helper function to extract the ChangeLogEvent V1 events from a TransactionResponse
+ * @param txResponse - TransactionResponse from the `@solana/web3.js`
+ * @param programId - PublicKey of the program (aka `programId`) that utilized the leaf on the tree
```
ngundotra, line 60: Thank you for adding this documentation! This description is actually slightly misleading. This function only parses ChangeLogs from the **first** outer instruction that was sent to `programId`. If you want to get **all** changelogs for **all** invocations of the given `programId`, then need to add a loop over outerInstructions **and** also attempt to parse the inner instruction nesting 🫠

account-compression/sdk/src/events/index.ts
```diff
@@ -39,10 +46,79 @@ export function deserializeApplicationDataEvent(
...
+  if (!txResponse) throw Error("No txResponse provided");
+
+  // flatten the array of all account keys (e.g. static, readonly, writable)
+  const accountKeys = txResponse.transaction.message
+    .getAccountKeys()
+    .keySegments()
+    .flat();
+
+  // find the correct index of the `programId` instruction
+  const relevantIndex =
```
ngundotra, line 79: If you want to scan over all `programId` instructions this should be a loop

account-compression/sdk/src/events/index.ts
```diff
@@ -39,10 +46,79 @@ export function deserializeApplicationDataEvent(
...
+        )
+      );
+    } catch (__) {
+      // this noop cpi is not a changelog event. do nothing with it.
+    }
+  }
+
+  // when no changeLogEvents were found, throw an error
+  if (changeLogEvents.length == 0)
+    throw Error("Unable to locate any `ChangeLogEventV1` events");
```
ngundotra, line 121: I think this should just be an empty array, rather than an Error. Someone might use this to scan if they need to store transactions, and it might be annoying to catch an error for a normal output case.

ngundotra (changes requested):
Can you either change the description, remove `programId` argument, or change implementation to look for *all* invocations of the `programId` in both outer & inner instructions?

nickfrosty:
the `programId` argument was removed and now this function parses (only) all inner instructions for noop calls

ngundotra (approved):

PR #4661, opened 2023-06-30
token-2022: Add account-level alloc and realloc

joncinque:
#### Problem

We have many helpers now for allocating and reallocating individual entries in the TLV buffer, but we still can't realloc the underlying `AccountInfo` whenever we add or update an unsized extension.

In the token-metadata interface, the metadata is written after the mint is initialized. During mint initialization, we check to make sure that there's no extra allocated space, to be safe.

This means that during token-metadata initialization, we have to realloc the underlying account in order to write the new data. During updates, we have to realloc the extension to fit the new data size.

#### Solution

Add a few sets of functions and helpers to get this done:

* get the current number of bytes used in the account
* get the future number of bytes used in the account
* for init: realloc the account, alloc the extension, then write the data
* for update: realloc the account, realloc the extension, and write the data (the order here is tricky, which is why we test a bit)

The commits should show a clear story, but let me know if there's anything unclear or if you want me to break anything up! This is the last step, since after this we can confidently do anything with the accounts and TLV buffer. The token-metadata implementation in token-2022 just needs to call into these functions.

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -376,6 +376,39 @@ pub trait BaseStateWithExtensions<S: BaseState> {
             Ok(adjust_len_for_multisig(total_len))
         }
     }
+
+    /// Calculate the new expected size if the state allocates the given number
+    /// of bytes for the given extension type.
+    ///
+    /// Provides the correct answer regardless if the extension is already present
+    /// in the TLV data.
+    fn get_new_account_len<V: UnsizedExtension>(
+        &self,
+        new_extension_len: usize,
+    ) -> Result<usize, ProgramError> {
+        // get the new length used by the extension
+        let new_extension_len = add_type_and_length_to_len(new_extension_len);
+        let tlv_info = get_tlv_data_info(self.get_tlv_data())?;
+        // If we're adding an extension, then we must have at least BASE_ACCOUNT_LENGTH
+        // and account type
+        let current_len = tlv_info
+            .used_len
+            .saturating_add(BASE_ACCOUNT_LENGTH)
+            .saturating_add(size_of::<AccountType>());
+        let new_len = if tlv_info.extension_types.is_empty() {
+            current_len.saturating_add(new_extension_len)
+        } else {
+            // get the current length used by the extension
+            let current_extension_len = self
+                .get_extension_bytes::<V>()
+                .map(|x| add_type_and_length_to_len(x.len()))
+                .unwrap_or(0);
+            current_len
+                .saturating_sub(current_extension_len)
+                .saturating_add(new_extension_len)
+        };
+        Ok(adjust_len_for_multisig(new_len))
+    }
 }
 
 /// Encapsulates owned immutable base state data (mint or account) with possible extensions
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -2274,8 +2274,9 @@ mod test {
             )
         }
     }
+
     #[test]
-    fn alloc_new_tlv_in_account_info() {
+    fn alloc_new_tlv_in_account_info_from_base_size() {
         const VALUE_LEN: usize = 10;
         let base_account_size = Mint::LEN;
         let mut buffer = vec![0; base_account_size];
@@ -2289,14 +2290,59 @@ mod test {
         let value_bytes = [255; VALUE_LEN];
 
         alloc_and_serialize::<Mint, UnsizedMintTest>(&account_info, &value_bytes).unwrap();
-        let new_account_len =
-            BASE_ACCOUNT_LENGTH + size_of::<AccountType>() + add_type_and_length_to_len(VALUE_LEN);
+        let new_account_len = BASE_ACCOUNT_AND_TYPE_LENGTH + add_type_and_length_to_len(VALUE_LEN);
         assert_eq!(data.len(), new_account_len);
         let state = StateWithExtensions::<Mint>::unpack(data.data()).unwrap();
         assert_eq!(
             state.get_extension_bytes::<UnsizedMintTest>().unwrap(),
             value_bytes
         );
+
+        // alloc again fails
+        let account_info = (&key, &mut data).into_account_info();
+        assert_eq!(
+            alloc_and_serialize::<Mint, UnsizedMintTest>(&account_info, &value_bytes).unwrap_err(),
+            TokenError::ExtensionAlreadyInitialized.into()
+        );
+    }
+
+    #[test]
+    fn alloc_new_tlv_in_account_info_from_extended_size() {
+        const VALUE_LEN: usize = 10;
+        let account_size =
+            ExtensionType::try_calculate_account_len::<Mint>(&[ExtensionType::MetadataPointer])
+                .unwrap()
+                + add_type_and_length_to_len(VALUE_LEN);
+        let mut buffer = vec![0; account_size];
+        let mut state = StateWithExtensionsMut::<Mint>::unpack_uninitialized(&mut buffer).unwrap();
+        state.base = TEST_MINT;
+        state.pack_base();
+        state.init_account_type().unwrap();
+
+        let test_key =
+            OptionalNonZeroPubkey::try_from(Some(Pubkey::new_from_array([20; 32]))).unwrap();
+        let extension = state.init_extension::<MetadataPointer>(false).unwrap();
+        extension.authority = test_key;
+        extension.metadata_address = test_key;
+
+        let mut data = SolanaAccountData::new(&buffer);
+        let key = Pubkey::new_unique();
+        let account_info = (&key, &mut data).into_account_info();
+        let value_bytes = [255; VALUE_LEN];
+
+        alloc_and_serialize::<Mint, UnsizedMintTest>(&account_info, &value_bytes).unwrap();
+        let new_account_len = BASE_ACCOUNT_AND_TYPE_LENGTH
+            + add_type_and_length_to_len(VALUE_LEN)
+            + add_type_and_length_to_len(size_of::<MetadataPointer>());
+        assert_eq!(data.len(), new_account_len);
+        let state = StateWithExtensions::<Mint>::unpack(data.data()).unwrap();
+        assert_eq!(
+            state.get_extension_bytes::<UnsizedMintTest>().unwrap(),
+            value_bytes
+        );
+        let extension = state.get_extension::<MetadataPointer>().unwrap();
+        assert_eq!(extension.authority, test_key);
+        assert_eq!(extension.metadata_address, test_key);
 
         // alloc again fails
         let account_info = (&key, &mut data).into_account_info();
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -326,16 +326,16 @@ fn get_extension_bytes_mut<S: BaseState, V: Extension>(
         return Err(ProgramError::InvalidAccountData);
     }
     let TlvIndices {
-        type_start,
+        type_start: _,
         length_start,
         value_start,
     } = get_extension_indices::<V>(tlv_data, false)?;
-
-    if tlv_data[type_start..].len() < add_type_and_length_to_len(V::TYPE.try_get_type_len()?) {
-        return Err(ProgramError::InvalidAccountData);
-    }
+    // get_extension_indices has checked that tlv_data is long enough to include these indices
     let length = pod_from_bytes::<Length>(&tlv_data[length_start..value_start])?;
     let value_end = value_start.saturating_add(usize::from(*length));
+    if tlv_data.len() < value_end {
+        return Err(ProgramError::InvalidAccountData);
+    }
     Ok(&mut tlv_data[value_start..value_end])
 }
 
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -30,6 +30,7 @@ use {
         program_error::ProgramError,
         program_pack::{IsInitialized, Pack},
     },
+    spl_type_length_value::variable_len_pack::VariableLenPack,
     std::{
         cmp::Ordering,
         convert::{TryFrom, TryInto},
@@ -385,7 +386,7 @@ pub trait BaseStateWithExtensions<S: BaseState> {
     ///
     /// Provides the correct answer regardless if the extension is already present
     /// in the TLV data.
-    fn try_get_new_account_len<V: UnsizedExtension>(
+    fn try_get_new_account_len<V: Extension + VariableLenPack>(
         &self,
         new_extension_len: usize,
     ) -> Result<usize, ProgramError> {
@@ -611,7 +612,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     ///
     /// Returns an error if the extension is not present, or if this is not enough
     /// space in the buffer.
-    pub fn realloc<V: UnsizedExtension>(
+    pub fn realloc<V: Extension + VariableLenPack>(
         &mut self,
         length: usize,
     ) -> Result<&mut [u8], ProgramError> {
@@ -663,7 +664,7 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     ///
     /// This can only be used for variable-sized types, such as `String` or `Vec`.
     /// `Pod` types must use `init_extension`
-    pub fn alloc<V: UnsizedExtension>(
+    pub fn alloc<V: Extension + VariableLenPack>(
         &mut self,
         length: usize,
         overwrite: bool,
@@ -862,7 +863,7 @@ pub enum ExtensionType {
     TokenMetadata,
     /// Test unsized mint extension
     #[cfg(test)]
-    UnsizedMintTest = u16::MAX - 2,
+    VariableLenMintTest = u16::MAX - 2,
     /// Padding extension used to make an account exactly Multisig::LEN, used for testing
     #[cfg(test)]
     AccountPaddingTest,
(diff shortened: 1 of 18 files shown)


buffalojoec:
Looking slick !

token/program-2022/src/extension/mod.rs
```diff
@@ -1124,12 +1125,51 @@ impl Extension for AccountPaddingTest {
...
 }
 
+/// Packs an unsized extension into a new TLV space
+///
+/// This function reallocates the account as needed to accommodate for the
+/// change in space, then allocates in the TLV buffer, and finally writes the
+/// bytes.
+pub fn alloc_and_serialize<S: BaseState, V: UnsizedExtension>(
+    account_info: &AccountInfo,
+    value_bytes: &[u8],
```
CriesofCarrots, line 1135: I wonder if we could lock this down a little, rather than accepting any arbitrary bytes. But like you alluded to [here](https://github.com/solana-labs/solana-program-library/pull/4704), I too don't love the idea of being tied to borsh for serialization of UnsizedExtensions.
So I don't know if it's worth the trouble, but I'm imagining something like:
```
// maybe this could just be in an SPL library instead of waiting on solana-program
trait UnsizedPack {
    fn pack_into_slice(&self, dst: &mut [u8]) -> Result<(), ProgramError>;
    fn unpack_from_slice(src: &[u8]) -> Result<Self, ProgramError>;
    fn get_packed_len() -> usize;

    fn pack(&self, dst: &mut [u8]) -> Result<(), ProgramError> {
        if dst.len() != self.get_packed_len() {
            return Err(ProgramError::InvalidAccountData);
        }
        self.pack_into_slice(dst);
        Ok(())
    }
}
```

```
trait UnsizedExtension: Extension + UnsizedPack {}

pub fn alloc_and_serialize<S: BaseState, V: UnsizedExtension(
    account_info: &AccountInfo,
    extension: V,
) -> Result<(), ProgramError> {
    ...
    let mut state = StateWithExtensionsMut::<S>::unpack(&mut buffer)?;
    let data = state.alloc::<V>(extension.get_packed_len(), false)?;
    extension.pack(data);
}
```

Then the TokenMetadata extension/interface could implement something like:
```
fn pack_into_slice(&self, dst: &mut [u8]) -> Result<(), ProgramError> {
    let bytes = self.try_to_vec()?;
    dst.copy_from_slice(value_bytes);
}
```

token/program-2022/src/extension/mod.rs
```diff
@@ -1124,12 +1125,51 @@ impl Extension for AccountPaddingTest {
     const TYPE: ExtensionType = ExtensionType::AccountPaddingTest;
 }
 
+/// Packs an unsized extension into a new TLV space
```
CriesofCarrots, line 1128: Actually this just packs arbitrary bytes. It's not clear what would fail if the bytes were incompatible with unsized extension type V.

token/program-2022/src/extension/mod.rs
```diff
@@ -2122,4 +2162,89 @@ mod test {
...
+                Epoch::default(),
+            )
+        }
+    }
+    #[test]
+    fn alloc_new_tlv_in_account_info() {
+        const VALUE_LEN: usize = 10;
+        let base_account_size = Mint::LEN;
+        let mut buffer = vec![0; base_account_size];
+        let mut state = StateWithExtensionsMut::<Mint>::unpack_uninitialized(&mut buffer).unwrap();
```
CriesofCarrots, line 2274: It would be nice if this test had a sized extension also, to ensure alloc_and_serialize plays nicely with existing extensions.

token/program-2022/src/extension/mod.rs
```diff
@@ -360,6 +360,20 @@ pub trait BaseStateWithExtensions<S: BaseState> {
...
+    /// Get the total number of bytes used by TLV entries and the base type
+    fn get_account_len(&self) -> Result<usize, ProgramError> {
+        let info = get_tlv_data_info(self.get_tlv_data())?;
+        if info.extension_types.is_empty() {
+            Ok(S::LEN)
+        } else {
+            let total_len = info
+                .used_len
+                .saturating_add(BASE_ACCOUNT_LENGTH)
+                .saturating_add(size_of::<AccountType>());
```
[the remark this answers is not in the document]
joncinque: After making the function, I realized it's simpler to make a new constant for `BASE_ACCOUNT_LENGTH + size_of::<AccountType>()`.

token/program-2022/src/extension/mod.rs
```diff
@@ -374,6 +374,39 @@ pub trait BaseStateWithExtensions<S: BaseState> {
             Ok(adjust_len_for_multisig(total_len))
         }
     }
+
+    /// Calculate the new expected size if the state allocates the given number
+    /// of bytes for the given extension type.
+    ///
+    /// Provides the correct answer regardless if the extension is already present
+    /// in the TLV data.
+    fn get_new_account_len<V: UnsizedExtension>(
```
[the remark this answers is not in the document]
joncinque: I think adding `try_` is fine, but we can keep it `try_get_new_account_len` to make it correspond well with `try_get_account_len`

token/program-2022/src/extension/mod.rs
```diff
@@ -360,6 +360,20 @@ pub trait BaseStateWithExtensions<S: BaseState> {
     fn get_first_extension_type(&self) -> Result<Option<ExtensionType>, ProgramError> {
         get_first_extension_type(self.get_tlv_data())
     }
+
+    /// Get the total number of bytes used by TLV entries and the base type
+    fn get_account_len(&self) -> Result<usize, ProgramError> {
```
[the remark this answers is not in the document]
joncinque: That works for me!

token/program-2022/src/extension/mod.rs
```diff
@@ -1124,12 +1125,51 @@ impl Extension for AccountPaddingTest {
...
+    if previous_account_len < new_account_len {
+        // size increased, so realloc the account first
+        account_info.realloc(new_account_len, false)?;
+    }
+
+    let mut buffer = account_info.try_borrow_mut_data()?;
+    // write the account type if needed, so that the next unpack works
+    if previous_account_len <= BASE_ACCOUNT_LENGTH {
+        buffer[BASE_ACCOUNT_LENGTH] = S::ACCOUNT_TYPE.into();
+    }
```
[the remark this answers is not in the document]
joncinque: Ah nice spot, I thought I did this for a good reason, but swapping it with `set_account_type` seems to work, so going with that!

token/program-2022/src/extension/mod.rs
joncinque: This is intentional -- it caught an implementation bug where I forgot to set the account type when going from the base 82 bytes to the extended 165+. The realloc test starts with a 165+ account that already has an extension, so that case is covered there.

Either way, I can add a separate alloc test for that and a comment explaining why this test matters too

token/program-2022/src/extension/mod.rs
joncinque: That's fair, will update

token/program-2022/src/extension/mod.rs
joncinque: I like this approach! I put in #4719 to add that trait to the TLV library so that it can be used here

joncinque: And I put in #4720, thanks so much for the suggestion! Once that lands, I can rebase this PR and just use `UnsizedPack` directly. I'll aim to do it in new commits to make this a bit easier for re-review

token/program-2022/src/extension/mod.rs
CriesofCarrots: Perfect :)

token/program-2022/src/extension/mod.rs
```diff
@@ -368,10 +371,7 @@ pub trait BaseStateWithExtensions<S: BaseState> {
         if info.extension_types.is_empty() {
             Ok(S::LEN)
         } else {
-            let total_len = info
-                .used_len
-                .saturating_add(BASE_ACCOUNT_LENGTH)
-                .saturating_add(size_of::<AccountType>());
+            let total_len = info.used_len.saturating_add(BASE_ACCOUNT_AND_TYPE_LENGTH);
```
CriesofCarrots: Yes, much better, thanks!

CriesofCarrots (commented):
Excellent, new commits all look good! Will review once more after the rebase.

token/program-2022/src/extension/mod.rs
```diff
@@ -181,8 +183,12 @@ fn get_tlv_data_info(tlv_data: &[u8]) -> Result<TlvDataInfo, ProgramError> {
...
         let tlv_indices = get_tlv_indices(start_index);
         if tlv_data.len() < tlv_indices.length_start {
-            // not enough bytes to store the type, malformed
-            return Err(ProgramError::InvalidAccountData);
+            // There aren't enough bytes to store the next type, which means we
+            // got to the end. The last byte could be used during a realloc!
+            return Ok(TlvDataInfo {
+                extension_types,
+                used_len: tlv_indices.type_start,
+            });
```
joncinque, line 191: This was a bug! When we wanted to reallocate  to add 1 more byte to fill a buffer, it would error out here, even though it's a valid situation to have one unused byte at the end.

joncinque:
Ok, the commits starting with https://github.com/solana-labs/solana-program-library/pull/4661/commits/6958b7a5e573318b1d73b3dd5b518ebe40b773f0 integrate `VariableLenPack` in the following steps:

* Rename everything to `VariableLen` where appropriate, and remove `UnsizedExtension` totally
* Add new functions to init / pack variable length extensions into the TLV buffer
* Lock down alloc and realloc via these new functions

This all uncovered a bug, which I've pointed out. Let me know if you have any questions!

token/program-2022/src/extension/mod.rs
```diff
@@ -1077,22 +1159,125 @@ impl Extension for AccountPaddingTest {
...
+    let mut state = StateWithExtensionsMut::<S>::unpack(&mut buffer)?;
+    state.init_variable_len_extension(value, false)
+}
+
+/// Packs arbitrary bytes for a variable-length extension into an existing TLV space
+///
+/// This function reallocates the account as needed to accommodate for the
+/// change in space, then reallocates in the TLV buffer, and finally writes the
+/// bytes.
+pub fn realloc_and_serialize<S: BaseState, V: Extension + VariableLenPack>(
```
CriesofCarrots, line 1199: There's a bit of duplication between `alloc_and_serialize()` and `realloc_and_serialize()`; I'm still not sure we need both and to force the caller to check if the extension already exists. I think we could get by with one, something like this?
```
pub fn alloc_and_serialize<S: BaseState, V: Extension + VariableLenPack>(
    account_info: &AccountInfo,
    new_extension: &V,
) -> Result<(), ProgramError> {
    let previous_account_len = account_info.try_data_len()?;
    let (new_account_len, extension_already_exists) = {
        let data = account_info.try_borrow_data()?;
        let state = StateWithExtensions::<S>::unpack(&data)?;
        let new_account_len = state.try_get_new_account_len(new_extension)?;
        let extension_already_exists = state.get_extension().is_ok();
        (new_account_len, extension_already_exists)
    };

    if previous_account_len < new_account_len {
        // size increased, so realloc the account first
        account_info.realloc(new_account_len, false)?;

        let mut buffer = account_info.try_borrow_mut_data()?;

        if extension_already_exists {
            let mut state = StateWithExtensionsMut::<S>::unpack(&mut buffer)?;
            state.realloc_variable_len_extension(new_extension)?;
        } else {
            // write the account type if needed, so that the next unpack works
            if previous_account_len <= BASE_ACCOUNT_LENGTH {
                set_account_type::<S>(*buffer)?;
            }
            // now alloc in the TLV buffer and write the data
            let mut state = StateWithExtensionsMut::<S>::unpack(&mut buffer)?;
            state.init_variable_len_extension(new_extension, false)
        }
    } else {
        let mut buffer = account_info.try_borrow_mut_data()?;
        let mut state = StateWithExtensionsMut::<S>::unpack(&mut buffer)?;

        if extension_already_exists {
            // do it backwards otherwise, write the state, realloc TLV, then the account
            state.realloc_variable_len_extension(new_extension)?;

            let removed_bytes = previous_account_len
                .checked_sub(new_account_len)
                .ok_or(ProgramError::AccountDataTooSmall)?;
            if removed_bytes > 0 {
                // this is probably fine, but be safe and avoid invalidating references
                drop(buffer);
                account_info.realloc(new_account_len, false)?;
            }
        } else {
            // now alloc in the TLV buffer and write the data
            state.init_variable_len_extension(new_extension, false)
        }
    }
    Ok(())
}
```

CriesofCarrots: I dunno, do you think it's too hard to follow with "one method to rule them all"?

token/program-2022/src/extension/mod.rs
```diff
@@ -1951,82 +2137,379 @@ mod test {
...
+                .unwrap(),
+            variable_len
+        );
+
+        // alloc again fails
+        let account_info = (&key, &mut data).into_account_info();
+        assert_eq!(
+            alloc_and_serialize::<Mint, _>(&account_info, &variable_len).unwrap_err(),
+            TokenError::ExtensionAlreadyInitialized.into()
+        );
```
CriesofCarrots, line 2387: I guess the question above (about unifying methods) is whether we need a failure on a second alloc.

CriesofCarrots (commented):
Getting really close! Mostly nits. Only one substantive question, about whether we need both helper fns.

token/program-2022/src/extension/mod.rs
joncinque: As long as we're clear on the behavior in the comment, then I'm happy to combine both! Since it's pretty low-level, I think it's perfectly legit to add a "thar be dragons here, be sure to check if you're planning reallocating" comment

token/program-2022/src/extension/mod.rs
joncinque: The realloc function is pretty complicated as it is, and the combination doesn't make it much more complicated. Your code here looks legit, gonna test it out and see how it goes!

token/program-2022/src/extension/mod.rs
```diff
@@ -383,7 +384,7 @@ pub trait BaseStateWithExtensions<S: BaseState> {
     ///
     /// Provides the correct answer regardless if the extension is already present
     /// in the TLV data.
-    fn try_get_new_account_len<V: UnsizedExtension>(
+    fn try_get_new_account_len<V: Extension + VariableLenPack>(
```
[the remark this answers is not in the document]
joncinque: Yep! It felt very nice to be able to remove it totally

token/program-2022/src/extension/mod.rs
```diff
@@ -2333,33 +2337,34 @@ mod test {
         let mut data = SolanaAccountData::new(&buffer);
         let key = Pubkey::new_unique();
         let account_info = (&key, &mut data).into_account_info();
-        let value_bytes = [255; VALUE_LEN];
 
-        alloc_and_serialize::<Mint, VariableLenMintTest>(&account_info, &value_bytes).unwrap();
-        let new_account_len = BASE_ACCOUNT_AND_TYPE_LENGTH + add_type_and_length_to_len(VALUE_LEN);
+        alloc_and_serialize::<Mint, _>(&account_info, &variable_len).unwrap();
```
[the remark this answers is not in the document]
joncinque: It's one of my gripes about Rust :disappointed:

token/program-2022/src/extension/mod.rs
CriesofCarrots: I haven't even tried to compile it, so hopefully it's not too horrible 😅

joncinque: It turned out pretty well!

joncinque:
Let me know how you like the combined impl, thanks for suggesting it :partying_face:

token/program-2022/src/extension/mod.rs
```diff
@@ -1166,64 +1166,57 @@ impl Extension for AccountPaddingTest {
...
+/// Packs a variable-length extension into a TLV space
 ///
 /// This function reallocates the account as needed to accommodate for the
 /// change in space, then reallocates in the TLV buffer, and finally writes the
 /// bytes.
-pub fn realloc_and_serialize<S: BaseState, V: Extension + VariableLenPack>(
+///
+/// NOTE: This function does not protect against double-alloc, and will realloc
+/// an existing extension. The calling code must check if the extension already
+/// exists, and return an error if appropriate.
```
CriesofCarrots, line 1177: I'm fine with this approach as-is, but if you think we need more guardrails, we could pass an `overwrite: bool` like we do with `StateWithExtensionsMut::init_extension()` to make things more obvious and controllable. Wdyt?

token/program-2022/src/extension/mod.rs
```diff
@@ -1166,64 +1166,57 @@ impl Extension for AccountPaddingTest {
...
-        state.realloc_variable_len_extension(new_extension)?;
+        if extension_already_exists {
+            state.realloc_variable_len_extension(new_extension)?;
+        } else {
+            // this situation can happen if we have an overallocated buffer
+            state.init_variable_len_extension(new_extension, false)?;
+        }
 
         let removed_bytes = previous_account_len
             .checked_sub(new_account_len)
```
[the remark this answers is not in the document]
joncinque: Yeah that part made me really happy. And for sure, will do!

token/program-2022/src/extension/mod.rs
joncinque: Do you mean something like:

```
if extension_already_exists && !overwrite {
    return Err(...);
}
```

That seems reasonable to me!

CriesofCarrots: Yes, exactly!

joncinque:
Ok I think this is the last round :crossed_fingers:

CriesofCarrots (approved):
:shipit: 
Thanks for all the polish! I'm pleased with where this ended up :)

joncinque:
>  Thanks for all the polish! I'm pleased with where this ended up :)

Thanks for all of your great reviews, I really like it too! Everything resolved itself so neatly

PR #4662, opened 2023-07-02
token-2022: Return error in `get_account_data_size` for non-account extension types

### LINKED ISSUE #4587: token-2022: `get_account_data_size` accepts mint extensions
#### Problem

The `get_account_data_size` instruction accepts all any extension type, including `Uninitialized` and mint extensions. It will happily calculate the number, even if it isn't possible to initialize an account with those extensions later.

#### Solution

For safety, return an error for `Uninitialized` or any extension types that describe mints.

cc @ryoqun

dhruvdabhi101:
Closes #4587 

Returning error for `Uninitialized` and mint extension types. 

Please review and tell if I need to change anything. 
Thank You.


token/program-2022/src/processor.rs
```diff
@@ -1,5 +1,7 @@
 //! Program state processor
 
+use crate::extension::Extension;
```
joncinque, line 3: Can you remove this line? Looks like it isn't used according to clippy

token/program-2022/src/processor.rs
```diff
@@ -1244,6 +1246,13 @@ impl Processor {
         accounts: &[AccountInfo],
         new_extension_types: Vec<ExtensionType>,
     ) -> ProgramResult {
+        if new_extension_types
+            .iter()
+            .any(|&t| t == ExtensionType::MintCloseAuthority || t == ExtensionType::Uninitialized || t == ExtensionType::ConfidentialTransferMint)
+        {
+            return Err(TokenError::ExtensionTypeMismatch.into());
+        }
```
joncinque, line 1254: This only covers two of the mint extensions. To make this more robust, you can change this to
```suggestion
        if new_extension_types
            .iter()
            .any(|&t| t.get_account_type() != AccountType::Account)
        {
            return Err(TokenError::ExtensionTypeMismatch.into());
        }
```

Using the specific helper for this https://github.com/solana-labs/solana-program-library/blob/5900d4d85f3662dd910abe84bafe0656ae4ba495/token/program-2022/src/extension/mod.rs#L930

token/program-2022/src/processor.rs
```diff
@@ -7524,6 +7533,32 @@ mod tests {
...
+                vec![&mut invalid_mint_account],
+            ),
+            Err(TokenError::ExtensionTypeMismatch.into())
+        );
+        assert_eq!(
+            do_process_instruction(
+                get_account_data_size(
+                    &program_id,
+                    &mint_key,
+                    &[ExtensionType::MintCloseAuthority]
```
joncinque, line 7555: to be totally sure, can you add an account extension type here to make sure that the error is still getting triggered?

joncinque (commented):
Just a couple of comments, but looks like it's mostly there!

token/program-2022/src/processor.rs
dhruvdabhi101: Okay. Thanks!

token/program-2022/src/processor.rs
```diff
@@ -7524,6 +7532,44 @@ mod tests {
...
+                get_account_data_size(
+                    &program_id,
+                    &mint_key,
+                    &[ExtensionType::MemoTransfer]
+                )
+                .unwrap(),
+                vec![&mut invalid_mint_account],
+            ),
+            Err(TokenError::ExtensionTypeMismatch.into())
+        );
```
joncinque, line 7572: Sorry, I wasn't clear, I meant to mix a valid account extension with an invalid mint extension, ie to do `&[ExtensionType::MemoTransfer, ExtensionType::MintCloseAuthority]`.

Also, these tests are using an invalid mint, so could you move these tests to after 7483 and use the valid mint?

joncinque (commented):
Just one last comment to sort out. After that, be sure to run `cargo fmt` from `token/program-2022` since CI is complaining about the formatting of the commit

token/program-2022/src/processor.rs
dhruvdabhi101: yes sure

joncinque (approved):
Thanks for your work!

PR #4650, opened 2023-06-29
[token-2022] Update confidential transfer for Solana 1.16.1

samkim-crypto:
#### Problem
There are a number of changes to be made to the confidential transfer with the Solana 1.16 upgrade. Instead of making all the changes in a single PR, it would be nice to break up the changes in smaller PRs for each instruction. Each PR could make changes to the program, token client, and tests related to a particular instruction in the confidential transfer extension.

#### Summary of Changes
This PR is a precursor to the individual PR changes for each instruction.
- [99429af](https://github.com/solana-labs/solana-program-library/pull/4650/commits/99429af1d002cf1ff40d42367a0ec7d1e5ba25db): Currently, the entire confidential transfer test is featurized out with `#![cfg(feature = "proof-program")]` under `proof-program`. This commit assigns `#[cfg(feature = "proof-program")]` to each components in the tests. Subsequent PRs will remove these one by one.
- [3cdb6b4](https://github.com/solana-labs/solana-program-library/pull/4650/commits/3cdb6b48084546960b41e336cf9a0f9e5da745b7) Removed the struct `ConfidentialTransferMintWithKeypairs` in the tests. This is a helper struct to simplify the tests. With recent updates to the program, it seemed like this struct was adding more complexity than it was removing.
- [acee591](https://github.com/solana-labs/solana-program-library/pull/4650/commits/acee591f7e5bfe12a24131f7be7c7ea77bfe8fed): Enabled the tests for initializing and updating confidential transfer mint. This part does not need the proof program, so the tests can be enabled.
- [b489eb4](https://github.com/solana-labs/solana-program-library/pull/4650/commits/b489eb428eb92b6872279ce1317c58253f99d69b): This is a smaller commit. Every place that we use `pod::ElGamalPubkey` or `pod::AeKey`, I just switched to using `ElGamalPubkey` and `AeKey` directly to make the code a little cleaner.
[...]

PR #4663, opened 2023-07-02
[token-2022] Update confidential transfer configure account for 1.16

samkim-crypto:
#### Problem
Confidential transfer configure account instruction is not updated for the new Solana 1.16 changes.

#### Summary of changes
- [b9a5682](https://github.com/solana-labs/solana-program-library/pull/4663/commits/b9a56828ecdad6625f1d62a28081ce63aa6fe04d): The `UpdateMint` instruction constructor and toiken client functions have not been touched for some time and they currently do not support multisig signers. I updated functions to support multisig signers. This should have been a part of the previous PR https://github.com/solana-labs/solana-program-library/pull/4650, which was missed.
- [3cc330b](https://github.com/solana-labs/solana-program-library/pull/4663/commits/3cc330b5b59aa51553c1042e0eb198887e454ffe):
  - Updated `ConfigureAccount` instruction constructor and processor and brought them out of `proof-program` feature.
  - Updated configure account token client. Previously, the token client function was a mess having three different functions each one accepting different parameters. I merged these functions into a single function in-line with the rest of the token client functions for the other instructions. 
    - It takes in an optional `maximum_pending_balance_credit_counter`, which is set to be a default value if not provided.  I put the default constant inside the client function itself since I did not see a clean place to put it. I can move it elsewhere if there are suggestions. The maximum credit counter limits how many times a confidential transfer account can be creditted (via deposits or confidential transfers). This is to prevent ElGamal ciphertexts from overflowing, making them difficult to decrypt.
    - The function also takes in `elgamal_keypair` and `aes_key` as input. Ideally, these parameters can also be optional so that they can be derived from signers when they are `None`, but unfortunately, the zk-token-sdk does not yet support these keys from being derived from `Signers` (only `Signer`). This part can perhaps be updated in the future in a separate PR.
- [21f2b49](https://github.com/solana-labs/solana-program-library/pull/4663/commits/21f2b49829053e40900b18b9b5e9fad5f11da6f2): I unfeaturized and updated `ConfidentialTokenAccountMeta`, which is a private helper type in tests. The changes are straightforward, just updating the interface for solana 1.16.
- [5b92b73](https://github.com/solana-labs/solana-program-library/pull/4663/commits/5b92b732967e21941d6d88dc2b99d3bf91b2ec3b): Updated the tests for the configure account instruction. I these should be straightforward.
- [464dfd6](https://github.com/solana-labs/solana-program-library/pull/4663/commits/464dfd6166005ea187b1f57a8aec4eb62ad119c2): I also updated the `ApproveAccount` instruction constructor and token client. This is a separate instruction, but since it is simple, I thought I could squeeze it into this PR.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -25,6 +25,7 @@ use {
             ExtensionType, StateWithExtensionsOwned,
         },
         instruction, offchain,
+        solana_zk_token_sdk::encryption::{auth_encryption::AeKey, elgamal::ElGamalKeypair},
         solana_zk_token_sdk::{errors::ProofError, zk_token_elgamal::pod::ElGamalPubkey},
         state::{Account, AccountState, Mint, Multisig},
     },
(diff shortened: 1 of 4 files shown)


joncinque (approved):
Looks great! Thanks for breaking down the commits and the reasoning for each

PR #4677, opened 2023-07-03
token-client: Unpin solana deps, and use `solana-rpc-client`

joncinque:
#### Problem

People who want to use the `spl-token-client` crate must use the exact same version of the solana tools, which limits downstream usage.

#### Solution

Relax the dependencies. I noticed that it's always been pinned, but I can't see why we can't relax them for the client library.

And while I was at it, I noticed that we could downgrade the `solana-client` dependency to `solana-rpc-client`.

### DIFF
--- a/token/client/Cargo.toml
+++ b/token/client/Cargo.toml
@@ -5,15 +5,15 @@ edition = "2021"
 license = "Apache-2.0"
 name = "spl-token-client"
 repository = "https://github.com/solana-labs/solana-program-library"
-version = "0.5.0"
+version = "0.5.1"
 
 [dependencies]
 async-trait = "0.1"
 futures-util = "0.3"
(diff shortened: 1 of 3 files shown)


token/client/Cargo.toml
```diff
@@ -5,15 +5,15 @@ edition = "2021"
...
 [dependencies]
 async-trait = "0.1"
 futures-util = "0.3"
-solana-cli-output = { version = "=1.16.1", optional = true }
-solana-client = "=1.16.1"
-solana-program-test = "=1.16.1"
-solana-sdk = "=1.16.1"
+solana-cli-output = { version = "1.16.1", optional = true }
+solana-program-test = "1.16.1"
+solana-rpc-client = "1.16.1"
```
CriesofCarrots, line 15: ❤️

CriesofCarrots (approved):

PR #4678, opened 2023-07-04
[token-2022] Update confidential transfer deposit and apply pending balance for Solana 1.16

samkim-crypto:
#### Problem
Confidential transfer deposit and apply pending balance instructions are not updated for the new Solana 1.16 changes.

#### Summary of changes
- [521c39d](https://github.com/solana-labs/solana-program-library/pull/4678/commits/521c39d99f4d2db8c92a8162ded148167f4ad8c6): Unfeaturized and updated the deposit instruction client to be in-line with the rest of the client functions. I also cleaned up the way a deposit amount is split up into the lo and hi part, which was quite messy before.
- [dbf1b97](https://github.com/solana-labs/solana-program-library/pull/4678/commits/dbf1b970b5018af0c04a917adf6321202ae0207f): Enabled the tests for the deposit instruction. Should be pretty straightforward.
- [ef78511](https://github.com/solana-labs/solana-program-library/pull/4678/commits/ef78511e60c6847fb1ad10aef87006b78dab1467): Unfeaturized and updated the client for the `ApplyPendingBalance` instruction.
  - One issue was that to generate the `ApplyPendingBalance` instruction, one needs the account ciphertext information. The client must (1) read on-chain data, (2) decrypt the info and generate proof, and then (3) create the instruction. I was trying to fit all this logic into the client function, but the function ended up becoming quite large. So I created a separate module `extension_info` and shoved the logic for steps (1) and (2) there. The main client function `confidential_transfer_apply_pending_balance` takes in an optional `ApplyPendingBalanceExtensionInfo`, which is defined in the `extension_info` module.
- [8c8e6ed](https://github.com/solana-labs/solana-program-library/pull/4678/commits/8c8e6ed07fbaffea7791e174e8c3d4585d9de1b8): Since the deposit client function was featurized out, the tests for `{enable,disable}_confidential_credit` tests did not verify that deposit fails after confidential credits are disabled. Now that the deposit function works, I extended the tests to check that deposit fails after confidential credits are disabled.


token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -890,22 +913,6 @@ async fn confidential_transfer_deposit() {
...
-    let state = token
-        .get_account_info(&alice_meta.token_account)
-        .await
-        .unwrap();
-    let extension = state
-        .get_extension::<ConfidentialTransferAccount>()
-        .unwrap();
-    assert_eq!(extension.pending_balance_credit_counter, 0.into());
-    assert_eq!(extension.expected_pending_balance_credit_counter, 2.into());
-    assert_eq!(extension.actual_pending_balance_credit_counter, 2.into());
```
joncinque: These asserts seem important -- did you mean to add them back in?

token/client/src/token.rs
```diff
@@ -2206,57 +2213,42 @@ where
...
+    pub async fn confidential_transfer_apply_pending_balance<S: Signers>(
         &self,
-        token_account: &Pubkey,
-        authority: &S,
-        available_balance: u64,
-        pending_balance: u64,
-        expected_pending_balance_credit_counter: u64,
+        account: &Pubkey,
+        authority: &Pubkey,
+        account_info: Option<ApplyPendingBalanceExtensionInfo>,
```
joncinque: This can take the extension directly if you add the functions on the extension struct type rather than using a new type

joncinque (commented):
Just a couple of refactoring-style questions, but the guts look great

token/client/src/token.rs
samkim-crypto: Looking at it now, implementing the functions directly on `ConfidentialTransferAccount` could be quite clean if we are to refactor the `extension_info` module to token-2022.

I think the main benefit of having a separate struct `ApplyPendingBalanceExtensionInfo` (and for the client function to take in this struct) would be that it holds just the fields from a `ConfidentialTransferAccount` that are needed to generate an `ApplyPendingBalance` instruction and hence it is much shorter than `ConfidentialTransferAccount`. 

I thought this could be useful for something like offline signing in the cli. Users can first look-up the relevant data from a confidential transfer account encoded as `ApplyPendingBalanceExtensionInfo`, copy it, and then use it as input to sign a transaction offline. In this case, a shorter `ApplyPendingBalanceExtensionInfo` would be more convenient than copying the entire `ConfidentialTransferAccount`.

What do you think about this aspect?

token/client/src/extension_info.rs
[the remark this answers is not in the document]
samkim-crypto: Yeah that is a good idea. I think there is some things to decide on whether we want to have separate structs like `ApplyPendingBalanceExtensionInfo`, `WithdrawExtensionInfo`, and `TransferExtensionInfo` (https://github.com/solana-labs/solana-program-library/pull/4678#discussion_r1258963523), but once that is resolved, I will refactor them to token-2022.

token/program-2022-test/tests/confidential_transfer.rs
samkim-crypto: Oh yes, I added these back in. There was actually a small defining the maximum deposit amount constant, so I fixed that as well.

token/client/src/token.rs
joncinque: I see what you mean -- it makes sense to encapsulate this in a separate type for clean separation. How many of these types will you need to support the rest of the instructions? Will it be one per instruction?

samkim-crypto: Yeah, so I think we will need it for the instructions that need zkp's (`EmiptyAccount`, `Withdraw`, `Transfer`, `TransferWithFee` and `WithdrawWithheldTokens`) + `ApplyPendingBalance`.

token/program-2022/src/extension/confidential_transfer/mod.rs
```diff
@@ -6,18 +11,18 @@ use {
...
+    solana_zk_token_sdk::zk_token_elgamal::pod::{
+        AeCiphertext as PodAeCiphertext, ElGamalCiphertext, ElGamalPubkey,
+    },
 };
 
 /// Maximum bit length of any deposit or transfer amount
 ///
 /// Any deposit or transfer amount must be less than 2^48
-pub const MAXIMUM_DEPOSIT_TRANSFER_AMOUNT_BIT_LENGTH: usize = 48;
+pub const MAXIMUM_DEPOSIT_TRANSFER_AMOUNT: u64 = (u16::MAX as u64) + (2 << 16) * (u32::MAX as u64);
```
joncinque, line 22: If the max is `2^48 - 1`, shouldn't this be?
```suggestion
pub const MAXIMUM_DEPOSIT_TRANSFER_AMOUNT: u64 = (u16::MAX as u64) + (1 << 16) * (u32::MAX as u64);
```

Otherwise you're actually doing `2^49 - 1` -- I tried this out to be sure :sweat_smile:

joncinque (commented):
R+ after one last comment

token/program-2022/src/extension/confidential_transfer/mod.rs
samkim-crypto: Oh oops! Yes, this actually seems to be a weird 49-bit number with all 1's except the 17th position :sweat_smile:

PR #4679, opened 2023-07-04
Upgrade to solana 1.16.3

samkim-crypto:


joncinque (commented):
You'll also need to commit the Cargo.lock changes, and be sure to maintain borsh's dependency on hashbrown to 0.12.3, otherwise on-chain programs can't build.

Otherwise, this is good to go!

KirillLykov:
I think this change is relevant -- https://github.com/solana-labs/solana-program-library/pull/4703
I'm not sure if sticking with particular version of hashbrown is better than adding getrandom or not. If this getrandom custom doesn't break anything, than I would argue that sooner or later we will have to upgrage hashbrown and, hence, getrandom. Maybe there is something else I'm not aware of

PR #4698, opened 2023-07-05
ci: Add all Solana crates to update-solana-dependencies.sh

joncinque:
#### Problem

`update-solana-dependencies.sh` misses some of the new crates, like `solana-rpc-client`, like during the upgrade to 1.16.2 https://github.com/solana-labs/solana-program-library/pull/4679/files#diff-a160be2082921ef5bd4bc4ee704c95cf79cb248ad2acf9e1d0b2bfc53206e469R15

#### Solution

Update the update script with all of the crates in the patch script.

### DIFF
--- a/update-solana-dependencies.sh
+++ b/update-solana-dependencies.sh
@@ -22,25 +22,62 @@ while IFS='' read -r line; do tomls+=("$line"); done < <(find . -name Cargo.toml
 crates=(
   solana-account-decoder
   solana-banks-client
+  solana-banks-interface
   solana-banks-server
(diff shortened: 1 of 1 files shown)


samkim-crypto (approved):
Thanks!

update-solana-dependencies.sh
```diff
@@ -22,25 +22,62 @@ while IFS='' read -r line; do tomls+=("$line"); done < <(find . -name Cargo.toml
 crates=(
```
buffalojoec, line 22: heh, this file is nifty

PR #4703, opened 2023-07-05
Add getrandom dependency

KirillLykov:
### Problem

`borsh` update lead to `hashbrown` update which adds dependency on `getrandom`.
All this lead to turning off spl downstream build in solana ci.
Details are in https://github.com/solana-labs/solana/issues/32378

### Solution

Following [this giude](https://github.com/solana-labs/solana/blob/52fe7eb1cf1dd48893dece4e947987dfca192fac/docs/src/developing/on-chain-programs/developing-rust.md?plain=1#L240), add explicit dependency on getrandom where needed with custom features


KirillLykov:
solved in https://github.com/solana-labs/solana/pull/32382

PR #4733, opened 2023-07-08
[token-2022] Update confidential transfer withdraw for Solana 1.16

samkim-crypto:
#### Problem
Confidential transfer withdraw instruction is not u9pdated for the new Solana 1.16 changes.

#### Summary of Changes
Updated the processor, client, and tests relating to the withdraw instruction. These are relatively simpler changes with just two commits. The withdraw instruction does fit into a single transaction, so I did not divide up the proof.

Some notable change is in the error. In the client, I changed the error variant `Proof(ProofError)` to `ProofGeneration`. Previously, `Proof(ProofError)` held the specific proof error from the zk-token-sdk. Now, I simplified it to just `ProofGeneration` because that is technically the only type of proof related error in the client and since the proof generation logic for the withdraw proof is now shoved into token-2022, propagating the error further could be a little complicated. This technically changes the error interface, so I am not sure if it is better to deprecate the variant and create a new variant.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -22,17 +22,20 @@ use {
         extension::{
             confidential_transfer::{
                 self, ApplyPendingBalanceAccountInfo, ConfidentialTransferAccount,
+                WithdrawAccountInfo,
             },
             cpi_guard, default_account_state, interest_bearing_mint, memo_transfer,
             metadata_pointer, transfer_fee, transfer_hook, BaseStateWithExtensions, ExtensionType,
             StateWithExtensionsOwned,
         },
         instruction, offchain,
-        solana_zk_token_sdk::encryption::{
-            auth_encryption::AeKey,
-            elgamal::{ElGamalKeypair, ElGamalSecretKey},
+        solana_zk_token_sdk::{
+            encryption::{
+                auth_encryption::AeKey,
+                elgamal::{ElGamalKeypair, ElGamalSecretKey},
+            },
+            zk_token_elgamal::pod::ElGamalPubkey,
         },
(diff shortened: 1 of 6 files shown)


token/client/src/token.rs
```diff
@@ -68,8 +71,8 @@ pub enum TokenError {
     AccountInvalidAssociatedAddress,
     #[error("invalid auxiliary account address")]
     AccountInvalidAuxiliaryAddress,
-    #[error("proof error: {0}")]
-    Proof(ProofError),
+    #[error("proof generation")]
+    ProofGeneration,
```
joncinque, line 75: Are there plans for token-client to do more with proofs than just generation? Ie is it also going to decrypt / verify / deserialize and all that? If the answer is "yes", I would lean towards keeping the error as it is, and have functions like `generate_proof_data` functions return `ProofError` instead of `TokenError`. That way you don't need `TokenError::ProofGeneration`. What do you think?

joncinque (commented):
Everything else looks good, just the one question about the ProofError

token/client/src/token.rs
samkim-crypto: Yeah, the client would want to decrypt account data (`TokenError::AccountDecryption`) and generate proofs (`TokenError::ProofGeneration`), but I can't really think of a case where it would verify proofs.

joncinque: Per offline conversation, this approach is ok since the other `ProofError`s are either no longer used or already covered

joncinque (approved):
R+ after `token22-confidential-transfer-deposit` goes in

PR #4742, opened 2023-07-10
token-2022: Add scaffolding to implement token-metadata interface

joncinque:
#### Problem

We want to implement the token-metadata interface in token-2022, but we've only laid the groundwork for handling variable-length extensions, and none of the actual implementation.

#### Solution

This one's pretty simple, just putting in the scaffolding for handling a token-metadata instruction, no real functional bits yet:

* adds `ExtensionType::TokenMetadata` and marks it as variable-length
* parses incoming instructions as `TokenMetadataInstruction` *after* the token-2022 instructions -- this is the bulk of the change, and it's all white-space
* forwards `TokenMetadataInstruction` to specific processors, which are all empty

### DIFF
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -18,7 +18,7 @@ use {
             mint_close_authority::MintCloseAuthority,
             non_transferable::{NonTransferable, NonTransferableAccount},
             permanent_delegate::{get_permanent_delegate, PermanentDelegate},
-            reallocate,
+            reallocate, token_metadata,
             transfer_fee::{self, TransferFeeAmount, TransferFeeConfig},
             transfer_hook::{self, TransferHook, TransferHookAccount},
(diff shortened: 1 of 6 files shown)


token/program-2022/src/processor.rs
```diff
@@ -1465,188 +1466,206 @@ impl Processor {
...
+                }
+                TokenInstruction::MetadataPointerExtension => {
+                    metadata_pointer::processor::process_instruction(
+                        program_id,
+                        accounts,
+                        &input[1..],
+                    )
+                }
             }
+        } else if let Ok(instruction) = TokenMetadataInstruction::unpack(input) {
```
buffalojoec, line 1665: Sick! This line here is literally keying on the interface's instruction set (imported). So clean 🧼

buffalojoec (approved):

PR #4766, opened 2023-07-11
discriminator: Remove dependency on "full" feature

joncinque:
#### Problem

There are many recompilations when working on spl crates, starting with `spl-discriminator-syn`, probably due to crates that are recompiled with different feature sets.

#### Solution

I tried a few different things, including specifying all `syn` usage to use the features `full` and `extra-traits`, to line up with solana-sdk-macro. I also tried to remove the dependency on the full in one crate, which also didn't really have an effect.

This is the result of that second effort. Unfortunately, it means that generics, lifetimes, and where clauses are no longer supported. The `ItemEnum` and `ItemStruct` parsers are really nifty!

### DIFF
--- a/libraries/discriminator/syn/src/lib.rs
+++ b/libraries/discriminator/syn/src/lib.rs
@@ -7,11 +7,11 @@ mod error;
 pub mod parser;
 
 use {
-    crate::{error::SplDiscriminateError, parser::parse_hash_input},
+    crate::parser::parse_hash_input,
     proc_macro2::{Span, TokenStream},
     quote::{quote, ToTokens},
     solana_program::hash,
-    syn::{parse::Parse, Generics, Ident, Item, ItemEnum, ItemStruct, LitByteStr, WhereClause},
+    syn::{ext::IdentExt, parse::Parse, Attribute, Ident, LitByteStr, Token, Visibility},
 };
 
 /// "Builder" struct to implement the `SplDiscriminate` trait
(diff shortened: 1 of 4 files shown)


libraries/discriminator/syn/src/lib.rs
```diff
@@ -7,74 +7,43 @@ mod error;
...
+        }
+        let ident = input.parse::<Ident>()?;
+        // just consume the rest
+        input.step(|cursor| {
+            let mut rest = *cursor;
+            while let Some((_, next)) = rest.token_tree() {
+                rest = next;
+            }
+            Ok(((), rest))
+        })?;
```
buffalojoec, line 45: Once we have the `Ident` can't we just return?

buffalojoec (commented):
So where does this leave us then? 😅

libraries/discriminator/syn/src/lib.rs
joncinque: Nope! `parse_macro_input` will barf if the whole thing isn't consumed, saying `{` wasn't expected.

joncinque:
> So where does this leave us then?

I'd prefer to not merge this to be honest, but you said you wanted to see it. We can keep it up, and maybe if we do the same thing with `spl-program-error`, we can avoid those recompilations.

But I think the better approach is to figure out the superset of features to enable on our crates everywhere.

Maybe I'll just close this

buffalojoec:
> > So where does this leave us then?
> 
> I'd prefer to not merge this to be honest, but you said you wanted to see it. We can keep it up, and maybe if we do the same thing with `spl-program-error`, we can avoid those recompilations.
> 
> But I think the better approach is to figure out the superset of features to enable on our crates everywhere.
> 
> Maybe I'll just close this

Happy to dive into this tonight so we can get it ironed out!

joncinque:
> Happy to dive into this tonight so we can get it ironed out!

Honestly, there's no rush. I'll put up an issue about it and we can figure it out when we have time

joncinque:
Closing this since we have an issue at #4769

PR #4767, opened 2023-07-11
token-client: Add simulation support to ProgramClient

joncinque:
#### Problem

While working on the token-metadata `emit` instruction, which writes the token-metadata bytes into return data, I realized that we can't simulate transactions using the token-client.

#### Solution

Support transaction simulation! I went with a separate set of traits rather than adding simulation to the `SendTransaction` trait, in the case that we want to add tpu-client support. In that case, it would be a new client that *does not* support simulation, so it wouldn't implement those traits.

Since the RPC simulation results don't have any normal formatters, I just went with the default `Debug` one. Where would be the best place to implement that?

I'm not sure if either of you was already thinking about this, but I hope it lines up with your expectations! If not, I'm happy to drop it and go another route.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -1,5 +1,5 @@
 use {
-    crate::client::{ProgramClient, ProgramClientError, SendTransaction},
+    crate::client::{ProgramClient, ProgramClientError, SendTransaction, SimulateTransaction},
     futures_util::TryFutureExt,
     solana_program_test::tokio::time,
     solana_sdk::{
@@ -319,7 +319,7 @@ fn native_mint_decimals(program_id: &Pubkey) -> u8 {
 
 impl<T> Token<T>
 where
-    T: SendTransaction,
+    T: SendTransaction + SimulateTransaction,
 {
     pub fn new(
         client: Arc<dyn ProgramClient<T>>,
@@ -510,6 +510,21 @@ where
             .map_err(|error| TokenError::Client(error.into()))?;
 
         Ok(transaction)
+    }
+
+    pub async fn simulate_ixs<S: Signers>(
+        &self,
+        token_instructions: &[Instruction],
+        signing_keypairs: &S,
+    ) -> TokenResult<T::SimulationOutput> {
+        let transaction = self
+            .construct_tx(token_instructions, signing_keypairs)
+            .await?;
+
+        self.client
+            .simulate_transaction(&transaction)
+            .await
+            .map_err(TokenError::Client)
     }
 
     pub async fn process_ixs<S: Signers>(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7010,9 +7010,11 @@ version = "0.5.1"
 dependencies = [
  "async-trait",
  "futures-util",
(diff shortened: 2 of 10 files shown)


2501babe:
not only had i been thinking about it but i was expecting to pr it eventually but didnt want to mess up the box future pin etc stuff for lack of understanding

will review in full soon but is there a reason not to let a future tpu client error if simulate is called like how the sign-only sender errors if send is called?

also this is what i used for a prettyprinter in my own cli:
```
        let simulation_data = config.rpc_client.simulate_transaction(&transaction).await?;

        if config.verbose() {
            if let Some(logs) = simulation_data.value.logs {
                for log in logs {
                    println!("    {}", log);
                }
            }       

            println!(
                "\nSimulation succeeded, consumed {} compute units",
                simulation_data.value.units_consumed.unwrap()
            );      
        } else {
            println_display(config, "Simulation succeeded".to_string());
        }
```

joncinque:
> will review in full soon but is there a reason not to let a future tpu client error if simulate is called like how the sign-only sender errors if send is called?

The sign-only sender actually just returns the transaction so that we can dump it right after. But it's a good point, we could just error for the tpu-client.

The whole token-client crate is so thoughtful with the typing that it seemed like an antipattern to do that though!  I was thinking it could be nice to be able to specify when you need just a `SendTransaction`, and when you need `SendTransaction + SimulateTransaction`, but maybe that's too confusing for users.

> also this is what i used for a prettyprinter in my own cli:

Ah thanks, I might rip that off and add the other fields. I can put in a monorepo PR for it and also do it in token-client in the meantime.

2501babe:
>The whole token-client crate is so thoughtful with the typing that it seemed like an antipattern to do that though!
this is a good point yea 

>specify when you need just a SendTransaction, and when you need SendTransaction + SimulateTransaction, but maybe that's too confusing for users
when we move this to monorepo on thing ill be thinking about is helper functions and exporting type aliases. i already do this in my tests for convenience: `type PClient = Arc<dyn ProgramClient<ProgramRpcClientSendTransaction>>;`

token/client/src/token.rs
```diff
@@ -502,6 +502,21 @@ where
         Ok(transaction)
     }
 
+    pub async fn simulate_ixs<S: Signers>(
```
CriesofCarrots, line 505: Oh dear, I hate that we didn't spell out `instructions` below. I guess we have to stay consistent?
I also dislike `construct_tx`, incidentally.

CriesofCarrots (approved):
I like the trait approach, personally. Lgtm, but do get ✅ from 2501babe, who's been in the trenches with this functionality more and more recently than I have.

2501babe (approved):

2501babe:
lgtm, when this lands ill update my cli to use it

2501babe:
fixed the conflict on Cargo.toml because of the upgrade to 1.16.3

joncinque:
> lgtm, when this lands ill update my cli to use it

Sorry, this kinda fell off my radar, thanks for helping! I've rebased and will merge once it passes CI

PR #4768, opened 2023-07-11
TLV Library Expansion

buffalojoec:
Related to conversations around ideas for TLV from the past week or so.
I can break this up if it's too much all at once, but mainly wanted you to check out the "non-strict" approach to TLV entries.

This PR aims to expand the functionality of the general-use TLV state management library `spl-type-length-value`.

In it I'm proposing a set of new features:
- The [`SplBorshVariableLenPack` derive macro](https://github.com/solana-labs/solana-program-library/pull/4768/files#diff-39641fcb5374c9ddd0214138ec71f71632f8ccf896670dc86d4ab476aaaa55b0R70-R91) for implementing `VaribleLenPack` with borsh-implemented types
- `strict` and `non-strict` APIs for `TlvState` and associated types
    - I just split your original `TlvState` into two traits: `TlvStateStrict` and `TlvStateNonStrict`
- The [`TlvIterator` custom iterator](https://github.com/solana-labs/solana-program-library/pull/4768/files#diff-661ea247691b8f67ae907bd8ac0f5f29190f5bc258b9926f5c8a51beccda776bR906-R1014) for iterating over TLV state (inspired by comment on `get_discriminators_and_end_index(...)`)

All major changes are in [`libraries/type-length-value/src/state.rs`](https://github.com/solana-labs/solana-program-library/pull/4768/files#diff-661ea247691b8f67ae907bd8ac0f5f29190f5bc258b9926f5c8a51beccda776b). Everything else is mostly just updating the struct names.


joncinque (commented):
It looks like there's three separate PRs in this, so maybe we can split them up and discuss each there? The PRs would be:

* borsh macro: to be honest I'm not keen to add a dependency on borsh in TLV, but we could add it in another package, like `spl-borsh` or something
* iterator: this is really cool! My one note is that we shouldn't expose `TlvIndices`, and instead the iterator gives the type as a `Discriminator`, length as `Length`, and value bytes as the byte slice. Also, it would be great if functions that call `get_indices` can use the iterator instead under the hood
* allowing duplicate types: this turned out much neater than expected! Rather than create a whole set of new types that I'll definitely forget to update properly, can we just add a new API for functions that permit duplicates and have the old API call into that? for example, `get_repeated_extension` or `get_extension_with_duplicate_type`?

buffalojoec:
This got worked in through different PRs!

PR #4772, opened 2023-07-12
[token-2022] Add serde support for the rest of token-2022 instructions

serbangv:
Issue #3325

Would appreciate someone to have a look over this PR before I move on with this approach. Thanks.


fromanother:
@joncinque  Can offer some feedback on this?

token/program-2022/src/pod.rs
```diff
@@ -95,16 +110,69 @@ impl From<OptionalNonZeroElGamalPubkey> for Option<ElGamalPubkey> {
...
+        let bytes = BASE64_STANDARD.decode(v).map_err(de::Error::custom)?;
+
+        if bytes.len() != 32 {
+            return Err(de::Error::custom("Length of base64 decoded bytes is not 32"));
+        }
+
+        let mut array = [0; 32];
+        array.copy_from_slice(&bytes[0..32]);
+
+        let elgamal_pubkey = ElGamalPubkey(array);
```
buffalojoec, line 150: Do you need this copy here? Or can we just do `ElGamalPubkey::from_bytes(&bytes[..32])`?

token/program-2022/src/pod.rs
```diff
@@ -95,16 +110,69 @@ impl From<OptionalNonZeroElGamalPubkey> for Option<ElGamalPubkey> {
...
+            s.serialize_some(&self.0.to_string())
+        }
+    }
+}
+
+#[cfg(feature = "serde-traits")]
+struct OptionalNonZeroElGamalPubkeyVisitor;
+#[cfg(feature = "serde-traits")]
+impl<'de> Visitor<'de> for OptionalNonZeroElGamalPubkeyVisitor {
+    type Value = OptionalNonZeroElGamalPubkey;
```
buffalojoec, line 131: Curious why the need for a new struct instead of implementing `Visitor` directly on `OptionalNonZeroElGamalPubkey` and making `Value` of type `String` ?

buffalojoec (commented):
I think this is a pretty good start! Thanks for picking this up.

Assuming you plan to keep going on these?

token/program-2022/src/pod.rs
```diff
@@ -95,16 +110,69 @@ impl From<OptionalNonZeroElGamalPubkey> for Option<ElGamalPubkey> {
...
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("an ElGamal public key as base64")
+    }
+
+    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+    {
+        let bytes = BASE64_STANDARD.decode(v).map_err(de::Error::custom)?;
```
joncinque, line 141: By convention, we typically use base58 for pubkeys. Is there a reason to use base64 for these?

token/program-2022/src/pod.rs
```diff
@@ -117,6 +185,31 @@ impl From<PodBool> for bool {
...
+#[cfg(feature = "serde-traits")]
+impl<'de> Deserialize<'de> for PodBool {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: Deserializer<'de>,
+    {
+        let serde_bool = bool::deserialize(deserializer)?;
+        Ok(PodBool::from(serde_bool))
+    }
+}
```
joncinque, line 208: Rather than implementing serialize and deserialize by hand, would it be possible to use a `from` and `into` bound whenever the field is encountered? ie:

```
#[serde(from = "bool", into = "bool")] 
my_field: PodBool
```

joncinque (commented):
Just a couple more comments -- this is on the right track!

token/program-2022/src/pod.rs
fromanother: @buffalojoec 

I implemented `Visitor` on a new struct because as far as I can tell it's the common serde pattern but I can move it on `OptionalNonZeroElGamalPubkey` no problem.

As far as making `Value` of type `String`, could you help me by elaborating a bit? The deserializer converts serialized data into `OptionalNonZeroElGamalPubkey` so this is why I set `type Value = OptionalNonZeroElGamalPubkey;`.

token/program-2022/src/pod.rs
fromanother: I will switch to base58. 
The reason I used base64 because that's how `fmt::Display` was implemented for it in the `solana-zk-token-sdk` crate.

joncinque: Ah I see, you might be right then! Let's see what @samkim-crypto thinks, since he did the work on `solana-zk-token-sdk`

samkim-crypto: Yes, base58 is typicaly used for public keys and should really be used for ElGamal public keys as well. But currently, base64 is used for `fmt::Display` in the `solana-zk-token-sdk`, so here, I think it would make sense to stick with base64.

joncinque: Since this is meant to be consumed by external users, maybe we should do whatever makes the most sense at the top-level, since we can easily fixup `solana-zk-token-sdk` right away.

@jordansexton from a wallet perspective, do you think ElGamal keys for encryption should use base58 or base64?

token/program-2022/src/pod.rs
buffalojoec: Actually I think what you have is correct, my bad! (I had to go read some more on `Visitor`) 
Resolving

token/program-2022/src/pod.rs
serbangv: Another `solana-zk-token-sdk` type that is used in instruction data is `AeCiphertext` and it also uses base64. Just FYI.

joncinque: We discussed this a bit and decided that base64 is ok https://discord.com/channels/428295358100013066/774014770402689065/1138612960332349532 -- all good then!

token/program-2022/src/pod.rs
```diff
@@ -6,11 +6,26 @@ use {
...
 /// A Pubkey that encodes `None` as all `0`, meant to be usable as a Pod type,
 /// similar to all NonZero* number types from the bytemuck library.
+#[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
 #[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
 #[repr(transparent)]
-pub struct OptionalNonZeroPubkey(Pubkey);
+pub struct OptionalNonZeroPubkey(
+    #[cfg_attr(feature = "serde-traits", serde(with = "As::<DisplayFromStr>"))]
+    Pubkey
+);
```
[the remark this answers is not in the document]
serbangv: Updated, please check

token/program-2022/src/pod.rs
serbangv: `pod::ElGamalPubkey` does not implement `from_bytes` but `decoded::ElGamalPubkey` (the non-POD ElGamalPubkey type) does. So I could use `decoded::ElGamalPubkey::from_bytes()` to create a new `decoded::ElGamalPubkey` instance, and then finally convert that to `pod::ElGamalPubkey`. I could do that, but `decoded::ElGamalPubkey::from_bytes()` also uses a copy, so the benefit of doing it might not be there.

Note that I move the code your comment was made under here: token/program-2022/src/serialization.rs in `fn base64_to_bytes`

token/program-2022/src/pod.rs
serbangv: Good catch, updated code, please check

serbangv:
This is ready for review, no longer WIP.

Quick recap & summary: 
* this PR is for issue #3325 
* in it I added serde's Serialize & Deserialize to data types that are passed in instructions
* I covered the remaining instructions not included in the original PR, https://github.com/solana-labs/solana-program-library/pull/3291

Notes:
* in serialization.rs I followed the pattern started by the prev PR, and everything is a module inside that file
* token/program-2022/src/extension/token_metadata/ -- haven't implemented anything for the token metadata instructions processed here, should I?

token/program-2022/src/serialization.rs
```diff
@@ -1,4 +1,22 @@
...
+/// helper function to convert base64 encoded string into a bytes array
+fn base64_to_bytes<const N: usize, E: Error>(v: &str) -> Result<[u8; N], E> {
+    let bytes = BASE64_STANDARD.decode(v).map_err(Error::custom)?;
+
+    if bytes.len() != N {
+        return Err(Error::custom(format!("Length of base64 decoded bytes is not {}", N)));
+    }
+
+    let mut array = [0; N];
+    array.copy_from_slice(&bytes[0..N]);
```
joncinque, line 17: nit: How about using `decode_slice` on the pre-allocated array rather than getting a vector and then writing it to an array? https://docs.rs/base64/latest/base64/engine/trait.Engine.html#method.decode_slice

token/program-2022/src/serialization.rs
```diff
@@ -76,3 +94,174 @@ pub mod coption_fromstr {
...
+                E: Error,
+        {
+            OptionalNonZeroPubkey::try_from(None)
+                .map_err(|e| Error::custom(e.to_string()))
+        }
+    }
+
+    /// Visitor for deserializing OptionalNonZeroElGamalPubkey
+    pub(crate) struct OptionalNonZeroElGamalPubkeyVisitor;
+    const OPTIONAL_NZ_ELG_LEN: usize = 32;
```
joncinque, line 242: nit: please spell this out in its entirety, otherwise I'll always wonder what New Zealand has to do with this code
```suggestion
    const OPTIONAL_NONZERO_ELGAMAL_PUBKEY_LEN: usize = 32;
```

Or you can just call it `ELGAMAL_PUBKEY_LEN` again

token/program-2022/src/pod.rs
```diff
@@ -60,6 +73,30 @@ impl From<OptionalNonZeroPubkey> for COption<Pubkey> {
...
+
+#[cfg(feature = "serde-traits")]
+impl<'de> Deserialize<'de> for OptionalNonZeroPubkey {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: Deserializer<'de>,
+    {
+        deserializer.deserialize_any(OptionalNonZeroPubkeyVisitor)
+    }
+}
```
joncinque, line 98: Is there a reason that the serde traits are implemented here rather than in `serialization.rs`? It seems like it would be cleaner to follow the same model as everywhere else by defining the type + visitor there and then using `serde(with = ...`

token/program-2022/src/pod.rs
```diff
@@ -95,16 +132,43 @@ impl From<OptionalNonZeroElGamalPubkey> for Option<ElGamalPubkey> {
...
+
+#[cfg(feature = "serde-traits")]
+impl<'de> Deserialize<'de> for OptionalNonZeroElGamalPubkey {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: Deserializer<'de>,
+    {
+        deserializer.deserialize_any(OptionalNonZeroElGamalPubkeyVisitor)
+    }
+}
```
joncinque, line 157: Same here, can we move the implementation in `serialization.rs`?

token/program-2022/tests/serialization.rs
```diff
@@ -1,11 +1,27 @@
...
+                AeCiphertext,
+                ElGamalPubkey,
+            },
+        },
+    },
+    std::str::FromStr,
 };
 
 #[test]
+// tests serde of ix containing COption and Pubkey
```
joncinque, line 24: nit: can you move all of the pre-test comments into the test or rename the test if the name isn't clear enough? I usually prefer long test names for self-documentation if they ever fail, so for example I'd call this one `serde_instruction_coption_pubkey`

token/program-2022/tests/serialization.rs
```diff
@@ -35,3 +52,115 @@ fn token_program_serde_with_none() {
...
+    use spl_token_2022::extension::confidential_transfer_fee::{
+        instruction::InitializeConfidentialTransferFeeConfigData
+    };
+
+    let withdraw_withheld_authority_elgamal_pubkey = ElGamalPubkey([162, 23, 108, 36, 130, 143,
+        18, 219, 196, 134, 242, 145, 179, 49, 229, 193, 74, 64, 3, 158, 68, 235, 124, 88, 247, 144,
+        164, 254, 228, 12, 173, 85]);
+
+    let inst = InitializeConfidentialTransferFeeConfigData {
+        authority: Default::default(),
```
joncinque, line 137: nit: can you avoid `Default::default()` and spell out the type?
```suggestion
        authority: OptionalNonZeroPubkey::default(),
```

joncinque (commented):
Really great work! Mostly some small nits and suggestions about moving things.

Also, be sure to run `cargo fmt` to format the code and `cargo +nightly-2023-04-19 clippy -Zunstable-options --all-targets` to check for lints.

joncinque:
> token/program-2022/src/extension/token_metadata/ -- haven't implemented anything for the token metadata instructions processed here, should I?

We can take care of that in a separate PR, this one's already getting pretty big.

token/program-2022/src/serialization.rs
serbangv: Done

token/program-2022/src/pod.rs
serbangv: I was trying to use `serde_with` only when working with external types since I cannot `impl Serialize`/`impl Deserialize` on them as they are imported from external crates. 

Another thing is that for `OptionalNonZeroPubkey` it's inner `Pubkey` is private which makes it harder to instantiate it from a separate module. Currently in it's `Visitor` in `serialization.rs` I am using  `OptionalNonZeroPubkey::try_from(Some(pkey))` to instantiate it.

If you prefer I can move the current serialization for `OptionalNonZeroPubkey`/`OptionalNonZeroElGamalPubkey` into `serialization.rs` and then use `serde(with = ...`, but if you could advise what is the best way to instantiate them since their inner `Pubkey` is private.

token/program-2022/tests/serialization.rs
serbangv: Updated, please check

token/program-2022/tests/serialization.rs
serbangv: Done

serbangv:
> Really great work! I especially appreciate all of the careful tests. Mostly some small nits and suggestions about moving things.

Glad to contribute, appreciate you and @buffalojoec guiding me.

token/program-2022/src/pod.rs
joncinque: Ah I see what you mean. Makes sense to me, thanks for the explanation!

token/program-2022/src/serialization.rs
joncinque: Ah I see what you mean, your solution makes the most sense. Thanks for the explanation!

joncinque:
Sorry, looks like you'll just need to rebase onto master to pick up the audit fix, then this should be good to go!

serbangv:
@joncinque Branch updated. Anything else needed before this can be merged? 

Also, I think I will look into the `token_metadata` instructions ([mentioned here](https://github.com/solana-labs/solana-program-library/pull/4772#issuecomment-1675541871)) after this PR is done, unless you have a different suggestion.

joncinque (approved):
This is finally ready to go in, thanks so much!

Thank you

joncinque:
If you have the desire and the time to work on the metadata instructions, that would be much appreciated!

Issue #4648, opened 2023-06-28
token-2022: Implement token-metadata interface


joncinque:
#### Problem

People need to use token metadata with token-2022.

#### Solution

The token-metadata interface was added to partly address that through external programs, but the interface can be implemented by any program, including directly by token-2022, so we're going to implement that whole interface in token-2022!

Here are the steps:

- [x] #4646
- [x] #4661
  * add `alloc` in the public interface, for allocating bytes to an extension
  * add `UnsizedExtension` trait, which specifies that an extension can use `alloc` and `realloc`. This is for compile-time safety, to keep using the in-place `get_extension` for all other extensions
  * add `realloc`, for an existing extension (ripping off https://github.com/solana-labs/solana-program-library/blob/ed8818c53438b32f96a77f86752e98db02a764ef/libraries/type-length-value/src/state.rs#L358)
  * add helpers for fetching the sizes of all TLV entries, along with calculating a new size for a realloc
  * add `realloc` for the whole account (ripping off https://github.com/solana-labs/solana-program-library/blob/ed8818c53438b32f96a77f86752e98db02a764ef/libraries/type-length-value/src/state.rs#L414)
  * add init helper for unsized extension: given an instance of an unsized extension, allocate the bytes for that instance and write them into the TLV buffer
  * add realloc helper for unsized extension: given an instance of an unsized extension, reallocate the TLV *and* account bytes in order to fit the new instance 
- [x] #4744
- [x] #4745
- [x] #4776
- [x] #4774
- [x] #4773

joncinque:
This is now done, token-2022 implements the spl-token-metadata-interface !

PR #4773, opened 2023-07-12
token-2022: Implement `TokenMetadataInstruction::Emit`

joncinque:
#### Problem

As part of #4648, token-2022 needs to implement `TokenMetadataInstruction::Emit` to be a token-metadata program.

#### Solution

Very similar to the example implementation https://github.com/solana-labs/solana-program-library/blob/142708116b57ae11bcfc33758962059c7ddf11c6/token-metadata/example/src/processor.rs#L175-L192

And tests (https://github.com/solana-labs/solana-program-library/blob/master/token-metadata/example/tests/emit.rs

Implement them!

### DIFF
--- a/token/program-2022-test/tests/token_metadata_emit.rs
+++ b/token/program-2022-test/tests/token_metadata_emit.rs
@@ -1,0 +1,140 @@
+#![cfg(feature = "test-sbf")]
+
+mod program_test;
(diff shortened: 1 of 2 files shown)


buffalojoec (approved):
lgtm !

PR #4774, opened 2023-07-12
token-2022: Implement `TokenMetadataInstruction::UpdateAuthority`

joncinque:
#### Problem

As part of #4648, token-2022 needs to implement `TokenMetadataInstruction::UpdateAuthority` to be a token-metadata program.

#### Solution

Very similar to the example implementation https://github.com/solana-labs/solana-program-library/blob/142708116b57ae11bcfc33758962059c7ddf11c6/token-metadata/example/src/processor.rs#L149-L172

And tests (https://github.com/solana-labs/solana-program-library/blob/master/token-metadata/example/tests/update_authority.rs

Implement them!

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -2701,4 +2701,23 @@ where
         ));
         self.process_ixs(&instructions, signing_keypairs).await
     }
+
+    /// Update the token-metadata authority in a mint
+    pub async fn update_authority_in_token_metadata<S: Signers>(
+        &self,
+        current_authority: &Pubkey,
+        new_authority: Option<Pubkey>,
+        signing_keypairs: &S,
+    ) -> TokenResult<T::Output> {
+        self.process_ixs(
+            &[spl_token_metadata_interface::instruction::update_authority(
+                &self.program_id,
+                &self.pubkey,
+                current_authority,
+                new_authority.try_into()?,
+            )],
+            signing_keypairs,
+        )
+        .await
+    }
 }
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -2703,7 +2703,7 @@ where
     }
 
     /// Update the token-metadata authority in a mint
-    pub async fn update_authority_in_token_metadata<S: Signers>(
+    pub async fn token_metadata_update_authority<S: Signers>(
         &self,
         current_authority: &Pubkey,
         new_authority: Option<Pubkey>,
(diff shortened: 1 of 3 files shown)


token/program-2022/src/extension/token_metadata/processor.rs
```diff
@@ -142,9 +142,26 @@ pub fn process_remove_key(
...
-    _data: UpdateAuthority,
+    accounts: &[AccountInfo],
+    data: UpdateAuthority,
 ) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+    let metadata_info = next_account_info(account_info_iter)?;
+    let update_authority_info = next_account_info(account_info_iter)?;
+
+    // deserialize the metadata, but scope the data borrow since we'll probably
+    // realloc the account
```
buffalojoec, line 153: nit: This comment doesn't make sense if we never need to realloc

token/client/src/token.rs
```diff
@@ -2701,4 +2701,23 @@ where
         ));
         self.process_ixs(&instructions, signing_keypairs).await
     }
+
+    /// Update the token-metadata authority in a mint
+    pub async fn update_authority_in_token_metadata<S: Signers>(
```
buffalojoec, line 2706: nit: Does this function name sound a little strange?

I guess no stranger than `update_token_metadata_update_authority`, but the latter sounds a bit more like what folks are used to.

Lmk what you think

buffalojoec (commented):
lgtm just 2 little nits/questions

token/program-2022/src/extension/token_metadata/processor.rs
joncinque: Ah nice call, thanks!

token/client/src/token.rs
joncinque: I wasn't sure for these. I was thinking it either goes "do_thing_in_token_metadata" or "token_metadata_do_thing", so for the second one, we'd have:

* `token_metadata_initialize`
* `token_metadata_update_field`
* `token_metadata_emit`
* `token_metadata_update_authority`
* `token_metadata_remove_key`

Maybe those are better now that I write them out. What do you think?

buffalojoec: Not a hill I'd die on but I will say those do mimic the Rust full path quite nicely, ie:

```rust
token_metadata::instruction::Initialize,
token_metadata::instruction::UpdateAuthority,
...
```

It only just jumped out at me with update authority (not the prior two PRs)

joncinque: works for me!

buffalojoec (approved):

PR #4610, opened 2023-06-26
Add support for required proposal signatories

wizzeh:
This PR adds support for specifying signing addresses that are required to move a proposal to the voting stage. This is determined at the governance level, so all proposals belonging to a particular governance will inherit that governance's set of required signatures.

This mechanism could be used to have trusted administrators of a governance or with program signers to set up some kind of automated constitution.

Here's the flow:
- Call AddRequiredSignatoryToGovernance
- (Create a Proposal)
- Call CreateSignatoryRecordFromGovernance for each signatory
- Each signatory calls SignOffProposal

I've kept the existing sign-off mechanism, based on AddSignatory, where each proposal has at most one delegated sign-off that's set per-proposal, for backwards compatibility reasons.

I did this work several months ago at the suggestion of @SebastianBor. It should still be good to go, though.

PR #4775, opened 2023-07-12
Governance config nonce

wizzeh:
Adds a configuration nonce to Governance.

The idea is that changing the Governance configuration can leave proposals in an invalid state. Currently it's up to the administrators of a Governance to ensure that config changes happen only when there are no active proposals.

This PR adds a new ProposalState, Invalidated, to which Proposals can be moved with `finalize` if they are invalidated by a governance configuration change. If a proposal should be invalidated, actions that advance proposal state result in an error. Completed proposals remain valid.

### DIFF
--- a/governance/program/src/error.rs
+++ b/governance/program/src/error.rs
@@ -498,6 +498,10 @@ pub enum GovernanceError {
     /// Invalid multi choice proposal parameters
     #[error("Invalid multi choice proposal parameters")]
     InvalidMultiChoiceProposalParameters, // 620
+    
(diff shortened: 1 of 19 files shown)


wizzeh:
attn @SebastianBor

wizzeh:
Huh?? I have no idea how the RegisterForRewards commit got in here. Let me take that out...

wizzeh:
https://github.com/solana-labs/solana-program-library/pull/4610 linking to this PR

PR #4777, opened 2023-07-12
token-client: Rename metadata functions

joncinque:
#### Problem

The function names for the token-metadata processors in the token client are kinda goofy, see https://github.com/solana-labs/solana-program-library/pull/4774#discussion_r1261811662

#### Solution

Rename the functions as `token_metadata_instruction` rather than `instruction_in_token_metadata`.

I'll rebase the open PRs onto this after it lands.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -2547,7 +2547,7 @@ where
     }
 
     /// Initialize token-metadata on a mint
-    pub async fn initialize_token_metadata<S: Signers>(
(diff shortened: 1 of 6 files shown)


buffalojoec (dismissed):
yeah, I think these look a lil better

buffalojoec (dismissed):

buffalojoec (approved):
sorry, didn't see new commits!

PR #4778, opened 2023-07-13
ci: Blast directories in 2xTx build

joncinque:
#### Problem

The 2xTx has recently started erroring due to running out of disk space: https://github.com/solana-labs/solana-program-library/actions/runs/5537335983/jobs/10106051181

#### Solution

Just use the solution built during #4505


joncinque:
Welp, the job didn't fail, which is good enough for me!

PR #4779, opened 2023-07-13
[token-2022] Update confidential transfer `Transfer` and `TransferWithFee` for Solana 1.16

samkim-crypto:
#### Problem
Confidential transfer `Transfer` and `TransferWithFee` instructions are not updated for the new Solana 1.16 changes.

#### Summary of changes
- [bdaf1fe](https://github.com/solana-labs/solana-program-library/pull/4779/commits/bdaf1fe9baf70944146f556bc3c98663898ee97b): The logic for the account info types needed for instructions was getting long, so I refactored these from `state.rs` into a separate module

- [e88c9e4](https://github.com/solana-labs/solana-program-library/pull/4779/commits/e88c9e4031e1bd912c3c934a7231f7ae322e0b26): I updated the processor and client logic for the `Transfer` instruction. This should be pretty standard except that I added helper functions that extract the relevant bits from the transfer amount and fee ciphertexts. The transfer amount and fee are encrypted under the source, destination, auditor, and/or withdraw withheld authority ElGamal pubkeys simultaneously. These helper functions extract the relevant components that are relevant for the source, destination, and/or withdraw withheld authority. The helper functions belong more naturally in the zk-token-sdk, but I inlined these for now until these are added on the monorepo side.

- [b07d480](https://github.com/solana-labs/solana-program-library/pull/4779/commits/b07d48076c3126317bbec5896b767ba8fdad1a04): The same set of changes as the previous commit for the `TransferWithFee` instruction.

- [2ce0429](https://github.com/solana-labs/solana-program-library/pull/4779/commits/2ce0429c0c4d63f3c35c500494461022e8902013): Pretty straightforward updates to the `Transfer` instruction tests.

- [8150248](https://github.com/solana-labs/solana-program-library/pull/4779/commits/8150248724406c0323d78039f0609961f98c3a65): The same set of changes as the previous commit for the `Transfer` with memo tests.

For this PR, I did not update the tests for the `TransferWithFee` since that would involve initializing the mint and account for the confidential transfer fee extension, which probably belong in a separate PR.

### DIFF
--- a/token/program-2022-test/tests/confidential_transfer.rs
+++ b/token/program-2022-test/tests/confidential_transfer.rs
@@ -140,7 +140,7 @@ impl ConfidentialTokenAccountMeta {
         }
     }
 
-    #[cfg(all(feature = "zk-ops", feature = "proof-program"))]
+    #[cfg(feature = "zk-ops")]
     async fn new_with_required_memo_transfers<T>(token: &Token<T>, owner: &Keypair) -> Self
     where
         T: SendTransaction,
@@ -1870,24 +1870,21 @@ async fn ct_withdraw_withheld_tokens_from_accounts() {
         .await;
 }
 
-#[cfg(all(feature = "zk-ops", feature = "proof-program"))]
+#[cfg(feature = "zk-ops")]
 #[tokio::test]
-async fn ct_transfer_memo() {
-    let ConfidentialTransferMintWithKeypairs {
-        ct_mint,
-        ct_mint_transfer_auditor_elgamal_keypair,
-        ..
-    } = ConfidentialTransferMintWithKeypairs::new();
+async fn confidential_transfer_transfer_memo() {
+    let authority = Keypair::new();
+    let auto_approve_new_accounts = true;
+    let auditor_elgamal_keypair = ElGamalKeypair::new_rand();
+    let auditor_elgamal_pubkey = (*auditor_elgamal_keypair.pubkey()).into();
+
     let mut context = TestContext::new().await;
     context
         .init_token_with_mint(vec![
             ExtensionInitializationParams::ConfidentialTransferMint {
-                authority: ct_mint.authority.into(),
-                auto_approve_new_accounts: ct_mint.auto_approve_new_accounts.try_into().unwrap(),
-                auditor_elgamal_pubkey: ct_mint.auditor_elgamal_pubkey.into(),
-                withdraw_withheld_authority_elgamal_pubkey: ct_mint
-                    .withdraw_withheld_authority_elgamal_pubkey
-                    .into(),
+                authority: Some(authority.pubkey()),
+                auto_approve_new_accounts,
+                auditor_elgamal_pubkey: Some(auditor_elgamal_pubkey),
             },
         ])
         .await
@@ -1901,31 +1898,32 @@ async fn ct_transfer_memo() {
         decimals,
         ..
     } = context.token_context.unwrap();
-    let alice_meta =
-        ConfidentialTokenAccountMeta::with_tokens(&token, &alice, &mint_authority, 42, decimals)
-            .await;
+
+    let alice_meta = ConfidentialTokenAccountMeta::new_with_tokens(
+        &token,
+        &alice,
+        &mint_authority,
+        42,
+        decimals,
+    )
+    .await;
+
     let bob_meta =
         ConfidentialTokenAccountMeta::new_with_required_memo_transfers(&token, &bob).await;
-
-    let state = token
-        .get_account_info(&alice_meta.token_account)
-        .await
-        .unwrap();
-    let extension = state
-        .get_extension::<ConfidentialTransferAccount>()
-        .unwrap();
 
     // transfer without memo
     let err = token
         .confidential_transfer_transfer(
             &alice_meta.token_account,
             &bob_meta.token_account,
-            &alice,
-            42, // amount
+            &alice.pubkey(),
             42,
-            &extension.available_balance.try_into().unwrap(),
-            &bob_meta.elgamal_keypair.public,
-            Some(ct_mint_transfer_auditor_elgamal_keypair.public),
+            None,
+            &alice_meta.elgamal_keypair,
+            &alice_meta.aes_key,
+            bob_meta.elgamal_keypair.pubkey(),
+            Some(auditor_elgamal_keypair.pubkey()),
+            &[&alice],
         )
         .await
         .unwrap_err();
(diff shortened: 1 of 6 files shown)


token/program-2022/src/extension/confidential_transfer/account_info.rs
```diff
@@ -0,0 +1,276 @@
+#[cfg(not(target_os = "solana"))]
```
joncinque, line 1: nit for simplicity: does it make sense to mark this whole file with `#![cfg(not(target_os = "solana")))]`? and not have all of the individual annotations? or even mark the module declaration in `confidential_transfer/mod.rs` with the cfg?

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -586,70 +554,123 @@ fn process_transfer(
...
+/// Extract the transfer amount ciphertext encrypted under the source ElGamal public key.
+#[cfg(feature = "zk-ops")]
+fn transfer_amount_source_ciphertext(
+    transfer_amount_ciphertext: &TransferAmountCiphertext,
+) -> ElGamalCiphertext {
+    let transfer_amount_ciphertext_bytes = bytemuck::bytes_of(transfer_amount_ciphertext);
+
+    let mut source_ciphertext_bytes = [0u8; 64];
+    source_ciphertext_bytes[..32].copy_from_slice(&transfer_amount_ciphertext_bytes[..32]);
+    source_ciphertext_bytes[32..].copy_from_slice(&transfer_amount_ciphertext_bytes[32..64]);
```
joncinque, line 617: nit: This is definitely done purposefully  -- can you add comments about specifically which bits you're extracting for these lines?

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -586,70 +554,123 @@ fn process_transfer(
...
+/// Extract the transfer amount ciphertext encrypted under the destination ElGamal public key.
+#[cfg(feature = "zk-ops")]
+fn transfer_amount_destination_ciphertext(
+    transfer_amount_ciphertext: &TransferAmountCiphertext,
+) -> ElGamalCiphertext {
+    let transfer_amount_ciphertext_bytes = bytemuck::bytes_of(transfer_amount_ciphertext);
+
+    let mut destination_ciphertext_bytes = [0u8; 64];
+    destination_ciphertext_bytes[..32].copy_from_slice(&transfer_amount_ciphertext_bytes[..32]);
+    destination_ciphertext_bytes[32..].copy_from_slice(&transfer_amount_ciphertext_bytes[64..96]);
```
joncinque, line 631: Same here, just a little comment about which parts you're copying

token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -1106,7 +1103,7 @@ async fn confidential_transfer_withdraw() {
         .await;
 
     // attempt to withdraw without enough funds
-    let err = token
+    token
```
joncinque, line 1106: Can you check that the error is correct here?

token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -1237,43 +1229,16 @@ async fn ct_transfer() {
...
-        err,
-        TokenClientError::Client(Box::new(TransportError::TransactionError(
-            TransactionError::InstructionError(
-                0,
-                InstructionError::Custom(
-                    TokenError::MaximumPendingBalanceCreditCounterExceeded as u32
-                ),
-            )
-        )))
-    );
```
joncinque, line 1272: This seems like an important case to check, can you add it back in?

joncinque (approved):
Looks really great! Really just a couple of nits to cleanup, but nothing that should stop this from going in after that

token/program-2022/src/extension/confidential_transfer/account_info.rs
samkim-crypto: Ah yes, this will be much simpler! I marked the module declaration to be `#![cfg(not(target_os = "solana")))]`.

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: Done!

token/program-2022-test/tests/confidential_transfer.rs
samkim-crypto: This check is verifying that a confidential transfer fails if the maximum credit limit in the destination account is reached. In practice the default destination will be `2^16`, but in order to test this, we need to set the destination account maximum credit limit to be low (say `2`). 

In order to simplify the tests, to initialize the `alice_meta`, I used the helper constructor  `ConfidentialTokenAccountMeta::new_with_tokens`, which deposits some funds to `alice`'s account. The `new_with_tokens` interface unfortunately does not allow us to set custom maximum credit limit. Instead of adding a new constructor `ConfidentialTokenAccountMeta::new_with_tokens_and_maximum_pending_balance_credit_counter`, I just initialize `bob_meta` with the constructor `ConfidentialTokenAccountMeta::new_with_maximum_pending_balance_credit_counter` and tested that a transfer to `bob`'s account fails if the maximum credit counter is reached.

So in summary, this check on `alice`'s account is not removed, but it is just moved to work on `bob`'s account purely instead for simplicity.

token/program-2022-test/tests/confidential_transfer.rs
samkim-crypto: Yes, now I remember why I didn't check for the error. I forgot to mention it in the PR description :pensive: . For some reason, the error assert does not hold even though both the left and the right hand side of the assert is `TokenClientError::ProofGeneration`. No matter what I try, I can't seem to make the assert to be true.

For testing, I tried changing the returned [error type](https://github.com/solana-labs/solana-program-library/pull/4779/files#diff-465a63e60c1758459d46845310375141037eda9b9043efe9a44f3e55063ac72eR1981) to be any other variant to `TokenClientError` (and correspondingly in the assert statement) and the assert statement holds true. It seems like there is a very weird behavior going on with `TokenClientError::ProofGeneration`. Have you encountered this type of behavior anywhere else?

joncinque: oh, you know what, this might be because of the error equality in `TokenError`, which does a goofy string equality. This is entirely my fault. If you want to implement something better, please do! Either way, your change is good.

joncinque: https://github.com/solana-labs/solana-program-library/blob/4e13218b41211a9a83397b4f6dea0135241a102f/token/client/src/token.rs#L92-L116 just add the missing ones!

joncinque (approved):
You'll probably just need to add those cases for the TokenClientError equality check. Once that's in and this passes CI, this is good to go on my side!

token/program-2022-test/tests/confidential_transfer.rs
samkim-crypto: Oh, this was right below where the error type is define and I missed it! Thanks for the pointer! I think the CI should pass now.

PR #4787, opened 2023-07-14
Bump anchor

ChewingGlass:
Bumping to the latest version of anchor

### DIFF
--- a/libraries/concurrent-merkle-tree/Cargo.toml
+++ b/libraries/concurrent-merkle-tree/Cargo.toml
@@ -12,7 +12,7 @@ log = []
 sol-log = [ "log" ]
 
 [dependencies]
-solana-program = "1.16"
+solana-program = "1.16.3"
 bytemuck = "1.13.1"
 thiserror = "1.0.43"
 
--- a/libraries/concurrent-merkle-tree/Cargo.toml
+++ b/libraries/concurrent-merkle-tree/Cargo.toml
@@ -12,7 +12,7 @@ log = []
 sol-log = [ "log" ]
 
 [dependencies]
-solana-program = "1.10.33"
+solana-program = "1.16"
 bytemuck = "1.13.1"
 thiserror = "1.0.43"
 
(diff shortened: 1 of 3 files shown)


libraries/concurrent-merkle-tree/Cargo.toml
```diff
@@ -12,7 +12,7 @@ log = []
 sol-log = [ "log" ]
 
 [dependencies]
-solana-program = "1.10.33"
+solana-program = "1.16"
```
2501babe, line 15:
```suggestion
solana-program = "1.16.3"
```

2501babe:
cc @ngundotra for review in case theres a reason to be on 1.10. the 1.14 -> 1.16 change in anchor and the rest of spl also entailed a borsh version change which breaks builds. so getting everything on one version will hopefully help downstream

ngundotra:
@2501babe I want to approve this, but with the version bump the `account-compression/sdk` tests no longer pass. They throw the following error: `Invoked an instruction with too many accounts (14693582971755349667 > 255)`

2501babe:
lol im having problems just building it myself because anchor 0.28 is pulling in borsh 0.9 instead of 0.10 because it has a too-permissive version constraint, probably as a hack to be able to keep people on 1.14...

ngundotra:
@ChewingGlass is this even building for you?

2501babe:
the Cargo.lock file is bad for me, it should build if you delete it and let it be regenerated

oh my god i wrote out a whole thing about the issues ive been having getting the tests to connect to the validator properly but its because `localhost` was routing over ipv6 lol... im getting the too many accounts error as expected now and should have time to try and debug soon-ish

ChewingGlass:
It was building for me, yes. But I ended up giving up on the path of forking and upgrading everyone. The Borsh upgrade just breaks too much stuff. Anchor-gen ended up being the solution. Feel free to close this PR and not do the upgrade.

ngundotra:
Closed by #4857

PR #4790, opened 2023-07-17
TLV Derive Macro for Borsh Variable Length

buffalojoec:
Broken off of #4768 (1/3)

---

This PR adds a derive macro to `spl-type-length-value` for simply adding the `VariableLenPack` trait with borsh methods for those who do not wish to implement a custom implementation of the tratit.

It simply mimics the `TokenMetadata` state implementation of `VariableLenPack`:

```rust
impl VariableLenPack for TokenMetadata {
    fn pack_into_slice(&self, dst: &mut [u8]) -> Result<(), ProgramError> {
        borsh::to_writer(&mut dst[..], self).map_err(Into::into)
    }
    fn unpack_from_slice(src: &[u8]) -> Result<Self, ProgramError> {
        try_from_slice_unchecked(src).map_err(Into::into)
    }
    fn get_packed_len(&self) -> Result<usize, ProgramError> {
        get_instance_packed_len(self).map_err(Into::into)
    }
}
```

---

I've also taken special care to set up a separate crate for testing the derive macro in order to keep `borsh` out of the dependencies for `spl-type-length-value`

### DIFF
--- a/libraries/type-length-value-derive-test/Cargo.toml
+++ b/libraries/type-length-value-derive-test/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "spl-type-length-value-derive-test"
+version = "0.0.1"
+description = "Testing Derive Macro Library for SPL Type Length Value traits"
+authors = ["Solana Labs Maintainers <maintainers@solanalabs.com>"]
+repository = "https://github.com/solana-labs/solana-program-library"
+license = "Apache-2.0"
+edition = "2021"
+
+[dev-dependencies]
+borsh = "0.10"
+solana-program = "1.16"
+spl-discriminator = { version = "0.1.0", path = "../discriminator" }
+spl-type-length-value = { version = "0.2.0", path = "../type-length-value", features = ["derive"] }
--- a/libraries/type-length-value-derive-test/Cargo.toml
+++ b/libraries/type-length-value-derive-test/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "spl-type-length-value-derive-test"
-version = "0.0.1"
+version = "0.1.0"
 description = "Testing Derive Macro Library for SPL Type Length Value traits"
 authors = ["Solana Labs Maintainers <maintainers@solanalabs.com>"]
 repository = "https://github.com/solana-labs/solana-program-library"
--- a/libraries/type-length-value/derive/Cargo.toml
+++ b/libraries/type-length-value/derive/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "spl-type-length-value-derive"
+version = "0.0.1"
+description = "Derive Macro Library for SPL Type Length Value traits"
+authors = ["Solana Labs Maintainers <maintainers@solanalabs.com>"]
+repository = "https://github.com/solana-labs/solana-program-library"
+license = "Apache-2.0"
+edition = "2021"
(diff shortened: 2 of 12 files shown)


libraries/type-length-value/derive/src/builder.rs
```diff
@@ -0,0 +1,91 @@
...
+
+impl From<&SplBorshVariableLenPackBuilder> for TokenStream {
+    fn from(builder: &SplBorshVariableLenPackBuilder) -> Self {
+        let ident = &builder.ident;
+        let generics = &builder.generics;
+        let where_clause = &builder.where_clause;
+        quote! {
+            impl #generics spl_type_length_value::variable_len_pack::VariableLenPack for #ident #generics #where_clause {
+                fn pack_into_slice(&self, dst: &mut [u8]) -> Result<(), solana_program::program_error::ProgramError> {
+                    borsh::to_writer(&mut dst[..], self).map_err(Into::into)
```
joncinque, line 78: It's pretty neat that you don't need to pull in borsh in the TLV lib thanks to macros!

libraries/type-length-value/derive/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+#![cfg_attr(not(test), forbid(unsafe_code))]
+
+extern crate proc_macro;
+
+mod builder;
+
+use builder::SplBorshVariableLenPackBuilder;
+use proc_macro::TokenStream;
+use quote::ToTokens;
+use syn::parse_macro_input;
```
joncinque, line 14: nit: can you combine these into one `use` for consistency with the tlv lib?

libraries/type-length-value-derive-test/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "spl-type-length-value-derive-test"
+version = "0.0.1"
```
joncinque, line 3: nit: May as well start with "0.1.0"

libraries/type-length-value/derive/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "spl-type-length-value-derive"
+version = "0.0.1"
```
joncinque, line 3: nit: same here, "0.1.0"

joncinque (dismissed):
Just a couple of little nits, then this is good to go -- great job!

libraries/type-length-value/derive/src/builder.rs
buffalojoec: Agreed but if you want to use this macro you have to pull borsh into your crate but that's less annoying lol

buffalojoec:
All comments resolved. I would like to note I kind of hate having a "test" crate - I think it looks weird. But we have some for program crates, so it's not unusual. Perhaps this crate itself can, over time, morph into a general testing crate for all derive macros in the `libraries` dir?

joncinque:
> I would like to note I kind of hate having a "test" crate

I totally appreciate that, but it's definitely needed in some cases. Typically it's whenever your dependencies would get screwed otherwise. And in this case, I want to avoid pulling borsh into the crate, so moving it into a test crate is a perfect solution.

I don't know if we need to worry about having some large test crate, but we can see how things shake out.

joncinque (approved):

Issue #4791, opened 2023-07-17
missing dependency @project-serum/anchor in the last release 0.1.9 of @solana/spl-account-compression


dmshvetsov:
1. This  commit https://github.com/solana-labs/solana-program-library/commit/3fb0c5d3c343d78d578c3e501f7411dedd11f8c7 introduced  `@project-serum/anchor` dist dependency in the 0.1.9 release.

2. Which is now causing this error:

```
2023-07-15T16:16:42.924234+00:00 app[web.1]: Error: Cannot find module '@project-serum/anchor/dist/cjs/utils/bytes'
2023-07-15T16:16:42.924258+00:00 app[web.1]: Require stack:
2023-07-15T16:16:42.924260+00:00 app[web.1]: - /app/node_modules/.pnpm/@solana+spl-account-compression@0.1.9_@solana+web3.js@1.76.0/node_modules/@solana/spl-account-compression/dist/cjs/events/index.js
2023-07-15T16:16:42.924260+00:00 app[web.1]: - /app/node_modules/.pnpm/@solana+spl-account-compression@0.1.9_@solana+web3.js@1.76.0/node_modules/@solana/spl-account-compression/dist/cjs/index.js
2023-07-15T16:16:42.924261+00:00 app[web.1]: - /app/node_modules/.pnpm/@metaplex-foundation+js@0.19.4/node_modules/@metaplex-foundation/js/dist/cjs/plugins/nftModule/operations/createCompressedNft.cjs
2023-07-15T16:16:42.924261+00:00 app[web.1]: - /app/node_modules/.pnpm/@metaplex-foundation+js@0.19.4/node_modules/@metaplex-foundation/js/dist/cjs/plugins/nftModule/operations/createNft.cjs
2023-07-15T16:16:42.924263+00:00 app[web.1]: - /app/node_modules/.pnpm/@metaplex-foundation+js@0.19.4/node_modules/@metaplex-foundation/js/dist/cjs/plugins/nftModule/NftBuildersClient.cjs
2023-07-15T16:16:42.924263+00:00 app[web.1]: - /app/node_modules/.pnpm/@metaplex-foundation+js@0.19.4/node_modules/@metaplex-foundation/js/dist/cjs/plugins/nftModule/NftClient.cjs
[...]

PR #4792, opened 2023-07-17
account-compression-js: Set @project-serum/anchor as a peer dependency

dmshvetsov:
resolves https://github.com/solana-labs/solana-program-library/issues/4791


dmshvetsov:
@ngundotra @nickfrosty I assume this PR is no longer need after #4864 change.

PR #4804, opened 2023-07-17
Unique & Allow Repeating APIs for TLV

buffalojoec:
Broken off of #4768 (2/3)

---

This PR adds support for duplicate TLV entries in the `spl-type-length-value` library.

I've simply taken almost all existing methods for `TlvStateMut` and some of the helpers and made them take a parameter `repetition_number: usize` or `allow_repetition: bool`. 

I've introduced an API of sorts for dealing with unique values, where each method has the prefix `first`. The functionality where duplicates are allowed has the suffix `*_with_repetition`.

With this change, developers can now use this library as we are using it in crates such as `token-metadata` (where we are dealing with unique non-repeating TLV entries) or their own use case where they might want multiple TLV entries for a certain type to be present in a buffer.


libraries/type-length-value/src/state.rs
```diff
@@ -300,37 +357,87 @@ impl<'data> TlvStateMut<'data> {
...
+    }
+
+    /// Packs a variable-length value into its appropriate data segment, where
+    /// no repeating discriminators are allowed
+    pub fn pack_variable_len_value_unique<V: SplDiscriminate + VariableLenPack>(
+        &mut self,
+        value: &V,
+    ) -> Result<(), ProgramError> {
+        self.pack_variable_len_value_at_entry::<V>(value, 0)
+    }
```
joncinque, line 425: Does this actually enforce uniqueness? It seems like you could call it on a TLV with multiple of the same type and it would work, correct? Maybe this should be `pack_first_variable_len_value`?

libraries/type-length-value/src/state.rs
```diff
@@ -300,37 +357,87 @@ impl<'data> TlvStateMut<'data> {
...
-        let data = self.get_bytes_mut::<V>()?;
+        let data = self.get_bytes_mut::<V>(entry_number)?;
         // NOTE: Do *not* use `pack`, since the length check will cause
         // reallocations to smaller sizes to fail
         value.pack_into_slice(data)
     }
 
+    /// Packs a variable-length value into its appropriate data segment, where
+    /// repeating discriminators _are_ allowed
+    pub fn pack_variable_len_value<V: SplDiscriminate + VariableLenPack>(
```
joncinque, line 410: naming nit: this one is less important since it has the new parameter, but better to be clear
```suggestion
    pub fn pack_variable_len_value_with_repetition<V: SplDiscriminate + VariableLenPack>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -346,25 +453,111 @@ impl<'data> TlvStateMut<'data> {
...
+        &mut self,
+        length: usize,
+    ) -> Result<&mut [u8], ProgramError> {
+        let (bytes, _) = self.alloc::<V>(length, Some(0))?;
+        Ok(bytes)
+    }
+
+    /// Allocate space for the given type, where repeating discriminators _are_
+    /// allowed
+    pub fn alloc_allow_repeating<V: SplDiscriminate>(
```
joncinque, line 474:
```suggestion
    pub fn alloc_with_repetition<V: SplDiscriminate>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -300,37 +357,87 @@ impl<'data> TlvStateMut<'data> {
...
+        let (extension_ref, _) = self.init_value_at_entry::<V>(Some(0))?;
         Ok(extension_ref)
     }
 
-    /// Packs a variable-length value into its appropriate data segment. Assumes that
-    /// space has already been allocated for the given type
-    pub fn pack_variable_len_value<V: SplDiscriminate + VariableLenPack>(
+    /// Packs a variable-length value into its appropriate data segment. Assumes
+    /// that space has already been allocated for the given type
+    fn pack_variable_len_value_at_entry<V: SplDiscriminate + VariableLenPack>(
```
joncinque, line 397: nit: to simplify things, you can remove this function  and just have `pack_variable_len_value`

libraries/type-length-value/src/state.rs
```diff
@@ -346,25 +453,111 @@ impl<'data> TlvStateMut<'data> {
...
+    /// Allocates and serializes a new TLV entry from a Pod type
+    fn add_entry<V: SplDiscriminate + Pod>(
+        &mut self,
+        value: &V,
+        entry_number: Option<usize>,
+    ) -> Result<usize, ProgramError> {
+        let (data, entry_number) = self.alloc::<V>(size_of::<V>(), entry_number)?;
+        data.copy_from_slice(bytemuck::bytes_of(value));
+        Ok(entry_number)
+    }
```
joncinque, line 490: This seems unnecessary since you already have `init_value_with_repetition`, probably just remove this and use that? It's kind of an antipattern to use zero-copy serde and then copy, so it's probably best to not encourage it

libraries/type-length-value/src/state.rs
```diff
@@ -346,25 +453,111 @@ impl<'data> TlvStateMut<'data> {
...
+
+    /// Allocates and serializes a new TLV entry from a Pod type, where no
+    /// repeating discriminators are allowed
+    pub fn add_entry_unique<V: SplDiscriminate + Pod>(
+        &mut self,
+        value: &V,
+    ) -> Result<(), ProgramError> {
+        self.add_entry(value, Some(0))?;
+        Ok(())
+    }
```
joncinque, line 500: Can also remove this and just use `init_value_unique`, right?

libraries/type-length-value/src/state.rs
```diff
@@ -346,25 +453,111 @@ impl<'data> TlvStateMut<'data> {
...
+
+    /// Allocates and serializes a new TLV entry from a Pod type, where
+    /// repeating discriminators _are_ allowed. Will add a new entry to the
+    /// next open slot provided there is enough space.
+    pub fn add_entry_allow_repeating<V: SplDiscriminate + Pod>(
+        &mut self,
+        value: &V,
+    ) -> Result<usize, ProgramError> {
+        self.add_entry(value, None)
+    }
```
joncinque, line 510: Can remove this too

libraries/type-length-value/src/state.rs
```diff
@@ -346,25 +453,111 @@ impl<'data> TlvStateMut<'data> {
...
+        value: &V,
+    ) -> Result<usize, ProgramError> {
+        self.add_variable_len_entry(value, None)
+    }
+
+    /// Reallocate the given number of bytes for the given SplDiscriminate. If
+    /// the new length is smaller, it will compact the rest of the buffer
+    /// and zero out the difference at the end. If it's larger, it will move
+    /// the rest of the buffer data and zero out the new data.
     pub fn realloc<V: SplDiscriminate>(
```
joncinque, line 548:
```suggestion
    pub fn realloc_with_repetition<V: SplDiscriminate>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -414,14 +617,18 @@ impl<'a> TlvState for TlvStateMut<'a> {
 pub fn realloc_and_pack_variable_len<V: SplDiscriminate + VariableLenPack>(
```
joncinque, line 617:
```suggestion
pub fn realloc_and_pack_variable_len_with_repetition<V: SplDiscriminate + VariableLenPack>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -346,25 +453,111 @@ impl<'data> TlvStateMut<'data> {
...
+        value: &V,
+    ) -> Result<(), ProgramError> {
+        self.add_variable_len_entry(value, Some(0))?;
+        Ok(())
+    }
+
+    /// Allocates and serializes a new TLV entry from a `VariableLenPack` type,
+    /// where repeating discriminators _are_ allowed. Will add a new entry
+    /// to the next open slot provided there is enough space.
+    pub fn add_variable_len_entry_allow_repeating<V: SplDiscriminate + VariableLenPack>(
```
joncinque, line 537: nit naming to make this clearer what's happening
```suggestion
    pub fn alloc_and_pack_variable_len_entry_with_repetition<V: SplDiscriminate + VariableLenPack>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -300,37 +357,87 @@ impl<'data> TlvStateMut<'data> {
...
-    pub fn alloc<V: SplDiscriminate>(&mut self, length: usize) -> Result<&mut [u8], ProgramError> {
-        let TlvIndices {
-            type_start,
-            length_start,
-            value_start,
-        } = get_indices(self.data, V::SPL_DISCRIMINATOR, true)?;
+    fn alloc<V: SplDiscriminate>(
+        &mut self,
+        length: usize,
+        entry_number: Option<usize>,
```
joncinque, line 431: This parameter is a bit confusing -- you only ever use `None` or `Some(0)`, so how about just passing a bool flag like `allow_repetition` and then set `entry_number` based on that?

libraries/type-length-value/src/state.rs
```diff
@@ -346,25 +453,111 @@ impl<'data> TlvStateMut<'data> {
...
+    ) -> Result<usize, ProgramError> {
+        let length = value.get_packed_len()?;
+        let (data, entry_number) = self.alloc::<V>(length, entry_number)?;
+        value.pack_into_slice(data)?;
+        Ok(entry_number)
+    }
+
+    /// Allocates and serializes a new TLV entry from a `VariableLenPack` type,
+    /// where no repeating discriminators _are_ allowed
+    pub fn add_variable_len_entry_unique<V: SplDiscriminate + VariableLenPack>(
```
joncinque, line 526:
```suggestion
    pub fn alloc_and_pack_variable_len_entry_unique<V: SplDiscriminate + VariableLenPack>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -346,25 +453,111 @@ impl<'data> TlvStateMut<'data> {
...
+    /// next open slot provided there is enough space.
+    pub fn add_entry_allow_repeating<V: SplDiscriminate + Pod>(
+        &mut self,
+        value: &V,
+    ) -> Result<usize, ProgramError> {
+        self.add_entry(value, None)
+    }
+
+    /// Allocates and serializes a new TLV entry from a `VariableLenPack` type
+    fn add_variable_len_entry<V: SplDiscriminate + VariableLenPack>(
```
joncinque, line 513:
```suggestion
    fn alloc_and_pack_variable_len_entry<V: SplDiscriminate + VariableLenPack>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -402,7 +595,17 @@ impl<'data> TlvStateMut<'data> {
 
         Ok(&mut self.data[value_start..new_value_end])
     }
+
+    /// Reallocate the given number of bytes for the given SplDiscriminate,
+    /// where no repeating discriminators are allowed
+    pub fn realloc_unique<V: SplDiscriminate>(
```
joncinque, line 601: I don't think this has any uniqueness guarantees either, right?
```suggestion
    pub fn realloc_first<V: SplDiscriminate>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -455,6 +662,15 @@ pub fn realloc_and_pack_variable_len<V: SplDiscriminate + VariableLenPack>(
     Ok(())
 }
 
+/// Packs a variable-length value into an existing TLV space, where no repeating
+/// discriminators are allowed
+pub fn realloc_and_pack_variable_len_unique<V: SplDiscriminate + VariableLenPack>(
```
joncinque, line 667: I don't think this has uniqueness guarantees either, right?
```suggestion
pub fn realloc_and_pack_first_variable_len<V: SplDiscriminate + VariableLenPack>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -467,10 +683,15 @@ fn check_data(tlv_data: &[u8]) -> Result<(), ProgramError> {
     Ok(())
 }
 
+// Tests for the base-use case of the TLV structure, where entries do NOT repeat
 #[cfg(test)]
 mod test {
-    use super::*;
-    use bytemuck::{Pod, Zeroable};
+    use solana_program::pubkey::Pubkey;
```
joncinque, line 689: nit: may as well fold the `use` statement in since you did it for the others

libraries/type-length-value/src/state.rs
```diff
@@ -467,10 +683,15 @@ fn check_data(tlv_data: &[u8]) -> Result<(), ProgramError> {
     Ok(())
 }
 
+// Tests for the base-use case of the TLV structure, where entries do NOT repeat
```
joncinque, line 686: nit: Can probably remove this comment since the tests seem interspersed

libraries/type-length-value/src/state.rs
```diff
@@ -191,23 +205,45 @@ pub trait TlvState {
     /// Get the full buffer containing all TLV data
     fn get_data(&self) -> &[u8];
 
-    /// Unpack a portion of the TLV data as the desired Pod type
-    fn get_value<V: SplDiscriminate + Pod>(&self) -> Result<&V, ProgramError> {
-        let data = get_bytes::<V>(self.get_data())?;
+    /// Unpack a portion of the TLV data as the desired Pod type for the entry
+    /// number specified
+    fn get_value<V: SplDiscriminate + Pod>(&self, entry_number: usize) -> Result<&V, ProgramError> {
```
joncinque, line 210: nit naming
```suggestion
    fn get_value_with_repetition<V: SplDiscriminate + Pod>(&self, entry_number: usize) -> Result<&V, ProgramError> {
```

libraries/type-length-value/src/state.rs
```diff
@@ -191,23 +205,45 @@ pub trait TlvState {
...
+    /// Unpacks a portion of the TLV data as the desired variable-length type
+    /// for the first entry found
+    fn get_first_variable_len_value<V: SplDiscriminate + VariableLenPack>(
+        &self,
+    ) -> Result<V, ProgramError> {
+        self.get_variable_len_value::<V>(0)
+    }
+
+    /// Unpack a portion of the TLV data as bytes for the entry number specified
+    fn get_bytes<V: SplDiscriminate>(&self, entry_number: usize) -> Result<&[u8], ProgramError> {
```
joncinque, line 240:
```suggestion
    fn get_bytes_with_repetition<V: SplDiscriminate>(&self, entry_number: usize) -> Result<&[u8], ProgramError> {
```

libraries/type-length-value/src/state.rs
```diff
@@ -191,23 +205,45 @@ pub trait TlvState {
...
 
+    /// Unpack a portion of the TLV data as the desired Pod type for the first
+    /// entry found
+    fn get_first_value<V: SplDiscriminate + Pod>(&self) -> Result<&V, ProgramError> {
+        self.get_value::<V>(0)
+    }
+
     /// Unpacks a portion of the TLV data as the desired variable-length type
+    /// for the entry number specified
     fn get_variable_len_value<V: SplDiscriminate + VariableLenPack>(
```
joncinque, line 223:
```suggestion
    fn get_variable_len_value_with_repetition<V: SplDiscriminate + VariableLenPack>(
```

libraries/type-length-value/src/state.rs
```diff
@@ -861,19 +1103,71 @@ mod test {
...
-        let value = state.get_value::<TestNonZeroDefault>().unwrap();
+        let value = state.get_first_value::<TestNonZeroDefault>().unwrap();
         assert_eq!(*value, TestNonZeroDefault::default());
 
         // realloc to smaller, still all 0
-        let data = state.realloc::<TestValue>(SMALL_SIZE).unwrap();
+        let data = state.realloc::<TestValue>(SMALL_SIZE, 0).unwrap();
         assert_eq!(data, [0; SMALL_SIZE]);
-        let value = state.get_value::<TestNonZeroDefault>().unwrap();
+        let value = state.get_first_value::<TestNonZeroDefault>().unwrap();
```
joncinque, line 1170: You're checking `TestNonZeroDefault`, but to be sure, can you also check that the other `TestValue` entry is all zeroes after each realloc?

libraries/type-length-value/src/state.rs
```diff
@@ -936,23 +1231,132 @@ mod test {
...
+            .unwrap();
+        assert_eq!(deser, test_variable_len);
+
+        // Alloc and get entry number (second entry)
+        let (_, entry_number) = state
+            .alloc_allow_repeating::<TestVariableLen>(tlv_size)
+            .unwrap();
+        let test_variable_len = TestVariableLen {
+            data: initial_data.to_string(),
+        };
```
joncinque, line 1291: To make these checks more useful, could you have different data for the second value? That way we're sure you're fetching and checking the second one

joncinque (commented):
Really great work! Don't be fooled by the number of comments, it's mostly nits on the API, naming, and which functions to expose.

libraries/type-length-value/src/state.rs
```diff
@@ -300,37 +357,87 @@ impl<'data> TlvStateMut<'data> {
...
-    /// Packs the default TLV data into the first open slot in the data buffer.
-    /// If extension is already found in the buffer, it returns an error.
-    pub fn init_value<V: SplDiscriminate + Pod + Default>(
+    /// Unpack a portion of the TLV data as mutable bytes for the first entry
+    /// found
+    pub fn get_first_bytes_mut<V: SplDiscriminate>(&mut self) -> Result<&mut [u8], ProgramError> {
+        self.get_bytes_mut::<V>(0)
+    }
+
+    fn init_value_at_entry<V: SplDiscriminate + Pod + Default>(
```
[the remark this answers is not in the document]
buffalojoec: It's just re-used code! You're correct you just use those two, and this one is shared across both. I just renamed to `inner_init_value` to see if that makes more sense

libraries/type-length-value/src/state.rs
buffalojoec: Yep that's fair, it's always going to do first, so `first` makes more sense

libraries/type-length-value/src/state.rs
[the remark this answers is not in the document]
buffalojoec: Actually I ended up changing this altogether

libraries/type-length-value/src/state.rs
buffalojoec: That's fair. I was trying to get around having to set each field or set the data in an extra step. But better off removing prob

buffalojoec:
> Really great work! Don't be fooled by the number of comments, it's mostly nits on the API, naming, and which functions to expose.

Thanks, I figured we'd def have some discussion about naming conventions. I resolved all your comments and sliced out some unnecessary/redundant functions. Lmk! Thanks

libraries/type-length-value/src/state.rs
```diff
@@ -14,14 +14,10 @@ use {
 };
 
 /// Get the current TlvIndices from the current spot
-const fn get_indices_unchecked(type_start: usize) -> TlvIndices {
+const fn get_indices_unchecked(type_start: usize) -> (usize, usize, usize) {
```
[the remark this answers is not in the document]
buffalojoec: I had this written and then removed it, because to me it didn't really make sense. `get_indices_unchecked` is just adding to `type_start` to get `length_start` and `value_start`, and doesn't know or care what the discriminator is, so it can't possibly determine which repetition number it's on.

It would have to be just a literal pass-through value, which I can do, but it felt weird.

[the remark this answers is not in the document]
joncinque: it's an internal function, so I'm fine with just doing a pass-through for that repetition considering it'll avoid a big tuple in favor of a struct, and also help readability in the other functions. do you have a particular concern with it?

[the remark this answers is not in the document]
buffalojoec: Nope, just seemed weird that's all! 
I just added it, and it cleans up (and makes sense with) `get_indices` since it returns the same indices, but it looks a little weird with `get_discriminators_and_end_index`. However, since that function doesn't return any indices or repetition number, I think it's all good.

libraries/type-length-value/src/state.rs
```diff
@@ -85,28 +85,34 @@ fn get_discriminators_and_end_index(
     let mut discriminators = vec![];
     let mut start_index = 0;
     while start_index < tlv_data.len() {
-        let (type_start, length_start, value_start) = get_indices_unchecked(start_index);
-        if tlv_data.len() < length_start {
+        let tlv_indices = get_indices_unchecked(start_index, 0);
```
joncinque, line 88: I see what you mean here... up to you, but you could add a comment here to explain the weirdness on `0` and why it's there

joncinque (dismissed):
Just an up-to-you nit, but otherwise this is good to go!

libraries/type-length-value/src/state.rs
buffalojoec: Yep, added a comment

joncinque (approved):

Issue #4701, opened 2023-07-05
token-cli: Support updating metadata pointer extension


joncinque:
#### Problem

#4636 added support to initialize the metadata pointer extension in a new mint, but it still isn't possible to update the metadata pointer address.

#### Solution

Add a new top-level command, called `UpdateMetadataPointerAddress`, which calls `update_metadata_pointer_address`: https://github.com/solana-labs/solana-program-library/blob/fccc0919513d7939ff624d3a80db2b1b883ca8b3/token/client/src/token.rs#L1576

Currently, the most similar instruction is `update_default_account_state` https://github.com/solana-labs/solana-program-library/blob/fccc0919513d7939ff624d3a80db2b1b883ca8b3/token/cli/src/main.rs#L2305

Rhythm-08:
Hello @joncinque , I am interested in this issue
Can you assign this issue to me.

joncinque:
@Rhythm-08 Go for it! I typically only assign issues to team members, but if you send a PR for this, I'll gladly take a look!

dhruvdabhi101:
If this issue is open to fix then I can give it a try. React to this comment if it's available.

buffalojoec:
> If this issue is open to fix then I can give it a try. React to this comment if it's available.

Might just want to check with @Rhythm-08 , possibly collab ?!

Rhythm-08:
> > If this issue is open to fix then I can give it a try. React to this comment if it's available.
> 
> Might just want to check with @Rhythm-08 , possibly collab ?!

Sure

PR #4810, opened 2023-07-18
token-cli: add command for updating metadata address

sanjsingh07:
resolved https://github.com/solana-labs/solana-program-library/issues/4701


token/cli/src/main.rs
```diff
@@ -3624,6 +3652,44 @@ fn app<'a, 'b>(
...
+                        .help(
+                            "Specify the token's authority. \
+                            This may be a keypair file or the ASK keyword. \
+                            Defaults to the client keypair.",
+                        ),
+                )
+                .arg(owner_address_arg())
+                .arg(multisig_signer_arg())
+                .nonce_args(true)
+                .offline_args(),
```
joncinque, line 3691: Can remove the offline args, since the instruction doesn't support sign-only

token/cli/src/main.rs
```diff
@@ -3624,6 +3652,44 @@ fn app<'a, 'b>(
...
+                        .value_name("KEYPAIR")
+                        .validator(is_valid_signer)
+                        .takes_value(true)
+                        .help(
+                            "Specify the token's authority. \
+                            This may be a keypair file or the ASK keyword. \
+                            Defaults to the client keypair.",
+                        ),
+                )
+                .arg(owner_address_arg())
```
joncinque, line 3688: You're using the `authority` arg and not the `owner` arg, so this can be removed

token/cli/src/main.rs
```diff
@@ -3624,6 +3652,44 @@ fn app<'a, 'b>(
...
+                        ),
+                )
+                .arg(
+                    Arg::with_name("authority")
+                        .long("authority")
+                        .value_name("KEYPAIR")
+                        .validator(is_valid_signer)
+                        .takes_value(true)
+                        .help(
+                            "Specify the token's authority. \
```
joncinque, line 3683: nit: Let's be clear
```suggestion
                            "Specify the token's metadata-pointer authority. \
```

token/cli/src/main.rs
```diff
@@ -3624,6 +3652,44 @@ fn app<'a, 'b>(
...
+                )
+                .arg(
+                    Arg::with_name("metadata_address")
+                        .long("metadata-address")
+                        .value_name("ADDRESS")
+                        .takes_value(true)
+                        .help(
+                            "Specify address that stores token metadata."
+                        ),
+                )
```
joncinque, line 3675: This will introduce a footgun. As it stands, if someone just accidentally forgets to put in a metadata address, the command will clear it.

Instead, this should look like the `new_authority` flag in `Authorize`, where someone can specify `--disable` to clear the metadata pointer address. Then if `--disable` is *not* provided, the `metadata_address` argument is required.

token/cli/src/main.rs
```diff
@@ -144,6 +144,7 @@ pub enum CommandName {
     EnableCpiGuard,
     DisableCpiGuard,
     UpdateDefaultAccountState,
+    UpdateMetadataPointerAddress,
```
joncinque, line 147: nit for ease of use
```suggestion
    UpdateMetadataAddress,
```

joncinque (commented):
Thanks for your contribution! A few bits to firm up the interface, and also, could you add a test for this? You can use the existing `metadata_pointer()` test for creating, and then just add a step where you update the metadata address, and check that the change was properly propagated.

token/cli/src/main.rs
sanjsingh07: done!

token/cli/src/main.rs
sanjsingh07: done!

token/cli/src/main.rs
sanjsingh07: done!

token/cli/src/main.rs
```diff
@@ -3624,6 +3652,44 @@ fn app<'a, 'b>(
                 .nonce_args(true)
                 .offline_args(),
         )
+        .subcommand(
+            SubCommand::with_name(CommandName::UpdateMetadataPointerAddress.into())
+                .about("Updates default account state for the mint. Requires the default account state extension.")
```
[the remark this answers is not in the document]
sanjsingh07: done!

token/cli/src/main.rs
```diff
@@ -3624,6 +3652,44 @@ fn app<'a, 'b>(
...
+                .arg(
+                    Arg::with_name("token")
+                        .validator(is_valid_pubkey)
+                        .value_name("TOKEN_MINT_ADDRESS")
+                        .takes_value(true)
+                        .index(1)
+                        .required(true)
+                        .help("The address of the token mint to update default account state"),
```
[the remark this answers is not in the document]
sanjsingh07: done!

token/cli/src/main.rs
sanjsingh07: done!

sanjsingh07:
updated `metadata_pointer()`  to have update-metadata-pointer address test case.
image

token/cli/src/main.rs
```diff
@@ -3624,6 +3652,41 @@ fn app<'a, 'b>(
...
+                        .index(1)
+                        .required(true)
+                        .help("The address of the token mint to update the metadata pointer address"),
+                )
+                .arg(
+                    Arg::with_name("metadata_address")
+                        .long("metadata-address")
+                        .value_name("METADATA_ADDRESS")
+                        .takes_value(true)
+                        .required_unless("disable")
```
joncinque, line 3672: You'll have to add the `"disable"` flag in order to make this work properly, ie:

```
                .arg(
                    Arg::with_name("disable")
                        .long("disable")
                        .takes_value(false)
                        .conflicts_with("metadata_address")
                        .help("Unset metadata pointer address.")
                )
```

token/cli/src/main.rs
```diff
@@ -3624,6 +3652,41 @@ fn app<'a, 'b>(
...
+                        .validator(is_valid_pubkey)
+                        .value_name("TOKEN_MINT_ADDRESS")
+                        .takes_value(true)
+                        .index(1)
+                        .required(true)
+                        .help("The address of the token mint to update the metadata pointer address"),
+                )
+                .arg(
+                    Arg::with_name("metadata_address")
+                        .long("metadata-address")
```
joncinque, line 3669: Since this is normally required (unless disable is set), rather than setting it with a flag, let's make a positional argument, ie:
```suggestion
                        .index(2)
```

joncinque (commented):
We're almost there! When you add the disable flag, can you add one more step to the test to disable the metadata pointer?

token/cli/src/main.rs
sanjsingh07: done! also update `metadata_pointer()` test-case so it contain `--disable` scenario as well.

token/cli/src/main.rs
sanjsingh07: done.

token/cli/src/main.rs
```diff
@@ -3624,6 +3652,48 @@ fn app<'a, 'b>(
...
+                        .validator(is_valid_pubkey)
+                        .value_name("TOKEN_MINT_ADDRESS")
+                        .takes_value(true)
+                        .index(1)
+                        .required(true)
+                        .help("The address of the token mint to update the metadata pointer address"),
+                )
+                .arg(
+                    Arg::with_name("metadata_address")
+                        .index(2)
```
joncinque, line 3669: Can you add `.validator(is_valid_pubkey)` for this arg too? Otherwise, if someone inputs an invalid pubkey, it'll disable the metadata pointer address!

joncinque (commented):
One last bit then this should be good to go!

token/cli/src/main.rs
sanjsingh07: Yupp, Done! and also sorry about this many commits. :)

joncinque (approved):
Looks great, thanks for your patience! I'll ping you if CI has any issues.

And a lot of commits is really good, it makes it much easier for me to review. Do avoid merge commits in the future though, and prefer rebasing + force-pushing your branch whenever you need to.

joncinque:
All good, thanks for your work!

PR #4816, opened 2023-07-18
[token-2022] Add support for context state accounts

samkim-crypto:
#### Problem
The token-2022 confidential transfer extension instructions that require zero-knowledge proofs do not yet support context states.

#### Summary of changes
Added support for context state accounts. All the confidential transfer instructions that require zkp's have an `i8` proof instruction offset field that represents the position of the zkp instruction relative to the token-2022 instruction. Since the offset cannot be zero, I updated the program so that it if the offset is zero, it checks for a context state account.

- [8bb4d2a](https://github.com/solana-labs/solana-program-library/pull/4816/commits/8bb4d2aec8841a7689e1ca2980491fad6a5c1844): The previous PR that updated the`EmptyAccount` instruction for 1.16 did not add account info for it, so I added it in line with the rest of the zkp requiring instructions and to simplify the `generate_proof_data` interface.
- [1a99b80](https://github.com/solana-labs/solana-program-library/pull/4816/commits/1a99b8075b20bc619d612fa853cf32a3ec8cb639): Added support for context states for the `ConfigureAccount` instruction. The main changes are in `instruction.rs` and `processor.rs`. The rest of the commits model this commit.
- [f7772ea](https://github.com/solana-labs/solana-program-library/pull/4816/commits/f7772ea1e015066ae8802cba2fd05a3bcdd58c12): Added support for context states for the ``EmptyAccount` instruction.
- [6a9a77d](https://github.com/solana-labs/solana-program-library/pull/4816/commits/6a9a77dc4b40ebb146072c9312ab42f19f0c6f8b): Added support for context states for the `Withdraw` instruction.
- [426589a](https://github.com/solana-labs/solana-program-library/pull/4816/commits/426589a7521ea145ef456563bbb93d7e8f477ec0): Added support for context states for the `Transfer` instruction. The transfer instruction also requires processing of smaller zkp proof context states. I started making the changes in this PR until I realized this PR is quite long, so I decided to do it in a subsequent PR.

### DIFF
--- a/token/program-2022-test/tests/confidential_transfer.rs
+++ b/token/program-2022-test/tests/confidential_transfer.rs
@@ -764,6 +764,7 @@ async fn confidential_transfer_empty_account() {
         .confidential_transfer_empty_account(
             &alice_meta.token_account,
             &alice.pubkey(),
+            None,
             &alice_meta.elgamal_keypair,
             &[&alice],
         )
@@ -1123,6 +1124,7 @@ async fn confidential_transfer_withdraw() {
         .confidential_transfer_empty_account(
             &alice_meta.token_account,
             &alice.pubkey(),
+            None,
             &alice_meta.elgamal_keypair,
             &[&alice],
         )
--- a/token/program-2022-test/tests/confidential_transfer.rs
+++ b/token/program-2022-test/tests/confidential_transfer.rs
@@ -777,6 +777,7 @@ async fn confidential_transfer_empty_account() {
             &alice_meta.token_account,
             &alice.pubkey(),
             None,
+            None,
             &alice_meta.elgamal_keypair,
             &[&alice],
         )
@@ -1136,6 +1137,7 @@ async fn confidential_transfer_withdraw() {
         .confidential_transfer_empty_account(
             &alice_meta.token_account,
             &alice.pubkey(),
+            None,
             None,
             &alice_meta.elgamal_keypair,
             &[&alice],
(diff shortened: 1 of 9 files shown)


token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -2106,3 +2139,688 @@ async fn ct_transfer_with_fee_memo() {
...
+    assert_eq!(
+        err,
+        TokenClientError::Client(Box::new(TransportError::TransactionError(
+            TransactionError::InstructionError(0, InstructionError::InvalidArgument,)
+        )))
+    );
+}
+
+#[tokio::test]
+async fn confidential_transfer_transfer_with_proof_context() {
```
joncinque, line 2635: Can you add a test case for transfer with fee?

token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -489,18 +503,28 @@ pub fn inner_configure_account(
     mint: &Pubkey,
     decryptable_zero_balance: AeCiphertext,
     maximum_pending_balance_credit_counter: u64,
+    context_state_account: Option<&Pubkey>,
     authority: &Pubkey,
     multisig_signers: &[&Pubkey],
     proof_instruction_offset: i8,
```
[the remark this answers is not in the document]
samkim-crypto: Great! I created a type along the lines of
```
enum ProofDataLocation {
    Instruction(NonZeroI8, ProofData),
    ContextAccount(Pubkey)
}
```
This seems to simplify things a lot!

token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -527,23 +551,39 @@ pub fn configure_account(
     mint: &Pubkey,
     decryptable_zero_balance: AeCiphertext,
     maximum_pending_balance_credit_counter: u64,
+    context_state_account: Option<&Pubkey>,
     authority: &Pubkey,
     multisig_signers: &[&Pubkey],
-    proof_data: &PubkeyValidityData,
+    proof_data: Option<&PubkeyValidityData>,
```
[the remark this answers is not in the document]
samkim-crypto: I think this should be addressed by the same type `ProofDataLocation`. Let me know what you think!

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -127,12 +134,25 @@ fn process_configure_account(
...
+        let context_state_account_data = context_state_account_info.data.borrow();
+        let context_state = pod_from_bytes::<ProofContextState<PubkeyValidityProofContext>>(
+            &context_state_account_data,
+        )?;
+
+        if context_state.proof_type != ProofType::PubkeyValidity.into() {
+            return Err(ProgramError::InvalidInstructionData);
+        }
+
+        context_state.proof_context
```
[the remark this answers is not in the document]
samkim-crypto: Sorry I missed this! Added.

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -525,12 +605,24 @@ fn process_transfer(
...
+            )?;
+            if context_state.proof_type != ProofType::Transfer.into() {
+                return Err(ProgramError::InvalidInstructionData);
+            }
+            context_state.proof_context
+        } else {
+            let zkp_instruction =
+                get_instruction_relative(proof_instruction_offset, instructions_sysvar_info)?;
+            *decode_proof_instruction_context::<TransferWithFeeData, TransferWithFeeProofContext>(
+                ProofInstruction::VerifyTransfer,
```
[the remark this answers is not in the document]
samkim-crypto: Yes thanks for the catch. This should now be addressed!

token/program-2022-test/tests/confidential_transfer.rs
samkim-crypto: Since this involves the confidential transfer fee extension, I will add the tests on a subsequent PR when I update all the fee component to Solana 1.16.

token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -551,39 +555,34 @@ pub fn configure_account(
...
+    )?];
+
+    if let ProofLocation::Instruction(proof_instruction_offset, proof_data) = proof_data_location {
+        // This constructor appends the proof instruction right after the `ConfigureAccount`
+        // instruction. This means that the proof instruction offset must be always be 1. To
+        // use an arbitrary proof instruction offset, use the `inner_configure_account`
+        // constructor.
+        let proof_instruction_offset: i8 = proof_instruction_offset.into();
+        if proof_instruction_offset != 1 {
+            return Err(ProgramError::InvalidArgument);
```
joncinque, line 588: Can you add another error for this case since it's very specific? I agree with the behavior, but for someone calling this from the outside, it'll be much clearer to get a more special error that they can eventually look up to find out that they need to use `1`

token/program-2022/src/proof.rs
```diff
@@ -25,3 +25,31 @@ pub fn decode_proof_instruction_context<T: Pod + ZkProofData<U>, U: Pod>(
...
+        } else {
+            Ok(Self(n))
+        }
+    }
+}
+impl From<NonZeroI8> for i8 {
+    fn from(n: NonZeroI8) -> Self {
+        n.0
+    }
+}
```
joncinque, line 46: Feel free to use the `std` version! https://doc.rust-lang.org/std/num/struct.NonZeroI8.html

token/program-2022/src/proof.rs
```diff
@@ -25,3 +25,31 @@ pub fn decode_proof_instruction_context<T: Pod + ZkProofData<U>, U: Pod>(
...
+
+/// A proof location type meant to be used for arguments to instruction constructors.
+#[derive(Clone, Copy)]
+pub enum ProofLocation<'a, T> {
+    /// The proof is included in the same transaction of a corresponding token-2022 instruction.
+    Instruction(NonZeroI8, &'a T),
```
joncinque, line 52: nit so that it's totally clear where to find it
```suggestion
    InstructionOffset(NonZeroI8, &'a T),
```

Or maybe even better, *don't* have this take an i8 value, and instead expose `get_instruction_offset()` function which returns 1 for `InstructionOffset` and 0 for `ContextStateAccount`, and that way you can remove all of the repeated validation logic.

joncinque (commented):
Everything is pretty much there! I'll approve so that you can move forward and address the last points as you like.

And sorry, just to double-check... will you need *multiple* context state accounts later for transfer? If so, `ProofLocation` will need a new version for `Transfer` which takes in multiple pubkeys, is that it?

joncinque (dismissed):
Oops, forgot to hit "approve"

samkim-crypto:
> And sorry, just to double-check... will you need _multiple_ context state accounts later for transfer? If so, `ProofLocation` will need a new version for `Transfer` which takes in multiple pubkeys, is that it?

Yeah, looking at it now, it seems like it might make more sense to change `ProofLocation::ContextStateAccount(&'a, Pubkey)` to `ProofLocation::ContextStateAccount(&'a &[&Pubkey])` instead.

token/program-2022/src/proof.rs
samkim-crypto: I changed this to `InstructionOffset`. I did think about removing the instruction offset, but there could be an application where the proof instruction is 2 or 3 places out from the token instruction. In this case, they would call the `inner_configure_account()` constructor with a custom instruction offset, so it might make sense to leave the offset field in there.

token/program-2022/src/proof.rs
samkim-crypto: Ah I did not know it existed. Thanks!

PR #4818, opened 2023-07-18
Interfaces Directory

buffalojoec:
OK this one is going to be annoying, but it may be necessary.

With the introduction of two interfaces thus far - `transfer-hook` and `token-metadata` - and more on the roadmap ahead (I'm currently developing `collections` and `editions`), it might be a good idea to start keeping our interfaces in their own standalone collection within SPL.

Rather than have interfaces strewn all over the repo as we add more to them, why not condense them all into an `interfaces` directory, where each has an `interface` for the interface itself and an `example` (as the two existing interfaces currently do)?

I'm going to leave this one as a draft for now, but let me know what you think!


joncinque:
We can probably condense `transfer-hook-interface` and `transfer-hook-example` into `transfer-hook/interface` and `transfer-hook/example`, but other than that, we might just want to create a whole new repo for interfaces.

buffalojoec:
> We can probably condense `transfer-hook-interface` and `transfer-hook-example` into `transfer-hook/interface` and `transfer-hook/example`, but other than that, we might just want to create a whole new repo for interfaces.

Sounds good, I think I'll hold off until I PR the two new ones and we can discuss a new repo

buffalojoec:
Closing in favor of future work re: #5035

PR #4821, opened 2023-07-19
Editions Interface

buffalojoec:
This PR introduces an interface for Token Editions. With this interface, on-chain programs can create tokens that are "originals" and then "print copies" of these tokens using specific additional metadata - dubbed `Original` and `Reprint`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7039,6 +7039,32 @@ dependencies = [
 ]
 
 [[package]]
+name = "spl-token-editions-example"
+version = "0.1.0"
+dependencies = [
+ "solana-program",
+ "solana-program-test",
+ "solana-sdk",
+ "spl-token-2022 0.7.0",
+ "spl-token-client",
+ "spl-token-editions-interface",
+ "spl-token-metadata-interface",
+ "spl-type-length-value",
+ "test-case",
+]
+
+[[package]]
+name = "spl-token-editions-interface"
+version = "0.1.0"
+dependencies = [
+ "borsh 0.10.3",
+ "solana-program",
+ "spl-discriminator",
+ "spl-program-error",
+ "spl-type-length-value",
+]
+
+[[package]]
 name = "spl-token-lending"
 version = "0.2.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,8 @@ members = [
   "associated-token-account/program-test",
   "binary-option/program",
   "binary-oracle-pair/program",
+  "editions/example",
+  "editions/interface",
   "examples/rust/cross-program-invocation",
   "examples/rust/custom-heap",
   "examples/rust/logging",
--- a/editions/README.md
+++ b/editions/README.md
@@ -1,0 +1,97 @@
+# Editions
+
+This interface is designed to allow on-chain programs to make use of printed "copies" of token metadata, with parent-child relationships.
+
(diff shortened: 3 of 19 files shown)


joncinque (commented):
I haven't looked through the code in depth, but I don't think I understand the motivation. Why would someone use this interface exactly? It seems easier to have tooling in `spl-token-metadata` which just copies an existing metadata account.

buffalojoec:
> I haven't looked through the code in depth, but I don't think I understand the motivation. Why would someone use this interface exactly? It seems easier to have tooling in `spl-token-metadata` which just copies an existing metadata account.

I think the direct answer to your question revolves around the pointers (parent-child relationship), however it's a valid point.

My thoughts on `spl-token-editions` as a standalone interface were basically:
- Super simple, highly-composable interfaces that can be stacked on each other for combined functionality
- An interface specifically for parent-child relationships amongst metadata/assets (could be combined to serve Collections, too)
- Specific TLV entries corresponding solely to editions
  - One can separate the authorities on `Metadata` and `Edition`
  - One can choose to store this information in separate accounts, the metadata account, or the mint account (if they've implemented `spl-token`)
- Can leverage other programs that implement `spl-token-metadata`
  - This means you could deploy just an editions interface and other metadata-implemented programs can leverage it. A token could have different editions across many implementations!

However, we can definitely include this in the `spl-token-metadata` interface, but I'm curious on a couple of things:
- Are you thinking we just include these instructions, or have extensions on the existing instructions somehow, and devs can optionally include the extended tooling for editions?
- Could we possibly take advantage of the additional key-value pairs included in `spl-token-metadata` and just provide types for editions?
  - Could this cause collision concerns? (ie. someone uses the reserved state key but doesn't actually implement the interface)

This does give me a couple of neat ideas like "interfaces within interfaces", where `spl-token-metadata` has one barebones set of instructions/accounts/parameters, but with enabled "sub-interfaces" can be expanded for more functionality, however this is a bit different than what we originally discussed.

A good example of what I mean might be: "I want to be able to create a `Metadata` _and_ `Original` state (possibly combined into one state entry) in the same instruction".

I was sticking with the original idea of having separate interfaces for each type of use case, and programs only implement those they want. But if you think this makes more sense to be wrapped into Token Metadata (possibly Collections as well given they're quite similar), I'm down

buffalojoec:
We discussed this a bit offline, but wondering if we wanted to land on the idea of a more generalized parent-child related interface for different state types.

Did you still want to check this code out a bit? Should I just throw said generalized interface together? 
I can link out to it from here

joncinque:
I'd prefer the more general parent -> child relationship that we discussed, since that should cover both collections and editions.

Let's see... to do copies, maybe with another marker for what kind of child it is? And for collections, just call them children? ie.

```
enum MemberType {
    Child,
    Copy,
}
```

But since I still know very little about the NFT world, maybe we can run this by someone else to make sure that it covers what's needed / expected?

buffalojoec:
> But since I still know very little about the NFT world, maybe we can run this by someone else to make sure that it covers what's needed / expected?

If anyone versed in NFTs sees this and wants to elbow-drop in, by all means!

However, it seems as though an sRFC is in order, to follow [00017](https://forum.solana.com/t/srfc-00017-token-metadata-interface/283). 

I’m happy to produce a spec and publish an sRFC!

buffalojoec:
Closing in favor of #5036 !

PR #4827, opened 2023-07-20
Collections Interface

buffalojoec:
This PR introduces an interface for Token Collections. With this interface, on-chain programs can create groups of tokens - dubbed "Collections" - that identify members of such a group through parent-child relationships.

Note: This interface is almost identical to the SPL Token Editions interface #4821 , but with some slight differences that are described in the `README.md`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7039,6 +7039,32 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


buffalojoec:
Closing in favor of #5036 !

PR #4828, opened 2023-07-21
single-pool: js

2501babe:
using the new experimental web3js


2501babe:
closing this and letting the branch sit dormant. my current plan now is to flesh out the js lib on web3legacy and aim to make its web3next-first at a later date, when both libraries can be used together. its not possible for existing wallets to work with web3next outputs, this will ahve to wait until the web3legacy interface is reimplemented on top of web3next

PR #4833, opened 2023-07-21
js: Update tests to use 127.0.0.1 instead of localhost

joncinque:
#### Problem

As noticed at https://github.com/solana-labs/solana-program-library/pull/4787#issuecomment-1640714523, if you use node 18, it defaults `localhost` over ipv6, and the solana test validator only binds to the ipv4 address.

#### Solution

I took a look at updating solana-test-validator to bind to the ipv6 address, and it works great for JS, but breaks all of the Rust tests, so that'll be a longer endeavor.

In the meantime, to at least properly support node 18, let's just explicitly use 127.0.0.1 instead of localhost in the tests.


2501babe (approved):

2501babe:
wow this feels like concierge service, id written it off as an issue with my environment :sweat_smile:

joncinque:
haha nah, i've been meaning to make this change for a bit, but your issue made it more pressing

PR #4504, opened 2023-06-07
Update Stake with the additional deactivation flag

HaoranYi:
Update StakeState unpack with additional deactivation flag.

PR #4841, opened 2023-07-24
Revert "Update Stake with the additional deactivation flag (#4504)"

joncinque:
This reverts commit 19645b6cea1752bd1725e2170f3578950b0eda3a.

#### Problem

#4504 was merged before the upstream changes landed. We might want to keep it in draft to make it clear that it's not ready. Master currently doesn't build.

#### Solution

Revert the commit.

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -683,7 +683,7 @@ fn command_deposit_stake(
         println!("Depositing stake account {:?}", stake_state);
     }
     let vote_account = match stake_state {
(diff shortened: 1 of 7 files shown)


joncinque:
The stake pool tests and clippy all passed, which is what broke, so merging this early to unblock!

yihau:
hey. I found we have already committed the StakeState related code to mono repo.

https://github.com/solana-labs/solana/blob/de4ff348b404be5bba2ee62463d785fb070a0a57/sdk/program/src/stake/state.rs#L26

so I think this one should restore. if someone encounter a similar error again, they need to rebase their mono repo.

PR #4842, opened 2023-07-24
Revert "Revert "Update Stake with the additional deactivation flag (#4504)""

yihau:
Reverts solana-labs/solana-program-library#4841

actually I found we have included required changes in monorepo here
https://github.com/solana-labs/solana/blob/de4ff348b404be5bba2ee62463d785fb070a0a57/sdk/program/src/stake/state.rs#L26

so I think maybe we can try to restore this one. if someone encounter a similar error, they can try to rebase their monorepo first.

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -683,7 +683,7 @@ fn command_deposit_stake(
         println!("Depositing stake account {:?}", stake_state);
     }
     let vote_account = match stake_state {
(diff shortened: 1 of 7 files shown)


joncinque:
We can't merge this until the new crates are released. So rather than making this change now, we should disable the SPL downstream tests from the monorepo, since it's a breaking change in the SDK. And we have to properly version the SDK to version 2.0.0. Once the new version of the sdk crates is released, we can update the SPL to use it

yihau:
yeah. this flow sounds more correct. thank you for pointing this out. will open another PR to disable spl downstream project on monorepo!

PR #4857, opened 2023-07-25
Bump account compression to use anchor 0.28.0

ngundotra:
Closing the loop on community PR to upgrade account compression #4787

### DIFF
--- a/libraries/concurrent-merkle-tree/Cargo.toml
+++ b/libraries/concurrent-merkle-tree/Cargo.toml
@@ -12,8 +12,8 @@ log = []
 sol-log = [ "log" ]
 
 [dependencies]
-solana-program = "1.10.33"
-bytemuck = "1.13.1"
+solana-program = "1.16"
+bytemuck = "1.13"
 thiserror = "1.0.44"
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


joncinque (commented):
Just one little thing, looks good otherwise!

libraries/concurrent-merkle-tree/src/concurrent_merkle_tree.rs
```diff
@@ -46,6 +46,7 @@ fn check_leaf_index(leaf_index: u32, max_depth: usize) -> Result<(), ConcurrentM
 /// An additional key property of ConcurrentMerkleTree is support for [append](ConcurrentMerkleTree::append) operations,
 /// which do not require any proofs to be passed. This is accomplished by keeping track of the
 /// proof to the rightmost leaf in the tree (`rightmost_proof`).
+#[repr(C)]
```
joncinque, line 49: Btw, looks like you were caught by `unsafe` code!

ngundotra: But I'm curious what the trigger was for completely different internal representation? Was it `rustc` change / upgrade?

joncinque: Yeah it could be, we'd have to look at the byte representation of old and new to check.

libraries/concurrent-merkle-tree/Cargo.toml
```diff
@@ -12,8 +12,8 @@ log = []
 sol-log = [ "log" ]
 
 [dependencies]
```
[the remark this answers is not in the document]
ngundotra: Bumped!

ngundotra:
Bumped CMT version, lmk if this looks good to go. I think Metaplex / Bubblegum is waiting on this upgrade before they can bump to 0.28 & 1.16 as well.

joncinque (approved):
Looks good to me, as long as you're sure about doing a patch release rather than a new minor version. So you're aware, this change breaks semver for your crates too now.

ngundotra:
My brain is not fully functioning today. Got it, I see. I will update the minor version on all the packages instead of the patch version.

joncinque (approved):

PR #4878, opened 2023-07-31
[token-2022] Add transfer with fee tests

samkim-crypto:
#### Problem
There are no tests for the confidential transfer with fee instruction.

#### Summary of changes
- [bcf2141](https://github.com/solana-labs/solana-program-library/pull/4878/commits/bcf21419a9f4e4ac498f0c5fae39afd0c3b411c3): There was no initialization parameters for the confidential fee extension so I added it her.
- [317aa8e](https://github.com/solana-labs/solana-program-library/pull/4878/commits/317aa8e8e8f1c1bc0c1d6baab0227dbaf89bf228): Reorganized the `ConfidentialTokenAccountMeta` type constructor used for tests. Previously, we used different constructors for whether we are initializing the account with memo, tokens, pending balance credit counter, etc. As I needed to write another constructor for initializing it with fees, I just restructured it so that there is one constructor that takes in inputs for each options.
- [10ffac2](https://github.com/solana-labs/solana-program-library/pull/4878/commits/10ffac2a12e0b2a87cc890068d7e7fdd0e418fde): There was a bug in the transfer with fee processor for self-transfers. If a transfer-with-fee is a self-transfer, then the processor executes the instruction as if it is a transfer (without fee) instruction. This made self-transfer with fees that have the `TransferWithFee` to fail since the processor will look for `Transfer` proof instead. So I fixed this.
- [4fb7821](https://github.com/solana-labs/solana-program-library/pull/4878/commits/4fb78215b3af680b594530365b3f92d9bc42f2cb): I updated the tests for transfer with fee. The previous test was quite bad/insufficient in that it didn't really test self-transfers.
- [f488c81](https://github.com/solana-labs/solana-program-library/pull/4878/commits/f488c81c01a6150bcd15cd455cb787b443eb99f0): I refactored the test component that involve the `ConfidentialTransferFee` extension to a separate file dedicated for confidential transfer fee tests.
- [3640085](https://github.com/solana-labs/solana-program-library/pull/4878/commits/36400854df6474ad386d6ed29ed67400611569ef): I added the initialize transfer fee config tests in the newly created test file.
- [7f27aa2](https://github.com/solana-labs/solana-program-library/pull/4878/commits/7f27aa2798d67b9b33e97d2797c9d5c35c18c8bb): There were some parts in the instruction doc that did not have the updated names for teh proof instructions, so I fixed these.


token/program-2022-test/tests/confidential_transfer_fee.rs
```diff
@@ -15,6 +61,143 @@ async fn check_withheld_amount_in_mint<T>(
...
+
+    assert_eq!(
+        err,
+        TokenClientError::Client(Box::new(TransportError::TransactionError(
+            TransactionError::InstructionError(
+                3,
+                InstructionError::Custom(TokenError::InvalidExtensionCombination as u32),
+            )
+        )))
+    );
```
joncinque, line 157: nit: what is this test-case for? You're doing just `ConfidentialTransferFeeConfig` right under, so this seems unnecessary

joncinque (dismissed):
Looks great! Thanks for breaking it up so nicely

token/program-2022-test/tests/confidential_transfer_fee.rs
samkim-crypto: Oh thanks for the catch. I removed it!

Issue #4817, opened 2023-07-18
[token-2022] Remaining tasks for the confidential transfer extension


samkim-crypto:
Adding a list of remaining TODOs for the confidential transfer and confidential transfer with fee extension programs.

#### Confidential Transfer
- [x] Update confidential transfer `Transfer` and `TransferWithFee` instructions for the Solana 1.16 updates
  - https://github.com/solana-labs/solana-program-library/pull/4779
- [x] Add support for context states (not split)
  - https://github.com/solana-labs/solana-program-library/pull/4816
- [x] Add support for context states (split)

#### Confidential Transfer with Fee
- [x] Divide the withheld amount in an account into lo and hi parts
  - Currently, the encrypted withheld amounts is encrypted as a full number. To ease decryption, this amount should be split into lo and hi parts like in the pending balance of the confidential transfer extension.
- [x] Update the `WithdrawWithheldTokensFromMint` and `WithdrawWithheldTokensFromAccounts` instructions for the Solana 1.16 updates (https://github.com/solana-labs/solana-program-library/pull/4896)
- [x] Add support for context states (not split)
- [ ] Add support for context states (split)

PR #4896, opened 2023-08-01
[token-2022] Update confidential transfer fee extension for solana 1.16

samkim-crypto:
#### Problem
The confidential transfer fee extension is not updated for solana 1.16

#### Summary of changes
The most importants changes in the first, second, and last commits.

[b96f57a](https://github.com/solana-labs/solana-program-library/pull/4896/commits/b96f57a632187bc21c28418bb4df8d84dbbac1fb): Updated the `WithdrawWithheldTokensFromMint` instruction for Solana 1.16 and for it to support context states. This is pretty straightforward except for the part about adding the withheld feeds directly into a destination available balance.

The original plan (https://github.com/solana-labs/solana-program-library/issues/4817) was to divide the withheld amount into low and high parts. Transfer amounts and fees in the instruction are divided into low and high bits so that they are easier to decrypt. When the transfer amount is accumulated into a destination account, the transfer amounts are added into the lo and hi pending balance fields. However, for the case of fees, they are merged and added to the single withheld fees field. Similarly, when these withheld fees are harvest into the mint, it is added to a single withheld fee field. If the withheld fees grow large then it could be inefficient for a withdraw withheld authority to decrypt the ciphertext. 

We can choose to also break up the withheld fees into lo and hi parts like the pending balance to make it easier for withdraw withheld authority to decrypt the withheld fees in an account for mint. However, I ended up deciding against it  because
- Having two ciphertexts for withheld fees in each account requires extra space
- More importantly, the instruction `WithdrawWithheldTokensFromMint` requires a `CiphertextCiphertextEqualityProof`. If there are two ciphertexts for withheld fees, then we have to either attach two instances of this zkp or batch them, which would require a new instruction in the proof program.
- In the worst case that the withdraw withheld authority cannot decrypt the withheld fees directly from the mint info, it can look up the previously list of transactions and decrypt. 

Previously, when the withheld fees in the mint are withdrawn to an account, they were added into the "lo" part of the pending balance of the destination account. This is fine, but a little unnatural. If the fees are already merged, then it should really be added into the available balance. So in this PR, I made the change. Since we are making changes to the available balance, I also added the new decryptable available balance field in the instruction.

[e64c9f0](https://github.com/solana-labs/solana-program-library/pull/4896/commits/e64c9f0b58d6b7a6b730d5f35409d31a2c026733): Updated the `WithdrawWithheldTokensFromAccount` instruction for Solana 1.16 and for it to support context states as above.

[0ed218e](https://github.com/solana-labs/solana-program-library/pull/4896/commits/0ed218e832e97802cc8968cb2d8ee77f055771e2): Small PR to unfeaturize the client function for `HarvestWithheldTokensToMint` instruction.

[dcc8574](https://github.com/solana-labs/solana-program-library/pull/4896/commits/dcc857472611006aa1346d74bd482a7e5093cf1a): Updated the tests for the `WithdrawWithheldTokensFromMint` instruction. I added the `ConfidentialTokenAccountMeta` helper struct in `tests/confidential_transfer_fee.rs`, which is a stripped down version of `ConfidentialTokensAccountMeta` in `tests/confidential_transfer.rs`. I can re-use the helper struct from `tests/confidential_transfer.rs`, but I decided to just have a separate helper struct in `test/confidential_transfer_fee.rs` that does not deal with fields like memo. The rest of the tests were straightforward.

[c582a53](https://github.com/solana-labs/solana-program-library/pull/4896/commits/c582a536fbab7a22a95ad85dcd1ce333685a2565): Updated the tests for the `WithdrawWithheldTokensFromAccounts` instruction.

[a00c50e](https://github.com/solana-labs/solana-program-library/pull/4896/commits/a00c50e23df62a1acd9f363de60cc1be6d6a1afd): Added tests that deal with context states for the `WithdrawWithheldTokensFromMint` and `WithdrawWithheldTokensFromAccounts` instructions.

[3a6b3e0](https://github.com/solana-labs/solana-program-library/pull/4896/commits/3a6b3e0dec00e10492d080f31921b6c667815635): Right now, the instruction `WithdrawWithheldTokensFromMint` could be susceptible to front-running. The zero-knowledge proof that is needed for the instruction depends on the currently withheld amount ciphertext in the mint. If the withheld amount is constantly harvested into the mint, the zkp can be nullified. 

This PR added `EnableHarvestToMint` and `DisableHarvestToMint`, which acts like `{Enable,Disable}ConfidentialTransfers` and `{Enable,Disable}NonConfidentialTransfers` for harvest. If the harvest fails due to front-running, the harvest authority can disable harvest, withdraw withheld tokens, and then enable harvest.

### DIFF
--- a/token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
+++ b/token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
@@ -153,6 +153,40 @@ pub enum ConfidentialTransferFeeInstruction {
     ///   None
     ///
     HarvestWithheldTokensToMint,
+
+    /// Configure a confidential transfer fee mint to accept harvested confidential fees.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[writable]` The SPL Token account.
+    ///   1. `[signer]` The confidential transfer fee authority.
+    ///
+    ///   *Multisignature owner/delegate
+    ///   0. `[writable]` The SPL Token account.
+    ///   1. `[]` The confidential transfer fee multisig authority,
+    ///   2. `[signer]` Required M signer accounts for the SPL Token Multisig account.
+    ///
+    /// Data expected by this instruction:
+    ///   None
+    EnableHarvestToMint,
+
+    /// Configure a confidential transfer fee mint to reject any harvested confidential fees.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[writable]` The SPL Token account.
+    ///   1. `[signer]` The confidential transfer fee authority.
+    ///
+    ///   *Multisignature owner/delegate
+    ///   0. `[writable]` The SPL Token account.
+    ///   1. `[]` The confidential transfer fee multisig authority,
+    ///   2. `[signer]` Required M signer accounts for the SPL Token Multisig account.
+    ///
+    /// Data expected by this instruction:
+    ///   None
+    DisableHarvestToMint,
 }
 
 /// Data expected by `InitializeConfidentialTransferFeeConfig`
@@ -418,6 +452,58 @@ pub fn harvest_withheld_tokens_to_mint(
         accounts,
         TokenInstruction::ConfidentialTransferFeeExtension,
         ConfidentialTransferFeeInstruction::HarvestWithheldTokensToMint,
+        &(),
+    ))
+}
+
+/// Create an `EnableHarvestToMint` instruction
+pub fn enable_harvest_to_mint(
+    token_program_id: &Pubkey,
+    mint: &Pubkey,
+    authority: &Pubkey,
+    multisig_signers: &[&Pubkey],
+) -> Result<Instruction, ProgramError> {
+    check_program_account(token_program_id)?;
+    let mut accounts = vec![
+        AccountMeta::new(*mint, false),
+        AccountMeta::new_readonly(*authority, multisig_signers.is_empty()),
+    ];
+
+    for multisig_signer in multisig_signers.iter() {
+        accounts.push(AccountMeta::new_readonly(**multisig_signer, true));
+    }
+
+    Ok(encode_instruction(
+        token_program_id,
+        accounts,
+        TokenInstruction::ConfidentialTransferFeeExtension,
+        ConfidentialTransferFeeInstruction::EnableHarvestToMint,
+        &(),
+    ))
+}
+
+/// Create a `DisableHarvestToMint` instruction
+pub fn disable_harvest_to_mint(
+    token_program_id: &Pubkey,
+    mint: &Pubkey,
+    authority: &Pubkey,
+    multisig_signers: &[&Pubkey],
+) -> Result<Instruction, ProgramError> {
+    check_program_account(token_program_id)?;
+    let mut accounts = vec![
+        AccountMeta::new(*mint, false),
+        AccountMeta::new_readonly(*authority, multisig_signers.is_empty()),
+    ];
+
+    for multisig_signer in multisig_signers.iter() {
+        accounts.push(AccountMeta::new_readonly(**multisig_signer, true));
+    }
+
+    Ok(encode_instruction(
+        token_program_id,
+        accounts,
+        TokenInstruction::ConfidentialTransferFeeExtension,
+        ConfidentialTransferFeeInstruction::DisableHarvestToMint,
         &(),
     ))
 }
--- a/token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
+++ b/token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
@@ -416,7 +416,7 @@ pub fn harvest_withheld_tokens_to_mint(
     Ok(encode_instruction(
         token_program_id,
         accounts,
-        TokenInstruction::ConfidentialTransferExtension,
+        TokenInstruction::ConfidentialTransferFeeExtension,
         ConfidentialTransferFeeInstruction::HarvestWithheldTokensToMint,
         &(),
     ))
--- a/token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
+++ b/token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
@@ -87,6 +87,9 @@ pub enum ConfidentialTransferFeeInstruction {
     /// authority. This instruction is susceptible to front-running. Use
     /// `HarvestWithheldTokensToMint` and `WithdrawWithheldTokensFromMint` as an alternative.
     ///
+    /// The withheld confidential tokens are aggregated directly into the destination available
+    /// balance.
+    ///
     /// Note on front-running: This instruction requires a zero-knowledge proof verification
     /// instruction that is checked with respect to the account state (the currently withheld
     /// fees). Suppose that a withdraw withheld authority generates the
@@ -104,7 +107,7 @@ pub enum ConfidentialTransferFeeInstruction {
     ///
     /// In order for this instruction to be successfully processed, it must be accompanied by the
     /// `VerifyWithdrawWithheldTokens` instruction of the `zk_token_proof` program in the same
-    /// transaction.
+    /// transaction or the address of a context state account for the proof must be provided.
     ///
     /// Accounts expected by this instruction:
     ///
@@ -112,7 +115,9 @@ pub enum ConfidentialTransferFeeInstruction {
     ///   0. `[]` The token mint. Must include the `TransferFeeConfig` extension.
     ///   1. `[writable]` The fee receiver account. Must include the `TransferFeeAmount` and
     ///      `ConfidentialTransferAccount` extensions.
-    ///   2. `[]` Instructions sysvar.
+    ///   2. `[]` Instructions sysvar if `VerifyCiphertextCiphertextEquality` is included in the
+    ///      same transaction or context state account if `VerifyCiphertextCiphertextEquality` is
+    ///      pre-verified into a context state account.
     ///   3. `[signer]` The mint's `withdraw_withheld_authority`.
     ///   4. ..3+N `[writable]` The source accounts to withdraw from.
     ///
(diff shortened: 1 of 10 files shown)


token/program-2022/src/extension/confidential_transfer_fee/mod.rs
```diff
@@ -34,6 +40,9 @@ pub struct ConfidentialTransferFeeConfig {
     /// fee parameters, the withheld fee amounts can reveal information about transfer amounts.
     pub withdraw_withheld_authority_elgamal_pubkey: ElGamalPubkey,
 
+    /// If `false`, the harvest of withheld tokens to mint is rejected.
+    pub harvest_to_mint_disabled: PodBool,
```
joncinque, line 44: nit: Let's flip it for ease of understanding, because the comment seems to be incorrect
```suggestion
    /// If `false`, the harvest of withheld tokens to mint is rejected.
    pub harvest_to_mint_enabled: PodBool,
```

token/program-2022/src/extension/confidential_transfer_fee/mod.rs
```diff
@@ -42,6 +51,17 @@ impl Extension for ConfidentialTransferFeeConfig {
...
+impl ConfidentialTransferFeeConfig {
+    /// Return the account information needed to construct a `WithdrawWithheldTokensFromMint`
+    /// instruction.
+    #[cfg(not(target_os = "solana"))]
+    pub fn withheld_tokens_info(&self) -> WithheldTokensInfo {
+        WithheldTokensInfo {
+            withheld_amount: self.withheld_amount,
+        }
+    }
+}
```
joncinque, line 63: this seems a bit backwards -- how about having a function on `WithheldTokensInfo` that takes `&ConfidentialTransferFeeConfig`? That way you can keep the `#[cfg(not(target_os = "solana"))]` bits all in `account_info.rs`

token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
```diff
@@ -135,6 +153,40 @@ pub enum ConfidentialTransferFeeInstruction {
     ///   None
     ///
     HarvestWithheldTokensToMint,
+
+    /// Configure a confidential transfer fee mint to accept harvested confidential fees.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[writable]` The SPL Token account.
```
joncinque, line 162: right?
```suggestion
    ///   0. `[writable]` The token mint.
```

token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
```diff
@@ -135,6 +153,40 @@ pub enum ConfidentialTransferFeeInstruction {
...
+    ///   1. `[signer]` The confidential transfer fee authority.
+    ///
+    ///   *Multisignature owner/delegate
+    ///   0. `[writable]` The SPL Token account.
```
joncinque, line 166: right?
```suggestion
    ///   0. `[writable]` The token mint.
```

token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
```diff
@@ -135,6 +153,40 @@ pub enum ConfidentialTransferFeeInstruction {
...
+    /// Data expected by this instruction:
+    ///   None
+    EnableHarvestToMint,
+
+    /// Configure a confidential transfer fee mint to reject any harvested confidential fees.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[writable]` The SPL Token account.
```
joncinque, line 179:
```suggestion
    ///   0. `[writable]` The token mint.
```

token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
```diff
@@ -135,6 +153,40 @@ pub enum ConfidentialTransferFeeInstruction {
...
+    ///   1. `[signer]` The confidential transfer fee authority.
+    ///
+    ///   *Multisignature owner/delegate
+    ///   0. `[writable]` The SPL Token account.
```
joncinque, line 183:
```suggestion
    ///   0. `[writable]` The token mint.
```

token/program-2022/src/extension/confidential_transfer_fee/processor.rs
```diff
@@ -65,22 +60,31 @@ fn process_initialize_confidential_transfer_fee_config(
     extension.authority = *authority;
     extension.withdraw_withheld_authority_elgamal_pubkey =
         *withdraw_withheld_authority_elgamal_pubkey;
+    extension.harvest_to_mint_disabled = false.into();
```
joncinque, line 63: So we don't forget to initialize it properly
```suggestion
    extension.harvest_to_mint_enabled = true.into();
```

token/program-2022/src/extension/confidential_transfer_fee/processor.rs
```diff
@@ -347,6 +390,64 @@ fn process_harvest_withheld_tokens_to_mint(accounts: &[AccountInfo]) -> ProgramR
...
+        mint.get_extension_mut::<ConfidentialTransferFeeConfig>()?;
+
+    let maybe_confidential_transfer_fee_authority: Option<Pubkey> =
+        confidential_transfer_fee_mint.authority.into();
+    let confidential_transfer_fee_authority =
+        maybe_confidential_transfer_fee_authority.ok_or(TokenError::NoAuthorityExists)?;
+
+    if !authority_info.is_signer {
+        return Err(ProgramError::MissingRequiredSignature);
+    }
```
joncinque, line 412: The comment mentions that this works with multisigs, but this check means that it only works with single authorities. It's fine if we don't want to support multisigs, but we should be clear in the instruction interface

token/program-2022/src/extension/confidential_transfer_fee/processor.rs
```diff
@@ -347,6 +390,64 @@ fn process_harvest_withheld_tokens_to_mint(accounts: &[AccountInfo]) -> ProgramR
...
```
joncinque, line 441: Same here

token/program-2022/src/extension/confidential_transfer_fee/processor.rs
```diff
@@ -367,41 +468,28 @@ pub(crate) fn process_instruction(
             )
         }
         ConfidentialTransferFeeInstruction::WithdrawWithheldTokensFromMint => {
-            msg!("ConfidentialTransferInstruction::WithdrawWithheldTokensFromMint");
-            #[cfg(all(feature = "zk-ops", feature = "proof-program"))]
```
joncinque, line 371: This removes the featurizing on `zk-ops`, so I think it'll fail to build with `zk-ops` disabled. I think we might need one more release using the feature off unfortunately, so could you restore that?

token/program-2022/src/extension/confidential_transfer_fee/processor.rs
```diff
@@ -367,41 +468,28 @@ pub(crate) fn process_instruction(
...
+            process_withdraw_withheld_tokens_from_mint(
+                program_id,
+                accounts,
+                &data.new_decryptable_available_balance,
+                data.proof_instruction_offset as i64,
+            )
         }
         ConfidentialTransferFeeInstruction::WithdrawWithheldTokensFromAccounts => {
-            msg!("ConfidentialTransferInstruction::WithdrawWithheldTokensFromAccounts");
-            #[cfg(all(feature = "zk-ops", feature = "proof-program"))]
```
joncinque, line 387: Same here, please restore the feature check

joncinque (commented):
Looks really great overall! The core logic is good, just a few little things to sort out

token/program-2022/src/extension/confidential_transfer_fee/processor.rs
samkim-crypto: Yes, sorry I missed this!

token/program-2022/src/extension/confidential_transfer_fee/mod.rs
samkim-crypto: Yes good call.

token/program-2022/src/extension/confidential_transfer_fee/instruction.rs
samkim-crypto: Yes, thanks for the catch!

token/program-2022/src/extension/confidential_transfer_fee/mod.rs
samkim-crypto: Yeah that is actually a good idea. Since `WithheldTokensInfo` just has one field `withheld_amount`, I think we can just work with the existing constructor `new(withheld_amount: EncryptedWithheldAmount) -> Self` rather than taking in the entire confidential transfer fee config.

But the suggestion appiles for the account info types in the regular confidential transfer extension, so I will apply the suggestion there.

samkim-crypto: https://github.com/solana-labs/solana-program-library/pull/4982

token/program-2022/src/extension/confidential_transfer_fee/processor.rs
```diff
@@ -367,41 +470,35 @@ pub(crate) fn process_instruction(
...
                     data.num_token_accounts,
+                    &data.new_decryptable_available_balance,
                     data.proof_instruction_offset as i64,
-                );
-            }
-            #[cfg(not(all(feature = "zk-ops", feature = "proof_program")))]
-            {
-                Err(ProgramError::InvalidInstructionData)
+                )
             }
```
joncinque, line 498: You'll also need the Error variant if the feature is off

token/program-2022/src/extension/confidential_transfer_fee/processor.rs
```diff
@@ -367,41 +470,35 @@ pub(crate) fn process_instruction(
...
                     accounts,
+                    &data.new_decryptable_available_balance,
                     data.proof_instruction_offset as i64,
-                );
-            }
-            #[cfg(not(all(feature = "zk-ops", feature = "proof_program")))]
-            {
-                Err(ProgramError::InvalidInstructionData)
+                )
             }
```
joncinque, line 483: You'll also need the error variant if the feature is off

joncinque (approved):
Just one last little nit, then this is good to go!

token/program-2022/src/extension/confidential_transfer_fee/processor.rs
samkim-crypto: Oh sorry I missed it!

PR #4913, opened 2023-08-01
stake-pool: Update tests for 1.17

joncinque:
#### Problem

The SPL downstream tests were disabled in https://github.com/solana-labs/solana/pull/32524, but we need to re-enable the downstream tests to avoid breaking people in the future, per https://github.com/solana-labs/solana/pull/32655.

After doing the work to reinstate the old `StakeState` and re-enabling the downstream tests, there were errors in the stake pools tests.

#### Solution

The biggest change relates to the async rewards payout being implemented for https://github.com/solana-foundation/solana-improvement-documents/pull/15 -- we have to warp to `slot + 1` in order to get out of the rewards payout time.

After that, there was one issue where we weren't serializing 200 bytes for stake accounts in a test.

And a little extra thing, I noticed that `retain` in the stake pool program was using a function pointer instead of a closure, so I changed to be a closure, same as all the other `big_vec` functions. This should improve compute-unit performance because the Rust compiler can inline closures, but not calling into function pointers.

### DIFF
--- a/stake-pool/program/tests/helpers/mod.rs
+++ b/stake-pool/program/tests/helpers/mod.rs
@@ -2379,12 +2379,14 @@ pub fn add_validator_stake_account(
         credits_observed: 0,
     };
 
+    let mut data = bincode::serialize::<stake::state::StakeState>(
+        &stake::state::StakeState::Stake(meta, stake),
+    )
+    .unwrap();
+    data.extend_from_slice(&[0, 0, 0, 0]);
     let stake_account = SolanaAccount::create(
         stake_amount + STAKE_ACCOUNT_RENT_EXEMPTION,
-        bincode::serialize::<stake::state::StakeState>(&stake::state::StakeState::Stake(
-            meta, stake,
-        ))
-        .unwrap(),
+        data,
         stake::program::id(),
         false,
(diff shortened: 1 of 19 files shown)


stake-pool/program/tests/helpers/mod.rs
```diff
@@ -2379,12 +2379,14 @@ pub fn add_validator_stake_account(
         credits_observed: 0,
     };
 
+    let mut data = bincode::serialize::<stake::state::StakeState>(
+        &stake::state::StakeState::Stake(meta, stake),
+    )
+    .unwrap();
+    data.extend_from_slice(&[0, 0, 0, 0]);
```
2501babe, line 2386: i havent run the code to see what happens but shouldnt this have a check that the length is 196 before extending? otherwise when we move to 1.17 we would extend to 204

2501babe (dismissed):

2501babe:
otherwise lgtm

stake-pool/program/tests/helpers/mod.rs
joncinque: good point, I'll instead allocate a bigger vec and then write into it

PR #4916, opened 2023-08-02
token-cli: Support transfer hook extension

### LINKED ISSUE #4548: token-cli: Support transfer hook fully
#### Problem

The transfer-hook used by token-2022 complicates transfers since you may need to fetch additional accounts before creating a transfer instruction, and those accounts may be provided for offline signing.

On top of that, the regular create + update functionality needs to be provided too.

#### Solution

Add CLI support for the transfer hook extension

joncinque:
#### Problem

There's no support for the transfer hook extension in the token CLI.

#### Solution

Add basic support for setting a transfer hook at mint creation, updating the program, and specifying transfer-hook-accounts for offline signing.

Fixes #4548

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -3156,7 +3156,10 @@ fn app<'a, 'b>(
                         .takes_value(true)
                         .multiple(true)
                         .min_values(0u64)
-                        .help("Additional pubkey(s) required for a transfer hook and their role. The role must be \"readonly\", \"writable\". \"readonly-signer\", or \"wrUsed for offline transaction creation and signing.")
+                        .help("Additional pubkey(s) required for a transfer hook and their \
+                            role, in the format \"<PUBKEY>:<ROLE>\". The role must be \
+                            \"readonly\", \"writable\". \"readonly-signer\", or \"writable-signer\".\
+                            Used for offline transaction creation and signing.")
                 )
                 .arg(multisig_signer_arg())
                 .arg(mint_decimals_arg())
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -28,6 +28,7 @@ use solana_cli_output::{
 use solana_client::rpc_request::TokenAccountsFilter;
 use solana_remote_wallet::remote_wallet::RemoteWalletManager;
 use solana_sdk::{
+    instruction::AccountMeta,
     native_token::*,
     program_option::COption,
     pubkey::Pubkey,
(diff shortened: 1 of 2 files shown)


token/cli/src/main.rs
```diff
@@ -2747,6 +2832,43 @@ fn app<'a, 'b>(
...
+                    Arg::with_name("token")
+                        .validator(is_valid_pubkey)
+                        .value_name("TOKEN_MINT_ADDRESS")
+                        .takes_value(true)
+                        .required(true)
+                        .index(1)
+                        .help("The token address with an existing transfer hook"),
+                )
+                .arg(
+                    Arg::with_name("new_transfer_hook_program_id")
```
buffalojoec, line 2848: Would just `--program-id` be bad? I don't see where it would collide, and the subcommand is already within the transfer hook scope

token/cli/src/main.rs
```diff
@@ -2978,6 +3101,16 @@ fn app<'a, 'b>(
...
+                .arg(
+                    Arg::with_name("transfer_hook_account")
+                        .long("transfer-hook-account")
+                        .validator(is_valid_pubkey)
+                        .value_name("PUBKEY")
+                        .takes_value(true)
+                        .multiple(true)
+                        .min_values(0u64)
+                        .help("Additional pubkey(s) required for a transfer hook. Used for offline transaction creation and signing.")
+                )
```
buffalojoec, line 3113: In regards to your reply in #4548 , I'm much more in favor of the second suggestion:

```
--transfer-hook-account <PUBKEY>:readonlysigner
```

However, I'm also curious about using an array?

buffalojoec (commented):
Code lgtm, I asked a couple questions and kicked off the chat about arg naming convention

token/cli/src/main.rs
joncinque: It's entirely a code thing, since you have to pass this parameter unless `--disable` is passed, and not via `--new-transfer-hook-program-id`. We can split the difference and go `"new_program_id"`?

token/cli/src/main.rs
joncinque: Yeah that makes sense, I'm just lazy :sweat_smile: 

What do you mean about using an array?

token/cli/src/main.rs
```diff
@@ -1060,7 +1132,12 @@ async fn command_transfer(
         Some(mint_info.decimals)
     };
 
-    let token = token_client_from_config(config, &token_pubkey, decimals)?;
+    let token = if let Some(transfer_hook_accounts) = transfer_hook_accounts {
+        token_client_from_config(config, &token_pubkey, decimals)?
+            .with_transfer_hook_accounts(transfer_hook_accounts)
+    } else {
+        token_client_from_config(config, &token_pubkey, decimals)?
+    };
```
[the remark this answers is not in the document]
joncinque: It's mostly for offline signing, since the validation account is required for the interface. maybe that means we can just include it if any transfer hook accounts are included :thinking:

token/cli/src/main.rs
buffalojoec: Ahh I see. Yeah, I think shorter the better for CLI args; `new_program_id` is cool

token/cli/src/main.rs
buffalojoec: Does `clap` support - without too much clunkiness - a value that's an array? like:

```
--transfer-hook-accounts [<PUBKEY>:readonlysigner, <PUBKEY>:readonly, <PUBKEY>:readonly]
```

and then maybe optionally this can accept a path to a JSON or something

token/cli/src/main.rs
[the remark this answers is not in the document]
buffalojoec: Ahh right because transfer hook program and validation account are expected in this vector, ie. https://github.com/solana-labs/solana-program-library/blob/37d21c6493de474f72e247931dbb42274ea002ba/token/transfer-hook-interface/src/offchain.rs#L57C5-L61C78

Because of the way that offchain helper is written, you'd have to validate that they haven't included either already and add them for them, or don't if they have. But might be a nice touch for DX :)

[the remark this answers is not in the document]
buffalojoec: Idk, I am leaning towards it might not be necessary to try to automagically include any metas for them for offline signing. Keep it simple. If you want to sign offline add each key according to: https://github.com/solana-labs/solana-program-library/blob/master/token/transfer-hook-interface/src/instruction.rs#L22C5-L27C82

token/cli/src/main.rs
joncinque: By convention our tools use repeated args in case we want to provide it more than once. And personally, I find array arguments extremely unclear and clunky because there's no real convention. Typically I've seen it as:

```
--transfer-hook-accounts pubkey:type,pubkey:type,pubkey:type
```

So you split on the comma, which makes writing long arguments very ugly. To implement your suggestion, we'd need to write a specialized parser that looks at the `[` and `]` characters and goes through spaces, which is not typically how arg parsing is implemented.

Anyway, I'll add a new enum for the different meta combinations and parse it properly, ie:

```
enum AccountMetaRole {
    Readonly,
    Writable,
    ReadonlySigner,
    WritableSigner,
}
```

token/cli/src/main.rs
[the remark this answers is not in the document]
joncinque: Actually, we don't even know the program id when we're offline, so we can't even derive the proper address, unless we keep the transfer hook program id as a separate argument. Yeah, let's keep it simple

token/cli/src/main.rs
buffalojoec: Fair point(s)! Probably an anti-pattern and looks pretty ugly the more I stare at the comma-separated items.

If repeat args is the pattern then all good

token/cli/src/main.rs
joncinque: Done!

token/cli/src/main.rs
joncinque: Done!

joncinque:

token/cli/src/main.rs
```diff
@@ -3104,12 +3151,12 @@ fn app<'a, 'b>(
...
                         .long("transfer-hook-account")
-                        .validator(is_valid_pubkey)
-                        .value_name("PUBKEY")
+                        .validator(validate_transfer_hook_account)
+                        .value_name("PUBKEY:ROLE")
                         .takes_value(true)
                         .multiple(true)
                         .min_values(0u64)
-                        .help("Additional pubkey(s) required for a transfer hook. Used for offline transaction creation and signing.")
+                        .help("Additional pubkey(s) required for a transfer hook and their role. The role must be \"readonly\", \"writable\". \"readonly-signer\", or \"wrUsed for offline transaction creation and signing.")
```
buffalojoec, line 3159: I think this got cut off here?

buffalojoec (commented):
Just one thing I noticed then should be good to go!

token/cli/src/main.rs
joncinque: ah nice, thanks for noticing! fixed

buffalojoec (approved):
ship it !

PR #4982, opened 2023-08-10
[token-2022] Add constructors for confidential transfer account info types

samkim-crypto:
#### Problem
This is a follow-up to https://github.com/solana-labs/solana-program-library/pull/4896#discussion_r1289197648.

#### Summary of changes
Added the constructors for the account info types in the confidential transfer instructions. Previously, the account information types had to be constructed using the associated function of `ConfidentialTransferAccount` type, which was quite unnatural.


joncinque (dismissed):
Just looked at the last commit, and it's good!

PR #4983, opened 2023-08-10
ci: Test purging more packages

joncinque:
#### Problem

https://github.com/solana-labs/solana/pull/32677 is failing due to lack of disk space, but there's definitely more stuff to blast in the ubuntu runner.

#### Solution

Remove more packages to get another 7GB

### DIFF
--- a/.github/workflows/pull-request-stake-pool.yml
+++ b/.github/workflows/pull-request-stake-pool.yml
@@ -6,6 +6,7 @@ on:
     - 'stake-pool/**'
     - 'token/**'
     - 'ci/*-version.sh'
+    - 'ci/warning/purge-ubuntu-runner.sh'
     - '.github/workflows/pull-request-stake-pool.yml'
   push:
     branches: [master]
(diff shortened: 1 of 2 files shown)


joncinque:
@yihau what do you think? Wanna try the downstream PR with another 7GB and see if we can get it to pass? If not, we can break up the SPL downstream job into a few pieces, ie. one for token / ATA / token-2022, one for stake pools, one for governance

yihau:
hey! thank you for the updating! tbh, I'm struggling with this one today. I delete more things each time. I'm waiting for the last try. if it still can't make it. I think I will try to split the test as you said.

the build: https://github.com/yihau/solana/actions/runs/5821637132/job/15784476314
the purge script: https://github.com/yihau/solana/blob/master/.github/scripts/purge-ubuntu-runner.sh
(this script get 45G+ more space)

yihau:
it failed again. I will spilt it later. 😞

joncinque:
I ended up just using your better script here :heart:

yihau (approved):
🪖

PR #4986, opened 2023-08-10
Tighten Up Account Resolution Methods

buffalojoec:
This PR swaps methods from `ExtraAccountMetaList` to `ExtraAccountMeta` and tightens up implementation.

This sets the stage for:
- Removal of `AccountInfo` helpers
- Updated integration with `spl-transfer-hook-example`

Summary:
- Moved `de_escalate_account_meta` and `resolve_pda` into `account.rs` as private helpers
- Added `resolve` to `ExtraAccountMeta`
- Trimmed implementation of `add_to_instruction` for `ExtraAccountMetaList`

Note: As it relates to #4963:
- An `is_pda()` or enum is no longer required
- An `AccountKeyable` trait is no longer required

### DIFF
--- a/libraries/tlv-account-resolution/src/account.rs
+++ b/libraries/tlv-account-resolution/src/account.rs
@@ -7,11 +7,65 @@ use {
     crate::{error::AccountResolutionError, seeds::Seed},
     bytemuck::{Pod, Zeroable},
     solana_program::{
-        account_info::AccountInfo, instruction::AccountMeta, program_error::ProgramError,
+        account_info::AccountInfo,
+        instruction::{AccountMeta, Instruction},
+        program_error::ProgramError,
         pubkey::Pubkey,
     },
     spl_type_length_value::pod::PodBool,
 };
+
+/// De-escalate an account meta if necessary
+fn de_escalate_account_meta(account_meta: &mut AccountMeta, account_metas: &[AccountMeta]) {
+    // This is a little tricky to read, but the idea is to see if
+    // this account is marked as writable or signer anywhere in
+    // the instruction at the start. If so, DON'T escalate it to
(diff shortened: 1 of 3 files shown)


libraries/tlv-account-resolution/src/account.rs
```diff
@@ -61,6 +115,23 @@ impl ExtraAccountMeta {
...
+                pubkey: resolve_pda(&seeds, instruction)?,
+                is_signer: self.is_signer.into(),
+                is_writable: self.is_writable.into(),
+            }
+        } else {
+            AccountMeta::try_from(self)?
+        };
+        de_escalate_account_meta(&mut account_meta, &instruction.accounts);
+        Ok(account_meta)
+    }
```
joncinque, line 134: Is there a reason that this changed from a match statement to an if statement? The match statement is more exhaustive, so I'd prefer to stick with that.

joncinque (commented):
Just the question about the match statement. In general, if you're just moving functions around, it's easier to review if there's one commit that just moves everything, and then a separate commit with functional changes.

libraries/tlv-account-resolution/src/account.rs
buffalojoec: It can definitely be a match statement!

In #4963 I had changed it to `if meta.is_pda()`, and when you do `AccountMeta::try_from(self)?` it will error on any discriminators outside of `0`, so I figured cases were covered.

But happy to convert to match

buffalojoec:
> Just the question about the match statement. In general, if you're just moving functions around, it's easier to review if there's one commit that just moves everything, and then a separate commit with functional changes.

Cool sounds good, I can do that next time

joncinque (approved):
great, thanks!

PR #4987, opened 2023-08-10
Feature Gate Program

buffalojoec:
This PR adds a new program to SPL: The Feature Gate Program.

This is the exact program proposed in [this issue on the Solana monorepo](https://github.com/solana-labs/solana/issues/32780).

In summary, this program takes the place of the otherwise empty and currently utilized account address at `Feature111111111111111111111111111111111111 `. Once we've replaced that account with this program, it will be able to control (and most importantly **revoke**) pending feature activations, as described in [the linked issue](https://github.com/solana-labs/solana/issues/32780).

### DIFF
--- a/feature-gate/cli/src/main.rs
+++ b/feature-gate/cli/src/main.rs
@@ -1,0 +1,268 @@
+#![allow(clippy::integer_arithmetic)]
+
+use {
+    clap::{crate_description, crate_name, crate_version, App, AppSettings, Arg, SubCommand},
+    solana_clap_utils::{
+        input_parsers::{keypair_of, pubkey_of},
+        input_validators::{is_keypair, is_url, is_valid_pubkey},
+    },
+    solana_client::rpc_client::RpcClient,
+    solana_sdk::{
+        commitment_config::CommitmentConfig,
+        feature::Feature,
+        pubkey::Pubkey,
+        rent::Rent,
+        signature::{read_keypair_file, Keypair, Signer},
+        system_instruction,
+        transaction::Transaction,
+    },
+    spl_feature_gate::instruction::{activate, revoke},
+};
+
+fn keypair_clone(kp: &Keypair) -> Keypair {
+    Keypair::from_bytes(&kp.to_bytes()).expect("failed to copy keypair")
+}
+
+#[allow(dead_code)]
+struct Config {
+    keypair: Keypair,
+    json_rpc_url: String,
+    verbose: bool,
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let app_matches = App::new(crate_name!())
+        .about(crate_description!())
+        .version(crate_version!())
+        .setting(AppSettings::SubcommandRequiredElseHelp)
+        .arg({
+            let arg = Arg::with_name("config_file")
+                .short("C")
+                .long("config")
+                .value_name("PATH")
+                .takes_value(true)
+                .global(true)
+                .help("Configuration file to use");
+            if let Some(ref config_file) = *solana_cli_config::CONFIG_FILE {
+                arg.default_value(config_file)
+            } else {
+                arg
+            }
+        })
+        .arg(
+            Arg::with_name("keypair")
+                .long("keypair")
+                .value_name("KEYPAIR")
+                .validator(is_keypair)
+                .takes_value(true)
+                .global(true)
+                .help("Filepath or URL to a keypair [default: client keypair]"),
+        )
+        .arg(
+            Arg::with_name("verbose")
+                .long("verbose")
+                .short("v")
+                .takes_value(false)
+                .global(true)
+                .help("Show additional information"),
+        )
+        .arg(
+            Arg::with_name("json_rpc_url")
+                .long("url")
+                .value_name("URL")
+                .takes_value(true)
+                .global(true)
+                .validator(is_url)
+                .help("JSON RPC URL for the cluster [default: value from configuration file]"),
+        )
+        .subcommand(
+            SubCommand::with_name("activate")
+                .about("Activate a feature")
+                .arg(
+                    Arg::with_name("feature_keypair")
+                        .value_name("FEATURE_KEYPAIR")
+                        .validator(is_keypair)
+                        .index(1)
+                        .required(true)
+                        .help("Path to keypair of the feature"),
+                )
+                .arg(
+                    Arg::with_name("authority_keypair")
+                        .value_name("AUTHORITY_KEYPAIR")
+                        .validator(is_keypair)
+                        .required(true)
+                        .help("Path to keypair of the authority"),
+                )
+                .arg(
+                    Arg::with_name("payer_keypair")
+                        .value_name("PAYER_KEYPAIR")
+                        .validator(is_keypair)
+                        .help(
+                            "Path to keypair of the payer to fund the feature account (defaults \
+                             to authority)",
+                        ),
+                ),
+        )
+        .subcommand(
+            SubCommand::with_name("revoke")
+                .about("Revoke a pending feature activation")
+                .arg(
+                    Arg::with_name("feature_id")
+                        .value_name("FEATURE_ID")
+                        .validator(is_valid_pubkey)
+                        .index(1)
+                        .required(true)
+                        .help("The address of the feature (feature ID)"),
+                )
+                .arg(
+                    Arg::with_name("destination")
+                        .value_name("DESTINATION")
+                        .validator(is_valid_pubkey)
+                        .index(2)
+                        .required(true)
+                        .help("The address of the destination for the refunded lamports"),
+                )
+                .arg(
+                    Arg::with_name("authority_keypair")
+                        .value_name("AUTHORITY_KEYPAIR")
+                        .validator(is_keypair)
+                        .required(true)
+                        .help("Path to keypair of the authority"),
+                ),
+        )
+        .get_matches();
+
+    let (sub_command, sub_matches) = app_matches.subcommand();
+    let matches = sub_matches.unwrap();
+
+    let config = {
+        let cli_config = if let Some(config_file) = matches.value_of("config_file") {
+            solana_cli_config::Config::load(config_file).unwrap_or_default()
+        } else {
+            solana_cli_config::Config::default()
+        };
+
+        Config {
+            json_rpc_url: matches
+                .value_of("json_rpc_url")
+                .unwrap_or(&cli_config.json_rpc_url)
+                .to_string(),
+            keypair: read_keypair_file(
+                matches
+                    .value_of("keypair")
+                    .unwrap_or(&cli_config.keypair_path),
+            )?,
+            verbose: matches.is_present("verbose"),
+        }
+    };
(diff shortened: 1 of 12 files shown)


feature-gate/program/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "spl-feature-gate"
+version = "0.0.1"
```
joncinque, line 3: nit: maybe start with 0.1.0?

feature-gate/program/src/instruction.rs
```diff
@@ -0,0 +1,87 @@
...
+    borsh::{BorshDeserialize, BorshSchema, BorshSerialize},
+    solana_program::{
+        instruction::{AccountMeta, Instruction},
+        pubkey::Pubkey,
+        system_program,
+    },
+};
+
+/// Feature Gate program instructions
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, BorshSchema, PartialEq)]
```
joncinque, line 13: Can we use `num_enum` for this instead of `borsh`? It'll be much easier for encoding a u8

feature-gate/program/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+//! Feature Gate program
+
+#![allow(clippy::integer_arithmetic)]
```
joncinque, line 3: Unless I'm missing something, I don't see a need to allow integer arithmetic in this program. Let's be tough!

feature-gate/program/src/instruction.rs
```diff
@@ -0,0 +1,87 @@
...
+pub enum FeatureGateInstruction {
+    /// Submit a feature for activation.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[ws]` Feature account (must be a system account)
+    ///   1. `[s]` Authority
```
joncinque, line 20: What's this supposed to do?

feature-gate/program/src/instruction.rs
```diff
@@ -0,0 +1,87 @@
...
+    ///   3. `[]` System program
```
joncinque, line 21: nit:
```suggestion
    ///   2. `[]` System program
```

feature-gate/program/src/instruction.rs
```diff
@@ -0,0 +1,87 @@
...
+    Activate,
+    /// Revoke a pending feature activation.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[w]` Feature account
+    ///   1. `[w]` Destination (for rent lamports)
+    ///   2. `[s]` Authority
+    Revoke,
```
joncinque, line 30: Let's make the action really clear everywhere
```suggestion
    RevokePendingActivation,
```

feature-gate/program/src/processor.rs
```diff
@@ -0,0 +1,89 @@
...
+pub fn process_revoke(program_id: &Pubkey, accounts: &[AccountInfo]) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+
+    let feature_info = next_account_info(account_info_iter)?;
+    let destination_info = next_account_info(account_info_iter)?;
+    let authority_info = next_account_info(account_info_iter)?;
+
+    if !authority_info.is_signer {
+        return Err(ProgramError::MissingRequiredSignature);
+    }
```
joncinque, line 56: This currently allows *anyone* to revoke a feature as long as you have any random signature aftewards, which could be pretty catastrophic :sweat_smile: 

Since activation requires a signature from the feature id keypair, let's require a signature from the feature id keypair for revoke, and not worry about some other authority. What do you think?

Unless I'm missing something -- have there been discussions around an extra authority?

feature-gate/program/src/processor.rs
```diff
@@ -0,0 +1,89 @@
...
+    let new_destination_lamports = feature_info
+        .lamports()
+        .checked_add(destination_info.lamports())
+        .ok_or::<ProgramError>(FeatureGateError::Overflow.into())?;
+
+    **feature_info.try_borrow_mut_lamports()? = 0;
+    **destination_info.try_borrow_mut_lamports()? = new_destination_lamports;
+
+    feature_info.realloc(0, true)?;
+    feature_info.assign(&system_program::id());
```
joncinque, line 71: :pray: thank you for doing this

feature-gate/program/src/processor.rs
```diff
@@ -0,0 +1,89 @@
...
+    let destination_info = next_account_info(account_info_iter)?;
+    let authority_info = next_account_info(account_info_iter)?;
+
+    if !authority_info.is_signer {
+        return Err(ProgramError::MissingRequiredSignature);
+    }
+
+    if feature_info.owner != program_id {
+        return Err(ProgramError::IllegalOwner);
+    }
```
joncinque, line 60: On top of this, we need to make sure that the data is all zeroes, otherwise this also allows someone to deactivate an *active* feature.

feature-gate/program/src/instruction.rs
```diff
@@ -0,0 +1,87 @@
...
+        system_program,
+    },
+};
+
+/// Feature Gate program instructions
+#[derive(Clone, Debug, BorshSerialize, BorshDeserialize, BorshSchema, PartialEq)]
+pub enum FeatureGateInstruction {
+    /// Submit a feature for activation.
+    ///
+    /// Accounts expected by this instruction:
```
joncinque, line 17: Let's make it clear that this doesn't fund the rent-exempt minimum, and isn't required since you can just create the account separately.

Unless I missed some conversations about how this should work

feature-gate/program/src/instruction.rs
```diff
@@ -0,0 +1,87 @@
...
+    ///
+    ///   0. `[w]` Feature account
+    ///   1. `[w]` Destination (for rent lamports)
+    ///   2. `[s]` Authority
+    Revoke,
+}
+
+/// Creates an 'Activate' instruction.
+pub fn activate(program_id: &Pubkey, feature: &Pubkey, authority: &Pubkey) -> Instruction {
```
joncinque, line 34: Since the funding isn't done here, let's add another helper which also includes the `system_program::transfer` instruction

feature-gate/program/tests/functional.rs
```diff
@@ -0,0 +1,235 @@
...
+        transaction::{Transaction, TransactionError},
+    },
+    spl_feature_gate::{
+        error::FeatureGateError,
+        instruction::{activate, revoke},
+    },
+};
+
+#[tokio::test]
+async fn test_functional() {
```
joncinque, line 21: nit: we can separate the activate and revoke tests

joncinque (commented):
Great start! I did a pass over the whole thing, but let's nail down the interface first and foremost, then we can look at the logic more closely.

feature-gate/cli/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+                )
+                .arg(
+                    Arg::with_name("payer_keypair")
+                        .value_name("PAYER_KEYPAIR")
+                        .validator(is_keypair)
+                        .help(
+                            "Path to keypair of the payer to fund the feature account (defaults \
+                             to authority)",
+                        ),
+                ),
```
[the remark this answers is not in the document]
buffalojoec: I could do that! I thought it might be nice to give the option to separate the rent funder from authority

feature-gate/program/src/instruction.rs
buffalojoec: I'm basically giving the program the ability to do [this](https://github.com/solana-labs/solana/blob/master/sdk/program/src/feature.rs#L53-L57)

joncinque: Sorry, I meant the `authority` specifically

feature-gate/program/src/processor.rs
buffalojoec: Correct and nope, I honestly have no idea how the current authority setup works, that's kind where I'm leaving off on this thing until we see how the talks go

buffalojoec: Feature keypair for revoke makes sense, since the contributor is instructed to hold onto this keypair right?

buffalojoec: Do we require an authority to activate? I guess to my understanding, anyone can make a feature account and mark it owned by `Feature111111111111111111111111111111111111 `, but if it's not in the [feature set](https://github.com/solana-labs/solana/blob/f4287d70bb78307f16949509e013fb5e6f15a751/sdk/src/feature_set.rs) it doesn't get picked up (?)

joncinque: Yeah that's exactly it. So it's all done by the feature keypair. I think for v1 of this program, we keep the same paradigm

feature-gate/program/src/instruction.rs
buffalojoec: Prob going to remove as a result of your other comment below. I was under the impression there was some authority additional to the feature keypair involved. I get the picture now I think

buffalojoec:
> Great start! I did a pass over the whole thing, but let's nail down the interface first and foremost, then we can look at the logic more closely.

Sounds good. I addressed any feedback you had for now, and we can just let this sit until talks finalize on the linked issue/discussion.

mvines:
> Great start! I did a pass over the whole thing, but let's nail down the interface first and foremost, then we can look at the logic more closely.

Do you mean the Revoke instruction interface? If so, wouldn't it be better to land just an empty noop feature gate program for now?  And deal with the Revoke instruction once we're waiting for the noop program to be shipped to all clusters?

joncinque:
> Do you mean the Revoke instruction interface? If so, wouldn't it be better to land just an empty noop feature gate program for now? And deal with the Revoke instruction once we're waiting for the noop program to be shipped to all clusters?

I've been wondering why you bring up the no-op program so much... I might be missing something, but we can simply ship the feature in the validator without any program deployed at the address. By the time the feature's ready to be enabled, we'll have the program ready

Do you expect the no-op program uploaded to the source address before shipping the validator change?

mvines:
Got it. Yeah so given how changing this part of the system is incredibly sensitive, my mind immediately wants to break down the change into many small steps that we can trivially (or close to) conclude will be safe.

Step 1. Install a no-op feature gate program at the right address. This requires the runtime changes and a bpf program to swap in. If the bpf program is empty, now we only need to narrowly focus on thinking about and auditing the runtime changes.

Step 2. With the scary runtime changes out of the way, and an upgradable feature-gate program installed, we can then turn our narrow focus on thinking about and auditing the program-specific changes to support revoking a pending activation.

Combining these steps to me just unnecessarily increases the risk that we'll overlook something important

joncinque:
Maybe I'm being a little :cowboy_hat_face: but those two steps are well-separated, with or without a no-op program. On the flipside, if the upgrade goes wrong, then a no-op program can't do any damage, so maybe that's the biggest benefit.

Either way, we can table this until https://github.com/solana-labs/solana/pull/32783 goes in if it's too much all at once

buffalojoec:
Closing in favor of #5510

PR #4988, opened 2023-08-10
condense TLV AR state to one init

buffalojoec:
This PR sets the stage for modifying the Transfer Hook Interface `InitializeExtraAccountMetas` instruction.

Here I'm condensing the act of initializing `ExtraAccountMetaList` state into one `init` function, rather than 3

### DIFF
--- a/libraries/tlv-account-resolution/src/state.rs
+++ b/libraries/tlv-account-resolution/src/state.rs
@@ -191,7 +191,7 @@ mod tests {
                 .iter()
                 .map(ExtraAccountMeta::from)
                 .collect::<Vec<_>>(),
-            metas.to_vec()
+            metas
         );
     }
 
@@ -257,12 +257,12 @@ mod tests {
                 .iter()
                 .map(ExtraAccountMeta::from)
                 .collect::<Vec<_>>(),
-            account_infos.to_vec()
+            account_infos
         );
     }
 
     #[test]
-    fn init_with_pod_account_metas() {
+    fn init_with_extra_account_metas() {
         let program_id = Pubkey::new_unique();
 
         let extra_meta3_literal_str = "seed_prefix";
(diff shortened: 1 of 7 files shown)


libraries/tlv-account-resolution/src/state.rs
```diff
@@ -232,7 +191,7 @@ mod tests {
                 .iter()
                 .map(ExtraAccountMeta::from)
                 .collect::<Vec<_>>(),
-            metas.iter().map(ExtraAccountMeta::from).collect::<Vec<_>>()
+            metas.to_vec()
```
joncinque, line 194: micro-nit: is the `to_vec()` needed? it should be able to do the equality, maybe by slicing?

libraries/tlv-account-resolution/src/state.rs
```diff
@@ -299,10 +257,7 @@ mod tests {
                 .iter()
                 .map(ExtraAccountMeta::from)
                 .collect::<Vec<_>>(),
-            account_infos
-                .iter()
-                .map(ExtraAccountMeta::from)
-                .collect::<Vec<_>>()
+            account_infos.to_vec()
```
joncinque, line 260: micro-nit: same here, maybe no `to_vec()` needed?

libraries/tlv-account-resolution/src/state.rs
```diff
@@ -661,10 +603,7 @@ mod tests {
...
```
joncinque, line 606: micro-nit: And last one

joncinque (approved):
Beautiful! Just the nits on the test

libraries/tlv-account-resolution/src/state.rs
buffalojoec: Weird I swore it barfed on me hah! Dropped

joncinque (approved):
Great job!

PR #4989, opened 2023-08-10
Add link to doc: instruction.rs

skaunov:
(Seems like I forgot to create the PR after editing.) :roll_eyes:


skaunov:
Oh it's actually a duplicate. Sorry, was showing like a change!

PR #4999, opened 2023-08-11
Transfer Hook: Provide extra metas as instruction data

buffalojoec:
This PR modifies the Transfer Hook Interface to take configurations for extra required account metas as instruction data rather than in the list of account infos.


joncinque (approved):
Beautiful! Simple and clean, thanks

Issue #4942, opened 2023-08-04
Update docs for Transfer Hook & TLV Account Resolution


buffalojoec:
### Problem
Now that PDA account resolution has been added to `spl-tlv-account-resolution`, we need some updated documentation!

### Solution
- Update the documentation for Transfer Hook
- Update the documentation for TLV Account Resolution
- Release new versions

PR #5000, opened 2023-08-11
Transfer Hook Example with PDAs

buffalojoec:
This PR modifies the transfer hook example program to demonstrate use of PDAs as extra account configs

Ref #4942

### DIFF
--- a/libraries/tlv-account-resolution/src/state.rs
+++ b/libraries/tlv-account-resolution/src/state.rs
@@ -3,7 +3,10 @@
 use {
     crate::{account::ExtraAccountMeta, error::AccountResolutionError},
     solana_program::{
-        account_info::AccountInfo, instruction::Instruction, program_error::ProgramError,
+        account_info::AccountInfo,
+        instruction::{AccountMeta, Instruction},
+        program_error::ProgramError,
+        pubkey::Pubkey,
     },
     spl_discriminator::SplDiscriminate,
     spl_type_length_value::{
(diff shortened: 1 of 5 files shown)


joncinque (commented):
Very cool to see this in action!

For an easier downstream dev experience, what do you think about refactoring the resolution bit from `ExtraAccountMeta::resolve` to take in the instruction components separately? So rather than taking `Instruction`, it takes the instruction data, pubkeys, and program id.

That way, an on-chain program can re-resolve to make sure all the inputs are correct. And then we won't have to hard-code anything in the example program, and people can copy it right away.

buffalojoec:
> Very cool to see this in action!
> 
> 
> 
> For an easier downstream dev experience, what do you think about refactoring the resolution bit from `ExtraAccountMeta::resolve` to take in the instruction components separately? So rather than taking `Instruction`, it takes the instruction data, pubkeys, and program id.
> 
> 
> 
> That way, an on-chain program can re-resolve to make sure all the inputs are correct. And then we won't have to hard-code anything in the example program, and people can copy it right away.

Ahhh you're too quick!!

Yeah, I was just looking into this right now after seeing the test failure for `program-2022-test`. 

I’m curious your thoughts, but sounds like we are mostly in agreement. 

I was kind of thinking:
- On one hand, we shouldn't need to loop over TLV state to validate the provided extra metas, since with any on-chain program that requires them, it's much more likely they'd have a specific use for these accounts, therefore a specific check and/or business logic for each
- On the other, some form of client libraries may want to easily resolve TLV state without the specific use of an `Instruction`, so we want some extra sugar there
- I could be dead wrong on point #1

joncinque:
I think you're right on both points -- while people won't want to loop over their accounts, they'll probably want an easy way to check that each one is the expected address without learning about all the contained types in tlv-account-resolution.

We could even include a `check_addresses` call which takes the raw slices to start with, and then they can extract them in the order expected. This way, they can also totally opt out of doing that and validate each address themselves. What do you think?

buffalojoec:
> I think you're right on both points -- while people won't want to loop over their accounts, they'll probably want an easy way to check that each one is the expected address without learning about all the contained types in tlv-account-resolution.
> 
> We could even include a `check_addresses` call which takes the raw slices to start with, and then they can extract them in the order expected. This way, they can also totally opt out of doing that and validate each address themselves. What do you think?

OK I see where you're coming from. I guess the weird part corresponds to your original point of busting `Instruction` into components. Let me see what I can whip up

buffalojoec:
Perhaps that will do it!

libraries/tlv-account-resolution/src/state.rs
```diff
@@ -105,6 +143,35 @@ impl ExtraAccountMetaList {
...
+            .map(account_meta_from_info)
+            .collect::<Vec<_>>();
+
+        for config in extra_account_metas.iter() {
+            let meta = config.resolve(&provided_metas, instruction_data, program_id)?;
+            provided_metas
+                .iter()
+                .find(|&x| *x == meta)
+                .ok_or(AccountResolutionError::IncorrectAccount)?;
+        }
```
joncinque, line 170: Is it also possible to ensure the order is correct? I think if you pass in `account_info_iter.as_slice()` you should also be able to enumerate and make sure the order is respected. And can you add a test that correctly errors if they're provided out of order?

joncinque (commented):
Just one last little bit, then this should be good to go!

libraries/tlv-account-resolution/src/state.rs
buffalojoec: Sure thing!

joncinque (approved):
Looks great!

PR #5001, opened 2023-08-12
[token-2022] Update confidential transfer instruction to support split proof contexts

samkim-crypto:
#### Summary of changes
- [779eaba](https://github.com/solana-labs/solana-program-library/pull/5001/commits/779eaba30d3cecb91ba2b890eca389555f8667f1): The verify proof helper function for `Transfer` is going to be a bit more involved, so I created a `verify_proof` submodule and refactored the existing `verify_..._proof`()` helper functions into this submodule.
- [7ee73cb](https://github.com/solana-labs/solana-program-library/pull/5001/commits/7ee73cb34e210440541ff63c04f82cc9e0059fea): Modified `instruction.rs` modules to support split proof context states. The most notable change here is the addition of `SplitContextStateAccounts` variant of `ProofLocation` argument type.
- [c26ec15](https://github.com/solana-labs/solana-program-library/pull/5001/commits/c26ec15c93e4c6288e934c427e2c9caf42f1ad75): I updated the `verify_..._proof()` helper functions to take in the iterator of `account_info`'s since functions like `verify_transfer_proof` need to take in variable number of context state accounts.
- [fee378f](https://github.com/solana-labs/solana-program-library/pull/5001/commits/fee378f28427a78b7321447f60e377eafaee7f3f): I refactored ciphertext extraction related helper functions into a separate submodule. The code in the `ciphertext_extraction` module should really belong to the `solana-zk-token-sdk`, so they can be removed on a future Solana upgrade.
- [4de36f4](https://github.com/solana-labs/solana-program-library/pull/5001/commits/4de36f4ed665ccd2b30bb16952c7b523423c9329): This is commit adds the main logic that checks consistency between the split context state proofs. This is unfortunately a bit technical and lower-level in terms of zkp that should really belong to the `solana-zk-token-sdk` (and hence added to `ciphertext_extraction` submodule, which could be removed in the future).
- [f0049ce](https://github.com/solana-labs/solana-program-library/pull/5001/commits/f0049cec0c16723cf2c7b0cfb2b173d82bcd89fc): So this commit relates to a really annoying mistake that I made. The split proof context accounts for equality, ciphertext validity, and range proofs do not contain all the information that is needed to reconstruct a full transfer proof context state due to some optimizations that I mistakenly made. So the "decryption handle" components under the source public key is missing from the three split proof instructions... I added these components directly as part of the transfer instruction data for now, but they should really be removed once this is fixed on the zk token sdk side.
- [78ce6eb](https://github.com/solana-labs/solana-program-library/pull/5001/commits/78ce6eb19d84979e7162a406fc271a12718d6395): Updated the token client to account for the above changes.
- [684ca38](https://github.com/solana-labs/solana-program-library/pull/5001/commits/684ca3859da5fda79a8b1ebf1c97050427de36ca): This is another unfortunately quite technical commit that should really belong to the zk token sdk. This adds logic that generates the individual split proof data from the token account info.
- [c3be8fa](https://github.com/solana-labs/solana-program-library/pull/5001/commits/c3be8faff5cc6f9ff117aee68ccaee49acdd6dd4): Added tests for the split proof
- [655bac5](https://github.com/solana-labs/solana-program-library/pull/5001/commits/655bac546574e6b3be4b26d0d83ba87d5427d678): Took me a while to figure out why the test was keep failing, but I realized that I had the conditional state flipped backwards...
- [11c9c1f](https://github.com/solana-labs/solana-program-library/pull/5001/commits/11c9c1fafb3de967b11e1ce0333ed10d6ef5fb02): Added two additional error variants to suitably describe common error when working with ciphertexts

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -27,6 +27,8 @@ use {
                     ApplyPendingBalanceAccountInfo, EmptyAccountAccountInfo, TransferAccountInfo,
                     WithdrawAccountInfo,
                 },
+                ciphertext_extraction::SourceDecryptHandles,
+                instruction::TransferContextStateAccounts,
                 ConfidentialTransferAccount, DecryptableBalance,
             },
             confidential_transfer_fee::{
@@ -2139,7 +2141,7 @@ where
         source_account: &Pubkey,
         destination_account: &Pubkey,
         source_authority: &Pubkey,
-        context_state_account: Option<&Pubkey>,
+        context_state_accounts: Option<TransferContextStateAccounts<'_>>,
         transfer_amount: u64,
         account_info: Option<TransferAccountInfo>,
         source_elgamal_keypair: &ElGamalKeypair,
@@ -2147,6 +2149,7 @@ where
         destination_elgamal_pubkey: &ElGamalPubkey,
         auditor_elgamal_pubkey: Option<&ElGamalPubkey>,
         signing_keypairs: &S,
+        source_decrypt_handles: Option<&SourceDecryptHandles>,
     ) -> TokenResult<T::Output> {
         let signing_pubkeys = signing_keypairs.pubkeys();
         let multisig_signers = self.get_multisig_signers(source_authority, &signing_pubkeys);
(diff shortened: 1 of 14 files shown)


joncinque (approved):
Thanks for breaking up all the commits and explaining each step. Even the really technical stuff was very digestible!

I added a commit to fix the serde tests, so this is good to go!

samkim-crypto:
Thanks for the help @joncinque!

PR #5011, opened 2023-08-14
add single pool to spl.solana.com

2501babe:
urls are correct after #5010 and https://github.com/solana-labs/security-audits/pull/13 land. also opening an issue for todos which can only be done after crates.io/npm releases

incidentally i dont know how to ensure this deploys after the pr is merged


joncinque (approved):
Looks great!

After this is merged, https://github.com/solana-labs/solana-program-library/blob/master/.github/workflows/spl_action.yml is run, which will build and publish the docs right away to spl.solana.com, so you can check on the CI job that runs on master to make sure that all went well

PR #5019, opened 2023-08-14
ci: Add dalek advisory to ignore list

joncinque:
#### Problem

CI is failing because of the Rust security advisory on ed25519-dalek.

#### Solution

Since the fix requires a major version bump on ed25519-dalek, and SPL picks up these crates from the Solana SDK, we can't just fix it.

Instead, ignore the vulnerability until SPL updates to 1.17.

While I was at it, I noticed the openssl vulernability no longer applied, so I removed it from the list


yihau (approved):
🪖

PR #5031, opened 2023-08-15
stake-pool: Update minimum reserve balance to 0

joncinque:
#### Problem

The `MINIMUM_RESERVE_BALANCE` in the stake pool program is `1`, which is no longer the case after https://github.com/solana-labs/solana/issues/24669 was enabled on mainnet.

#### Solution

Update `MINIMUM_RESERVE_BALANCE` to `0`. There was one check during increase that needed to be changed too from `<=` to `<`, and then a few places in the testing where we added an additional lamport to make stake accounts "valid".


joncinque:
Merging after getting the :+1: from Neodyme, who noticed this

PR #5032, opened 2023-08-15
docs: Clarify stake pool fees more

joncinque:
#### Problem

There can still be some confusion about when stake pool fees are set and whether it's possible to rug people.

#### Solution

Make it clearer!


Benhawkins18:
This is a much clearer explanation. Good to know there are solid controls here.

joncinque:
Thanks for the quick look!

Issue #5015, opened 2023-08-14
finish single pool


2501babe:
various tasks to complete once #5033 is merged:
* grind a cute key (thinking "svsp" leading prefix, i dont have a gpu tho)
* figure out the multisig story for upgrade authority (squads?)
* do whatever bullshit to webify the node code
* publish on cargo/npm
* docs
  - update readme with versions indicating svsp launched on cargo
  - update readme with svsp cli/js
  - update readme with link to latest audit once available
  - update docs to indicate not "upcoming"
  - update docs with cli/js code examples
  - update docs with canonical program address

PR #5033, opened 2023-08-15
single-pool: js (take two)

2501babe:


2501babe:
@joncinque sorry its a bit long. this should be the final nontrivial single-pool pr before v1 launch. the remaining tasks i see ill detail in #5015 

@steveluscher feel free to take a look if you have any suggestions on what would make a web3next conversion easier. i went ahead and did it in web3classic but my intent is that when the webs three can coexist in one codebase via whatever glue code mechanism is intended, id just change the guts to web3next and make the function signatures polymorphic over both types of connection/instruction/transaction. and also replace buffer-layout with something else more web3next-friendly

single-pool/js/src/transactions.ts
```diff
@@ -0,0 +1,302 @@
...
+      transaction.add(SinglePoolInstruction.createTokenMetadata(pool, payer));
+    }
+
+    return transaction;
+  }
+
+  static async deposit(params: DepositParams) {
+    const { connection, pool, userWallet } = params;
+
+    if (!params.userStakeAccount == !params.depositFromDefaultAccount) {
```
joncinque, line 109: I thought this was wrong for a bit, but then I read it some more and realized you're very clever -- nicely done!

single-pool/js/src/transactions.ts
```diff
@@ -0,0 +1,302 @@
...
+    );
+
+    return transaction;
+  }
+
+  static async withdraw(params: WithdrawParams) {
+    const { connection, pool, userWallet, userStakeAccount, tokenAmount, createStakeAccount } =
+      params;
+
+    if (typeof tokenAmount == 'bigint' && tokenAmount > BigInt(Number.MAX_SAFE_INTEGER)) {
```
joncinque, line 175: Shouldn't this be:
```suggestion
    if (typeof tokenAmount != 'bigint' && tokenAmount > BigInt(Number.MAX_SAFE_INTEGER)) {
```

If it's a bigint, it should be safe if it's big, right?

single-pool/js/src/transactions.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    const transaction = new Transaction();
+
+    const poolMintAuthority = findPoolMintAuthorityAddress(this.programId, pool);
+
+    const userStakeAuthority = params.userStakeAuthority || userWallet;
+    const userTokenAccount =
+      params.userTokenAccount ||
+      getAssociatedTokenAddressSync(findPoolMintAddress(this.programId, pool), userWallet);
+    const userTokenAuthority = params.userTokenAuthority || userWallet;
```
joncinque, line 187: I'm surprised to see this! But I guess it makes sense since there's no signers... you could use a throwaway keypair and sign, but then they're screwed if they want to add anything else.  Was that your thinking roughly? Or am I missing something?

single-pool/js/src/transactions.ts
```diff
@@ -0,0 +1,302 @@
...
+    voteAccount: PublicKey,
+    userWallet: PublicKey,
+    stakeAmount: number | bigint,
+  ) {
+    const transaction = new Transaction();
+
+    const pool = findPoolAddress(this.programId, voteAccount);
+    const stakeAccount = await findDefaultDepositAccountAddress(pool, userWallet);
+
+    const seed = 'svsp' + pool.toString().slice(0, 28);
```
joncinque, line 263: nit: How about having this return the seed?

single-pool/js/src/transactions.ts
```diff
@@ -0,0 +1,302 @@
...
+    const stakeRent = await connection.getMinimumBalanceForRentExemption(StakeProgram.space);
+
+    if (
+      typeof stakeAmount == 'bigint' &&
```
joncinque, line 267: Again, shouldn't this be?
```suggestion
      typeof stakeAmount != 'bigint' &&
```

single-pool/js/tests/fixtures/spl_single_validator_pool.so
```diff
@@ -0,0 +1 @@
+../../../../target/deploy/spl_single_validator_pool.so
```
joncinque, line 1: Is it possible to specify another search directory in bankrun?

single-pool/js/tests/transactions.test.ts
```diff
@@ -0,0 +1,385 @@
...
+    connection,
+    pool: poolAddress,
+    userWallet: payer.publicKey,
+    userStakeAccount,
+  });
+  await processTransaction(context, transaction);
+
+  const minimumDelegation = (await connection.getStakeMinimumDelegation()).value;
+  const poolStakeAccount = await client.getAccount(poolStakeAddress);
+  t.true(poolStakeAccount.lamports > minimumDelegation * 2, 'stake has been deposited');
```
joncinque, line 189: up to you, but it might be nice to include the rent exemption and make this a full `===` test

single-pool/js/tests/transactions.test.ts
```diff
@@ -0,0 +1,385 @@
...
+  transaction = await SinglePoolProgram.deposit({
+    connection,
+    pool: poolAddress,
+    userWallet: payer.publicKey,
+    depositFromDefaultAccount: true,
+  });
+  await processTransaction(context, transaction);
+
+  const poolStakeAccount = await client.getAccount(poolStakeAddress);
+  t.true(poolStakeAccount.lamports > minimumDelegation * 2, 'stake has been deposited');
```
joncinque, line 226: Same here

single-pool/js/tests/transactions.test.ts
```diff
@@ -0,0 +1,385 @@
...
+    pool: poolAddress,
+    userWallet: payer.publicKey,
+    userStakeAccount: withdrawAccount.publicKey,
+    tokenAmount: minimumDelegation,
+    createStakeAccount: true,
+  });
+  await processTransaction(context, transaction, [withdrawAccount]);
+
+  const userStakeAccount = await client.getAccount(withdrawAccount.publicKey);
+  t.true(userStakeAccount.lamports > minimumDelegation, 'stake has been withdrawn');
```
joncinque, line 274: And same here

single-pool/js/tests/transactions.test.ts
```diff
@@ -0,0 +1,385 @@
...
+  // initialize pool
+  let transaction = await SinglePoolProgram.initialize(
+    connection,
+    voteAccountAddress,
+    payer.publicKey,
+  );
+  await processTransaction(context, transaction);
+
+  // update metadata
+  const newName = 'hana wuz here';
```
joncinque, line 337: nice :sunglasses:

joncinque (commented):
Really great work overall! Mostly tiny comments, this is really close to ready. I'll let Steve comment on any JS specific bits

single-pool/js/src/transactions.ts
2501babe: no, we have to convert to number when passing to web3 because thats what it supports, so this check is "if its a bigint, error if that conversion would be lossy"

single-pool/js/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
+    "strict": true,
+    "forceConsistentCasingInFileNames": true,
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitReturns": true,
+    "skipLibCheck": true // needed to avoid re-export errors from borsh
```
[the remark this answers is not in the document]
2501babe: and i spent approximately five seconds copying the file verbatim from you :smile:

single-pool/js/src/transactions.ts
2501babe: maybe too clever if its confusing! to me its just a logical xor

single-pool/js/src/transactions.ts
2501babe: sorry im not clear on whats surprising about this, or what you mean by signing with a throwaway keypair. the args here are basically "we assume a normal user will have a wallet address which they use for their token and stake account ownership, but we support people with more niche use-cases who need to specify the pubkeys explicitly." these functions all generate unsigned transactions, the caller is responsible for signing

.github/dependabot.yml
```diff
@@ -65,3 +65,12 @@ updates:
   open-pull-requests-limit: 3
   labels:
     - "automerge"
+- package-ecosystem: pnpm
```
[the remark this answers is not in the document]
2501babe: oops, i had only found this which i interpreted to imply a separate designation. ty! https://github.blog/changelog/2023-06-12-dependabot-version-updates-now-supports-pnpm/

single-pool/js/tests/fixtures/spl_single_validator_pool.so
2501babe: i can set an environment var but doing it in js seemed hacky and doing it in the github script would break testing locally

single-pool/js/src/transactions.ts
joncinque: `tokenAmount` is passed to your own instruction, which supports `bigint`, and to `createApproveInstruction`, which also supports `bigint`, so there's no raw web3.js stuff to worry about from what I can see. Is there something else that I'm missing?

single-pool/js/tsconfig.json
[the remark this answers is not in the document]
joncinque: Ok great, I'm glad that you didn't waste any time on it!

single-pool/js/src/transactions.ts
joncinque: Nah, I'm just a bit dense when it comes to xor checks

single-pool/js/src/transactions.ts
joncinque: Oh nvm, this is even more clever than I thought. I know you're a big fan of delegates, so I was surprised to see you default an approval to the wallet itself. I thought this would look just like a normal transfer signed by the wallet then, which would defeat the purpose of the delegate.

But if you approve to yourself, you still limit the amount of damage! E.g. your account has 1000 SOL, you approve 100 to yourself, so when you call into another program, the most that it can take is 100 SOL.

We should talk up this flow to wallets more, because it's very clever!

single-pool/js/tests/fixtures/spl_single_validator_pool.so
joncinque: Ah gotcha, a link makes sense then

single-pool/js/src/transactions.ts
2501babe: oh yes youre right, spl-token supports bigint. web3.js doesnt, and i had made this "fix" to `withdraw` after discovering i needed to convert to number in `createAndDelegateUserStake`

single-pool/js/src/transactions.ts
joncinque: Per offline discussion, I misunderstood what this is doing. This is approving to the pool's authority, and not to the wallet. So this is perfect as is. Resolving!

single-pool/js/src/instructions.ts
```diff
@@ -0,0 +1,206 @@
...
+import { MPL_METADATA_PROGRAM_ID, findMplMetadataAddress } from './mpl_metadata';
+import {
+  findPoolAddress,
+  findPoolStakeAddress,
+  findPoolMintAddress,
+  findPoolStakeAuthorityAddress,
+  findPoolMintAuthorityAddress,
+  findPoolMplAuthorityAddress,
+} from './addresses';
+import { SinglePoolProgram } from './transactions';
```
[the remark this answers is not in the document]
2501babe: going to put in layouts because that file isnt reexported at the toplevel

single-pool/js/src/transactions.ts
2501babe: why? the user can get the address via `findDefaultDepositAccountAddress`. if theres a use for them having the seed, id return it from that instead tho

single-pool/js/tests/transactions.test.ts
2501babe: i thought it would be potentially brittle because the amount depends on whether the pool is activating or active at time of deposit, but i think bankrun lets me jump epochs

joncinque: that's fair, it's up to you!

2501babe:
@joncinque presume the audit fail will be fixed by a rebase but ill wait for you to inspect the latest commit if youd like

steveluscher:
Uh… actually it looks like you might be async already, everywhere there's a key operation.

steveluscher:
I wonder if @lorisleiva would like to work with you to give Kinobi a shot, to see if it could straight up auto-generate the client based on your program?

lorisleiva:
I would love to help with that! ☺️

2501babe:
well forcing people to mark certain sync things async in an upgrade isnt that huge an ask. but you will need to have the ability to include both web3 libraries in a single project, plus conversion functions between (at minimum) `Transaction`, `TransactionInstruction`, `Connection`... the issue i realized while writing a web3next version (https://github.com/2501babe/solana-program-library/blob/578b624d9802684dd9e1e9b3e34f1e8334c0481e/stake-pool/single-pool-nextjs/src/index.ts) is that i _cant_ be on web3next because then it would be impossible for any existing wallet or dapp to use the code. and conversely, early adopters of web3next will need a way to use web3classic libraries, otherwise onboarding will be impossible. i talked a bit about this in discord with joe but figured you had a plan for this

lorisleiva:
I think a Umi-compatible library (generated by Kinobi) could help with that transition.

Umi provides a set of abstract interfaces that both libraries and end-users adhere to. Then, it's up to the end-user to decide which implementations they want to use for these interfaces via plugins. Umi also provide "bundle" plugins and more specifically a "default bundle" plugin that can help end-users get started very quickly. 

One of these interfaces is the `RpcInterface` that enables users and libraries to query the RPC and send/confirm transaction. The "default bundle" currently includes the `Web3JsRpc` implementation which uses the legacy `web3.js` library under the hood. We could offer a new RPC implementation that uses the new experimental `web3.js` and users could switch to it without even realising it.

I'm also more than happy to tweak and upgrade these interfaces if needed to make that transition smoother.

steveluscher:
> well forcing people to mark certain sync things async in an upgrade isnt that huge an ask.

Super depends on how they've built their app. Changing from `Keypair.generate()` to `await generateKeyPair()` is fine if the outer method is already `async`, but if it's not then that method and _every method that calls it_ has to be made async. I've seen apps in which that refactor is impossible given all the time and engineers in the universe.

So, first of all I totally get two things:

* this needs to work with existing wallets, today
* you probably want to get this out fast, _more_ than you want this to be used as a workshop for Kinobi/web3next

Even without getting into Kinobi codegen, I think that we can have it all here, with some architectural sleight of hand.

Let me use, as an example, the `getVoteAccountAddressForPool` function.

```ts
export async function getVoteAccountAddressForPool(connection: Connection, poolAddress: PublicKey) {
  const poolAccount = await connection.getAccountInfo(poolAddress);
  if (!(poolAccount && poolAccount.data[0] === 1)) {
    throw 'invalid pool address';
  }

  return new PublicKey(poolAccount.data.slice(1));
}
```

This function falls into the classic trap of [overspecifiying its interface](https://twitter.com/steveluscher/status/1555460016361992192). `Connection` contains a _shitload_ of stuff, from two network clients, to `HTTPAgents`, to the implementation of 100 RPC methods, a full state machine for subscriptions, and much much more. Literally all this function cares about is:

* getting `AccountInfo` for an address
* returning an address from some account data

If I were to strip this function down to its raw, hot core, it would look like this:

```ts
// Let's presume this is released on npm as a module called `single-pool-core`

type Config<TAddress> = Readonly<{
  createAddressFromBytes(bytes: Uint8Array): Promise<TAddress>;
  getAccountInfo(address: TAddress): Promise<AccountInfo>;
}>;

export async function getVoteAccountAddressForPoolBase<TAddress>(
  poolAddress: TAddress,
  { createAddressFromBytes, getAccountInfo }: Config<TAddress>
) {
  const poolAccount = await getAccountInfo(poolAddress);
  if (!(poolAccount && poolAccount.data[0] === 1)) {
    throw "invalid pool address";
  }
  return await createAddressFromBytes(poolAccount.data.slice(1));
}
```

This core function has nfc what a `PublicKey` is or what an RPC/Connection is and it doesn't gaf.

Now you make two outer wrappers and release them as npm packages.

```ts
/** `single-pool` on npm */
export async function getVoteAccountAddressForPool(
  rpc: Rpc<GetAccountInfoApi>,
  poolAddress: Base58EncodedAddress,
  abortSignal?: AbortSignal
) {
  return await getVoteAccountAddressForPoolBase(poolAddress, {
    createAddressFromBytes: (bytes) =>
      getBase58EncodedAddressCodec().deserialize(bytes),
    getAccountInfo: (address) => rpc.getAccountInfo(address).send(abortSignal),
  });
}

/** `single-pool/legacy` on npm */
export async function getVoteAccountAddressForPool(
  connection: Connection,
  poolAddress: PublicKey
) {
  return await getVoteAccountAddressForPoolBase(poolAddress, {
    createAddressFromBytes: (bytes) => new PublicKey(bytes),
    getAccountInfo: (address) => connection.getAccountInfo(address),
  });
}
```

Pros about this:

* web3.js is completely gone from your core ‘base’ program
* that `rpc: Rpc<GetAccountInfoApi>` typehint says ‘I just need an RPC that supports at least this _one_ method’
* devs can import `single-pool` or `single-pool/legacy` and get only the code they need and not a bit more.

You can do this _everywhere_. For instance in `SinglePoolProgram#initialize` you don't _actually_ need a `Connection`; you just need getters for the rent exemption and stake delegation minimums, and then some transaction/instruction builders.

There's a _second_ approach you could take. You could make web3next the de-facto base implementation. Here's how you would do that.

```ts
/** `single-pool` on npm */
export async function getVoteAccountAddressForPool(
  rpc: Rpc<GetAccountInfoApi>,
  poolAddress: Base58EncodedAddress,
  abortSignal?: AbortSignal
) {
  const poolAccount = await rpc.getAccountInfo(poolAddress).send(abortSignal);
  if (!(poolAccount && poolAccount.data[0] === 1)) {
    throw "invalid pool address";
  }
  return getBase58EncodedAddressCodec().deserialize(poolAccount.data.slice(1));
}

/** `single-pool/legacy` on npm */
export async function getVoteAccountAddressForPoolLegacy(
  connection: Connection,
  poolAddress: PublicKey
) {
  const voteAccountAddress = await getVoteAccountAddressForPool(
    {
      getAccountInfo(address: Base58EncodedAddress) {
        return {
          async send() {
            const poolPublicKey = new PublicKey(address);
            return await connection.getAccountInfo(poolPublicKey);
          },
        };
      },
    },
    poolAddress.toBase58() as Base58EncodedAddress
  );
  return new PublicKey(voteAccountAddress);
}
```

Your legacy wrapper just commutes between the legacy/modern inputs and outputs.

2501babe:
after more conversatoin with luscher im going to merge this pr as-is, then attempt approach 2 as a separate pr. if it proves infeasible or i need to refocus on other priorities then this is serviceable as-is

as of now the single pool program, cli, and js client are ready to ship. so its just improvements and logistics to finish for a v1 launch

PR #5036, opened 2023-08-15
Token Group Interface Spec

buffalojoec:
The following is a spec for a "Token Group Interface" - which serves to outline an interface for grouping tokens together.

This interface is designed to be generic over whatever type of asset (represented by tokens) you might be implementing an interface over.

Two common examples are demonstrated in this PR/spec:

- Token Collections: Collections of tokens where one token identifies the collection (or group), and all other tokens are simply "members" of the collection
- Token Editions: Token "editions" where one token serves as the original, and all others are "reprints"

> Note: These are implementations of the interface specific only to this example, and interfaces that follow the guise of "collections" or "editions" can be implemented in a variety of ways.

For more information about the nuances of this interface, see the `README.md` of each crate!

---

Marking as a draft for now until I can circle back and work on this some more - but anyone is free to take a look in the meantime!

(It will not pass CI yet)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6697,6 +6697,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "spl-interface-base"
+version = "0.1.0"
+dependencies = [
+ "borsh 0.10.3",
+ "solana-program",
+ "spl-discriminator",
+ "spl-tlv-account-resolution",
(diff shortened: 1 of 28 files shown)


buffalojoec:
Closing for #5190 

lfg

PR #5043, opened 2023-08-16
docs: Specify that the stake pool staker can update the staker

joncinque:
#### Problem

In the "roles" section of the stake pool docs, it doesn't specify that the staker can update the staker.

#### Solution

Specify it!


joncinque:
Merging after getting the 👍 from Neodyme, who noticed this

PR #5050, opened 2023-08-17
[token-metadata] Add serde support for token metadata instructions

serbangv:
Issue #3325

Adds serde support for instructions from `token-metadata`. Continuation of #4772

Notes:
* Added `serde-traits` feature flag to `token-metadata`, all serde code is behind this flag
* Noticed `OptionalNonZeroPubkey` is virtually identical to the type with the same name from `token-2022` so I basically copy/pasted the de/serialization logic from there. Let me know if a different approach is preferred.

@joncinque @buffalojoec

### DIFF
--- a/token-metadata/interface/src/instruction.rs
+++ b/token-metadata/interface/src/instruction.rs
@@ -16,6 +16,7 @@ use serde::{Deserialize, Serialize};
 
 /// Initialization instruction data
 #[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
 #[derive(Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize, SplDiscriminate)]
 #[discriminator_hash_input("spl_token_metadata_interface:initialize_account")]
 pub struct Initialize {
@@ -29,6 +30,7 @@ pub struct Initialize {
 
 /// Update field instruction data
 #[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
 #[derive(Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize, SplDiscriminate)]
 #[discriminator_hash_input("spl_token_metadata_interface:updating_field")]
 pub struct UpdateField {
@@ -40,6 +42,7 @@ pub struct UpdateField {
 
 /// Remove key instruction data
 #[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
 #[derive(Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize, SplDiscriminate)]
 #[discriminator_hash_input("spl_token_metadata_interface:remove_key_ix")]
 pub struct RemoveKey {
@@ -53,6 +56,7 @@ pub struct RemoveKey {
 /// Update authority instruction data
 #[derive(Clone, Debug, PartialEq, BorshSerialize, BorshDeserialize, SplDiscriminate)]
 #[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
 #[discriminator_hash_input("spl_token_metadata_interface:update_the_authority")]
 pub struct UpdateAuthority {
     /// New authority for the token metadata, or unset if `None`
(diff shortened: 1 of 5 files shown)


buffalojoec (commented):
lgtm! I'll leave the stamp for @joncinque

token-metadata/interface/src/instruction.rs
```diff
@@ -384,4 +396,104 @@ mod test {
...
+
+    #[cfg(feature = "serde-traits")]
+    #[test]
+    fn update_authority_serde_with_none() {
+        let data = UpdateAuthority {
+            new_authority: OptionalNonZeroPubkey::default(),
+        };
+        let ix = TokenMetadataInstruction::UpdateAuthority(data);
+        let serialized = serde_json::to_string(&ix).unwrap();
+        let serialized_expected = "{\"UpdateAuthority\":{\"new_authority\":null}}";
```
joncinque, line 477: I'm just now realizing that we did these a bit wrong... since it's meant to be consumed by JSON, we should really use JS idioms and do everything `camelCase`. We'll fix token-2022 separately, but for this one, can you make everything camelCase? Ie. add another line like:

```
#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
```

More info at https://serde.rs/attr-rename.html

joncinque (commented):
Looks really great! Just the bit about camelCasing.

Also, can you add a step in CI to run the tests with `serde-traits` enabled? You can add it right before this step https://github.com/solana-labs/solana-program-library/blob/30d0e1b654414b5a2aca6c6d3a90cd66c820c31e/.github/workflows/pull-request-token-metadata.yml#L61, and it would look similar to https://github.com/solana-labs/solana-program-library/blob/30d0e1b654414b5a2aca6c6d3a90cd66c820c31e/.github/workflows/pull-request-token.yml#L96

token-metadata/interface/src/instruction.rs
serbangv: Sure thing.

The attributes we're adding for `serde` are quite a few now. I was thinking maybe it's worth investing time in creating a `proc_macro_attribute` that will automatically add the appropriate attributes depending on the type.

So we would go from this:

```
#[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
#[derive(Clone, Copy, Debug, PartialEq)]
#[repr(u8)]
pub enum TransferFeeInstruction {
    InitializeTransferFeeConfig {
        #[cfg_attr(feature = "serde-traits", serde(with = "coption_fromstr"))]
        transfer_fee_config_authority: COption<Pubkey>,
        #[cfg_attr(feature = "serde-traits", serde(with = "coption_fromstr"))]
        withdraw_withheld_authority: COption<Pubkey>,
        transfer_fee_basis_points: u16,
        maximum_fee: u64,
    },
...
}

#[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
#[cfg_attr(feature = "serde-traits", serde(from = "u64", into = "u64"))]
#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
#[repr(transparent)]
pub struct PodU64([u8; 8]);

```

To something like this:

```
#[cfg_attr(feature = "serde-traits", spl_serde)]
#[derive(Clone, Copy, Debug, PartialEq)]
#[repr(u8)]
pub enum TransferFeeInstruction {
    InitializeTransferFeeConfig {
        transfer_fee_config_authority: COption<Pubkey>,
        withdraw_withheld_authority: COption<Pubkey>,
        transfer_fee_basis_points: u16,
        maximum_fee: u64,
    },
...
}

#[cfg_attr(feature = "serde-traits", spl_serde)]
#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
#[repr(transparent)]
pub struct PodU64([u8; 8]);
```

Since `serde` logic is spread out in a few crates (`program-2022` and `token-metadata` now) this `proc_macro_attribute` would reside in `/libraries` so it can be used in several crates.

Do you think this is worth it? I could implement it.

joncinque: I think that makes sense. Maybe you can work with @buffalojoec to create a little crate with the macro? For this PR, let's stick with the two lines

serbangv: Done. 

Will sync with @buffalojoec regarding the proc_macro_attribute.

joncinque (approved):
Beautiful! I'll let @buffalojoec give the final approval and merge it in. And please rebase to fix the unrelated audit failure

serbangv:
Audit check passes now.

buffalojoec:
> Beautiful! I'll let @buffalojoec give the final approval and merge it in. And please rebase to fix the unrelated audit failure

Ping-pong!

Yeah, lgtm @serbangv thank you mate!

As far as the macro stuff goes, I think the proc macro you suggested makes sense, and I'm happy to review if you need some eyes on it. I like the concept of kinda wrapping all of those field-level attributes and the special treatment of `COption` into the proc macro.

Feel free to throw a PR together!

buffalojoec (approved):

PR #5052, opened 2023-08-18
single pool: minor cleanups

2501babe:
when implementing `create_and_delegate_user_stake` in js i realized providing both vote and pool was weird (a result of hastiliy adding the vote param when i realized i needed it), so i normalized it to have the same interface as all the other functions

also neodyme pointed out `find_address_and_bump` is clever for no reason, and yea, i forget why i did this originally


joncinque (approved):
Makes sense~

PR #5054, opened 2023-08-20
[token-2022] Confidential transfer parallel split proofs

samkim-crypto:
#### Summary of changes
- [838e6be](https://github.com/solana-labs/solana-program-library/pull/5054/commits/838e6becd52a67634233420804774ed8950f167b): I removed the split proofs from the `Transfer` instruction and added a separate `TransferWithSplitProofs` that support transfer with split proof contexts.
- [cda6078](https://github.com/solana-labs/solana-program-library/pull/5054/commits/cda6078fd81e8db5d5331efa196650868e1b6137): Updated the main program logic for split proof contexts to support no-op if the account is not initialized and also auto close on full transfer execution. Since CPI into the proof program is not currently supported, I currently disabled auto close on transfer execution with an invalid instruction data error. The notable change in this PR is to the result type of `verify_transfer_proof` in `verify_proof.rs`. I made it so that the function returns an `Option` that is wrapped inside another `Result`. I considered adding a custom type for the return type, but I thought this could make things even more complicated.
- [98b171e](https://github.com/solana-labs/solana-program-library/pull/5054/commits/98b171e9f7740451f7ce581fa8514c4b31644a23): I updated the token client to support the `TransferWithSplitProofs` instruction. This PR adds the following functions:
  - `confidential_transfer_create_equality_and_ciphertext_validity_proofs_for_transfer`: This function creates context state accounts for the equality and ciphertext validity proofs that are needed for a transfer
  - `confidential_transfer_create_range_proof_for_transfer`: This function creates context state account for the range proof that is needed for a transfer
  - `confidential_transfer_transfer_with_split_proofs`: This submits the `TransferWithSplitProofs` instruction assuming that all the context states have been previously created.
  - `confidential_transfer_close_context_state`: This function closes a context state account
  - `confidential_transfer_transfer_with_split_proofs_in_parallel`: This function submits two transactions in parallel. The first transaction consist of instructions needed to create the relevant equality and ciphertext validity proof with the transfer instruction attached. The second transaction consists of instructions to needed to create range proof also with the transfer instruction attached. Whichever instruction lands later will fully execute the transfer.
- [8589746](https://github.com/solana-labs/solana-program-library/pull/5054/commits/85897462dcdd650114544cc562abb23a4231a698): Updated tests for the `TransferWithSplitProofs` instruction.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -2249,7 +2249,7 @@ where
                     new_decryptable_available_balance.into(),
                     source_authority,
                     context_state_accounts,
-                    &source_decrypt_handles,
+                    source_decrypt_handles,
                 )?,
             ],
             &[source_authority_keypair],
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -1,6 +1,6 @@
 use {
     crate::client::{ProgramClient, ProgramClientError, SendTransaction, SimulateTransaction},
-    futures::future::join_all,
+    futures::{future::join_all, try_join},
     futures_util::TryFutureExt,
     solana_program_test::tokio::time,
     solana_sdk::{
@@ -28,7 +28,7 @@ use {
                     WithdrawAccountInfo,
                 },
                 ciphertext_extraction::SourceDecryptHandles,
-                instruction::TransferContextStateAccounts,
+                instruction::TransferSplitContextStateAccounts,
                 ConfidentialTransferAccount, DecryptableBalance,
             },
             confidential_transfer_fee::{
@@ -46,13 +46,18 @@ use {
                 auth_encryption::AeKey,
                 elgamal::{ElGamalCiphertext, ElGamalKeypair, ElGamalPubkey, ElGamalSecretKey},
             },
+            instruction::*,
             zk_token_elgamal::pod::ElGamalPubkey as PodElGamalPubkey,
+            zk_token_proof_instruction::{self, ContextStateInfo, ProofInstruction},
+            zk_token_proof_program,
+            zk_token_proof_state::ProofContextState,
         },
         state::{Account, AccountState, Mint, Multisig},
     },
     spl_token_metadata_interface::state::{Field, TokenMetadata},
     std::{
         fmt, io,
+        mem::size_of,
         sync::{Arc, RwLock},
         time::{Duration, Instant},
     },
@@ -2141,7 +2146,7 @@ where
         source_account: &Pubkey,
         destination_account: &Pubkey,
         source_authority: &Pubkey,
-        context_state_accounts: Option<TransferContextStateAccounts<'_>>,
+        context_state_account: Option<&Pubkey>,
         transfer_amount: u64,
         account_info: Option<TransferAccountInfo>,
         source_elgamal_keypair: &ElGamalKeypair,
@@ -2149,7 +2154,6 @@ where
         destination_elgamal_pubkey: &ElGamalPubkey,
         auditor_elgamal_pubkey: Option<&ElGamalPubkey>,
         signing_keypairs: &S,
-        source_decrypt_handles: Option<&SourceDecryptHandles>,
     ) -> TokenResult<T::Output> {
         let signing_pubkeys = signing_keypairs.pubkeys();
         let multisig_signers = self.get_multisig_signers(source_authority, &signing_pubkeys);
@@ -2163,7 +2167,7 @@ where
             TransferAccountInfo::new(confidential_transfer_account)
         };
 
-        let proof_data = if context_state_accounts.is_some() {
+        let proof_data = if context_state_account.is_some() {
             None
         } else {
             Some(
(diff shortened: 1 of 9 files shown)


token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -1279,3 +1273,82 @@ pub fn disable_non_confidential_credits(
...
 }
+
+/// Create a `TransferWithSplitProof` instruction without fee
+#[allow(clippy::too_many_arguments)]
+#[cfg(not(target_os = "solana"))]
+pub fn transfer_with_split_proofs(
+    token_program_id: &Pubkey,
+    source_token_account: &Pubkey,
+    destination_token_account: &Pubkey,
+    mint: &Pubkey,
```
joncinque, line 1284: nit: I'm just noticing this now, so you don't need to fix it in this PR, but all of the confidential transfer functions order these as "source, destination, mint", whereas the normal transfer functions order it as "source, mint, destination". We should make that consistent at some point

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -454,55 +457,61 @@ fn process_transfer(
...
+        let maybe_proof_context = verify_transfer_proof(
             account_info_iter,
             proof_instruction_offset,
             split_proof_context_state_accounts,
+            no_op_on_uninitialized_split_context_state,
+            close_split_context_state_on_execution,
             source_decrypt_handles,
         )?;
+        // If `maybe_proof_context` is `None`, then this means that
+        // `close_split_context_state_on_execution` is true and a required context state account is
```
joncinque, line 469: This should be `no_op_on_uninitialized_split_context_state`, right?

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -454,55 +457,61 @@ fn process_transfer(
...
-            &proof_context.transfer_pubkeys.destination,
-            &destination_ciphertext_lo,
-            &destination_ciphertext_hi,
-            None,
+            maybe_proof_context.as_ref(),
         )?;
+
+        if maybe_proof_context.is_none() {
+            msg!("Context states not fully initialized: return with no op");
+            msg!("WARNING: transfer is NOT yet executed");
```
joncinque, line 513: nit: Maybe "WARNING" is a bit strong, since it is expected operation. Let's just say "transfer is NOT yet executed" and combine it with the earlier `msg`

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -648,7 +782,7 @@ fn process_source_for_transfer(
         source_transfer_amount_lo,
         source_transfer_amount_hi,
     )
-    .ok_or(TokenError::CiphertextArithmeticFailed)?;
+    .ok_or(ProgramError::InvalidInstructionData)?;
```
joncinque, line 785: Did you mean to revert all of these errors? Unless there's a good reason, let's stick with the more specific ones

token/client/src/token.rs
```diff
@@ -2212,13 +2204,308 @@ where
...
+            transfer_with_range_proof
+        )
+    }
+
+    /// Create equality and ciphertext validity proofs for a confidential transfer.
+    ///
+    /// If an optional transfer instruction is provided, then the transfer instruction is attached
+    /// to the same transaction.
+    #[allow(clippy::too_many_arguments)]
+    pub async fn confidential_transfer_create_equality_and_ciphertext_validity_proofs_for_transfer<
```
joncinque, line 2351: nit: instead of `proofs` can you say `context_state` since that's what's being done? And rather than overloading this public function to optionally add a transfer, how about have three functions:

* `confidential_transfer_create_equality_and_ciphertext_validity_context_states_for_transfer` -- just the creation of the context state accounts
* `confidential_transfer_create_equality_and_ciphertext_validity_context_states_and_transfer_parallel` -- specifically for parallel execution, requires the transfer instruction
* some internal helper function that they both call into to create the proofs, up to you for naming!

token/client/src/token.rs
```diff
@@ -2212,13 +2204,308 @@ where
...
+        }
+
+        self.process_ixs(&instructions, &signers).await
+    }
+
+    /// Create a range proof context state account for a confidential transfer.
+    ///
+    /// If an optional transfer instruction is provided, then the transfer instruction is attached
+    /// to the same transaction.
+    pub async fn confidential_transfer_create_range_proof_for_transfer<S: Signer>(
```
joncinque, line 2440: Same thing here, can we split it into three functions?

joncinque (approved):
Looks really good! Only some minor comments, R+ after that

token/program-2022/src/extension/confidential_transfer/instruction.rs
samkim-crypto: Yep, thanks for the catch. I'll reorder the inputs in the confidential transfer functions in a follow-up PR.

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: Thanks for the catch! Fixed.

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: You are right. I removed the warning and merged the "transfer is NOT yet executed" to the previous message.

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: This must have been introduced when I was refactoring. I reverted these changes.

token/client/src/token.rs
samkim-crypto: Yes, that is a great idea!

PR #5083, opened 2023-08-22
ci: Ignore webpki RUSTSEC issue

joncinque:
#### Problem

There's a RUSTEC advisory on webpki, which is used by upstream dependenices like tokio, but there's no upgrade path.

#### Solution

We'll have to ignore it for now unfortunately.


joncinque:
Merging without review to unblock other PRs

joncinque:
No worries at all!

PR #5088, opened 2023-08-23
tlv-account-resolution: Support PDAs from other program ids

joncinque:
#### Problem

The seeds in tlv-account-resolution are really useful for getting dynamic accounts into an instruction, but they only derive PDAs for the executing program, and not for others. This limits composability for devs who want to build a sample transfer-hook program on top of another program.

#### Solution

When resolving PDAs, if the top bit of the discriminator is set (it's  >= 128), that means the rest of the bits encode the index of an account in the instruction.

While I was at it, in a separate commit, I used some seeds in the examples / README

cc @dankelleher

### DIFF
--- a/libraries/tlv-account-resolution/README.md
+++ b/libraries/tlv-account-resolution/README.md
@@ -12,7 +12,11 @@ into a TLV entry in an account, you can do the following:
 use {
     solana_program::{account_info::AccountInfo, instruction::{AccountMeta, Instruction}, pubkey::Pubkey},
     spl_discriminator::{ArrayDiscriminator, SplDiscriminate},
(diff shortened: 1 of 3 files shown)


buffalojoec (approved):
Simple and clean, I like it!

Only negative here is that the program ID _has_ to be included in the accounts list, and it may not be needed for the program. But unless we decide to increase the size of the `address_config`, this is the best approach

joncinque:
> Only negative here is that the program ID has to be included in the accounts list, and it may not be needed for the program. But unless we decide to increase the size of the address_config, this is the best approach

That's a good point... I think the idea is that you would also CPI into that program to do any other checks / logic, so hopefully it won't always be too bad!

PR #5093, opened 2023-08-23
token-2022: enable camelCase for serde

serbangv:
Issue #3325

Addresses this comment: https://github.com/solana-labs/solana-program-library/pull/5050#discussion_r1300031644

I still plan on implementing a macro to replace a lot of these serde attributes, but until that is done it feels like the camelCase should not be postponed too much so I added it in this PR.

@joncinque @buffalojoec


joncinque (approved):
Looks good to me! We'll need to mention this as a breaking change for the next version of token-2022.

PR #5094, opened 2023-08-23
token-cli: Integrate token-metadata in the mint

joncinque:
#### Problem

The metadata extension in token-2022 is great, but currently unusable except in the token client.

#### Solution

Support initializing and updating metadata!

Unfortunately, we can't also initialize the metadata during that first call to `create-token` because the mint authority doesn't sign that command, and it's more confusing than anything else to have it take a maybe-keypair and then maybe-do-it. The solution there was to explain the next steps.

Otherwise, the rest should be straightforward!

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -5419,7 +5419,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore]
     #[serial]
     async fn create_token_default() {
         let (test_validator, payer) = new_validator_for_test().await;
@@ -5440,7 +5439,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore]
     #[serial]
     async fn create_token_interest_bearing() {
         let (test_validator, payer) = new_validator_for_test().await;
@@ -5474,7 +5472,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore]
     #[serial]
     async fn set_interest_rate() {
         let (test_validator, payer) = new_validator_for_test().await;
@@ -5512,7 +5509,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore]
     #[serial]
     async fn supply() {
         let (test_validator, payer) = new_validator_for_test().await;
@@ -5532,7 +5528,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore]
     #[serial]
     async fn create_account_default() {
         let (test_validator, payer) = new_validator_for_test().await;
@@ -5554,7 +5549,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore]
     #[serial]
     async fn account_info() {
         let (test_validator, payer) = new_validator_for_test().await;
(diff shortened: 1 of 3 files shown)


token/cli/src/main.rs
```diff
@@ -2916,6 +3028,97 @@ fn app<'a, 'b>(
...
+                        .alias("owner")
+                        .value_name("KEYPAIR")
+                        .validator(is_valid_signer)
+                        .takes_value(true)
+                        .help(
+                            "Specify the mint authority keypair. \
+                             This may be a keypair file or the ASK keyword. \
+                             Defaults to the client keypair."
+                        ),
+                )
```
buffalojoec, line 3076: re: my other comment: update authority is missing from this list of args

token/cli/src/main.rs
```diff
@@ -5171,6 +5419,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore]
```
buffalojoec, line 5422: Did you mean to leave these `ignore`s in here?

token/cli/src/main.rs
```diff
@@ -7989,4 +8286,146 @@ mod tests {
...
+
+    #[tokio::test]
+    #[serial]
+    async fn metadata() {
+        let (test_validator, payer) = new_validator_for_test().await;
+        let program_id = spl_token_2022::id();
+        let config = test_config_with_default_signer(&test_validator, &payer, &program_id);
+        let name = "this";
+        let symbol = "is";
+        let uri = "METADATA!";
```
buffalojoec, line 8298: 😂 ⚔️

buffalojoec (commented):
Looking nice! Just left a few comments

token/cli/src/main.rs
joncinque: Ah yeah, good point. Sorry, I was doing this by default during initialize before, and then it didn't work, so I created this other instruction. I'll definitely add it in

token/cli/src/main.rs
joncinque: Oops! Removing them

token/cli/src/main.rs
```diff
@@ -698,6 +720,88 @@ async fn command_set_transfer_hook_program(
...
+    uri: String,
+    bulk_signers: Vec<Arc<dyn Signer>>,
+) -> CommandResult {
+    let token = token_client_from_config(config, &token_pubkey, None)?;
+
+    let res = token
+        .token_metadata_initialize_with_rent_transfer(
+            &config.fee_payer()?.pubkey(),
+            &mint_authority,
+            &mint_authority,
```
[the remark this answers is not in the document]
joncinque: yeah good point

joncinque:
Should be good for another look

buffalojoec:
> Should be good for another look

lgtm I think you just need to add that annotation to quiet clippy

joncinque:
Bah sorry, I went a little too quickly on this one

buffalojoec (approved):

PR #5095, opened 2023-08-23
docs: Explain the token-2022 metadata and pointer extensions

joncinque:
#### Problem

The metadata and metadata-pointer extensions have been in token-2022 for some time, but there's no way to know.

#### Solution

Add some text and examples about how to use them. This uses the flows from #5094.


buffalojoec (commented):
This lgtm just waiting to see your thoughts on #5094

joncinque:
I rebased and updated the "initialize" command to specify the update authority

buffalojoec (approved):
ship it!

PR #5096, opened 2023-08-24
token-js: Added support for transfer hook instructions

wjthieme:
resolves: #4337

PR #5102, opened 2023-08-24
token-js: added the missing AuthorityTypes and tests

### LINKED ISSUE #5100: token-js: Support all authority types for set-authority
#### Problem

There's a huge disparity between the supported Rust authorities https://github.com/solana-labs/solana-program-library/blob/47f09c903f567bfee1e9d5a1e9ab83770208449e/token/program-2022/src/instruction.rs#L1042-L1070

and the JS supported authorities https://github.com/solana-labs/solana-program-library/blob/47f09c903f567bfee1e9d5a1e9ab83770208449e/token/js/src/instructions/setAuthority.ts#L16-L21

#### Solution

Add all of the authorities to JS too!

wjthieme:
resolves #5100

The following tests are still missing since the js-client does not support initializing these extensions yet
* `ConfidentialTransferMint`
* `ConfidentialTransferFeeConfig`
* `MetadataPointer`

~~Will add a test for `TransferHookProgramId` once #5096 is merged~~


joncinque:
Thanks!

PR #5112, opened 2023-08-25
token-js: added extra account resolution for transfer hook extension

### LINKED ISSUE #5108: token-js: Fetch extra account metas for transfer-hooks on transfer
#### Problem

The transfer hook system requires fetching additional accounts, and the Rust libraries expose ways to do this off-chain and on-chain. Currently, the JS library has no idea how to fetch any additional required accounts for the transfer hook.

#### Solution

Implement something similar to the transfer-hook-interface offchain helpers to fetch additional accounts. This takes a few levels.

First, we need to fetch the mint, and if a transfer hook is required, the additional validation account: https://github.com/solana-labs/solana-program-library/blob/678b1e562e8484a7a08042fba794ca3bbe863f92/token/transfer-hook-interface/src/offchain.rs#L43

Then we need to read the data contained in the `ExtraAccountMetasList` to add those extra account metas into the instruction: https://github.com/solana-labs/solana-program-library/blob/678b1e562e8484a7a08042fba794ca3bbe863f92/libraries/tlv-account-resolution/src/state.rs#L205

And finally, for each element in the list, we have to `resolve` it properly: https://github.com/solana-labs/solana-program-library/blob/678b1e562e8484a7a08042fba794ca3bbe863f92/libraries/tlv-account-resolution/src/account.rs#L122

wjthieme:
resolves #5108

Couple notes:
* Not sure exactly how multisigners is implemented for extra account resolution in the program since it changes the ordering of the account keys. Does it mean that if a user wants multisigners to work they explicitly need to code this into their transfer hook program?
* What is done by `tlv-account-resolution` in rust is now baked into `@solana/spl-token` for js. Might be a nice improvement to separate that into its own package?
* I have written a few tests to the best of my understanding of what exactly the data structure of the extra accounts data account is. Might be worth for @joncinque or @buffalojoec to check if my implementation actually matches that of the rust client.
* The way `transferCheckedWithHook` and `addExtraAccountsToInstruction` are structured is a little bit made up by me. Let me know if it matches your expectations @joncinque

### DIFF
--- a/token/js/src/errors.ts
+++ b/token/js/src/errors.ts
@@ -70,9 +70,9 @@ export class TokenUnsupportedInstructionError extends TokenError {
     name = 'TokenUnsupportedInstructionError';
 }
 
-/** Thrown if the transfer hook extra accounts contains an invalid discriminator */
-export class TokenTransferHookInvalidDiscriminator extends TokenError {
-    name = 'TokenTransferHookInvalidDiscriminator';
+/** Thrown if the transfer hook extra accounts contains an invalid account index */
+export class TokenTransferHookAccountNotFound extends TokenError {
+    name = 'TokenTransferHookAccountNotFound';
 }
 
 /** Thrown if the transfer hook extra accounts contains an invalid seed */
--- a/token/js/src/errors.ts
+++ b/token/js/src/errors.ts
@@ -15,6 +15,11 @@ export class TokenInvalidAccountError extends TokenError {
     name = 'TokenInvalidAccountError';
 }
 
+/** Thrown if a program state account does not contain valid data */
+export class TokenInvalidAccountDataError extends TokenError {
+    name = 'TokenInvalidAccountDataError';
+}
+
 /** Thrown if a program state account is not owned by the expected token program */
 export class TokenInvalidAccountOwnerError extends TokenError {
     name = 'TokenInvalidAccountOwnerError';
@@ -63,4 +68,14 @@ export class TokenInvalidInstructionTypeError extends TokenError {
 /** Thrown if the program does not support the desired instruction */
 export class TokenUnsupportedInstructionError extends TokenError {
     name = 'TokenUnsupportedInstructionError';
+}
+
+/** Thrown if the transfer hook extra accounts contains an invalid discriminator */
+export class TokenTransferHookInvalidDiscriminator extends TokenError {
+    name = 'TokenTransferHookInvalidDiscriminator';
+}
+
+/** Thrown if the transfer hook extra accounts contains an invalid seed */
+export class TokenTransferHookInvalidSeed extends TokenError {
+    name = 'TokenTransferHookInvalidSeed';
 }
--- a/token/js/src/extensions/transferHook/actions.ts
+++ b/token/js/src/extensions/transferHook/actions.ts
@@ -1,22 +1,18 @@
 import type { Commitment, ConfirmOptions, Connection, Signer, TransactionSignature } from '@solana/web3.js';
 import { TransactionInstruction } from '@solana/web3.js';
-import { PublicKey } from '@solana/web3.js';
+import type { PublicKey } from '@solana/web3.js';
 import { sendAndConfirmTransaction, Transaction } from '@solana/web3.js';
 import { getSigners } from '../../actions/internal.js';
 import { TOKEN_2022_PROGRAM_ID, programSupportsExtensions } from '../../constants.js';
-import { createInitializeTransferHookInstruction, createUpdateTransferHookInstruction } from './instructions.js';
+import {
+    createInitializeTransferHookInstruction,
+    createTransferCheckedWithFeeAndTransferHookInstruction,
+    createTransferCheckedWithTransferHookInstruction,
+    createUpdateTransferHookInstruction,
+} from './instructions.js';
 import { getExtraAccountMetaAccount, getExtraAccountMetas, getTransferHook, resolveExtraAccountMeta } from './state.js';
 import { getMint } from '../../state/index.js';
-import {
-    TokenInvalidAccountDataError,
-    TokenInvalidAccountError,
-    TokenInvalidMintError,
-    TokenUnsupportedInstructionError,
-} from '../../errors.js';
-import {
-    createTransferCheckedInstruction,
-    decodeTransferCheckedInstruction,
-} from '../../instructions/transferChecked.js';
+import { TokenUnsupportedInstructionError } from '../../errors.js';
 
 /**
  * Initialize a transfer hook on a mint
@@ -131,7 +127,7 @@ export async function addExtraAccountsToInstruction(
 }
 
 /**
- * Transfer tokens from one account to another, asserting the transfer fee, token mint, and decimals
+ * Transfer tokens from one account to another, asserting the token mint, and decimals
  *
  * @param connection     Connection to use
  * @param payer          Payer of the transaction fees
@@ -162,26 +158,73 @@ export async function transferCheckedWithTransferHook(
 ): Promise<TransactionSignature> {
     const [authorityPublicKey, signers] = getSigners(authority, multiSigners);
 
-    const rawInstruction = createTransferCheckedInstruction(
-        source,
-        mint,
-        destination,
-        authorityPublicKey,
-        amount,
-        decimals,
-        signers,
-        programId
+    const transaction = new Transaction().add(
+        await createTransferCheckedWithTransferHookInstruction(
+            connection,
+            source,
+            mint,
+            destination,
+            authorityPublicKey,
+            amount,
+            decimals,
+            signers,
+            confirmOptions?.commitment,
+            programId
+        )
     );
 
-    const hydratedInstruction = await addExtraAccountsToInstruction(
-        connection,
-        rawInstruction,
-        mint,
-        confirmOptions?.commitment,
-        programId
-    );
+    return await sendAndConfirmTransaction(connection, transaction, [payer, ...signers], confirmOptions);
+}
 
-    const transaction = new Transaction().add(hydratedInstruction);
+/**
+ * Transfer tokens from one account to another, asserting the transfer fee, token mint, and decimals
+ *
+ * @param connection     Connection to use
+ * @param payer          Payer of the transaction fees
+ * @param source         Source account
+ * @param mint           Mint for the account
+ * @param destination    Destination account
+ * @param authority      Authority of the source account
+ * @param amount         Number of tokens to transfer
+ * @param decimals       Number of decimals in transfer amount
+ * @param fee            The calculated fee for the transfer fee extension
+ * @param multiSigners   Signing accounts if `owner` is a multisig
+ * @param confirmOptions Options for confirming the transaction
+ * @param programId      SPL Token program account
+ *
+ * @return Signature of the confirmed transaction
+ */
+export async function transferCheckedWithFeeAndTransferHook(
+    connection: Connection,
+    payer: Signer,
+    source: PublicKey,
+    mint: PublicKey,
+    destination: PublicKey,
+    authority: Signer | PublicKey,
+    amount: bigint,
+    decimals: number,
+    fee: bigint,
+    multiSigners: Signer[] = [],
+    confirmOptions?: ConfirmOptions,
+    programId = TOKEN_2022_PROGRAM_ID
+): Promise<TransactionSignature> {
+    const [authorityPublicKey, signers] = getSigners(authority, multiSigners);
+
+    const transaction = new Transaction().add(
+        await createTransferCheckedWithFeeAndTransferHookInstruction(
+            connection,
+            source,
+            mint,
+            destination,
+            authorityPublicKey,
+            amount,
+            decimals,
+            fee,
+            signers,
+            confirmOptions?.commitment,
+            programId
+        )
+    );
 
     return await sendAndConfirmTransaction(connection, transaction, [payer, ...signers], confirmOptions);
 }
--- a/token/js/src/extensions/transferHook/actions.ts
+++ b/token/js/src/extensions/transferHook/actions.ts
@@ -1,18 +1,14 @@
-import type { Commitment, ConfirmOptions, Connection, Signer, TransactionSignature } from '@solana/web3.js';
-import { TransactionInstruction } from '@solana/web3.js';
+import type { ConfirmOptions, Connection, Signer, TransactionSignature } from '@solana/web3.js';
 import type { PublicKey } from '@solana/web3.js';
 import { sendAndConfirmTransaction, Transaction } from '@solana/web3.js';
 import { getSigners } from '../../actions/internal.js';
-import { TOKEN_2022_PROGRAM_ID, programSupportsExtensions } from '../../constants.js';
+import { TOKEN_2022_PROGRAM_ID, TOKEN_PROGRAM_ID } from '../../constants.js';
 import {
     createInitializeTransferHookInstruction,
     createTransferCheckedWithFeeAndTransferHookInstruction,
     createTransferCheckedWithTransferHookInstruction,
     createUpdateTransferHookInstruction,
 } from './instructions.js';
-import { getExtraAccountMetaAccount, getExtraAccountMetas, getTransferHook, resolveExtraAccountMeta } from './state.js';
-import { getMint } from '../../state/index.js';
-import { TokenUnsupportedInstructionError } from '../../errors.js';
 
 /**
  * Initialize a transfer hook on a mint
@@ -77,56 +73,6 @@ export async function updateTransferHook(
 }
 
 /**
- * Add extra accounts needed for transfer hook to an instruction
- *
- * @param connection      Connection to use
- * @param instruction     The transferChecked instruction to add accounts to
- * @param commitment      Commitment to use
- * @param programId       SPL Token program account
- *
- * @return Instruction to add to a transaction
- */
-export async function addExtraAccountsToInstruction(
-    connection: Connection,
-    instruction: TransactionInstruction,
-    mint: PublicKey,
-    commitment?: Commitment,
-    programId = TOKEN_2022_PROGRAM_ID
-): Promise<TransactionInstruction> {
-    if (!programSupportsExtensions(programId)) {
-        throw new TokenUnsupportedInstructionError();
-    }
-
-    const mintInfo = await getMint(connection, mint, commitment, programId);
-    const transferHook = getTransferHook(mintInfo);
-    if (transferHook == null) {
-        return instruction;
-    }
-
-    const extraAccountsAccount = getExtraAccountMetaAccount(transferHook.programId, mint);
-    const extraAccountsInfo = await connection.getAccountInfo(extraAccountsAccount, commitment);
-    if (extraAccountsInfo == null) {
-        return instruction;
-    }
-
-    const extraAccountMetas = getExtraAccountMetas(extraAccountsInfo);
-
-    const accountMetas = instruction.keys;
-
-    for (const extraAccountMeta of extraAccountMetas) {
-        const accountMeta = resolveExtraAccountMeta(
-            extraAccountMeta,
-            accountMetas,
-            instruction.data,
-            transferHook.programId
-        );
-        accountMetas.push(accountMeta);
-    }
-
-    return new TransactionInstruction({ keys: accountMetas, programId, data: instruction.data });
-}
-
-/**
  * Transfer tokens from one account to another, asserting the token mint, and decimals
  *
  * @param connection     Connection to use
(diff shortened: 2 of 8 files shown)


token/js/src/extensions/transferHook/state.ts
```diff
@@ -49,3 +52,64 @@ export function getTransferHookAccount(account: Account): TransferHookAccount |
...
+    seq<ExtraAccountsMeta>(ExtraAccountsMetaLayout, greedy(ExtraAccountsMetaLayout.span), 'extraAccounts'),
+]);
+
+/** Unpack an extra account metas account and parse the data into a list of ExtraAccountMetas */
+export function getExtraAccountMetas(account: AccountInfo<Buffer> | null): ExtraAccountsMeta[] | null {
+    if (account == null) {
+        throw null;
+    }
+    return ExtraAccountsMetaListLayout.decode(account.data).extraAccounts;
+}
```
buffalojoec, line 89: It doesn't look like these bad boys need to be exported. Can prob keep them private

token/js/src/extensions/transferHook/state.ts
```diff
@@ -49,3 +52,64 @@ export function getTransferHookAccount(account: Account): TransferHookAccount |
...
+            isSigner: extraMeta.isSigner,
+            isWritable: extraMeta.isWritable,
+        };
+    }
+
+    const accountIndex = extraMeta.discriminator - (1 << 7);
+    if (previousMetas.length <= accountIndex) {
+        throw new TokenTransferHookInvalidDiscriminator();
+    }
+    const programId = extraMeta.discriminator === 1 ? transferHookProgramId : previousMetas[accountIndex].pubkey;
```
buffalojoec, line 110: What do you think of not trying to evaluate the index for the account without first checking the discriminator?

```suggestion
    let programId;
    if (extraMeta.discriminator === 1) {
        programId = transferHookProgramId;
    } else {
        const accountIndex = extraMeta.discriminator - (1 << 7);
        if (previousMetas.length <= accountIndex) {
            throw new TokenTransferHookAccountNotFound(); // New error
        }
        programId = previousMetas[accountIndex].pubkey
    }
```

Note this uses a new error type - which matches the one in the Rust lib - for `AccountNotFound`

token/js/src/errors.ts
```diff
@@ -64,3 +69,13 @@ export class TokenInvalidInstructionTypeError extends TokenError {
 export class TokenUnsupportedInstructionError extends TokenError {
     name = 'TokenUnsupportedInstructionError';
 }
+
+/** Thrown if the transfer hook extra accounts contains an invalid discriminator */
+export class TokenTransferHookInvalidDiscriminator extends TokenError {
+    name = 'TokenTransferHookInvalidDiscriminator';
+}
```
buffalojoec, line 76: Might be able to cut this error, as per other comments

token/js/src/extensions/transferHook/state.ts
```diff
@@ -49,3 +52,64 @@ export function getTransferHookAccount(account: Account): TransferHookAccount |
...
+}
+
+/** Buffer layout for de/serializing a list of ExtraAccountsMeta prefixed by a u32 length */
+export const ExtraAccountsMetaListLayout = struct<ExtraAccountsMetaList>([
+    u32('count'),
+    seq<ExtraAccountsMeta>(ExtraAccountsMetaLayout, greedy(ExtraAccountsMetaLayout.span), 'extraAccounts'),
+]);
+
+/** Unpack an extra account metas account and parse the data into a list of ExtraAccountMetas */
+export function getExtraAccountMetas(account: AccountInfo<Buffer> | null): ExtraAccountsMeta[] | null {
```
buffalojoec, line 84: You have a null check in `addExtraAccountsToInstruction`, do you need this parameter to be nullable? I think it makes sense for it to be a valid `AccountInfo` if you are attempting to deserialize it

token/js/test/unit/transferHook.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    };
+    const pdaExtraAccountWithProgramId = Buffer.concat([
+        Buffer.from([128]), // u8 discriminator
+        addressConfig, // 32 bytes address config
+        Buffer.from([0]), // bool isSigner
+        Buffer.from([1]), // bool isWritable
+    ]);
+
+    const extraAccountList = Buffer.concat([
+        Buffer.from([3, 0, 0, 0]), // u32 count
```
buffalojoec, line 71: The TLV library is using little-endian format for `u32` length, so you'll want to make sure your tests _and_ your buffer layouts are using little-endian.

https://github.com/solana-labs/solana-program-library/blob/737a37a52850a4990c73d16f5c3657844fc2c71a/libraries/type-length-value/src/pod.rs#L35-L48

buffalojoec (commented):
Nice work mate!

The Rust lib separates serialization/deserialization of the seed configs into the `seeds` module, since it has to be used to _pack_ and _unpack_ both on-chain and off-chain, but JavaScript doesn't need that! I appreciate your condensed version of the seed resolution, doing both deserialization and resolution in one step 🤌🏼

I left a handful of comments, but this is shaping up nicely!

buffalojoec:
> * Not sure exactly how multisigners is implemented for extra account resolution in the program since it changes the ordering of the account keys. Does it mean that if a user wants multisigners to work they explicitly need to code this into their transfer hook program?

Multisigs are evaluated at the Token-2022 level. The `TransferChecked` instruction expects - if you're working with multisig - the multisig account _first_ and then the signers immediately preceding.

https://github.com/solana-labs/solana-program-library/blob/737a37a52850a4990c73d16f5c3657844fc2c71a/token/program-2022/src/instruction.rs#L319-L330

You can see here we're reading the data of the authority account and then using `validate_owner` to evaluate the mutisigs from the remaining accounts.

https://github.com/solana-labs/solana-program-library/blob/737a37a52850a4990c73d16f5c3657844fc2c71a/token/program-2022/src/processor.rs#L396-L402

Then your extra metas would come after!

> * What is done by `tlv-account-resolution` in rust is now baked into `@solana/spl-token` for js. Might be a nice improvement to separate that into its own package?

We could separate it, but I mentioned in my comment above that it's actually not necessary at the moment for the JavaScript library to be able to handle (de)serialization of the seed configs and TLV state. However, the more I think about it, the more I think we will eventually want a JavaScript TLV library so we don't have to keep writing the same `count` and `values` structures with buffer layout. Perhaps just a split-out TLV JS library?

> * I have written a few tests to the best of my understanding of what exactly the data structure of the extra accounts data account is. Might be worth for @joncinque or @buffalojoec to check if my implementation actually matches that of the rust client.

Actually you're pretty damn close, I just had the comment about little-endian. It's all facilitated by the [`PodSlice` type](https://github.com/solana-labs/solana-program-library/blob/737a37a52850a4990c73d16f5c3657844fc2c71a/libraries/type-length-value/src/pod.rs#L87) in the `spl-type-length-value` library, which uses bytemuck zero-copy serde to do exactly what you're doing, iterate over deterministic bytes in a slice with a `u32` length at the front to know where to stop.

> * The way `transferCheckedWithHook` and `addExtraAccountsToInstruction` are structured is a little bit made up by me. Let me know if it matches your expectations @joncinque

I think these look pretty good! I don't think we'll really be able to test these handlers properly until we have the instructions from the transfer hook interface baked into the JS client, so we can tell a program to pack some extra metas and then try to transfer with them.

Once there, we can probably pluck these commands out of the `package.json` and into a shell script so we can also deploy a hook program and maybe even another CPI program like the Rust tests do.

https://github.com/solana-labs/solana-program-library/blob/737a37a52850a4990c73d16f5c3657844fc2c71a/token/js/package.json#L44-L47

token/js/src/extensions/transferHook/seeds.ts
```diff
@@ -0,0 +1,80 @@
...
+function unpackSeedInstructionArg(seeds: Uint8Array, instructionData: Buffer): Seed {
+    if (seeds.length < 2) {
+        throw new TokenTransferHookInvalidSeed();
+    }
+    const [index, length] = seeds;
+    if (instructionData.length < length) {
+        throw new TokenTransferHookInvalidSeed();
+    }
+    return {
+        data: instructionData.subarray(index, index + length),
```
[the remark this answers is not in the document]
wjthieme: Good catch!

token/js/src/extensions/transferHook/state.ts
wjthieme: My thinking for these ones being exposed was that potentially someone (probably wallet software) would want to decode the ExtraAccountMeta account manually if they do not want to use `addExtraAccountsToInstruction` which is async.

One such case could be because when someone is fetching multiple accounts using `Connection.getMultipleAccountsInfo` instead of fetching each account sequentially.

Now that I think of it they could also use `getExtraAccountMetas` and `resolveExtraAccountMeta` directly which makes more sense.

Only thing you can do with `ExtraAccountsMetaListLayout` that you cannot `getExtraAccountMetas` is encode the data. This might be nice to have for e2e testing purposes. For example to set arbitrary accounts when using a tool like `solana-bankrun`.

token/js/src/extensions/transferHook/state.ts
wjthieme: Yep this makes more sense

wjthieme:
Is it correct that there is a limitation in` tlv-account-resolution` that accounts using other accounts as seeds need to be ordered after the account that needs them?

i.e. if you are resolving the account at index 3 it can only use accounts 0..2 as seed inputs.

wjthieme:
Extra clarification question

Is it correct that` tlv-account-resolution` has a limitation that accounts using other accounts as seeds need to be ordered after the account that needs them? At least that is what I am seeing in the rust implementation so just copied that logic into js as well.

i.e. if you are resolving the account at index 3 it can only use accounts 0..2 as seed inputs.

token/js/test/unit/transferHook.test.ts
wjthieme: For some reason I always get confused with LE and BE. Isn't this already in LE?

```ts
const buffer = Buffer.from([ 3, 0, 0, 0 ])

buffer.readUInt32LE()
// 4

buffer.readUInt32BE()
// 67108864
```

buffalojoec: Ah shit, apparently I do too 😂. Looks like you're correct

buffalojoec:
> Extra clarification question
> 
> Is it correct that `tlv-account-resolution` has a limitation that accounts using other accounts as seeds need to be ordered after the account that needs them? At least that is what I am seeing in the rust implementation so just copied that logic into js as well.
> 
> i.e. if you are resolving the account at index 3 it can only use accounts 0..2 as seed inputs.

Yep that's correct. I have an implementation where this constraint is no longer required but it was a bit complicated. If we ever need it, I'll pull that bad boy out

token/js/src/extensions/transferHook/seeds.ts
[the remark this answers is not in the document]
buffalojoec: Good catch by you ser! #5120

token/js/src/extensions/transferHook/state.ts
buffalojoec: yeah I guess that's fair. After all you can just use the Rust types directly as well. Carry on mate!

buffalojoec:
> Awesome! So the js implementation for this is correct and a transfer hook program would need to take these extra multisig accounts into consideration if they want to support it.

Yep
 
> > Perhaps just a split-out TLV JS library?
> 
> Yeah I think that would make more sense than a tlv-account-resolution library. Perhaps we can create a issue for it but guess it is not that important for now.

Agreed I swore I made one already but I'll throw one up now #5127

> > so we can tell a program to pack some extra metas and then try to transfer with them
> 
> Yeah was thinking about that as well but figured that was beyond the scope of this PR/issue

Agreed not in this PR. Also made an issue for this! #5128

token/js/src/extensions/transferHook/state.ts
```diff
@@ -49,3 +52,68 @@ export function getTransferHookAccount(account: Account): TransferHookAccount |
...
+/** Buffer layout for de/serializing a list of ExtraAccountMeta prefixed by a u32 length */
+export const ExtraAccountMetaListLayout = struct<ExtraAccountMetaList>([
+    u32('count'),
+    seq<ExtraAccountMeta>(ExtraAccountMetaLayout, greedy(ExtraAccountMetaLayout.span), 'extraAccounts'),
+]);
+
+/** Unpack an extra account metas account and parse the data into a list of ExtraAccountMetas */
+export function getExtraAccountMetas(account: AccountInfo<Buffer>): ExtraAccountMeta[] | null {
+    return ExtraAccountMetaListLayout.decode(account.data).extraAccounts;
+}
```
joncinque, line 86: This won't quite work unfortunately -- it's taking in the full account data buffer and interpreting those bytes as the extra account metas.

The extra account metas are actually encoded as a TLV structure, where `type` is 8 bytes, `length` is 4 bytes, and the `value` is the following `length` bytes.

So in your case, for the `TransferHookInstruction::Execute` instruction, you're looking for the discriminant `[105, 37, 101, 197, 75, 251, 102, 26]`. You'll take the first 8 bytes as the type, then the next 4 bytes as the length. If the first 8 bytes match, then you decode the next `length` bytes (after the length) as `ExtraAccountMetaList`. If they don't match, you'll skip those next `length` bytes, then read the next 8 bytes as the type, 4 bytes as the length, etc etc

Also, be sure to *not* return the whole slice as `ExtraAccountMetas`, just the first `count` items, ie:

```
const decoded = ExtraAccountMetaListLayout.decode(account.data);
return decoded.extraAccounts.slice(0, decoded.count);
```

token/js/test/unit/transferHook.test.ts
```diff
@@ -0,0 +1,139 @@
+import { getExtraAccountMetas, resolveExtraAccountMeta } from '../../src';
+import { expect } from 'chai';
+import { AccountMeta, Keypair, PublicKey } from '@solana/web3.js';
+
+describe('transferHookExtraAccounts', () => {
+    const testProgramId = new PublicKey('7N4HggYEJAtCLJdnHGCtFqfxcB5rhQCsQTze3ftYstVj');
+    const instructionData = Buffer.from(Array.from(Array(32).keys()));
+    const plainAccount = new PublicKey('6c5q79ccBTWvZTEx3JkdHThtMa2eALba5bfvHGf8kA2c');
+    const pdaPublicKey = new PublicKey('8Zibh9ywkKCZc3h23vwxvdMPG9yhg4LYYzky2yRkLHBJ');
+    const pdaPublicKeyWithProgramId = new PublicKey('9Apejge9XvJ3toZ8PX2sKKgMoyN7sSegzdq8n1kMzRae');
```
joncinque, line 10: To give some more confidence in the test, can you derive these PDAs dynamically from the inputs rather than putting them in statically?

joncinque (commented):
Really great work! This is pretty close to ready, thanks so much for contributing. We'll just need to fixup the decoding bit and clean up a few nits, then it should be good to go!

token/js/src/extensions/transferHook/actions.ts
```diff
@@ -65,3 +79,116 @@ export async function updateTransferHook(
...
+        signers,
+        programId
+    );
+
+    const hydratedInstruction = await addExtraAccountsToInstruction(
+        connection,
+        rawInstruction,
+        confirmOptions?.commitment,
+        programId
+    );
```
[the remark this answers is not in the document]
wjthieme: Now that I think of it `transferCheckedWithTransferHook` will only be useful in a specific use case: when you have a Signer (so not in browser) and when there is no transfer fee on the mint.

Would it make sense to have two instruction creators (`createTransferCheckedWithTransferHook` and `createTransferCheckedWithFeeAndTransferHook`) and two instruction actions that just call into those creators?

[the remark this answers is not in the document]
joncinque: Yeah that makes sense to me!

token/js/src/extensions/transferHook/state.ts
joncinque: I don't like merging code that we know doesn't work -- can you at least port the bit to skip the first 12 bytes?

token/js/src/extensions/transferHook/actions.ts
```diff
@@ -65,3 +75,156 @@ export async function updateTransferHook(
...
+ * Add extra accounts needed for transfer hook to an instruction
+ *
+ * @param connection      Connection to use
+ * @param instruction     The transferChecked instruction to add accounts to
+ * @param commitment      Commitment to use
+ * @param programId       SPL Token program account
+ *
+ * @return Instruction to add to a transaction
+ */
+export async function addExtraAccountsToInstruction(
```
joncinque, line 89: nit: Can you move this function down to `instructions.ts`? The `actions` code lives above everything else, so moving this function will break a circular dependency

token/js/src/extensions/transferHook/actions.ts
```diff
@@ -65,3 +75,156 @@ export async function updateTransferHook(
...
+    payer: Signer,
+    source: PublicKey,
+    mint: PublicKey,
+    destination: PublicKey,
+    authority: Signer | PublicKey,
+    amount: bigint,
+    decimals: number,
+    multiSigners: Signer[] = [],
+    confirmOptions?: ConfirmOptions,
+    programId = TOKEN_2022_PROGRAM_ID
```
joncinque, line 157: Let's change this to `TOKEN_PROGRAM_ID`, that way we can recommend people to just switch over to use this function instead, and it'll work without any other changes

token/js/src/extensions/transferHook/actions.ts
```diff
@@ -65,3 +75,156 @@ export async function updateTransferHook(
...
+    source: PublicKey,
+    mint: PublicKey,
+    destination: PublicKey,
+    authority: Signer | PublicKey,
+    amount: bigint,
+    decimals: number,
+    fee: bigint,
+    multiSigners: Signer[] = [],
+    confirmOptions?: ConfirmOptions,
+    programId = TOKEN_2022_PROGRAM_ID
```
joncinque, line 209: Same here, let's use `TOKEN_PROGRAM_ID`

token/js/src/extensions/transferHook/instructions.ts
```diff
@@ -1,11 +1,14 @@
 import { struct, u8 } from '@solana/buffer-layout';
-import type { PublicKey, Signer } from '@solana/web3.js';
+import type { Commitment, Connection, PublicKey, Signer } from '@solana/web3.js';
 import { TransactionInstruction } from '@solana/web3.js';
 import { programSupportsExtensions, TOKEN_2022_PROGRAM_ID } from '../../constants.js';
 import { TokenUnsupportedInstructionError } from '../../errors.js';
 import { addSigners } from '../../instructions/internal.js';
 import { TokenInstruction } from '../../instructions/types.js';
 import { publicKey } from '@solana/buffer-layout-utils';
+import { createTransferCheckedInstruction } from '../../instructions/index.js';
```
joncinque, line 9: nit: can you import this from the actual file for consistency with everything else?
```suggestion
import { createTransferCheckedInstruction } from '../../instructions/transferChecked.js';
```

token/js/src/extensions/transferHook/instructions.ts
```diff
@@ -112,3 +115,106 @@ export function createUpdateTransferHookInstruction(
...
+    connection: Connection,
+    source: PublicKey,
+    mint: PublicKey,
+    destination: PublicKey,
+    authority: PublicKey,
+    amount: bigint,
+    decimals: number,
+    multiSigners: (Signer | PublicKey)[] = [],
+    commitment?: Commitment,
+    programId = TOKEN_2022_PROGRAM_ID
```
joncinque, line 145: Same as the others
```suggestion
    programId = TOKEN_PROGRAM_ID
```

token/js/src/extensions/transferHook/instructions.ts
```diff
@@ -112,3 +115,106 @@ export function createUpdateTransferHookInstruction(
...
+    source: PublicKey,
+    mint: PublicKey,
+    destination: PublicKey,
+    authority: PublicKey,
+    amount: bigint,
+    decimals: number,
+    fee: bigint,
+    multiSigners: (Signer | PublicKey)[] = [],
+    commitment?: Commitment,
+    programId = TOKEN_2022_PROGRAM_ID
```
joncinque, line 197: same as the others
```suggestion
    programId = TOKEN_PROGRAM_ID
```

joncinque (commented):
This is extremely close, just a few last little things! Thanks for your patience and quick responses

token/js/src/extensions/transferHook/actions.ts
wjthieme: Good catch!

joncinque (approved):
This looks good to me, thanks so much for taking this on! You:

rock

I'll give @buffalojoec a chance to add any other comments

buffalojoec (approved):
Beast! Ship it 🚢

Issue #4915, opened 2023-08-02
Introduce Library for Pod Types & Utils


buffalojoec:
## Problem

`bytemuck` and `Pod` types are repeated across SPL.

### Description

SPL crates have been making more persistent use of `bytemuck` and it's associated `Pod` trait to perform byte-wise operations on account state and other buffers.

As this demand for `bytemuck` and Pod-compatible types grows, we've seen some code be duplicated across SPL crates.

Some examples can be seen in:
- [`spl-type-length-value`](https://github.com/solana-labs/solana-program-library/blob/master/libraries/type-length-value/src/pod.rs)
- [`spl-token-metadata-interface`](https://github.com/solana-labs/solana-program-library/blob/master/token-metadata/interface/src/state.rs#L22)
- [`spl-token-2022`](https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/src/pod.rs)

## Solution

Create a library for all `bytemuck` and `Pod` types and utils to be shared across these crates, and also publicly available to others.

qiweiii:
I would like to try this!

serbangv:
Started working on this.

buffalojoec:
> Started working on this.

Thank you mate!

PR #5119, opened 2023-08-25
Move Pod types to separate library

serbangv:
Issue #4915

Created library includes: 
* `OptionalNonZeroPubkey`
* `OptionalNonZeroElGamalPubkey`
* `PodBool`
* `PodU16`, `PodI16`, `PodU32`, `PodI32`, `PodU64`, `PodI64`
* \+ some util functions

Notes:
* There is this change in the code that might seem a bit out of context: `expected_update_authority.clone()` to `*expected_update_authority`. Motive for the change is:
    * `OptionalNonZeroPubkey` was defined in both `token-metadata` and `program-2022`
    * the version defined in `token-metadata` did not derive `Pod`
    * merging these two types led to the following clippy error, `warning: using clone on type OptionalNonZeroPubkey which implements the Copy trait` -- this is why I changed `expected_update_authority.clone()` to `*expected_update_authority`, since by deriving `Pod`, we also derive `Copy` so we don't need the `.clone()` anymore.
* This could be added in this PR or a new one, but should `dep:` be added in `serde-traits` dependencies list? Like `serde-traits = ["dep:serde", "dep:base64"]`? Instead of `serde-traits = ["serde", "base64"]`. Without adding `dep:` it exposes the de optional dependencies as standalone features of the crates, so for example, right now I could run this in `program-2022` without error: `cargo build --features base64`.

@buffalojoec @joncinque


libraries/type-length-value/README.md
```diff
@@ -67,11 +67,11 @@ assert_eq!(other_value.data, 10);
 other_value.data = 2;
 
 // Later on, to work with it again
-let value = state.get_value_mut::<MyPodValue>().unwrap();
+let value = state.get_first_value_mut::<MyPodValue>().unwrap();
```
buffalojoec, line 70: Nice catch. You don't need to worry about it in this PR, but that README needs a whole once-over. I just created #5125 for the rest of this doc!

libraries/type-length-value/src/length.rs
```diff
@@ -1,8 +1,8 @@
 //! Module for the length portion of a Type-Length-Value structure
 use {
-    crate::pod::PodU32,
     bytemuck::{Pod, Zeroable},
     solana_program::program_error::ProgramError,
+    spl_pod::PodU32,
```
buffalojoec, line 5: Same comment as before on the import syntax

token-metadata/example/src/processor.rs
```diff
@@ -31,7 +32,7 @@ fn check_update_authority(
     if !update_authority_info.is_signer {
         return Err(ProgramError::MissingRequiredSignature);
     }
-    let update_authority = Option::<Pubkey>::from(expected_update_authority.clone())
+    let update_authority = Option::<Pubkey>::from(*expected_update_authority)
```
buffalojoec, line 35: 👀

token-metadata/example/tests/program_test.rs
```diff
@@ -101,7 +101,7 @@ pub async fn setup_metadata(
             initialize(
                 metadata_program_id,
                 &metadata_keypair.pubkey(),
-                &Option::<Pubkey>::from(token_metadata.update_authority.clone()).unwrap(),
+                &Option::<Pubkey>::from(token_metadata.update_authority).unwrap(),
```
buffalojoec, line 104: 👀 👀

token-metadata/example/tests/update_authority.rs
```diff
@@ -68,14 +67,14 @@ async fn success_update() {
...
-    token_metadata.update_authority = new_update_authority_pubkey.clone();
+    token_metadata.update_authority = new_update_authority_pubkey;
 
     let transaction = Transaction::new_signed_with_payer(
         &[update_authority(
             &program_id,
             &metadata_pubkey,
             &authority.pubkey(),
-            new_update_authority_pubkey.clone(),
+            new_update_authority_pubkey,
```
buffalojoec, line 77: Thank you for these

buffalojoec (commented):
This is awesome, thanks for doing this!

I left a few comments in my review, and I also wanted to add a couple more questions here.

What do you think about moving the `OptionalNonZeroElGamalPubkey` and `OptionalNonZeroPubkey` into a shared module called something like `non_zero_keys`?

Can we include `PodSlice` and `PodSliceMut` in this lib as well?

Btw, I stopped repeating comments after 2 or 3, but the `Cargo.toml` path and `use` statement syntax ones apply to all crates.

buffalojoec:
Closes #5119 

^ For Actions

libraries/pod/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+//! Crate containing `Pod` types and `bytemuck` utils used in SPL
+
+pub mod non_zero_pubkeys;
+pub mod pod_slice;
```
joncinque, line 4: nit: maybe we call this module just `slice` since none of the others have `pod` in the name?

libraries/pod/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+//! Crate containing `Pod` types and `bytemuck` utils used in SPL
+
+pub mod non_zero_pubkeys;
```
joncinque, line 3: nit: to me, the optionality is the most important part, and not that they're non-zero, so I'd go with `optional_keys` or just `keys`

libraries/pod/src/non_zero_pubkeys.rs
```diff
@@ -0,0 +1,344 @@
...
+    {
+        deserializer.deserialize_any(OptionalNonZeroElGamalPubkeyVisitor)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use {super::*, crate::utils::pod_from_bytes};
+
+    const OPTIONAL_NONZERO_PUBKEY_LEN: usize = 32;
```
joncinque, line 237: nit: you can just use https://github.com/solana-labs/solana/blob/37887d487ce9b87297d9a20dd07a89adf4c66cd5/sdk/program/src/pubkey.rs#L18

libraries/pod/src/pod_slice.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// Error in checked math operation
+    #[error("Error in checked math operation")]
+    CalculationFailure,
+    /// Provided byte buffer too small for expected type
+    #[error("Provided byte buffer too small for expected type")]
+    BufferTooSmall,
+    /// Provided byte buffer too large for expected type
+    #[error("Provided byte buffer too large for expected type")]
+    BufferTooLarge,
+}
```
joncinque, line 28: nit: How about moving this to an `errors.rs` module?

libraries/pod/src/pod_slice.rs
```diff
@@ -0,0 +1,230 @@
...
+            pod_from_bytes, pod_from_bytes_mut, pod_slice_from_bytes, pod_slice_from_bytes_mut,
+        },
+    },
+    bytemuck::Pod,
+    solana_program::program_error::ProgramError,
+};
+
+use spl_program_error::*;
+
+/// Errors that may be returned by the Token program.
```
joncinque, line 16: :eyes: oops, can say "by the spl-pod library"

libraries/pod/src/utils.rs
joncinque, line 1: nit: I typically dislike files called `utils` or `helpers`, since they become a catch-all and you have no idea what's in them. These are all just pass-throughs to bytemuck functions, so how about just calling it `bytemuck.rs`?

joncinque (commented):
Great work, thanks for taking this on! Mostly some nits. I didn't look super closely at the implementations, and I'm assuming that they're just copy-pasted from the existing ones

joncinque:
It also looks like the token-2022 tests are failing for transfer hook stuff:

```
thread 'fail_without_transferring_flag' panicked at 'called `Result::unwrap()` on an `Err` value: TransactionError(InstructionError(1, Custom(1)))', token/transfer-hook-example/tests/functional.rs:719:10
test fail_without_transferring_flag ... FAILED
...
thread 'fail_incorrect_derivation' panicked at 'assertion failed: `(left == right)`
test fail_incorrect_derivation ... FAILED
  left: `InstructionError(1, Custom(1))`,
 right: `InstructionError(1, InvalidSeeds)`', token/transfer-hook-example/tests/functional.rs:485:5
```

serbangv:
> Great work, thanks for taking this on! Mostly some nits. I didn't look super closely at the implementations, and I'm assuming that they're just copy-pasted from the existing ones

Some are modified, I'll add context for those when I get the branch ready.

joncinque:
> Some are modified, I'll add context for those when I get the branch ready.

In that case, please split up the commits so that one commit *just* moves implementations, and another that actually changes them. That way we can accurately review things

serbangv:
> > Some are modified, I'll add context for those when I get the branch ready.
> 
> In that case, please split up the commits so that one commit _just_ moves implementations, and another actually changes them. That way we can accurately review things

Sure, no problem. I'll add a comment when it's done.

libraries/pod/src/pod_slice.rs
serbangv: Done, but used filename `error.rs` to match what other crates use

serbangv:
@buffalojoec @joncinque All comments added so far have been addressed, please check.

libraries/pod/src/optional_keys.rs
```diff
@@ -0,0 +1,340 @@
...
+        if self.0 == ElGamalPubkey::default() {
+            s.serialize_none()
+        } else {
+            s.serialize_some(&self.0.to_string())
+        }
+    }
+}
+
+#[cfg(feature = "serde-traits")]
+pub(crate) struct OptionalNonZeroElGamalPubkeyVisitor;
```
joncinque, line 186: micro-nit: does this need to be `pub` at all? seems like it's only used in the `Serialize` / `Deserialize` impls

libraries/pod/src/slice.rs
```diff
@@ -0,0 +1,217 @@
...
+        // 1 `TestStruct` + length = 37 bytes
+        // we pass 38 to trigger BufferTooLarge
+        let pod_slice_bytes = [1; 38];
+        let result = PodSlice::<TestStruct>::unpack(&pod_slice_bytes);
+        match result {
+            Ok(_) => panic!("Expected an `PodSliceError::BufferTooLarge` error, but got Ok"),
+            Err(err) => {
+                assert_eq!(err, PodSliceError::BufferTooLarge.into());
+            }
+        }
```
joncinque, line 179: nit: is it possible to do `expect_err`? that'll keep things tidier, ie
```suggestion
        let err = PodSlice::<TestStruct>::unpack(&pod_slice_bytes).expect_err("Expected an `PodSliceError::BufferTooLarge` error");
        assert_eq!(err, PodSliceError::BufferTooLarge.into());
```

libraries/pod/src/slice.rs
```diff
@@ -0,0 +1,217 @@
...
+        assert_eq!(*pod_slice.length, PodU32::from(1));
+        pod_slice.push(TestStruct::default()).unwrap();
+        assert_eq!(*pod_slice.length, PodU32::from(2));
+        let result = pod_slice.push(TestStruct::default());
+        match result {
+            Ok(_) => panic!("Expected an `PodSliceError::BufferTooSmall` error, but got Ok"),
+            Err(err) => {
+                assert_eq!(err, PodSliceError::BufferTooSmall.into());
+            }
+        }
```
joncinque, line 215: nit: and last time, `expect_err`?

joncinque (commented):
Thanks for splitting that up, it made a huge difference. Just a few last comments, then it's all good from my side once CI passes!

libraries/pod/src/slice.rs
```diff
@@ -14,6 +14,7 @@ use {
 
 const LENGTH_SIZE: usize = std::mem::size_of::<PodU32>();
 /// Special type for using a slice of `Pod`s in a zero-copy way
+#[derive(Debug)]
```
buffalojoec, line 17: This get left in? 👀

serbangv: Required for `.expect_err()`, which was added to the test below. 🤷‍♂️

buffalojoec: Ohh that's weird. I hate to say it, but I'd rather see you use `unwrap_err` like you did here than add this trait:

https://github.com/solana-labs/solana-program-library/pull/5119/commits/da7e6e5fba7e302f60c7f412279b4fcb76a8411a#diff-715e96a717478b935eac0729cd96d09beaaa5426483ce87d791e82a882bd8f48R251-R262

buffalojoec: Oh that's silly. It will let you do `.err().unwrap()` though, but not `unwrap_err()`.

> No worries, I know these things matter.

Yeah, I just would hate to add extra code/methods just for tests, but `Debug` isn't the worst thing I guess

joncinque (approved):
Looks good to me!

libraries/pod/src/slice.rs
serbangv: Updated, please check.

buffalojoec: Just make `rustfmt` happy and I'm STAMPIN' it

serbangv: My bad, fixed now.

buffalojoec (approved):
Awesome work mate! Shippppp it 🚢

joncinque (approved):
Thank you

serbangv:
> Awesome work mate! Shippppp it 🚢

Thanks! I added you on discord for a quick chat about next steps.

PR #5138, opened 2023-08-28
token-js: added an e2e test for transferring using a mint with a transfer hook extension

### LINKED ISSUE #5128: Transfer Hook Interface Example JS Tests
#### Problem
Currently transfer hook is only supported in JavaScript through the Token2022 library. We don't have any JS helpers for creating instructions for the transfer hook interface (`InitializeExtraAccountMetaList` & `Execute`).

One of the main purposes would be to test the Token2022 account resolution code being shipped by #5112. 

Ideally, we should be able to use [Solana Bankrun](https://github.com/kevinheavey/solana-bankrun) to deploy the example program and test all of our JavaScript transfer hook magic!

#### Solution
Lay down the proper JS helpers for the transfer hook instructions and integrate Solana Bankrun to test against the example program like in [the Rust tests](https://github.com/solana-labs/solana-program-library/blob/master/token/transfer-hook-example/tests/functional.rs)

wjthieme:
resolves #5128
~blocked by #5112~

* Added an e2e test for transferring spl-tokens with a transfer hook extension.
* ~Added a `getExtraAccountMetaAccount` helper function.~ fixed in #5112
* ~Fixed a bug in `transfer-hook-example` where no lamports were transferred to the extraAccountMeta account in the initialize instruction causing it to not be rent-exempt.~ wasn't actually a bug :)
* ~State encode/decode for transfer hooks has become a little bit tricky as there are now multiple structs that require each other. `ExtraAccountMeta` is an individual extra account meta, `ExtraAccountMetaList` is a PodSlice of ExtraAccountMeta and `ExtraAccountMetaAccountData` which is the actual data inside the extraAccountMeta data account which apparently wraps `ExtraAccountMetaList` in a tlv. Any suggestions on naming/structure would be appreciated.~ fixed in #5112
* ~Had to add the extraAccountMeta and transferHookProgram accounts to `addExtraAccountsToInstruction` as they were still missing.~ fixed in #5112

> Ideally, we should be able to use [Solana Bankrun](https://github.com/kevinheavey/solana-bankrun) to deploy the example program and test all of our JavaScript transfer hook magic!

Introducing solana-bankrun as described in the issue for just this single test felt a little bit out of place for this PR so I left that part out.


.github/workflows/pull-request-token.yml
```diff
@@ -197,6 +197,58 @@ jobs:
...
+
+      - name: Build and test transfer-hook-example
+        run: ./ci/cargo-test-sbf.sh token/transfer-hook-example
+
+      - name: Upload program
+        uses: actions/upload-artifact@v2
+        with:
+          name: spl-transfer-hook-example
+          path: "target/deploy/*.so"
+          if-no-files-found: error
```
joncinque, line 250: We're already doing this starting at line 103 -- can you just tack on the last step to upload the program there?

token/transfer-hook-example/src/processor.rs
```diff
@@ -114,14 +116,17 @@ pub fn process_initialize_extra_account_meta_list(
...
+        program_id,
+    );
     invoke_signed(
-        &system_instruction::allocate(extra_account_metas_info.key, account_size as u64),
-        &[extra_account_metas_info.clone()],
-        &[&signer_seeds],
-    )?;
-    invoke_signed(
-        &system_instruction::assign(extra_account_metas_info.key, program_id),
-        &[extra_account_metas_info.clone()],
```
joncinque, line 124: This is done deliberately, and not a bug -- since we're defining an interface, we *cannot* expose an instruction that takes a wallet as a signer and writable, since any malicious program implementing the interface can just immediately rug all funds in the user's wallet.

In fact, it's an antipattern to have people pass in their wallet as a signer + writable to any program, especially if the system program is included! They should just transfer the lamports directly to the account separately, that way there's no risk of getting rugged ever.

Was there a reason to change this?

token/js/test/e2e-2022/transferHook.test.ts
```diff
@@ -15,31 +15,57 @@ import {
...
         payer = await newAccountWithLamports(connection, 1000000000);
+        destinationAuthority = Keypair.generate().publicKey;
         transferHookAuthority = Keypair.generate();
-        transferHookProgramId = Keypair.generate().publicKey;
-        newTransferHookProgramId = Keypair.generate().publicKey;
     });
     beforeEach(async () => {
         const mintKeypair = Keypair.generate();
         mint = mintKeypair.publicKey;
+        pdaExtraAccountMeta = getExtraAccountMetaAccount(TRANSFER_HOOK_TEST_PROGRAM_ID, mint);
```
joncinque, line 52: nit: This can be done separately, but I'm realizing now that this method isn't actually fetching an account, it's deriving an address, so it should be called `getExtraAccountMetaAddress`, and we should reverse the arguments for consistency with the address deriving functions

token/js/test/e2e-2022/transferHook.test.ts
```diff
@@ -106,4 +133,96 @@ describe('transferHook', () => {
...
+            TEST_PROGRAM_ID
+        );
+
+        const transferCheckedWithExtraAccountMeta = await addExtraAccountsToInstruction(
+            connection,
+            transferCheckedInstruction,
+            mint,
+            undefined,
+            TEST_PROGRAM_ID
+        );
```
joncinque, line 213: nit: what do you think about using the action helper that you created instead? that way we exercise a bit more code

joncinque (commented):
The test itself looks great! Just a few other little things to address, we should be able to get this in soon

token/transfer-hook-example/src/processor.rs
wjthieme: Never looked at it this way but it makes sense. I changed this to make the test work which in itself should have been a red flag :))).  Let me change this back.

.github/workflows/pull-request-token.yml
wjthieme: Overlooked that! Will change

token/transfer-hook-example/src/processor.rs
joncinque: Ok great, thanks!

token/js/test/e2e-2022/transferHook.test.ts
wjthieme: Will create a new PR for that!

token/js/test/e2e-2022/transferHook.test.ts
wjthieme: Yep makes sense!

joncinque (approved):
Looks great, thanks!

thank hook

PR #5152, opened 2023-08-29
token-2022: Featurize `zk-ops` to build without it

joncinque:
#### Problem

It isn't possible to build token-2022 with `--no-default-features`, since some parts depend on the `zk-ops` feature being enabled.

#### Solution

Add the feature check wherever needed to get `cargo build-sbf --no-default-features` to work.


joncinque:
Merging early to unblock

PR #5169, opened 2023-08-30
SPL errors from hashes

### LINKED ISSUE #5042: [Token-2022] Transfer hook interface error reporting

buffalojoec:
This PR adds unique, custom error codes to every error in SPL libraries - including the two interfaces `transfer-hook` and `token-metadata`.

The ability to derive unique `u32` error codes from hash inputs is powered by these changes to the `spl-program-error` crate, which simply hashes the following combination of inputs and takes bytes 13 through 16 as the little-endian `u32`:

```text
"spl_program_error:" + <enum name> + ":" + <variant name>
```

This creates unique error codes for each error, as requested in [this comment](https://github.com/solana-labs/solana-program-library/pull/5168#issuecomment-1697853488)

Closes #5042 
Solves discussions in #5168

### DIFF
--- a/libraries/program-error/derive/src/lib.rs
+++ b/libraries/program-error/derive/src/lib.rs
@@ -64,12 +64,17 @@ pub fn print_program_error(input: TokenStream) -> TokenStream {
 /// - `solana_program::decode_error::DecodeError`
 /// - `solana_program::program_error::PrintProgramError`
 ///
-/// Optionally, you can add `hash_error_codes: bool` argument to create unique
-/// `u32` error codes from the names of the enum variants.
+/// Optionally, you can add `hash_error_code_start: u32` argument to create
+/// a unique `u32` _starting_ error codes from the names of the enum variants.
+/// Notes:
+/// - The _error_ variant will start at this value, and the rest will be
+/// incremented by one
+/// - The value provided is only for code readability, the actual error code
+/// will be a hash of the input string and is checked against your input
 ///
-/// Syntax: `#[spl_program_error(hash_error_codes = true)]`
+/// Syntax: `#[spl_program_error(hash_error_code_start = 1275525928)]`
 /// Hash Input: `spl_program_error:<enum name>:<variant name>`
-/// Value: `u32::from_le_bytes(<hash of input>[8..12])`
+/// Value: `u32::from_le_bytes(<hash of input>[13..17])`
 #[proc_macro_attribute]
 pub fn spl_program_error(attr: TokenStream, input: TokenStream) -> TokenStream {
     let args = parse_macro_input!(attr as SplProgramErrorArgs);
--- a/libraries/program-error/derive/src/lib.rs
+++ b/libraries/program-error/derive/src/lib.rs
@@ -14,39 +14,67 @@
 extern crate proc_macro;
 
 mod macro_impl;
+mod parser;
 
-use macro_impl::MacroType;
-use proc_macro::TokenStream;
-use syn::{parse_macro_input, ItemEnum};
+use {
+    crate::parser::SplProgramErrorArgs,
+    macro_impl::MacroType,
+    proc_macro::TokenStream,
+    syn::{parse_macro_input, ItemEnum},
+};
 
-/// Derive macro to add `Into<solana_program::program_error::ProgramError>` traits
+/// Derive macro to add `Into<solana_program::program_error::ProgramError>`
+/// trait
 #[proc_macro_derive(IntoProgramError)]
 pub fn into_program_error(input: TokenStream) -> TokenStream {
-    MacroType::IntoProgramError
-        .generate_tokens(parse_macro_input!(input as ItemEnum))
+    let ItemEnum { ident, .. } = parse_macro_input!(input as ItemEnum);
+    MacroType::IntoProgramError { ident }
+        .generate_tokens()
         .into()
 }
 
 /// Derive macro to add `solana_program::decode_error::DecodeError` trait
 #[proc_macro_derive(DecodeError)]
 pub fn decode_error(input: TokenStream) -> TokenStream {
-    MacroType::DecodeError
-        .generate_tokens(parse_macro_input!(input as ItemEnum))
-        .into()
+    let ItemEnum { ident, .. } = parse_macro_input!(input as ItemEnum);
+    MacroType::DecodeError { ident }.generate_tokens().into()
 }
 
 /// Derive macro to add `solana_program::program_error::PrintProgramError` trait
 #[proc_macro_derive(PrintProgramError)]
 pub fn print_program_error(input: TokenStream) -> TokenStream {
-    MacroType::PrintProgramError
-        .generate_tokens(parse_macro_input!(input as ItemEnum))
+    let ItemEnum {
+        ident, variants, ..
+    } = parse_macro_input!(input as ItemEnum);
+    MacroType::PrintProgramError { ident, variants }
+        .generate_tokens()
         .into()
 }
 
 /// Proc macro attribute to turn your enum into a Solana Program Error
+///
+/// Adds:
+/// - `Clone`
+/// - `Debug`
+/// - `Eq`
+/// - `PartialEq`
+/// - `thiserror::Error`
+/// - `num_derive::FromPrimitive`
+/// - `Into<solana_program::program_error::ProgramError>`
+/// - `solana_program::decode_error::DecodeError`
+/// - `solana_program::program_error::PrintProgramError`
+///
+/// Optionally, you can add `hash_error_codes: bool` argument to create unique
+/// `u32` error codes from the names of the enum variants.
+///
+/// Syntax: `#[spl_program_error(hash_error_codes = true)]`
+/// Hash Input: `spl_program_error:<enum name>:<variant name>`
+/// Value: `u32::from_le_bytes(<hash of input>[8..12])`
 #[proc_macro_attribute]
-pub fn spl_program_error(_: TokenStream, input: TokenStream) -> TokenStream {
-    MacroType::SplProgramError
-        .generate_tokens(parse_macro_input!(input as ItemEnum))
+pub fn spl_program_error(attr: TokenStream, input: TokenStream) -> TokenStream {
+    let args = parse_macro_input!(attr as SplProgramErrorArgs);
+    let item_enum = parse_macro_input!(input as ItemEnum);
+    MacroType::SplProgramError { args, item_enum }
+        .generate_tokens()
         .into()
 }
--- a/libraries/program-error/derive/src/macro_impl.rs
+++ b/libraries/program-error/derive/src/macro_impl.rs
@@ -11,7 +11,7 @@ use {
 };
 
 const SPL_ERROR_HASH_NAMESPACE: &str = "spl_program_error";
-const SPL_ERROR_HASH_MIN_VALUE: u32 = 7_0000;
+const SPL_ERROR_HASH_MIN_VALUE: u32 = 7_000;
 
 /// The type of macro being called, thus directing which tokens to generate
 #[allow(clippy::enum_variant_names)]
--- a/libraries/program-error/derive/src/macro_impl.rs
+++ b/libraries/program-error/derive/src/macro_impl.rs
@@ -120,8 +120,8 @@ pub fn spl_program_error(
     args: &SplProgramErrorArgs,
     item_enum: &mut ItemEnum,
 ) -> proc_macro2::TokenStream {
-    if args.hash_error_codes {
-        build_discriminants(item_enum);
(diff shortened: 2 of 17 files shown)


token/transfer-hook-interface/src/error.rs
```diff
@@ -1,17 +1,9 @@
...
-        msg,
-        program_error::{PrintProgramError, ProgramError},
-    },
-    thiserror::Error,
-};
+use spl_program_error::*;
 
 /// Errors that may be returned by the interface.
-#[derive(Clone, Debug, Eq, Error, FromPrimitive, PartialEq)]
+#[spl_program_error(hash_error_codes = true)]
```
joncinque, line 6: :tada:

libraries/program-error/derive/src/macro_impl.rs
```diff
@@ -114,3 +143,37 @@ pub fn spl_program_error(input: ItemEnum) -> proc_macro2::TokenStream {
...
+        let variant_ident = &variant.ident;
+        let discriminant = u32_from_hash(enum_ident, variant_ident);
+        let eq = Token![=](Span::call_site());
+        let expr = Expr::Lit(ExprLit {
+            attrs: Vec::new(),
+            lit: Lit::Int(LitInt::new(&discriminant.to_string(), Span::call_site())),
+        });
+        variant.discriminant = Some((eq, expr));
+    }
+}
```
joncinque, line 164: This is really neat!

joncinque (commented):
The implementation looks perfect! I did just realize though... we'll need a way for someone at the top-level to get more information about the error, since they're essentially random numbers :sweat: 

Ok, hear me out, what if instead of giving *each error* a random number, we give the *first* error a random hashed number of `namespace:error_name`, and everything counts up from there? That way someone just needs to find out that first number, which we force people to add to the source code somewhere to make it easier to discover, similar to `solana-frozen-abi`, ie:

```
#[spl_program_error(hash_error_codes = true, error_start = 42)]
pub enum MyError {
    OhNoes,
}
```

Where `42` is `hash('spl_program_error:MyError')[13..17]`. If it's incorrect, the macro tells you what number to set as `error_start`.

We could maybe condense those two into one arg, but I can't come up with any good names. Maybe just have `expected_error_start`? Or expose a new macro, like `spl_program_error_with_unique_error_codes` which does the hashing and also takes in the expected start number? These names all stink, so I'll noodle on something better.

joncinque:
> On the topic of collisions, I felt like it was probably also overkill, but I have one particular concern about a hash randomly resulting in, say [1, 0, 0, 0] and colliding with (literally any) program errors. For this reason I suggest we lean toward a fixed 9 on the MSB and take only bytes 13, 14, and 15 from the hash. wdyt?

Ah right, that's a good point. Rather than remove a whole byte from the randomness, how about we just make sure that the number is >= 256 to be totally safe? and if it happens to be 0 <= x < 256, just shift over 1 byte until you get to a valid set of bytes.

> It smells a bit like a macro anti-pattern to me

If you have a better way of surfacing that number easily for people looking at the source, I'm game! But I couldn't come up with anything else

buffalojoec:
> > On the topic of collisions, I felt like it was probably also overkill, but I have one particular concern about a hash randomly resulting in, say [1, 0, 0, 0] and colliding with (literally any) program errors. For this reason I suggest we lean toward a fixed 9 on the MSB and take only bytes 13, 14, and 15 from the hash. wdyt?

smh, I meant `256` not `9`

> Ah right, that's a good point. Rather than remove a whole byte from the randomness, how about we just make sure that the number is >= 256 to be totally safe? and if it happens to be 0 <= x < 256, just shift over 1 byte until you get to a valid set of bytes.

Which number are you referring to? The entire `u32` integer itself? In my opinion we want to clear at least 5000 for Anchor errors, and likely even go a little further just to be safe for people who like to categorize with leading digits. Say, `10_000` or `100_000`?

> > It smells a bit like a macro anti-pattern to me
> 
> If you have a better way of surfacing that number easily for people looking at the source, I'm game! But I couldn't come up with anything else

Nah I think it's okay. Unless I find some Rustaceans barking about it online, it's going in

joncinque:
> Which number are you referring to? The entire u32 integer itself? In my opinion we want to clear at least 5_000 for Anchor errors, and likely even go a little further just to be safe for people who like to categorize with leading digits. Say, 10_000 or 100_000?

Sorry, yeah I mean the whole 4 bytes interpreted as a little-endian u32. Does Anchor start at 5000 for all programs? That seems very odd. If that's the case, we can just do 6000 or something, right? And that's already being plenty generous I'd say. If people use their own start, that's similar to just grinding a bunch of error names until you get something you like.

buffalojoec:
> > Which number are you referring to? The entire u32 integer itself? In my opinion we want to clear at least 5_000 for Anchor errors, and likely even go a little further just to be safe for people who like to categorize with leading digits. Say, 10_000 or 100_000?
> 
> Sorry, yeah I mean the whole 4 bytes interpreted as a little-endian u32. Does Anchor start at 5000 for all programs? That seems very odd. If that's the case, we can just do 6000 or something, right? And that's already being plenty generous I'd say. If people use their own start, that's similar to just grinding a bunch of error names until you get something you like.

They use the first digit as a category https://docs.rs/anchor-lang/latest/anchor_lang/error/enum.ErrorCode.html
and then users start at 6000 https://docs.rs/anchor-lang/latest/anchor_lang/error/constant.ERROR_CODE_OFFSET.html

I have 10,000 in my latest commit, I guess we could get away with 7,000 ?

libraries/program-error/derive/src/macro_impl.rs
```diff
@@ -114,3 +144,53 @@ pub fn spl_program_error(input: ItemEnum) -> proc_macro2::TokenStream {
...
+
+/// Hashes the `SPL_ERROR_HASH_NAMESPACE` constant, the enum name and variant
+/// name and returns four middle bytes (13 through 16) as a u32.
+fn u32_from_hash(enum_ident: &Ident, variant_ident: &Ident) -> u32 {
+    let hash_input = format!(
+        "{}:{}:{}",
+        SPL_ERROR_HASH_NAMESPACE, enum_ident, variant_ident
+    );
+
+    // We don't want our error code to start at any number below `10_0000`!
```
joncinque, line 182: nit:
```suggestion
    // We don't want our error code to start at any number below `SPL_ERROR_HASH_MIN_VALUE`!
```

libraries/program-error/derive/src/macro_impl.rs
```diff
@@ -114,3 +144,53 @@ pub fn spl_program_error(input: ItemEnum) -> proc_macro2::TokenStream {
...
+    if discriminant == error_code_start {
+        let eq = Token![=](Span::call_site());
+        let expr = Expr::Lit(ExprLit {
+            attrs: Vec::new(),
+            lit: Lit::Int(LitInt::new(&discriminant.to_string(), Span::call_site())),
+        });
+        first_variant.discriminant = Some((eq, expr));
+    } else {
+        panic!(
+            "Error code start value from hash is {}. Update your macro attribute.",
```
joncinque, line 168: Let's tell people *exactly* what to do
```suggestion
            "Error code start value from hash must be {0}. Update your macro use to `#[spl_program_error(hash_error_code_start = {0})]`.",
```

libraries/program-error/derive/src/lib.rs
```diff
@@ -14,39 +14,72 @@
...
+/// Optionally, you can add `hash_error_code_start: u32` argument to create
+/// a unique `u32` _starting_ error codes from the names of the enum variants.
+/// Notes:
+/// - The _error_ variant will start at this value, and the rest will be
+/// incremented by one
+/// - The value provided is only for code readability, the actual error code
+/// will be a hash of the input string and is checked against your input
+///
+/// Syntax: `#[spl_program_error(hash_error_code_start = 1275525928)]`
+/// Hash Input: `spl_program_error:<enum name>:<variant name>`
```
joncinque, line 76: I might be missing something, but is there a reason to include the variant name? I'm thinking it would be clearer to just use the enum name.

joncinque (commented):
Just few tiny last little questions, this is extremely close!

libraries/program-error/derive/src/lib.rs
buffalojoec: Yeah good point, probably not needed now that we're only hashing variant 0

libraries/program-error/derive/src/macro_impl.rs
```diff
@@ -1,32 +1,52 @@
...
+    proc_macro2::Span,
+    quote::quote,
+    syn::{
+        punctuated::Punctuated, token::Comma, Expr, ExprLit, Ident, ItemEnum, Lit, LitInt, LitStr,
+        Token, Variant,
+    },
+};
+
+const SPL_ERROR_HASH_NAMESPACE: &str = "spl_program_error";
+const SPL_ERROR_HASH_MIN_VALUE: u32 = 10_0000;
```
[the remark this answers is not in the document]
buffalojoec: Some people must have said "but `u32` is huuuuge!"

buffalojoec:
Comments resolved!

This is a breaking change for anything using these libraries downstream. I'm thinking:
- `spl-program-error`: `1.0.0` or `0.3.0` if we want to keep it in development
- Any downstream crates: Bump minor version

wdyt?

libraries/program-error/derive/src/macro_impl.rs
```diff
@@ -11,7 +11,7 @@ use {
 };
 
 const SPL_ERROR_HASH_NAMESPACE: &str = "spl_program_error";
-const SPL_ERROR_HASH_MIN_VALUE: u32 = 10_0000;
+const SPL_ERROR_HASH_MIN_VALUE: u32 = 7_0000;
```
joncinque, line 14: Haha I just realized that this has an extra 0, should be `7_000`. Good thing we have multiple rounds of reviews

joncinque:
> This is a breaking change for anything using these libraries downstream

Sorry, maybe I'm missing something, but how is this change breaking? It passes CI without any changes to other libraries that are using it currently

libraries/program-error/derive/src/macro_impl.rs
buffalojoec: Ah! my bad

buffalojoec:
> > This is a breaking change for anything using these libraries downstream
> 
> Sorry, maybe I'm missing something, but how is this change breaking? It passes CI without any changes to other libraries that are using it currently

Is it not _technically_ breaking since the underlying error codes are changing?

CI is passing because I included changes to the tests where the expected error changed.

For example, anyone expecting `TransferHookError::IncorrectAccount` is now getting `AccountResolutionError::IncorrectAccount`, and furthermore anyone expecting some `u32` like `1` is now getting `13498579` or something wild.

joncinque:
> Is it not technically breaking since the underlying error codes are changing?

But that's only if you use `hash_error_code_start`, right?

buffalojoec:
> > Is it not technically breaking since the underlying error codes are changing?
> 
> But that's only if you use `hash_error_code_start`, right?

I'm thinking about if I'm working with `spl_token_2022` and I'm expecting certain codes from, say transfer hook helpers. Also `spl_token_client` for that matter

joncinque (approved):
Looks good!

hash

buffalojoec:
Lmao Lennon would approve

PR #5177, opened 2023-08-30
Fix incorrect code doc

CriesofCarrots:
`pod_slice_from_bytes()` return type isn't mutable, so let's fix the doc that suggests it will be.

Also, I find the docs in this file a little hard to read, because the word `slice` is used so many times. Do you think it is worth saying `slice of bytes` or `byte slice` or something to distinguish those slices from the slices of `Pod`?


joncinque (approved):
Thanks for catching this! And `slice of bytes` would be much clearer overall

PR #5179, opened 2023-08-31
stake-pool: Use unaligned types for safe pointer cast

joncinque:
#### Problem

Under Rust 1.72, the cast done by the stake pool program from raw bytes into a `ValidatorStakeInfo` panics because the underlying pointer might not be aligned to an address with a multiple of 8. See the offending line: https://github.com/solana-labs/solana-program-library/blob/c79c727f88710e565cd47bde0049f1b449aef0eb/stake-pool/program/src/big_vec.rs#L163C1-L163C80

And see https://github.com/solana-labs/solana/actions/runs/6029425589/job/16358927677?pr=32961 for a failing run.

#### Solution

It causes loads of churn, but do the right thing: use the alignment-safe `PodU64` and `PodU32` instead of `u64` and `u32`, respectively.

Note: this is blocking the upgrade of the monorepo to 1.72 because of downstream build errors

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6898,6 +6898,7 @@ dependencies = [
  "solana-sdk",
  "solana-vote-program",
  "spl-math",
+ "spl-pod",
  "spl-token 4.0.0",
  "spl-token-2022 0.7.0",
  "test-case",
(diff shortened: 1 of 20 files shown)


joncinque:
Merging to unblock the monorepo

PR #5181, opened 2023-08-31
Serde optional dependencies clean-up

serbangv:
When declaring an optional dependency in `Cargo.toml` like so: `serde_with = { version = "3.3.0", optional = true }`, it will automatically define a new feature, in this case `serde_with`.  Adding the `dep:` prefix prevents this.

This is why some spl crates have extra feature flags, for example `spl-token-2022` has `serde` and `serde_with` features, https://docs.rs/crate/spl-token-2022/0.7.0/features

This PR should fix it.

@buffalojoec @joncinque

### DIFF
--- a/libraries/pod/Cargo.toml
+++ b/libraries/pod/Cargo.toml
@@ -8,7 +8,7 @@ license = "Apache-2.0"
 edition = "2021"
 
 [features]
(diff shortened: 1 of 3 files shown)


joncinque (approved):
Thanks for cleaning things up!

clean

PR #5184, opened 2023-09-01
Added a TLVData parsing helper class

### LINKED ISSUE #5127: Type-Length-Value JS Lib
#### Problem
Right now, we have two nifty libraries for managing type-length-value entries in account state, but neither of them are replicated for our JavaScript friends!

#### Solution
Create a TLV library in JavaScript that mimics the Rust lib at `libraries/type-length-value`.

Note: We may also want to either create a library for TLV Account Resolution as well, or combine these two libraries somehow.

See discussion in #5112

wjthieme:
resolves #5127

In this PR I have isolated parsing of TLVData into its own file. For ticket #5127, I think this is fine for now as this allows us to very quickly extract TLVData parsing into its own library if we ever need it.

The `TLVData` class adds two helper functions for reading TLVData.
* `entry(type)` returns the first entry where the type matches the supplied type (preferred)
* `entries()` returns a map of all entries in the TLVData.

The TLVData class allows you to specify the length/span of both the type and length part of the tlv which should make parsing tlv data much easier in the future.

I added a couple simple tests for parsing and reading TLVData. 

This new class is not actually being used yet anywhere in the code. I plan to create (at least) two follow up PRs that switch out the old implementation for this new one once this PR is merged:
* `getExtensionData` in `./token/js/extension/extensionType.ts`
* `getExtraAccountMetaAccount` in `./token/js/exension/transferHook/state.ts`
* Anything else?

### DIFF
--- a/token/js/src/extensions/tlvData.ts
+++ b/token/js/src/extensions/tlvData.ts
@@ -1,0 +1,95 @@
+export type TLVNumberSize = 1 | 2 | 4 | 8;
+
+function readTLVNumberSize<T>(
+    buffer: Buffer,
+    size: TLVNumberSize,
+    offset: number,
+    constructor: (x: number | bigint) => T
+): T {
+    switch (size) {
+        case 1:
+            return constructor(buffer.readUInt8(offset));
+        case 2:
+            return constructor(buffer.readUInt16LE(offset));
+        case 4:
+            return constructor(buffer.readUInt32LE(offset));
+        case 8:
+            return constructor(buffer.readBigUInt64LE(offset));
+    }
+}
+
+export class TLVData {
+    private readonly tlvData: Buffer;
+    private readonly typeSize: TLVNumberSize;
+    private readonly lengthSize: TLVNumberSize;
+
+    public constructor(buffer: Buffer, typeSize: TLVNumberSize = 2, lengthSize: TLVNumberSize = 2, offset: number = 0) {
+        this.tlvData = buffer.subarray(offset);
+        this.typeSize = typeSize;
+        this.lengthSize = lengthSize;
+    }
+
+    /**
+     * Get the raw tlv data
+     *
+     * @return the raw tlv data
+     */
+    public get data(): Buffer {
+        return this.tlvData;
+    }
+
+    /**
+     * Get a single entry from the tlv data. This function performs a greedy search returning the first entry with the given type.
+     *
+     * @param type the type of the entry to get
+     *
+     * @return the entry from the tlv data or null
+     */
+    public entry(type: Buffer | Uint8Array | number | bigint): Buffer | null {
+        let offset = 0;
+        while (offset + this.typeSize + this.lengthSize <= this.tlvData.length) {
+            let typeMatches = false;
+            switch (typeof type) {
+                case 'number':
+                    typeMatches = readTLVNumberSize(this.tlvData, this.typeSize, offset, Number) === type;
+                    break;
+                case 'bigint':
+                    typeMatches = readTLVNumberSize(this.tlvData, this.typeSize, offset, BigInt) === type;
+                    break;
+                default:
+                    typeMatches = this.tlvData.subarray(offset, offset + this.typeSize).equals(type);
+                    break;
+            }
+            offset += this.typeSize;
+            const entryLength = readTLVNumberSize(this.tlvData, this.lengthSize, offset, Number);
+            offset += this.lengthSize;
+            if (typeMatches && offset + entryLength <= this.tlvData.length) {
+                return this.tlvData.subarray(offset, offset + entryLength);
+            }
+            offset += entryLength;
+        }
+        return null;
+    }
+
+    /**
+     * Get all entries from the tlv data. If you need to get a single entry, use the entry function instead.
+     *
+     * @return a map of all entries in the tlv data
+     */
+    public entries(): Map<bigint, Buffer> {
+        let offset = 0;
+        const entries = new Map<bigint, Buffer>();
+        while (offset + this.typeSize + this.lengthSize <= this.tlvData.length) {
+            const entryType = readTLVNumberSize(this.tlvData, this.typeSize, offset, BigInt);
+            offset += this.typeSize;
+            const entryLength = readTLVNumberSize(this.tlvData, this.lengthSize, offset, Number);
+            offset += this.lengthSize;
+            if (offset + entryLength <= this.tlvData.length) {
+                const entryData = this.tlvData.subarray(offset, offset + entryLength);
+                entries.set(entryType, entryData);
+            }
+            offset += entryLength;
+        }
+        return entries;
+    }
+}
--- a/token/js/test/unit/tlvData.test.ts
+++ b/token/js/test/unit/tlvData.test.ts
@@ -1,0 +1,117 @@
+import type { TLVNumberSize } from '../../src/extensions/tlvData';
+import { TLVData } from '../../src/extensions/tlvData';
+import { expect } from 'chai';
+
+describe('tlvData', () => {
+    // typeLength 1, lengthSize 2
+    const tlvData1 = Buffer.concat([
+        Buffer.from([0]),
+        Buffer.from([0, 0]),
+        Buffer.from([]),
+        Buffer.from([1]),
+        Buffer.from([1, 0]),
+        Buffer.from([1]),
+        Buffer.from([2]),
+        Buffer.from([2, 0]),
+        Buffer.from([1, 2]),
+    ]);
+
+    // typeLength 2, lengthSize 1
+    const tlvData2 = Buffer.concat([
+        Buffer.from([0, 0]),
+        Buffer.from([0]),
+        Buffer.from([]),
+        Buffer.from([1, 0]),
+        Buffer.from([1]),
+        Buffer.from([1]),
+        Buffer.from([2, 0]),
+        Buffer.from([2]),
+        Buffer.from([1, 2]),
+    ]);
+
+    // typeLength 4, lengthSize 8
+    const tlvData3 = Buffer.concat([
+        Buffer.from([0, 0, 0, 0]),
+        Buffer.from([0, 0, 0, 0, 0, 0, 0, 0]),
+        Buffer.from([]),
+        Buffer.from([1, 0, 0, 0]),
+        Buffer.from([1, 0, 0, 0, 0, 0, 0, 0]),
+        Buffer.from([1]),
+        Buffer.from([2, 0, 0, 0]),
+        Buffer.from([2, 0, 0, 0, 0, 0, 0, 0]),
+        Buffer.from([1, 2]),
+    ]);
+
+    // typeLength 8, lengthSize 4
+    const tlvData4 = Buffer.concat([
+        Buffer.from([0, 0, 0, 0, 0, 0, 0, 0]),
+        Buffer.from([0, 0, 0, 0]),
+        Buffer.from([]),
+        Buffer.from([1, 0, 0, 0, 0, 0, 0, 0]),
+        Buffer.from([1, 0, 0, 0]),
+        Buffer.from([1]),
+        Buffer.from([2, 0, 0, 0, 0, 0, 0, 0]),
+        Buffer.from([2, 0, 0, 0]),
+        Buffer.from([1, 2]),
+    ]);
+
+    const testRawData = (tlvData: Buffer, typeSize: TLVNumberSize, lengthSize: TLVNumberSize) => {
+        const tlv = new TLVData(tlvData, typeSize, lengthSize);
+        expect(tlv.data).to.be.deep.equal(tlvData);
+        const tlvWithOffset = new TLVData(tlvData, typeSize, lengthSize, typeSize + lengthSize);
+        expect(tlvWithOffset.data).to.be.deep.equal(tlvData.subarray(typeSize + lengthSize));
+    };
+
+    it('should get the raw tlv data', () => {
+        testRawData(tlvData1, 1, 2);
+        testRawData(tlvData2, 2, 1);
+        testRawData(tlvData3, 4, 8);
+        testRawData(tlvData4, 8, 4);
+    });
+
+    const testIndividualEntries = (tlvData: Buffer, typeSize: TLVNumberSize, lengthSize: TLVNumberSize) => {
+        const tlv = new TLVData(tlvData, typeSize, lengthSize);
+        expect(tlv.entry(Number(0))).to.be.deep.equal(Buffer.alloc(0));
+        expect(tlv.entry(Number(1))).to.be.deep.equal(Buffer.from([1]));
+        expect(tlv.entry(Number(2))).to.be.deep.equal(Buffer.from([1, 2]));
+        expect(tlv.entry(Number(3))).to.be.null;
+        expect(tlv.entry(BigInt(0))).to.be.deep.equal(Buffer.alloc(0));
+        expect(tlv.entry(BigInt(1))).to.be.deep.equal(Buffer.from([1]));
+        expect(tlv.entry(BigInt(2))).to.be.deep.equal(Buffer.from([1, 2]));
+        expect(tlv.entry(BigInt(3))).to.be.null;
+
+        const type = Buffer.alloc(typeSize);
+        type[0] = 0;
+        expect(tlv.entry(type)).to.be.deep.equal(Buffer.alloc(0));
+        type[0] = 1;
+        expect(tlv.entry(type)).to.be.deep.equal(Buffer.from([1]));
+        type[0] = 2;
+        expect(tlv.entry(type)).to.be.deep.equal(Buffer.from([1, 2]));
+        type[0] = 3;
+        expect(tlv.entry(type)).to.be.null;
+    };
+
+    it('should get the entries individually', () => {
+        testIndividualEntries(tlvData1, 1, 2);
+        testIndividualEntries(tlvData2, 2, 1);
+        testIndividualEntries(tlvData3, 4, 8);
+        testIndividualEntries(tlvData4, 8, 4);
+    });
+
+    const testEntries = (tlvData: Buffer, typeSize: TLVNumberSize, lengthSize: TLVNumberSize) => {
+        const tlv = new TLVData(tlvData, typeSize, lengthSize);
+        const entries = tlv.entries();
+        expect(entries).to.have.length(3);
+        expect(entries.get(BigInt(0))).to.be.deep.equal(Buffer.alloc(0));
+        expect(entries.get(BigInt(1))).to.be.deep.equal(Buffer.from([1]));
+        expect(entries.get(BigInt(2))).to.be.deep.equal(Buffer.from([1, 2]));
+        expect(entries.get(BigInt(3))).to.be.undefined;
+    };
+
+    it('should get the entries', () => {
+        testEntries(tlvData1, 1, 2);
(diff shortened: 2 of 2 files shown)


token/js/src/extensions/tlvData.ts
```diff
@@ -0,0 +1,95 @@
...
+        case 1:
+            return constructor(buffer.readUInt8(offset));
+        case 2:
+            return constructor(buffer.readUInt16LE(offset));
+        case 4:
+            return constructor(buffer.readUInt32LE(offset));
+        case 8:
+            return constructor(buffer.readBigUInt64LE(offset));
+    }
+}
```
buffalojoec, line 19: This is a neat function here, but our TLV library is using `u32` for length and Token2022 is using `u16`, so we can probably do away with handling `u64` and eliminate the need for `bigint`. What do you think?

https://github.com/solana-labs/solana-program-library/blob/f35dc5fdb6fc01d054afc51a6ea5a10307a48e3b/token/program-2022/src/extension/mod.rs#L81

https://github.com/solana-labs/solana-program-library/blob/f35dc5fdb6fc01d054afc51a6ea5a10307a48e3b/libraries/type-length-value/src/length.rs#L11

token/js/src/extensions/tlvData.ts
```diff
@@ -0,0 +1,95 @@
...
+            offset += this.typeSize;
+            const entryLength = readTLVNumberSize(this.tlvData, this.lengthSize, offset, Number);
+            offset += this.lengthSize;
+            if (typeMatches && offset + entryLength <= this.tlvData.length) {
+                return this.tlvData.subarray(offset, offset + entryLength);
+            }
+            offset += entryLength;
+        }
+        return null;
+    }
```
buffalojoec, line 72: I see what you're doing here, and this function does, given a type (discriminator), return the value at that TLV entry, however I think making the distinction in the code as specifically a **discriminator** is a worthy inclusion.

Similar to length, we use specific-length arrays for discriminators. Token2022 is using `u16`s converted to bytes, while TLV is using an SPL discriminator. Maybe we can somehow include this? FWIW, I know Token2022 seems to duplicate a lot of code from TLV 😭 but the JS lib doesn't have to!

https://github.com/solana-labs/solana-program-library/blob/f35dc5fdb6fc01d054afc51a6ea5a10307a48e3b/token/program-2022/src/extension/mod.rs#L862-L866

https://github.com/solana-labs/solana-program-library/blob/f35dc5fdb6fc01d054afc51a6ea5a10307a48e3b/libraries/discriminator/src/discriminator.rs#L24-L26

token/js/test/unit/tlvData.test.ts
```diff
@@ -0,0 +1,117 @@
...
+        const type = Buffer.alloc(typeSize);
+        type[0] = 0;
+        expect(tlv.entry(type)).to.be.deep.equal(Buffer.alloc(0));
+        type[0] = 1;
+        expect(tlv.entry(type)).to.be.deep.equal(Buffer.from([1]));
+        type[0] = 2;
+        expect(tlv.entry(type)).to.be.deep.equal(Buffer.from([1, 2]));
+        type[0] = 3;
+        expect(tlv.entry(type)).to.be.null;
+    };
```
buffalojoec, line 92: Am I reading this correctly in that it appears you've combined grabbing by index or by discriminator?

That's pretty interesting. I guess someone would have to know the ordering of the TLV entries, but if they do it's not bad to have. I would sort of lean toward a separate function though, for brevity. Curious your thoughts.

With that being said I wonder if there's a similarly slick way to do grabbing an entry for a certain discriminator (that repeats) at a certain repetition number? That's what this bad boy does. Given a repetition number, say 2, it will return the second TLV entry for that discriminator:

https://github.com/solana-labs/solana-program-library/blob/f35dc5fdb6fc01d054afc51a6ea5a10307a48e3b/libraries/type-length-value/src/state.rs#L243-L249

buffalojoec (commented):
This is definitely heading in the right direction. Thanks again for continuing to contribute here!

I will say I think it's worth matching a little bit more of the behavioral aspects of the Rust libraries, namely:

> Parsing still succeeds if the tlv data is not the right length. entry just returns null for a partial last entry and entries just excludes the last partial entry.

In Rust, this would fail to parse, so we probably want this to happen in JS as well. We might even need to introduce some errors similar to the Rust ones.

https://github.com/solana-labs/solana-program-library/blob/f35dc5fdb6fc01d054afc51a6ea5a10307a48e3b/libraries/type-length-value/src/state.rs#L143-L145

I also think we should try to use similar nomenclature for the class and method names, to avoid confusion, ie: `TlvState`, `get_first_bytes`, etc. Obviously we don't need the separated `mut` API in here, though.

buffalojoec:
I can see that you're going for a more generalized approach here, and I'm not sure this idea I'm going to suggest will be as slick as I imagine it, but what do you think about instead using an interface and/or super class that together mock the `TlvState` trait? Ideally then you'd just supply the sizes and get all of the TLV stuff out of the box?

That's what a lot of [these tests](https://github.com/solana-labs/solana-program-library/blob/f35dc5fdb6fc01d054afc51a6ea5a10307a48e3b/libraries/type-length-value/src/state.rs#L633-L677) are doing; just creating objects that implement `SplDiscriminate` and can then be used as TLV entries.

In JS, I understand we would have to somehow replicate Rust's `size_of` or the `solana_program::borsh::get_instance_packed_len` functions, but that could be roped in here.

buffalojoec:
Oh, I'm just now remembering I have this sloppy code that might be helpful!

https://github.com/buffalojoec/keyring-program/blob/main/js/src/keystore.ts

token/js/src/extensions/tlvData.ts
wjthieme: But the discriminator could still by 8 bytes, right? This function is currently used for both parsing the discriminator and the length

token/js/src/extensions/tlvData.ts
wjthieme: My goal was to make it so generic that you can basically supply anything as a discriminator and this helper class would then figure out how to parse / read that from bytes. I think it might make sense to add a simple version of splDiscriminate to the js library as well (more on that below)

token/js/test/unit/tlvData.test.ts
wjthieme: The `entry` method currently only works by discriminator (not by index). It transforms anything (number, bigint, bytes, etc.) you provide into raw bytes and then matches that to each discriminator it encounters. It might be nice to add grabbing by index but I think that for now this is not (yet) needed?

wjthieme:
> In Rust, this would fail to parse, so we probably want this to happen in JS as well. We might even need to introduce some errors similar to the Rust ones.

Agreed. Let's keep it similar to the rust implementation. Also think it makes sense to keep the naming the same. i.e. `TlvState` instead of `TLVData`, `get_first_bytes` and `get_bytes_with_repetition` and maybe `get_discriminators` which would then replace the existing functions.

> I can see that you're going for a more generalized approach here, and I'm not sure this idea I'm going to suggest will be as slick as I imagine it, but what do you think about instead using an interface and/or super class that together mock the TlvState trait? Ideally then you'd just supply the sizes and get all of the TLV stuff out of the box?

I think it would definitely make sense to add some form of `spl_discriminate` to the js library as well. Mainly because it makes it easier to read and verify the code. i.e. you can just match the discriminator string to the rust code instead of having the take bytes/bigint/number and turning that into raw bytes and then trying to find the matching discriminator.

I think probably the easiest (and fastest) implementation could be a single function that takes a discriminator string and turns it into bytes/bigint/number which you can then throw into the `TlvState` to grab the entry. Something like:

```ts
export function splDiscriminator(key: string, length = 8): Buffer {
    // TODO: make sure this also works in browser without needing to polyfill node crypto/buffer?
    const digest = createHash('sha256').update(key).digest();
    return digest.subarray(0, length);
}
```

token/js/src/extensions/tlvData.ts
wjthieme: Yep that makes sense. This was only added because you can then supply a BigInt or number in the `entry` function as well as a Buffer.

buffalojoec:
Ok I think we managed to fix the CI!

Let me know when you're ready for another review

token/js/src/extensions/tlvData.ts
buffalojoec: You think maybe define SplDiscriminate separately and then let Token2022 or some other lib drive this TLV lib with whatever size discriminator they need?

Ideally we could then re-write this whole file https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/extensions/extensionType.ts

token/js/test/unit/tlvData.test.ts
buffalojoec: The Rust library supports either grabbing the first or grabbing by which ever "repetition count" (for repeating entries), but it's currently not used by anything in SPL.

For that reason I can live without it for now, but it should eventually go in. Doesn't have to be this PR though

buffalojoec:
> I think it would definitely make sense to add some form of `spl_discriminate` to the js library as well. Mainly because it makes it easier to read and verify the code. i.e. you can just match the discriminator string to the rust code instead of having the take bytes/bigint/number and turning that into raw bytes and then trying to find the matching discriminator.
> 
> I think probably the easiest (and fastest) implementation could be a single function that takes a discriminator string and turns it into bytes/bigint/number which you can then throw into the `TlvState` to grab the entry. Something like:
> 
> ```ts
> export function splDiscriminator(key: string, length = 8): Buffer {
>     // TODO: make sure this also works in browser without needing to polyfill node crypto/buffer?
>     const digest = createHash('sha256').update(key).digest();
>     return digest.subarray(0, length);
> }
> ```

Yes, I like this approach! I think that's perfect. The more I think about it, the more I think a more generalized approach as you've written makes sense, because as I mentioned above we want to actually _not_ do what we did in Rust for Token2022 and TLV, and shared all TLV operations across this library.

If Token2022 wasn't being audited, I'd vote to do this in Rust as well using the Rust TLV lib, and modify the Rust `spl-discriminator` library to support whatever length discriminator you want. This may happen in the future, so what you've got for any length discriminator makes sense to me!

buffalojoec:
Can we roll a separate library for this stuff, rather than working it into Token JS?

I think after some of our discussion it makes sense to introduce new ones, and then they can be imported into the metadata library as well as token (#5228).

I think, for now, let's just keep SplDiscriminate and TLV stuff in one JS library - we'll call it `@solana/spl-type-length-value`, and then import it where we need it. What do you think?

wjthieme:
> Can we roll a separate library for this stuff, rather than working it into Token JS?

I think that makes sense. What location in the repository would you suggest for that?

There is the `libraries` folder but it currently only contains rust libraries so wouldn't necessarily say it makes sense to put it there.

I'm leaning towards either a `type-length-value` or `js-type-length-value` in the root or something along the lines of `libraries-js/type-length-value`. What do you think?

buffalojoec:
> There is the `libraries` folder but it currently only contains rust libraries so wouldn't necessarily say it makes sense to put it there.

You're right but that's only because none of those libraries have any JS counterparts 😂

I'd say just stick a `js` folder into `libraries/type-length-value`. Cargo won't even look at it

.github/workflows/pull-request-libraries.yml
```diff
@@ -56,3 +56,21 @@ jobs:
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+      - uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: node-${{ hashFiles('libraries/type-length-value/js/package-lock.json') }}
+          restore-keys: |
+            node-
+      - run: ./ci/js-test-tlv.sh
```
buffalojoec, line 76: Nice! Thanks for snagging CI and dependabot in here too 💪🏼

libraries/type-length-value/js/package.json
```diff
@@ -0,0 +1,68 @@
...
+        "test": "mocha test"
+    },
+    "peerDependencies": {
+        "@solana/web3.js": "^1.47.4"
+    },
+    "dependencies": {
+        "buffer": "^6.0.3"
+    },
+    "devDependencies": {
+        "@solana/web3.js": "^1.47.4",
```
buffalojoec, line 50: We can cut `@solana/web3.js`

libraries/type-length-value/js/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Library with utilities for working with Type-Length-Value structures in js.
+
+## Example usage
+
+```ts
+import { TlvState, SplDiscriminator } from '@solana/spl-type-length-value';
+
+const tlv = new TlvState(tlvData, typeSize, lengthSize);
+const discriminator = new SplDiscriminator("<discriminator-key>");
```
buffalojoec, line 11: nit: the verbage we use in Rust is "discriminator-hash-input"

libraries/type-length-value/js/src/splDiscriminate.ts
```diff
@@ -0,0 +1,17 @@
...
+    public constructor(discriminator: string, length = 8) {
+        this.discriminator = discriminator;
+        this.length = length;
+    }
+
+    public get bytes(): Buffer {
+        const digest = createHash('sha256').update(this.discriminator).digest();
+        return digest.subarray(0, this.length);
+    }
+}
```
buffalojoec, line 17: This looks good, but can you run a manual test to make sure this function creates the same hash as the `solana_program` hash? Yes, they're both sha256 but want to be absolutely sure.

You don't have to commit anything for that though.

libraries/type-length-value/js/src/tlvState.ts
```diff
@@ -0,0 +1,132 @@
...
+                if (type instanceof Buffer) {
+                    return this.tlvData.subarray(offset, offset + this.typeSize).equals(type);
+                }
+                if ('bytes' in type) {
+                    return this.tlvData.subarray(offset, offset + this.typeSize).equals(type.bytes);
+                }
+                throw new TlvInvalidDiscriminatorError();
+            default:
+                throw new TlvInvalidDiscriminatorError();
+        }
```
buffalojoec, line 60: Let's rename this error to be `TlvTypeNotFoundError`

https://github.com/solana-labs/solana-program-library/blob/c2c516f6584f2c685e8e69eed5b3adbbf7fe5121/libraries/type-length-value/src/state.rs#L60-L64

libraries/type-length-value/js/src/errors.ts
```diff
@@ -0,0 +1,16 @@
...
+export abstract class TlvError extends Error {
+    constructor(message?: string) {
+        super(message);
+    }
+}
+
+/** Thrown if the byte length of an tlv buffer doesn't match the expected size */
+export class TlvInvalidAccountSizeError extends TlvError {
+    name = 'TlvInvalidAccountSizeError';
+}
```
buffalojoec, line 11: nit: `TlvInvalidAccountDataError`

https://github.com/solana-labs/solana-program-library/blob/c2c516f6584f2c685e8e69eed5b3adbbf7fe5121/libraries/type-length-value/src/state.rs#L44-L46

libraries/type-length-value/js/src/tlvState.ts
```diff
@@ -0,0 +1,132 @@
+import { TlvInvalidAccountSizeError, TlvInvalidDiscriminatorError } from './errors.js';
+
+export type TlvNumberSize = 1 | 2 | 4 | 8;
+
+export interface TlvDiscriminator {
+    bytes: Buffer;
+}
+
+export type TlvType = Buffer | Uint8Array | number | bigint | TlvDiscriminator;
```
buffalojoec, line 9: I think we can simplify this and provide good readability by just having one type:

```typescript
type Discriminator = Uint8Array;
```

Since `Buffer` is a subtype of `Uint8Array`, it will work as a `Discriminator`, and whenever someone is using an `SplDiscriminator`, they can just provide `.bytes`.

Then we can forgo the switch case later on and just match against the bytes straight up.

libraries/type-length-value/js/src/tlvState.ts
```diff
@@ -0,0 +1,132 @@
...
+                return this.readTlvNumberSize(this.typeSize, offset, BigInt) === type;
+            case 'object':
+                if (type instanceof Buffer) {
+                    return this.tlvData.subarray(offset, offset + this.typeSize).equals(type);
+                }
+                if ('bytes' in type) {
+                    return this.tlvData.subarray(offset, offset + this.typeSize).equals(type.bytes);
+                }
+                throw new TlvInvalidDiscriminatorError();
+            default:
```
buffalojoec, line 58: Referencing my other comment, this collapses to just the one conditional

libraries/type-length-value/js/src/tlvState.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Get a multiple entries from the tlv data. This function returns `count` or less entries with the given type.
+     *
+     * @param type the type of the entry to get
+     * @param count the number of entries to get (0 for all entries)
+     *
+     * @return the entry from the tlv data or null
+     */
+    public bytesRepeating(type: TlvType, count = 0): Buffer[] {
```
buffalojoec, line 83:
```suggestion
    public bytesRepeating(discriminator: Discriminator, count = 0): Buffer[] {
```

libraries/type-length-value/js/src/tlvState.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export interface TlvDiscriminator {
+    bytes: Buffer;
+}
+
+export type TlvType = Buffer | Uint8Array | number | bigint | TlvDiscriminator;
+
+export class TlvState {
+    private readonly tlvData: Buffer;
+    private readonly typeSize: TlvNumberSize;
```
buffalojoec, line 13: We should probably call this `discriminatorLength` and just make it a `number`

libraries/type-length-value/js/src/tlvState.ts
```diff
@@ -0,0 +1,132 @@
...
+        return entries;
+    }
+
+    /**
+     * Get all the discriminators from the tlv data. This function will return a type multiple times if it occurs multiple times in the tlv data.
+     *
+     * @return a list of the discriminators.
+     */
+    public discriminators(): Buffer[] {
+        const types: Buffer[] = [];
```
buffalojoec, line 114: Just so it's easier to follow
```suggestion
        const discriminators: Buffer[] = [];
```

buffalojoec (commented):
This is coming together nicely! I left comments mostly around simplifying the provided discriminator(s). I think if we can strip some of that complexity away, we'll be in good shape.

Thanks again!

libraries/type-length-value/js/src/splDiscriminate.ts
wjthieme: Can add a simple test case with static bytes!

libraries/type-length-value/js/src/tlvState.ts
wjthieme: What about when we want to use `ExtensionType` from spl-token? In js this is a number (which we would then need to explicitly turn into bytes every time we want to use it).

That's why I added the switch + all the supported types in the first place.

Now that I think of it we could just do that in the `getExtensionData(extension: ExtensionType, tlvData: Buffer): Buffer | null` function

buffalojoec: Yep! Exactly. It's already quite similar to what you've got

buffalojoec (approved):
Lgtm. Thanks for putting this together and addressing all feedback!

PR #5186, opened 2023-09-01
token-js: renamed `getExtraAccountMetaAccount` to `getExtraAccountMetaAddress`

wjthieme:
The method is not actually getting any account. Rather it is computing the address for the account. Alongside that I have flipped the parameters around to be consistent with other address getter functions.


joncinque (dismissed):
Thanks for the quick turnaround!

joncinque:
Oh actually, I'm just realizing that this wasn't rebased on master, so you'll need to change the calls in your new test too

joncinque (approved):
And never mind, you did that too, looks good!

joncinque:
Failure is unrelated, merging on red

PR #5218, opened 2023-09-07
bump twoxtx ci rust version to 1.72.0

buffalojoec:
The monorepo is running 1.72 and the setup steps involved in the `twoxtx` CI job are pulling in the latest Solana version, which causes some non-compatibility between crates.

This repairs failing CI job for `twoxtx`

https://github.com/solana-labs/solana-program-library/actions/runs/6111174177/job/16592442873#step:8:756

### DIFF
--- a/.github/workflows/pull-request-token.yml
+++ b/.github/workflows/pull-request-token.yml
@@ -214,7 +214,7 @@ jobs:
 
       - name: Set env vars
         run: |
-          echo "RUST_STABLE_VERSION=1.69.0" >> $GITHUB_ENV
+          echo "RUST_STABLE_VERSION=1.72.0" >> $GITHUB_ENV
           source ci/rust-version.sh
           echo "RUST_STABLE=$rust_stable" >> $GITHUB_ENV
           source ci/solana-version.sh


buffalojoec:
@2501babe is this just failing _some_ of the time? Thus this might not actually be a fix? Saw you merge a lot of those dependabot PRs and many were successful

2501babe:
spl ci is intricately scoped such that the prs i have been merging probably dont trigged the twotx test because they dont change anything it needs to pass for. a lot were just js libs

buffalojoec:
> spl ci is intricately scoped such that the prs i have been merging probably dont trigged the twotx test because they dont change anything it needs to pass for. a lot were just js libs

Ahh makes sense!

2501babe:
im gonna be honest i dont know what to do about the failure here. maybe bumping the rust version on this installs both toolchains into one block device at the same time, which is too much for github because 1.72 just is too large now? google indicates theres no way to allocate it more space...

2501babe:
going to merge this because the failure is unrelated but getting twotx to succeed again will be nice

PR #5230, opened 2023-09-10
[type-length-value]: update README

### LINKED ISSUE #5125: Update the Type Length Value README
#### Problem

Since `spl-type-length-value` was expanded to support duplicate TLV entries and subsequently imposed a `allow_repeating` and `first` API, we need to update the README.

#### Solution

Update the README!

buffalojoec:
Updates the README for `spl-type-length-value` to reflect the addition of the "allow repetition" API.

Closes #5125

### DIFF
--- a/libraries/type-length-value/README.md
+++ b/libraries/type-length-value/README.md
@@ -21,7 +21,7 @@ use {
 struct MyPodValue {
     data: [u8; 32],
 }
-impl SpleDiscriminates for MyPodValue {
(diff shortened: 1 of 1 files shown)


2501babe:
we allow repetition now? does this have any ugly implications for parsing? this is opt-in only right?

2501babe (approved):

buffalojoec:
> does this have any ugly implications for parsing?

Parsing where exactly?

PR #5231, opened 2023-09-11
ci broke

2501babe:
the token job runs out of disk space. pain

PR #5265, opened 2023-09-13
ci: Make space in token test-sbf run

joncinque:
#### Problem

As noted in #5231, CI keeps breaking because of a lack of disk space.

#### Solution

Use the `purge-ubuntu-runner.sh` script in token's `cargo-test-sbf` step.

I added `/imagegeneration` to the script, which I wasn't aware of.


2501babe (approved):

joncinque:
Merging this as is, will address the other failure separately

PR #5268, opened 2023-09-13
Fix docs in confidential transfer account definition.

harsh4786:
Fixed a possible typo inside the confidential extension’s `mod.rs` in the `ConfidentialTransferAccount` definition.


joncinque (approved):
Thanks for fixing this up!

joncinque:
And in the future, please avoid tagging team members. Someone will get to it when they have a moment, and calling one person out makes it less likely that anyone will help.

PR #5285, opened 2023-09-15
stake-pool: Prefund split account during redelegate

joncinque:
#### Problem

During redelegate, the first split account should have the rent-exempt reserve populated. This makes it clearer how many lamports are actually moving to the new validator.

#### Solution

Include the stake pool reserve in redelegate, and withdraw the required rent during redelegate.

### DIFF
--- a/stake-pool/program/src/processor.rs
+++ b/stake-pool/program/src/processor.rs
@@ -1944,18 +1944,22 @@ impl Processor {
                 stake_space,
             )?;
 
-            // pre-fund the rent-exempt reserve from the reserve stake account
-            Self::stake_withdraw(
-                stake_pool_info.key,
-                reserve_stake_info.clone(),
-                withdraw_authority_info.clone(),
-                AUTHORITY_WITHDRAW,
-                stake_pool.stake_withdraw_bump_seed,
-                source_transient_stake_account_info.clone(),
-                clock_info.clone(),
-                stake_history_info.clone(),
-                stake_rent,
-            )?;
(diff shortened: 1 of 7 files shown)


stake-pool/program/src/processor.rs
```diff
@@ -1945,6 +1944,25 @@ impl Processor {
                 stake_space,
             )?;
 
+            // if needed, pre-fund the rent-exempt reserve from the reserve stake
+            let required_lamports_for_rent_exemption =
+                stake_rent.saturating_sub(source_transient_stake_account_info.lamports());
+            if required_lamports_for_rent_exemption >= reserve_stake_info.lamports() {
+                return Err(StakePoolError::ReserveDepleted.into());
+            }
```
2501babe, line 1952: this case is probably not worth caring about but note it fails if the reserve is empty but the transient account is fully funded

joncinque: I see what you mean... the logic is really touchy around here. Note that the reserve shouldn't ever be empty, since it needs to have at least the rent-exemption to exist, so this line shouldn't ever trip unnecessarily.

I put this in because some later tests showed that I was accidentally drawing everything out of the reserve and destroying it, so the intent is to make sure that we don't delete the reserve ever.

How about this then?

```
if required_lamports_for_rent_exemption > 0 {
    if required_lamports_for_rent_exemption >= reserve_stake_info.lamports() {
        return Err(StakePoolError::ReserveDepleted.into());
    }
    Self::stake_withdraw(.....)?; 
}
```

2501babe (approved):

Issue #5228, opened 2023-09-10
Create SPL Token Metadata JS Library


buffalojoec:
With the addition of token metadata to Token2022 and the SPL Token Metadata interface itself to the SPL library, we need to support this cool feature & interface in JavaScript!

The goal here is to lay down a JavaScript library for SPL Token Metadata and then import it into `@solana/spl-token` for Token2022 metadata support.

This issue is for the Token Metadata library `@solana/spl-token-metadata`.

#### Problem
We have no JavaScript support for SPL Token Metadata.

#### Solution
Create a JavaScript library for the SPL Token Metadata Interface!
Dubbed `@solana/spl-token-metadata`

mistersimon:
Hey @buffalojoec, would love to pick this up. 

Trying to wrap my head around the boundary between the functionality in @solana/spl-token vs @solana/spl-token-metadata - any pointers/thoughts?

Is the @solana/spl-token-metadata purely meant to be an interface, and then the functionality implemented in @solana/spl-token?

buffalojoec:
> Is the @solana/spl-token-metadata purely meant to be an interface, and then the functionality implemented in @solana/spl-token?

Yeah, that's it! Think about it like this: other people should be able to use `@solana/spl-token-metadata` to interact with their own on-chain program, if it implements the interface, without needing to use `@solana/spl-token` (conceptually speaking).

mistersimon:
Hey @buffalojoec thanks for the pointers - much appreciated. 

Let me work on this, probably take me a few days to get up to speed - hope to share something early next week. 

I always felt doing an interface in javascript a bit clunky, I think  @solana/spl-token-metadata could either be:
1. Pure typescript interface
2. Javascript class with no implementation - users will need to extend the class with functionality to get it to work. 

Would opt for option 1, unless you got any opinions - doesn't force people using the library to implement a class and haven't been able to find any precedence in the code base

buffalojoec:
> Hey @buffalojoec thanks for the pointers - much appreciated.
> 
> Let me work on this, probably take me a few days to get up to speed - hope to share something early next week.
> 
> I always felt doing an interface in javascript a bit clunky, I think @solana/spl-token-metadata could either be:
> 
> 1. Pure typescript interface
> 2. Javascript class with no implementation - users will need to extend the class with functionality to get it to work.
> 
> Would opt for option 1, unless you got any opinions - doesn't force people using the library to implement a class and haven't been able to find any precedence in the code base

[...]

Issue #5229, opened 2023-09-10
Add Token Metadata Support to `@solana/spl-token`


buffalojoec:
With the addition of token metadata to Token2022 and the SPL Token Metadata interface itself to the SPL library, we need to support this cool feature & interface in JavaScript!

The goal here is to lay down a JavaScript library for SPL Token Metadata and then import it into @solana/spl-token for Token2022 metadata support.

This issue is importing and implementing `@solana/spl-token-metadata`.

Depends on #5228.

Problem
We have no JavaScript support for SPL Token Metadata when working with Token2022 in the `@solana/spl-token` JS library.

Solution
Import and implement `@solana/spl-token-metadata` to provide `@solana/spl-token` with support for Token2022's token metadata.

Depends on #5228.

jedusor1:
This would be super useful indeed. Any news on the progress with this one?

buffalojoec:
> This would be super useful indeed. Any news on the progress with this one?

#5292

mistersimon:
Hey @buffalojoec, can check if i'm understanding what needs to be done correctly:

1. Import `createInstruction` functions from `@solana/spl-token-metadata` and re-export

1. Create `actions` that makes the call to the blockchain for each of the instructions, similar to:
https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/actions/burn.ts

2. Create a function `getTokenMetadata` which calls gets the token instruction, similar to:
https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/state/mint.ts#L75

buffalojoec:
> Hey @buffalojoec, can check if i'm understanding what needs to be done correctly:
> 
> 1. Import `createInstruction` functions from `@solana/spl-token-metadata` and re-export
> 2. Create `actions` that makes the call to the blockchain for each of the instructions, similar to:
>    https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/actions/burn.ts
> 3. Create a function `getTokenMetadata` which calls gets the token instruction, similar to:
>    https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/state/mint.ts#L75

1 & 2 yes, you're correct. For 3, it's actually going to look a bit different.

Since the metadata in Token2022 is a `Mint` extension, we'll want to follow the existing pattern for fetching a mint (using the function you linked), and then unpacking the extension data for `TokenMetadata`. Like this:

[...]

PR #5292, opened 2023-09-17
WIP: SPL Token Metadata JS Library

mistersimon:
JavaScript library for SPL Token Metadata

https://github.com/solana-labs/solana-program-library/issues/5228

Took the base scaffolding from token/js


token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,53 @@
...
+    // The associated mint, used to counter spoofing to be sure that metadata belongs to a particular mint
+    mint: PublicKey;
+    // The longer name of the token
+    name: string;
+    // The shortened symbol for the token
+    symbol: string;
+    /// The URI pointing to richer metadata
+    uri: string;
+    /// Any additional metadata about the token as key-value pairs
+    additionalMetadata: [string, string][];
```
mistersimon, line 19: Is this the right schema for the additional properties?

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,53 @@
+import type { PublicKey } from '@solana/web3.js';
+
+/** The field can be one of the required fields (name, symbol, URI), or a
+ * totally new field denoted by a "key" string. */
+type Field = 'name' | 'symbol' | 'uri' | string;
```
mistersimon, line 5: I followed the rust implementation, but would personally just type it as "string"

token-metadata/js/src/instruction.ts
buffalojoec: This is close! The `updateAuthority` is an optional public key, so typically in Rust we'd represent it with `Option<Pubkey>`, but we're using `OptionalNonZeroPubkey`, which just means for `None` it's going to be a 32-length array of `0`.

So we need to reflect that optional pubkey here somehow. Maybe all we need to do is make the field optional?

token-metadata/js/src/instruction.ts
buffalojoec: It might seem gratuitous to include in the type definition, but it does provide a sense of reminder/security to users. Essentially, we want to make it abundantly clear that `'name' | 'symbol' | 'uri'` are fundamental parts of the interface, while any other key is additional.

This might come in handy if you're typing out a function that will attempt to update one of these keys, or erroneously trying to delete them.

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,53 @@
...
+    updateAuthority: (
+        programId: PublicKey,
+        metadata: PublicKey,
+        oldAuthority: PublicKey,
+        newAuthority: PublicKey
+    ) => Promise<void>;
+
+    // Emits the token-metadata as return data
+    emit: (programId: PublicKey, metadata: PublicKey) => Promise<TokenMetadata>;
+}
```
buffalojoec, line 53: I don't think this actually needs to be an interface itself. I think maybe we should just define a set of functions for creating the interface instructions, similar to the functions that are laid out here:

https://github.com/solana-labs/solana-program-library/blob/d1f06df779bbf52eb7df85300b619ed6d6675776/token-metadata/interface/src/instruction.rs#L225-L319

Think about it like this: The interface itself is for on-chain programs, so the JS library doesn't _itself_ have to be an actual interface. Rather, it's a library designed for working with programs that implement the _Rust interface_.

Lmk if this makes sense!

buffalojoec: You are also missing one: `initialize`.

And we'll need to roll some `pack/unpack` stuff as well using TLV or manually packing the bytes (likely the latter).

buffalojoec (commented):
This is on the right track, thanks!

I left a few comments, and I also have a couple of review thoughts I'll put here:

- I think `TokenMetadata` should be in a separate module from the interface instructions, namely `state.ts`.
- Same with `Field`, just to keep with the Rust layout.
- I also think we need to include the 8-byte discriminator for token metadata, since we're going to need to use the TLV library to read it from some slab of bytes.
- There's a number of implemented helpers and traits on `state::TokenMetadata` that allow consumers to easily interact with serialized metadata. We should provide similar support in the JS library for the interface. We won't have all of the robust Rust TLV goodness in JS (yet), so we'll have to lean on the [JS TLV library](https://github.com/solana-labs/solana-program-library/tree/master/libraries/type-length-value/js) and make any suggestions for furthering its development in order to provide similar helpers for metadata here in this library.

token-metadata/js/src/instruction.ts
mistersimon: Okay follow your rationale, lets stick with how it is. Thanks for the explanation

token-metadata/js/src/instruction.ts
mistersimon: Can i check if the value will ever be `undefined`?

From what i'm understanding from you (and trying to read the rust code), `updateAuthority` is actually always defined, but could be a 32-length array.

If this is true, while contextually i think it makes sense to make it optional field - if it never actually is optional you force to users to check for undefined (to avoid typescript complaining) when its never true. 

We could create our own type like this below, just to serve as more context
```
// If not defined, will be the default public key with 32-length array of 0.
export type OptionalNonZeroPubkey = PublicKey
```

token-metadata/js/src/instruction.ts
mistersimon: Makes sense - functions provides more flexibility for sure. 

I've added `initialize`, but unsure what the return type should be - got any input? Should it be done as a generic? 

I've added the `pack/unpack` interfaces, still need to figure out the implementation so will add later

mistersimon:
Hey @buffalojoec - thanks for the feedback and review. 

I've most of the changes, left some comments to clarify things and have not implemented the pack/unpack functions as i'm still trying to figure out how to do them. 

Do you have any sample data i can create a test case around for the pack/unpack data?

Thanks!

token-metadata/js/src/instruction.ts
buffalojoec: I think making it undefined and then defaulting to the all zeroes is the right move, but the user doesn't really have to know that's happening under the hood, since all they really care about is the fact that the update authority is either the address they've set, or effectively `None`.

token-metadata/js/src/instruction.ts
buffalojoec: Left comments in the code on the types/functions you laid out

token-metadata/js/package.json
```diff
@@ -0,0 +1,70 @@
...
+        "docs": "shx rm -rf docs && typedoc && shx cp .nojekyll docs/",
+        "deploy:docs": "npm run docs && gh-pages --dest token-metadata/js --dist docs --dotfiles"
+    },
+    "peerDependencies": {
+        "@solana/web3.js": "^1.47.4"
+    },
+    "dependencies": {},
+    "devDependencies": {
+        "@solana/spl-memo": "^0.2.2",
+        "@solana/web3.js": "^1.47.4",
```
buffalojoec, line 51: We should make sure we're only using the bare minimum necessary libs

token-metadata/js/README.md
```diff
@@ -0,0 +1,61 @@
+# `@solana/spl-token-metadata`
+
+A TypeScript interface describing the instructions required for a program to implement to be considered a "token-metadata" program for SPL token mints. The interface can be implemented by any program.
+
+## Links
+
+- [TypeScript Docs](https://solana-labs.github.io/solana-program-library/token/js/)
```
buffalojoec, line 7: Gotta update this path

token-metadata/js/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Please ask questions in the Solana Stack Exchange: https://solana.stackexchange.com/
+
+If you've found a bug or you'd like to request a feature, please
+[open an issue](https://github.com/solana-labs/solana-program-library/issues/new).
+
+## Install
+
+```shell
+npm install --save @solana/spl-token-metadata @solana/web3.js
```
buffalojoec, line 24: If the dependency tree is set up right, they shouldn't have to install both. Odds are they probably have `@solana/web3.js` already, and if they don't need it, our lib will bring only what it needs from `@solana/web3.js` when installed.

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,56 @@
...
+
+
+// TODO:- Library should impliment Pack/unpack helpers
+export type Pack = (meta: TokenMetadata) => Promise<Buffer>
+export type Unpack = (input: Buffer) => Promise<TokenMetadata>
+
+export type Initialize = (
+    programId: PublicKey,
+    metadata: PublicKey,
+    updateAuthority: PublicKey,
```
buffalojoec, line 19: this could also be optional

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/*
+Not implimented as a strict interface, by function signatures so that the 
+library can work with programs that implement the Rust interface.
+*/
+
+
+// TODO:- Library should impliment Pack/unpack helpers
+export type Pack = (meta: TokenMetadata) => Promise<Buffer>
+export type Unpack = (input: Buffer) => Promise<TokenMetadata>
```
buffalojoec, line 14: So, you've actually got two pack-unpack operations in this interface. One is for the instructions, and the other is for the state.
So, as a user of this library I'll need to do both:
- Create instructions to send to an SPL Token Metadata program
- Deserialize account data to read SPL Token Metadata from a metadata account

Some of this work has been done already in the JavaScript TLV library, so you should just need to import it and drive it's unpacking functions to do the _state_ part of things.

https://github.com/solana-labs/solana-program-library/tree/master/libraries/type-length-value/js

For instructions, we're just building a `TransactionInstruction` and making sure we build the bytes the same way as we do in the Rust lib, so we get the right instructions!

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/** If the field does not exist on the account, it will be created.
+ * If the field does exist, it will be overwritten. */
+export type UpdateField = (
+    programId: PublicKey,
+    metadata: PublicKey,
+    updateAuthority: PublicKey,
+    field: Field,
+    value: string
+) => Promise<void>;
```
buffalojoec, line 35: I'm not sure I follow what these types are actually defining.
Rather than a type that takes a defined set of parameters and returns `Promise<void>`, why not just have a set of types for the parameters (pretty much the exact types you have without the arrow function or return bits), and then just define functions to create the instructions?

Something like this:

```typescript
export type UpdateField = {
   /* Type spec */
}

export function createUpdateFieldInstruction(args: UpdateField): TransactionInstruction {
   /* Function body */
}
```

buffalojoec (commented):
Thanks for the continued effort. I left some comments hopefully clearing up a few of your questions!

token-metadata/js/README.md
mistersimon: Changed to:
`https://solana-labs.github.io/solana-program-library/token-metadata/js/`

token-metadata/js/package.json
mistersimon: Yes good idea. I've removed "@solana/spl-memo" for now. Will leave this conversation open for now, as a reminder to revisit. 

There are some like "start-server-and-test" i think we can remove, but want to implement the tests first

token-metadata/js/README.md
mistersimon: I don't believe it behaves that way. `@solana/web3.js` is listed as a peer dependency, so it won't get installed when consumers of the library install this package and it will need to be installed. If you already have "@solana/web3.js" installed its safe to install it again - so i think it just serves as a good reminder that its a peer dependency and helps people setting for the first time.

I think this should be safe/preferred and its the same as this package https://github.com/solana-labs/solana-program-library/blob/master/token/js/README.md.

I believe its also standard practices to set it as a peer dependency, so the library version doesn't conflict with installed versions in the consumers code (it can be quite annoying to fix these!). 



Can read more here:
https://www.solutelabs.com/blog/peer-dependencies-in-npm

token-metadata/js/src/instruction.ts
mistersimon: I don't think exporting function definitions in the library will work as intended, theres no way to inherit/extend functions in javascript/typescript. 

The intention with these types was that the consumer would write something like the below in their code, and typescript would know arguments and return type:

```
const createUpdateFieldInstruction: UpdateField = (args) => {
  /* Function body*/
}
```

mistersimon: But you comment raises another question, I assumed each of the arguments were to be passed individually, but your example has it being passed as a object  - not sure if this is intentional

The different is like this:
1. `createUpdateFieldInstruction(0x000, 0x000, .....)`
2. `createUpdateFieldInstruction({ programId: 0x000, metadata:0x000 })`.


I did 1, because thats what I thought the rust code was doing - but i'm not familiar with rust. Do you know which it is?

I think 2 might be better for us. It would allow updateAuthority to be optional (you can't have optional parameters in the middle of an argument list)

token-metadata/js/src/instruction.ts
mistersimon: Thanks for the explanation, its getting a bit clearer to me, but still a bit unsure where to start.

Do you know a good way to get example byte data to build a test case around?

buffalojoec: For the instructions, you'll want to take note of the 8-byte discriminator for each:

https://github.com/solana-labs/solana-program-library/blob/f76276e91df169d045d5c9c1b240c1923bdd72e3/token-metadata/interface/src/instruction.rs#L22

Then it's just unpacking the rest with Borsh:

https://github.com/solana-labs/solana-program-library/blob/f76276e91df169d045d5c9c1b240c1923bdd72e3/token-metadata/interface/src/instruction.rs#L172-L175

For the state, there's one discriminator for token metadata state:

https://github.com/solana-labs/solana-program-library/blob/f76276e91df169d045d5c9c1b240c1923bdd72e3/token-metadata/interface/src/state.rs#L45

And the rest is also unpacked with Borsh:

https://github.com/solana-labs/solana-program-library/blob/f76276e91df169d045d5c9c1b240c1923bdd72e3/token-metadata/interface/src/state.rs#L104

The thing to keep in mind about the Token Metadata state is the fact that it can exist alongside other account data, such as a mint with extensions!

For this reason, it's unpacked with Borsh, but within a Type-Length-Value configuration, meaning the Token Metadata is one entry of some arbitrary possible number of entries, each of which are discriminated using an 8-byte `ArrayDiscriminator`.

As I mentioned, you should be able to use the linked JS library to unpack TLV, and as far as creating bytes for test cases, you might have to roll some buffers manually unfortunately!

token-metadata/js/src/instruction.ts
buffalojoec: > But you comment raises another question, I assumed each of the arguments were to be passed individually, but your example has it being passed as a object - not sure if this is intentional

You're correct, in Rust we're doing individual parameters:

https://github.com/solana-labs/solana-program-library/blob/f76276e91df169d045d5c9c1b240c1923bdd72e3/token-metadata/interface/src/instruction.rs#L225-L236

But it's also not uncommon to pass objects in JS for functions with lots of parameters. I think it can be either one.

buffalojoec: > I don't think exporting function definitions in the library will work as intended, theres no way to inherit/extend functions in javascript/typescript.

I'm not really sure why you would need to inherit anything from this library, maybe I'm missing something?

All this library needs to do, in general, is provide the ability to build instructions and deserialize state for the SPL Token Metadata interface. When working with a program that implements SPL Token Metadata, like Token2022, this is the library you use to interact with it.

token-metadata/js/package.json
```diff
@@ -0,0 +1,79 @@
...
+        "reinstall": "npm run nuke && npm install",
+        "release": "npm run clean && npm run build",
+        "test": "mocha test",
+        "watch": "tsc --build --verbose --watch tsconfig.all.json"
+    },
+    "peerDependencies": {
+        "@solana/web3.js": "^1.47.4"
+    },
+    "dependencies": {
+        "@solana/spl-type-length-value": "link:../../libraries/type-length-value/js",
```
mistersimon, line 50: @solana/spl-type-length-value needs to be published (or made public). 

Change to a normal "package" once done

token-metadata/js/package.json
```diff
@@ -0,0 +1,79 @@
...
+        "borsh": "^1.0.0"
```
mistersimon, line 51: Somewhere else in this repo, `https://www.npmjs.com/package/@coral-xyz/borsh` is being used. 

There wasn't any docs for coral-xyz package, and this packaged looked easier to use + more downloads on npm. 

Any rules surrounding this?

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,166 @@
...
+    const data = { field, value };
+    return new TransactionInstruction({
+        programId,
+        keys: [
+            { isSigner: false, isWritable: true, pubkey: metadata },
+            { isSigner: true, isWritable: false, pubkey: updateAuthority },
+        ],
+        data: Buffer.concat([
+            DISCRIMNATOR.UpdateField,
+            Buffer.from([3]), // TODO explain this. It comes from field being typed as "Field" rather than string
```
mistersimon, line 98: This appears to arise from the fact that field is of type "Field" rather than "string" in the rust code. If i convert it to a string argument, the 3 no longer appears. 
https://github.com/solana-labs/solana-program-library/blob/master/token-metadata/interface/src/instruction.rs#L255

My guess is it is some additional encoding borsh is doing, but can't figure it out. Any insights? 

I think it should be safe, because it will just be treated as a string instead - but not confident.

token-metadata/js/src/instruction.ts
mistersimon: Just want to confirm my understanding/implementation on the optionality of the updateAuthority.

For all instructions updateAuthority is not optional. It could be optional for `updateAuthority`, but i think its better to have users explicitly set the new Authority explicitly as the null address.

When retrieving the TokenMetadata - updateAuthority might the be null address, for these cases i have implimented such that it shows up as undefined, rather than the default/null public key.

token-metadata/js/src/instruction.ts
mistersimon: Thanks for the directions - I think i managed to figure out with your help. 

I've implemented the Instructions such that they return TransactionInstructions with the right data bytes layout.

I managed to cross check against the rust by running the code locally and printing some of the buffers

token-metadata/js/src/instruction.ts
mistersimon: Take a look at the latest code - I think its closer to what you had in mind. 

I think I had the wrong understanding about what the library was suppose to be, but its a bit clearer now and should be understanding your comments correctly now!

token-metadata/js/test/instruction.test.ts
```diff
@@ -0,0 +1,147 @@
...
+                    { isSigner: false, isWritable: true, pubkey: metadata },
+                    { isSigner: false, isWritable: false, pubkey: updateAuthority },
+                    { isSigner: false, isWritable: false, pubkey: mint },
+                    { isSigner: true, isWritable: false, pubkey: mintAuthority },
+                ],
+                data: Buffer.from([
+                    // Output of rust implimentation
+                    210, 225, 30, 162, 88, 184, 77, 141, 13, 0, 0, 0, 77, 121, 32, 116, 101, 115, 116, 32, 116, 111,
+                    107, 101, 110, 4, 0, 0, 0, 84, 69, 83, 84, 16, 0, 0, 0, 104, 116, 116, 112, 58, 47, 47, 116, 101,
+                    115, 116, 46, 116, 101, 115, 116,
```
buffalojoec, line 50: You pulled the bytes from the Rust implementation and matched them up here? Impressive! 🙌🏼

buffalojoec (commented):
This is looking great, thanks for a continued effort! Almost ready to go

I left a few comments, many of them around some of the comments you left. Keep in mind comments are super helpful for downstream users to understand the code, so we want to make sure they are accurate, clean, and make sense!

token-metadata/js/src/instruction.ts
buffalojoec: Actually after checking out the instruction it's apparently not optional when you initialize, so we're good here. It's only optional when you go to update it!

token-metadata/js/package.json
buffalojoec: I've published this package: https://www.npmjs.com/package/@solana/spl-type-length-value

token-metadata/js/package.json
buffalojoec: We _do_ use buffer-layout in `@solana/spl-token` though

https://github.com/solana-labs/solana-program-library/blob/5ce6eace7c128c3eb5686e02de983ee431b73ce1/token/js/src/extensions/transferFee/instructions.ts#L39-L49

token-metadata/js/test/instruction.test.ts
mistersimon: Yeah - Thought it was the best way to make sure the code was working correct!

token-metadata/js/src/errors.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/** Thrown if Token metadata has no update authority */
+export class ImmutableMetadataError extends TokenMetadataError {
+    name = 'ImmutableMetadataError';
+}
+
+/** Thrown if Key not found in metadata account */
+export class KeyNotFoundError extends TokenMetadataError {
+    name = 'KeyNotFoundError';
+}
```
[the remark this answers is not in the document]
mistersimon: Added - Thanks for spotting :)

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,166 @@
...
+            { isSigner: false, isWritable: false, pubkey: mint },
+            { isSigner: true, isWritable: false, pubkey: mintAuthority },
+        ],
+        data: Buffer.concat([DISCRIMNATOR.Initialize, serialize(SCHEMA.Initialize, data)]),
+    });
+}
+
+/** If the field does not exist on the account, it will be created.
+ * If the field does exist, it will be overwritten. */
+interface UpdateFeild {
```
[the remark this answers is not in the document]
mistersimon: Fixed - Added spellcheck to my editor as well to find others

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,166 @@
+import { PublicKey, TransactionInstruction } from '@solana/web3.js';
+import { splDiscriminate } from '@solana/spl-type-length-value';
+import { serialize } from 'borsh';
+
+import { Field } from './state';
+
+// Taken from https://github.com/solana-labs/solana-program-library/blob/master/token-metadata/interface/src/instruction.rs
+export const DISCRIMNATOR = {
```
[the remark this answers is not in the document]
mistersimon: Renamed

token-metadata/js/package.json
mistersimon: Awesome thank you. Switched to the npm package

token-metadata/js/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './errors.js';
+export * from './instruction.js';
+export * from './state.js';
```
buffalojoec, line 3: Shouldn't these be `*.ts`?

mistersimon: Yes I agree its a bit weird, but i think its a quirk of the current prettier setup. I would usually not put the `.js/.ts` extension - but the prettier config forces it.  

If we do `.ts`, it raises the following typescript error:
`An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.ts`

For what its worth its the same in this file:
https://github.com/solana-labs/solana-program-library/blob/5ce6eace7c128c3eb5686e02de983ee431b73ce1/token/js/src/index.ts

token-metadata/js/src/instruction.ts
mistersimon: Yeah you are right. Didn't realise that `Field::Key("Name".to_string())` is different from `Field::Name` (still a rust noobie)

Managed to extract the buffer when you use `Name` into a test case (but its failing).
https://github.com/solana-labs/solana-program-library/pull/5292/commits/30ed1be7bc1d5646f19eeedf859263447c803e8c

mistersimon: I haven't successfully serialised this yet - but i'll take this in another thread.

But it raises the question as to what should be matched against the rust enum "Name" etc...

I think there are two solutions, using an enum or string matching

**1. Enum**
Define a typescript enum, and encode as the enum only if the enum is provided
```
enum RequiredField {
  Name = 0 ,
  Symbol = 1,
  Uri = 2,
}
type Field = RequiredField | "string"
```

**2. string matching**
```
   if(field == "Name" || field == "name" ) {
    // encode as enum Field::name
  } else {
     // encode as user defined key
  }

```

Solution 1 more closely matches the rust implementation and allows there to both keys `Field::Name` and `Field::Key("Name".to_string())` (which i think can co-exist based on my limited understanding?). However Enums is a typescript feature only, and any javascript users would be out of like. I think they would need to pass in 0 as the field to get the correct behaviour'



I'm more inclined to go with Solution 1 - if its really true that `Field::Name` and `Field::Key("Name".to_string())` can co-exist

token-metadata/js/package.json
mistersimon: I tried to switch to `@solana/buffer-layouts` but two main notes:

1. `string` support is a bit clunky. I have to manually calculate the buffer length. The string length is variable, and doesn't seem like the library can handle it and always expect a buffer first. Its possible to over-allocate and trim after to what was used, but seems less efficient. Also seem to have to manually add the string length as a u32 before the data. (I also tried the utf8 type in the buffer-layouts, but didn't solve the above problems and kept giving me typescripts errors)
```
    const layout = struct<RemoveKeyData>([bool('idempotent'), u32('fieldLength'), cstr('field')]);

    const fieldLength = Buffer.byteLength(field, 'utf8');

    const buffer = Buffer.alloc(1 + 4 + fieldLength);
```

2. There doesn't seem to be a way to encode optional parameters in `@solana/buffer-layouts` such as whats needed for the `Emit` instruction.

mistersimon: Also I don't think the borsh library can encode the "Field" correctly, tried for a while but can't handle it. It can do enums, just not enums with a dynamic `Key(string)` value.

Not sure if you have encouraged similar problems before @buffalojoec ? 

I have 2 potential solutions:

1. Stick with `@solana/buffer-layouts` - but its going to be messy. Think i will have to manually handle the optional parameters and the `Field` parameter.

2. Try this alternative package - Seems to work similar to the existing `borsh` package, but we can define a custom encoder for the `Field` in `UpdateField`
https://www.npmjs.com/package/@dao-xyz/borsh


I'm leaning towards doing 1, avoid having to use another package and we still have to right custom serialiser/deserialiser - but thought I check before I head down this path.

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,166 @@
...
+};
+
+// Order of keys matters
+export const SCHEMA = {
+    Initialize: { struct: { name: 'string', symbol: 'string', uri: 'string' } },
+    UpdateField: { struct: { field: 'string', value: 'string' } },
+    RemoveKey: { struct: { idempotent: 'bool', key: 'string' } },
+    UpdateAuthorithy: { struct: { newAuthority: { array: { type: 'u8', len: 32 } } } },
+    Emit: { struct: { start: { option: 'u64' }, end: { option: 'u64' } } },
+};
```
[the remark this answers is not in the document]
mistersimon: This looks better - I've refactored to what you did in the playground, just changed to a generic type to avoid typing as `any`

token-metadata/js/src/instruction.ts
```diff
@@ -0,0 +1,205 @@
...
+        data: UpdateAuthority.pack(values),
+    });
+}
+
+// Emits the token-metadata as return data
+export interface Emit {
+    programId: PublicKey;
+    metadata: PublicKey;
+    start?: bigint;
+    end?: bigint;
```
mistersimon, line 195: Just want to highlight that i changed the type here to `bigint`. 

Realised `number` is smaller than `u64` it is being packed to

token-metadata/js/src/instruction.ts
mistersimon: Okay - lets go with object parameters then :)

mistersimon:
Hey @buffalojoec thanks for your review and comments - really helpful. 

I think the only real outstanding issue is serialisation/deserialisation code

token-metadata/js/src/instruction.ts
buffalojoec: Yeah perfect!

token-metadata/js/src/index.ts
buffalojoec: Weird, but ok!

buffalojoec:
Do you mind adding this library to CI? Similar to Token JS and Type-Length-Value JS

mistersimon:
> Do you mind adding this library to CI? Similar to Token JS and Type-Length-Value JS

That should be working now, but need your help to trigger it! Don't have the required permissions

buffalojoec:
Hey @mistersimon thanks for the awesome work on this library.

I pushed up a commit to integrate `@solana/codecs-*` - which are a few experimental libraries we're developing in-house at Labs for serializing Rust counterparts in JavaScript. This should solve your concerns with the serialization.

Take a look at my changes and let me know what you think!

.github/workflows/pull-request-token-metadata.yml
```diff
@@ -67,3 +67,26 @@ jobs:
...
+        with:
+          path: ~/.npm
+          key: node-${{ hashFiles('token-metadata/js/package-lock.json') }}
+          restore-keys: |
+            node-
+      - name: Download programs
+        uses: actions/download-artifact@v2
+        with:
+          name: token-lending-programs
+          path: target/deploy
```
buffalojoec, line 91: This is causing CI build problems.

token-metadata/js/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "dependencies": {
+        "@solana/codecs-core": "2.0.0-experimental.398c396",
+        "@solana/codecs-data-structures": "2.0.0-experimental.398c396",
+        "@solana/codecs-numbers": "2.0.0-experimental.398c396",
+        "@solana/codecs-strings": "2.0.0-experimental.398c396",
+        "@solana/options": "2.0.0-experimental.398c396",
+        "@solana/spl-type-length-value": "0.1.0"
+    },
+    "devDependencies": {
+        "@solana/web3.js": "^1.47.4",
```
buffalojoec, line 58: Is this supposed to be _both_ a `devDependency` and `peerDependency`?

token-metadata/js/src/index.ts
```diff
@@ -0,0 +1,6 @@
+// The .js extension is enforced by prettier (and it works).
+// Using .ts raises typescript errors
```
buffalojoec, line 2: We can get rid of this comment.

token-metadata/js/test/instruction.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    createRemoveKeyInstruction,
+    createUpdateAuthorityInstruction,
+    createUpdateFieldInstruction,
+} from '../src';
+
+// Buffer outputs are verified against rust implementation
+// To print packed values in rust file:
+// Add `use std::fmt::Write;` to the top of instructions.rs and
+// use `println!("{:?}", &packed);` inside check_pack_unpack to print buffer
+// Run with `cargo test -- --nocapture`
```
buffalojoec, line 16: We don't need this comment either.

buffalojoec (commented):
This is super close to being done! Just have some cleanup to do and have to get CI working, then we should be cruisin'!

token-metadata/js/src/index.ts
```diff
@@ -0,0 +1,5 @@
+// The .js extension is enforced by prettier (and it works). 
+// Using .ts raises typescript errors
```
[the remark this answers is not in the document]
mistersimon: Removed

token-metadata/js/src/index.ts
mistersimon: Removed

token-metadata/js/test/instruction.test.ts
mistersimon: Removed

token-metadata/js/package.json
mistersimon: Yes - I believe so. 

It needs to be a `devDependency ` so it exists when you running tests inside library, but when the library it is imported, it will be a `peerDependency` and use the version that already exists. 

Its the same setup as here as well:
https://github.com/solana-labs/solana-program-library/blob/master/token/js/package.json

.github/workflows/pull-request-token-metadata.yml
mistersimon: Removed, thanks for highlighting.

Hopefully fixes the issues - will need your help to trigger

buffalojoec: Looks like it worked!

buffalojoec:
> Not sure if this change was intentional on your part - happy to make it

Yeah it was intentional - I didn't really think exporting them was necessary. But if you think it's more useful to downstream users go for it! Then I think we're good to go.

mistersimon:
> > Not sure if this change was intentional on your part - happy to make it
> 
> Yeah it was intentional - I didn't really think exporting them was necessary. But if you think it's more useful to downstream users go for it! Then I think we're good to go.

@buffalojoec made that change. Agreed, think we are good to go! 

Do you take it from here to get it merged + deployed or is there anything else I need to do?

buffalojoec (approved):
Awesome work, thanks again!!

If you feel like also tackling #5229 as well, using your new library, let me know!

I'll publish this package later today.

PR #5323, opened 2023-09-20
Access epoch_schedule by reference

apfitzge:
- https://github.com/solana-labs/solana/pull/32767 removes `Copy` from `EpochSchedule` and `Rent` because they are inconsistent with `CloneZeroed`
- A test in `huge_pool.rs` was `Copy`ing an `EpochSchedule`. Changed to just reference it.


joncinque (approved):
Looks great, thanks!

PR #5324, opened 2023-09-20
Bump repo to 1.16.13

joncinque:
#### Problem

We need features in later Solana releases than 1.16.3, namely the ability to CPI into the zero-knowledge proof program.

#### Solution

Bump the whole repo to use 1.16.13! The first commit is just the result of running `update-solana-dependencies.sh` with a `cargo check` to update the lockfile.

After that, I went through and updated all instances of deprecated functions.

Note that this might not pass CI right away.


samkim-crypto (approved):
Thanks for the updates!

PR #5326, opened 2023-09-21
token-cli: Support offline signing for update-metadata

### LINKED ISSUE #5291: token-cli: update-metadata doesn't support offline signing

joncinque:
#### Problem

Update-metadata doesn't support offline signing in the token CLI.

#### Solution

Thankfully it's pretty simple -- just add the offline args!

Fixes #5291


joncinque:
The audit failure was fixed in #5327 -- merging

PR #5329, opened 2023-09-21
Remove EpochSchedule Copy

apfitzge:
Another fix for https://github.com/solana-labs/solana/pull/32767


joncinque (approved):
Thanks for fixing downstream!

PR #5335, opened 2023-09-24
[token-cli] Add confidential transfer mint commands

samkim-crypto:
#### Problem
There are no cli support for confidential transfers yet.

#### Summary of changes
Updated/added cli commands regarding the mint. The `create-token` with `--enable-confidential-transfers` flag should already be supported.

[f616231](https://github.com/solana-labs/solana-program-library/pull/5335/commits/f616231da8b821759220b30072a01de9ba7605fc): Updated the `display` command to support the confidential transfer mint extension. Unfortunately, the `solana-account-decoder` is still using the older version of token-2022 before the confidential transfer and confidential transfer fee extensions were split, which means we can't use the `UiConfidentialTransferMint` type from `solana-account-decoder`. I hard-coded a temporary `UiConfidentialTransferMint` type in `output.rs` to account for this. This part can be removed in the next monorepo upgrade.

[09ad0d9](https://github.com/solana-labs/solana-program-library/pull/5335/commits/09ad0d9075ce04a0a23eeb6420b3701426cc5c55): ElGamal keys are [supported](https://github.com/solana-labs/solana/blob/master/clap-v3-utils/src/keypair.rs#L1049) in `solana-clap-v3-utils`, but an upgrade to clap-v3 for token-cli is blocked until 1.17 due to a new panicking behavior that was introduced with clap-v3 (https://github.com/solana-labs/solana/pull/33184). So I added temporary functions to read ElGamal keypairs and pubkeys to unblock work on the token-cli. These functions only support either reading from ElGamal keypair files or reading form base64-encoded pubkey string. Once we upgrade to `solana-clap-v3-utils`, we can support a more variety of ways to take ElGamal arguments.

I didn't write tests for these hard-coded functions since they will eventually be replaced anyway, but I can add them in if suggested.

[7e52fc5](https://github.com/solana-labs/solana-program-library/pull/5335/commits/7e52fc5ca4b7fa7801509f3bd7f4c083d06395ff): Added `update-confidential-transfer-settings`, which allows updating the confidential transfer mint configuration. 

To try out the new commands:
```
> cargo run -- create-token --enable-confidential-transfers auto -p TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb

Address:  [MINT-ADDRESS]
Decimals:  9

... the auditing feature is set to none by default
```

```
> cargo run -- display [MINT-ADDRESS]
... will display the confidential transfer mint settings

> cargo run -- update-confidential-transfer-settings [MINT-ADDRESS] manual 5DcMEnLGCFYe2qChv9uexI8tuM6sDJlrbe+tMeIf+UA=
... will update the confidential transfer mint settings to manual approve and auditor key

> cargo run --display [MINT-ADDRESS]
... verify that the update was done correctly
```

To generate an arbitrary ElGamal public key, I just did
```rust
let elgamal_keypair = ElGamalKeypair::new_rand();
println!("{}", elgamal_keypair.pubkey());
// this prints base64 encoding of ElGamal public key like `5DcMEnLGCFYe2qChv9uexI8tuM6sDJlrbe+tMeIf+UA=`
```

Once the issues in `solana-clap-v3-utils` are resolved and we publish `solana-zk-keygen`, then we can use that instead to generate ElGamal keypairs/pubkeys.

### DIFF
--- a/token/cli/Cargo.toml
+++ b/token/cli/Cargo.toml
@@ -12,6 +12,7 @@ version = "3.1.1"
 walkdir = "2"
 
 [dependencies]
+base64 = "0.21.4"
 clap = "2.33.3"
 console = "0.15.7"
 serde = "1.0.188"
@@ -26,12 +27,20 @@ solana-logger = "=1.16.13"
 solana-remote-wallet = "=1.16.13"
 solana-sdk = "=1.16.13"
 solana-transaction-status = "=1.16.13"
-spl-token = { version = "4.0", path="../program", features = [ "no-entrypoint" ] }
-spl-token-2022 = { version = "0.9", path="../program-2022", features = [ "no-entrypoint" ] }
-spl-token-client = { version = "0.7", path="../client" }
-spl-token-metadata-interface = { version = "0.2", path="../../token-metadata/interface" }
-spl-associated-token-account = { version = "2.0", path="../../associated-token-account/program", features = [ "no-entrypoint" ] }
-spl-memo = { version = "4.0.0", path="../../memo/program", features = ["no-entrypoint"] }
+spl-token = { version = "4.0", path = "../program", features = [
+  "no-entrypoint",
+] }
+spl-token-2022 = { version = "0.9", path = "../program-2022", features = [
+  "no-entrypoint",
+] }
+spl-token-client = { version = "0.7", path = "../client" }
+spl-token-metadata-interface = { version = "0.2", path = "../../token-metadata/interface" }
+spl-associated-token-account = { version = "2.0", path = "../../associated-token-account/program", features = [
+  "no-entrypoint",
+] }
+spl-memo = { version = "4.0.0", path = "../../memo/program", features = [
+  "no-entrypoint",
+] }
 strum = "0.25"
 strum_macros = "0.25"
 tokio = "1.14"
--- a/token/cli/src/encryption_keypair.rs
+++ b/token/cli/src/encryption_keypair.rs
@@ -1,0 +1,55 @@
+use {
+    base64::{prelude::BASE64_STANDARD, Engine},
+    clap::ArgMatches,
+    spl_token_2022::solana_zk_token_sdk::{
+        encryption::elgamal::{ElGamalKeypair, ElGamalPubkey},
+        zk_token_elgamal::pod::ElGamalPubkey as PodElGamalPubkey,
+    },
+};
+
+const ELGAMAL_PUBKEY_MAX_BASE64_LEN: usize = 44;
+
+pub(crate) fn elgamal_pubkey_or_none(
+    matches: &ArgMatches,
+    name: &str,
+) -> Result<Option<PodElGamalPubkey>, String> {
+    let arg_str = matches.value_of(name).unwrap();
+    if arg_str == "none" {
+        return Ok(None);
+    }
+    elgamal_pubkey_of(matches, name).map(|pubkey| Some(pubkey))
+}
+
+pub(crate) fn elgamal_pubkey_of(
+    matches: &ArgMatches,
+    name: &str,
+) -> Result<PodElGamalPubkey, String> {
+    if let Ok(keypair) = elgamal_keypair_of(matches, name) {
+        let elgamal_pubkey = (*keypair.pubkey()).into();
+        Ok(elgamal_pubkey)
+    } else {
+        let arg_str = matches.value_of(name).unwrap();
+        if let Some(pubkey) = elgamal_pubkey_from_str(arg_str) {
+            Ok(pubkey)
+        } else {
+            Err("failed to read ElGamal pubkey".to_string())
+        }
+    }
+}
+
+pub(crate) fn elgamal_keypair_of(
+    matches: &ArgMatches,
+    name: &str,
+) -> Result<ElGamalKeypair, String> {
+    let path = matches.value_of(name).unwrap();
+    ElGamalKeypair::read_json_file(path).map_err(|e| e.to_string())
+}
+
+fn elgamal_pubkey_from_str(s: &str) -> Option<PodElGamalPubkey> {
+    if s.len() > ELGAMAL_PUBKEY_MAX_BASE64_LEN {
+        return None;
+    }
+    let pubkey_vec = BASE64_STANDARD.decode(s).ok()?;
+    let elgamal_pubkey = ElGamalPubkey::from_bytes(&pubkey_vec)?;
+    Some(elgamal_pubkey.into())
+}
--- a/token/cli/src/encryption_keypair.rs
+++ b/token/cli/src/encryption_keypair.rs
@@ -13,15 +13,31 @@ use {
 
 const ELGAMAL_PUBKEY_MAX_BASE64_LEN: usize = 44;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum ElGamalPubkeyOrNone {
+    ElGamalPubkey(PodElGamalPubkey),
+    None,
+}
+
+impl Into<Option<PodElGamalPubkey>> for ElGamalPubkeyOrNone {
+    fn into(self) -> Option<PodElGamalPubkey> {
+        match self {
+            ElGamalPubkeyOrNone::ElGamalPubkey(pubkey) => Some(pubkey),
+            ElGamalPubkeyOrNone::None => None,
+        }
+    }
+}
+
 pub(crate) fn elgamal_pubkey_or_none(
     matches: &ArgMatches,
     name: &str,
-) -> Result<Option<PodElGamalPubkey>, String> {
+) -> Result<ElGamalPubkeyOrNone, String> {
     let arg_str = matches.value_of(name).unwrap();
     if arg_str == "none" {
-        return Ok(None);
+        return Ok(ElGamalPubkeyOrNone::None);
     }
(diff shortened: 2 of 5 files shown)


token/cli/src/output.rs
```diff
@@ -288,6 +297,7 @@ pub(crate) struct CliMint {
     pub(crate) epoch: u64,
     #[serde(flatten)]
     pub(crate) mint: UiMint,
+    pub(crate) ui_confidential_transfer_extension: Option<UiConfidentialTransferExtension>,
```
joncinque, line 300: nit: Can you put a note to remove it here too?

token/cli/src/output.rs
```diff
@@ -683,21 +699,47 @@ fn display_ui_extension(
...
+                        )?;
+                        writeln!(
+                            f,
+                            "    {}: {}",
+                            style("Audit key:").bold(),
+                            if let Some(auditor_pubkey) = ui_mint.auditor_encryption_pubkey.as_ref()
+                            {
+                                auditor_pubkey
+                            } else {
+                                "audits are not enabled"
```
joncinque, line 730: nit: should we just say "disabled" here?

token/cli/src/output.rs
```diff
@@ -708,3 +750,39 @@ fn flattened<S: Serializer>(
...
+                        .auto_approve_new_accounts
+                        .into(),
+                    auditor_encryption_pubkey: auditor_encryption_pubkey
+                        .map(|pubkey| pubkey.to_string()),
+                },
+            ));
+        }
+    }
+    None
+}
```
joncinque: This should all get removed too, right? If so, please put a comment about it

token/cli/Cargo.toml
```diff
@@ -26,12 +27,20 @@ solana-logger = "=1.16.13"
...
+  "no-entrypoint",
+] }
+spl-token-client = { version = "0.6", path = "../client" }
+spl-token-metadata-interface = { version = "0.2", path = "../../token-metadata/interface" }
+spl-associated-token-account = { version = "2.0", path = "../../associated-token-account/program", features = [
+  "no-entrypoint",
+] }
+spl-memo = { version = "4.0.0", path = "../../memo/program", features = [
+  "no-entrypoint",
+] }
```
joncinque, line 43: Did your formatter pick these up? I haven't seen this before!

token/cli/src/encryption_keypair.rs
joncinque, line 1: Can you add a comment here too to remove and use the proper parsers from `clap-v3-utils`?

token/cli/src/main.rs
```diff
@@ -7644,6 +7822,38 @@ mod tests {
...
+        // update confidential transfer mint settings
+        let auditor_keypair = ElGamalKeypair::new_rand();
+        let auditor_pubkey: ElGamalPubkey = (*auditor_keypair.pubkey()).into();
+        let new_auto_approve = false;
+
+        command_update_confidential_transfer_settings(
+            &config,
+            token_pubkey,
+            confidential_transfer_mint_authority,
+            false,                // auto approve
```
joncinque, line 7834: Micro-nit: 
```suggestion
            new_auto_approve,
```

token/cli/src/main.rs
```diff
@@ -4252,6 +4345,61 @@ fn app<'a, 'b>(
...
+                )
+                .arg(
+                    Arg::with_name("approve_policy")
+                        .long("approve-policy")
+                        .value_name("APPROVE_POLICY")
+                        .takes_value(true)
+                        .index(2)
+                        .possible_values(&["auto", "manual"])
+                        .help(
+                            "Enable accounts to make confidential transfers. If \"auto\" \
```
joncinque, line 4368: micro-nit
```suggestion
                            "Policy for enabling accounts to make confidential transfers. If \"auto\" \
```

token/cli/src/main.rs
```diff
@@ -4252,6 +4345,61 @@ fn app<'a, 'b>(
...
+                        .index(1)
+                        .required(true)
+                        .help("The address of the token mint to update confidential transfer configuration for")
+                )
+                .arg(
+                    Arg::with_name("approve_policy")
+                        .long("approve-policy")
+                        .value_name("APPROVE_POLICY")
+                        .takes_value(true)
+                        .index(2)
```
joncinque, line 4365: Rather than setting these by index, which can be annoying if you just want to update one parameter, how about making them optional? That is, unless `sign-only` is present, in which case you'll always need it

token/cli/src/main.rs
```diff
@@ -4252,6 +4345,61 @@ fn app<'a, 'b>(
...
+                            confidential transfer mint authority must approve each account \
+                            before it can make confidential transfers."
+                        )
+                )
+                .arg(
+                    Arg::with_name("auditor_pubkey")
+                        .long("auditor-pubkey")
+                        .value_name("AUDITOR_PUBKEY")
+                        .takes_value(true)
+                        .index(3)
```
joncinque, line 4380: Same here, how about making it optional? Then you only set the specified params. Unless `sign-only` is present, of course

token/cli/src/main.rs
```diff
@@ -4252,6 +4345,61 @@ fn app<'a, 'b>(
...
+                    Arg::with_name("confidential_transfer_authority")
+                        .long("confidential-transfer-authority")
+                        .validator(is_valid_signer)
+                        .value_name("SIGNER")
+                        .takes_value(true)
+                        .help(
+                            "Specify the confidential transfer authority keypair. \
+                            Defaults to the client keypair address."
+                        )
+                )
```
joncinque, line 4401: You've done the sign-only work already, so how about adding
```suggestion
                )
                .nonce_args(true)
                .offline_args(),
```

token/cli/src/main.rs
```diff
@@ -2738,6 +2747,90 @@ async fn command_withdraw_withheld_tokens(
...
+    if let Some(auditor_pubkey) = auditor_pubkey {
+        println_display(
+            config,
+            format!(
+                "  auditor encryption pubkey set to {}",
+                auditor_pubkey.to_string(),
+            ),
+        );
+    } else {
+        println_display(config, format!("  auditability set to disabled",))
```
joncinque, line 2815: nit:
```suggestion
        println_display(config, format!("  auditability disabled",))
```

or maybe just `auditor set to none` is fine?

joncinque (commented):
Looks great! Mostly tiny nits

token/cli/Cargo.toml
samkim-crypto: Yeah I recently updated my neovim config (I now use [lazyvim](https://www.lazyvim.org/)) to enable lsp-server with [taplo](https://taplo.tamasfe.dev/) and it automatically validates cargo files as in an IDE, which is pretty cool. I also enabled auto-[format](https://taplo.tamasfe.dev/cli/usage/formatting.html) on save as well. I guess should have filtered these format changes since it is orthogonal to this PR, sorry 😅.

token/cli/src/output.rs
samkim-crypto: Yep, good idea!

token/cli/src/main.rs
samkim-crypto: Oops! Thanks!

token/cli/src/main.rs
samkim-crypto: Oh yes, we should add that as well!

token/cli/src/main.rs
```diff
@@ -2738,6 +2748,120 @@ async fn command_withdraw_withheld_tokens(
...
+                    token_pubkey,
+                    expected_authority
+                        .map(|pubkey| pubkey.to_string())
+                        .unwrap_or_else(|| "disabled".to_string()),
+                    authority
+                )
+                .into());
+            }
+
+            let current_auto_approve = if let Some(auto_approve) = auto_approve {
```
joncinque, line 2783: micro-nit for consistency between the branches, and for what this actually encodes:
```suggestion
            let new_auto_approve = if let Some(auto_approve) = auto_approve {
```

token/cli/src/main.rs
```diff
@@ -2738,6 +2748,120 @@ async fn command_withdraw_withheld_tokens(
...
+                auto_approve
+            } else {
+                bool::from(confidential_transfer_mint.auto_approve_new_accounts)
+            };
+
+            let current_auditor_pubkey = if let Some(auditor_pubkey) = auditor_pubkey {
```
joncinque, line 2789: micro-nit: same here
```suggestion
            let new_auditor_pubkey = if let Some(auditor_pubkey) = auditor_pubkey {
```

token/cli/src/main.rs
```diff
@@ -2738,6 +2748,120 @@ async fn command_withdraw_withheld_tokens(
...
+                auditor_pubkey.into()
+            } else {
+                Option::<ElGamalPubkey>::from(confidential_transfer_mint.auditor_elgamal_pubkey)
+            };
+
+            (current_auto_approve, current_auditor_pubkey)
```
joncinque, line 2795: Adding this one too so you can just commit them as a batch and not have to touch an editor
```suggestion
            (new_auto_approve, new_auditor_pubkey)
```

joncinque (approved):
Just one last nit, but otherwise looks great!

token/cli/src/main.rs
samkim-crypto: Thanks for the catch!

PR #5367, opened 2023-09-26
release: Bump tlv-account-resolution and dependents

joncinque:
#### Problem

We want to use a new version of spl-token-2022, but its version on master depends on spl-tlv-account-resolution, which made a breaking change to support deriving PDAs from account data.

#### Solution

Bump tlv-account-resolution and everything that depends on it to a new minor version. The idea was to do a minor version bump on things that we want to release, and things that would break because of the change.

### DIFF
--- a/token-metadata/example/Cargo.toml
+++ b/token-metadata/example/Cargo.toml
@@ -13,7 +13,7 @@ test-sbf = []
 
 [dependencies]
 solana-program = "1.16.13"
-spl-token-2022 = { version = "0.8", path = "../../token/program-2022" }
+spl-token-2022 = { version = "0.9", path = "../../token/program-2022" }
 spl-token-metadata-interface = { version = "0.2.0", path = "../interface" }
 spl-type-length-value = { version = "0.3.0" , path = "../../libraries/type-length-value" }
 spl-pod = { version = "0.1.0", path = "../../libraries/pod" }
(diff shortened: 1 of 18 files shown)


token-metadata/example/Cargo.toml
```diff
@@ -13,15 +13,15 @@ test-sbf = []
...
 spl-pod = { version = "0.1.0", path = "../../libraries/pod" }
 
 [dev-dependencies]
 solana-program-test = "1.16.13"
 solana-sdk = "1.16.13"
-spl-token-client = { version = "0.6", path = "../../token/client" }
+spl-token-client = { version = "0.7", path = "../../token/client" }
 test-case = "3.2"
 
 [lib]
```
buffalojoec, line 27: Do we want to bump the minor version of the Token Metadata example considering these dependency changes?

buffalojoec: I guess I don't see you doing that in a few other libs, maybe not necessary? Like `stake-pool` and the swap/upgrade/wrap programs.

buffalojoec (commented):
Whew! Just asking for clarification on the minor bump/release plan(s)

token-metadata/example/Cargo.toml
buffalojoec: Ok sgtm!

buffalojoec (approved):
🎉

PR #5372, opened 2023-09-26
token-2022: Fix build with zk-ops disabled

joncinque:
#### Problem

Token-2022 currently doesn't build with the zk-ops feature disabled.

#### Solution

Mark a function as requiring `zk-ops` to get it to work.


samkim-crypto (approved):
Oops, sorry I missed this!

PR #5375, opened 2023-09-27
macro: Remove dependency on solana-program

joncinque:
#### Problem

While updating the monorepo to use the new spl crates, like token-2022, I got a really strange build error:

```
warning: output filename collision.
The lib target `solana-program` in package `solana-program v1.17.0 (/home/jon/src/solana/solana/sdk/program)` has the same output filename as the lib target `solana-program` in package `solana-program v1.17.0 (/home/jon/src/solana/solana/sdk/program)`.
Colliding filename is: /home/jon/src/solana/solana/target/debug/deps/libsolana_program.so
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
```

After the warning, it complained that `solana_sdk` didn't exist.

I dug around for a bit, and narrowed it down to the usage of `solana_program` in the new derive crates. Since the derive crates are `proc-macro = true`, cargo treats it and its dependencies as completely different libraries, which meant that we were building `solana-program` twice: once in the proc-macro context, and another time in the normal build context.

#### Solution

Thankfully, both crates were only using the hasher from `solana-program`, so this change just uses `Sha256` directly. Some tests might fail here, so I'll fix those up as needed.

I've also bumped the versions so we can publish new versions and use those immediately.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6471,7 +6471,7 @@ dependencies = [
 
 [[package]]
 name = "spl-discriminator-derive"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "quote 1.0.29",
  "spl-discriminator-syn",
@@ -6480,7 +6480,7 @@ dependencies = [
 
 [[package]]
 name = "spl-discriminator-syn"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "proc-macro2 1.0.63",
  "quote 1.0.29",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6484,7 +6484,7 @@ version = "0.1.0"
 dependencies = [
  "proc-macro2 1.0.63",
  "quote 1.0.29",
- "solana-program",
+ "sha2 0.10.7",
  "syn 2.0.28",
  "thiserror",
 ]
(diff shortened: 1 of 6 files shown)


buffalojoec:
> Since the derive crates are `proc-macro = true`, cargo treats it and its dependencies as completely different libraries, which meant that we were building `solana-program` twice: once in the proc-macro context, and another time in the normal build context.

How can we use the same dependency within a proc macro crate and the library that re-exports the macro crate, then? And is this only a problem if we re-export?
Strange that it causes collisions for `solana-program` but not `quote`, `syn`, or `thiserror`.

joncinque:
I think the issue might be because we're patching solana-program in the monorepo, to force the SPL crates to share the same version of solana-program:

https://github.com/solana-labs/solana/blob/ca92e9c3877e2a17374d59b634ab98837212e0fb/Cargo.toml#L421-L443

buffalojoec:
> I think the issue might be because we're patching solana-program in the monorepo, to force the SPL crates to share the same version of solana-program:
> 
> https://github.com/solana-labs/solana/blob/ca92e9c3877e2a17374d59b634ab98837212e0fb/Cargo.toml#L421-L443

Yeah that's definitely it. For now I think your fix is fine since we're just using sha256, but we may run into this in the future again, and might need to do something a little different if we actually need `solana-program`.

buffalojoec (approved):
This is fine

PR #5404, opened 2023-10-02
ci: Ignore tungstenite security advisory

joncinque:
#### Problem

There's a RUSTSEC advisory on tungstenite https://rustsec.org/advisories/RUSTSEC-2023-0065 causing CI failures

#### Solution

Same as 1.16, ignore the advisory: https://github.com/solana-labs/solana/pull/33457


joncinque:
Audit passed, merging early

PR #5405, opened 2023-10-03
token-2022: Refresh blockhash in flaky test

joncinque:
#### Problem

There was a spurious failure during CI in a confidential transfer test: https://github.com/solana-labs/solana-program-library/actions/runs/6383061276/job/17322950719

Essentially, we're doing the same thing twice in a row without refreshing the blockhash, so there's a chance that the program-test will return the same old response.

#### Solution

Refresh the blockhash


samkim-crypto (approved):
Looks good to me!

PR #5414, opened 2023-10-03
single-pool: update docs with examples

2501babe:


2501babe:
well this is driving me crazy...

as-is, this block:
```
const transaction = await SinglePoolProgram.deposit({
  connection,
  pool: poolAddress,
  userWallet,
  userStakeAccount,
});

// sign with the fee payer and stake account withdraw authority, if these signers differ
// userWallet is a convenience parameter to use one account as a payer, authority, and lamport recipient
```

produces this error:
```
SyntaxError: /home/hana/work/solana/spl/docs/src/single-pool.mdx: Unexpected token, expected "}" (240:6)
  238 | const transaction = await SinglePoolProgram.deposit({
  239 |   connection,
> 240 |   pool: poolAddress,
      |       ^
  241 |   userWallet,
  242 |   userStakeAccount,
  243 | });
[ERROR] Client bundle compiled with errors therefore further build is impossible.
```

at first i thought maybe it doesnt understand the js autofill thing (where you type `a,` for `a: a,` but its not that because replacing all of them with explicit colon syntax errors on the first colon. deleting the line with the colon instead produces this:

```
SyntaxError: /home/hana/work/solana/spl/docs/src/single-pool.mdx: Unexpected token, expected "}" (239:12)
  237 | ```typescript
  238 | const transaction = await SinglePoolProgram.deposit({
> 239 |   connection: a,
      |             ^
  240 |   pool: poolAddress,
  241 |   userWallet: b,
  242 |   userStakeAccount: c,
[ERROR] Client bundle compiled with errors therefore further build is impossible.
```

the only things i can think of is i have a syntax error somewhere (im 100% sure there are no mismatched brackets anywhere in the file because i can use my editor to check for them, and 90% sure there are no mismatched backticks) or the docusaurus parser is broken in some way...? @joncinque you wouldnt happen to have any insight would you...

2501babe:
something extremely strange is happening because if i put `const x = {a: 1};` into that typescript block, it gives me the error, but if it put it in any prior typescript block, its fine, and the error shows up at `pool: poolAddress,` like normal. im reduced to deleting random things in between to see if theres some mismatched... anything?

2501babe:
ok this is 100% a parser bug in docusaurus because the error moves from `x` to `transaction` in this:

```
const x = {a: 1}; 
const transaction = await SinglePoolProgram.deposit({
  connection,
  pool: poolAddress,
  userWallet,
  userStakeAccount,
});
```

when i add a newline between `<TabItem value="jsc" label="WEB3.JS CLASSIC">` and `` ```typescript ``

2501babe:
i fixed it by adding tons of newlines to the file and i am now a supporter of any totalitarian regime that would force all web programmers to take an intro-level computational linguistics course

2501babe:
im ready for review with the caveat that i need to fill in whatever the npm package name will be where it says `XXX TODO`

docs/src/single-pool.mdx
```diff
@@ -0,0 +1,477 @@
...
+staking directly while holding the value in a tokenized form.
+It also allows you to buy or sell stakes smaller than the minimum delegation on the market.
+
+Assuming a vote account `Ammgaa2iZfA745BmZMhkcS27uh87fEVDC6Gm2RXz5hrC` exists, we create a pool at address `DkE6XFGbqSyYzRugLVSmmB42F9BQZ7mZU837e2Cti7kb`:
+
+<Tabs className="unique-tabs" groupId="language-selection">
+  <TabItem value="cli" label="CLI" default>
+
+```console
+spl-single-pool initialize Ammgaa2iZfA745BmZMhkcS27uh87fEVDC6Gm2RXz5hrC
```
joncinque, line 172: nit-up-to-you-for-consistency: the earlier examples start the line with `$`, and this one doesn't. maybe it's best to be consistent and not use them anywhere, and instead have an empty line between the command and the output

docs/src/single-pool.mdx
```diff
@@ -0,0 +1,477 @@
...
+```typescript
+const transaction = await SinglePoolProgram.createTokenMetadata(poolAddress, feePayerAddress);
+
+// sign with the fee payer
+```
+
+  </TabItem>
+</Tabs>
+
+The default token metadata is only minimally helpful, spotlighting the address of the pool.
```
joncinque, line 243: is this correct?
```suggestion
The default token metadata is only minimally helpful, spotlighting the address of the validator vote account.
```

docs/src/single-pool.mdx
```diff
@@ -0,0 +1,477 @@
...
+  'cPool',
+);
+
+// sign with the fee payer and authorized withdrawer
+```
+
+  </TabItem>
+</Tabs>
+
+A URL may also be included as an optional parameter.
```
joncinque, line 288: Could you include the URL in the examples? I think most people will want to do that

docs/src/single-pool.mdx
```diff
@@ -0,0 +1,477 @@
...
+The user retains full authority on the stake account until they decide to deposit.
+
+<Tabs className="unique-tabs" groupId="language-selection">
+  <TabItem value="cli" label="CLI" default>
+
+```console
+spl-single-pool create-default-stake --pool DkE6XFGbqSyYzRugLVSmmB42F9BQZ7mZU837e2Cti7kb 1000000000
+```
+
+(some time passes)
```
joncinque, line 357: nit: this can be more specific
```suggestion
Once the stake becomes active, typically after crossing one epoch boundary.
```

docs/src/single-pool.mdx
```diff
@@ -0,0 +1,477 @@
...
+  connection,
+  voteAccountAddress,
+  userWallet,
+  minimumDelegation,
+);
+
+// sign with user wallet
+```
+
+(some time passes)
```
joncinque, line 379: nit: same here
```suggestion
Once the stake becomes active, typically after crossing one epoch boundary.
```

docs/src/single-pool.mdx
```diff
@@ -0,0 +1,477 @@
...
+```console
+spl-single-pool deposit --pool DkE6XFGbqSyYzRugLVSmmB42F9BQZ7mZU837e2Cti7kb --default-stake-account
+```
+
+  </TabItem>
+
+  <TabItem value="jsc" label="WEB3.JS CLASSIC">
+
+```typescript
+const minimumDelegation = (await connection.getStakeMinimumDelegation()).value;
```
joncinque, line 368: nit-up-to-you: how about putting a comment that any value greater than or equal to the minimum delegation is valid?

docs/src/single-pool.mdx
```diff
@@ -0,0 +1,477 @@
...
+  rpc,
+  voteAccountAddress,
+  userWallet,
+  minimumDelegation,
+);
+
+// sign with user wallet, which is used as the fee payer and as the base address for a seeded account
+```
+
+(some time passes)
```
joncinque, line 408:
```suggestion
Once the stake becomes active, typically after crossing one epoch boundary.
```

joncinque (commented):
Looks great overall! Just a few small comments, but this will be great to get people started

docs/src/single-pool.mdx
2501babe: will add the $ actually, ty for catching this, i copied the above stuff from the token docs

docs/src/single-pool.mdx
2501babe: oops youre right! i was looking at the create with seed call when i wrote this..

docs/src/single-pool.mdx
2501babe: you know, it occurs to me that until the minimum for stake program is activated, they can actually do this with any amount. the only enforcement done anywhere is that the _pool_ doesnt go below max(1sol, minimum). maybe to allay confusion ill just hardcode the number...

2501babe:
alright merging after i have npm urls to add

2501babe:
latest commit per #5488

joncinque (approved):
Looks good to me once the links are updated with the deployed packages.

joncinque:
I'm pretty sure we still want this, right?

2501babe:
yes i just needed to finally get the js on npm, which is now done!

PR #5437, opened 2023-10-05
token-2022(feat): add serialisation option for the extensions

### LINKED ISSUE #5431: [spl-token-2022]: Allow serialisation/deserialisation for token-2022 extensions
Hey guys,  I am planning to add support for `token-2022` indexing. In order to do this, I would like to have all the extensions support serialising/de-serialising. Is it ok if I create a PR to support this via the `serde-trait` feature-flag? Is there any reason not to do this?

nikhil-helius:
Fixes #5431 

Note that for extension `confidentional_transfer` and `confidential_transfer_fee`,  some the fields are in a separate sdk which doesn't  have the serialisation enabled. We would have to change there first before setting it here 

Tested by turning on the compile time flags `serde-trait` by default and using it in our codebase.

### DIFF
--- a/token/program-2022/src/extension/cpi_guard/mod.rs
+++ b/token/program-2022/src/extension/cpi_guard/mod.rs
@@ -8,6 +8,9 @@ use {
     spl_pod::primitives::PodBool,
 };
 
+#[cfg(feature = "serde-traits")]
+use serde::{Deserialize, Serialize};
+
 /// CPI Guard extension instructions
 pub mod instruction;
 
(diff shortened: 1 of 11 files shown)


buffalojoec (approved):
Lgtm!

joncinque (approved):
Looks great, thanks! If you're interested in doing the zk bits, you'll find those types at https://github.com/solana-labs/solana/tree/master/zk-token-sdk

nikhil-helius:
> Looks great, thanks! If you're interested in doing the zk bits, you'll find those types at https://github.com/solana-labs/solana/tree/master/zk-token-sdk

Done: https://github.com/solana-labs/solana/pull/33760 
This was tested in our private fork.

joncinque (commented):
This is my fault, I'm just realizing that all of these should have `rename_all = "camelCase"` configured too. I'll get a PR up for that

PR #5439, opened 2023-10-06
stake-pool: Allow removal of force-destaked validator

joncinque:
#### Problem

As mentioned in a Discord discussion https://discord.com/channels/428295358100013066/859540542608900127/1157393252119478332, stake pools cannot properly handle a validator that has been force-destaked through the new `DeactivateDelinquent` instruction on the stake program.

#### Solution

It's pretty simple -- don't try to deactivate a stake account again during removal. This way, the manager simply needs to remove the validator and then re-run an update to reabsorb the account.

While going through this, I also relaxed the check on an activating / active transient stake account, so that it'll just deactivate it. This way, we prevent a potential malfunction during this edge case:

* someone deactivates-delinquent the main stake account, but *not* an activating transient stake account
* the delinquent validator comes back online
* it's now impossible to deactivate-delinquent the activating transient stake account
* pool is confused and can't remove the validator

### DIFF
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -777,7 +777,7 @@ pub fn remove_validator_from_pool(
         AccountMeta::new_readonly(*stake_pool_withdraw, false),
         AccountMeta::new(*validator_list, false),
         AccountMeta::new(*stake_account, false),
-        AccountMeta::new_readonly(*transient_stake_account, false),
+        AccountMeta::new(*transient_stake_account, false),
         AccountMeta::new_readonly(sysvar::clock::id(), false),
         AccountMeta::new_readonly(stake::program::id(), false),
     ];
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -112,7 +112,7 @@ pub enum StakePoolInstruction {
     ///   2. `[]` Stake pool withdraw authority
     ///   3. `[w]` Validator stake list storage account
     ///   4. `[w]` Stake account to remove from the pool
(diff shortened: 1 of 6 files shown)


2501babe:
will review this next!

stake-pool/program/src/instruction.rs
```diff
@@ -777,7 +777,7 @@ pub fn remove_validator_from_pool(
         AccountMeta::new_readonly(*stake_pool_withdraw, false),
         AccountMeta::new(*validator_list, false),
         AccountMeta::new(*stake_account, false),
-        AccountMeta::new_readonly(*transient_stake_account, false),
+        AccountMeta::new(*transient_stake_account, false),
```
2501babe, line 780: i cant figure out how comment or suggest corrections on lines github considers out of scope but line 115 should change to `///   5. `[w]` Transient stake account`

2501babe (dismissed):

2501babe:
i love how simple this change is! it handles the case its meant to without needing to check for it, just two independent deactivations that accomplish the broader goal of removing the validator. very nice!

stake-pool/program/src/instruction.rs
joncinque: Ah great point, will fix, thanks!

joncinque:
That was the hope! I go cross-eyed making these changes sometimes, so I'm glad that you liked the simplicity of this one

CriesofCarrots:
@joncinque , seems like `ssuccess_remove_validator` is failing on downstream CI pretty consistently:
https://github.com/solana-labs/solana/actions/runs/6489539720/job/17624866919?pr=33668
https://github.com/solana-labs/solana/actions/runs/6487946093/job/17619381251

PR #5488, opened 2023-10-10
single-pool: add `ReactivatePool` instruction

2501babe:
handles the issue described in #5439. the first commit is the important one (changes the program itself), the others just add to the cli. also i moved half the commands into a subcommand, 99% of users are only ever going to need `deposit` and `withdraw` so im trying to minimize visual clutter

### DIFF
--- a/single-pool/cli/src/cli.rs
+++ b/single-pool/cli/src/cli.rs
@@ -71,15 +71,9 @@ pub struct Cli {
 
 #[derive(Clone, Debug, Subcommand)]
 pub enum Command {
-    /// Permissionlessly create the single-validator stake pool for a given validator vote account
-    /// if one does not already exist. The fee payer also pays rent-exemption for accounts,
-    /// along with the cluster-configured minimum stake delegation
-    Initialize(InitializeCli),
-
-    /// Permissionlessly re-stake the pool stake account in the case when it has been deactivated.
-    /// This may happen if the validator is force-deactivated, and then later reactivated using
-    /// the same address for its vote account.
-    Reactivate(ReactivateCli),
+    /// Commands used to initialize or manage existing single-validator stake pools.
+    /// Other than initializing new pools, most users should never need to use these.
+    Manage(ManageCli),
 
     /// Deposit delegated stake into a pool in exchange for pool tokens, closing out
     /// the original stake account. Provide either a stake account address, or a
@@ -92,6 +86,32 @@ pub enum Command {
     /// or the ALL keyword to burn all.
     Withdraw(WithdrawCli),
 
+    /// Create and delegate a new stake account to a given validator, using a default address
+    /// linked to the intended depository pool
+    CreateDefaultStake(CreateStakeCli),
+
+    /// Display info for one or all single single-validator stake pool(s)
+    Display(DisplayCli),
+}
+
+#[derive(Clone, Debug, Parser)]
+pub struct ManageCli {
+    #[clap(subcommand)]
+    pub manage: ManageCommand,
+}
+
+#[derive(Clone, Debug, Subcommand)]
+pub enum ManageCommand {
+    /// Permissionlessly create the single-validator stake pool for a given validator vote account
+    /// if one does not already exist. The fee payer also pays rent-exemption for accounts,
+    /// along with the cluster-configured minimum stake delegation
+    Initialize(InitializeCli),
+
+    /// Permissionlessly re-stake the pool stake account in the case when it has been deactivated.
+    /// This may happen if the validator is force-deactivated, and then later reactivated using
+    /// the same address for its vote account.
+    Reactivate(ReactivateCli),
+
     /// Permissionlessly create default MPL token metadata for the pool mint. Normally this is done
     /// automatically upon initialization, so this does not need to be called.
     CreateTokenMetadata(CreateMetadataCli),
@@ -99,13 +119,6 @@ pub enum Command {
     /// Modify the MPL token metadata associated with the pool mint. This action can only be
     /// performed by the validator vote account's withdraw authority
     UpdateTokenMetadata(UpdateMetadataCli),
-
-    /// Create and delegate a new stake account to a given validator, using a default address
-    /// linked to the intended depository pool
-    CreateDefaultStake(CreateStakeCli),
-
-    /// Display info for one or all single single-validator stake pool(s)
-    Display(DisplayCli),
 }
 
 #[derive(Clone, Debug, Args)]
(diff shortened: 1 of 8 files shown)


single-pool/program/src/instruction.rs
```diff
@@ -43,6 +43,19 @@ pub enum SinglePoolInstruction {
     ///  12. `[]` Stake program
     InitializePool,
 
+    ///   Restake the pool stake account if its validator is force-deactivated and later
+    ///   recreated with the same vote account.
```
2501babe, line 47: feel free to suggest different wording for this, im not familiar enough with the mechanism that causes this to happen to craft proper verbiage

single-pool/program/src/instruction.rs
```diff
@@ -43,6 +43,19 @@ pub enum SinglePoolInstruction {
...
+    ///
+    ///   0. `[]` Validator vote account
+    ///   1. `[]` Pool account
+    ///   2. `[w]` Pool stake account
+    ///   3. `[]` Pool stake authority
+    ///   4. `[]` Clock sysvar
+    ///   5. `[]` Stake history sysvar
+    ///   6. `[]` Stake config sysvar
+    ///   7. `[]` Stake program
+    ReactivatePool,
```
joncinque, line 57: I got a little tripped up on the name, but I see what you use it, since `redelegate` means something else. The only thing I was able to come up with is `DelegatePoolStake` which would say literally what's happening, but wouldn't be as clear as to when it would be useful. If you prefer `ReactivatePool`, I'm good with it

single-pool/program/tests/accounts.rs
```diff
@@ -196,6 +196,7 @@ async fn fail_account_checks(test_mode: TestMode) {
 
 // make an individual instruction for all program instructions
 // the match is just so this will error if new instructions are added
+// if you are reading this because of that error, add the case to the `consistent_account_order` test!!!
```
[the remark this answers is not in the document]
joncinque: Haha we all have goldfish-brain moments

single-pool/program/src/processor.rs
```diff
@@ -707,6 +707,64 @@ impl Processor {
...
+            program_id,
+            pool_info.key,
+            pool_stake_authority_info.key,
+        )?;
+        check_stake_program(stake_program_info.key)?;
+
+        let (_, pool_stake_state) = get_stake_state(pool_stake_info)?;
+        if pool_stake_state.delegation.deactivation_epoch >= clock.epoch {
+            return Err(SinglePoolError::WrongStakeState.into());
+        }
```
joncinque, line 738: I might be missing something, but with this check, it makes it impossible for someone to reactivate a stake that was *just* `DeactivateDelinquent`ed in the same epoch. I think it should be safe to remove.

single-pool/program/tests/reactivate.rs
```diff
@@ -0,0 +1,149 @@
...
+        },
+    ))
+    .unwrap();
+
+    let mut stake_account = get_account(&mut context.banks_client, &accounts.stake_account).await;
+    stake_account.data = account_data;
+    context.set_account(
+        &accounts.stake_account,
+        &AccountSharedData::from(stake_account),
+    );
```
joncinque, line 51: I should really use this in my testing too, this was a nice reminder, thanks!

single-pool/cli/src/main.rs
```diff
@@ -157,6 +164,61 @@ async fn command_initialize(config: &Config, command_config: InitializeCli) -> C
...
+            .await?
+            .unwrap()
+            .1
+            .delegation
+            .deactivation_epoch
+            < current_epoch;
+
+        if !pool_stake_deactivated {
+            return Err("Pool stake account is not deactivated".into());
+        }
```
joncinque, line 201: Same as in the processor: I think we can relax the check to `<= current_epoch`, so that it's possible to reactivate something that was just deactivated in the same epoch

single-pool/cli/src/cli.rs
```diff
@@ -114,6 +132,22 @@ pub struct InitializeCli {
...
+    #[clap(short, long = "pool", value_parser = |p: &str| parse_address(p, "pool_address"))]
+    pub pool_address: Option<Pubkey>,
+
+    /// The vote account corresponding to the pool to reactivate
+    #[clap(long = "vote-account", value_parser = |p: &str| parse_address(p, "vote_account_address"))]
+    pub vote_account_address: Option<Pubkey>,
+
+    // backdoor for testing, theres no reason to ever use this
+    #[clap(long, hide = true)]
+    pub yolo: bool,
```
joncinque, line 148: Even though it's a backdoor, we can be clearer about its usage, just in case:
```suggestion
    // backdoor for testing, theres no reason to ever use this
    #[clap(long, hide = true)]
    pub skip_deactivation_check: bool,
```

single-pool/cli/src/main.rs
```diff
@@ -157,6 +164,61 @@ async fn command_initialize(config: &Config, command_config: InitializeCli) -> C
...
+        Some(&payer.pubkey()),
+        &vec![payer],
+        config.program_client.get_latest_blockhash().await?,
+    );
+
+    let signature = process_transaction(config, transaction).await?;
+
+    Ok(format_output(
+        config,
+        "Reactivate".to_string(),
```
joncinque, line 217: nit: to be consistent with the instruction name?
```suggestion
        "ReactivatePool".to_string(),
```

joncinque (commented):
Looks really good! Just a few little comments

single-pool/program/src/instruction.rs
2501babe: `ReactivateStake`? `ReactivatePoolStake`?

single-pool/program/src/processor.rs
2501babe: i just always err toward being as restrictive as possible to minimize the chances of corner cases. ill relax it to `>`

single-pool/cli/src/main.rs
2501babe: theres a mystery convention in my head where `InitializePool` -> `Initialize`, `DepositStake` -> `Deposit` etc because these have longer "instruction" names and shorter "transaction" names, but since this has no multi-instruction transaction i guess that makes more sense

single-pool/program/src/instruction.rs
2501babe: last one seems fine

2501babe:
normally id merge because the fixes are trivial but wanna get the stamp because i changed the program. also i copied your fix from #5514

PR #5494, opened 2023-10-10
chore: Bump Solana crates to 1.16.16

joncinque:
#### Problem

The Solana crates have moved up to 1.16.16, which contains the updated account decoders for token-2022 from https://github.com/solana-labs/solana/pull/33465. With the newer crates, we'll be able to `spl-token display` all extensions.

#### Solution

Just run the script to bump the crates.


samkim-crypto (approved):
Thanks!

PR #5495, opened 2023-10-11
token-cli: Display all mint and account extensions

joncinque:
#### Problem

There are many extensions that aren't displayed by the token CLI, because it's using the older parsers.

#### Solution

The parsers were updated in https://github.com/solana-labs/solana/pull/33453 though, so use all of those!

Before, a devnet mint with metadata showed no extensions, and now it'll show:

```
SPL Token Mint
  Address: 25HxeJ6NKj8akSKLZmnxzLtxKnSK1wZEwjuCEJjsijqJ
  Program: TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb
  Supply: 3000000
  Decimals: 6
  Mint authority: DtkFBodq67HK9tzFgkjnWw2ZS4LZZgAVqeerukxp6rZs
  Freeze authority: GkatiJtcJWVToqYgqQKRx6qUbcTAzMm5fLVgzZCeS3xH
Extensions
  Permanent delegate: GkatiJtcJWVToqYgqQKRx6qUbcTAzMm5fLVgzZCeS3xH
  Metadata Pointer:
    Authority: G8ENSYKVGPbRTbcN1BxXuRMYeCq13271UjCUrrpZTJ4X
    Metadata address: 25HxeJ6NKj8akSKLZmnxzLtxKnSK1wZEwjuCEJjsijqJ
  Metadata:
    Update Authority: Acwmy2AMbcgaJs2Zt7bWPVGfWnWquVEcKz4j9rbuVoKn
    Mint: 25HxeJ6NKj8akSKLZmnxzLtxKnSK1wZEwjuCEJjsijqJ
    Name: RAIN_Coin
    Symbol: RAIN
    URI: https://arweave.net/wyrYETI0J5ys3QVfTajFcMDkqDlRnlroK02YEGoBuRs
```

While I was at it, I removed the ad-hoc confidential parser.

### DIFF
--- a/token/cli/src/output.rs
+++ b/token/cli/src/output.rs
@@ -4,20 +4,14 @@ use serde::{Deserialize, Serialize, Serializer};
 use solana_account_decoder::{
     parse_token::{UiAccountState, UiMint, UiMultisig, UiTokenAccount, UiTokenAmount},
     parse_token_extension::{
-        UiCpiGuard, UiDefaultAccountState, UiExtension, UiInterestBearingConfig, UiMemoTransfer,
-        UiMintCloseAuthority, UiPermanentDelegate, UiTransferFeeAmount, UiTransferFeeConfig,
+        UiConfidentialTransferAccount, UiConfidentialTransferFeeAmount,
+        UiConfidentialTransferFeeConfig, UiConfidentialTransferMint, UiCpiGuard,
+        UiDefaultAccountState, UiExtension, UiInterestBearingConfig, UiMemoTransfer,
+        UiMetadataPointer, UiMintCloseAuthority, UiPermanentDelegate, UiTokenMetadata,
+        UiTransferFeeAmount, UiTransferFeeConfig, UiTransferHook, UiTransferHookAccount,
     },
 };
 use solana_cli_output::{display::writeln_name_value, OutputFormat, QuietDisplay, VerboseDisplay};
-use spl_token_2022::{
-    extension::{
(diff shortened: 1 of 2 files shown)


token/cli/src/output.rs
```diff
@@ -698,49 +686,230 @@ fn display_ui_extension(
...
+            pending_balance_credit_counter,
+            maximum_pending_balance_credit_counter,
+            expected_pending_balance_credit_counter,
+            actual_pending_balance_credit_counter,
+        }) => {
+            writeln!(f, "  {}", style("Confidential transfer:").bold())?;
+            writeln_name_value(f, "    Approved:", &format!("{approved}"))?;
+            writeln_name_value(f, "    Encryption key:", elgamal_pubkey)?;
+            writeln_name_value(f, "    Pending Balance Low:", pending_balance_lo)?;
+            writeln_name_value(f, "    Pending Balance Hi:", pending_balance_hi)?;
```
samkim-crypto, line 707: Seems like we should stick to either `lo`/`hi` or `low`/`high`?

samkim-crypto (approved):
Looks good to me! Thanks!

token/cli/src/output.rs
joncinque: Ah right, good call, what do you prefer? Maybe we go `Low` / `High` for clarity?

samkim-crypto: Yeah, I think Low/High will be nice.

PR #5512, opened 2023-10-11
token group: init interface

buffalojoec:
This PR initializes a newer, slimmed down version of the SPL Token Group interface.

Previous PR: #5190

### DIFF
--- a/token-group/interface/Cargo.toml
+++ b/token-group/interface/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "spl-token-group-interface"
+version = "0.1.0"
+description = "Solana Program Library Token Group Interface"
+authors = ["Solana Labs Maintainers <maintainers@solanalabs.com>"]
+repository = "https://github.com/solana-labs/solana-program-library"
+license = "Apache-2.0"
+edition = "2021"
+
+[dependencies]
+bytemuck = "1.14.0"
+solana-program = "1.16.3"
+spl-discriminator = { version = "0.1.0" , path = "../../libraries/discriminator" }
+spl-pod = { version = "0.1.0" , path = "../../libraries/pod", features = ["borsh"] }
+spl-program-error = { version = "0.3.0" , path = "../../libraries/program-error" }
+
+[dev-dependencies]
+spl-type-length-value = { version = "0.3.0", path = "../../libraries/type-length-value", features = ["derive"] }
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
--- a/token-group/interface/Cargo.toml
+++ b/token-group/interface/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 
 [dependencies]
 bytemuck = "1.14.0"
-solana-program = "1.16.3"
+solana-program = "1.16.16"
 spl-discriminator = { version = "0.1.0" , path = "../../libraries/discriminator" }
 spl-pod = { version = "0.1.0" , path = "../../libraries/pod", features = ["borsh"] }
 spl-program-error = { version = "0.3.0" , path = "../../libraries/program-error" }
--- a/token-group/interface/src/instruction.rs
+++ b/token-group/interface/src/instruction.rs
@@ -18,7 +18,7 @@ use {
 /// Instruction data for initializing a new `Group`
 #[repr(C)]
 #[derive(Clone, Copy, Debug, PartialEq, Pod, Zeroable, SplDiscriminate)]
-#[discriminator_hash_input("spl_token_group_interface:initialize_group")]
+#[discriminator_hash_input("spl_token_group_interface:initialize_token_group")]
 pub struct InitializeGroup {
     /// Update authority for the group
     pub update_authority: OptionalNonZeroPubkey,
@@ -38,7 +38,7 @@ pub struct UpdateGroupMaxSize {
 /// Instruction data for updating the authority of a `Group`
 #[repr(C)]
 #[derive(Clone, Copy, Debug, PartialEq, Pod, Zeroable, SplDiscriminate)]
-#[discriminator_hash_input("spl_token_group_interface:update_group_authority")]
+#[discriminator_hash_input("spl_token_group_interface:update_authority")]
 pub struct UpdateGroupAuthority {
     /// New authority for the group, or unset if `None`
     pub new_authority: OptionalNonZeroPubkey,
@@ -89,7 +89,8 @@ pub enum TokenGroupInstruction {
     /// Accounts expected by this instruction:
     ///
     ///   0. `[w]`  Member
-    ///   1. `[s]`  Member update authority
+    ///   1. `[]`   Member mint
+    ///   1. `[s]`  Member mint authority
     ///   2. `[w]`  Group
     ///   3. `[s]`  Group update authority
     InitializeMember(InitializeMember),
@@ -155,7 +156,6 @@ pub fn initialize_group(
     mint_authority: &Pubkey,
     update_authority: Option<Pubkey>,
     max_size: u32,
-    extra_account_metas: &[AccountMeta],
 ) -> Instruction {
     let update_authority = OptionalNonZeroPubkey::try_from(update_authority)
         .expect("Failed to deserialize `Option<Pubkey>`");
@@ -164,16 +164,13 @@ pub fn initialize_group(
         max_size: max_size.into(),
     })
     .pack();
-    let mut accounts = vec![
-        AccountMeta::new(*group, false),
-        AccountMeta::new_readonly(*mint, false),
-        AccountMeta::new_readonly(*mint_authority, true),
-    ];
-    accounts.extend_from_slice(extra_account_metas);
-
     Instruction {
         program_id: *program_id,
-        accounts,
+        accounts: vec![
+            AccountMeta::new(*group, false),
+            AccountMeta::new_readonly(*mint, false),
+            AccountMeta::new_readonly(*mint_authority, true),
+        ],
         data,
     }
 }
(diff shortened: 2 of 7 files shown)


token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,337 @@
...
+
+/// Instruction data for initializing a new `Member` of a `Group`
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Pod, Zeroable, SplDiscriminate)]
+#[discriminator_hash_input("spl_token_group_interface:initialize_member")]
+pub struct InitializeMember {
+    /// The pubkey of the `Group`
+    pub group: Pubkey,
+    /// The member number
+    pub member_number: u32,
```
joncinque, line 54: Why is this needed? Seems like it should be able to figure it out from the group, no?

token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,337 @@
...
+    pub new_authority: OptionalNonZeroPubkey,
+}
+
+/// Instruction data for initializing a new `Member` of a `Group`
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Pod, Zeroable, SplDiscriminate)]
+#[discriminator_hash_input("spl_token_group_interface:initialize_member")]
+pub struct InitializeMember {
+    /// The pubkey of the `Group`
+    pub group: Pubkey,
```
joncinque, line 52: Why is this needed? It's already provided as an account

token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,337 @@
...
+    UpdateGroupAuthority(UpdateGroupAuthority),
+
+    /// Initialize a new `Member` of a `Group`
+    ///
+    /// Assumes the `Group` has already been initialized,
+    /// as well as the mint for the member.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[w]`  Member
```
joncinque, line 100: Interesting, if I'm reading this right, since this can just be any account, then a mint can be a member of multiple groups. Is that correct?

token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,337 @@
...
+    ///   1. `[]`   Member Mint
+    ///   2. `[s]`  Member Mint authority
+    ///   3. `[w]`  Group
+    ///   4. `[]`   Group Mint
+    ///   5. `[s]`  Group Mint authority
```
joncinque, line 105: I'm thinking this should be the group update authority since the group exists. What do you think?

token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// Assumes the `Group` has already been initialized,
+    /// as well as the mint for the member.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[w]`  Member
+    ///   1. `[]`   Member Mint
+    ///   2. `[s]`  Member Mint authority
+    ///   3. `[w]`  Group
+    ///   4. `[]`   Group Mint
```
joncinque, line 104: What will this be needed for?

joncinque (commented):
This is looking really good! Just some bits to iron out on the interface, then this should be good to start on

token-group/interface/src/instruction.rs
buffalojoec: Yep, if that's what we want! Right now I left it up for discussion.

token-group/interface/src/instruction.rs
buffalojoec: I could go either way tbh. I think update authority makes more since only because you might have an NFT as the group, and the mint authority is `None`, so that's prob preferred.

Note I was also considering removing the member mint authority, wdyt?

token-group/interface/src/instruction.rs
buffalojoec: So no group mint then? I'm sure it doesn't _need_ to be part of the interface yeah.

token-group/interface/src/instruction.rs
buffalojoec: Good point ha

token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// The member number
+    pub member_number: u32,
+}
+
+/// All instructions that must be implemented in the SPL Token Group Interface
+///
+/// Note: Any instruction can be extended using additional required accounts by
+/// using the `InitializeExtraAccountMetaList` instruction to write
+/// configurations for extra required accounts into validation data
+/// corresponding to an instruction's unique discriminator.
```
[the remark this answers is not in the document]
buffalojoec: This snuck in there 🫠

buffalojoec:
OK I've resolved most of the comments, but there's still a few items to discuss:
- The `u32` in place of `u64` for things like `max_size` is a result of #5518 
- Multiple groups for one member
- The instructions discriminators being a high number

token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,337 @@
...
+    spl_pod::{
+        bytemuck::{pod_bytes_of, pod_from_bytes},
+        optional_keys::OptionalNonZeroPubkey,
+    },
+};
+
+/// Instruction data for initializing a new `Group`
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Pod, Zeroable, SplDiscriminate)]
+#[discriminator_hash_input("spl_token_group_interface:initialize_group")]
```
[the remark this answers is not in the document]
joncinque: Correct, I just check for a first number that's "pretty big", at least in the 100s

token-group/interface/src/instruction.rs
joncinque: Hm, I'd say let's keep the member mint authority, because then someone else can include your mint in all sorts of groups without your permission, which would probably help confuse people a lot.

token-group/interface/src/instruction.rs
joncinque: Yeah correct, I don't think we need any information from the mint

token-group/interface/src/instruction.rs
joncinque: awesome, that works!

token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,308 @@
...
+pub struct InitializeGroup {
+    /// Update authority for the group
+    pub update_authority: OptionalNonZeroPubkey,
+    /// The maximum number of group members
+    pub max_size: u32,
```
joncinque, line 25: You'll likely need to use `PodU32` for the instructions and structs to avoid padding / alignment issues

token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+/// Creates a `InitializeMember` instruction
+#[allow(clippy::too_many_arguments)]
+pub fn initialize_member(
+    program_id: &Pubkey,
+    group: &Pubkey,
+    group_update_authority: &Pubkey,
+    member: &Pubkey,
```
joncinque, line 227: micro-nit: I like to have the params match the instruction
```suggestion
    member: &Pubkey,
    group: &Pubkey,
    group_update_authority: &Pubkey,
```

token-group/interface/src/state.rs
```diff
@@ -0,0 +1,180 @@
...
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable, SplDiscriminate)]
+#[discriminator_hash_input("spl_token_group_interface:group")]
+pub struct Group {
+    /// The authority that can sign to update the group
+    pub update_authority: OptionalNonZeroPubkey,
+    /// The current number of group members
+    pub size: u32,
+    /// The maximum number of group members
+    pub max_size: u32,
```
joncinque, line 21: Same here, these should probably be `PodU32`s

token-group/interface/src/instruction.rs
```diff
@@ -0,0 +1,308 @@
...
+    pub max_size: u32,
+}
+
+/// Instruction data for updating the max size of a `Group`
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq, Pod, Zeroable, SplDiscriminate)]
+#[discriminator_hash_input("spl_token_group_interface:update_group_max_size")]
+pub struct UpdateGroupMaxSize {
+    /// New max size for the group
+    pub max_size: u32,
```
joncinque, line 34: `PodU32` here

joncinque (commented):
Just a few last things, then this should be good to go!

buffalojoec:
> Edit: oh, and we'll also need the member mint authority to sign the `initialize_member` instruction

Would you say mint authority or update authority? A case could be made for either imo.

buffalojoec:
Also, I think we're good on discriminators:

- `InitializeGroup`: [88, 13, 213, 85, 206, 182, 83, 46]
- `UpdateGroupAuthority`: [67, 150, 244, 200, 81, 37, 35, 55]
- `UpdateGroupMaxSize`: [108, 37, 171, 143, 248, 30, 18, 110]
- `InitializeMember`: [152, 32, 222, 176, 223, 237, 116, 134]

buffalojoec:
* master
  * **PR #5512** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5512" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5512?utm_source=stack-comment).

joncinque:
> Would you say mint authority or update authority? A case could be made for either imo.

There is no update authority here since we're just initializing the member, right? I think that means we have to go with member mint authority

joncinque (approved):
Looks good!

token-group/interface/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "0.1.0"
+description = "Solana Program Library Token Group Interface"
+authors = ["Solana Labs Maintainers <maintainers@solanalabs.com>"]
+repository = "https://github.com/solana-labs/solana-program-library"
+license = "Apache-2.0"
+edition = "2021"
+
+[dependencies]
+bytemuck = "1.14.0"
+solana-program = "1.16.3"
```
joncinque, line 12: Just need to get this correct so that the automatic updater works later
```suggestion
solana-program = "1.16.16"
```

joncinque (approved):
Looks great!

buffalojoec:
Sorry @joncinque! Had to fix a `rustfmt` error!

joncinque (approved):
No worries! Feel free to merge without an explicit re-approval if you just fixed a rustfmt issue

PR #5514, opened 2023-10-12
stake-pool: Fix test for monorepo downstream job

joncinque:
#### Problem

As noted in https://github.com/solana-labs/solana-program-library/pull/5439#issuecomment-1758777522, the new force-destake test was failing in the downstream job from the monorepo.

This is happening because the test is serializing 196 bytes for the stake account, rather than the expected 200 bytes. This works fine in 1.16, where we don't use the extra 4 bytes. In 1.17 and 1.18, however, those extra 4 bytes matter for the stake flags.

#### Solution

Overallocate the vector used when serializing a `StakeState`.

### DIFF
--- a/stake-pool/program/tests/force_destake.rs
+++ b/stake-pool/program/tests/force_destake.rs
@@ -41,10 +41,12 @@ async fn setup(
     let (mut stake_pool, mut validator_list) = stake_pool_accounts.state();
 
     let _ = add_vote_account_with_pubkey(voter_pubkey, &mut program_test);
+    let mut data = vec![0; std::mem::size_of::<StakeState>()];
+    bincode::serialize_into(&mut data[..], forced_stake).unwrap();
 
     let stake_account = Account::create(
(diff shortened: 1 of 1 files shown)


2501babe (approved):

2501babe:
do i need to worry about breaking monorepo with this?
```
    // deactivate the pool stake account
    let (meta, stake, _) =
        get_stake_account(&mut context.banks_client, &accounts.stake_account).await;
    let delegation = Delegation {
        activation_epoch: 0,
        deactivation_epoch: 0,
        ..stake.unwrap().delegation
    };
    let account_data = serialize(&StakeState::Stake(
        meta,
        Stake {
            delegation,
            ..stake.unwrap()
        },
    ))
    .unwrap();
 ```

in #5488 ? im a little confused why yours was broken, the downstream tests patch spl to the current monorepo version, so shouldnt `StakeState` have the flag space on it when you create it?

PR #5540, opened 2023-10-15
[token-cli] Enable configuring confidential transfer account

samkim-crypto:
#### Problem
The token-cli does not yet support initializing confidential transfer account

#### Summary of changes
Added configuring cli commands for confidential transfer accounts. I think each commits should be pretty straightforward and independent with the following note:
- For `configure-confidential-transfer-account`, I made it so that the elgamal and aes keys are derived from the signer by default. This can be updated so that users have the option to initialize the account with custom keys once we upgrade to clap-v3-utils.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -2966,24 +2966,6 @@ async fn command_enable_disable_confidential_transfers(
         );
     }
 
-    // Reallocation (if needed)
-    let mut existing_extensions: Vec<ExtensionType> = state_with_extension.get_extension_types()?;
-    if !existing_extensions.contains(&ExtensionType::ConfidentialTransferAccount) {
-        existing_extensions.push(ExtensionType::ConfidentialTransferAccount);
-        let needed_account_len =
-            ExtensionType::try_calculate_account_len::<Account>(&existing_extensions)?;
-        if needed_account_len > current_account_len {
-            token
-                .reallocate(
-                    &token_account_address,
-                    &owner,
-                    &[ExtensionType::ConfidentialTransferAccount],
-                    &bulk_signers,
-                )
-                .await?;
-        }
-    }
-
     let res = if let Some(allow_confidential_credits) = allow_confidential_credits {
         let extension_state = state_with_extension
             .get_extension::<ConfidentialTransferAccount>()?
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -38,7 +38,7 @@ use solana_sdk::{
 use spl_associated_token_account::get_associated_token_address_with_program_id;
 use spl_token_2022::{
     extension::{
-        confidential_transfer::ConfidentialTransferMint,
+        confidential_transfer::{ConfidentialTransferAccount, ConfidentialTransferMint},
         confidential_transfer_fee::ConfidentialTransferFeeConfig,
         cpi_guard::CpiGuard,
         default_account_state::DefaultAccountState,
@@ -52,7 +52,10 @@ use spl_token_2022::{
         BaseStateWithExtensions, ExtensionType, StateWithExtensionsOwned,
     },
     instruction::*,
-    solana_zk_token_sdk::zk_token_elgamal::pod::ElGamalPubkey,
+    solana_zk_token_sdk::{
+        encryption::{auth_encryption::AeKey, elgamal::ElGamalKeypair},
+        zk_token_elgamal::pod::ElGamalPubkey,
+    },
     state::{Account, AccountState, Mint},
 };
 use spl_token_client::{
@@ -166,6 +169,7 @@ pub enum CommandName {
     InitializeMetadata,
     UpdateMetadata,
     UpdateConfidentialTransferSettings,
+    ConfigureConfidentialTransferAccount,
 }
 impl fmt::Display for CommandName {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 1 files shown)


token/cli/src/main.rs
```diff
@@ -4459,6 +4651,109 @@ fn app<'a, 'b>(
...
+                .about("Configure confidential transfers for token account")
+                .arg(
+                    Arg::with_name("account")
+                        .validator(is_valid_pubkey)
+                        .value_name("TOKEN_ACCOUNT_ADDRESS")
+                        .takes_value(true)
+                        .index(1)
+                        .required(true)
+                        .help("The address of the token account to configure confidential transfers for")
+                )
```
joncinque, line 4665: For all of these, similar to `create-account`, how about taking in a `"mint"` and using the owner's ATA for that mint if `"account"` isn't specified?

token/cli/src/main.rs
```diff
@@ -5308,6 +5603,125 @@ async fn process_command<'a>(
...
+
+            if config.multisigner_pubkeys.is_empty() {
+                push_signer_with_dedup(owner_signer, &mut bulk_signers);
+            }
+
+            let maximum_credit_counter = value_t!(
+                arg_matches.value_of("maximum_pending_balance_credit_counter"),
+                u64
+            )
+            .ok();
```
joncinque, line 5628: Unless I'm missing something, if someone puts in an invalid number here, it'll silently fallback to the default. You'll probably want to properly handle the error if the arg is present, ie:

```
arg_matches.value_of("maximum_pending_balance_credit_counter").map(|v| v.parse::<u64>().unwrap_or_else(print_error_and_exit))

File: token/cli/src/main.rs [hunk 3]
```diff
@@ -5308,6 +5603,125 @@ async fn process_command<'a>(
...
+                token_account,
+                owner,
+                maximum_credit_counter,
+                &elgamal_keypair,
+                &aes_key,
+                bulk_signers,
+            )
+            .await
+        }
+        (CommandName::EnableConfidentialCredits, arg_matches) => {
```
joncinque, line 5641: nit: You could probably combine all of these by doing:

```
(c @ CommandName::EnableConfidentialCredits, arg_matches) |
(c @ CommandName::DisableConfidentialCredits, arg_matches) |
...
```

Then

```
let (allow_confidential_credits, allow_non_confidential_credits) = match c {
    CommandName::EnableConfidentialCredits => (Some(true), None),
    ...
};
```

token/cli/src/main.rs
```diff
@@ -2875,6 +2880,193 @@ async fn command_update_confidential_transfer_settings(
...
+            token
+                .reallocate(
+                    &token_account_address,
+                    &owner,
+                    &[ExtensionType::ConfidentialTransferAccount],
+                    &bulk_signers,
+                )
+                .await?;
+        }
+    }
```
joncinque, line 2985: Considering the command errors if it doesn't already contain the extension right before this, it's probably best to remove this logic since it'll never get hit

joncinque (commented):
Looks great! Just a few little comments, but this should be ready to go soon

token/cli/src/main.rs
samkim-crypto: Oops, thanks for the catch! Removed.

token/cli/src/main.rs
samkim-crypto: Oh I think I never used match bindings before. I learned something new. Thanks!

token/cli/src/main.rs
samkim-crypto: Yes, this is a good point. In clap-v3, this will panic, but in clap-v2, this will set `maximum_credit_counter` to default. Fixed.

token/cli/src/main.rs
samkim-crypto: Yeah, that is a good idea. It turned out it is slightly clunky since `account` and `token` can all be optional. Please take a look. I think this suggestion also applies for all other commands that take in `account` as a required argument.

joncinque (commented):
Sorry, one last bit to make these flags consistent with other commands

token/cli/src/main.rs
```diff
@@ -4459,6 +4653,144 @@ fn app<'a, 'b>(
...
+                .arg(
+                    owner_address_arg()
+                )
+                .arg(
+                    Arg::with_name("token")
+                        .long("token")
+                        .validator(is_valid_pubkey)
+                        .value_name("TOKEN_MINT_ADDRESS")
+                        .takes_value(true)
+                        .help("The token address with confidential transfers enabled"),
```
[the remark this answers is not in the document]
joncinque: Sorry it wasn't clear -- can you make these changes in all of the new instructions?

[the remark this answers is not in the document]
samkim-crypto: Oh my mistake! I updated the other commands.

token/cli/src/main.rs
```diff
@@ -4459,6 +4653,164 @@ fn app<'a, 'b>(
...
+                )
+                .arg(
+                    Arg::with_name("address")
+                        .long("address")
+                        .validator(is_valid_pubkey)
+                        .value_name("TOKEN_ACCOUNT_ADDRESS")
+                        .takes_value(true)
+                        .conflicts_with("token")
+                        .help("The address of the token account to configure confidential transfers for \
```
joncinque, line 4718:
```suggestion
                        .help("The address of the token account to enable confidential transfers for \
```

token/cli/src/main.rs
```diff
@@ -4459,6 +4653,164 @@ fn app<'a, 'b>(
...
```
joncinque, line 4747:
```suggestion
                        .help("The address of the token account to disable confidential transfers for \
```

token/cli/src/main.rs
```diff
@@ -4459,6 +4653,164 @@ fn app<'a, 'b>(
...
```
joncinque, line 4776:
```suggestion
                        .help("The address of the token account to enable non-confidential transfers for \
```

token/cli/src/main.rs
```diff
@@ -4459,6 +4653,164 @@ fn app<'a, 'b>(
...
```
joncinque, line 4805:
```suggestion
                        .help("The address of the token account to disable non-confidential transfers for \
```

joncinque (approved):
I noticed a few of the help texts got changed, so just changing those back. After that, this is good to go!

samkim-crypto:
Oh man sorry for all the mess and hassle this PR and thanks for your patience.

joncinque (approved):

PR #5510, opened 2023-10-11
feature gate: init program with `RevokePendingActivation` instruction

buffalojoec:
This PR is the first in a series of re-written commit history to facilitate easier discussion on the Feature Gate Program, as [SIMD 0066]() progresses.

Original PR: #4987 

Notable changes:
- Condensed feedback to one commit, defining the "base" program
- Only defines the `RevokePendingActivation` instruction
- Removed CLI (for now)

PR #5541, opened 2023-10-16
feature gate: add activate instruction

buffalojoec:
This PR adds the `ActivateFeature` instruction to the Feature Gate program.

This instruction is the activation process via feature keypair that we are familiar with today.

The next PR in this stack will address [this comment](https://github.com/solana-labs/solana-program-library/pull/5510#discussion_r1357539741) on #5510 with an instruction designed for multi-sig support.

### DIFF
--- a/feature-gate/README.md
+++ b/feature-gate/README.md
@@ -2,7 +2,8 @@
 
 This program serves to manage new features on Solana.
 
-It serves one purpose: revoking features that are pending activation.
+It serves two main purposes: activating new features and revoking features that
+are pending activation.
 
 More information & documentation will follow as this program matures, but you
 can follow the discussions
--- a/feature-gate/README.md
+++ b/feature-gate/README.md
@@ -2,9 +2,9 @@
 
 This program serves to manage new features on Solana.
 
-It serves two main purposes: activating new features and revoking features that
-are pending activation.
+It serves two main purposes: queuing new features for activation and revoking
+features that are already queued and pending activation.
 
 More information & documentation will follow as this program matures, but you
 can follow the discussions
-[here](https://github.com/solana-labs/solana/issues/32780)!
+[here](https://github.com/solana-foundation/solana-improvement-documents/pull/72)!
--- a/feature-gate/program/src/instruction.rs
+++ b/feature-gate/program/src/instruction.rs
@@ -16,7 +16,8 @@ use {
 #[derive(Clone, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
 #[repr(u8)]
 pub enum FeatureGateInstruction {
-    /// Submit a feature for activation.
+    /// Queue a feature for activation by allocating and assigning a feature
+    /// account.
     ///
     /// Note: This instruction expects the account to exist and be owned by the
     /// system program. The account should also have enough rent-exempt lamports
--- a/feature-gate/program/src/instruction.rs
+++ b/feature-gate/program/src/instruction.rs
@@ -3,12 +3,10 @@
 use {
     num_enum::{IntoPrimitive, TryFromPrimitive},
     solana_program::{
-        feature::Feature,
         instruction::{AccountMeta, Instruction},
         program_error::ProgramError,
         pubkey::Pubkey,
-        rent::Rent,
-        system_instruction, system_program,
+        system_program,
     },
 };
 
@@ -19,16 +17,14 @@ pub enum FeatureGateInstruction {
     /// Queue a feature for activation by allocating and assigning a feature
     /// account.
     ///
-    /// Note: This instruction expects the account to exist and be owned by the
-    /// system program. The account should also have enough rent-exempt lamports
-    /// to cover the cost of the account creation for a
-    /// `solana_program::feature::Feature` state prior to invoking this
-    /// instruction.
+    /// Note: This instruction expects the account to be owned by the system
+    /// program.
     ///
     /// Accounts expected by this instruction:
     ///
     ///   0. `[w+s]`    Feature account (must be a system account)
-    ///   1. `[]`       System program
+    ///   1. `[w+s]`    Payer (for rent lamports)
+    ///   2. `[]`       System program
     ActivateFeature,
     /// Revoke a pending feature activation.
     ///
@@ -62,9 +58,10 @@ impl FeatureGateInstruction {
 }
 
 /// Creates an 'ActivateFeature' instruction.
-pub fn activate_feature(feature_id: &Pubkey) -> Instruction {
+pub fn activate_feature(feature_id: &Pubkey, payer: &Pubkey) -> Instruction {
     let accounts = vec![
         AccountMeta::new(*feature_id, true),
+        AccountMeta::new(*payer, true),
         AccountMeta::new_readonly(system_program::id(), false),
     ];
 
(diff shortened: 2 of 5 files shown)


buffalojoec:
* master
  * **PR #5510** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5510" target="_blank">Graphite</a> 
    * **PR #5541** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5541" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5541?utm_source=stack-comment).

feature-gate/program/src/feature_id.rs
```diff
@@ -0,0 +1,37 @@
...
+// &crate::id()).0;     while feature_exists(&feature_id) {
+//         nonce.checked_add(1).ok_or(FeatureGateError::Overflow)?;
+//         feature_id = Pubkey::find_program_address(
+//             &[b"feature", &nonce.to_le_bytes(), &authority.to_bytes()],
+//             &crate::id(),
+//         )
+//         .0;
+//     }
+//     Ok(feature_id)
+// }
```
buffalojoec, line 27: I _really_ wish this would work, but a program can't import `solana_sdk` :(

Any other ideas?

feature-gate/program/src/feature_id.rs
```diff
@@ -0,0 +1,37 @@
...
+//     while feature_exists(&feature_id) {
+//         nonce.checked_add(1).ok_or(FeatureGateError::Overflow)?;
+//         feature_id = Pubkey::find_program_address(
+//             &[b"feature", &nonce.to_le_bytes(), &authority.to_bytes()],
+//             &crate::id(),
+//         )
+//         .0;
+//     }
+//     Ok(feature_id)
+// }
```
buffalojoec, line 27: I _really_ wish this would work, but a program can't import solana_sdk :(

Any other ideas @joncinque  ?

CriesofCarrots:
If possible, I think we should continue to support the current process, where a single signer is the feature account, so that we can upgrade to this version of the program without making engs immediately change process to having to derive an address to put in the feature_set list. Make `authority` optional and only derive the address when Some?

feature-gate/program/src/feature_id.rs
CriesofCarrots: It also won't work because once an eng calls `derive_feature_id()` once and puts an entry in FEATURE_NAMES to start using the gate for runtime changes, `derive_feature_id()` won't return that same key again for the actual activation.
Probably have to let authorities handle their own noncing or seed determination in some fashion...?

buffalojoec:
> If possible, I think we should continue to support the current process, where a single signer is the feature account, so that we can upgrade to this version of the program without making engs immediately change process to having to derive an address to put in the feature_set list. Make `authority` optional and only derive the address when Some?

Yeah that's totally possible!

I'm going to re-write this PR to be the activate instruction as we currently know it, so nothing changes for contributors. Then I'll introduce another with the multi-sig support and we can check them both out in tandem.

feature-gate/program/src/instruction.rs
```diff
@@ -3,16 +3,32 @@
...
+        rent::Rent,
+        system_instruction, system_program,
     },
 };
 
 /// Feature Gate program instructions
 #[derive(Clone, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
 #[repr(u8)]
 pub enum FeatureGateInstruction {
+    /// Submit a feature for activation.
```
CriesofCarrots, line 19:
```suggestion
    /// Queue a feature for activation by allocating and assigning a feature account.
```

feature-gate/README.md
```diff
@@ -2,7 +2,8 @@
 
 This program serves to manage new features on Solana.
 
-It serves one purpose: revoking features that are pending activation.
+It serves two main purposes: activating new features and revoking features that
+are pending activation.
```
CriesofCarrots, line 6:
```suggestion
It serves two main purposes: queuing new features for activation and revoking features that
are already queued and pending activation.
```
I'm not totally happy with my suggestion here (and I don't love adding another new term: "queued"), but I'm trying to express that the new instruction sets a feature to "pending". Thoughts?

feature-gate/program/src/instruction.rs
```diff
@@ -44,6 +60,36 @@ impl FeatureGateInstruction {
...
+/// Creates a set of two instructions:
+///   * One to fund the feature account with rent-exempt lamports
+///   * Another is the Feature Gate Program's 'ActivateFeature' instruction
+pub fn activate_feature_with_rent_transfer(
+    feature_id: &Pubkey,
+    payer: &Pubkey,
+) -> [Instruction; 2] {
+    let lamports = Rent::default().minimum_balance(Feature::size_of());
+    [
+        system_instruction::transfer(payer, feature_id, lamports),
```
CriesofCarrots, line 88: I think we should move transfer handling into `process_activate_feature()`. That way we can use the Rent syscall to be sure we're using correct rates. `Rent::default()` may not always be accurate.
Would be something like what we do in spl-assocated-token: https://github.com/solana-labs/solana-program-library/blob/5ce6eace7c128c3eb5686e02de983ee431b73ce1/associated-token-account/program/src/tools/account.rs#L28-L42

Unless there's a reason I'm missing that the transfer needs to be separate?

feature-gate/program/src/processor.rs
```diff
@@ -7,12 +7,41 @@ use {
...
+    let _system_program_info = next_account_info(account_info_iter)?;
+
+    if !feature_info.is_signer {
+        return Err(ProgramError::MissingRequiredSignature);
+    }
+
+    if feature_info.owner != &system_program::id() {
+        return Err(FeatureGateError::InvalidFeatureAccount.into());
+    }
+
```
CriesofCarrots, line 32: I'm advocating for moving the transfer into this processor, but if we don't we definitely need to check the lamport balance here.

feature-gate/README.md
buffalojoec: Yeah despite the fact it's still doing the same thing until SIMD 0066, I agree we should take the opportunity to describe it more clearly.

I think "queue" is the best I can think of in this context as well, so I'm good with it!

feature-gate/program/src/instruction.rs
buffalojoec: I was following Token2022's example with how account initialization is handled, specifically initializing token metadata:

https://github.com/solana-labs/solana-program-library/blob/acabef3680e972baf8e9a47531c35eaa45405740/token/client/src/token.rs#L3614-L3637

https://github.com/solana-labs/solana-program-library/blob/acabef3680e972baf8e9a47531c35eaa45405740/token/client/src/token.rs#L3657-L3695

But in hindsight I'm thinking that might be the process for some of the Token2022 stuff since the account in question _may_ or _may not_ exist prior to the instruction (?).

Considering these feature accounts will not exist prior to sending the instruction, and our CLI command `solana feature activate` will always use the "with rent" one, I think it makes sense to combine in the program's processor, yep.

buffalojoec: Note this also requires a payer though

buffalojoec: I've added a commit for this. If we decide we don't like it I can always pluck it!

buffalojoec:
Closing for now, since whenever SIMD talks conclude, we'll want to upgrade the program one feature at a time, starting with the existing `RevokePendingActivation` instruction.

PR #5560, opened 2023-10-16
single-pool: rename program to spl_single_pool

2501babe:
actually for real final pr before i tag my 1.0 versions and release... this is so we can be `spl-single-pool` and `spl-single-pool-cli` on crates.io. ive removed all uses of "single-validator pool," officially we are "single-pool" or "svsp" for short and "single-validator stake pool" in full

once this is in i will:
* make three tags and three releases for program, cli, js (js-classic depends on js-modern so these should advance in lockstep, hence no fourth release)
* cargo and npm publish
* update docs with npm url and reactivate example and merge #5414 
* program on devnet!!!

ill hold the devnet upgrade authority for now, we need to figure out who will ultimately be the uprgrader for mainnet before we deploy there. but the finish line is in sight

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6874,6 +6874,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "spl-single-pool"
+version = "1.0.0"
+dependencies = [
+ "approx",
+ "arrayref",
+ "bincode",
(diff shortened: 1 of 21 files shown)


joncinque (approved):
Looks great!

PR #5594, opened 2023-10-18
single-pool-js: Fix ci

joncinque:
#### Problem

The single-pool js tests are failing in CI, e..g https://github.com/solana-labs/solana-program-library/actions/runs/6560481835/job/17818975411

#### Solution

It was fiddly, but I needed to make two changes -- one to use the `import` flag instead of `loader`, and then to use the common js loader instead of esm. Who knows why this works? Not me! :upside_down_face:

### DIFF
--- a/single-pool/js/packages/classic/package.json
+++ b/single-pool/js/packages/classic/package.json
@@ -5,7 +5,7 @@
     "build": "pnpm tsc",
     "lint": "eslint --max-warnings 0 .",
     "lint:fix": "eslint . --fix",
-    "test": "NODE_OPTIONS='--loader=tsx' ava"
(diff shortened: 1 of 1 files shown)


2501babe:
uhh that's really weird... im just gonna check what happens without the commonjs change in a separate pr because i have no idea why it would want that, considering files are... not cjs

2501babe:
ok nevermind I reproduced locally. god this is infuriating. the loader flag is necessary to make everything use ts, but the import flag doesnt actually do what the loader flag is supposed to do, and they removed the loader flag. i hate this platform so much and i also dont care either, if it works it works

joncinque:
Ah nice job tracking that down! Looks like I got frustrated and quit sooner than you

PR #5596, opened 2023-10-18
token 2022: make `IncorrectMintAuthority` a `TokenError`

buffalojoec:
This PR makes `IncorrectMintAuthority` a `TokenError`, rather than a `TokenMetadataError`, since it will likely be used across multiple extension processors (such as SPL Token Group).

PR #5595, opened 2023-10-18
token 2022: move update authority check to shared file

buffalojoec:
This PR moves the `check_update_authority` function into its own module, for use across multiple extention processors.

PR #5593, opened 2023-10-18
token 2022: add support for groups

buffalojoec:
This PR adds support to Token2022 for token groups, made possible by implementing the SPL Token Group interface.

With this change, developers can use Token2022 for groups of tokens, such as NFT Collections and Editions.

PR #5549, opened 2023-10-16
token-group: create editions example

buffalojoec:
This PR adds another use case for the SPL Token Group example program: Editions!

It demonstrates using the interface to print "reprints" of an original token's metadata, using the Token Group interface to mark these reprints as part of the original group.

PR #5601, opened 2023-10-18
token 2022: add `InitializeGroup` instruction from SPL Token Group interface

buffalojoec:
This PR adds the `InitializeGroup` instruction to Token2022 - from the SPL Token Group interface.


buffalojoec:
* master
  * **PR #5512** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5512" target="_blank">Graphite</a> 
        * **PR #5595** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5595" target="_blank">Graphite</a> 
          * **PR #5596** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5596" target="_blank">Graphite</a> 
            * **PR #5601** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5601" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5601?utm_source=stack-comment).

token/client/src/token.rs
```diff
@@ -3846,4 +3846,76 @@ where
...
+        let account_lamports = account.lamports;
+        let mint_state = self.unpack_mint_info(account)?;
+        let new_account_len = mint_state
+            .try_get_account_len()?
+            .checked_add(ExtensionType::try_calculate_account_len::<Mint>(&[
+                ExtensionType::TokenGroup,
+            ])?)
+            .ok_or(TokenError::Program(
+                spl_token_2022::error::TokenError::Overflow.into(),
+            ))?;
```
buffalojoec, line 3883: Is this really the way to do this for fixed-length types or am I missing something?

token/program-2022/src/extension/token_group/processor.rs
```diff
@@ -0,0 +1,101 @@
...
+        ExtensionType::try_calculate_account_len::<Mint>(&[ExtensionType::TokenGroup])?;
+    let new_account_len = group_info
+        .data_len()
+        .checked_add(group_extension_len)
+        .ok_or(TokenError::Overflow)?;
+    group_info.realloc(new_account_len, false)?;
+    let mut buffer = group_info.try_borrow_mut_data()?;
+    let mut state = StateWithExtensionsMut::<Mint>::unpack(&mut buffer)?;
+    let group = state.init_extension::<TokenGroup>(false)?;
+    *group = TokenGroup::new(mint_info.key, data.update_authority, data.max_size.into());
```
buffalojoec, line 83: This _can't_ be the right way to do this 🫠

token/program-2022/src/extension/token_group/processor.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    // Reallocate the mint for the new extension
+    realloc_mint(mint_info, ExtensionType::TokenGroup)?;
+
+    // Allocate a TLV entry for the space and write it in
+    // Assumes that there's enough SOL for the new rent-exemption
+    let mut mint_data = mint_info.try_borrow_mut_data()?;
+    let mut mint = StateWithExtensionsMut::<Mint>::unpack(&mut mint_data)?;
+    let group = mint.init_extension::<TokenGroup>(false)?;
+    *group = TokenGroup::new(mint_info.key, data.update_authority, data.max_size.into());
```
buffalojoec, line 89: @joncinque I'm currently debugging why this bit of code doesn't work properly only _some of the time_. I'm wondering if it's got anything to do with the `realloc` involved here.

I'm expecting `init_extension(..)` to fail if the extension exists, and most of the time it does, but every so often (particularly in CI) it doesn't fail, and my test expecting `ExtensionAlreadyInitialized` fails.

Any ideas lmk.

token/program-2022/src/extension/token_group/processor.rs
```diff
@@ -0,0 +1,107 @@
...
+
+fn realloc_mint(mint_info: &AccountInfo, extension: ExtensionType) -> Result<(), ProgramError> {
+    let extension_len = extension.try_get_tlv_len()?;
+    let new_account_len = mint_info
+        .data_len()
+        .checked_add(extension_len)
+        .ok_or::<ProgramError>(TokenError::Overflow.into())?;
+    mint_info.realloc(new_account_len, false)?;
+    Ok(())
+}
```
joncinque, line 36: This function makes a potentially dangerous assumption, that the account buffer is fully used. I think we're typically very careful to avoid any overallocations, but this does worry me a little bit.

Rather than doing the realloc at this level, where we're adding more bytes without knowing how much of the buffer is used, how about adding a version of `alloc_and_serialize` that works for fixed-length extensions?

token/program-2022/src/extension/mod.rs
```diff
@@ -907,10 +910,12 @@ pub enum ExtensionType {
     MetadataPointer,
     /// Mint contains token-metadata
     TokenMetadata,
-    /// Test variable-length mint extension
     /// Mint contains a pointer to another account (or the same account) that holds group
     /// configurations
     GroupPointer,
+    /// Mint contains token group configurations
+    TokenGroup,
+    /// Test variable-length mint extension
```
joncinque, line 918: Oops, I totally missed that one in the last review :disappointed:

token/program-2022-test/tests/token_group_initialize.rs
```diff
@@ -0,0 +1,271 @@
...
+        .unwrap_err();
+    assert_eq!(
+        error,
+        TokenClientError::Client(Box::new(TransportError::TransactionError(
+            TransactionError::InstructionError(
+                1,
+                InstructionError::Custom(TokenError::ExtensionAlreadyInitialized as u32)
+            )
+        )))
+    );
```
joncinque, line 147: This is probably failing in CI because it looks like the exact same transaction it's already seen. This is a strange case though, because it should be tacking on a transfer for a different amount :thinking: but either way, to make it more secure, just change `max_size` to some other number, and that way you're guaranteed to get a unique transaction.

joncinque (commented):
Looks great for the most part! Just the one bit about realloc, wondering if we can tighten that up some more

token/program-2022/src/extension/mod.rs
buffalojoec: Ngl I thought I rebased it to the previous commit lol.

token/program-2022-test/tests/token_group_initialize.rs
buffalojoec: Actually the amount would be the same, since it's not checking if the extension exists in the client-side function. I'm not sure if we want to add that to the client, or just let the program handle that, but since `TokenGroup` is fixed it will always be the same rent each time that's called.

Still seems weird that it would work about 50% of the time, but maybe it's all about timing? 😅

token/program-2022/src/extension/token_group/processor.rs
buffalojoec: Yeah, will do!

token/program-2022-test/tests/token_group_initialize.rs
joncinque: Ah, there's your answer then! We should probably smarten that up. And yeah, especially in CI, you'll often get the results of a previous transaction if it has the same signature as a previous one.

token/client/src/token.rs
```diff
@@ -3846,4 +3847,77 @@ where
...
+            .saturating_add(size_of::<ExtensionType>())
+            .saturating_add(size_of::<Length>())
+            .saturating_add(size_of::<V>());
+        let new_rent_exempt_minimum = self
+            .client
+            .get_minimum_balance_for_rent_exemption(new_account_len)
+            .await
+            .map_err(TokenError::Client)?;
+        Ok(new_rent_exempt_minimum.saturating_sub(account_lamports))
+    }
```
joncinque, line 3890: Since you're already fetching the account here, you can do a quick to see if the account already has the extension, and return 0 if it does.

buffalojoec: Yep totally, I just wasn't sure how useful it would be since it's basically like saying "hey you don't need any rent because the extension exists, but the program is going to barf on you when this goes through".

buffalojoec: But this call will return the program's error so I guess the 0 approach is fine.

token/program-2022/src/extension/token_group/processor.rs
buffalojoec: Done in #5672 !

token/client/src/token.rs
```diff
@@ -3649,4 +3650,77 @@ where
...
+    async fn get_additional_rent_for_fixed_len_extension<V: Extension + Pod>(
+        &self,
+    ) -> TokenResult<u64> {
+        let account = self.get_account(self.pubkey).await?;
+        let account_lamports = account.lamports;
+        let mint_state = self.unpack_mint_info(account)?;
+        if mint_state.get_extension::<V>().is_ok() {
+            Ok(0)
+        } else {
+            let new_account_len = mint_state.try_get_new_account_len::<V>()?;
```
joncinque, line 3685: very neat :sunglasses:

joncinque (approved):
Looks great! Let's get the groups going

group

token/client/src/token.rs
```diff
@@ -3649,4 +3650,77 @@ where
...
+                .client
+                .get_minimum_balance_for_rent_exemption(new_account_len)
+                .await
+                .map_err(TokenError::Client)?;
+            Ok(new_rent_exempt_minimum.saturating_sub(account_lamports))
+        }
+    }
+
+    /// Initialize token-group on a mint
+    #[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
buffalojoec: What the 🤨
I think I had another arg in here or something. Changed!

joncinque (approved):

PR #5624, opened 2023-10-20
token-2022: Use serde with `rename_all = "camelCase"`

joncinque:
#### Problem

#5437 did some great work with deriving the serde traits, but we forgot to specify `rename_all = "camelCase"` as a separate attribute, which means that we'd get some inconsistency in the serialization.

#### Solution

Just add the attribute config. There were also a couple of transfer-fee structs that didn't have it, so I added those.


buffalojoec (approved):
Lgtm mate!

PR #5627, opened 2023-10-21
token 2022: add `UpdateGroupAuthority` instruction from SPL Token Group interface

buffalojoec:
This PR adds the `UpdateGroupAuthority` instruction to Token2022 - from the SPL Token Group interface.

PR #5621, opened 2023-10-20
token 2022: add `GroupPointer` extension

buffalojoec:
This PR adds the `GroupPointer` extension to Token2022.

This extension behaves exactly like the existing `MetadataPointer` extension, serving to denote the corresponding account containing a mint's group configurations and prevent any unauthorized configurations.

PR #5616, opened 2023-10-20
token group: add mint to interface

buffalojoec:
I realized I didn't add the mint as a required field in the state of a `TokenGroup` or `TokenGroupMember`.

This PR adds the mint!

PR #5628, opened 2023-10-22
token 2022: add `GroupMemberPointer` extension

buffalojoec:
This PR adds the `GroupMemberPointer` extension to Token2022.

This extension behaves exactly like the existing `GroupPointer` extension, serving to denote the corresponding account containing a mint's group member configurations and prevent any unauthorized configurations.


buffalojoec:
* master
  * **PR #5616** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5616" target="_blank">Graphite</a> 
      * **PR #5621** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5621" target="_blank">Graphite</a> 
              * **PR #5628** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5628" target="_blank">Graphite</a>  👈
      * **PR #5614** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5614" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5628?utm_source=stack-comment).

token/program-2022/src/extension/group_pointer/processor.rs
```diff
@@ -47,6 +44,9 @@ fn process_initialize(
         );
         Err(TokenError::InvalidInstruction)?;
     }
+
+    let extension = mint.init_extension::<GroupPointer>(true)?;
+    extension.authority = *authority;
```
buffalojoec, line 49: Just moved this down here for uniformity!

buffalojoec (commented):
Ok @joncinque this should be ready for another look. Thanks for the previous review! RIP to repeating extensions 🪦

token/program-2022/src/extension/group_member_pointer/instruction.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///
+    /// The mint must have exactly enough space allocated for the base mint (82
+    /// bytes), plus 83 bytes of padding, 1 byte reserved for the account type,
+    /// then space required for this extension, plus any others.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   0. `[writable]` The mint to initialize.
+    ///   1. `[]`         The group mint.
+    ///   2. `[signer]`   The group's update authority.
```
joncinque, line 38: Why are these required? This should work just like the group pointer, right?

token/program-2022/src/extension/group_member_pointer/instruction.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// Update the group member pointer address. Only supported for mints that
+    /// include the `GroupMemberPointer` extension.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single authority
+    ///   0. `[writable]` The mint.
+    ///   1. `[]`         The group mint.
+    ///   2. `[signer]`   The mint's group member pointer authority.
+    ///   3. `[signer]`   The group's update authority.
```
joncinque, line 52: I think we just need the mint and group member pointer authority, right?

token/program-2022/src/extension/group_member_pointer/processor.rs
```diff
@@ -0,0 +1,145 @@
...
+    let group_mint = StateWithExtensions::<Mint>::unpack(&group_data)?;
+    let group_state = group_mint.get_extension::<TokenGroup>()?;
+    if Option::<Pubkey>::from(group_state.update_authority)
+        != Some(*group_update_authority_info.key)
+    {
+        msg!("Incorrect update authority for group");
+        Err(TokenError::OwnerMismatch)?;
+    }
+    Ok(())
+}
```
joncinque, line 47: This doesn't work with an external group program, where we can't figure out the authority, so we should probably get rid of the group authority requirement. Let's let any mint point at any member account, and it's up to clients to make sure that the accounts point at each other, same as with token metadata

token/program-2022-test/tests/group_member_pointer.rs
```diff
@@ -0,0 +1,554 @@
...
+    authority: &Pubkey,
+    group_update_authority: &Keypair,
+) -> TestContext {
+    let mut context = TestContext {
+        context,
+        token_context: None,
+    };
+    let member_address = Some(mint.pubkey());
+    context
+        .init_token_with_mint_keypair_and_freeze_authority_and_additional_signers(
```
joncinque, line 104: I don't think we'll need this new function if the group member pointer extension looks like metadata pointer / group pointer

joncinque (commented):
There's a few repeated comments, mostly on the interface and structs. Once we agree there, I'll take a closer look at the implementation and test

token/program-2022/src/extension/group_member_pointer/instruction.rs
joncinque: I don't agree with this -- for the other pointer extensions, the pointer is always allowed to point anywhere, and it's up to the other extension to do the proper check, or for clients to make sure that the accounts point to each other.

Also, by requiring the authority check, this extension becomes incompatible with a token group that is owned by another program, which makes the pointer part unnecessary.

buffalojoec: I think I see what you're saying. The responsibility of validating a pointer is up to the program adding the extension being pointed to. In other words:
- `MetadataPointer`: Whoever creates `TokenMetadata` (Token2022 extension or elsewhere)
- `GroupPointer`: Whoever creates `TokenGroup` (Token2022 extension or elsewhere)
- `GroupMemberPointer`: Whoever creates `TokenGroupMember` (Token2022 extension or elsewhere)

Those programs are then responsible for validating one pointer over all the others. I think I got confused when it came to two related extensions. But I think we're on the same page now.

buffalojoec:
@joncinque considering the discussion [here](https://github.com/solana-labs/solana-program-library/pull/5628#discussion_r1397692619) I'm going to re-work this PR and remove the group auth stuff.

buffalojoec:
Done. Note I deleted the prior branch in the stack, so I re-wrote the commit here.

token/program-2022/src/extension/group_member_pointer/instruction.rs
joncinque: I'd argue that the program being pointed to doesn't need to care either. It comes down to an agreement between two programs, which manifests as two accounts pointing at each other, ie a mint pointing at metadata, and metadata pointing at the mint. That's the *only* valid case, and everything else should be deemed invalid.

As metadata and collection programs proliferate, those programs can start doing whatever they want, but it's only "official" if the two point at each other.

You could make the argument that a metadata / collection program should validate the pointer before allowing the creation, but that could be limiting for some programs.

token/program-2022/src/extension/group_member_pointer/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+};
+
+/// Instructions for the GroupMemberPointer extension
+pub mod instruction;
+/// Instruction processor for the GroupMemberPointer extension
+pub mod processor;
+
+/// Group member pointer extension data for mints.
+#[repr(C)]
+#[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
```
joncinque, line 16: Don't forget the rename_all bit!
```suggestion
#[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
```

token/program-2022/src/extension/group_member_pointer/processor.rs
```diff
@@ -0,0 +1,103 @@
...
+    extension.member_address = *new_member_address;
+    Ok(())
+}
+
+pub(crate) fn process_instruction(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    input: &[u8],
+) -> ProgramResult {
+    check_program_account(program_id)?;
```
joncinque, line 87: I'm just noticing now that we do this in a few places, but I have no idea why -- we're checking that the executing program is token-2022, but that seems unnecessary. And it makes it annoying for other people to deploy their own version of the program. We can change in a separate PR, but it seems strange.

token/program-2022-test/tests/group_member_pointer.rs
```diff
@@ -0,0 +1,425 @@
...
+        .init_token_with_mint_keypair_and_freeze_authority(
+            Keypair::new(),
+            vec![ExtensionInitializationParams::GroupMemberPointer {
+                authority: None,
+                member_address: None,
+            }],
+            None,
+        )
+        .await
+        .unwrap_err();
```
joncinque, line 223: nit: this can be simplified
```suggestion
    let err = context
        .init_token_with_mint(
            vec![ExtensionInitializationParams::GroupMemberPointer {
                authority: None,
                member_address: None,
            }],
        )
        .await
        .unwrap_err();
```

token/program-2022-test/tests/group_member_pointer.rs
```diff
@@ -0,0 +1,425 @@
...
+        &group_update_authority.pubkey(),
+        &payer,
+    )
+    .await;
+
+    // fail with all none
+    let mut context = TestContext {
+        context,
+        token_context: None,
+    };
```
joncinque, line 212: nit: This can probably be simplified quite a bit too
```suggestion
    // fail with all none
    let mut context = TestContext::new().await;
```

token/program-2022-test/tests/group_member_pointer.rs
```diff
@@ -0,0 +1,425 @@
...
+            vec![ExtensionInitializationParams::GroupPointer {
+                authority: Some(*authority),
+                group_address,
+            }],
+            None,
+        )
+        .await
+        .unwrap();
+    context
+}
```
joncinque, line 65: nit: Is this function still needed?

token/program-2022-test/tests/group_member_pointer.rs
```diff
@@ -0,0 +1,425 @@
...
+        .token_group_initialize_with_rent_transfer(
+            &payer.pubkey(),
+            &mint_authority.pubkey(),
+            update_authority,
+            3,
+            &[&payer, &mint_authority],
+        )
+        .await
+        .unwrap();
+}
```
joncinque, line 84: nit: is this function still needed?

token/program-2022-test/tests/group_member_pointer.rs
```diff
@@ -0,0 +1,425 @@
...
+    .take()
+    .unwrap();
+
+    setup_token_group(
+        &group_token,
+        &group_token.mint_authority,
+        &group_update_authority.pubkey(),
+        &payer,
+    )
+    .await;
```
joncinque, line 146: Is any of this needed?

token/program-2022-test/tests/group_member_pointer.rs
```diff
@@ -0,0 +1,425 @@
...
```
joncinque, line 271: Is any of this needed?

token/program-2022-test/tests/group_member_pointer.rs
```diff
@@ -0,0 +1,425 @@
...
```
joncinque, line 362: Is this needed?

joncinque (commented):
The base logic looks good! Just one point on the code, and then a few places to simplify the tests

token/program-2022/src/extension/group_member_pointer/processor.rs
buffalojoec: It appears to be coupled to Token-2022 based on the variable name and comment, but technically it's only checking if the provided program ID matches the one generated by `declare_id!(..)`.

https://github.com/solana-labs/solana-program-library/blob/d6bf8540316aff8540700ba3a7f202a55f49cb4c/token/program-2022/src/lib.rs#L100-L106

So, this should work for someone who deploys their own version, since they'll want to change that anyway.

joncinque: It depends. If someone just wants to build the program from source and immediately deploy it, it won't work for certain instructions, which is really weird. Anyway, I put in https://github.com/solana-labs/solana-program-library/issues/5909 for it

joncinque (approved):
Ship it!

member

PR #5629, opened 2023-10-22
token 2022: add `InitializeMember` instruction from SPL Token Group interface

buffalojoec:
This PR adds the `InitializeMember` instruction to Token2022 - from the SPL Token Group interface.

### DIFF
--- a/token/program-2022-test/tests/token_group_initialize_member.rs
+++ b/token/program-2022-test/tests/token_group_initialize_member.rs
@@ -1,4 +1,4 @@
-#![cfg(feature = "test-sbf")]
+// #![cfg(feature = "test-sbf")]
 
 mod program_test;
 use {
@@ -81,7 +81,7 @@ async fn setup(group: SetupConfig, members: Vec<SetupConfig>) -> (TestContext, V
             &payer_pubkey,
             &group_token_context.mint_authority.pubkey(),
             &group_authority,
-            5,
+            2,
             &[&group_token_context.mint_authority],
         )
         .await
@@ -136,10 +136,21 @@ async fn success_initialize() {
         )
         .await
         .unwrap_err();
+    let member_index = if group_mint_keypair
+        .pubkey()
+        .cmp(&member1_mint_keypair.pubkey())
+        .is_le()
+    {
+        4
+    } else {
+        3
+    };
     assert_eq!(
         error,
         TokenClientError::Client(Box::new(TransportError::TransactionError(
-            TransactionError::InsufficientFundsForRent { account_index: 4 }
+            TransactionError::InsufficientFundsForRent {
+                account_index: member_index
+            }
         )))
     );
 
@@ -241,6 +252,11 @@ async fn success_initialize() {
     );
 
     // fail double-init
+    {
+        let mut context = member_contexts[0].context.lock().await;
+        context.get_new_latest_blockhash().await.unwrap();
+        context.get_new_latest_blockhash().await.unwrap();
+    }
     let error = member1_token_context
         .token
         .token_group_initialize_member(
@@ -261,7 +277,7 @@ async fn success_initialize() {
         )))
     );
 
-    // Now the others
+    // Now the second
     let member2_token_context = member_contexts[1].token_context.take().unwrap();
     member2_token_context
         .token
@@ -286,8 +302,9 @@ async fn success_initialize() {
         }
     );
 
+    // Third should fail on max size
     let member3_token_context = member_contexts[2].token_context.take().unwrap();
-    member3_token_context
+    let error = member3_token_context
         .token
         .token_group_initialize_member_with_rent_transfer(
             &payer_pubkey,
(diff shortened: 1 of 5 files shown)


buffalojoec:
* master
  * **PR #5616** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5616" target="_blank">Graphite</a> 
      * **PR #5621** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5621" target="_blank">Graphite</a> 
              * **PR #5628** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5628" target="_blank">Graphite</a> 
                * **PR #5629** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5629" target="_blank">Graphite</a>  👈
      * **PR #5614** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5614" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5629?utm_source=stack-comment).

buffalojoec (commented):
This one's ready for another look as well!

buffalojoec:
@joncinque I'm still debugging some weird behavior on this one. It seems like every once in a while, the transaction error for `InsufficientFundsForRent` throws on account at index 3 instead of 4.

Quite frankly, I'm not even sure why 4 is the more common index. The fee payer should be index 0, and the mint - which should be throwing the insufficient lamports error - is at index 1 on the instruction, or index 2 if we consider the fee payer being bumped to 0.

Marking as a draft until I can debug. Note, this may also affect other tests for `TokenMetadata::Initialize` and `TokenGroup::Initialize`.

token/program-2022-test/tests/token_group_initialize_member.rs
```diff
@@ -0,0 +1,587 @@
...
+    //         ],
+    //     )
+    //     .await
+    //     .unwrap_err();
+    // assert_eq!(
+    //     error,
+    //     TokenClientError::Client(Box::new(TransportError::TransactionError(
+    //         TransactionError::InsufficientFundsForRent { account_index: 1 }
+    //     )))
+    // );
```
joncinque, line 149: If you re-enable this, be sure to refresh the blockhash before trying your `fail double-init` case again, since you may get the error from this transaction instead.

token/program-2022-test/tests/token_group_initialize_member.rs
```diff
@@ -0,0 +1,587 @@
...
+    let member_bytes = mint_info.get_extension_bytes::<TokenGroupMember>().unwrap();
+    let fetched_member = pod_from_bytes::<TokenGroupMember>(member_bytes).unwrap();
+    assert_eq!(
+        fetched_member,
+        &TokenGroupMember {
+            mint: member3_mint_keypair.pubkey(),
+            group: group_mint_keypair.pubkey(),
+            member_number: 3.try_into().unwrap(),
+        }
+    );
```
joncinque, line 304: It would exercise more code if the test sets a max size of 2, which makes this fail. What do you think?

joncinque (commented):
Just a few small points, this is looking good!

Regarding your question: the issue likely has to do with the ordering of the pubkeys in the transaction -- since you're generating random keypairs, sometimes one account will appear before another one in the BTreeMap of compiled account keys. If you want to test it for sure, you can check the ordering of the keypairs that you generate, or maybe simpler, just use keypairs from static arrays like `[1; 32]`

buffalojoec:
> Regarding your question: the issue likely has to do with the ordering of the pubkeys in the transaction -- since you're generating random keypairs, sometimes one account will appear before another one in the BTreeMap of compiled account keys.

Are we at all concerned with this behavior? In other words, is it worth looking into ensuring the order of accounts is first dictated by role and then secondarily dictated by the order in which it appears in an instruction? As I type it out, probably not, but thought I'd ask.

joncinque:
> Are we at all concerned with this behavior? In other words, is it worth looking into ensuring the order of accounts is first dictated by role and then secondarily dictated by the order in which it appears in an instruction? As I type it out, probably not, but thought I'd ask.

The behavior is actually dictated in that way, and it's also sorted by pubkey. So if you have two writable accounts in a transaction, by definition the "lower" pubkey will appear first.

buffalojoec:
> > Are we at all concerned with this behavior? In other words, is it worth looking into ensuring the order of accounts is first dictated by role and then secondarily dictated by the order in which it appears in an instruction? As I type it out, probably not, but thought I'd ask.
> 
> The behavior is actually dictated in that way, and it's also sorted by pubkey. So if you have two writable accounts in a transaction, by definition the "lower" pubkey will appear first.

I think I'm specifically tunnel-visioning it around this use case I encountered. Even if I did attempt to change it to _something_ else, it wouldn't make debugging transaction errors like `InsufficientLamportsForRent` any easier. So forget it!

token/program-2022-test/tests/token_group_initialize_member.rs
```diff
@@ -0,0 +1,490 @@
+// #![cfg(feature = "test-sbf")]
```
joncinque, line 1: Was this a mistake? It looks like we do want this for `test-sbf`

joncinque (approved):
Just one last comment, but then it's good to go!

club

token/program-2022-test/tests/token_group_initialize_member.rs
buffalojoec: Ahh yeah, it doesn't give me IntelliSense with the feature flag on

joncinque: You should be able to configure it to enable crate features. At least, you can in vim :grin:

buffalojoec: I think it's purely procrastination 😭

buffalojoec:
### Merge activity

* **Nov 29, 12:50 PM**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5629).

PR #5630, opened 2023-10-23
[token-cli] Add confidential transfer deposit, withdraw, apply pending balance, and confidential transfer

samkim-crypto:
#### Problem
The token-cli does not yet support confidential transfer instructions to deposit and withdraw confidential transfer tokens, apply pending balance to available balance, and make a confidential transfer.

#### Summary of changes
Add the four commands for the four instructions listed above. For confidential transfers, the functionality is added into the existing `transfer` command instead of creating a dedicated command for confidential transfers.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -39,8 +39,8 @@ use spl_associated_token_account::get_associated_token_address_with_program_id;
 use spl_token_2022::{
     extension::{
         confidential_transfer::{
-            account_info::WithdrawAccountInfo, ConfidentialTransferAccount,
-            ConfidentialTransferMint,
+            account_info::{ApplyPendingBalanceAccountInfo, WithdrawAccountInfo},
+            ConfidentialTransferAccount, ConfidentialTransferMint,
         },
         confidential_transfer_fee::ConfidentialTransferFeeConfig,
         cpi_guard::CpiGuard,
@@ -3367,11 +3367,14 @@ async fn command_apply_pending_balance(
     let state_with_extension = StateWithExtensionsOwned::<Account>::unpack(account.data)?;
     let token = token_client_from_config(config, &state_with_extension.base.mint, None)?;
 
+    let extension_state = state_with_extension.get_extension::<ConfidentialTransferAccount>()?;
+    let account_info = ApplyPendingBalanceAccountInfo::new(extension_state);
+
     let res = token
         .confidential_transfer_apply_pending_balance(
             &token_account_address,
             &owner,
-            None,
+            Some(account_info),
             elgamal_keypair.secret(),
             aes_key,
             &bulk_signers,
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -1647,67 +1647,76 @@ async fn command_transfer(
     };
 
     // ...and, finally, the transfer
-    let res = if let Some(recipient_owner) = fundable_owner {
-        token
-            .create_recipient_associated_account_and_transfer(
-                &sender,
-                &recipient_token_account,
-                &recipient_owner,
-                &sender_owner,
-                transfer_balance,
-                maybe_fee,
-                &bulk_signers,
-            )
-            .await?
-    } else if maybe_fee.is_some() && confidential_transfer_args.is_none() {
-        token
-            .transfer_with_fee(
-                &sender,
-                &recipient_token_account,
-                &sender_owner,
-                transfer_balance,
-                maybe_fee.unwrap(),
-                &bulk_signers,
-            )
-            .await?
-    } else if maybe_fee.is_none() && confidential_transfer_args.is_some() {
-        // deserialize `pod` ElGamal pubkeys
-        let recipient_elgamal_pubkey: elgamal::ElGamalPubkey = recipient_elgamal_pubkey
-            .unwrap()
-            .try_into()
-            .expect("Invalid recipient ElGamal pubkey");
-        let auditor_elgamal_pubkey = auditor_elgamal_pubkey.map(|pubkey| {
-            let auditor_elgamal_pubkey: elgamal::ElGamalPubkey =
-                pubkey.try_into().expect("Invalid auditor ElGamal pubkey");
-            auditor_elgamal_pubkey
-        });
-        token
-            .confidential_transfer_transfer(
-                &sender,
-                &recipient_token_account,
-                &sender_owner,
-                None,
-                transfer_balance,
-                None,
-                &confidential_transfer_args.unwrap().sender_elgamal_keypair,
-                &confidential_transfer_args.unwrap().sender_aes_key,
-                &recipient_elgamal_pubkey,
-                auditor_elgamal_pubkey.as_ref(),
-                &bulk_signers,
-            )
-            .await?
-    } else if maybe_fee.is_some() && confidential_transfer_args.is_some() {
-        panic!("Confidential transfer with fee is not yet supported.");
-    } else {
-        token
-            .transfer(
-                &sender,
-                &recipient_token_account,
-                &sender_owner,
-                transfer_balance,
-                &bulk_signers,
-            )
-            .await?
+    let res = match (fundable_owner, maybe_fee, confidential_transfer_args) {
+        (Some(recipient_owner), None, None) => {
+            token
+                .create_recipient_associated_account_and_transfer(
+                    &sender,
+                    &recipient_token_account,
+                    &recipient_owner,
+                    &sender_owner,
+                    transfer_balance,
+                    maybe_fee,
+                    &bulk_signers,
+                )
+                .await?
+        }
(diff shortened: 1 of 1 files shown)


token/cli/src/main.rs
```diff
@@ -5739,6 +6198,94 @@ async fn process_command<'a>(
             )
             .await
         }
+        (c @ CommandName::DepositConfidentialTokens, arg_matches)
+        | (c @ CommandName::WithdrawConfidentialTokens, arg_matches) => {
+            let token = pubkey_of_signer(arg_matches, "token", &mut wallet_manager)
+                .unwrap()
+                .unwrap();
+            let amount = match arg_matches.value_of("amount").unwrap() {
+                "ALL" => None,
```
joncinque, line 6207: I can't wait until we can start using `Amount` from the clap-v3 utils!

token/cli/src/main.rs
```diff
@@ -4813,6 +5141,113 @@ fn app<'a, 'b>(
...
+            SubCommand::with_name(CommandName::WithdrawConfidentialTokens.into())
+                .about("Withdraw amounts for confidential transfers")
+                .arg(
+                    Arg::with_name("token")
+                        .long("token")
+                        .validator(is_valid_pubkey)
+                        .value_name("TOKEN_MINT_ADDRESS")
+                        .takes_value(true)
+                        .index(1)
+                        .required(true)
```
joncinque, line 5193: Same here, does this still work if no mint is provided? ie:
```suggestion
                        .required_unless("address")
```

token/cli/src/main.rs
```diff
@@ -4813,6 +5141,113 @@ fn app<'a, 'b>(
...
+            SubCommand::with_name(CommandName::DepositConfidentialTokens.into())
+                .about("Deposit amounts for confidential transfers")
+                .arg(
+                    Arg::with_name("token")
+                        .long("token")
+                        .validator(is_valid_pubkey)
+                        .value_name("TOKEN_MINT_ADDRESS")
+                        .takes_value(true)
+                        .index(1)
+                        .required(true)
```
joncinque, line 5154: Does this work without the mint provided? Ie:
```suggestion
                        .required_unless("address")
```
Or is there an issue because of `index(2)` on the amount parameter?

token/cli/src/main.rs
```diff
@@ -3071,6 +3181,208 @@ async fn command_enable_disable_confidential_transfers(
...
+                amount,
+                decimals,
+                None,
+                elgamal_keypair,
+                aes_key,
+                &bulk_signers,
+            )
+            .await?
+    } else {
+        panic!("Command not supported");
```
joncinque, line 3321: nit: You could do a `match` statement and not need this case!

token/cli/src/main.rs
```diff
@@ -3071,6 +3181,208 @@ async fn command_enable_disable_confidential_transfers(
...
+    let maybe_amount =
+        ui_amount.map(|ui_amount| spl_token::ui_amount_to_amount(ui_amount, mint_info.decimals));
+
+    // the amount we will deposit or withdraw, as a u64
+    let amount = if !config.sign_only && instruction_type == ConfidentialInstructionType::Deposit {
+        let current_balance = token
+            .get_account_info(&token_account_address)
+            .await?
+            .base
+            .amount;
```
joncinque, line 3248: The account was just fetched earlier as `state_with_extension` -- can you just re-use that here?

token/cli/src/main.rs
```diff
@@ -3071,6 +3181,208 @@ async fn command_enable_disable_confidential_transfers(
...
+        let aes_key = aes_key.expect("AES key must be provided");
+
+        token
+            .confidential_transfer_withdraw(
+                &token_account_address,
+                &owner,
+                None,
+                amount,
+                decimals,
+                None,
```
joncinque, line 3314: Can you pass the `WithdrawAccountInfo` here based on `state_with_extension` that was fetched earlier to avoid a second fetch?

token/cli/src/main.rs
```diff
@@ -8456,6 +9003,87 @@ mod tests {
...
+                "spl-token",
+                CommandName::ConfigureConfidentialTransferAccount.into(),
+                "--address",
+                &destination_account.to_string(),
+            ],
+        )
+        .await
+        .unwrap(); // configure destination account for confidential transfers first
+
+        // NOTE: the test fails due to transaction size limit
```
joncinque, line 9051: The next step is to add support for split proofs for this to pass, right?

token/cli/src/main.rs
```diff
@@ -8456,6 +9003,87 @@ mod tests {
...
+        //         &destination_account.to_string(),
+        //         "--confidential",
+        //     ],
+        // )
+        // .await
+        // .unwrap();
+
+        // withdraw confidential tokens
+        //
+        // NOTE: the test fails due to transaction size limit :(
```
joncinque, line 9070: We'll just need another helper on the token client to generate the proof here, right?

joncinque (commented):
This is looking really good! I got too excited and gave an early review, so I hope that's ok :sweat_smile:

token/cli/src/main.rs
```diff
@@ -1558,6 +1576,70 @@ async fn command_transfer(
         token.with_memo(text, vec![config.default_signer()?.pubkey()]);
     }
 
+    // fetch confidential transfer info for recipient and auditor
+    let (recipient_elgamal_pubkey, auditor_elgamal_pubkey) = if let Some(args) =
+        confidential_transfer_args
+    {
+        if !config.sign_only {
+            let confidential_transfer_mint = config.get_account_checked(&token_pubkey).await?;
```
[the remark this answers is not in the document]
samkim-crypto: Yes, I think it is a good idea to note it in a comment and potentially improve it in the future.

token/cli/src/main.rs
samkim-crypto: Oops! Fixed!

token/cli/src/main.rs
samkim-crypto: Yes, that is a great idea.

token/cli/src/main.rs
samkim-crypto: Yep that would be the next step. Since this PR is quite long already, should we land this PR once the existing issues are addressed and then add the split proofs in a follow-up?

token/cli/src/main.rs
samkim-crypto: Yeah, it should suffice to build a token-client function that (1) creates a withdraw proof context, and then (2) follow through with the withdraw instruction.

token/cli/src/main.rs
```diff
@@ -3071,6 +3181,208 @@ async fn command_enable_disable_confidential_transfers(
...
+    let account = config.get_account_checked(&token_account_address).await?;
+
+    let state_with_extension = StateWithExtensionsOwned::<Account>::unpack(account.data)?;
+    let token = token_client_from_config(config, &state_with_extension.base.mint, None)?;
+
+    let res = token
+        .confidential_transfer_apply_pending_balance(
+            &token_account_address,
+            &owner,
+            None,
```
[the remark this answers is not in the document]
samkim-crypto: Yep, thanks for the catch. Updated.

token/cli/src/main.rs
samkim-crypto: I am not sure how I did not think of that. Thanks!

token/cli/src/main.rs
```diff
@@ -1571,17 +1653,45 @@ async fn command_transfer(
...
                 &sender,
                 &recipient_token_account,
                 &sender_owner,
                 transfer_balance,
-                fee,
+                maybe_fee.unwrap(),
                 &bulk_signers,
             )
             .await?
+    } else if maybe_fee.is_none() && confidential_transfer_args.is_some() {
```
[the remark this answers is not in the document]
samkim-crypto: Yes, I think this is much cleaner!

token/cli/src/main.rs
samkim-crypto: Hm yeah, I think this is a bit tricky unless we put `amount` before the mint. Not sure if it is worth breaking consistency with the rest of the commands.

token/cli/src/main.rs
samkim-crypto: Yeah, if we omit mint, then clap will interpret the amount as mint. It doesn't seem like clap provides ways to addresses this type of parsing...

token/cli/src/main.rs
joncinque: Yeah that's fine, it'll require #5663 anyway!

token/cli/src/main.rs
joncinque: Ah darn, no worries then!

joncinque (approved):
Looks great, we're so close!

PR #5640, opened 2023-10-23
transfer-hook: Move interface and example into subfolders

joncinque:
#### Problem

The transfer hook interface and example program don't follow conventions by being in two different subdirs: `transfer-hook-interface` and `transfer-hook-example`.

#### Solution

Make them into `token/transfer-hook/example` and `token/transfer-hook/interface`.

Note that I want to make a super-simple example CLI for people to deploy, and it made more sense to put that in a `transfer-hook/cli` subdir, rather than also have `transfer-hook-cli`, which is why I'm doing this now.


buffalojoec (approved):
Clean diffs!! `|-|/|` 😂

PR #5642, opened 2023-10-24
token metadata js: bump version to 0.1.0

buffalojoec:
Bumps the version of the new SPL Token Metadata JS library to `0.1.0` for release.


joncinque (dismissed):
Looks good, but I think the version is also in the lockfile -- can you also just run `npm install` to update the version there too?

joncinque (approved):

PR #5659, opened 2023-10-24
transfer-hook-cli: Add a tool for creating account metas

joncinque:
#### Problem

I want to demo transfer hooks at breakpoint, but there's no easy way to create / manage the extra account metas required for transfer hooks.

#### Solution

Create a simple CLI that can do exactly that. It even comes with a test and everything! The parsing of extra account metas was lifted directly from the token-cli, with some small modifications for clap v3. The tests were mostly lifted from the token-upgrade-cli.

### DIFF
--- a/token/transfer-hook/cli/src/main.rs
+++ b/token/transfer-hook/cli/src/main.rs
@@ -1,0 +1,405 @@
+use {
+    clap::{crate_description, crate_name, crate_version, Arg, Command},
+    solana_clap_v3_utils::{
+        input_parsers::{parse_url_or_moniker, pubkey_of_signer},
+        input_validators::{is_valid_pubkey, is_valid_signer, normalize_to_url_if_moniker},
+        keypair::DefaultSigner,
+    },
+    solana_client::nonblocking::rpc_client::RpcClient,
+    solana_remote_wallet::remote_wallet::RemoteWalletManager,
+    solana_sdk::{
+        commitment_config::CommitmentConfig,
+        instruction::AccountMeta,
+        pubkey::Pubkey,
+        signature::{Signature, Signer},
+        system_instruction,
+        transaction::Transaction,
+    },
+    spl_tlv_account_resolution::state::ExtraAccountMetaList,
+    spl_transfer_hook_interface::{
+        get_extra_account_metas_address, instruction::initialize_extra_account_meta_list,
+    },
+    std::{fmt, process::exit, rc::Rc, str::FromStr},
+    strum_macros::{EnumString, IntoStaticStr},
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, EnumString, IntoStaticStr)]
+#[strum(serialize_all = "kebab-case")]
+pub enum AccountMetaRole {
+    Readonly,
+    Writable,
+    ReadonlySigner,
+    WritableSigner,
+}
+impl fmt::Display for AccountMetaRole {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+fn parse_transfer_hook_account(arg: &str) -> Result<AccountMeta, String> {
+    match arg.split(':').collect::<Vec<_>>().as_slice() {
+        [address, role] => {
+            let address = Pubkey::from_str(address).map_err(|e| format!("{e}"))?;
+            let meta = match AccountMetaRole::from_str(role).map_err(|e| format!("{e}"))? {
+                AccountMetaRole::Readonly => AccountMeta::new_readonly(address, false),
+                AccountMetaRole::Writable => AccountMeta::new(address, false),
+                AccountMetaRole::ReadonlySigner => AccountMeta::new_readonly(address, true),
+                AccountMetaRole::WritableSigner => AccountMeta::new(address, true),
+            };
+            Ok(meta)
+        }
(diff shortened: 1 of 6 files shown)


joncinque:
Oh also, reserved the name: https://crates.io/crates/spl-transfer-hook-cli

token/transfer-hook/cli/src/main.rs
```diff
@@ -0,0 +1,405 @@
...
+                .short('u')
+                .long("url")
+                .value_name("URL")
+                .takes_value(true)
+                .global(true)
+                .value_parser(parse_url_or_moniker)
+                .help("JSON RPC URL for the cluster [default: value from configuration file]"),
+        )
+        .subcommand(
+            Command::new("create-extra-account-metas")
```
buffalojoec, line 177: That's a big ol' command (`spl-transfer-hook create-extra-account-metas ...`)!

token/transfer-hook/cli/src/main.rs
```diff
@@ -0,0 +1,405 @@
...
+
+    if config.verbose {
+        println!("JSON RPC URL: {}", config.json_rpc_url);
+    }
+    let rpc_client =
+        RpcClient::new_with_commitment(config.json_rpc_url.clone(), config.commitment_config);
+
+    match (command, matches) {
+        ("create-extra-account-metas", arg_matches) => {
+            let program_id = pubkey_of_signer(arg_matches, "program_id", &mut wallet_manager)
```
buffalojoec, line 266: Shouldn't this be `try_pubkey_of`?

token/transfer-hook/cli/src/main.rs
```diff
@@ -0,0 +1,405 @@
...
+                .unwrap()
+                .unwrap();
+            let token = pubkey_of_signer(arg_matches, "token", &mut wallet_manager)
```
buffalojoec, line 269: Also curious why this one involves signer/keypair. Maybe I'm not understanding how these methods work, but they appear to attempt to find the keypair at some path.

buffalojoec (commented):
Lgtm. wen PDAs? 👀

token/transfer-hook/cli/src/main.rs
joncinque: And to think, I shortened it from `create-extra-account-metas-list`. I couldn't come up with anything better, but if you have a suggestion, please let me know!

token/transfer-hook/cli/src/main.rs
joncinque: It's more of a convenience, allowing me to specify `mint.json` on the command line instead of only its address

token/transfer-hook/cli/src/main.rs
buffalojoec: I think it's fine! Best I could come up with is `transfer-hook init-extra-metas`.

buffalojoec (dismissed):
Cool CLI! Curious if you plan to roll an `update` as well, in case I want to modify my list?

token/transfer-hook/cli/src/main.rs
```diff
@@ -0,0 +1,405 @@
...
+    let account_size = ExtraAccountMetaList::size_of(length)?;
+    let required_lamports = rpc_client
+        .get_minimum_balance_for_rent_exemption(account_size)
+        .await
+        .map_err(|err| format!("error: unable to fetch rent-exemption: {err}"))?;
+    let current_lamports = rpc_client
+        .get_account(&extra_account_metas_address)
+        .await
+        .map(|a| a.lamports)
+        .unwrap_or(0);
```
buffalojoec, line 90: Should we maybe have a check in case the account exists?

buffalojoec: I know the `init_extension` does this, but just curious if we want it on the CLI, too.

token/transfer-hook/cli/src/main.rs
joncinque: How about we split the difference with `create-extra-metas`? `init` typically means that you're *just* writing to an allocated + assigned account, and here we're also doing the allocation + assign

token/transfer-hook/cli/src/main.rs
joncinque: Yep that's the idea! I'm pretty lazy when I write shell scripts and just want to use keypair files haha

token/transfer-hook/cli/src/main.rs
joncinque: Ah good call, yeah I'll add the check

joncinque:
> Lgtm. wen PDAs? 👀

Oh dang, that's a good point. I'll add an issue for it

> Cool CLI! Curious if you plan to roll an update as well, in case I want to modify my list?

That would also require a new instruction on the example program, which isn't technically required by the interface, so I would probably avoid adding it.

token/transfer-hook/cli/src/main.rs
buffalojoec: Leaving that totally up to you, but that sounds good to me!

buffalojoec:
> That would also require a new instruction on the example program, which isn't technically required by the interface, so I would probably avoid adding it.

Makes sense. I guess it kind of makes me wonder if we _should_ support this in the interface/example eventually? 😅

joncinque:
> Makes sense. I guess it kind of makes me wonder if we should support this in the interface/example eventually? 😅

I'll write up a good-first-issue for it :wink:

buffalojoec (approved):
meta

PR #5663, opened 2023-10-25
token-client: Use `Signers` instead of `Signer` for confidential transfer functions

joncinque:
#### Problem

A lot of the confidential transfer functions in the token client parametrize on `Signer` and accept multiple keypairs for the required signatures (account authority, context state accounts, etc). This does not work with the token CLI, which uses a huge `bulk_signers` everywhere.

#### Solution

Change all of the functions to take in `Signers` instead! This works very nicely for the most part, but the split-proofs-in-parallel functions get a little tricky since you need to craft the correct `Signers` arrays for each transaction.

Let me know how it looks! You should be able to track each commit pretty easily, since each one actually works on its own.

### DIFF
--- a/token/program-2022-test/tests/confidential_transfer.rs
+++ b/token/program-2022-test/tests/confidential_transfer.rs
@@ -2791,10 +2791,10 @@ async fn confidential_transfer_transfer_with_fee_and_split_proof_context() {
             transfer_context_state_accounts,
             &equality_proof_data,
             &transfer_amount_ciphertext_validity_proof_data,
-            None,
-            &equality_proof_context_state_account,
-            &transfer_amount_ciphertext_validity_proof_context_state_account,
-            None,
+            &[
+                &equality_proof_context_state_account,
+                &transfer_amount_ciphertext_validity_proof_context_state_account,
+            ],
         )
         .await
         .unwrap();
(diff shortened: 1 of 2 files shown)


token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -3006,13 +3016,9 @@ async fn confidential_transfer_transfer_with_fee_and_split_proof_context_in_para
...
             TEST_FEE_BASIS_POINTS,
             TEST_MAXIMUM_FEE,
-            &alice,
-            &equality_proof_context_state_account,
-            &transfer_amount_ciphertext_validity_proof_context_state_account,
-            &fee_sigma_proof_context_state_account,
-            &fee_ciphertext_validity_proof_context_state_account,
-            &range_proof_context_state_account,
-            None,
+            &equality_and_ciphertext_signers,
```
samkim-crypto, line 3019:
```suggestion
            &equality_and_ciphertext_proof_signers,
```

token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -2991,6 +2990,17 @@ async fn confidential_transfer_transfer_with_fee_and_split_proof_context_in_para
         close_split_context_state_accounts: None,
     };
 
+    let equality_and_ciphertext_signers = vec![
```
samkim-crypto, line 2993: Very minor: since we refer to the second type of proof as `ciphertext_validity` proof throughout the program, let's use that for consistency.
```suggestion
    let equality_and_ciphertext_proof_signers = vec![
```

samkim-crypto (approved):
Looks good to me!

PR #5667, opened 2023-10-26
Adding Token Metadata Support to @solana/spl-token 

mistersimon:
Addresses #5229 

> With the addition of token metadata to Token2022 and the SPL Token Metadata interface itself to the SPL library, we need to support this cool feature & interface in JavaScript!
> 
> The goal here is to lay down a JavaScript library for SPL Token Metadata and then import it into @solana/spl-token for Token2022 metadata support.
> 
> This issue is importing and implementing @solana/spl-token-metadata.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -722,6 +722,9 @@ importers:
       '@solana/buffer-layout-utils':
         specifier: ^0.2.0
         version: 0.2.0
+      '@solana/spl-token-metadata':
+        specifier: ^0.1.1
+        version: link:../../token-metadata/js
       buffer:
         specifier: ^6.0.3
         version: 6.0.3
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
 importers:
 
   .:
@@ -723,12 +727,15 @@ importers:
         specifier: ^0.2.0
         version: 0.2.0
       '@solana/spl-token-metadata':
-        specifier: ^0.1.1
+        specifier: ^0.1.2
         version: link:../../token-metadata/js
       buffer:
         specifier: ^6.0.3
         version: 6.0.3
     devDependencies:
+      '@solana/codecs-strings':
+        specifier: 2.0.0-experimental.9741939
+        version: 2.0.0-experimental.9741939(fastestsmallesttextencoderdecoder@1.0.22)
       '@solana/spl-memo':
         specifier: 0.2.3
         version: link:../../memo/js
@@ -2274,7 +2281,6 @@ packages:
 
   /@solana/codecs-core@2.0.0-experimental.9741939:
     resolution: {integrity: sha512-7E51aEwLW+1ta6VWbEq0CbvwSUOcIwmaQCLpkOKsF6cwSNqE5GDv/jw9zKuiYOynlBFQO1Ws59a/hlb2wwwWKg==}
-    dev: false
 
   /@solana/codecs-data-structures@2.0.0-experimental.9741939:
     resolution: {integrity: sha512-Ghjx0pFEA22T/cpqg3zXHYlnxqKytjTr1QjgfhWvBrlVFkqr3ZQL4av9/V6sHKvpoi+JNkEwM1wE+tlpt/54CA==}
@@ -2287,7 +2293,6 @@ packages:
     resolution: {integrity: sha512-VXBvw8LZdGUJGuC33EFzvaUxCvgNtr9y9Td8zjK9wJDqKSzR0+G53CJv5K4AF25LUu8du8XHBK3VEz3YHl44nQ==}
     dependencies:
       '@solana/codecs-core': 2.0.0-experimental.9741939
-    dev: false
 
   /@solana/codecs-strings@2.0.0-experimental.9741939(fastestsmallesttextencoderdecoder@1.0.22):
     resolution: {integrity: sha512-J1DCTJsAMhFcIIvys/yNWM2Vr4HEWi+LzFB9xxdH+FHTynApjTMd4vI7frgnckuWH7ynp8EZFPnBTje1BphQrw==}
@@ -2297,7 +2302,6 @@ packages:
       '@solana/codecs-core': 2.0.0-experimental.9741939
       '@solana/codecs-numbers': 2.0.0-experimental.9741939
       fastestsmallesttextencoderdecoder: 1.0.22
-    dev: false
 
   /@solana/eslint-config-solana@1.0.2(@typescript-eslint/eslint-plugin@6.14.0)(@typescript-eslint/parser@6.14.0)(eslint-plugin-jest@27.6.0)(eslint-plugin-react-hooks@4.6.0)(eslint-plugin-simple-import-sort@10.0.0)(eslint-plugin-sort-keys-fix@1.1.2)(eslint@8.55.0)(typescript@5.3.3):
     resolution: {integrity: sha512-/mBi2xFsX0fq3ukZ3ugA+Zh2+Ct+bh713ZSdOJAv1G7ARzyOEDxmiUz8TLzHu4joOC/3Ok4Yl7cjupOcIRzMHw==}
@@ -4585,7 +4589,6 @@ packages:
 
   /fastestsmallesttextencoderdecoder@1.0.22:
     resolution: {integrity: sha512-Pb8d48e+oIuY4MaM64Cd7OW1gt4nxCHs7/ddPPZ/Ic3sg8yVGM7O9wDvZ7us6ScaUupzM+pfBolwtYhN1IxBIw==}
-    dev: false
 
   /fastq@1.15.0:
     resolution: {integrity: sha512-wBrocU2LCXXa+lWBt8RoIRD89Fi8OdABODa/kEnyeyjS5aZO5/GNvI5sEINADqP/h8M29UHTHUb53sUu5Ihqdw==}
@@ -8014,7 +8017,3 @@ packages:
     resolution: {integrity: sha512-rVksvsnNCdJ/ohGc6xgPwyN8eheCxsiLM8mxuE/t/mOVqJewPuO1miLpTHQiRgTKCLexL4MeAFVagts7HmNZ2Q==}
     engines: {node: '>=10'}
     dev: true
-
-settings:
-  autoInstallPeers: true
-  excludeLinksFromLockfile: false
--- a/token/js/src/extensions/extensionType.ts
+++ b/token/js/src/extensions/extensionType.ts
@@ -45,6 +45,10 @@ export enum ExtensionType {
 export const TYPE_SIZE = 2;
 export const LENGTH_SIZE = 2;
 
+function addTypeAndLengthToLen(len: number): number {
+    return len + TYPE_SIZE + LENGTH_SIZE;
+}
+
 // NOTE: All of these should eventually use their type's Span instead of these
 // constants.  This is provided for at least creation to work.
 export function getTypeLen(e: ExtensionType): number {
@@ -181,7 +185,7 @@ function getLen(extensionTypes: ExtensionType[], baseSize: number): number {
             ACCOUNT_TYPE_SIZE +
             extensionTypes
                 .filter((element, i) => i === extensionTypes.indexOf(element))
-                .map((element) => getTypeLen(element) + TYPE_SIZE + LENGTH_SIZE)
+                .map((element) => addTypeAndLengthToLen(getTypeLen(element)))
                 .reduce((a, b) => a + b);
         if (accountLength === MULTISIG_SIZE) {
             return accountLength + TYPE_SIZE;
@@ -201,10 +205,10 @@ export function getAccountLen(extensionTypes: ExtensionType[]): number {
 
 export function getExtensionData(extension: ExtensionType, tlvData: Buffer): Buffer | null {
     let extensionTypeIndex = 0;
-    while (extensionTypeIndex + TYPE_SIZE + LENGTH_SIZE <= tlvData.length) {
+    while (addTypeAndLengthToLen(extensionTypeIndex) <= tlvData.length) {
         const entryType = tlvData.readUInt16LE(extensionTypeIndex);
         const entryLength = tlvData.readUInt16LE(extensionTypeIndex + TYPE_SIZE);
-        const typeIndex = extensionTypeIndex + TYPE_SIZE + LENGTH_SIZE;
+        const typeIndex = addTypeAndLengthToLen(extensionTypeIndex);
         if (entryType == extension) {
             return tlvData.slice(typeIndex, typeIndex + entryLength);
         }
@@ -220,7 +224,7 @@ export function getExtensionTypes(tlvData: Buffer): ExtensionType[] {
         const entryType = tlvData.readUInt16LE(extensionTypeIndex);
         extensionTypes.push(entryType);
         const entryLength = tlvData.readUInt16LE(extensionTypeIndex + TYPE_SIZE);
-        extensionTypeIndex += TYPE_SIZE + LENGTH_SIZE + entryLength;
+        extensionTypeIndex += addTypeAndLengthToLen(entryLength);
     }
     return extensionTypes;
 }
@@ -241,8 +245,8 @@ export function getNewAccountLenForExtensionLen(
     const mint = unpackMint(address, info, programId);
     const extensionData = getExtensionData(e, mint.tlvData);
 
-    // 2 bytes type, 2 bytes length, extensionLen
-    const newExtensionLen = TYPE_SIZE + LENGTH_SIZE + extensionLen;
+    const currentExtensionLen = extensionData ? addTypeAndLengthToLen(extensionData.length) : 0;
+    const newExtensionLen = addTypeAndLengthToLen(extensionLen);
 
-    return info.data.length + newExtensionLen - (extensionData?.length || 0);
+    return info.data.length + newExtensionLen - currentExtensionLen;
 }
--- a/token/js/src/extensions/extensionType.ts
+++ b/token/js/src/extensions/extensionType.ts
@@ -36,6 +36,7 @@ export enum ExtensionType {
     // ConfidentialTransferFee, // Not implemented yet
     // ConfidentialTransferFeeAmount, // Not implemented yet
     MetadataPointer = 18, // Remove number once above extensions implemented
+    TokenMetadata = 19, // Remove number once above extensions implemented
 }
 
 export const TYPE_SIZE = 2;
@@ -46,6 +47,7 @@ export const LENGTH_SIZE = 2;
 export function getTypeLen(e: ExtensionType): number {
     switch (e) {
         case ExtensionType.Uninitialized:
+        case ExtensionType.TokenMetadata:
             return 0;
         case ExtensionType.TransferFeeConfig:
             return TRANSFER_FEE_CONFIG_SIZE;
@@ -95,6 +97,7 @@ export function isMintExtension(e: ExtensionType): boolean {
         case ExtensionType.PermanentDelegate:
         case ExtensionType.TransferHook:
         case ExtensionType.MetadataPointer:
+        case ExtensionType.TokenMetadata:
             return true;
         case ExtensionType.Uninitialized:
         case ExtensionType.TransferFeeAmount:
@@ -130,6 +133,7 @@ export function isAccountExtension(e: ExtensionType): boolean {
         case ExtensionType.PermanentDelegate:
         case ExtensionType.TransferHook:
         case ExtensionType.MetadataPointer:
+        case ExtensionType.TokenMetadata:
             return false;
         default:
             throw Error(`Unknown extension type: ${e}`);
@@ -154,6 +158,7 @@ export function getAccountTypeOfMintType(e: ExtensionType): ExtensionType {
         case ExtensionType.MemoTransfer:
         case ExtensionType.MintCloseAuthority:
         case ExtensionType.MetadataPointer:
+        case ExtensionType.TokenMetadata:
         case ExtensionType.Uninitialized:
         case ExtensionType.InterestBearingConfig:
         case ExtensionType.PermanentDelegate:
--- a/token/js/src/extensions/extensionType.ts
+++ b/token/js/src/extensions/extensionType.ts
@@ -238,12 +238,12 @@ export function getAccountLenForMint(mint: Mint): number {
 export function getNewAccountLenForExtensionLen(
     info: AccountInfo<Buffer>,
     address: PublicKey,
-    e: ExtensionType,
+    extensionType: ExtensionType,
     extensionLen: number,
     programId = TOKEN_2022_PROGRAM_ID
 ): number {
     const mint = unpackMint(address, info, programId);
-    const extensionData = getExtensionData(e, mint.tlvData);
+    const extensionData = getExtensionData(extensionType, mint.tlvData);
 
     const currentExtensionLen = extensionData ? addTypeAndLengthToLen(extensionData.length) : 0;
     const newExtensionLen = addTypeAndLengthToLen(extensionLen);
--- a/token/js/src/extensions/extensionType.ts
+++ b/token/js/src/extensions/extensionType.ts
@@ -47,7 +47,6 @@ export const LENGTH_SIZE = 2;
 export function getTypeLen(e: ExtensionType): number {
     switch (e) {
         case ExtensionType.Uninitialized:
-        case ExtensionType.TokenMetadata:
             return 0;
         case ExtensionType.TransferFeeConfig:
             return TRANSFER_FEE_CONFIG_SIZE;
@@ -81,6 +80,8 @@ export function getTypeLen(e: ExtensionType): number {
             return TRANSFER_HOOK_SIZE;
         case ExtensionType.TransferHookAccount:
             return TRANSFER_HOOK_ACCOUNT_SIZE;
+        case ExtensionType.TokenMetadata:
+            throw Error(`Variable length extension type: ${e}`);
         default:
             throw Error(`Unknown extension type: ${e}`);
     }
--- a/token/js/src/extensions/extensionType.ts
+++ b/token/js/src/extensions/extensionType.ts
@@ -214,6 +214,17 @@ export function getExtensionData(extension: ExtensionType, tlvData: Buffer): Buf
     return null;
 }
 
+export function getExtensionDataFromAccountInfo(
+    address: PublicKey,
+    accountInfo: AccountInfo<Buffer> | null,
+    e: ExtensionType,
+    programId = TOKEN_2022_PROGRAM_ID
+): Buffer | null {
+    const mint = unpackMint(address, accountInfo, programId);
+
+    return getExtensionData(e, mint.tlvData);
+}
+
 export function getExtensionTypes(tlvData: Buffer): ExtensionType[] {
     const extensionTypes = [];
     let extensionTypeIndex = 0;
@@ -232,50 +243,31 @@ export function getAccountLenForMint(mint: Mint): number {
     return getAccountLen(accountExtensions);
 }
 
-// Get new extension length, with current data as input
-export type GetUpdatedVariableExtensionLen = (current: Buffer | null) => number;
-
 export function getNewAccountLenForExtensionLen(
     info: AccountInfo<Buffer>,
     address: PublicKey,
     e: ExtensionType,
-    extensionLen: number | GetUpdatedVariableExtensionLen,
+    extensionLen: number,
     programId = TOKEN_2022_PROGRAM_ID
 ): number {
-    const mint = unpackMint(address, info, programId);
+    const data = getExtensionDataFromAccountInfo(address, info, e, programId);
 
-    const data = getExtensionData(e, mint.tlvData);
-
-    // Need 2 bytes extension discriminator, 2 bytes length
-    let newExtensionLen = 2 + 2;
-
-    newExtensionLen += typeof extensionLen === 'function' ? extensionLen(data) : extensionLen;
+    // 2 bytes discriminator, 2 bytes length, extensionLen
+    const newExtensionLen = 2 + 2 + extensionLen;
 
     return info.data.length + newExtensionLen - (data?.length || 0);
-}
-
-export async function getAdditionalRentForNewAccountLen(
-    connection: Connection,
-    info: AccountInfo<Buffer>,
-    newAccountLen: number
-) {
-    if (newAccountLen <= info.data.length) {
-        return 0;
-    }
-
-    const newRentExemptMinimum = await connection.getMinimumBalanceForRentExemption(newAccountLen);
-
-    return newRentExemptMinimum - info.lamports;
 }
 
 export async function getAdditionalRentForNewExtensionLen(
     connection: Connection,
     address: PublicKey,
     e: ExtensionType,
-    extensionLen: number | GetUpdatedVariableExtensionLen,
-    programId = TOKEN_2022_PROGRAM_ID
+    extensionLen: number,
+    programId = TOKEN_2022_PROGRAM_ID,
+    accountInfo?: AccountInfo<Buffer> | null
 ): Promise<number> {
-    const info = await connection.getAccountInfo(address);
+    // If account info was provided, don't fetch again
+    const info = accountInfo ?? (await connection.getAccountInfo(address));
 
     if (!info) {
         throw new TokenAccountNotFoundError();
@@ -283,5 +275,11 @@ export async function getAdditionalRentForNewExtensionLen(
 
     const newAccountLen = getNewAccountLenForExtensionLen(info, address, e, extensionLen, programId);
 
-    return await getAdditionalRentForNewAccountLen(connection, info, newAccountLen);
+    if (newAccountLen <= info.data.length) {
+        return 0;
+    }
+
+    const newRentExemptMinimum = await connection.getMinimumBalanceForRentExemption(newAccountLen);
+
+    return newRentExemptMinimum - info.lamports;
 }
--- a/token/js/src/extensions/extensionType.ts
+++ b/token/js/src/extensions/extensionType.ts
@@ -1,4 +1,4 @@
-import type { AccountInfo, Connection, PublicKey } from '@solana/web3.js';
+import type { AccountInfo, PublicKey } from '@solana/web3.js';
 
 import { ACCOUNT_SIZE } from '../state/account.js';
 import type { Mint } from '../state/mint.js';
@@ -17,7 +17,6 @@ import { PERMANENT_DELEGATE_SIZE } from './permanentDelegate.js';
 import { TRANSFER_FEE_AMOUNT_SIZE, TRANSFER_FEE_CONFIG_SIZE } from './transferFee/index.js';
 import { TRANSFER_HOOK_ACCOUNT_SIZE, TRANSFER_HOOK_SIZE } from './transferHook/index.js';
 import { TOKEN_2022_PROGRAM_ID } from '../constants.js';
-import { TokenAccountNotFoundError } from '../errors.js';
 
 // Sequence from https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/src/extension/mod.rs#L903
 export enum ExtensionType {
@@ -214,17 +213,6 @@ export function getExtensionData(extension: ExtensionType, tlvData: Buffer): Buf
     return null;
 }
 
-export function getExtensionDataFromAccountInfo(
-    address: PublicKey,
-    accountInfo: AccountInfo<Buffer> | null,
-    e: ExtensionType,
-    programId = TOKEN_2022_PROGRAM_ID
-): Buffer | null {
-    const mint = unpackMint(address, accountInfo, programId);
-
-    return getExtensionData(e, mint.tlvData);
-}
-
 export function getExtensionTypes(tlvData: Buffer): ExtensionType[] {
     const extensionTypes = [];
     let extensionTypeIndex = 0;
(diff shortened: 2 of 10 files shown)


token/js/test/e2e-2022/tokenmeta.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    createInitializeAccountInstruction,
+    getAccountLen,
+    ExtensionType,
+    initializeMetadata,
+} from '../../src';
+import { TEST_PROGRAM_ID, newAccountWithLamports, getConnection } from '../common';
+import { MINT_SIZE } from '../../src/state/mint';
+
+const TEST_TOKEN_DECIMALS = 2;
+const EXTENSIONS = [ExtensionType.CpiGuard];
```
mistersimon, line 18: How do we set the account length for token metadata? From what i understand its not fixed length
I tried reading in the rust to see how it worked, but couldn't figure it out

token/js/test/e2e-2022/tokenmeta.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+describe.only('meta', () => {
```
mistersimon, line 20: Note to self: Don't forget to remove this `.only`

token/js/test/e2e-2022/tokenmeta.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        await sendAndConfirmTransaction(connection, transaction, [payer, accountKeypair], undefined);
+    });
+
+    it('Can initialize token metadata', async () => {
+        const updateAuthority = Keypair.generate();
+
+        await initializeMetadata(
+            connection,
+            payer,
+            mintKeypair.publicKey, // metadata
```
mistersimon, line 74: I can't figure out if we need to create a new 'metadata' account, or if the mint account is used

token/js/test/e2e-2022/tokenmeta.test.ts
```diff
@@ -0,0 +1,86 @@
...
+                lamports,
+                programId: TEST_PROGRAM_ID,
+            }),
+            createInitializeAccountInstruction(account, mint, owner.publicKey, TEST_PROGRAM_ID)
+        );
+
+        await sendAndConfirmTransaction(connection, transaction, [payer, accountKeypair], undefined);
+    });
+
+    it('Can initialize token metadata', async () => {
```
mistersimon, line 68: Currently throwing this error:
```Error: failed to send transaction : Transaction simulation failed: Error processing Instruction 0: custom prog ram error: 0x33```

Think it maps to `Calculated fee does not match expected fee`
https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/src/error.rs#L126C14-L126C56

**Review by mistersimon — COMMENTED** [review]
Hey @buffalojoec, can't get the test case setup correct with the token metadata.

Thought I'd ask if you got any guidance

File: token/js/test/e2e-2022/tokenmeta.test.ts (hunk 3 above)
Comment (buffalojoec) [reply token/js/test/e2e-2022/tokenmeta.test.ts:74]: Technically there's two extensions that have to do with metadata.

- `MetadataPointer`: This extension lives inside the mint and tells Token2022 which account has this token's metadata. It's just a pointer to some address protected by an authority.
- `TokenMetadata`: This extension is the token metadata itself. If a mint has the `TokenMetadata` extension, it has metadata within the mint and not externally.

Token 2022's processor for the `TokenMetadata` extension does not deal with creating metadata outside of a mint.

**Review by buffalojoec — COMMENTED** [review]
What exactly is the issue with the test that you're having?

File: token/js/test/e2e-2022/tokenmeta.test.ts (hunk 3 above)
Comment (mistersimon) [reply token/js/test/e2e-2022/tokenmeta.test.ts:74]: I don't think these 2 extensions have been added to the js code - shall i add them as part of the PR?

File: token/js/test/e2e-2022/tokenmeta.test.ts (hunk 4 above)
Comment (mistersimon) [reply token/js/test/e2e-2022/tokenmeta.test.ts:68]: Yeah, i'm quite sure i'm not setting up the accounts correctly - just can't figure out the right way

File: token/js/test/e2e-2022/tokenmeta.test.ts (hunk 3 above)
Comment (buffalojoec) [reply token/js/test/e2e-2022/tokenmeta.test.ts:74]: Yes that's correct.

Comment (mistersimon) [reply token/js/test/e2e-2022/tokenmeta.test.ts:73]: Haven't managed to make much progress, still can't create the token with the metadata pointer inside.

How does the extension get "enabled"? Do I need to modify [createMint]( https://github.com/mistersimon/solana-program-library/blob/token-2022/js-incorp-metadata/token/js/src/actions/createMint.ts#L21) or is a separate transaction like [initializeMintCloseAuthority](https://github.com/mistersimon/solana-program-library/blob/token-2022/js-incorp-metadata/token/js/src/instructions/initializeMintCloseAuthority.ts#L38)?

Comment (buffalojoec) [reply token/js/test/e2e-2022/tokenmeta.test.ts:73]: So to answer your question, we basically need an instruction to create a `MetadataPointer` extension.

https://github.com/solana-labs/solana-program-library/blob/022e0c5d49d964ae59be56cc5b1532c4e6f70726/token/program-2022/src/extension/metadata_pointer/mod.rs#L20-L24

File: token/js/test/e2e-2022/tokenmeta.test.ts (hunk 1 above)
Comment (buffalojoec) [reply token/js/test/e2e-2022/tokenmeta.test.ts]: Yes, the metadata is comprised of multiple strings, which are variable-length. So you have to calculate the size of the metadata object using `@solana/spl-type-length-value`

File: token/js/test/e2e-2022/tokenmeta.test.ts (hunk 3 above)
Comment (mistersimon) [reply token/js/test/e2e-2022/tokenmeta.test.ts:73]: Hey - thanks for your guidance as always! I managed to make some progress. I've got the MetadataPointer instructions implemented now (verified the data packing against rust implementation)

Comment (mistersimon) [reply token/js/test/e2e-2022/tokenmeta.test.ts:73]: Will look next to address this `create an account with enough space for the extensions`

File: token/js/src/extensions/tokenMetadata/actions.ts [hunk 5]
```diff
@@ -0,0 +1,133 @@
...
+    multiSigners: Signer[] = [],
+    confirmOptions?: ConfirmOptions,
+    programId = TOKEN_2022_PROGRAM_ID
+): Promise<TransactionSignature> {
+    const [mintAuthorityPublicKey, signers] = getSigners(mintAuthority, multiSigners);
+
+    const transaction = new Transaction().add(
+        createInitializeInstruction({
+            programId,
+            metadata: mint,
```
mistersimon, line 48: Is it correct that we don't give users the option to have a separate metadata account? Understand for token-2022, the metadata can only live inside the mint, so we abstract that away from users

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ * @param connection       Connection to use
+ * @param payer            Payer of the transaction fees
+ * @param mint             Mint Account
+ * @param multiSigners     Signing accounts if `authority` is a multisig
+ * @param confirmOptions   Options for confirming the transaction
+ * @param programId        SPL Token program account
+ *
+ * @return Signature of the confirmed transaction
+ */
+export async function emitTokenMetadata(
```
mistersimon, line 118: Is this suppose to just return the signature, or should we actually return the metadata?

token/js/test/e2e-2022/tokenMetadata.test.ts
```diff
@@ -0,0 +1,181 @@
...
+const TEST_TOKEN_DECIMALS = 2;
+
+// TODO: Remove or move to src. Temp function so we can repack to use spl-token-metadata
+const lengthBuffer = (buffer: Buffer | Uint8Array): Buffer => {
+    const length = Buffer.alloc(4);
+    length.writeUIntLE(buffer.length, 0, 4);
+    return length;
+};
+
+describe('Token Metadata', () => {
```
mistersimon, line 35: I had difficulty with the unpacking of the metadata. I got it to work, but don't understand why and thinking perhaps i got lucky?

I assumed its needs to be done, but please correct me if i'm wrong.

token/js/test/e2e-2022/tokenMetadata.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    before(async () => {
+        connection = await getConnection();
+        payer = await newAccountWithLamports(connection, 1000000000);
+    });
+
+    beforeEach(async () => {
+        mint = Keypair.generate();
+
+        // TODO @mistersimon:- Change to instruction creator once merged
+        const metadataPointerInstruction = new TransactionInstruction({
```
mistersimon, line 61: Will switch to the functions created in https://github.com/solana-labs/solana-program-library/pull/5805, just wanted to decouple them for now and keep them smaller

token/js/test/e2e-2022/tokenMetadata.test.ts
```diff
@@ -0,0 +1,181 @@
...
+            'name',
+            'symbol',
+            'uri',
+            undefined,
+            undefined,
+            TEST_PROGRAM_ID
+        );
+    });
+
+    const getTokenMetadata = async () => {
```
mistersimon, line 104: 1. Do we need to a function like this in the library? Couldn't see anything similar - Would it be considered an action? 

1. I don't understand why ExtensionType needs to be 19 (which is GroupPointer) - Is it correct, or just a coincidence it works?

1. Seems like we can't use the `unpack` in the `@solana/spl-token-metadata`  library, because its not a tlv. Should we refactor `@solana/spl-token-metadata` to expose a function to decode the pure data? Or create the same functionality in this library?

token/js/test/e2e-2022/tokenMetadata.test.ts
```diff
@@ -0,0 +1,181 @@
...
+            symbol: 'symbol',
+            uri: 'uri',
+            additionalMetadata: [],
+        });
+    });
+
+    it('can successfully emit', async () => {
+        const signature = await emitTokenMetadata(connection, payer, mint.publicKey);
+        const tx: any = await connection.getTransaction(signature);
+        const data = Buffer.from(tx?.meta?.returnData?.data?.[0], 'base64');
```
mistersimon, line 132: Is the correct approach? The typescript doesn't have `returnData` defined on `tx.meta`, but its outputted when i run the test.

mistersimon (commented):
Hey @buffalojoec, cleaned up the PR and added the instructions to init, update and emit the token metadata

Wanted to check on the decoding of the Metadata before I continue to add the rest. It works, but I don't understand why and not sure if its correct. Please take a look at let me know

token/js/test/e2e-2022/tokenMetadata.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        });
+
+        const lamports = await connection.getMinimumBalanceForRentExemption(mintLen);
+
+        const transaction = new Transaction().add(
+            SystemProgram.createAccount({
+                fromPubkey: payer.publicKey,
+                newAccountPubkey: mint.publicKey,
+                space: mintLen,
+                lamports: lamports * 10,
```
mistersimon, line 74: Does the instructions need to handle topping up the mint account with lamports for rent? I saw some functionality in the rust code for this. 

Is there any examples in the repo you could share where this is done?

token/js/src/extensions/tokenMetadata/actions.ts
buffalojoec: No, you can have metadata in another account, but our CLI isn't going to do it for you.

In terms of metadata, this CLI is only handling Token2022 metadata (in mints).

token/js/src/extensions/tokenMetadata/actions.ts
buffalojoec: it's an instruction being sent to the program to emit the Token Metadata as the program's return data. Think of it like a view function.

For the return type, it should be treated like any other instruction/transaction function.

token/js/src/extensions/tokenMetadata/actions.ts
mistersimon: Okay - so the approach of not allowing users to pass in a separate metadata account in this action is correct?

token/js/test/e2e-2022/tokenMetadata.test.ts
buffalojoec: > 1. Do we need to a function like this in the library? Couldn't see anything similar - Would it be considered an action?
Yeah we probably do, even though I don't believe the Token Rust client has this yet!

> 2. I don't understand why ExtensionType needs to be 19 (which is GroupPointer) - Is it correct, or just a coincidence it works?
I count `TokenMetadata` as 19.

> 3. Seems like we can't use the `unpack` in the `@solana/spl-token-metadata`  library, because its not a tlv. Should we refactor `@solana/spl-token-metadata` to expose a function to decode the pure data? Or create the same functionality in this library?
Ahh okay, I think I see the problem here. Token2022 isn't using 8-byte discriminators, so we can't fix the `unpack` function in `@solana/spl-token-metadata` to use 8-byte discriminators. We'll have to modify that function so it can be used here, or refactor the serialization bits of the `TokenMetadata` state in the JS lib to be more friendly with the Rust trait `VariableLenPack`:

https://github.com/solana-labs/solana-program-library/blob/f542e00a8905eaf3f55dea1c08ca4787138e1788/token-metadata/interface/src/state.rs#L99-L109

token/js/test/e2e-2022/tokenMetadata.test.ts
buffalojoec: There's two versions of the command: one without the rent funding and one with. I'd just roll both like we do in Rust.

https://github.com/solana-labs/solana-program-library/blob/f542e00a8905eaf3f55dea1c08ca4787138e1788/token/client/src/token.rs#L3453
https://github.com/solana-labs/solana-program-library/blob/f542e00a8905eaf3f55dea1c08ca4787138e1788/token/client/src/token.rs#L3497

token/js/src/extensions/tokenMetadata/actions.ts
buffalojoec: Yes!

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ * @param name             Longer name of token
+ * @param symbol           Shortened symbol of token
+ * @param uri              URI pointing to more metadata (image, video, etc)
+ * @param multiSigners     Signing accounts if `authority` is a multisig
+ * @param confirmOptions   Options for confirming the transaction
+ * @param programId        SPL Token program account
+ *
+ * @return Signature of the confirmed transaction
+ */
+export async function initializeTokenMetadata(
```
buffalojoec, line 30: nit: I would rather you name these according to the same pattern we have in the Rust client, ie:

- `tokenMetadataInitialize(..)`
- `tokenMetadataUpdateField(..)`
- ...

buffalojoec (commented):
This is moving along nicely! Let's iron out the `MetadataPointer` PR first and then rebase this one on top, then I can give this another look.

Thanks for the continued effort!

mistersimon:
> This is moving along nicely! Let's iron out the `MetadataPointer` PR first and then rebase this one on top, then I can give this another look.
> 
> Thanks for the continued effort!

Sounds like a good idea. Once the `MetadataPointer` PR is merged, I'll come back to this - rebase and address your comments. Then if you can take a look then before i go add the other instructions, would be really helpful.

pnpm-lock.yaml
```diff
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
```
mistersimon, line 3: Seems to have automatically moved from bottom to top - might different pnpm version

token/js/test/e2e-2022/tokenMetadata.test.ts
```diff
@@ -0,0 +1,444 @@
...
+                null,
+                TEST_PROGRAM_ID
+            )
+        );
+
+        await sendAndConfirmTransaction(connection, transaction, [payer, mint], undefined);
+    });
+
+    it('can successfully initialize', async () => {
+        // await expect(
```
mistersimon, line 79: Whats the intended behaviour of there is insufficient funds for rent? It current gets a timeout error waiting for the block hash. Is it something we should test for?

token/js/test/e2e-2022/tokenMetadata.test.ts
```diff
@@ -0,0 +1,444 @@
...
+        expect(
+            await getEmittedTokenMetadata(
+                connection,
+                '4rPdnCjZkwt4CR2mih8HiepAxTmh4UFpks9vUjNLFJbrLrX3aJWss29NBzCwgxZrxVB1chYLC2YxqPfT44aRFVaG' // Random Transaction
+            )
+        ).to.deep.equal(null);
+    });
+});
+
+describe.only('Token Metadata operations', () => {
```
mistersimon, line 168: Note to self: remove the .only

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,390 @@
...
+ * Calculates additional lamports need to variable length extension
+ *
+ * @param connection       Connection to use
+ * @param address          Mint Account
+ * @param tokenMetadata    Token Metadata
+ * @param programId        SPL Token program account
+ *
+ * @return lamports to send
+ */
+async function getAdditionalRentForNewMetadata(
```
mistersimon, line 29: Do you have any suggestions for how to implement this? I don't think the implementation mimics very well how the rust code does it, its not very generic - and it doesn't have the complexity of handling updates yet.


For updates, the only way i can think of is basically writing "reducer" function like `(metadata, updates) => newMetadata` and then getting calculating the length on this new Metadata. I think its relatively straightforward to implemented but there are a few edge cases like handling Fields and adding/removing/changing additional metadata).

Wanted to check if there is a better way, before i go and do this cause it feels like we are re-implementing the rust code.

mistersimon (commented):
Hey @buffalojoec - hit a bit of blocker when trying to implement the `tokenMetadataUpdateFieldWithRentTransfer`. 

Wanted to see if there is some elegant way of calculating how much lamports is needed doing it before adding a bunch of code/logic to get it. 

Take a look at the comment above

token/js/src/extensions/tokenMetadata/actions.ts
buffalojoec: You'll have to make this function take a `connection: Connection` arg and then do exactly what the Rust client does.

First get the mint with `connection.getAccountInfo(..)`, check the new metadata size against the existing one to see if more rent is needed, and if so use `connection.getMinimumBalanceForRentExemption(..)` to get those additional required lamports.

token/js/test/e2e-2022/tokenMetadata.test.ts
buffalojoec: I'm not sure why you're experiencing a timeout. The proper error if a mint does not have enough lamports for rent should be the specific error that says "insufficient lamports for rent".

token/js/src/extensions/tokenMetadata/state.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @return Token Metadata information
+ */
+export async function getEmittedTokenMetadata(
+    connection: Connection,
+    signature: string,
+    commitment?: Finality,
+    programId = TOKEN_2022_PROGRAM_ID
+): Promise<TokenMetadata | null> {
+    const tx: any = await connection.getTransaction(signature, {
```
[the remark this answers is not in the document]
buffalojoec: Honestly I would just completely remove this function. `Emit` is intended for view functions which may or may not be relevant at the moment. As soon as there's a need for this type of function in JS, we can add it, but until then let's keep it simple.

token/js/src/extensions/tokenMetadata/actions.ts
buffalojoec: You can possibly piggyback on the `getTokenMetadata` which you've written already

token/js/src/extensions/extensionType.ts
```diff
@@ -79,6 +80,8 @@ export function getTypeLen(e: ExtensionType): number {
             return TRANSFER_HOOK_SIZE;
         case ExtensionType.TransferHookAccount:
             return TRANSFER_HOOK_ACCOUNT_SIZE;
+        case ExtensionType.TokenMetadata:
```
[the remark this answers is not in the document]
buffalojoec: Yeah, nice! How about a tad more verbose: `Cannot get type length for variable extension type: {e}`

token/js/test/e2e-2022/tokenMetadata.test.ts
mistersimon: Are we sure its suppose to throw an error? I can't figure out whats wrong.

I get the `insufficient lamports for rent` when I am creating the mint, but not when i'm just trying to initialise the metadata data in this test. The mint is created, but doesn't have enough lamports to hold the token metadata. 

From my basic understanding of the rust code - it actually doesn't throw an error, but assumes there is enough space?
https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/src/extension/token_metadata/processor.rs#L101-L101

token/js/src/extensions/tokenMetadata/state.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    if (field === Field.Uri || field === 'Uri' || field === 'uri') {
+        return 'uri';
+    }
+
+    return field;
+};
+
+export const updateTokenMetadata = (
```
mistersimon, line 26: I don't see any ways to simply this function but i'm open to ideas - it feels a bit overkill and might be missing some edge cases. 

We only really need the change in length,  concerned if we have this we the behaviour in sync with the rust implementation. Is this an issue?. 

Alternatively we could change this to just calculate the change in data len - I didn't do it because i thought this better maps to the domain logic.

mistersimon, line 26: Any ideas how to simplify this function. It feels a bit overkill cause we only really need the change in length, but this lets us reuse the `pack` function.

token/js/src/extensions/tokenMetadata/state.ts
```diff
@@ -0,0 +1,105 @@
+import type { Commitment, Connection } from '@solana/web3.js';
+import { PublicKey } from '@solana/web3.js';
+import type { TokenMetadata } from '@solana/spl-token-metadata';
+import { Field, unpack } from '@solana/spl-token-metadata';
+
+import { TOKEN_2022_PROGRAM_ID } from '../../constants.js';
+import { ExtensionType, getExtensionData } from '../extensionType.js';
+import { getMint } from '../../state/mint.js';
+
+export const getNormalizedTokenMetadataField = (field: Field | string): string => {
```
mistersimon, line 10: would have put this in `@solana/spl-token-metadata` but not sure if its worth the admin overhead to publish a new version

token/js/test/e2e-2022/tokenMetadata.test.ts
```diff
@@ -0,0 +1,421 @@
...
+            uri: 'uri',
+            additionalMetadata: [],
+        });
+    });
+
+    it('can successfully emit', async () => {
+        const signature = await tokenMetadataEmit(connection, payer, mint.publicKey);
+
+        // Get metadata emitted in the transaction to verify
+        const tx: any = await connection.getTransaction(signature, {
```
mistersimon, line 243: I removed the `getEmittedTokenMetadata` but kept the logic here just to check the emit worked. 

Can remove if you want - feels more like i'm testing the token program rather than the js client

token/js/src/extensions/tokenMetadata/actions.ts
mistersimon: Okay i've gotten something that works. I tried to make it as simple as possible, but still feels quite complicated. Let me know if you think theres anything that can be simplified

mistersimon (commented):
Hey @buffalojoec - think we should be close to wrapping this up. 

Would appreciate your input on the `...WithRentTransfer` actions. I tried to keep it as simple / understandable as possible, but its still quite a bit of logic

token/js/src/extensions/tokenMetadata/state.ts
buffalojoec: Actually I think it's fine to live in the token library. The metadata library already does its own "normalizing" within the `FieldCodec`, so it doesn't need this. Token needs this, and it shouldn't be exported, so it's fine here!

token/js/src/extensions/tokenMetadata/actions.ts
buffalojoec: Yeah, it does seem a little bit complicated. I think we can tighten up some of the null/error handling into fewer lines, and then we can simplify the size comparison.

I left a comment below.

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    const data = getExtensionData(ExtensionType.TokenMetadata, mint.tlvData);
+
+    const currentDataLen = data ? data.length : 0;
+
+    const newMeta = data === null ? getNewMetadata({} as TokenMetadata) : getNewMetadata(unpack(data));
+    const newDataLen = pack(newMeta).length;
+
+    let newAccountLen = info.data.length + (newDataLen - currentDataLen);
```
buffalojoec, line 53: We might need to roll something like this in the JS library, for general use:

https://github.com/solana-labs/solana-program-library/blob/b49a7ceebb900f64ec0389b0ce1e35579a5c2ade/token/program-2022/src/extension/mod.rs#L371-L391

token/js/src/extensions/tokenMetadata/state.ts
mistersimon: I removed support to update `mint` and `updateAuthority` - don't see why the js library would need it and it introduces some complexity

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,387 @@
...
+
+    if (currentDataLen === 0) {
+        // Extension not initialized
+        // Need 2 bytes extension discriminator, 2 bytes length
+        newAccountLen += 4;
+    }
+
+    const newRentExemptMinimum = await connection.getMinimumBalanceForRentExemption(newAccountLen);
+
+    return newRentExemptMinimum - info.lamports;
```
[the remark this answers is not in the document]
mistersimon: Thanks for the suggestions - managed to make it simpler. 

Also - you are completely right on that check. Did that small optimisation on the update field

mistersimon (commented):
Thanks for the suggestion to split new/update - I think its much easier to follow now. 

Still need that generic variable extension length handling.  Would like to get your thoughts what you had in mind as i'm not sure if it can be so generic as to handle anything, or calls to separate functions per extension

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,387 @@
...
+    programId = TOKEN_2022_PROGRAM_ID
+): Promise<number> {
+    const info = await connection.getAccountInfo(address);
+    const mint = unpackMint(address, info, programId);
+
+    if (!info) {
+        // unpack mint does error checking on info internally
+        // Check here for typescript and just in-case
+        throw new TokenAccountNotFoundError();
+    }
```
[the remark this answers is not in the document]
buffalojoec: Ha! Fair enough. If it's been used for mints in the library already, then we can use it here!

token/js/src/extensions/extensionType.ts
```diff
@@ -221,3 +231,57 @@ export function getAccountLenForMint(mint: Mint): number {
...
+    e: ExtensionType,
+    extensionLen: number | GetUpdatedVariableExtensionLen,
+    programId = TOKEN_2022_PROGRAM_ID
+): number {
+    const mint = unpackMint(address, info, programId);
+
+    const data = getExtensionData(e, mint.tlvData);
+
+    // Need 2 bytes extension discriminator, 2 bytes length
+    let newExtensionLen = 2 + 2;
```
mistersimon, line 250: You mentioned its a 4 byte discriminator - but I have it as 2. The test cases work which leads me to believe its correct. Should it be 4?
 
https://github.com/solana-labs/solana-program-library/pull/5667#discussion_r1412247485

token/js/src/extensions/tokenMetadata/actions.ts
mistersimon: I've put something together - let me know if its the right approach. 

Some things i want to check:

1. You mention a 4 byte discriminator - but I have a 2 byte one. The test cases passing leads me to believe its correct - but should it be 4?

2. Do we need to recalculate the base length in the js implementation?  I believe just taking the account length and subtracting the current extension length and adding the length length should be sufficient?

I started calling [`getExtensionTypes`](https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/extensions/extensionType.ts#L207) and feeding into [`getLen`](https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/extensions/extensionType.ts#L166C10-L166C16) - but needs some extra logic around variable length extensions - which i'm unsure we need.

buffalojoec: > 1. You mention a 4 byte discriminator - but I have a 2 byte one. The test cases passing leads me to believe its correct - but should it be 4?

Oops, my bad. 2 is correct.
https://github.com/solana-labs/solana-program-library/blob/5da184aed56cda46b0006ebf3ea5d617ef0501a8/token/program-2022/src/extension/mod.rs#L902

> 2. Do we need to recalculate the base length in the js implementation?  I believe just taking the account length and subtracting the current extension length and adding the length length should be sufficient (which is what I did)?

The base length is static for mints or accounts. What the function needs to do is check if the extension exists, if it does calculate the size difference, if it doesn't get the new size.

buffalojoec (commented):
I tried to suggest some light refactoring to simplify the additional rent approach and align more with the Rust client.

Let's tighten this stuff down, and then I'll give it a good once-over and we can hopefully get this thing stamped!

I appreciate you keeping on it!

buffalojoec:
Hey @mistersimon as usually I really appreciate the continued effort!

I wanted to reach out and see if you were interested in continuing this PR, or if you'd like me to take it over. I only ask because we may need this feature soon.

If I don't hear from you by the weekend I'll probably build off what you've got and wrap this thing up, not a problem! Again, thanks!

mistersimon:
> Hey @mistersimon as usually I really appreciate the continued effort!
> 
> I wanted to reach out and see if you were interested in continuing this PR, or if you'd like me to take it over. I only ask because we may need this feature soon.
> 
> If I don't hear from you by the weekend I'll probably build off what you've got and wrap this thing up, not a problem! Again, thanks!

Hey @buffalojoec - thanks for reaching out. 

Yeah have been a bit busy with other stuff and haven't had as much time to work on this as I had hope. I have carved out some time this weekend though to finish up the outstanding issues.

But completely understand you all have your own schedule, and i have taken a while to finish this. No worries if you feel the need to take over - just let me know!

buffalojoec:
> Yeah have been a bit busy with other stuff and haven't had as much time to work on this as I had hope. I have carved out some time this weekend though to finish up the outstanding issues.
> 
> But completely understand you all have your own schedule, and i have taken a while to finish this. No worries if you feel the need to take over - just let me know!

If you've got the time and you want to give it another go this weekend I'm totally cool with that!

If at any point you decide you're too busy or otherwise don't want to keep hammering on this, just throw a flare up.

token/js/src/extensions/extensionType.ts
```diff
@@ -221,3 +242,44 @@ export function getAccountLenForMint(mint: Mint): number {
...
+    info: AccountInfo<Buffer>,
+    address: PublicKey,
+    e: ExtensionType,
+    extensionLen: number,
+    programId = TOKEN_2022_PROGRAM_ID
+): number {
+    const data = getExtensionDataFromAccountInfo(address, info, e, programId);
+
+    // 2 bytes discriminator, 2 bytes length, extensionLen
+    const newExtensionLen = 2 + 2 + extensionLen;
```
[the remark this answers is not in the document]
mistersimon: Missed those - changed!

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,406 @@
...
+/**
+ * Calculates additional lamports need for variable length extension
+ *
+ * @param connection       Connection to use
+ * @param address          Mint Account
+ * @param tokenMetadata    Token Metadata
+ * @param programId        SPL Token program account
+ *
+ * @return lamports to send
+ */
```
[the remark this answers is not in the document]
mistersimon: Thanks for clarifying - wasn't sure when doc string was need and when not. This makes sense, removed

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,406 @@
...
+ * Calculates additional lamports to update field
+ *
+ * @param connection       Connection to use
+ * @param address          Mint Account
+ * @param field            Field to update in the metadata
+ * @param value            Value to write for the field
+ * @param programId        SPL Token program account
+ *
+ * @return lamports to send
+ */
```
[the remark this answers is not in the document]
mistersimon: Removed

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,406 @@
...
+    const transaction = new Transaction();
+
+    const lamports = await getAdditionalRentForNewMetadata(connection, mint, {
+        updateAuthority,
+        mint,
+        name,
+        symbol,
+        uri,
+        additionalMetadata: [],
+    });
```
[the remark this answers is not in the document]
mistersimon: Good spot. Also missed it on `getAdditionalRentForUpdatedMetadata` - which i added

token/js/test/e2e-2022/tokenMetadata.test.ts
```diff
@@ -0,0 +1,421 @@
...
+                payer,
+                authority,
+                mint.publicKey,
+                'name',
+                'TEST',
+                undefined,
+                undefined,
+                TEST_PROGRAM_ID
+            ),
+        ]);
```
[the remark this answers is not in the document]
mistersimon: Yes that is correct - mostly did it this way to make the tests run faster and I am checking the metadata once both complete so don't think the order should matter. 

But to make things easier I split it into sequential operations. I didn't split into separate tests to to avoid running the beforeEach hook an additional time

token/js/src/extensions/tokenMetadata/state.ts
```diff
@@ -0,0 +1,86 @@
+import type { Commitment, Connection } from '@solana/web3.js';
+import type { PublicKey } from '@solana/web3.js';
+import type { TokenMetadata } from '@solana/spl-token-metadata';
+import { Field, unpack } from '@solana/spl-token-metadata';
+
+import { TOKEN_2022_PROGRAM_ID } from '../../constants.js';
+import { ExtensionType, getExtensionData } from '../extensionType.js';
+import { getMint } from '../../state/mint.js';
+
+export const getNormalizedTokenMetadataField = (field: Field | string): string => {
```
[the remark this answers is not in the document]
mistersimon: Yep changed. 

I had to remove the test case though (which requires it to be exported) - but I thought the function is simple enough for it to be acceptable!

token/js/src/extensions/extensionType.ts
```diff
@@ -204,6 +214,17 @@ export function getExtensionData(extension: ExtensionType, tlvData: Buffer): Buf
...
+    address: PublicKey,
+    accountInfo: AccountInfo<Buffer> | null,
+    e: ExtensionType,
+    programId = TOKEN_2022_PROGRAM_ID
+): Buffer | null {
+    const mint = unpackMint(address, accountInfo, programId);
+
+    return getExtensionData(e, mint.tlvData);
+}
+
```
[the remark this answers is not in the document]
mistersimon: Fair enough - removed it

token/js/src/extensions/extensionType.ts
```diff
@@ -221,3 +242,44 @@ export function getAccountLenForMint(mint: Mint): number {
...
+): number {
+    const data = getExtensionDataFromAccountInfo(address, info, e, programId);
+
+    // 2 bytes discriminator, 2 bytes length, extensionLen
+    const newExtensionLen = 2 + 2 + extensionLen;
+
+    return info.data.length + newExtensionLen - (data?.length || 0);
+}
+
+export async function getAdditionalRentForNewExtensionLen(
```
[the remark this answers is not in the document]
mistersimon: I've seen your changes below and adopted your approach and removed this function

token/js/src/extensions/extensionType.ts
```diff
@@ -221,3 +242,44 @@ export function getAccountLenForMint(mint: Mint): number {
...
+    connection: Connection,
+    address: PublicKey,
+    e: ExtensionType,
+    extensionLen: number,
+    programId = TOKEN_2022_PROGRAM_ID,
+    accountInfo?: AccountInfo<Buffer> | null
+): Promise<number> {
+    // If account info was provided, don't fetch again
+    const info = accountInfo ?? (await connection.getAccountInfo(address));
```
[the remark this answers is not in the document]
mistersimon: Adopted your approach and removed this function

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,406 @@
...
+    programId = TOKEN_2022_PROGRAM_ID
+): Promise<number> {
+    return await getAdditionalRentForNewExtensionLen(
+        connection,
+        address,
+        ExtensionType.TokenMetadata,
+        pack(tokenMetadata).length,
+        programId
+    );
+}
```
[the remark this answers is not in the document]
mistersimon: I've changed to this approach. 

Only modification I made is to early return `0` if `newAccountLen` is less than or equal to the current account length based on this comment:
https://github.com/solana-labs/solana-program-library/pull/5667#discussion_r1409645659

(Assumed you omitted this for clarity - but let me know if it was intentional)

token/js/src/extensions/tokenMetadata/actions.ts
```diff
@@ -0,0 +1,406 @@
...
+
+    return await getAdditionalRentForNewExtensionLen(
+        connection,
+        address,
+        ExtensionType.TokenMetadata,
+        extensionLen,
+        programId,
+        info
+    );
+}
```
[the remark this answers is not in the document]
mistersimon: I've changed to this approach, but:

1. Why the `+ 1` in `connection.getMinimumBalanceForRentExemption(newAccountLen + 1);`? 

2. Don't follow your comment "unpacking the whole account data as TokenMetadata". I believe i was unpacking the extension data through my `getExtensionDataFromAccountInfo` function (which i have now removed). Perhaps the variable name was misleading? Should probably been `extensionData` vs `data`. Am i missing something?

3. Also changed to only check required rent exemption if `newAccountLen` is longer

token/js/src/extensions/extensionType.ts
```diff
@@ -221,3 +242,44 @@ export function getAccountLenForMint(mint: Mint): number {
...
+    extensionLen: number,
+    programId = TOKEN_2022_PROGRAM_ID
+): number {
+    const data = getExtensionDataFromAccountInfo(address, info, e, programId);
+
+    // 2 bytes discriminator, 2 bytes length, extensionLen
+    const newExtensionLen = 2 + 2 + extensionLen;
+
+    return info.data.length + newExtensionLen - (data?.length || 0);
+}
```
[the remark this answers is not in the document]
mistersimon: Made this changed. 

I refactored all the other uses of `+ TYPE_SIZE + LENGTH_SIZE` to also use this new function.

mistersimon:
Hey @buffalojoec - think i've picked up all your edits. Feel free to take a look and let me know if theres anything still left to do. 

I'll keep some time free next week to pick them up and get this closed out!

buffalojoec:
Hey @mistersimon thanks for the awesome work!

This PR was just about ready to go, but in the interest of time I did the little bit of cleanup and also refactored the tests a bit.

mistersimon:
Thanks @buffalojoec for getting this across the line and with all the help in reviews!

rholang:
@mistersimon 
Hi,
I am still getting "insufficient lamports" error, when executing the "create metadata" proposal on devnet (from params created). What could be the reason (there is more then enough SOL in the account)?
https://explorer.solana.com/tx/5nRBvH6zooKGZKWwxwYTab4ZEw12jh2y84VHUXQHXwvowf2MgPMUJWiC5GBRttwoqZJTpzXuG6tuULejTVv75Mzi?cluster=devnet

PR #5670, opened 2023-10-26
token-cli: Performing a confidential transfer sequentially

joncinque:
#### Problem

We need to do confidential transfers through the CLI.

#### Solution

This tries to do sequential transactions to setup the separate context state accounts, performing the transfer, and then closing the context state accounts.

The non-range proof transaction fails with:

```
called `Result::unwrap()` on an `Err` value: Client(Error { request: Some(SendTransaction), kind: RpcError(RpcResponseError { code: -32602, message: "base64 encoded solana_sdk::transaction::versioned::VersionedTransaction too large: 1672 bytes (max: encoded/raw 1644/1232)", data: Empty }) })
```

And the range proof transaction fails with:

```
called `Result::unwrap()` on an `Err` value: Client(Error { request: Some(SendTransaction), kind: RpcError(RpcResponseError { code: -32602, message: "base64 encoded solana_sdk::transaction::versioned::VersionedTransaction too large: 1812 bytes (max: encoded/raw 1644/1232)", data: Empty }) })
```

Note that this builds on #5663 -- hopefully there's a solution!


joncinque:
@samkim-crypto this rebased and ready for a look!

samkim-crypto (approved):
Great! Thanks for all the changes. Everything looks nice and clean to me!

PR #5672, opened 2023-10-26
token 2022: add alloc_and_serialize for fixed-len extensions

buffalojoec:
This PR adds the `alloc_and_serialize` function for fixed-length extensions to Token2022.

Note: I decided to follow already pre-established patterns and rename the existing `alloc_and_serialize` function to `alloc_and_serialize_variable_len_extension`. The existing pattern has been to add some form of `*variable_len*` suffix on functions that deal with variable-length methods, and give the more concisely named functions to the fixed-length extensions.

Regardless, this function allows reallocation of an account _only_ if the extension does not exist yet and requires more uninitialized bytes to write it in.

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -404,12 +404,12 @@ pub trait BaseStateWithExtensions<S: BaseState> {
     ///
     /// Provides the correct answer regardless if the extension is already present
     /// in the TLV data.
-    fn try_get_new_account_len<V: Extension + VariableLenPack>(
+    fn try_get_new_account_len<V: Extension>(
         &self,
-        new_extension: &V,
+        new_extension_len: usize,
     ) -> Result<usize, ProgramError> {
         // get the new length used by the extension
-        let new_extension_len = add_type_and_length_to_len(new_extension.get_packed_len()?);
+        let new_extension_tlv_len = add_type_and_length_to_len(new_extension_len);
         let tlv_info = get_tlv_data_info(self.get_tlv_data())?;
         // If we're adding an extension, then we must have at least BASE_ACCOUNT_LENGTH
         // and account type
@@ -417,7 +417,7 @@ pub trait BaseStateWithExtensions<S: BaseState> {
             .used_len
             .saturating_add(BASE_ACCOUNT_AND_TYPE_LENGTH);
         let new_len = if tlv_info.extension_types.is_empty() {
-            current_len.saturating_add(new_extension_len)
+            current_len.saturating_add(new_extension_tlv_len)
         } else {
             // get the current length used by the extension
             let current_extension_len = self
@@ -426,7 +426,7 @@ pub trait BaseStateWithExtensions<S: BaseState> {
                 .unwrap_or(0);
             current_len
                 .saturating_sub(current_extension_len)
-                .saturating_add(new_extension_len)
+                .saturating_add(new_extension_tlv_len)
         };
         Ok(adjust_len_for_multisig(new_len))
     }
@@ -1178,15 +1178,21 @@ impl Extension for AccountPaddingTest {
     const TYPE: ExtensionType = ExtensionType::AccountPaddingTest;
 }
 
-/// Packs a variable-length extension into a TLV space
+/// Packs a fixed-length extension into a TLV space
 ///
 /// This function reallocates the account as needed to accommodate for the
-/// change in space, then reallocates in the TLV buffer, and finally writes the
-/// bytes.
+/// change in space.
 ///
-/// NOTE: Unlike the `reallocate` instruction, this function will reduce the
-/// size of an account if it has too many bytes allocated for the given value.
-pub fn alloc_and_serialize<S: BaseState, V: Extension + VariableLenPack>(
+/// If the extension already exists, it will overwrite the existing extension
+/// if `overwrite` is `true`, otherwise it will return an error.
+///
+/// If the extension does not exist, it will reallocate the account and write
+/// the extension into the TLV buffer.
+///
+/// NOTE: Since this function deals with fixed-size extensions, it does not
+/// handle _decreasing_ the size of an account's data buffer, like the function
+/// `alloc_and_serialize_variable_len_extension` does.
+pub fn alloc_and_serialize<S: BaseState, V: Default + Extension + Pod>(
     account_info: &AccountInfo,
     new_extension: &V,
     overwrite: bool,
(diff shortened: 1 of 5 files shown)


buffalojoec:
* master
  * **PR #5672** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5672" target="_blank">Graphite</a>  👈
          * **PR #5628** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5628" target="_blank">Graphite</a> 
            * **PR #5629** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5629" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/5672?utm_source=stack-comment).

token/program-2022/src/extension/mod.rs
```diff
@@ -404,20 +404,20 @@ pub trait BaseStateWithExtensions<S: BaseState> {
     ///
     /// Provides the correct answer regardless if the extension is already present
     /// in the TLV data.
-    fn try_get_new_account_len<V: Extension + VariableLenPack>(
+    fn try_get_new_account_len<V: Extension>(
         &self,
-        new_extension: &V,
+        new_extension_len: usize,
```
joncinque, line 409: This function always kinda weirded me out -- this is much cleaner, thanks!

joncinque, line 409: This function always kinda weirded me out -- this is much cleaner, thanks!

token/program-2022/src/extension/mod.rs
```diff
@@ -404,20 +404,20 @@ pub trait BaseStateWithExtensions<S: BaseState> {
     ///
     /// Provides the correct answer regardless if the extension is already present
     /// in the TLV data.
-    fn try_get_new_account_len<V: Extension + VariableLenPack>(
+    fn try_get_new_account_len<V: Extension>(
```
joncinque, line 407: This is kind of a footgun now, since you can call `try_get_new_account_len` with a fixed-length extension and pass the wrong size.

How about making two versions, one for fixed-length, and one for variable-length, and have them call into a common private function? The variable length one can change to take in the new size if you think that's clearer

token/program-2022/src/extension/mod.rs
```diff
@@ -1178,6 +1178,58 @@ impl Extension for AccountPaddingTest {
...
+    account_info: &AccountInfo,
+    new_extension: &V,
+    overwrite: bool,
+) -> Result<(), ProgramError> {
+    let previous_account_len = account_info.try_data_len()?;
+    let (new_account_len, extension_already_exists) = {
+        let data = account_info.try_borrow_data()?;
+        let state = StateWithExtensions::<S>::unpack(&data)?;
+        let new_account_len = state.try_get_new_account_len::<V>(pod_get_packed_len::<V>())?;
+        let extension_already_exists = state.get_extension_bytes::<V>().is_ok();
```
joncinque, line 1205: nit: let's use `get_extension` so we're sure to use the correct trait bounds

token/program-2022/src/extension/mod.rs
```diff
@@ -1178,6 +1178,58 @@ impl Extension for AccountPaddingTest {
...
+        // Realloc the account, then write the new extension
+        account_info.realloc(new_account_len, false)?;
+        let mut buffer = account_info.try_borrow_mut_data()?;
+        if previous_account_len <= BASE_ACCOUNT_LENGTH {
+            set_account_type::<S>(*buffer)?;
+        }
+        let mut state = StateWithExtensionsMut::<S>::unpack(&mut buffer)?;
+        let extension = state.init_extension::<V>(false)?;
+        *extension = *new_extension;
+    }
```
joncinque, line 1229: We can probably shorten this some to avoid duplication of logic, ie. the checking during `init_extension`. If the previous block only gives you the `new_account_len`, then you should be able to do:

```suggestion
    // Realloc the account first, if needed
    if new_account_len > previous_account_len {
        account_info.realloc(new_account_len, false)?;
    }
    let mut buffer = account_info.try_borrow_mut_data()?;
    if previous_account_len <= BASE_ACCOUNT_LENGTH {
        set_account_type::<S>(*buffer)?;
    }
    let mut state = StateWithExtensionsMut::<S>::unpack(&mut buffer)?;
    
    // Write the extension
    let extension = state.init_extension::<V>(overwrite)?;
    *extension = *new_extension;
```

What do you think?

joncinque (commented):
This is looking very close! Just a few nits, but the core is good. Could you add a test for the new function?

token/program-2022/src/extension/mod.rs
buffalojoec: This is what I had, believe it or not, but I wanted to minimize changes to this code. But I can return to that!

token/program-2022/src/extension/mod.rs
buffalojoec: Ohhh so just providing overwrite into the `init_extension(..)`? I dig it! Nice touch-up.

buffalojoec:
> Could you add a test for the new function?

Yep, I was mostly waiting to see if you were okay with changing this many bits of the TLV/extension module. Didn't want to make too many changes.

Will definitely add some tests!

token/program-2022/src/extension/mod.rs
```diff
@@ -353,6 +353,37 @@ fn get_extension_bytes_mut<S: BaseState, V: Extension>(
...
+    // get the new length used by the extension
+    let new_extension_tlv_len = add_type_and_length_to_len(new_extension_len);
+    let tlv_info = get_tlv_data_info(tlv_data)?;
+    // If we're adding an extension, then we must have at least BASE_ACCOUNT_LENGTH
+    // and account type
+    let current_len = tlv_info
+        .used_len
+        .saturating_add(BASE_ACCOUNT_AND_TYPE_LENGTH);
+    let new_len = if tlv_info.extension_types.is_empty() {
+        current_len.saturating_add(new_extension_tlv_len)
```
joncinque, line 374: I did some testing around, and this branch could be safely removed without affecting the correctness of the code. How about we just remove it?

token/program-2022/src/extension/mod.rs
```diff
@@ -398,37 +429,25 @@ pub trait BaseStateWithExtensions<S: BaseState> {
             Ok(adjust_len_for_multisig(total_len))
         }
     }
+    /// Calculate the new expected size if the state allocates the given number
+    /// of bytes for the given fixed-length extension type.
```
joncinque, line 433: nit:
```suggestion
    /// Calculate the new expected size if the state allocates the given fixed-length extension type
    /// If the state already has the extension, the resulting account len will be unchanged.
```

token/program-2022/src/extension/mod.rs
```diff
@@ -398,37 +429,25 @@ pub trait BaseStateWithExtensions<S: BaseState> {
...
+        )
+    }
 
     /// Calculate the new expected size if the state allocates the given number
-    /// of bytes for the given extension type.
-    ///
-    /// Provides the correct answer regardless if the extension is already present
-    /// in the TLV data.
-    fn try_get_new_account_len<V: Extension + VariableLenPack>(
+    /// of bytes for the given variable-length extension type.
```
joncinque, line 442: nit: (also get rid of the last word on the previous line)
```suggestion
    /// variable-length extension instance.
```

token/program-2022/src/extension/mod.rs
```diff
@@ -907,12 +926,15 @@ pub enum ExtensionType {
...
     /// Mint contains a pointer to another account (or the same account) that holds group
     /// configurations
     GroupPointer,
+    /// Test variable-length mint extension
     #[cfg(test)]
-    VariableLenMintTest = u16::MAX - 2,
+    VariableLenMintTest = u16::MAX - 4,
+    /// Test fixed-length mint extension
+    #[cfg(test)]
+    FixedLenMintTest = u16::MAX - 3,
```
joncinque, line 937: What does this give you over `MintPaddingTest`?

token/program-2022/src/extension/mod.rs
```diff
@@ -907,12 +926,15 @@ pub enum ExtensionType {
...
     /// Mint contains token-metadata
     TokenMetadata,
-    /// Test variable-length mint extension
     /// Mint contains a pointer to another account (or the same account) that holds group
     /// configurations
     GroupPointer,
+    /// Test variable-length mint extension
     #[cfg(test)]
-    VariableLenMintTest = u16::MAX - 2,
+    VariableLenMintTest = u16::MAX - 4,
```
joncinque, line 934: nit: this is off-by-one, since `u16::MAX` should be allocated to the last one

token/program-2022/src/extension/mod.rs
```diff
@@ -1195,7 +1264,8 @@ pub fn alloc_and_serialize<S: BaseState, V: Extension + VariableLenPack>(
     let (new_account_len, extension_already_exists) = {
         let data = account_info.try_borrow_data()?;
         let state = StateWithExtensions::<S>::unpack(&data)?;
-        let new_account_len = state.try_get_new_account_len(new_extension)?;
+        let new_account_len =
+            state.try_get_new_account_len_for_variable_len_extension::<V>(new_extension)?;
```
joncinque, line 1268: nit: is the `V` needed here? ie can you do
```suggestion
            state.try_get_new_account_len_for_variable_len_extension(new_extension)?;
```

joncinque (commented):
Just a few last little things, then this is good to go! The tests look great

token/program-2022/src/extension/mod.rs
buffalojoec: That's weird I plucked this from the original. Removed.

token/program-2022/src/extension/mod.rs
buffalojoec: Just felt like it made more sense to have specific test extensions for specific tests, but if I can use `MintPaddingTest` in the fixed-length tests I can do that!

joncinque (approved):
Thanks for integrating those changes!
fixed

PR #5674, opened 2023-10-27
[token-cli] sequentially process confidential transfer withdraw in token-cli

samkim-crypto:
#### Problem
We can't do confidential token withdrawal through the cli due to the transaction size limit

#### Summary of changes
Divide up the process of creating proof context states and the actual token withdraw.

### DIFF
--- a/token/cli/src/main.rs
+++ b/token/cli/src/main.rs
@@ -3409,17 +3409,47 @@ async fn command_deposit_withdraw_confidential_tokens(
                 state_with_extension.get_extension::<ConfidentialTransferAccount>()?;
             let withdraw_account_info = WithdrawAccountInfo::new(extension_state);
 
+            let context_state_authority = config.fee_payer()?;
+            let context_state_keypair = Keypair::new();
+            let context_state_pubkey = context_state_keypair.pubkey();
+
+            let withdraw_proof_data =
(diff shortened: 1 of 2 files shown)


token/cli/src/main.rs
```diff
@@ -3409,18 +3409,48 @@ async fn command_deposit_withdraw_confidential_tokens(
...
+            // close context state account
+            let context_state_authority_pubkey = context_state_authority.pubkey();
+            let close_context_state_signers = &[context_state_authority];
+            token
+                .confidential_transfer_close_context_state(
+                    &context_state_pubkey,
+                    &token_account_address,
+                    &context_state_authority_pubkey,
+                    close_context_state_signers,
+                )
```
joncinque, line 3453: Is it possible to do the withdraw and close in the same transaction? If so, it would be good to do it in this PR

joncinque (approved):
Great work! Just one suggestion to speed up the withdrawal to go down to 3 dependent transactions instead of 4

token/cli/src/main.rs
samkim-crypto: Yes, good idea!

samkim-crypto: Added issue https://github.com/solana-labs/solana-program-library/issues/5685, which will be addressed in a follow-up PR.

PR #5688, opened 2023-10-29
token-cli: Cleanup `conflicts_with` and output colons

joncinque:
#### Problem

The `conflicts_with` on confidential token `deposit` and `withdraw` makes it unusable with `--address`.

Also, colons `:` appear twice while displaying confidential transfer mint info.

#### Solution

In separate commits:

* remove `conflicts_with("address")` for deposit and withdraw
* remove the extra colon


samkim-crypto (approved):
Oh oops. Thanks for the catch!

PR #5722, opened 2023-11-03
cli: spl-token, fix parameter parsing bug

norwnd:
Possibly resolves https://github.com/solana-labs/solana/issues/33938

I was browsing through this code, and this looks like a typo(bug) to me.


buffalojoec:
Hey @norwnd thanks for contributing and working to solve this issue!

I was just looking into this, too when @joncinque shipped this fix #5734 

The issue wasn't that it was supposed to be `mint_authority`, but rather the command was defined as `program_authority` but the parser was looking for `authority`. Obviously it doesn't find it, so it defaults to the local keypair, as mentioned in the issue you commented on.

Thanks again!

PR #5729, opened 2023-11-06
`rustfmt.toml` for SPL

buffalojoec:
This PR adds a `rustfmt.toml` to the SPL repository.

I've included only the file itself with configs for review. Once we agree on the configs for the whole workspace, I can run the formatter and PR the changes.


buffalojoec:
Note this will also include the mandatory use of `nightly`, thus CI modifications will be required as part of this change.

token/rustfmt.toml
```diff
@@ -0,0 +1,10 @@
+comment_width = 80
+edition = "2021"
+format_code_in_doc_comments = true
+format_strings = true
+group_imports = "One"
+ignore = [
+  "**/*/entrypoint.rs"
+]
```
joncinque, line 8: Why is this here? We should definitely format entrypoints too!

joncinque (commented):
How about we embrace these different features one by one? It seems like this will be chaos if we just enable all of them in one commit.

Let's start with `group_imports` and `imports_granularity`, and then we can see how each option impacts the repo before adding it.

Maybe we can even start with using nightly format in CI before adding this file. What do you think?

token/rustfmt.toml
buffalojoec: Because `rustfmt` will turn this:

```rust
use solana_program::{entrypoint, entrypoint::ProgramResult};
```

into this, which doesn't import the `entrypoint!` macro:

```rust
use solana_program::entrypoint::{self, ProgramResult};
```

I tried scouring the `rustfmt` docs for how to disable this import syntax alone, but didn't seem to find anything! But I agree, for the repo it should probably not be this general.

joncinque: bah that stinks. How about we change everywhere to call the macro with `solana_program::entrypoint!(process_instruction)`?

buffalojoec: This is the way

buffalojoec:
> Let's start with `group_imports` and `imports_granularity`, and then we can see how each option impacts the repo before adding it.
> 
> Maybe we can even start with using nightly format in CI before adding this file. What do you think?

yeah this is going to result in a bazillion changes so I'm in favor of splitting it up.

buffalojoec:
Closing in favor of #5741

PR #5734, opened 2023-11-06
token-cli: Fix set-transfer-hook-program "authority" flag

joncinque:
#### Problem

The `set-transfer-hook-program` command allows a flag called `program-authority` to set the current authority, but the code then looks for `authority`, making the flag unusable.

#### Solution

Just call it "authority" everywhere


joncinque:
Landing early to unblock

PR #5775, opened 2023-11-08
js: Move everything to pnpm

joncinque:
#### Problem

There's a lot of libraries in SPL, and managing all of their dependencies can take awhile, especially as all of the dependabot PRs come in.

#### Solution

Move everything to one pnpm-lock.yaml file! As I was going through the different packages, there were loads of issues, so I had to fix up a few other things. This is step 1, and after this, we can change all of the deps in the package.json files to use the version from `workspace:*`

Note: this is currently set to draft while CI gets happy


.github/workflows/pull-request-stake-pool.yml
```diff
@@ -99,11 +98,6 @@ jobs:
           key: node-${{ hashFiles('pnpm-lock.yaml') }}
           restore-keys: |
             node-
-      - name: Download programs
-        uses: actions/download-artifact@v2
-        with:
-          name: stake-pool-programs
-          path: target/deploy
```
buffalojoec, line 103: How come we don't need to download the programs anymore?

token-swap/js/package.json
```diff
@@ -32,7 +32,7 @@
   "scripts": {
     "build": "tsc -p tsconfig.json && tsc-esm -p tsconfig.json && tsc -p tsconfig.cjs.json",
     "postbuild": "echo '{\"type\":\"commonjs\"}' > dist/cjs/package.json && echo '{\"type\":\"module\"}' > dist/esm/package.json",
-    "test": "ts-node test/main.ts",
+    "test": "mocha test",
```
buffalojoec, line 36: This is fine, but I'm curious if we want align on either `jest` or `mocha` for all JS libs?

token-swap/js/.eslintrc
buffalojoec, line 1: We _could_ enforce `eslint` and `prettier` at the root level for all libs (like `rustfmt`). `pnpm` will pick it up.

Also, have you seen `@solana/eslint-config-solana`?

buffalojoec (commented):
I love it! This should make managing JS libraries in SPL much easier.

stake-pool/js/package.json
```diff
@@ -49,7 +49,8 @@
     "@solana/spl-token": "^0.3.7",
     "@solana/web3.js": "^1.76.0",
     "bn.js": "^5.2.0",
-    "buffer": "^6.0.3"
+    "buffer": "^6.0.3",
+    "superstruct": "^0.14.2"
```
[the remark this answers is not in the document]
joncinque: Yeah, I guess it was picking it up through a transitive dependency, but the older build tools didn't care?

.github/workflows/pull-request-stake-pool.yml
joncinque: The JS tests don't actually use the program, so I decided to simplify it for faster runs

token-swap/js/package.json
joncinque: We probably want to use `jest`, but it was such a nightmare to get right in the memo directory I decided to keep it simple here. We're probably dropping all of this stuff soon anyway

token-swap/js/.eslintrc
joncinque: Ah cool, I added that as an item in #5785, and I can definitely use the `@solana` version!

token-swap/js/package.json
buffalojoec: You mean the example programs?

joncinque: yeah, a lot of the unmaintained reference implementations

buffalojoec (approved):

PR #5779, opened 2023-11-08
single-pool-js: redo packaging

2501babe:
please work please work please work


2501babe:
alright i think... im finally free. cc @steveluscher @joncinque you dont need to code review im gonna merge as-is but you might be interested in the details of what i can only pray to god is the resolution to all this

the tldr is use tsc to produce separate es6 and cjs versions of the library, do not use `type` in the main `package.json` but monkeypatch it in `dist/`. the only weird fucking thing about all this is i cant for the life of me get `import './internal'` to work, so i added `.js` extensions (in ts files... lol...) as far as i can tell node _deliberately broke_ filename resolution in 19, which would explain why stake pool works fine on 18... i dont know

joncinque:
This removed `solana-bankrun` and still somehow passed, so I looked at the CI, and it seems like it's still broken unfortunately https://github.com/solana-labs/solana-program-library/actions/runs/6800097574/job/18488277665#step:6:317

2501babe:
well the good news is that's unrelated, i removed bankrun from the one of the two packages that doesnt have tests in it, so it shouldnt affect anything

the bad news... is:
* i have no idea why the script failing still makes ci think it passed...?
* this is because node removed _another_ flag i depend on to make node actually work. and they did it on a minor version of the lts release. this should pass on 20.5 but starts failing with 20.6

i will... take care of it

joncinque:
Ah ok, that makes sense! Sorry for not looking further into it. But yeah, something strange is going on...

2501babe:
oh i know why, its because i modified the test script to be `edit package.json && run test ; edit package.json` because the reversion needs to happen regardless of if the tests pass or fail but this means now the test command always returns 0. annoying

2501babe:
thank you for catching this!

PR #5788, opened 2023-11-09
docs: Update token-2022 extensions

joncinque:
#### Problem

We've added a lot of functionality for using extensions in the CLI, but it isn't exposed in the docs.

#### Solution

This is a bit of a kitchen sink of changes, but it should update everything roughly to where we are.

Someone also pointed out that the command were incorrect for transfer-hooks, so I decided to make those consistent


samkim-crypto (approved):
Looks great!

PR #5798, opened 2023-11-09
dependabot: Update npm to root to properly update dependencies

joncinque:
#### Problem

The dependabot PRs don't properly update the lockfile, probably because we have a job for each JS library, where it doesn't look for the pnpm lockfile.

#### Solution

Update the npm directory to the root of the repo and remove all of the copied jobs. While I was at it, I updated to UTC to make it clearer when it runs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,91 +4,19 @@ updates:
   directory: "/"
   schedule:
     interval: daily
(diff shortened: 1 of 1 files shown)


joncinque:
No need for CI here -- we'll see how it runs tomorrow

PR #5799, opened 2023-11-09
README: Use commit hashes for audits

joncinque:
#### Problem

As mentioned at https://github.com/solana-labs/solana-program-library/pull/5639#discussion_r1369284335, it's clearer to use the commit hash of each audit when mentioning when the program was audited.

#### Solution

Do that!


samkim-crypto (approved):
Looks fine to me!

PR #5804, opened 2023-11-11
token-cli: Refactor tests out of `main.rs`

joncinque:
#### Problem

The token-cli tests are very slow and often completely fail CI, probably because it's spinning up and killing so many test validators.

#### Solution

This doesn't solve it yet, but as a first step before using a custom test runner, we need to factor the tests out of `main.rs`. To summarize the changes:

* split main.rs into `clap_app.rs`, `command.rs`, and `tests/command.rs`
* only expose `process_command` from `command.rs`
* only use `process_command` in the tests rather than the direct commands
* fix an issue when trying to set the auditor key and the approve policy for a confidential mint at the same time 65bdeb4d07ad6bada6bd59e434d44af6f52624fd <-- only functional change

Up to you how much you want to read, it's kind messy, but it gets the job done! Instead of one file with 10k lines, we have 3 files with ~3k lines, so that's progress!


token/cli/src/clap_app.rs
buffalojoec, line 1: This commit really doesn't belong in here, since it seems unrelated to refactoring. But that's up to you, it's a small change!

buffalojoec (approved):
Lgtm, are you planning to move the parallelization to a later PR? I saw you added it then reverted it.

joncinque:
> Lgtm, are you planning to move the parallelization to a later PR? I saw you added it then reverted it.

Yeah if we just add it here, it's too much for the CI machines, but it was faster for local testing. The next step is to create a new "main" function which spins up the test validator and then passes a payer and RPC client to each test so they can run in parallel without overtaxing the system with 43 concurrent test validators.

buffalojoec:
> The next step is to create a new "main" function which spins up the test validator and then passes a payer and RPC client to each test so they can run in parallel without overtaxing the system

Awesome!! Very cool.

PR #5805, opened 2023-11-11
token 2022: add metadata pointer extension to js @solana/spl-token client

mistersimon:
Adding support for the metadata pointer extension

Split out of wider task to add metadata support for ease of review
https://github.com/solana-labs/solana-program-library/pull/5667

### DIFF
--- a/token/js/src/extensions/extensionType.ts
+++ b/token/js/src/extensions/extensionType.ts
@@ -8,6 +8,7 @@ import { DEFAULT_ACCOUNT_STATE_SIZE } from './defaultAccountState/index.js';
 import { IMMUTABLE_OWNER_SIZE } from './immutableOwner.js';
 import { INTEREST_BEARING_MINT_CONFIG_STATE_SIZE } from './interestBearingMint/state.js';
 import { MEMO_TRANSFER_SIZE } from './memoTransfer/index.js';
+import { METADATA_POINTER_SIZE } from './metadataPointer/state.js';
 import { MINT_CLOSE_AUTHORITY_SIZE } from './mintCloseAuthority.js';
 import { NON_TRANSFERABLE_SIZE, NON_TRANSFERABLE_ACCOUNT_SIZE } from './nonTransferable.js';
 import { PERMANENT_DELEGATE_SIZE } from './permanentDelegate.js';
@@ -15,22 +16,26 @@ import { TRANSFER_FEE_AMOUNT_SIZE, TRANSFER_FEE_CONFIG_SIZE } from './transferFe
 import { TRANSFER_HOOK_ACCOUNT_SIZE, TRANSFER_HOOK_SIZE } from './transferHook/index.js';
 
 export enum ExtensionType {
-    Uninitialized,
-    TransferFeeConfig,
-    TransferFeeAmount,
-    MintCloseAuthority,
-    ConfidentialTransferMint,
-    ConfidentialTransferAccount,
-    DefaultAccountState,
-    ImmutableOwner,
-    MemoTransfer,
-    NonTransferable,
-    InterestBearingConfig,
-    CpiGuard,
-    PermanentDelegate,
-    NonTransferableAccount,
-    TransferHook,
-    TransferHookAccount,
+    Uninitialized = 0,
+    TransferFeeConfig = 1,
+    TransferFeeAmount = 2,
+    MintCloseAuthority = 3,
+    ConfidentialTransferMint = 4,
+    ConfidentialTransferAccount = 5,
+    DefaultAccountState = 6,
+    ImmutableOwner = 7,
+    MemoTransfer = 8,
+    NonTransferable = 9,
+    InterestBearingConfig = 10,
+    CpiGuard = 11,
+    PermanentDelegate = 12,
+    NonTransferableAccount = 13,
+    TransferHook = 14,
+    TransferHookAccount = 15,
+    // ConfidentialTransferFee = 16, // Not implemented yet
+    // WithdrawExcessLamports = 17, // Not implemented yet
+    MetadataPointer = 18,
+    // GroupPointer = 19, // Not implemented yet
 }
 
 export const TYPE_SIZE = 2;
@@ -60,6 +65,8 @@ export function getTypeLen(e: ExtensionType): number {
             return IMMUTABLE_OWNER_SIZE;
         case ExtensionType.MemoTransfer:
             return MEMO_TRANSFER_SIZE;
+        case ExtensionType.MetadataPointer:
+            return METADATA_POINTER_SIZE;
         case ExtensionType.NonTransferable:
             return NON_TRANSFERABLE_SIZE;
         case ExtensionType.InterestBearingConfig:
@@ -87,6 +94,7 @@ export function isMintExtension(e: ExtensionType): boolean {
         case ExtensionType.InterestBearingConfig:
         case ExtensionType.PermanentDelegate:
         case ExtensionType.TransferHook:
+        case ExtensionType.MetadataPointer:
             return true;
         case ExtensionType.Uninitialized:
         case ExtensionType.TransferFeeAmount:
@@ -121,6 +129,7 @@ export function isAccountExtension(e: ExtensionType): boolean {
         case ExtensionType.InterestBearingConfig:
         case ExtensionType.PermanentDelegate:
         case ExtensionType.TransferHook:
+        case ExtensionType.MetadataPointer:
             return false;
         default:
             throw Error(`Unknown extension type: ${e}`);
@@ -144,6 +153,7 @@ export function getAccountTypeOfMintType(e: ExtensionType): ExtensionType {
         case ExtensionType.ImmutableOwner:
         case ExtensionType.MemoTransfer:
         case ExtensionType.MintCloseAuthority:
+        case ExtensionType.MetadataPointer:
         case ExtensionType.Uninitialized:
         case ExtensionType.InterestBearingConfig:
         case ExtensionType.PermanentDelegate:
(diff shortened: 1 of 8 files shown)


token/js/src/extensions/extensionType.ts
```diff
@@ -8,29 +8,34 @@ import { DEFAULT_ACCOUNT_STATE_SIZE } from './defaultAccountState/index.js';
...
-    ImmutableOwner,
-    MemoTransfer,
-    NonTransferable,
-    InterestBearingConfig,
-    CpiGuard,
-    PermanentDelegate,
-    NonTransferableAccount,
-    TransferHook,
-    TransferHookAccount,
+    Uninitialized = 0,
```
mistersimon, line 19: There were some typescript errors if I created the unused `ConfidentialTransferFee`... instructions. Would need to define them in the below functions to work, and i'm not sure the approriate ones. 

The explicit definition is the same ones that Typescript automatically generates, so there is no change in the transplied javacript code.

mistersimon (commented):
Hey @buffalojoec - I broke the metadata pointer extension into a separate PR. Thought it would be easier to review smaller changes.

token/js/src/extensions/extensionType.ts
```diff
@@ -8,29 +8,34 @@ import { DEFAULT_ACCOUNT_STATE_SIZE } from './defaultAccountState/index.js';
...
+    InterestBearingConfig = 10,
+    CpiGuard = 11,
+    PermanentDelegate = 12,
+    NonTransferableAccount = 13,
+    TransferHook = 14,
+    TransferHookAccount = 15,
+    // ConfidentialTransferFee = 16, // Not implemented yet
+    // WithdrawExcessLamports = 17, // Not implemented yet
+    MetadataPointer = 18,
+    // GroupPointer = 19, // Not implemented yet
```
mistersimon, line 38: TODO: Wrong numbers. Source of truth is here
https://github.com/solana-labs/solana-program-library/blob/master/token/program-2022/src/extension/mod.rs#L903

token/js/src/extensions/extensionType.ts
buffalojoec: Which TypeScript errors? I would much rather just add them to the enum and do nothing else, so that we don't have to track these numbers.

token/js/src/extensions/metadataPointer/instructions.ts
```diff
@@ -0,0 +1,95 @@
...
+import { TokenUnsupportedInstructionError } from '../../errors.js';
+import { TokenInstruction } from '../../instructions/types.js';
+import { addSigners } from '../../instructions/internal.js';
+
+export enum MetadataPointerInstruction {
+    Initialize = 0,
+    Update = 1,
+}
+
+export const initializeMetadataPointer = struct<{
```
buffalojoec, line 15: nit: Let's call this something with `*Encoder` or `*Data`, just so it's marked as a serialization const. `initializeMetadataPointerData`?

token/js/src/extensions/metadataPointer/instructions.ts
```diff
@@ -0,0 +1,95 @@
...
+    instruction: TokenInstruction.MetadataPointerExtension;
+    metadataPointerInstruction: number;
+    authority: PublicKey;
+    metadataAddress: PublicKey;
+}>([
+    u8('instruction'), // prettier let me wrap please
```
buffalojoec, line 21: At the top of the function block:

```javascript
// prettier-ignore
```

token/js/src/extensions/metadataPointer/state.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        // Only add keys if they are defined (ignored zero public keys)
+        if (!decoded.authority.equals(PublicKey.default)) {
+            state.authority = decoded.authority;
+        }
+        if (!decoded.metadataAddress.equals(PublicKey.default)) {
+            state.metadataAddress = decoded.metadataAddress;
+        }
+
+        return state;
```
buffalojoec, line 37: This is slick, but I think in this case it makes more sense to opt for `null`. That way, applications know they'll definitely have the fields defined, and `null` explicitly means the key is set to `None` (default).

buffalojoec (commented):
Nice! We are cruising right along with Token2022's JS support, thanks to you mate.

Just some small bits from my side, then we should be able to wrap soon.

token/js/src/extensions/metadataPointer/instructions.ts
mistersimon: Updated!

token/js/src/extensions/metadataPointer/state.ts
mistersimon: That makes sense. Made the change

token/js/src/extensions/metadataPointer/instructions.ts
mistersimon: Awesome, thats better. made the change

token/js/src/extensions/extensionType.ts
mistersimon: I think theres two parts to this discussion.


1. Want to clarify that the numbers are actually used and need to match the values in rust, so they do need to be tracked. For example the below function relies on the numbers to unpack the extension data. I prefer it explicitly defined, its a good enforcement that they actually have meaning, but will defer to you. If we remove the numbers, we just need to make sure the the order matches the rust implementation, so the automatic values given by TS matches the rust.
https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/extensions/extensionType.ts#L189

1. The reason that unimplemented extensions like `ConfidentialTransferFee` are commented out, is that if they are defined it needs to be added to this function https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/extensions/extensionType.ts#L130C33-L130C33. Theres no default case, so typescript will complain because the function can return  undefined. While its an easy fix to add a default cause to throw an error - i thought it would be better for whoever is implementing the extension to enable it and configure it correctly. Happy to add a default case if preferred.

mistersimon:
> Nice! We are cruising right along with Token2022's JS support, thanks to you mate.
> 
> Just some small bits from my side, then we should be able to wrap soon.

Thanks for the kind words @buffalojoec. 

Addressed most of your comments - wanted to clarify one thing on the enums before changing

token/js/src/extensions/extensionType.ts
buffalojoec: > 1. If we remove the numbers, we just need to make sure the the order matches the rust implementation

Yes, this is what we've been doing.

> 2\. While its an easy fix to add a default cause to throw an error

Yes do this! If the library doesn't support the extension yet, we just need to mark it as "unimplemented", which we can temporarily do with an error. Just so that it doesn't block progress on this PR.

mistersimon: Okay, good that we are on the same page. I've gone ahead and removed it. 

However for 2 - it wasn't as easy as adding default error. This test case requires every extension to be either a Mint or Account extension, and i'm not confident which ones they should be.
https://github.com/solana-labs/solana-program-library/blob/master/token/js/test/unit/index.test.ts#L244

I removed all the numbering and only explicitly given a number to `MetadataPointer` as a work around. Take a look and hopefully its an acceptable compromise.

token/js/src/extensions/extensionType.ts
```diff
@@ -31,6 +33,12 @@ export enum ExtensionType {
     NonTransferableAccount,
     TransferHook,
     TransferHookAccount,
+    // ConfidentialTransferFee, // Not implemented yet
+    // ConfidentialTransferFeeAmount, // Not implemented yet
+    MetadataPointer = 18, // Remove number once above extensions implemented
+    // TokenMetadata, // Not implemented yet
+    // GroupPointer, // Not implemented yet
+    // TokenGroup, //Not implemented yet
```
buffalojoec, line 41: Yeah this works.

Tiny nit: There really isn't any reason to include any after `MetadataPointer` here. And on the next PR we just add `TokenMetadata`, no others.

token/js/src/instructions/types.ts
```diff
@@ -37,4 +37,8 @@ export enum TokenInstruction {
     CpiGuardExtension = 34,
     InitializePermanentDelegate = 35,
     TransferHookExtension = 36,
+    // ConfidentialTransferFeeExtension = 37,
+    // WithdrawalExcessLamports = 38,
+    MetadataPointerExtension = 39,
+    // GroupPointerExtension = 40,
```
buffalojoec, line 43: Same tiny nit here on introducing trailing variants we don't need yet.

buffalojoec (commented):
This looks great! Two little things I noticed.

token/js/src/extensions/extensionType.ts
buffalojoec:
```suggestion
    // ConfidentialTransferFee, // Not implemented yet
    // ConfidentialTransferFeeAmount, // Not implemented yet
    MetadataPointer = 18, // Remove number once above extensions implemented
```

token/js/src/instructions/types.ts
buffalojoec:
```suggestion
    // ConfidentialTransferFeeExtension = 37,
    // WithdrawalExcessLamports = 38,
    MetadataPointerExtension = 39,
```

token/js/src/extensions/extensionType.ts
mistersimon: Makes sense. Changed!

token/js/src/instructions/types.ts
mistersimon: Makes sense. Changed!

mistersimon:
> This looks great! Two little things I noticed.

Made those changes. Let me know if there is anything else!

buffalojoec (approved):
Nice work!! JavaScript Token2022 support is shaping up nicely!

PR #5806, opened 2023-11-12
token 2022: Expose token-metadata state decoder in @solana/token-metata

mistersimon:
Token2022 doesn't using 8-byte discriminators, and can't use the unpack function current in @solana/spl-token-metadata which assumes 8-byte discriminators.

This PR exposes a function that just decodes to the data, so it can be used in Token2022.

Discussion where the need arose:
https://github.com/solana-labs/solana-program-library/pull/5667#discussion_r1390234922

### DIFF
--- a/token-metadata/js/package.json
+++ b/token-metadata/js/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@solana/spl-token-metadata",
     "description": "SPL Token Metadata Interface JS API",
-    "version": "0.1.0",
+    "version": "0.2.0",
     "author": "Solana Labs Maintainers <maintainers@solanalabs.com>",
     "repository": "https://github.com/solana-labs/solana-program-library",
     "license": "Apache-2.0",
--- a/token-metadata/js/package.json
+++ b/token-metadata/js/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@solana/spl-token-metadata",
     "description": "SPL Token Metadata Interface JS API",
-    "version": "0.2.0",
+    "version": "0.1.0",
     "author": "Solana Labs Maintainers <maintainers@solanalabs.com>",
     "repository": "https://github.com/solana-labs/solana-program-library",
     "license": "Apache-2.0",
--- a/token-metadata/js/src/state.ts
+++ b/token-metadata/js/src/state.ts
@@ -1,17 +1,17 @@
 import { PublicKey } from '@solana/web3.js';
-import {
-    getArrayDecoder,
-    getArrayEncoder,
-    getBytesDecoder,
-    getBytesEncoder,
-    getStructDecoder,
-    getStructEncoder,
-    getTupleDecoder,
-    getTupleEncoder,
-} from '@solana/codecs-data-structures';
-import { getStringDecoder, getStringEncoder } from '@solana/codecs-strings';
+import { getArrayCodec, getBytesCodec, getStructCodec, getTupleCodec } from '@solana/codecs-data-structures';
+import { getStringCodec } from '@solana/codecs-strings';
 
 export const TOKEN_METADATA_DISCRIMINATOR = Buffer.from([112, 132, 90, 90, 11, 88, 157, 87]);
+
+export const tokenMetadataCodec = getStructCodec([
+    ['updateAuthority', getBytesCodec({ size: 32 })],
+    ['mint', getBytesCodec({ size: 32 })],
+    ['name', getStringCodec()],
+    ['symbol', getStringCodec()],
+    ['uri', getStringCodec()],
+    ['additionalMetadata', getArrayCodec(getTupleCodec([getStringCodec(), getStringCodec()]))],
+]);
 
 export interface TokenMetadata {
     // The authority that can sign to update the metadata
@@ -40,18 +40,9 @@ function isNonePubkey(buffer: Uint8Array): boolean {
 
 // Pack TokenMetadata into byte slab
 export const pack = (meta: TokenMetadata): Uint8Array => {
-    const encoder = getStructEncoder([
-        ['updateAuthority', getBytesEncoder({ size: 32 })],
-        ['mint', getBytesEncoder({ size: 32 })],
-        ['name', getStringEncoder()],
-        ['symbol', getStringEncoder()],
-        ['uri', getStringEncoder()],
-        ['additionalMetadata', getArrayEncoder(getTupleEncoder([getStringEncoder(), getStringEncoder()]))],
-    ]);
-
     // If no updateAuthority given, set it to the None/Zero PublicKey for encoding
     const updateAuthority = meta.updateAuthority ?? PublicKey.default;
-    return encoder.encode({
+    return tokenMetadataCodec.encode({
         ...meta,
         updateAuthority: updateAuthority.toBuffer(),
         mint: meta.mint.toBuffer(),
@@ -60,16 +51,7 @@ export const pack = (meta: TokenMetadata): Uint8Array => {
 
 // unpack byte slab into TokenMetadata
 export function unpack(buffer: Buffer | Uint8Array): TokenMetadata {
-    const decoder = getStructDecoder([
-        ['updateAuthority', getBytesDecoder({ size: 32 })],
-        ['mint', getBytesDecoder({ size: 32 })],
-        ['name', getStringDecoder()],
-        ['symbol', getStringDecoder()],
-        ['uri', getStringDecoder()],
-        ['additionalMetadata', getArrayDecoder(getTupleDecoder([getStringDecoder(), getStringDecoder()]))],
-    ]);
-
-    const data = decoder.decode(buffer);
+    const data = tokenMetadataCodec.decode(buffer);
 
     return isNonePubkey(data[0].updateAuthority)
         ? {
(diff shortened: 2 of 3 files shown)


buffalojoec (changes requested):
Ok, so here's what we want to do (and what's in Rust):

In `@solana/spl-token-metadata`: Provide an `unpack` function that _just_ uses the deserialization to unpack a buffer into `TokenMetadata`; no discriminator _or_ length

In `@solana/spl-token`: Use that `unpack` in the token metadata extension to deserialize the token metadata from the extension. The discriminator and length bits are taken care of for you when you call `getExtensionData`, so we just need a function that calls that for the `TokenMetadata` extension, and then unpacks the returned buffer into `TokenMetadata`!

Here's an example:

https://github.com/solana-labs/solana-program-library/blob/f542e00a8905eaf3f55dea1c08ca4787138e1788/token/js/src/extensions/transferFee/state.ts#L92-L99

mistersimon:
> Ok, so here's what we want to do (and what's in Rust):
> 
> In `@solana/spl-token-metadata`: Provide an `unpack` function that _just_ uses the deserialization to unpack a buffer into `TokenMetadata`; no discriminator _or_ length


If i'm understanding correctly, the `unpack` function you describing is the function `decodeTokenMetadata` I added.
https://github.com/solana-labs/solana-program-library/pull/5806#pullrequestreview-1726249136

I left the behaviour of the `unpack` function the same to avoid a breaking change, but it uses `decodeTokenMetadata` under the hood. 

You suggesting we get rid of the `unpack` function as its implemented now and just leave `decodeTokenMetadata` but renamed to `unpack`?

buffalojoec:
> You suggesting we get rid of the `unpack` function as its implemented now and just leave `decodeTokenMetadata` but renamed to `unpack`?

Yes, we're basically just copying this, which is serializing/deserializing with `borsh`.

https://github.com/solana-labs/solana-program-library/blob/b1ba6493fb49a89fcbb0abcf0b5f05694a8bf7d1/token-metadata/interface/src/state.rs#L99-L109

We could even flop the `decoder` object from `@solana/codecs-data-structures` to use `codec` from and then we can roll `pack` and `unpack`! Although I don't know why we would need pack in JS at the moment.

Regardless, I think we should opt for `unpack` and `pack` naming conventions over `decode`, etc.

token-metadata/js/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@solana/spl-token-metadata",
     "description": "SPL Token Metadata Interface JS API",
-    "version": "0.1.0",
```
mistersimon, line 4: How do you all manage versions? Do you strictly follow semver? Cause then i think it should be 1.0.0

token-metadata/js/src/state.ts
```diff
@@ -1,9 +1,15 @@
...
+    getBytesDecoder,
+    getBytesEncoder,
+    getStructDecoder,
+    getStructEncoder,
+    getTupleDecoder,
+    getTupleEncoder,
+} from '@solana/codecs-data-structures';
+import { getStringDecoder, getStringEncoder } from '@solana/codecs-strings';
 
 export const TOKEN_METADATA_DISCRIMINATOR = Buffer.from([112, 132, 90, 90, 11, 88, 157, 87]);
```
mistersimon, line 14: I think it still makes sense to expose this even though it is not used, since i understand its a standardised value?

mistersimon:
> Regardless, I think we should opt for `unpack` and `pack` naming conventions over `decode`, etc.

Okay i changed the behaviour of unpack to what you described. 

Also added pack, as you say not sure there is a need but it was already there as a helper for the test cases - just had to move and expose it.

token-metadata/js/package.json
buffalojoec: I'll do the version bump separately!

token-metadata/js/src/state.ts
buffalojoec: Yeah sure.

token-metadata/js/test/state.test.ts
```diff
@@ -2,68 +2,31 @@ import { PublicKey } from '@solana/web3.js';
...
-            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0, 0, 0, 110, 97,
-            109, 101, 6, 0, 0, 0, 115, 121, 109, 98, 111, 108, 3, 0, 0, 0, 117, 114, 105, 0, 0, 0, 0,
-        ]);
-
-        const input = Buffer.concat([TOKEN_METADATA_DISCRIMINATOR, lengthBuffer(data), data]);
-
-        const meta = unpack(input);
-        expect(meta).to.deep.equal({
+    it('Can pack and unpack as rust implementation', () => {
+        const meta = {
```
buffalojoec, line 9: I just noticed this test doesn't use `updateAuthority` in the data structure. Can you add to the test a step where you add `updateAuthority` and test the bytes again? It doesn't have to be a separate test.

token-metadata/js/package.json
```diff
@@ -1,7 +1,7 @@
...
+    "version": "0.2.0",
```
buffalojoec, line 4: As I mentioned I'll bump this!

token-metadata/js/src/state.ts
```diff
@@ -32,12 +38,28 @@ function isNonePubkey(buffer: Uint8Array): boolean {
...
+// Pack TokenMetadata into byte slab
+export const pack = (meta: TokenMetadata): Uint8Array => {
+    const encoder = getStructEncoder([
+        ['updateAuthority', getBytesEncoder({ size: 32 })],
+        ['mint', getBytesEncoder({ size: 32 })],
+        ['name', getStringEncoder()],
+        ['symbol', getStringEncoder()],
+        ['uri', getStringEncoder()],
+        ['additionalMetadata', getArrayEncoder(getTupleEncoder([getStringEncoder(), getStringEncoder()]))],
+    ]);
```
buffalojoec, line 50: You can save the code duplication here by instead importing `getArrayCodec`, `getBytesCodec`, etc. and define this as `codec`. Then it will yield both `encode` and `decode` from the same constant object.

buffalojoec (commented):
Conceptually this is perfect, let's just tighten up a few bits.

token-metadata/js/package.json
mistersimon: Sure. Reverted back to 0.1.0 and i'll leave you to bump

token-metadata/js/src/state.ts
mistersimon: Oh i didn't know there was that functionality. That will be a lot cleaner. Changed it 👍

token-metadata/js/test/state.test.ts
mistersimon: I modified the last test which has the updateAuthority to check against the bytes. 

https://github.com/solana-labs/solana-program-library/pull/5806/commits/3839116439f00552319248f01d2af4206ae954ae#diff-3223c36003634dbc6a302f5e66b19f6aed12d8aa50040ebf6b79ec4fa3a45c5fR52

mistersimon:
> Conceptually this is perfect, let's just tighten up a few bits.

Thanks for the review - I've pick up those pieces. Let me know if there is anything else, otherwise will need to trouble you to merge + publish to npm. Thanks in advance!

token-metadata/js/src/state.ts
```diff
@@ -1,18 +1,18 @@
...
-    getTupleDecoder,
-    getTupleEncoder,
-} from '@solana/codecs-data-structures';
-import { getStringDecoder, getStringEncoder } from '@solana/codecs-strings';
+import { getArrayCodec, getBytesCodec, getStructCodec, getTupleCodec } from '@solana/codecs-data-structures';
+import { getStringCodec } from '@solana/codecs-strings';
 
 export const TOKEN_METADATA_DISCRIMINATOR = Buffer.from([112, 132, 90, 90, 11, 88, 157, 87]);
 
+export const tokenMetadataCodec = getStructCodec([
```
buffalojoec, line 7: Let's _not_ export this, since we are peppering in some custom logic for all-zero update authorities, etc. Someone might get confused if they try to use the codec in its raw form.

token-metadata/js/test/state.test.ts
```diff
@@ -50,23 +50,23 @@ describe('Token Metadata State', () => {
     });
 
     it('Can pack and unpack with mint and updateAuthority', () => {
-        const input = pack({
+        const meta = {
```
buffalojoec, line 53: We should also probably add a test with some additional metadata.

buffalojoec (commented):
Thanks! Couple small things then we're good to go.

token-metadata/js/src/state.ts
```diff
@@ -4,7 +4,8 @@ import { getStringCodec } from '@solana/codecs-strings';
 
 export const TOKEN_METADATA_DISCRIMINATOR = Buffer.from([112, 132, 90, 90, 11, 88, 157, 87]);
 
-export const tokenMetadataCodec = getStructCodec([
+// Don't export as you still need custom logic for all-zero update authorities
```
buffalojoec, line 7: Not necessary, lol.

buffalojoec (approved):
nice

buffalojoec (approved):

Issue #4550, opened 2023-06-15
token-js: Support metadata pointer extension


joncinque:
#### Problem

With #4549, token-2022 can also store a pointer to a metadata account, but this isn't exposed in the JS bindings.

#### Solution

Support the following instructions:

* initializing a mint with a metadata pointer: https://github.com/solana-labs/solana-program-library/blob/0a3686805d396f2ea8387758cf8ed7347c8fbc0a/token/program-2022/src/extension/metadata_pointer/instruction.rs#L77-L95
* updating the metadata pointer address: https://github.com/solana-labs/solana-program-library/blob/0a3686805d396f2ea8387758cf8ed7347c8fbc0a/token/program-2022/src/extension/metadata_pointer/instruction.rs#L98-L122

Support deserializing the extension data: https://github.com/solana-labs/solana-program-library/blob/0a3686805d396f2ea8387758cf8ed7347c8fbc0a/token/program-2022/src/extension/metadata_pointer/mod.rs#L17-L22

This is a good example to follow, based on a different extension: https://github.com/solana-labs/solana-program-library/pull/3833

pcellix:
@joncinque Can I work on it?

joncinque:
Go for it! I've updated the issue description to be more helpful

pcellix:
I'm on it! Thank you!

pcellix:
I installed the solana cli, when I try to execute cargo command I get the following message:

ubuntu@ubuntu-16gb-nbg1-1:~/solana-program-library$ cargo build-bpf
info: syncing channel updates for '1.69.0-x86_64-unknown-linux-gnu'
info: latest update on 2023-04-20, rust version 1.69.0 (84c898d65 2023-04-16)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
[2023-07-07T17:37:28.456416664Z ERROR cargo_build_sbf] Failed to execute /home/ubuntu/.cache/solana/v1.37/platform-tools/rust/bin/rustc: No such file or directory (os error 2)

ubuntu@ubuntu-16gb-nbg1-1:~/solana-program-library$ solana --version
[...]

PR #5813, opened 2023-11-13
[token js]: bump version

buffalojoec:
After adding the `MetadataPointer` extension to `@solana/spl-token` (#4550), we need to release a new version!

Here's the bump.


joncinque (approved):
Looks good!

PR #5814, opened 2023-11-13
token-cli: Make update confidential accept multiple args

joncinque:
#### Problem

As noted in #5804, it's currently not possible to update the approve policy and the auditor pubkey during `update-confidential-transfer-mint` because only one `update_field` is allowed at once.

#### Solution

Allow both to be set by setting `multiple(true)` on the `"update_field"` parameter.

cc @samkim-crypto


buffalojoec (approved):
😈

joncinque:
CI is hanging, so merging before it's done. Soon the token-cli won't be so heavy!

PR #5819, opened 2023-11-13
js: Add top-level package.json and turbo build

joncinque:
#### Problem

There's no top-level `package.json`, which means that dependabot doesn't know how to handle the sub-packages in the SPL repo.

#### Solution

Add a top-level package.json! This follows a very similar model to https://github.com/solana-labs/solana-web3.js/blob/master/package.json, and also adds turbo for building, which means that we don't need to declare the dependencies by hand in the CI scripts, which is *awesome*.

It's still a little messy, but it should be much better to work with overall. The next step might be to isolate all of the JS testing and have them all build their required programs themselves, rather than waiting on the `cargo-test-sbf` step to build it for them. That should greatly reduce the CI time.


joncinque:
Merging early to see how the dependabot run goes tomorrow morning, please feel free to still give feedback!

name-service/js/package.json
```diff
@@ -21,12 +21,12 @@
...
-    "lint": "yarn pretty && eslint --max-warnings 0 'src/*.ts'",
-    "lint:fix": "yarn pretty:fix && eslint 'src/*.ts' --fix",
+    "lint": "npm run pretty && eslint --max-warnings 0 'src/*.ts'",
+    "lint:fix": "npm run pretty:fix && eslint 'src/*.ts' --fix",
     "pretty": "prettier --check '{src/*.ts,test/*/*.ts}'",
     "pretty:fix": "prettier --write '{src/*.ts,test/*/*.ts}'",
-    "doc": "yarn typedoc src/index.ts",
-    "test": "yarn test:unit && yarn test:e2e",
+    "doc": "npm run typedoc src/index.ts",
+    "test": "npm run test:unit && npm run test:e2e",
```
2501babe, line 29: why the change to npm instead of either pnpm or leaving it as-is?

2501babe:
turbo is surprisingly straightforward

name-service/js/package.json
joncinque: pnpm fails if yarn is mentioned in the scripts, but works with npm. Since the other packages use npm when calling other scripts, I decided for consistency. We could certainly change it to pnpm though!

joncinque:
Yeah it's really neat. The default build doesn't figure out inter-package dependencies within your workspace, and just tries to build everything all at once.  Apparently turbo will parse everything out and build intelligently

PR #5821, opened 2023-11-14
token-cli: Use libtest-mimic to parallelize tests

joncinque:
#### Problem

The token-cli CI job takes a very long time (often around 25 minutes https://github.com/solana-labs/solana-program-library/actions/runs/6824131173/job/18559408031), and often times out https://github.com/solana-labs/solana-program-library/actions/runs/6813008074/job/18526633231

#### Solution

Rather than serializing all the tests, re-use two test validator instances to run all tests in parallel without overloading the system with ~40 concurrently running test validators.

A few other changes required with this:

* created an `async_trial!` macro to wrap the libtest-mimic `Trial` type into a normal runnable function, by copying what the `#[tokio::test]` macro is doing: https://docs.rs/tokio/latest/tokio/attr.test.html#using-current-thread-runtime
* all of the wrapped SOL tests are combined into one because we don't expose a nice way to wrap into a provided keypair
* the GC test is separated into its own test validator because it was cleaning up the aux accounts required by other tests

Registering tests still kind of stinks, so let me know if you have a better suggestion!


joncinque:
The CI job is now down to 18 minutes from 25 minutes, so that's some good progress! And hopefully it won't be so flaky

buffalojoec (approved):
speed

2501babe:
oh interesting, so the upshot here is instead of having toplevel test functions, we just have a main function, which creates the single test-validator and spawns threads for every "test," which is just a normal function wrapped by a new test harness lib? hmm i guess that means we cant use the `test_case` macros tho

2501babe:
hm i tried doing what i wanted with `lazy_static` and with `OnceLock` but immediately realized the core problem here is everything needs a _persistent_ ref to the validator, its only _calling_ it via a client that should be mutexed, but i dont have a good idea of how to express that. im not gonna worry about it for now... single-pool-cli tests are not a very pressing concern. i guess i could have a macro that accomplishes something like `test_case` tho by just writing several lines in the array passed to `libtest_mimic::run`

joncinque:
> hm i tried doing what i wanted with lazy_static and with OnceLock but immediately realized the core problem here is everything needs a persistent ref to the validator, its only calling it via a client that should be mutexed, but i dont have a good idea of how to express that

Haha that's funny, I went through the same exercise and got to exactly the same spot

> i guess i could have a macro that accomplishes something like test_case tho by just writing several lines in the array passed to libtest_mimic::run

Yeah that would be my solution too, unless we want to get fancy with macros

PR #5822, opened 2023-11-14
dependabot: Fix time format to add leading 0

joncinque:
#### Problem

The dependabot tab in the dependency graph is giving an error:

```
The property '#/updates/0/schedule/time' value "9:00" did not match the regex '^\d\d:\d\d$'
```

Because the time is specified "9:00".

#### Solution

Update it to "09:00"


joncinque:
No need for CI, merging straight away

PR #5835, opened 2023-11-14
untested, commiting to discuss with group

staccDOTsol:
There's a few things that we need to fixup, mainly calling into the token program with the PDA, doing more checks on the input data, and some tests, but submitted as a work in progress


buffalojoec:
@staccDOTsol Did you want to discuss this change still?

PR #5836, opened 2023-11-14
token 2022: add support for _reading_ repeating fixed-length extensions

buffalojoec:
In order to support Token2022 extensions that can repeat, we need to add support
to read such extensions!

Almost every read function for Token2022 extensions is expecting one single
entry, so it will return the first encountered.

Here. we're introducing an adjacent API for reading repeating extensions,
specifically fixed-length extensions.

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -173,6 +173,35 @@ fn get_extension_indices<V: Extension>(
     Err(ProgramError::InvalidAccountData)
 }
 
+fn get_all_extension_indices<V: Extension>(
+    tlv_data: &[u8],
+) -> Result<Vec<TlvIndices>, ProgramError> {
+    let mut indices_list = vec![];
+    let mut start_index = 0;
+    let v_account_type = V::TYPE.get_account_type();
+    while start_index < tlv_data.len() {
+        let tlv_indices = get_tlv_indices(start_index);
+        if tlv_data.len() < tlv_indices.value_start {
+            return Err(ProgramError::InvalidAccountData);
+        }
+        let extension_type =
+            ExtensionType::try_from(&tlv_data[tlv_indices.type_start..tlv_indices.length_start])?;
+        let account_type = extension_type.get_account_type();
+
+        let length =
+            pod_from_bytes::<Length>(&tlv_data[tlv_indices.length_start..tlv_indices.value_start])?;
+        let value_end_index = tlv_indices.value_start.saturating_add(usize::from(*length));
+        start_index = value_end_index;
+
+        if extension_type == V::TYPE {
+            indices_list.push(tlv_indices);
+        } else if v_account_type != account_type && extension_type != ExtensionType::Uninitialized {
+            return Err(TokenError::ExtensionTypeMismatch.into());
+        }
+    }
+    Ok(indices_list)
+}
+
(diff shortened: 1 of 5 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -397,11 +450,46 @@ pub trait BaseStateWithExtensions<S: BaseState> {
...
+    }
+
     /// Unpack a portion of the TLV data as the desired type
     fn get_extension<V: Extension + Pod>(&self) -> Result<&V, ProgramError> {
         pod_from_bytes::<V>(self.get_extension_bytes::<V>()?)
     }
 
+    /// Unpack a portion of the TLV data as the desired type, where repetitions
+    /// are allowed
+    fn get_repeating_extension<V: Extension + Pod>(
```
joncinque, line 478: nit: `get_extension_with_repetition`

token/program-2022/src/extension/mod.rs
```diff
@@ -397,11 +450,46 @@ pub trait BaseStateWithExtensions<S: BaseState> {
...
+                .ok_or::<ProgramError>(ProgramError::InvalidAccountData)
+                .unwrap()
+        })
+    }
+
+    /// Fetch all bytes for each entry of a particular TLV entry, where
+    /// repetitions are allowed
+    fn get_all_extension_bytes<V: Extension>(&self) -> Result<Vec<&[u8]>, ProgramError> {
+        get_all_extension_bytes::<S, V>(self.get_tlv_data())
+    }
```
joncinque, line 469: Are these needed? `get_extension_bytes` is only used for emitting a slice of bytes for token metadata, and we won't need to do that with the group interface, so I'd prefer to remove these functions if they're not needed

token/program-2022/src/extension/mod.rs
```diff
@@ -397,11 +450,46 @@ pub trait BaseStateWithExtensions<S: BaseState> {
         get_extension_bytes::<S, V>(self.get_tlv_data())
     }
 
+    /// Fetch the bytes for a TLV entry, where repetitions are allowed
+    fn get_repeating_extension_bytes<V: Extension>(
+        &self,
+        repetition: usize,
+    ) -> Result<&[u8], ProgramError> {
+        get_all_extension_bytes::<S, V>(self.get_tlv_data()).map(|x| {
+            *x.get(repetition.saturating_sub(1))
```
joncinque, line 459: The TLV library is 0-indexed, so let's keep that consistent here

token/program-2022/src/extension/mod.rs
```diff
@@ -1435,6 +1571,46 @@ mod test {
...
+        let all_extensions = state.get_all_extensions::<MintCloseAuthority>().unwrap();
+        assert_eq!(all_extensions.len(), 3);
+        assert_eq!(
+            state.get_extension::<TransferFeeConfig>(),
+            Err(ProgramError::InvalidAccountData)
+        );
+        assert_eq!(
+            StateWithExtensions::<Account>::unpack(MINT_WITH_DUPLICATED_EXTENSION_AND_ONE_EXTRA),
+            Err(ProgramError::InvalidAccountData)
+        );
```
joncinque, line 1611: Can you also check that you can fetch the extra extension, just to be safe?

joncinque (commented):
Looks good overall! Just some little comments that I hope make it better

PR #5837, opened 2023-11-14
token 2022: add support for _writing_ repeating fixed-length extensions

buffalojoec:
Similar to the PR before this one, we need to expand the API for Token2022's
extensions to support repeating extension entries.

Previously we addresses _reading_ these extensions, now we shall address
_writing_.

This PR introduces an adjacent API for writing repeatable fixed-length
extensions to Token2022 mints.

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -413,6 +413,37 @@ fn get_extension_bytes_mut<S: BaseState, V: Extension>(
     Ok(&mut tlv_data[value_start..value_end])
 }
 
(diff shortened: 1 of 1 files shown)


joncinque (commented):
I just noticed that `try_get_total_tlv_len` will also break because we dedupe extensions there. To change that, we'll have to go through all of the cases where we call `try_calculate_account_len` *without* deduping beforehand.

I'm starting to lean towards not having multiple extensions of one type. We made that assumption pretty strongly , and I worry that we'll break something. What do you think?

PR #5841, opened 2023-11-14
ci: Refactor JS tests to avoid running so often

joncinque:
#### Problem

The JS CI that kicks off for any change to pnpm-lock.yaml triggers essentially *everything* to run, when very often we just need to make sure that the packages build.

#### Solution

Refactor the CI jobs a bit:

* Make JS test scripts build the programs they rely on with `build:program`
* Avoid running main program-level workflows if there's just a JS change
* Add new workflow if there's only a JS change

While I was at it, I killed off the fee-specific token-swap JS tests, since they were more annoying than anything else to manage.

Note: Marking this a draft since it might take some time to get things right in CI.


joncinque:
Merging this early so that it can be part of the dependabot runs tomorrow and avoid using all of our runners at once. I'll happily integrate any feedback afterwards!

.github/workflows/pull-request-js.yml
```diff
@@ -0,0 +1,54 @@
...
+      - 'libraries/type-length-value/js/**'
+      - 'memo/js/**'
+      - 'single-pool/js/**'
+      - 'stake-pool/js/**'
+      - 'token/js/**'
+      - 'token-lending/js/**'
+      - 'token-metadata/js/**'
+      - 'token-swap/js/**'
+      - 'pnpm-lock.yaml'
+      - '.github/workflows/pull-request-js.yml'
```
2501babe, line 29: does this mean that any change to any js lib results in all of them running their tests? seems undesireable

joncinque: Currently, yes. Ideally we'd want to add a check in the matrix to only run that job if there's a change to the subdir OR the lockfile, but I haven't been able to find how GH actions exposes that at the step-level. Let me know if you find something!

Thankfully, these take max 10 minutes to run, so it seemed like an ok trade-off for now.

PR #5844, opened 2023-11-15
CI: Update GitHub actions to modern versions

joncinque:
#### Problem

Our GitHub Actions runs produce loads of warnings because we're using old versions of actions or deprecated actions.

#### Solution

Update them all! Everything was done in separate commits, but here's the list:

* actions/checkout -> v4
* actions/cache -> v3
* actions/upload-artifact -> v3
* actions/download-artifact -> v3
* actions/setup-node -> v4
* actions-rs/toolchain -> [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) which seems to be the next version used by everyone, maintained by a Rust OG
* actions-rs/cargo -> just call `cargo`
* actions-rs/install -> just call `cargo install`

### DIFF
--- a/.github/workflows/fuzz-nightly.yml
+++ b/.github/workflows/fuzz-nightly.yml
@@ -21,11 +21,9 @@ jobs:
           source ci/solana-version.sh
           echo "SOLANA_VERSION=$solana_version" >> $GITHUB_ENV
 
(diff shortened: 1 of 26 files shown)


buffalojoec (approved):
shiny

PR #5845, opened 2023-11-15
ci: Update actions/setup-python to v4

joncinque:
#### Problem

As a final follow-up to #5844, actions/setup-python is also outdated.

#### Solution

To keep this PR slim, just update the stake pool job to use actions/setup-python v4


buffalojoec (approved):
python

Issue #5092, opened 2023-08-23
[Compressed NFTs] Transfer with zero length proof


raulriera:
**Description**

From the [docs](https://docs.solana.com/developing/guides/compressed-nfts#prepare-the-asset-proof) the recommendation is to prune it from the proofs but that results in an empty array and the transfer will fail.

> In the TypeScript code example above, we are first taking a slice of our "full proof", starting at the beginning of the array, and ensuring we only have proof.length - canopyDepth number of proof values. This will remove the portion of the proof that is already stored on-chain in the tree's canopy.

There are situations where the canopyDepth and full proof are the same resulting in a zero length proof that will fail a transfer with `Proof is empty`. Example NFT would be any of the Tensorians NFTs https://www.tensor.trade/trade/tensorians

ngundotra:
@raulriera I haven't heard of any issues with zero length proofs, but if you can provide a code sample I can take a look.

I'll add a test to the test suite & then close this issue.

cc @richwardwu

PR #5847, opened 2023-11-15
[account compression] Clean up tests

ngundotra:
It is not eminently clear that zero-length instructions work, so add a test case for this in e2e tests.

Addresses concern in #5092


ngundotra:
@jnwng @buffalojoec @joncinque Before adding some features to SPL account compression, do you guys have any preference on how the SPL Account Compression test suite is written?

buffalojoec:
> @jnwng @buffalojoec @joncinque Before adding some features to SPL account compression (in follow up PRs), do you guys have any preference on how the SPL Account Compression test suite is written?

From what I can tell all of the JS libraries do tests their own way, but we are migrating all of them to use `pnpm` over `npm` or `yarn`, so if you wanted to configure your `package.json` and test suite for that, it would be a help. Shouldn't be much of a change!

buffalojoec:
@joncinque might not be the place to discuss but this did remind me of Solana Bankrun, which we can use to test our JS libs.
https://kevinheavey.github.io/solana-bankrun/

ngundotra:
Okay, I actually split up the tests into a few different spots, should run a bit faster on local now since they can be parallelized by jest. Also way easier to read now.

Should be evident from the `tests/accounts/canopy.test.ts` that zero length transfer works.

joncinque:
> Do you guys have any preference on how the SPL Account Compression test suite is written?

Jest is normally the recommended test suite over mocha. And definitely use bankrun rather than spinning up a test validator to test against, if you can!

PR #5862, opened 2023-11-20
token-cli: Support creating mint with fee and conf transfer

joncinque:
#### Problem

As reported at https://solana.stackexchange.com/questions/8106/is-it-possible-yet-to-use-mint-tokens-supporting-both-transfer-fees-and-confiden, the token-cli currently fails when creating a mint with a transfer fee and confidential transfers, even though we have tests that work for creating the mint.

#### Solution

It's a bit tricky since we're not receiving the ElGamal keypair, but this little bandaid solution uses the default signer's derived ElGamal keypair. This will unblock people who want to create a mint with both.

If you think this is a bad idea, however, I'll relent and we can just say it's unsupported until 1.18.

### DIFF
--- a/token/cli/src/command.rs
+++ b/token/cli/src/command.rs
@@ -255,6 +255,21 @@ async fn command_create_token(
             auto_approve_new_accounts: auto_approve,
             auditor_elgamal_pubkey: None,
         });
+        if transfer_fee.is_some() {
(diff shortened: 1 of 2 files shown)


samkim-crypto (approved):
Sorry for the delay on this one. I agree that it is a good idea to enable configuring transfer with fee. The tests look fine to me!

PR #5863, opened 2023-11-20
repo: Update to 1.17.6

joncinque:
#### Problem

There are some fixes past 1.17.2 that we want to integrate in the repo, namely the new compute unit usage for the confidential transfer instructions.

#### Solution

Update the repo to 1.17.6, and no longer force-disable a feature in the token-cli tests.


samkim-crypto (approved):
Looks good to me!

PR #5876, opened 2023-11-21
docs: Add CPI Guard extension info

### LINKED ISSUE #3771: token-2022: cpiguard docs/js/cli

joncinque:
#### Problem

CPI Guard has been around for some time, but there aren't any docs about it.

#### Solution

Using the great proposal at #3694 for inspiration (and copy-pasta), add the docs!

Fixes #3771

### DIFF
--- a/docs/src/token-2022/extensions.mdx
+++ b/docs/src/token-2022/extensions.mdx
@@ -1308,6 +1308,146 @@ import {
   </TabItem>
 </Tabs>
 
+### CPI Guard
+
+CPI Guard is an extension that prohibits certain actions inside cross-program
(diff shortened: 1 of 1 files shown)


2501babe (approved):

2501babe:
thank you so much for taking care of this! haha every time i tried i got writers block... should have just used what id already written!

PR #5883, opened 2023-11-22
docs: Add status of the token program

joncinque:
#### Problem

As evidenced by the comment at https://github.com/solana-labs/solana-program-library/issues/2742#issuecomment-1822038083, people don't know the status of the token program.

#### Solution

Add a couple of lines in the docs to essentially say that the token program isn't changing.


t-nelson (approved):
:shipit:

PR #5886, opened 2023-11-22
ci: Run docs publish when CI job is updated too

joncinque:
#### Problem

Even though the CI docs workflow was just updated, the publish step didn't run because it needs a change to `docs/**`.

#### Solution

Also have the job run when the workflow file is changed.


joncinque:
Merging early since the job only runs on merges to master

Issue #5662, opened 2023-10-25
transfer-hook: Add `update` instruction


joncinque:
#### Problem

A program implementing the transfer-hook interface only absolutely requires one instruction, the `execute` hook which is called by token-2022.

However, the interface does also include an initialization function, which is needed in some form to write data to the extra-account-metas account. An implementer may choose to do something else for that instruction.

Similarly, there's no way to `update` the extra-account-metas account with just the interface.

#### Solution

This behavior is fine as is, but it would be nicer to have a recommended-but-not-required `update` instruction for devs looking to implement the transfer hook.

cc @buffalojoec

buffalojoec:
> would it be possible for me to work on this issue? I'm a bit of a newcomer to this domain but wouldn't mind tackling it. 

Of course! I’d start with the interface change, then in separate PRs we can address the different clients.

Feel free to request my review.

tonton-sol:
> then in separate PRs we can address the different clients.

Which clients specifically would need to be updated?

tonton-sol:
I am also trying to implement this. I have added the new instruction to the interface and example program. I have also added a subcommand to the CLI. When testing with the CLI, the RPC simulation fails with custom program error: 0x47af3bc1.

How can I decode this? Also is there an easy way to show logs of the failed RPC simulation?

cjauton:
I am also trying to implement this. I have added the new instruction to the interface and example program. I have also added a subcommand to the CLI. When testing with the CLI, the RPC simulation fails with custom program error: 0x47af3bc1.

How can I decode this? Also is there an easy way to show logs of the failed RPC simulation?

tonton-sol:
I am also trying to implement this. I have added the new instruction to the interface and example program. I have also added a subcommand to the CLI. When testing with the CLI, the RPC simulation fails with custom program error: 0x47af3bc1.

How can I decode this? Also is there an easy way to show logs of the failed RPC simulation?

tonton-sol:
I figured out the error, it is `TlvError::TypeAlreadyExists`. 

I thought I could simply just create a new `TransferHookInstruction::UpdateExtraAccountMetaList` in the interface similar to `TransferHookInstruction::InitializeExtraAccountMetaList`, then change the example program handle this instruction. 

The error arises when:
```
ExtraAccountMetaList::init::<ExecuteInstruction>(&mut data, extra_account_metas)?;
``` 
is called in the new update instruction in the example program

[...]

PR #5894, opened 2023-11-26
Token 22: Add functionality to update extra_account_metas after initializing.

tonton-sol:
This is my first attempt at adding an update functionality, issue [#5662](https://github.com/solana-labs/solana-program-library/issues/5662). I added an update function for tlv-account-resolution::ExtraAccountMetaList that will realloc the original data then rewrite the account metas. I also added the correct instruction handling to the cli, interface, and program. 

The account size is increased if needed by the program. As far as I am aware there is no way to decrease the size of an account, only make it larger.

This is my first open-source contribution, so please let me know if I did something improper dev-ops wise.

Thank you!


tonton-sol:
Fixed two issues.

libraries/tlv-account-resolution/src/state.rs
```diff
@@ -182,6 +182,22 @@ impl ExtraAccountMetaList {
...
+        data: &mut [u8],
+        extra_account_metas: &[ExtraAccountMeta],
+    ) -> Result<(), ProgramError> {
+        let mut state = TlvStateMut::unpack(data).unwrap();
+        let tlv_size = PodSlice::<ExtraAccountMeta>::size_of(extra_account_metas.len())?;
+        let bytes = state.realloc_first::<T>(tlv_size)?;
+        let mut validation_data = PodSliceMut::init(bytes)?;
+        for meta in extra_account_metas {
+            validation_data.push(*meta)?;
+        }
```
buffalojoec, line 197: I _think_ this will work, but the unit tests will help here.

token/transfer-hook/cli/src/main.rs
```diff
@@ -127,6 +128,71 @@ async fn process_create_extra_account_metas(
...
+        .map_err(|err| format!("error: unable to fetch rent-exemption: {err}"))?;
+    let extra_account_metas_account = rpc_client.get_account(&extra_account_metas_address).await;
+    if extra_account_metas_account.is_err() {
+        return Err(format!(
+            "error: extra account metas for mint {token} and program {program_id} does not exist"
+        )
+        .into());
+    }
+    let current_lamports = extra_account_metas_account.map(|a| a.lamports).unwrap_or(0);
+    let transfer_lamports = required_lamports.saturating_sub(current_lamports);
```
buffalojoec, line 159: There's a ton of repeat code here compared with the `create` function above it. Can we extract some of this into a helper?

buffalojoec (commented):
Hey thanks for contributing! This is looking really good.

It would be great if you added some tests for the new `update` function and processor!

buffalojoec (commented):
Thanks for the continued awesome work! I left a few more comments, mostly little stuff, but this is moving along really nicely!

token/transfer-hook/cli/src/main.rs
```diff
@@ -127,6 +126,84 @@ async fn process_create_extra_account_metas(
...
+    }
+    transaction
+        .try_sign(&signers, blockhash)
+        .map_err(|err| format!("error: failed to sign transaction: {err}"))?;
+
+    rpc_client
+        .send_and_confirm_transaction_with_spinner(&transaction)
+        .await
+        .map_err(|err| format!("error: send transaction: {err}").into())
+}
```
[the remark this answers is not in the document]
tonton-sol: Agreed!

buffalojoec (commented):
Nice!! Just a few more small items and we should be cruising 💪🏼 💪🏼

libraries/tlv-account-resolution/src/state.rs
```diff
@@ -1292,6 +1308,64 @@ mod tests {
...
+            ExtraAccountMeta::new_with_pubkey(&Pubkey::new_unique(), false, false).unwrap(),
+            ExtraAccountMeta::new_with_pubkey(&Pubkey::new_unique(), false, true).unwrap(),
+        ];
+        initialize_update_and_assert_metas(&initial_metas, &updated_metas_2).await;
+
+        // Create updated and smaller list of metas
+        let updated_metas_3 =
+            [ExtraAccountMeta::new_with_pubkey(&Pubkey::new_unique(), true, true).unwrap()];
+        initialize_update_and_assert_metas(&initial_metas, &updated_metas_3).await;
+    }
```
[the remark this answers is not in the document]
buffalojoec: I would say including one or two PDAs maybe? But other than that I think just ensuring we can update more than once is fine for this test specifically.

buffalojoec:
@tonton-sol just re-request my review or tag me whenever you're ready for another look.

token/transfer-hook/example/src/processor.rs
```diff
@@ -134,6 +134,63 @@ pub fn process_initialize_extra_account_meta_list(
...
+    if expected_validation_address != *extra_account_metas_info.key {
+        return Err(ProgramError::InvalidSeeds);
+    }
+
+    let length = extra_account_metas.len();
+    let account_size = ExtraAccountMetaList::size_of(length)?;
+    let original_account_size = extra_account_metas_info.data_len();
+
+    // If the new extra_account_metas length is different, resize the account and update
+    if account_size >= original_account_size {
```
buffalojoec, line 178: We should probably add a check in here to ensure the account has previously been initialized as a validation account.

I'd say verify the owner is the current program and the account's data is at least enough space for a `T: Discriminator` and `L: Length` (TLV min size).

Maybe `ProgramErrror:: UninitializedAccount`?

tonton-sol:
> @tonton-sol just re-request my review or tag me whenever you're ready for another look.

I think its ready 👍

token/transfer-hook/example/tests/functional.rs
```diff
@@ -747,3 +750,625 @@ async fn fail_without_transferring_flag() {
...
+                Seed::InstructionData {
+                    index: 8,  // After instruction discriminator
+                    length: 8, // `u64` (amount)
+                },
+                Seed::AccountKey { index: 2 },
+            ],
+            false,
+            true,
+        )
+        .unwrap(),
```
buffalojoec, line 1105: It's still a valid test, but you're not really changing a whole lot in these updates. Let's up the stakes a little bit, eh?

buffalojoec (commented):
A few tiny things and we're ready to ship. Massive thanks again!

libraries/tlv-account-resolution/src/state.rs
```diff
@@ -1292,6 +1308,79 @@ mod tests {
...
+            ExtraAccountMeta::new_with_pubkey(&Pubkey::new_unique(), true, true).unwrap(),
+            ExtraAccountMeta::new_with_pubkey(&Pubkey::new_unique(), false, false).unwrap(),
+            ExtraAccountMeta::new_with_pubkey(&Pubkey::new_unique(), false, true).unwrap(),
+        ];
+        update_and_assert_metas(&mut buffer, &updated_metas_2).await;
+
+        // Create updated and smaller list of metas
+        let updated_metas_3 =
+            [ExtraAccountMeta::new_with_pubkey(&Pubkey::new_unique(), true, true).unwrap()];
+        update_and_assert_metas(&mut buffer, &updated_metas_3).await;
```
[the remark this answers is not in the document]
buffalojoec: No that's not expected. You should be able to define a PDA that way. What about something like this?

https://github.com/solana-labs/solana-program-library/blob/5da184aed56cda46b0006ebf3ea5d617ef0501a8/libraries/tlv-account-resolution/src/state.rs#L552-L567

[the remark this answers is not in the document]
buffalojoec: > Shouldn't it be possible to convert ExtraAccountMeta to AccountMeta?

Only if it's a fixed address! If it's a PDA you can't get the address without account resolution.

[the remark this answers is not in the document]
buffalojoec: In a few places in the tests you can see we can only properly convert to `AccountMeta` by running the resolution and grabbing the resulting `AccountMeta`s from the resolved instruction.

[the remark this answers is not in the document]
tonton-sol: > No that's not expected. You should be able to define a PDA that way. What about something like this?
> 
> https://github.com/solana-labs/solana-program-library/blob/5da184aed56cda46b0006ebf3ea5d617ef0501a8/libraries/tlv-account-resolution/src/state.rs#L552-L567

I can define the PDA ExtraAccountMeta, but I can't convert to AccountMeta using try_from(). But I understand why its not possible. I will probably just build an AccountMetas for each case to assert with instructions.accounts.

token/transfer-hook/example/src/processor.rs
tonton-sol: For the space comparison

token/transfer-hook/example/src/processor.rs
```diff
@@ -170,12 +170,17 @@ pub fn process_update_extra_account_meta_list(
...
     }
 
-    let length = extra_account_metas.len();
-    let account_size = ExtraAccountMetaList::size_of(length)?;
+    // Check if the extra metas have been initialized
+    let min_account_size = ExtraAccountMetaList::size_of(0)?;
     let original_account_size = extra_account_metas_info.data_len();
+    if program_id != extra_account_metas_info.owner || original_account_size <= min_account_size {
+        return Err(ProgramError::UninitializedAccount);
+    }
```
buffalojoec: Good check, but _technically_ an initialized list with zero entries is valid.

```suggestion
    // Check if the extra metas have been initialized
    let min_account_size = ExtraAccountMetaList::size_of(0)?;
    let original_account_size = extra_account_metas_info.data_len();
    if program_id != extra_account_metas_info.owner || original_account_size < min_account_size {
        return Err(ProgramError::UninitializedAccount);
    }
```

libraries/tlv-account-resolution/src/state.rs
```diff
@@ -1292,6 +1308,124 @@ mod tests {
...
+        // resize buffer if necessary
+        let account_size = ExtraAccountMetaList::size_of(updated_metas.len()).unwrap();
+        if account_size > buffer.len() {
+            buffer.resize(account_size, 0);
+        }
+
+        // update
+        ExtraAccountMetaList::update::<TestInstruction>(buffer, updated_metas).unwrap();
+
+        // retreive metas and assert
```
buffalojoec, line 1326: nit: "retrieve"

buffalojoec (commented):
Just have to update the check and we can punch this in 💪🏼

Again, much appreciated continued effort!

buffalojoec (approved):
Awesome job!! Nice contribution to SPL 🤙🏼

PR #5895, opened 2023-11-27
feat: Add compatibility for publicKeys

Gitmazter:
Ran into an issue where I was passing web3.publicKeys to createMintToInstruction with 3 multiSigners but received an anonymous.toString() is undefined error while trying to use serializeMessage() for offline signing. 

The error was that the docs allowed a (signer | publicKey) [ ] but the function only works with a signer[ ], my fix was to move my web3.publicKeys into an object. {publicKey: myKey} which allowed me to serialize the instruction. 

What I assume happened was that the logic 
`pubkey: signer instanceof PublicKey ? signer : signer.publicKey`
did not work as intended and saw my publicKeys as signers and thus tried publicKey.publicKey.toString() 

I believe the changes to the logic in this file will resolve the issue for anyone encountering it in the future.

I'm not entirely familiar with the SPL library yet so I'd be glad for suggestions for how to improve my PR or just have a discussion if anything else might have caused it

### DIFF
--- a/token/js/src/instructions/internal.ts
+++ b/token/js/src/instructions/internal.ts
@@ -13,7 +13,7 @@ export function addSigners(
         for (const signer of multiSigners) {
             if (signer instanceof PublicKey || signer instanceof Keypair) {
                 keys.push({
-                    pubkey: signer instanceof PublicKey ? signer : signer.publicKey,
+                    pubkey: 'publicKey' in signer ? signer.publicKey : signer,
                     isSigner: true,
                     isWritable: false,
                 });
--- a/token/js/src/instructions/internal.ts
+++ b/token/js/src/instructions/internal.ts
@@ -1,5 +1,5 @@
 import type { AccountMeta, Signer } from '@solana/web3.js';
-import { PublicKey } from '@solana/web3.js';
+import { PublicKey, Keypair } from '@solana/web3.js';
 
 /** @internal */
 export function addSigners(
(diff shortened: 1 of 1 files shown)


token/js/src/instructions/internal.ts
```diff
@@ -9,12 +9,28 @@ export function addSigners(
...
     if (multiSigners.length) {
         keys.push({ pubkey: ownerOrAuthority, isSigner: false, isWritable: false });
+        
         for (const signer of multiSigners) {
-            keys.push({
-                pubkey: signer instanceof PublicKey ? signer : signer.publicKey,
-                isSigner: true,
-                isWritable: false,
-            });
+            if (signer instanceof PublicKey || signer instanceof Keypair) {
```
buffalojoec, line 14: All `Keypair` are `Signer` but not all `Signer` are `Keypair`.

token/js/src/instructions/internal.ts
```diff
@@ -9,12 +9,28 @@ export function addSigners(
...
+                keys.push({
+                    pubkey: signer instanceof PublicKey ? signer : signer.publicKey,
```
buffalojoec, line 16: What if we just did this instead?

```suggestion
                    pubkey: 'publicKey' in signer ? signer.publicKey : signer,
```

buffalojoec (commented):
Thanks for reporting this and submitting a fix!

I wonder if the issue had something to do with a dual-package error. TL/DR, if two different versions of `@solana/web3.js` are somehow being added to the dependency tree, then `instanceof` is going to fail. Since the conditional defaults to a `signer` anytime `instanceof PublicKey` fails, that seems to make sense to me.

It looks like your fix will work, but I suggested a way to write this without using `instanceof` at all.

Gitmazter:
Will add comment this here as well outside of the oudated conversation: 

This is a better/simpler solution, will commit it! Thank you for paying attention to my PR :) Still new to cooperating over Github so it's definetly easing my social anxiety lol.

After reflecting on your comments the two versions of web3.js argument sounds the most likely to me, i think it applies to wallet adapters as well since I'm using the phantom adapter as part of my project which isnt named PublicKey but is named after a 2 byte hash e.g. ve{}|ey{} etc. Although I did try to convert these before submitting my pr but the problem remained.

buffalojoec:
> This is a better/simpler solution, will commit it! Thank you for paying attention to my PR :) Still new to cooperating over Github so it's definetly easing my social anxiety lol.

You are welcome to contribute here, mate.
 
> After reflecting on your comments the two versions of web3.js argument sounds the most likely to me, i think it applies to wallet adapters as well since I'm using the phantom adapter as part of my project which isnt named PublicKey but is named after a 2 byte hash e.g. ve{}|ey{} etc. Although I did try to convert these before submitting my pr but the problem remained.

Is this issue related to the fix you've PR'ed here?

PR #5900, opened 2023-11-27
token-2022: [OS-SPL-ADV-02] Fix instruction creator

joncinque:
#### Problem

The instruction creators for some of the confidential-transfer-with-fee instructions add the multisig signers as non-signers. An instruction created in this way will not work, since all of the multisig signers must sign the transaction.

#### Solution

Have the multisig signers declared as signing.


samkim-crypto (approved):
Thanks, looks good to me!

PR #5901, opened 2023-11-27
token-2022: [OS-SPL-ADV-03] Check account mint in approve

joncinque:
#### Problem

During confidential transfer approval, there's no check to make sure that the token account corresponds to the provided mint. This makes it possible to approve accounts on other mints.

#### Solution

Add a check that the provided token account belongs to the provided mint.

### DIFF
--- a/token/program-2022-test/tests/confidential_transfer.rs
+++ b/token/program-2022-test/tests/confidential_transfer.rs
@@ -3,7 +3,7 @@
 mod program_test;
 use {
     program_test::{TestContext, TokenContext},
-    solana_program_test::{processor, tokio, ProgramTest},
+    solana_program_test::tokio,
     solana_sdk::{
         instruction::InstructionError,
         pubkey::Pubkey,
@@ -39,7 +39,7 @@ use {
         proof_generation::transfer_with_fee_split_proof_data,
         token::{ExtensionInitializationParams, Token, TokenError as TokenClientError},
     },
-    std::{convert::TryInto, mem::size_of, sync::Arc},
+    std::{convert::TryInto, mem::size_of},
 };
 
 #[cfg(feature = "zk-ops")]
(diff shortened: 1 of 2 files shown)


token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -306,6 +306,75 @@ async fn confidential_transfer_configure_token_account() {
     );
 }
 
+#[tokio::test]
+async fn fail_approving_account_on_wrong_mint() {
```
samkim-crypto, line 310:
```suggestion
async fn confidential_transfer_fail_approving_account_on_wrong_mint() {
```
Just for consistency with the rest of the tests, should we append the `confidential_transfer_` prefix on the test names? This also makes filtering just the confidential transfer tests a little bit easier.

samkim-crypto (approved):
Looks good to me!

One small question: was there a reason why you didn't just use something like 
```
let mut context_a = TestContext::new().await;
context_a
    .init_token_with_mint(vec![
        ExtensionInitializationParams::ConfidentialTransferMint {
            authority: Some(authority.pubkey()),
            auto_approve_new_accounts,
            auditor_elgamal_pubkey: Some(auditor_elgamal_pubkey),
        },
    ])
    .await
    .unwrap();
```
like in the rest of the test functions?

token/program-2022-test/tests/confidential_transfer.rs
joncinque: Yeah sure that works for me! I typically avoid prefixing all the tests because the filename will do that prefixing for you, so I'll run `cargo test-sbf --test confidential_transfer -- fail_approve`, but I'll keep things consistent here

joncinque:
> One small question: was there a reason why you didn't just use something like:

Laziness entirely! I'll fix it up

PR #5910, opened 2023-11-28
account-compression: Move sdk to pnpm workspace

joncinque:
#### Problem

During the update to 1.17.6 in https://github.com/solana-labs/solana-program-library/pull/5863, it looks like the build barfs on account-compression not being part of the pnpm workspace

#### Solution

Move the account-compression sdk to the same workspace! This way it'll also pick up dependabot changes like the rest of the repo.

This also required some dependency bumps because the current build has some unmet dependencies according to pnpm. 

Note that the linting isn't working, so I can fix that in a follow-up PR.


.github/workflows/pull-request-account-compression.yml
```diff
@@ -8,6 +8,7 @@ on:
       - "ci/*-version.sh"
       - "ci/install-anchor.sh"
       - ".github/workflows/pull-request-account-compression.yml"
+      - "!account-compression/sdk/**"
```
ngundotra, line 11: what does this do & why add?

.github/workflows/pull-request-account-compression.yml
```diff
@@ -16,6 +17,7 @@ on:
...
```
ngundotra, line 20: same q as above

ngundotra (commented):
Overall looks good! Just 1 qq on the CI stuff

.github/workflows/pull-request-account-compression.yml
joncinque: The `!` means to not run if there's a change in one of those files.

The idea is to *not* run this workflow if there's *only* a change to the JS side, since `pull-request-js.yml` gets run if there's only a change on the JS side. But if there's a change in the program, you *do* want to run this workflow.

.github/workflows/pull-request-account-compression.yml
joncinque: Same a as above

.github/workflows/pull-request-account-compression.yml
ngundotra: But if there's a change in the SDK the tests should run. SDK can change test behavior

joncinque: Yep, the SDK tests are run by `pull-request-js.yml` if there's a change there

ngundotra (approved):
Thanks for explaining!

PR #5916, opened 2023-11-28
[docs]: add edit page button to GitHub PRs

buffalojoec:
This will open the editor on `master` and then prompt the user to commit directly to `master` or create a PR, and they _shouldn't_ have perms to do the former.


joncinque (approved):
Perfect, thanks!

perfect

PR #5926, opened 2023-11-29
[token cli]: add validation to metadata address

buffalojoec:
This PR adds the validation for a valid pubkey on the `--metadata-address <ADDRESS>` arg for `spl-token create-token`.


joncinque (approved):
Thanks!

clean

PR #5945, opened 2023-12-04
stake-pool: Update security and audit info

joncinque:
#### Problem

Even though the new stake pool program was tagged, the hash info and audit information was not included.

#### Solution

Now that the hash is known post-merge, update all the security information.


buffalojoec (approved):
security

PR #5984, opened 2023-12-11
transfer hook cli: bump

buffalojoec:
In step with #5970, bump the transfer hook CLI!


buffalojoec:
Note: Rebased after latest crate version changes.

PR #5987, opened 2023-12-12
associated-token-account: Bump to use token-2022 v1

joncinque:
#### Problem

We need to use token-2022 v1 in the monorepo. Associated-token-account uses it too, but there's no published crate version that uses it.

#### Solution

Bump the ATA crate for a new publish.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6420,7 +6420,7 @@ dependencies = [
  "serde_json",
  "solana-account-decoder",
  "solana-sdk",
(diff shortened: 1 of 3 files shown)


buffalojoec (commented):
Sorry, I always forget how you do this one, but are we bumping this crate in Token Client & CLI?

joncinque:
Those just declare a looser version of `"2.0"`, so they'll be OK. We can pin the version for the CLI in a separate piece of work though.

It probably would be better to do that, and I'm surprised it's not already pinning the SPL deps, strange :thinking:

buffalojoec (approved):

PR #5995, opened 2023-12-13
token-2022: Update audit and release info

joncinque:
#### Problem

The OtterSec audit is now published at https://github.com/solana-labs/security-audits/blob/master/spl/OtterSecToken2022Audit-2023-11-03.pdf, but there's no reference in the SPL repo.

#### Solution

Add the references everywhere.


t-nelson (approved):
lgtm! :shipit:

PR #5997, opened 2023-12-14
transfer hook cli: add support for PDAs as extra metas

### LINKED ISSUE #5661: transfer-hook-cli: Add support for PDA seeds
#### Problem

The simple CLI for transfer hook programs added in #5659 only allows users to specify literal pubkeys for the extra required accounts, and does not allow specifying any array of PDA seeds, similar to the example at https://github.com/solana-labs/solana-program-library/tree/master/libraries/tlv-account-resolution#example-usage

#### Solution

Add the ability to specify all kinds of seeds:

* `ExtraAccountMeta::new_with_seeds`
* `ExtraAccountMeta::new_external_pda_with_seeds`

I'm not sure what the format for these would take as command-line arguments, since they accept `&[Seed]`, where `Seed` is an enum with 4 possibilities (`Literal`, `InstructionData`, `AccountKey`, `AccountData`). We might need to convert these seeds with strum args, which could make it look like:

```
$ spl-transfer-hook create-extra-metas <PROGRAM> <MINT> [literal("test"), instruction-data(2, 2), account-key(1), account-data(3, 3, 3)]:readonly
```

This is hard, but I'm sure there's a solution!

cc @buffalojoec

buffalojoec:
This PR adds support to the Transfer Hook CLI to add extra account meta configurations that involve seed configs!

This feature supports JSON or YAML file types, with a flexible layout to represent the extra metas.

Closes #5661.

### DIFF
--- a/libraries/tlv-account-resolution/src/seeds.rs
+++ b/libraries/tlv-account-resolution/src/seeds.rs
@@ -34,7 +34,7 @@ use {crate::error::AccountResolutionError, solana_program::program_error::Progra
 /// Enum to describe a required seed for a Program-Derived Address
 #[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
-#[cfg_attr(feature = "serde-traits", serde(rename_all = "snake_case"))]
+#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
 pub enum Seed {
     /// Uninitialized configuration byte space
     Uninitialized,
@@ -57,7 +57,7 @@ pub enum Seed {
     ///     * 1 - Discriminator
     ///     * 1 - Start index of instruction data
     ///     * 1 - Length of instruction data starting at index
-    #[cfg_attr(feature = "serde-traits", serde(alias = "instructionData"))]
+    #[cfg_attr(feature = "serde-traits", serde(alias = "instruction_data"))]
     InstructionData {
         /// The index where the bytes of an instruction argument begin
         index: u8,
@@ -72,7 +72,7 @@ pub enum Seed {
     /// Packed as:
     ///     * 1 - Discriminator
     ///     * 1 - Index of account in accounts list
-    #[cfg_attr(feature = "serde-traits", serde(alias = "accountKey"))]
+    #[cfg_attr(feature = "serde-traits", serde(alias = "account_key"))]
     AccountKey {
         /// The index of the account in the entire accounts list
         index: u8,
@@ -83,13 +83,16 @@ pub enum Seed {
     ///     * 1 - Index of account in accounts list
     ///     * 1 - Start index of account data
     ///     * 1 - Length of account data starting at index
-    #[cfg_attr(feature = "serde-traits", serde(alias = "accountData"))]
+    #[cfg_attr(
+        feature = "serde-traits",
+        serde(rename_all = "camelCase", alias = "account_data")
+    )]
     AccountData {
         /// The index of the account in the entire accounts list
-        #[cfg_attr(feature = "serde-traits", serde(alias = "accountIndex"))]
+        #[cfg_attr(feature = "serde-traits", serde(alias = "account_index"))]
         account_index: u8,
         /// The index where the bytes of an account data argument begin
-        #[cfg_attr(feature = "serde-traits", serde(alias = "dataIndex"))]
+        #[cfg_attr(feature = "serde-traits", serde(alias = "data_index"))]
         data_index: u8,
         /// The length of the argument (number of bytes)
         ///
--- a/libraries/tlv-account-resolution/src/seeds.rs
+++ b/libraries/tlv-account-resolution/src/seeds.rs
@@ -57,7 +57,6 @@ pub enum Seed {
     ///     * 1 - Discriminator
     ///     * 1 - Start index of instruction data
     ///     * 1 - Length of instruction data starting at index
-    #[cfg_attr(feature = "serde-traits", serde(alias = "instruction_data"))]
     InstructionData {
         /// The index where the bytes of an instruction argument begin
         index: u8,
@@ -72,7 +71,6 @@ pub enum Seed {
     /// Packed as:
     ///     * 1 - Discriminator
     ///     * 1 - Index of account in accounts list
-    #[cfg_attr(feature = "serde-traits", serde(alias = "account_key"))]
     AccountKey {
         /// The index of the account in the entire accounts list
         index: u8,
@@ -89,10 +87,8 @@ pub enum Seed {
     )]
     AccountData {
         /// The index of the account in the entire accounts list
-        #[cfg_attr(feature = "serde-traits", serde(alias = "account_index"))]
         account_index: u8,
         /// The index where the bytes of an account data argument begin
-        #[cfg_attr(feature = "serde-traits", serde(alias = "data_index"))]
         data_index: u8,
         /// The length of the argument (number of bytes)
         ///
--- a/libraries/tlv-account-resolution/src/seeds.rs
+++ b/libraries/tlv-account-resolution/src/seeds.rs
@@ -27,10 +27,14 @@
 //! No matter which types of seeds you choose, the total size of all seed
 //! configurations must be less than or equal to 32 bytes.
 
+#[cfg(feature = "serde-traits")]
+use serde::{Deserialize, Serialize};
 use {crate::error::AccountResolutionError, solana_program::program_error::ProgramError};
 
 /// Enum to describe a required seed for a Program-Derived Address
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde-traits", serde(rename_all = "snake_case"))]
 pub enum Seed {
     /// Uninitialized configuration byte space
     Uninitialized,
@@ -53,6 +57,7 @@ pub enum Seed {
     ///     * 1 - Discriminator
     ///     * 1 - Start index of instruction data
     ///     * 1 - Length of instruction data starting at index
+    #[cfg_attr(feature = "serde-traits", serde(alias = "instructionData"))]
     InstructionData {
         /// The index where the bytes of an instruction argument begin
         index: u8,
@@ -67,6 +72,7 @@ pub enum Seed {
     /// Packed as:
     ///     * 1 - Discriminator
     ///     * 1 - Index of account in accounts list
+    #[cfg_attr(feature = "serde-traits", serde(alias = "accountKey"))]
     AccountKey {
         /// The index of the account in the entire accounts list
         index: u8,
@@ -77,10 +83,13 @@ pub enum Seed {
     ///     * 1 - Index of account in accounts list
     ///     * 1 - Start index of account data
     ///     * 1 - Length of account data starting at index
+    #[cfg_attr(feature = "serde-traits", serde(alias = "accountData"))]
     AccountData {
         /// The index of the account in the entire accounts list
+        #[cfg_attr(feature = "serde-traits", serde(alias = "accountIndex"))]
         account_index: u8,
         /// The index where the bytes of an account data argument begin
+        #[cfg_attr(feature = "serde-traits", serde(alias = "dataIndex"))]
         data_index: u8,
         /// The length of the argument (number of bytes)
         ///
--- a/token/transfer-hook/cli/src/main.rs
+++ b/token/transfer-hook/cli/src/main.rs
@@ -252,16 +252,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                         .help("The token mint address for the transfer hook"),
                 )
                 .arg(
-                    Arg::with_name("transfer_hook_account")
+                    Arg::with_name("transfer_hook_accounts")
                         .value_parser(parse_transfer_hook_account_arg)
-                        .value_name("PUBKEY:ROLE")
+                        .value_name("TRANSFER_HOOK_ACCOUNTS")
                         .takes_value(true)
                         .multiple(true)
                         .min_values(0)
                         .index(3)
-                        .help("Additional pubkey(s) required for a transfer hook and their \
-                            role, in the format \"<PUBKEY>:<ROLE>\". The role must be \
-                            \"readonly\", \"writable\". \"readonly-signer\", or \"writable-signer\".")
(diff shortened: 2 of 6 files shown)


token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,318 @@
...
+    serde::{Deserialize, Serialize},
+    solana_sdk::pubkey::Pubkey,
+    spl_tlv_account_resolution::{account::ExtraAccountMeta, seeds::Seed},
+    std::{fmt, path::Path, str::FromStr},
+    strum_macros::{EnumString, IntoStaticStr},
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, EnumString, IntoStaticStr, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[strum(serialize_all = "kebab-case")]
```
joncinque, line 11: how about we change this to camel case to be consistent with the rest? Right now, we've got snake-case for YAML, camel-case for JSON, and kebab-case for just this enum. That might be pretty confusing to people, so how about just using one standard for the whole thing?

token/transfer-hook/cli/src/main.rs
```diff
@@ -407,8 +378,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 .unwrap()
                 .unwrap();
             let transfer_hook_accounts = arg_matches
-                .get_many::<AccountMeta>("transfer_hook_account")
+                .get_many::<Vec<ExtraAccountMeta>>("transfer_hook_account")
```
joncinque, line 381: I don't think this works currently
```suggestion
                .get_many::<Vec<ExtraAccountMeta>>("transfer_hook_accounts")
```

token/transfer-hook/cli/src/main.rs
```diff
@@ -446,8 +418,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
```
joncinque, line 421: Same here, I don't think this works currently
```suggestion
                .get_many::<Vec<ExtraAccountMeta>>("transfer_hook_accounts")
```

token/transfer-hook/cli/src/main.rs
```diff
@@ -289,16 +252,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-                        .help("Additional pubkey(s) required for a transfer hook and their \
-                            role, in the format \"<PUBKEY>:<ROLE>\". The role must be \
-                            \"readonly\", \"writable\". \"readonly-signer\", or \"writable-signer\".")
+                        .help("Additional account(s) required for a transfer hook and their \
+                            respective configurations, whether they are a fixed address or PDA. \
+                            \nAdditional accounts with known fixed addresses can be passed at the \
+                            command line in the format \"<PUBKEY>:<ROLE>\". The role must be \
+                            \"readonly\", \"writable\". \"readonly-signer\", or \"writable-signer\". \
+                            \n Additional acounts requiring seed configurations can be defined in a \
+                            configuration file using either JSON or YAML.")
```
joncinque, line 268: Can you give an example of this format here?

token/transfer-hook/cli/src/main.rs
```diff
@@ -332,16 +299,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
```
joncinque, line 315: Same here, can you give an example of this format here?

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,318 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct ExtraAccountMetaConfigFile {
+    #[serde(alias = "extraMetas")]
+    extra_metas: Vec<ExtraAccountMetaConfig>,
+}
```
joncinque, line 38: Same here, let's just camel-case everywhere

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,318 @@
...
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct ExtraAccountMetaConfig {
+    pubkey: Option<String>,
+    seeds: Option<Vec<Seed>>,
+    role: Option<AccountMetaRole>,
+    #[serde(alias = "isSigner")]
+    is_signer: Option<bool>,
+    #[serde(alias = "isWritable")]
+    is_writable: Option<bool>,
+}
```
joncinque, line 49: same here, let's just camel-case everywhere

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,318 @@
...
+#[serde(rename_all = "kebab-case")]
+#[strum(serialize_all = "kebab-case")]
+enum AccountMetaRole {
+    Readonly,
+    Writable,
+    ReadonlySigner,
+    WritableSigner,
+}
+impl AccountMetaRole {
+    fn bools(&self) -> (/* is_signer */ bool, /* is_writable */ bool) {
```
joncinque, line 19: nit: since this is pretty easy to use incorrectly, let's make this self-documenting with a struct like `AccountMetaRoleFlags` with `is_signer` and `is_writable` bool fields

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,318 @@
...
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct ExtraAccountMetaConfigFile {
+    #[serde(alias = "extraMetas")]
+    extra_metas: Vec<ExtraAccountMetaConfig>,
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct ExtraAccountMetaConfig {
+    pubkey: Option<String>,
+    seeds: Option<Vec<Seed>>,
```
joncinque, line 43: nit: how about an enum here since you can't have both? It looks like serde makes this nifty https://serde.rs/enum-representations.html, and it'll make `discriminator_and_address` neater

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,318 @@
...
+        if let Some(pubkey_string) = &self.pubkey {
+            (0, Pubkey::from_str(pubkey_string).unwrap().to_bytes())
+        } else if let Some(seeds) = &self.seeds {
+            (1, Seed::pack_into_address_config(seeds).unwrap())
+        } else {
+            panic!("Either `pubkey` or `seeds` must be present");
+        }
+    }
+
+    fn bools(&self) -> (/* is_signer */ bool, /* is_writable */ bool) {
```
joncinque, line 64: Same here: can this be some new struct for more safety and ease of use?

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,318 @@
...
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+struct ExtraAccountMetaConfig {
+    pubkey: Option<String>,
+    seeds: Option<Vec<Seed>>,
+    role: Option<AccountMetaRole>,
+    #[serde(alias = "isSigner")]
+    is_signer: Option<bool>,
+    #[serde(alias = "isWritable")]
+    is_writable: Option<bool>,
```
joncinque, line 48: nit: Can also have these as an enum to make it harder to mess up

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,318 @@
...
+            let path = Path::new(arg);
+            let parse_fn = match path.extension().and_then(|s| s.to_str()) {
+                Some("json") => |v: &str| {
+                    serde_json::from_str::<ExtraAccountMetaConfigFile>(v)
+                        .map_err(|e| format!("Unable to parse file: {e}"))
+                },
+                Some("yaml") => |v: &str| {
+                    serde_yaml::from_str::<ExtraAccountMetaConfigFile>(v)
+                        .map_err(|e| format!("Unable to parse file: {e}"))
+                },
```
joncinque, line 114: nit-for-safety: let's also match on `yml`
```suggestion
                Some("yaml") | Some("yml") => |v: &str| {
                    serde_yaml::from_str::<ExtraAccountMetaConfigFile>(v)
                        .map_err(|e| format!("Unable to parse file: {e}"))
                },
```

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,318 @@
...
+}
+impl ExtraAccountMetaConfig {
+    fn discriminator_and_address(&self) -> (u8, [u8; 32]) {
+        if self.pubkey.is_some() && self.seeds.is_some() {
+            panic!("Only one of `pubkey` or `seeds` must be present");
+        }
+        if let Some(pubkey_string) = &self.pubkey {
+            (0, Pubkey::from_str(pubkey_string).unwrap().to_bytes())
+        } else if let Some(seeds) = &self.seeds {
+            (1, Seed::pack_into_address_config(seeds).unwrap())
```
joncinque, line 58: Now that the discriminants are being used outside of the crate, I think it's time for a separate enum in spl-tlv-account-resolution

joncinque (commented):
Looks great for the most part! just a lot of nits

token/transfer-hook/cli/src/meta.rs
buffalojoec: Ehh, this may be more trouble that it's worth, since you can't derive `Pod` on an enum, and the discriminator for external PDAs is not static.

Let me instead try to refactor this code.

buffalojoec: Yeah, I've got a better solution. Reminds me I also need to add support for external PDAs.

token/transfer-hook/cli/src/meta.rs
buffalojoec: I left this in for now, but I can change it. I figured since it was a string literal it was OK and made sense, especially because it's the same across JSON and YAML, and the CLI arg.

But I can change it to allow for both camel and snake instead.

token/transfer-hook/cli/src/main.rs
buffalojoec: Sure, once the format is finalized.

buffalojoec:
@joncinque taking another peek at this, can you let me know any more thoughts on the formatting, etc. now that I've pushed a few commits based on your feedback?

Once we agree on the remaining bits I'll add the examples as you requested!

joncinque:
Thanks for the ping! I'll take a look at this tomorrow

libraries/tlv-account-resolution/src/seeds.rs
```diff
@@ -53,6 +57,7 @@ pub enum Seed {
     ///     * 1 - Discriminator
     ///     * 1 - Start index of instruction data
     ///     * 1 - Length of instruction data starting at index
+    #[cfg_attr(feature = "serde-traits", serde(alias = "instruction_data"))]
```
joncinque, line 60: I still highly recommend just sticking with camel-case, because we *will* forget to add the snake-case alias at some point and it will be annoying to debug.

token/transfer-hook/cli/src/meta.rs
joncinque: I would eliminate all snake-case support if I were you. Maybe this can keep kebab-case for consistency with other command-line tools that we provide, but camel-case should be the priority. From what I see, YAML doesn't have hard conventions for casing, so let's use that to our advantage to avoid multiple formats

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,337 @@
...
+    Writable,
+    ReadonlySigner,
+    WritableSigner,
+}
+
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", untagged)]
+enum AccessLevel {
+    Access(Access),
+    Role { role: Role },
```
joncinque, line 32: I can't remember, is there a reason to allow for both of these formats? I'm worried it'll be more confusing than helpful to allow both. I don't have a strong preference for either format

token/transfer-hook/cli/src/meta.rs
```diff
@@ -0,0 +1,337 @@
...
+
+fn get_parse_function(path: &Path) -> Result<ParseFn, String> {
+    match path.extension().and_then(|s| s.to_str()) {
+        Some("json") => Ok(|v: &str| {
+            serde_json::from_str::<ConfigFile>(v).map_err(|e| format!("Unable to parse file: {e}"))
+        }),
+        Some("yaml") | Some("yml") => Ok(|v: &str| {
+            serde_yaml::from_str::<ConfigFile>(v).map_err(|e| format!("Unable to parse file: {e}"))
+        }),
+        _ => Err(format!("Unsupported file extension: {}", path.display())),
```
joncinque, line 114: nit: can you specify that only json and yaml are supported here?

joncinque (commented):
Looks really close! My opinion hasn't changed about the format, but the rest of the changes look great. If you have a strong reason to support all formats, then feel free to go forward with it, but I'm almost sure that I'll forget to maintain all of them at some point.

token/transfer-hook/cli/src/meta.rs
buffalojoec: Weird I swore YAML was always snake case. But you're right, looks like it's not strict. I agree let's get rid of snake case for sure.

token/transfer-hook/cli/src/meta.rs
buffalojoec: Basically the ideas is that it's easier to provide "role" on the CLI argument `<ADDRESS>:<ROLE>` and in the file you could decide to use either one.

I guess it really isn't super necessary, and you can just use the role in the files as well. Just giving people the option. Maybe they need specifically `isSigner` and `isWritable` for some other configuration or scripting, and they can just plug that file into transfer hook configs, too.

joncinque: In that case, I'd lean towards just having the `Role`.  Let's keep it simple for the reference implementation.

buffalojoec:
Alrighty, I went ahead and integrated all feedback!

Here's a screenshot of the help output for the CLI, in case you wanted to see what the JSON and YAML examples looked like:

Screenshot 2024-01-05 at 10 41 54 AM

joncinque (approved):
Looks great!

PR #6000, opened 2023-12-15
governance: Ignore test for changes in monorepo

joncinque:
#### Problem

According to https://github.com/solana-labs/solana/pull/34407, a governance test currently fails because of how solana-program-test manages forks.

#### Solution

To unblock that PR, ignore the test for now. Once solana-program-test does things correctly, we can re-enable it.


pgarg66 (approved):

joncinque:
Merging early to unblock

PR #6006, opened 2023-12-18
Check balances of both accounts after transfering lamports

macaujack:
Made initial balances of both accounts large enough to be rent exempt.

### DIFF
--- a/examples/rust/transfer-lamports/tests/functional.rs
+++ b/examples/rust/transfer-lamports/tests/functional.rs
@@ -2,6 +2,7 @@ use {
     solana_program::{
         instruction::{AccountMeta, Instruction},
         pubkey::Pubkey,
+        rent::Rent,
     },
     solana_program_test::*,
     solana_sdk::{account::Account, signature::Signer, transaction::Transaction},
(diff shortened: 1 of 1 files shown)


CriesofCarrots:
The existing test actually shows the functional bounds (account must be empty or rent-exempt at the end of processing the transaction). Can we add your changes as a separate example/case instead?

macaujack:
Oh, I didn't realize that the test wants to demonstrate functional bounds.If the test wants to show accounts must be empty or rent-exempt at the end of processing the transaction, why is there no verification? I think at least we should verify that the account corresponding to the `source_pubkey` doesn't exist anymore.

I'm just wondering why there is no `assert` or `assert_eq` in a Rust test function... I'm new to Solana so this somewhat confuses me. And sure, I don't mind these changes to be in a separate example. Thanks

CriesofCarrots:
> I think at least we should verify that the account corresponding to the `source_pubkey` doesn't exist anymore.

Well, I'm not sure the test was *intended* to demonstrate the bounds, but it does, and I think that has some value. If you'd like to add asserts on those accounts for existence/balance, I would support that, as well as a 2nd example using 2 rent-exempt accounts.

macaujack:
I think it's better to have the 2nd example to show the correctness after transferring lamports. And I also think it's better to assert the existence of both accounts in the current example. From my (as a newbie) perspective, these 2 improvements could help me get some basics of Solana quickly... Thanks :)

PR #6007, opened 2023-12-18
docs: language update

marctrem:
> close to the same cost

close to the cost
-or-
about the same cost

are suitable replacements


buffalojoec:
Hey thanks for the contribution! I don't think this is a necessary change, however.

marctrem:
It's not a _necessary_ change still the syntax isn't correct. Feel free to merge or close that PR.

PR #6015, opened 2023-12-20
token metadata js: update test suites

buffalojoec:
This PR updates the tests for Token Metadata JS to check its pack-unpack accuracy against hand-rolled code in JavaScript, like its Rust counterpart does.

This eliminates hard-coded bytes extracted from the Rust library's test runs.


buffalojoec:
Shipping since this was just a test change!

PR #6018, opened 2023-12-20
token js: bump to 0.3.10

buffalojoec:
After #5667 we need to bump Token JS so we can cut a new release and folks can start using the Token Metadata tooling!


2501babe:
can we update the packages in the repo to the latest version? looks like the pnpm lockfile switches from getting token locally to getting the old version from npm

2501babe (approved):

PR #6056, opened 2024-01-03
single-pool: Add security.txt

joncinque:
#### Problem

The other new SPL programs that we deploy to the chain have a security.txt, ie #5929 and #5928, but the single pool program doesn't.

#### Solution

Add it in! Since the final audit just completed, you can fill in the information for the hash and release name.


2501babe:
oh this is neat

PR #6057, opened 2024-01-04
token-cli: Allow burn command to accept ALL keyword for amount

steviez:
Several of the token commands, such as `transfer`, support using the `ALL` keyword for the `amount` argument in order to automatically choose all of the tokens in the account.

This PR adds support for the `ALL` keyword to the `burn` command as well.

~~TODO: haven't tested this, should probably do so before this PR is considered for submission~~
UPDATE: I have since tested this, and confirmed that running with the keyword `ALL` worked as expected and selected all available tokens in an account. Please let me know if any other level of testing is desired; it doesn't look like the CLI has unit testing but I'll also admit that I haven't spent much time looking in / contributing in this repo

### DIFF
--- a/token/cli/src/command.rs
+++ b/token/cli/src/command.rs
@@ -1679,10 +1679,16 @@ async fn command_burn(
 
     let token = token_client_from_config(config, &mint_info.address, decimals)?;
 
-    let balance = token.get_account_info(&account).await?.base.amount;
-    let amount = ui_amount
-        .map(|ui_amount| spl_token::ui_amount_to_amount(ui_amount, mint_info.decimals))
-        .unwrap_or(balance);
+    let amount = if let Some(ui_amount) = ui_amount {
+        spl_token::ui_amount_to_amount(ui_amount, mint_info.decimals)
+    } else {
+        if config.sign_only {
+            return Err("Use of ALL keyword to burn tokens requires online signing"
+                .to_string()
+                .into());
+        }
+        token.get_account_info(&account).await?.base.amount
+    };
 
     println_display(
         config,
--- a/token/cli/src/command.rs
+++ b/token/cli/src/command.rs
@@ -1662,21 +1662,15 @@ async fn command_burn(
     config: &Config<'_>,
     account: Pubkey,
     owner: Pubkey,
-    ui_amount: f64,
+    ui_amount: Option<f64>,
     mint_address: Option<Pubkey>,
     mint_decimals: Option<u8>,
     use_unchecked_instruction: bool,
     memo: Option<String>,
     bulk_signers: BulkSigners,
 ) -> CommandResult {
-    println_display(
-        config,
-        format!("Burn {} tokens\n  Source: {}", ui_amount, account),
-    );
-
     let mint_address = config.check_account(&account, mint_address).await?;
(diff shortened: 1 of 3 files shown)


token/cli/src/command.rs
```diff
@@ -1549,28 +1557,37 @@ async fn command_burn(
...
     } else {
         Some(mint_info.decimals)
     };
 
     let token = token_client_from_config(config, &mint_info.address, decimals)?;
+
+    let balance = token.get_account_info(&account).await?.base.amount;
+    let amount = ui_amount
+        .map(|ui_amount| spl_token::ui_amount_to_amount(ui_amount, mint_info.decimals))
+        .unwrap_or(balance);
```
joncinque, line 1580: This will break offline signing, since it'll always try to fetch an account. Can you only fetch this balance if `ui_amount` is `None` AND we're not in sign-only mode?

I couldn't come up with a neat way to do this because of the async function, so I'll accept whatever you come up with, but here's my example:

```suggestion
    let amount = if let Some(ui_amount) = ui_amount {
        spl_token::ui_amount_to_amount(ui_amount, mint_info.decimals)
    } else {
        if config.sign_only {
            return Err("Burning all tokens requires online signing").into();
        }
        token.get_account_info(&account).await?.base.amount
    };
```

joncinque (commented):
Just a little comment about being sure to support offline signing.

And it doesn't look like we have a simple test for burning, but you can see the CLI tests at https://github.com/solana-labs/solana-program-library/blob/master/token/cli/tests/command.rs

Would you mind adding a test for this? Be sure to add your new test to the `main()` function in that file. We use a custom test runner for the CLI to avoid spinning up so many test validators.

steviez:
>  it doesn't look like the CLI has unit testing but I'll also admit that I haven't spent much time looking in / contributing in this repo

> but you can see the CLI tests at https://github.com/solana-labs/solana-program-library/blob/master/token/cli/tests/command.rs

Well now I look silly 😅 , but yes, we should definitely add a unit test before considering this PR for submission. Thanks for pointing it out

steviez:
Have been preoccupied with other stuff the last few weeks, but I will finish out the additions Jon requested this week

token/cli/src/command.rs
steviez: Fixed with https://github.com/solana-labs/solana-program-library/pull/6057/commits/50a0f0606510feb92c0244d75943ee8597b9a561

steviez:
> @steviez , is this still on your list? It still seems useful to me, but if you won't return to it, I could pick up.

Ha yep, slipped off the radar again but this one didn't have much to do so bumped it up the list.

> And it doesn't look like we have a simple test for burning
> ...
> Would you mind adding a test for this?

Of course, and thank you for the code link / tips. I added a basic test, adding something to handle the case where we bail out if using `ALL` with offline signing looks a little more involved but I can take a crack at it if we think we should exercise that case

steviez:
Also, I shifted an unrelated test around in https://github.com/solana-labs/solana-program-library/pull/6057/commits/adaf452cd4a53772f9fdb92dfbaecbb82e1da4eb so that all of the burn tests would be next to each other. I hope this isn't too distracting, but I can remove it from the PR if reviewers would like

joncinque (approved):
Looks great, thanks for contributing!

happy flame

PR #6058, opened 2024-01-04
stake-pool: make `update_validator_list_balance()` more ergonomic

billythedummy:
# Problem

`spl_stake_pool::instruction::update_validator_list_balance()`'s API has a redundant param in `validator_vote_accounts` since this data is already present in the `validator_list: &ValidatorList` param. This also makes usage of the function more prone to error because the actual instruction processor only works on contiguous chunks of the list but this API allows users to pass in any slice of vote accounts.

The function also does several unnecessary `Vec` allocations that can be omitted if an iterator is used instead.

# Solution

Update `update_validator_list_balance()` to just take a `len` parameter that specifies the number of validators on the `ValidatorList` starting from `start_index` to include into the instruction.

Update `update_validator_list_balance()` to remove the unnecessary `Vec` allocations.

I've also changed the `start_index` param from `u32` to `usize` to reduce the number of `u32 <-> usize` conversions required. 

The test files have been updated accordingly, with `update_all()`'s API being simplified as well.

I've kept the fail-silently-and-just-return-an-instruction-that-does-nothing behaviour in order to keep the return type consistent.

### DIFF
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -2,15 +2,13 @@
 
 #![allow(clippy::too_many_arguments)]
 
-use crate::state::ValidatorStakeInfo;
-
 use {
     crate::{
         find_deposit_authority_program_address, find_ephemeral_stake_program_address,
         find_stake_program_address, find_transient_stake_program_address,
         find_withdraw_authority_program_address,
         inline_mpl_token_metadata::{self, pda::find_metadata_account},
-        state::{Fee, FeeType, StakePool, ValidatorList},
+        state::{Fee, FeeType, StakePool, ValidatorList, ValidatorStakeInfo},
         MAX_VALIDATORS_TO_UPDATE,
     },
     borsh::{BorshDeserialize, BorshSchema, BorshSerialize},
--- a/stake-pool/program/src/instruction.rs
+++ b/stake-pool/program/src/instruction.rs
@@ -2,6 +2,8 @@
 
 #![allow(clippy::too_many_arguments)]
 
+use crate::state::ValidatorStakeInfo;
+
 use {
     crate::{
         find_deposit_authority_program_address, find_ephemeral_stake_program_address,
@@ -1370,8 +1372,8 @@ pub fn update_validator_list_balance(
     validator_list_address: &Pubkey,
     reserve_stake: &Pubkey,
     validator_list: &ValidatorList,
-    validator_vote_accounts: &[Pubkey],
-    start_index: u32,
+    len: usize,
+    start_index: usize,
     no_merge: bool,
 ) -> Instruction {
     let mut accounts = vec![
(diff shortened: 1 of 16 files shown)


stake-pool/program/src/instruction.rs
```diff
@@ -1383,43 +1383,54 @@ pub fn update_validator_list_balance(
...
-    );
+    let data = StakePoolInstruction::UpdateValidatorListBalance {
+        start_index: start_index.try_into().unwrap(),
+        no_merge,
+    }
+    .try_to_vec()
+    .unwrap();
+    let validator_list_subslice = match validator_list
+        .validators
+        .get(start_index..start_index.saturating_add(len))
```
buffalojoec, line 1394: This bad boy results in `None` if the length provided results in a range that reaches out of bounds.

https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get

We might want to think about handling the case where I've provided a length that is too large. I would receive an instruction with no stake accounts but I'd be none the wiser until I found this code and inferred the issue I was having.

buffalojoec (commented):
Hey thanks for submitting this!

Although I can appreciate the ergonomics here, I think this approach would limit the set of keys that can be updated to be some subslice of the validator list.

For example, if I wanted to update the validators at indices 0, 5, and 9 of a 12-length validator list, this wouldn't let me do it.

billythedummy:
> For example, if I wanted to update the validators at indices 0, 5, and 9 of a 12-length validator list, this wouldn't let me do it.

Why not? Doesn't a `len` of 1 work?

stake-pool/program/src/instruction.rs
billythedummy: Yes, as mentioned:

> I've kept the fail-silently-and-just-return-an-instruction-that-does-nothing behaviour in order to keep the return type consistent.

The old code had a closure that would return empty vecs for vote accounts that weren't on the list, which was essentially doing the same thing.

I personally think this function would be better if it returned an `Option` or `Result` instead, but that would change the function signature.

billythedummy:
> For example, if I wanted to update the validators at indices 0, 5, and 9 of a 12-length validator list, this wouldn't let me do it.

Oh you're talking about doing it in 1 instruction? As mentioned,

> the actual instruction processor only works on contiguous chunks of the list

You could write a test that does just that: create a 12-length validator list stake pool, execute an `UpdateValidatorListBalance` instruction with validators at indices 0, 5, 9, then tell me what happens :smile: Do validator[5] and validator[9] actually get updated?

Hint: https://github.com/solana-labs/solana-program-library/blob/f36c2fb5a24bd87e04c60a509aec94304798c1a3/stake-pool/program/src/processor.rs#L2291C12-L2336

buffalojoec:
> > For example, if I wanted to update the validators at indices 0, 5, and 9 of a 12-length validator list, this wouldn't let me do it.
> 
> Oh you're talking about doing it in 1 instruction? As mentioned,
> 
> > the actual instruction processor only works on contiguous chunks of the list
> 
> You could write a test that does just that: create a 12-length validator list stake pool, execute an `UpdateValidatorListBalance` instruction with validators at indices 0, 5, 9, then tell me what happens 😄 Do validator[5] and validator[9] actually get updated?
> 
> Hint: https://github.com/solana-labs/solana-program-library/blob/f36c2fb5a24bd87e04c60a509aec94304798c1a3/stake-pool/program/src/processor.rs#L2291C12-L2336

Ahh gotcha. I'm curious why we had it set up like this in the first place, then. I'll take another look ASAP, but it seems like a nice enhancement!

stake-pool/program/src/instruction.rs
billythedummy: @buffalojoec lmk what lab's decision is on changing the function signature to an `Option` or `Result` and i'll make the necessary changes. Thank you!

buffalojoec: Hey @billythedummy we're already changing the function signature here so we're going to actually have to deprecate this and create a new one. Since that's the case, I think it's appropriate to add the `Option` or `Result` handling as you see fit!

buffalojoec:
@billythedummy Sorry I'm just circling back to this now. As far as I can tell, this seems like a nice enhancement!

I mentioned in my one comment above that since this changes the function signature for `instruction::update_validator_list_balance(..)`, we'll have to deprecate that and introduce a new one.

I'm not really sure on a good naming convention. Maybe something like `instruction::update_validator_list_balance_chunk(..)`? If you have anything better, by all means!

joncinque (commented):
Thanks for fixing this up! I can't remember why I did it this way, probably just wanted to get something out as fast as possible :sweat: 

As @buffalojoec mentioned, we'll need to deprecate the old function and create a new one with the change, so once that's done, we can get this merged in

stake-pool/program/src/instruction.rs
```diff
@@ -1363,6 +1363,10 @@ pub fn decrease_additional_validator_stake_with_vote(
 
 /// Creates `UpdateValidatorListBalance` instruction (update validator stake
 /// account balances)
+#[deprecated(
+    since = "1.1.0",
```
billythedummy, line 1367: @joncinque @buffalojoec please update this vers number if not appropriate

buffalojoec: I think that's good, since we're not removing the deprecated helper just yet. Whenever we decide to do so, we can bump to `2.0.0`.

stake-pool/program/src/instruction.rs
```diff
@@ -1423,6 +1427,73 @@ pub fn update_validator_list_balance(
...
 }
 
+/// Creates an `UpdateValidatorListBalance` instruction (update validator stake
+/// account balances) to update `validator_list[start_index..start_index +
+/// len]`.
+///
+/// Returns `None` if:
+/// - `start_index..start_index + len` is out of bounds for
+///   `validator_list.validators`
+pub fn update_validator_list_balance_chunk(
```
buffalojoec, line 1437: Just curious if you think `Option` or `Result` is better here. I know you mentioned both.

To me, it feels a bit like this should be a `Result`, since if the length is out of bounds, it's attempting to create an invalid instruction, which should be fallible. What do you think? Maybe you have some use cases in mind I can't think of.

billythedummy: I just chose `Option` bec thats what `slice.get()` returns and i'm lazy lol. Also because im not sure what the error type should be if it was a `Result`

buffalojoec: How about `ProgramError::InvalidInstructionData`?

billythedummy: Done. Yall could've just decided on this from the start instead of telling me to "add the Option or Result handling as you see fit" lol

buffalojoec (approved):
Looks good! Thanks for adding this!!

PR #6059, opened 2024-01-04
stake-pool-cli: don't send `UpdateValidatorListBalance` transactions for subslices of validator list that have already been updated

billythedummy:
# Problem

The `update` command on the `stake-pool-cli` always sends update transactions for the entire validator list. This is redundant and increases the odds of failure if a previous update command failed in the middle since you don't have to run the `UpdateValidatorListBalance` instruction again on the parts of the validator list that were successfully updated. This is especially true for large stake pools - e.g. solblaze has all 2950 validators on their list.

# Solution

In the `update` command, filter out `UpdateValidatorListBalance` instructions for subslices of the `ValidatorList` that have already been updated. Keep the current behaviour for the `--force` flag.

This builds on PR #6058


stake-pool/cli/src/main.rs
```diff
@@ -1158,14 +1161,55 @@ fn command_update(
 
     let validator_list = get_validator_list(&config.rpc_client, &stake_pool.validator_list)?;
 
-    let (mut update_list_instructions, final_instructions) =
+    let (mut update_list_instructions, final_instructions) = if force {
```
buffalojoec, line 1164: The `--force` flag here is specifically for forcing updates regardless of the last updated epoch. There might be some confusion for CLI users if `--force` suddenly provides more than one override.

Maybe we introduce `--all` or something similar for this?

stake-pool/cli/src/main.rs
```diff
@@ -1158,14 +1161,55 @@ fn command_update(
...
-        );
+        )
+    } else {
+        let (withdraw_authority, _) =
+            find_withdraw_authority_program_address(&spl_stake_pool::id(), stake_pool_address);
+        let update_list_instructions: Vec<Instruction> = validator_list
+            .validators
+            .chunks(MAX_VALIDATORS_TO_UPDATE)
+            .enumerate()
+            .filter_map(|(i, chunk)| {
```
buffalojoec, line 1179: Should we work this into a function in the program `instruction.rs` instead, alongside `update_stake_pool`? Then the functionality could be accessed from more than just the CLI.

buffalojoec (commented):
Nice idea! I left a few comments on CLI ergonomics. Once #6058 moves along we can circle back to this, or we can do this one first and just cherry pick the last commit and move it up, up to you!

stake-pool/cli/src/main.rs
billythedummy: Yes, i was thinking of doing the same. How's `update_stale_stake_pool` for the fn name?

stake-pool/cli/src/main.rs
billythedummy: Yeah that would probably be clearer. I wasn't sure if you guys would be ok with adding additional flags at first

stake-pool/cli/src/main.rs
buffalojoec: Yep, that was what I had in mind, too!

stake-pool/cli/src/main.rs
buffalojoec: Yeah, I'd say let's lean on the new flag to isolate the intention!

oasisMystre:
Anyone has an idea why SetTransferFee instruction is not being implemented? Weird 🤔🤔

billythedummy:
> Once https://github.com/solana-labs/solana-program-library/pull/6058 moves along we can circle back to this, or we can do this one first and just cherry pick the last commit and move it up, up to you!

yeah i think lets finish #6058 first

stake-pool/cli/src/main.rs
billythedummy: Added the `--fresh` flag that updates all validators when set, while default behavior is to only update the stales ones when the flag is not set

stake-pool/cli/src/main.rs
```diff
@@ -2426,6 +2437,12 @@ fn main() {
                     .takes_value(false)
                     .help("Do not automatically merge transient stakes. Useful if the stake pool is in an expected state, but the balances still need to be updated."),
             )
+            .arg(
```
buffalojoec, line 2440: Can you just update the `.help` for `force` to say "Update balances" rather than "Update all balances", since now we have some functionality that may not update _all_ of them?

stake-pool/cli/src/main.rs
```diff
@@ -1158,14 +1159,24 @@ fn command_update(
...
+    } else {
+        spl_stake_pool::instruction::update_stale_stake_pool(
+            &spl_stake_pool::id(),
+            &stake_pool,
+            &validator_list,
+            stake_pool_address,
+            no_merge,
+            epoch_info.epoch,
+        )
+    };
```
buffalojoec, line 1179: I know we chatted about this new flag previously, but what do you think about making this flag something like `--stale-only` and flipping this conditional around?

I think that might be less confusing considering the existing command with just `--force` will still behave the same as before, while `--force --stale-only` would be a little more explicit way to say "do the update anyway but only update stale validators".

buffalojoec (commented):
Nice! Thanks for adding this new functionality. I just had a question about the flag. Curious what you think!

stake-pool/cli/src/main.rs
billythedummy: Done. I didn't change the arg where `command_update()` is called before other commands tho so it defaults to `--stale-only false` for all the other commands

PR #6062, opened 2024-01-05
[spl-record] Remove `Data` type from `RecordData` type

samkim-crypto:
#### Problem
The record program uses the struct `RecordData` to represent record data in accounts.
```
pub struct RecordData {
    pub version: u8,
    pub authority: Pubkey,
    pub data: Data,
}
```
The `Data` type is an array of fixed size 8. It seems like this struct was added just for initial testing. The logic to support arbitrary length record data already exists in the processor logic.

By removing `Data` from the `RecordData` type and using it just as a header for data, we should be able to support arbitrary length record data.

#### Summary of changes
Removed the `Data` struct from state and updated the record program to support arbitrary length data.

This is a follow-up to https://github.com/solana-labs/solana-program-library/pull/6054#discussion_r1441021286.

### DIFF
--- a/record/program/src/instruction.rs
+++ b/record/program/src/instruction.rs
@@ -162,10 +162,7 @@ pub fn close_account(record_account: &Pubkey, signer: &Pubkey, receiver: &Pubkey
 
 #[cfg(test)]
 mod tests {
-    use {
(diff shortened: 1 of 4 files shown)


joncinque (approved):
Perfect, thanks for improving this!

PR #6063, opened 2024-01-05
[spl-record] Add `Reallocate` instruction

samkim-crypto:
#### Problem
The record program is useful when storing arbitrary length data on accounts. However, if the data is too long to fit inside an account, it is not possible to record the data without creating a new account with enough space. It would be nice to be able to reallocate more space to the existing account to store longer record data.

#### Summary of changes
Add a reallocate instruction. I followed the main logic of the reallocate instruction in the token-2022 program.

This PR builds on top of https://github.com/solana-labs/solana-program-library/pull/6062. Once https://github.com/solana-labs/solana-program-library/pull/6062 is merged, I will rebase to master to remove the first few commits.

### DIFF
--- a/record/program/src/instruction.rs
+++ b/record/program/src/instruction.rs
@@ -56,8 +56,8 @@ pub enum RecordInstruction<'a> {
 
     /// Reallocate additional space in a record account
     ///
-    /// If the record account already has enough space to hold the specified data length, then the
-    /// instruction does nothing.
+    /// If the record account already has enough space to hold the specified
+    /// data length, then the instruction does nothing.
     ///
     /// Accounts expected by this instruction:
     ///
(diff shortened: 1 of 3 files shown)


record/program/src/instruction.rs
```diff
@@ -52,19 +53,37 @@ pub enum RecordInstruction<'a> {
...
+    /// Reallocate additional space in a record account
+    ///
+    /// If the record account already has enough space to hold the specified
+    /// data length, then the instruction does nothing.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    /// 0. `[writable]` The record account to reallocate
+    /// 1. `[signer, writable]` The payer account to fund reallocation
+    /// 2. `[]` System program for reallocation funding
```
joncinque, line 66: Rather than passing these accounts in, how about we have the client library top up the account with the new required rent exemption?

joncinque (commented):
The implementation and tests look good! I think we just need to move the lamport transfer logic outside of the program. What do you think?

record/program/src/instruction.rs
samkim-crypto: Okay, yes agreed. I removed the lamports logic! Let me know what you think.

joncinque (approved):
Looks great, thanks!

PR #6065, opened 2024-01-05
Transfer hook offchain and onchain helpers bugfix

### LINKED ISSUE #6064: Transfer Hook: Off-chain and on-chain helpers are resolving keys incorrectly
#### Problem
Token2022's off-chain and on-chain helpers for adding the necessary extra account metas for a transfer instruction that requires a transfer hook are missing a crucial step.

**Note:** This crucial step is missing _only_ from the helpers, thus the on-chain program itself needs no change.

At the bottom of this issue is a test. In it, we're going to build instructions with extra account metas in three ways:
- Adding extra metas to a `TransferChecked` instruction using the Token2022 off-chain helper
- Adding extra metas to an `Execute` instruction using the TLV Account Resolution library's `add_to_instruction`
- Adding extra metas to an `Execute` CPI instruction using the TLV Account Resolution library's `add_to_cpi_instruction`

As you can probably infer, the issue is that we are adding the extra metas to a **transfer** instruction, rather than an `Execute` instruction.

Consider the instruction keys for `TransferChecked` (simplified):

0. [w] Source
1. [] Mint
2. [w] Destination
3. [s] Owner

Now consider the instruction keys as defined in the [SPL Transfer Hook Interface's `Execute` instruction](https://github.com/solana-labs/solana-program-library/blob/f36c2fb5a24bd87e04c60a509aec94304798c1a3/token/transfer-hook/interface/src/instruction.rs#L22-L34):

0. [] Source
1. [] Mint
2. [] Destination
3. [] Owner
4. [] Validation account

As you can see, `Execute` requires the validation (extra metas) account, while `TransferChecked` does not.

When extra account metas are hard-coded addresses, ie. `ExtraAccountMeta::new_with_pubkey(..)`, this is not an issue. However, once any extra account meta uses seed configurations that point to other account keys in the list, this difference in account keys bungles the whole resolution.

The problem here is that, at the time of account resolution, the list of account keys for a transfer instruction and the list of account keys for an `Execute` instruction will yield entirely different resolutions.

For example, if I have a PDA that's based off of the account at index 4, that will be two different accounts depending on the instruction. No bueno.

Consider the test.

```rust
const TRANSFER_HOOK_PROGRAM_ID: Pubkey = Pubkey::new_from_array([1; 32]);

const MINT_PUBKEY: Pubkey = Pubkey::new_from_array([2; 32]);

const MOCK_MINT_STATE: [u8; 234] = [
    0, 0, 0, 0, // COption (4): None = 0
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, // Mint authority (32)
    0, 0, 0, 0, 0, 0, 0, 0, // Supply (8)
    0, // Decimals (1)
    1, // Is initialized (1)
    0, 0, 0, 0, // COption (4): None = 0
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, // Freeze authority (32)
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, // Padding (83)
    1, // Account type (1): Mint = 1
    14, 0, // Extension type (2): Transfer hook = 14
    64, 0, // Extension length (2): 64
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, // Authority (32)
    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
    1, 1, // Transfer hook program ID (32)
];

const MOCK_EXTRA_METAS_STATE: [u8; 86] = [
    105, 37, 101, 197, 75, 251, 102, 26, // Discriminator for `ExecuteInstruction` (8)
    74, 0, 0, 0, // Length of pod slice (4): 74
    2, 0, 0, 0, // Count of account metas (4): 2
    1, // First account meta discriminator (1): PDA = 1
    3, 0, // First seed: Account key at index 0 (2)
    3, 1, // Second seed: Account key at index 1 (2)
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, // No more seeds (28)
    0, // First account meta is signer (1): false = 0
    0, // First account meta is writable (1): false = 0
    1, // Second account meta discriminator (1): PDA = 1
    3, 4, // First seed: Account key at index 4 (2)
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, // No more seeds (30)
    0, // Second account meta is signer (1): false = 0
    0, // Second account meta is writable (1): false = 0
];

async fn mock_fetch_account_data_fn(address: Pubkey) -> AccountDataResult {
    if address == MINT_PUBKEY {
        Ok(Some(MOCK_MINT_STATE.to_vec()))
    } else if address
        == get_extra_account_metas_address(&MINT_PUBKEY, &TRANSFER_HOOK_PROGRAM_ID)
    {
        Ok(Some(MOCK_EXTRA_METAS_STATE.to_vec()))
    } else {
        Ok(None)
    }
}

#[tokio::test]
async fn test_resolve_extra_transfer_account_metas() {
    let spl_token_2022_program_id = crate::id();
    let transfer_hook_program_id = TRANSFER_HOOK_PROGRAM_ID;

    let source_pubkey = Pubkey::new_unique();
    let mut source_data = vec![0; 165]; // Mock
    let mut source_lamports = 0; // Mock
    let source_account_info = AccountInfo::new(
        &source_pubkey,
        false,
        true,
        &mut source_lamports,
        &mut source_data,
        &spl_token_2022_program_id,
        false,
        0,
    );

    let mint_pubkey = MINT_PUBKEY;
    let mut mint_data = MOCK_MINT_STATE.to_vec();
    let mut mint_lamports = 0; // Mock
    let mint_account_info = AccountInfo::new(
        &mint_pubkey,
        false,
        true,
        &mut mint_lamports,
        &mut mint_data,
        &spl_token_2022_program_id,
        false,
        0,
    );

    let destination_pubkey = Pubkey::new_unique();
    let mut destination_data = vec![0; 165]; // Mock
    let mut destination_lamports = 0; // Mock
    let destination_account_info = AccountInfo::new(
        &destination_pubkey,
        false,
        true,
        &mut destination_lamports,
        &mut destination_data,
        &spl_token_2022_program_id,
        false,
        0,
    );

    let authority_pubkey = Pubkey::new_unique();
    let mut authority_data = vec![]; // Mock
    let mut authority_lamports = 0; // Mock
    let authority_account_info = AccountInfo::new(
        &authority_pubkey,
        false,
        true,
        &mut authority_lamports,
        &mut authority_data,
        &system_program::ID,
        false,
        0,
    );

    let validate_state_pubkey = get_extra_account_metas_address(&mint_pubkey, &transfer_hook_program_id);

    let extra_meta_1_pubkey = Pubkey::find_program_address(
        &[
            &source_pubkey.to_bytes(), // Account key at index 0
            &mint_pubkey.to_bytes(),   // Account key at index 1
        ],
        &transfer_hook_program_id,
    )
    .0;
    let mut extra_meta_1_data = vec![]; // Mock
    let mut extra_meta_1_lamports = 0; // Mock
    let extra_meta_1_account_info = AccountInfo::new(
        &extra_meta_1_pubkey,
        false,
        true,
        &mut extra_meta_1_lamports,
        &mut extra_meta_1_data,
        &transfer_hook_program_id,
        false,
        0,
    );

    let extra_meta_2_pubkey = Pubkey::find_program_address(
        &[
            &validate_state_pubkey.to_bytes(), // Account key at index 4
        ],
        &transfer_hook_program_id,
    )
    .0;
    let mut extra_meta_2_data = vec![]; // Mock
    let mut extra_meta_2_lamports = 0; // Mock
    let extra_meta_2_account_info = AccountInfo::new(
        &extra_meta_2_pubkey,
        false,
        true,
        &mut extra_meta_2_lamports,
        &mut extra_meta_2_data,
        &transfer_hook_program_id,
        false,
        0,
    );

    let mut validate_state_data = MOCK_EXTRA_METAS_STATE.to_vec();
    let mut validate_state_lamports = 0; // Mock
    let validate_state_account_info = AccountInfo::new(
        &validate_state_pubkey,
        false,
        true,
        &mut validate_state_lamports,
        &mut validate_state_data,
        &transfer_hook_program_id,
        false,
        0,
    );

    // First use the resolve function to add the extra account metas to the
    // transfer instruction from offchain
    let mut offchain_transfer_instruction = crate::instruction::transfer_checked(
        &spl_token_2022_program_id,
        &source_pubkey,
        &mint_pubkey,
        &destination_pubkey,
        &authority_pubkey,
        &[],
        2,
        9,
    )
    .unwrap();

    resolve_extra_transfer_account_metas(
        &mut offchain_transfer_instruction,
        mock_fetch_account_data_fn,
        &mint_pubkey,
    )
    .await
    .unwrap();

    // Then use the offchain function to add the extra account metas to the
    // _execute_ instruction from offchain
    let mut offchain_execute_instruction = spl_transfer_hook_interface::instruction::execute(
        &transfer_hook_program_id,
        &source_pubkey,
        &mint_pubkey,
        &destination_pubkey,
        &authority_pubkey,
        &validate_state_pubkey,
        0,
    );

    ExtraAccountMetaList::add_to_instruction::<ExecuteInstruction, _, _>(
        &mut offchain_execute_instruction,
        mock_fetch_account_data_fn,
        &MOCK_EXTRA_METAS_STATE,
    )
    .await
    .unwrap();

    // Finally, use the onchain function to add the extra account metas to
    // the _execute_ CPI instruction from onchain
    let mut onchain_execute_cpi_instruction = spl_transfer_hook_interface::instruction::execute(
        &transfer_hook_program_id,
        &source_pubkey,
        &mint_pubkey,
        &destination_pubkey,
        &authority_pubkey,
        &validate_state_pubkey,
        0,
    );
    let mut onchain_execute_cpi_account_infos = vec![
        source_account_info.clone(),
        mint_account_info.clone(),
        destination_account_info.clone(),
        authority_account_info.clone(),
        validate_state_account_info.clone(),
    ];
    let all_account_infos = &[
        source_account_info.clone(),
        mint_account_info.clone(),
        destination_account_info.clone(),
        authority_account_info.clone(),
        validate_state_account_info.clone(),
        extra_meta_1_account_info.clone(),
        extra_meta_2_account_info.clone(),
    ];

    ExtraAccountMetaList::add_to_cpi_instruction::<ExecuteInstruction>(
        &mut onchain_execute_cpi_instruction,
        &mut onchain_execute_cpi_account_infos,
        &MOCK_EXTRA_METAS_STATE,
        all_account_infos,
    )
    .unwrap();

    // Now let's evaluate the different lists of keys

    // The two `Execute` instructions should have the same accounts
    assert_eq!(
        offchain_execute_instruction.accounts,
        onchain_execute_cpi_instruction.accounts,
    );

    // However, the transfer instruction is going to be missing the
    // the validation account at index 4
    assert_ne!(
        offchain_transfer_instruction.accounts,
        offchain_execute_instruction.accounts,
    );
    assert_ne!(
        offchain_transfer_instruction.accounts[4].pubkey,
        validate_state_pubkey,
    );

    // Even though both execute instructions have the validation account
    // at index 4
    assert_eq!(
        offchain_execute_instruction.accounts[4].pubkey,
        validate_state_pubkey,
    );
    assert_eq!(
        onchain_execute_cpi_instruction.accounts[4].pubkey,
        validate_state_pubkey,
    );

    // Despite the differences in indices, we're also going to have entirely
    // different PDAs in each list, and this is the issue
    assert_ne!(
        offchain_transfer_instruction.accounts[4].pubkey,
        extra_meta_1_pubkey,
    );
    assert_eq!(
        offchain_execute_instruction.accounts[5].pubkey,
        extra_meta_1_pubkey,
    );
    assert_eq!(
        onchain_execute_cpi_instruction.accounts[5].pubkey,
        extra_meta_1_pubkey,
    );
    assert_ne!(
        offchain_transfer_instruction.accounts[5].pubkey,
        extra_meta_2_pubkey,
    );
    assert_eq!(
        offchain_execute_instruction.accounts[6].pubkey,
        extra_meta_2_pubkey,
    );
    assert_eq!(
        onchain_execute_cpi_instruction.accounts[6].pubkey,
        extra_meta_2_pubkey,
    );

    // For even more verbosity
    assert!(offchain_transfer_instruction
        .accounts
        .iter()
        .find(|account_meta| account_meta.pubkey == extra_meta_1_pubkey)
        .is_none());
    assert!(offchain_transfer_instruction
        .accounts
        .iter()
        .find(|account_meta| account_meta.pubkey == extra_meta_2_pubkey)
        .is_none());
}
```

#### Proposed Solution
We should be able to insert a few lines of code into the helpers that simply converts a transfer instruction into an `Execute` instruction _before_ resolving the extra account metas.

Since anyone storing extra account meta configs should be considering the list of accounts as it will arrive in their Transfer Hook program's `Execute` instruction, it makes sense to resolve for `Execute`.

After all, the main problem we have here is that we're asking TLV Account Resolution to resolve the extra account metas for an `Execute` instruction, but we're not actually feeding it a valid `Execute` instruction.

buffalojoec:
This PR introduces some modifications to Token2022's off-chain and on-chain helpers that simply convert a transfer instruction into an `Execute` instruction before resolving the extra account metas.

Since anyone storing extra account meta configs should be considering the list of accounts as it will arrive in their Transfer Hook program's `Execute` instruction, it makes sense to resolve for `Execute`.

After all, the main problem we have here is that we're asking TLV Account Resolution to resolve the extra account metas for an `Execute` instruction, but we're not actually feeding it a valid `Execute` instruction. Instead, we're feeding it a transfer instruction, typically `TransferChecked`.

I've also added some tests!

Closes #6064 

---
#### A Note on Token2022 Release
A release is necessary to publish this bugfix. However, since this change is specific to Token2022's off-chain and on-chain helpers, the on-chain program does not need to be modified. Developers will simply need to use the newer version of the crate, as well as the newer versions of the Token Client and CLI if they'd like to use the repaired helpers.

We can communicate the bug in question by describing the exact use cases one may encounter issues:
- Defining extra account metas that have seed configurations pointing to instruction data by index
- Defining extra account metas that have seed configurations pointing to other extra account metas by index (_not_ instruction accounts; these will work just fine)

This means the following extra account meta configurations will still work as expected:
- Defining extra account metas that have fixed addresses
- Defining extra account metas that have seed configurations with literal bytes

Also, while we're at it, I think we need to publish some extremely detailed documentation on how to use Transfer Hook extra metas.

### DIFF
--- a/token/program-2022-test/tests/transfer_hook.rs
+++ b/token/program-2022-test/tests/transfer_hook.rs
@@ -688,23 +688,20 @@ async fn success_transfers_using_onchain_helper() {
     let (source_b_account, destination_b_account) =
         setup_accounts(&token_b_context, Keypair::new(), Keypair::new(), amount).await;
     let authority_b = token_b_context.alice;
-    let account_metas = vec![
+
+    // Since we need to add extra account metas for our swap, which is a
+    // combination of two transfers, we need to resolve the extra metas
+    // for each transfer instruction.
+    let transfer_1_metas = vec![
         AccountMeta::new(source_a_account, false),
         AccountMeta::new_readonly(mint_a, false),
         AccountMeta::new(destination_a_account, false),
         AccountMeta::new_readonly(authority_a.pubkey(), true),
-        AccountMeta::new_readonly(spl_token_2022::id(), false),
-        AccountMeta::new(source_b_account, false),
-        AccountMeta::new_readonly(mint_b, false),
-        AccountMeta::new(destination_b_account, false),
-        AccountMeta::new_readonly(authority_b.pubkey(), true),
-        AccountMeta::new_readonly(spl_token_2022::id(), false),
     ];
-
-    let mut instruction = Instruction::new_with_bytes(swap_program_id, &[], account_metas);
-
+    let mut transfer_1_instruction =
+        Instruction::new_with_bytes(swap_program_id, &[], transfer_1_metas.clone());
     offchain::resolve_extra_transfer_account_metas(
-        &mut instruction,
+        &mut transfer_1_instruction,
         |address| {
             token_a.get_account(address).map_ok_or_else(
                 |e| match e {
@@ -719,8 +716,17 @@ async fn success_transfers_using_onchain_helper() {
     )
     .await
     .unwrap();
+
+    let transfer_2_metas = vec![
+        AccountMeta::new(source_b_account, false),
+        AccountMeta::new_readonly(mint_b, false),
+        AccountMeta::new(destination_b_account, false),
+        AccountMeta::new_readonly(authority_b.pubkey(), true),
+    ];
+    let mut transfer_2_instruction =
+        Instruction::new_with_bytes(swap_program_id, &[], transfer_2_metas.clone());
     offchain::resolve_extra_transfer_account_metas(
-        &mut instruction,
+        &mut transfer_2_instruction,
         |address| {
             token_a.get_account(address).map_ok_or_else(
                 |e| match e {
@@ -736,8 +742,23 @@ async fn success_transfers_using_onchain_helper() {
     .await
     .unwrap();
 
+    let mut swap_metas = vec![
+        AccountMeta::new(source_a_account, false),
+        AccountMeta::new_readonly(mint_a, false),
+        AccountMeta::new(destination_a_account, false),
+        AccountMeta::new_readonly(authority_a.pubkey(), true),
+        AccountMeta::new_readonly(spl_token_2022::id(), false),
+        AccountMeta::new(source_b_account, false),
+        AccountMeta::new_readonly(mint_b, false),
+        AccountMeta::new(destination_b_account, false),
+        AccountMeta::new_readonly(authority_b.pubkey(), true),
+        AccountMeta::new_readonly(spl_token_2022::id(), false),
+    ];
+    swap_metas.extend_from_slice(&transfer_1_instruction.accounts[4..]); // Remaining accounts from transfer 1
+    swap_metas.extend_from_slice(&transfer_2_instruction.accounts[4..]); // Remaining accounts from transfer 2
+    let swap_instruction = Instruction::new_with_bytes(swap_program_id, &[], swap_metas);
     token_a
-        .process_ixs(&[instruction], &[&authority_a, &authority_b])
+        .process_ixs(&[swap_instruction], &[&authority_a, &authority_b])
         .await
         .unwrap();
 }
--- a/token/program-2022-test/tests/transfer_hook.rs
+++ b/token/program-2022-test/tests/transfer_hook.rs
@@ -715,6 +715,7 @@ async fn success_transfers_using_onchain_helper() {
             )
         },
         &mint_a,
+        amount,
     )
     .await
     .unwrap();
@@ -730,6 +731,7 @@ async fn success_transfers_using_onchain_helper() {
             )
         },
         &mint_b,
+        amount,
     )
     .await
     .unwrap();
--- a/token/program-2022/src/offchain.rs
+++ b/token/program-2022/src/offchain.rs
@@ -3,11 +3,14 @@
 pub use spl_transfer_hook_interface::offchain::{AccountDataResult, AccountFetchError};
 use {
     crate::{
+        error::TokenError,
         extension::{transfer_hook, StateWithExtensions},
         state::Mint,
     },
-    solana_program::{instruction::Instruction, program_error::ProgramError, pubkey::Pubkey},
-    spl_transfer_hook_interface::offchain::resolve_extra_account_metas,
+    solana_program::{instruction::Instruction, msg, program_error::ProgramError, pubkey::Pubkey},
+    spl_transfer_hook_interface::{
+        get_extra_account_metas_address, offchain::resolve_extra_account_metas,
+    },
     std::future::Future,
 };
 
@@ -33,10 +36,18 @@ use {
 ///     &mint,
 /// ).await?;
 /// ```
+/// Note that this offchain helper will build a new `Execute` instruction,
+/// resolve the extra account metas, and then add them to the transfer
+/// instruction. This is because the extra account metas are configured
+/// specifically for the `Execute` instruction, which requires five accounts
+/// (source, mint, destination, authority, and validation state), wheras the
+/// transfer instruction only requires four (source, mint, destination, and
+/// authority) in addition to `n` number of multisig authorities.
 pub async fn resolve_extra_transfer_account_metas<F, Fut>(
     instruction: &mut Instruction,
     fetch_account_data_fn: F,
     mint_address: &Pubkey,
+    amount: u64,
 ) -> Result<(), AccountFetchError>
 where
     F: Fn(Pubkey) -> Fut,
@@ -46,14 +57,38 @@ where
         .await?
         .ok_or(ProgramError::InvalidAccountData)?;
     let mint = StateWithExtensions::<Mint>::unpack(&mint_data)?;
+
     if let Some(program_id) = transfer_hook::get_program_id(&mint) {
+        // Convert the transfer instruction into an `Execute` instruction,
+        // then resolve the extra account metas as configured in the validation
+        // account data, then finally add the extra account metas to the original
+        // transfer instruction.
+        if instruction.accounts.len() < 4 {
+            msg!("Not a valid transfer instruction");
+            Err(TokenError::InvalidInstruction)?;
+        }
+
+        let mut execute_ix = spl_transfer_hook_interface::instruction::execute(
+            &program_id,
+            &instruction.accounts[0].pubkey,
+            &instruction.accounts[1].pubkey,
+            &instruction.accounts[2].pubkey,
+            &instruction.accounts[3].pubkey,
+            &get_extra_account_metas_address(mint_address, &program_id),
+            amount,
+        );
+
         resolve_extra_account_metas(
-            instruction,
+            &mut execute_ix,
             fetch_account_data_fn,
             mint_address,
             &program_id,
         )
         .await?;
+
+        instruction
+            .accounts
+            .extend_from_slice(&execute_ix.accounts[5..]);
     }
     Ok(())
 }
--- a/token/program-2022/src/offchain.rs
+++ b/token/program-2022/src/offchain.rs
@@ -92,3 +92,490 @@ where
     }
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use {
+        super::*,
+        solana_program::{account_info::AccountInfo, system_program},
+        solana_program_test::tokio,
+        spl_tlv_account_resolution::state::ExtraAccountMetaList,
+        spl_transfer_hook_interface::instruction::ExecuteInstruction,
+    };
+
+    const TRANSFER_HOOK_PROGRAM_ID: Pubkey = Pubkey::new_from_array([1; 32]);
+
+    const MINT_PUBKEY: Pubkey = Pubkey::new_from_array([2; 32]);
+
+    const MOCK_MINT_STATE: [u8; 234] = [
+        0, 0, 0, 0, // COption (4): None = 0
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, // Mint authority (32)
+        0, 0, 0, 0, 0, 0, 0, 0, // Supply (8)
+        0, // Decimals (1)
+        1, // Is initialized (1)
+        0, 0, 0, 0, // COption (4): None = 0
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, // Freeze authority (32)
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, // Padding (83)
+        1, // Account type (1): Mint = 1
+        14, 0, // Extension type (2): Transfer hook = 14
+        64, 0, // Extension length (2): 64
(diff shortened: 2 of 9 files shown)


oasisMystre:
Anyone has an idea why SetTransferFee instruction is not being implemented? Weird 🤔🤔

buffalojoec:
Hey @joncinque I'm going to break this up into 3 PRs for easier review, but I'd like for you to have a look at the linked issue and the general proposed solution, as well as my notes on release.

joncinque:
Great catch! The issue is definitely valid, it's definitely best to split this up, the release plan makes sense, and we definitely need more docs :sweat_smile:

It makes me realize that seed configurations won't work in all cases. For example, if you're writing a swap program, and a token derives a PDA based on instruction data, then there's no way to get the right address, since the amount will be variable based on the swap calculation.

I'll make some comments on the interface

token/program-2022/src/offchain.rs
```diff
@@ -33,10 +36,18 @@ use {
...
+/// specifically for the `Execute` instruction, which requires five accounts
+/// (source, mint, destination, authority, and validation state), wheras the
+/// transfer instruction only requires four (source, mint, destination, and
+/// authority) in addition to `n` number of multisig authorities.
 pub async fn resolve_extra_transfer_account_metas<F, Fut>(
     instruction: &mut Instruction,
     fetch_account_data_fn: F,
     mint_address: &Pubkey,
+    amount: u64,
 ) -> Result<(), AccountFetchError>
```
joncinque, line 51: Since this is specific to transfers, let's deprecate this function, and create a new one which requires passing in all the transfer parameters, and that way we don't have to assume that a proper transfer instruction is provided.

token/program-2022/src/onchain.rs
```diff
@@ -3,19 +3,91 @@
...
+/// (source, mint, destination, authority, and validation state), wheras the
+/// transfer instruction only requires four (source, mint, destination, and
+/// authority) in addition to `n` number of multisig authorities.
+pub fn resolve_extra_transfer_account_metas_for_cpi<'a>(
+    cpi_instruction: &mut Instruction,
+    cpi_account_infos: &mut Vec<AccountInfo<'a>>,
+    mint_info: &AccountInfo<'a>,
+    additional_accounts: &[AccountInfo<'a>],
+    amount: u64,
+) -> Result<(), ProgramError> {
```
joncinque, line 42: Same here, let's pass the transfer parameters by name, rather than in the instruction

token/program-2022-test/tests/transfer_hook.rs
```diff
@@ -688,23 +688,20 @@ async fn success_transfers_using_onchain_helper() {
...
-        AccountMeta::new_readonly(spl_token_2022::id(), false),
     ];
-
-    let mut instruction = Instruction::new_with_bytes(swap_program_id, &[], account_metas);
-
+    let mut transfer_1_instruction =
+        Instruction::new_with_bytes(swap_program_id, &[], transfer_1_metas.clone());
     offchain::resolve_extra_transfer_account_metas(
-        &mut instruction,
+        &mut transfer_1_instruction,
```
joncinque, line 704: By updating the params on this function, this test should look much clearer!

buffalojoec:
> It makes me realize that seed configurations won't work in all cases. For example, if you're writing a swap program, and a token derives a PDA based on instruction data, then there's no way to get the right address, since the amount will be variable based on the swap calculation.

Actually I'm glad you brought this up. I have been mulling on an idea for a more dynamic approach to some account keys, but I haven't thought through all of the implementation yet.

The use case I had in mind was "what if I want to require an extra account meta that's _not_ a PDA but can be different every time someone invokes my program?". For example, let's say I require some program ID to CPI to, and maybe that program just has to implement some interface. Right now you'd have to make that a static fixed address, but it could be set to `[0u8; 32]` and that just means "wildcard". That way it can be whatever address!

Note if we did add more seed support we still wouldn't have to change Token2022's on-chain program. 🙏🏼

As far as the instruction data issue you presented, I think that's probably okay and not our concern at the moment. Technically no matter what program you're writing, you have to implement the `Execute` instruction. That means your instruction data is always going to be `<discriminator (8)> + <amount (8)>`. If you extend inputs through account state, you're covered with `Seed::AccountData` and if you are doing calculations within the program, that's a really far-out edge case in my opinion that's hard to use as a seed even natively.

Curious your thoughts on either of these points.

buffalojoec:
FYI I've added the comprehensive docs to my list, and I'll produce something next week.

joncinque:
>  That way it can be whatever address!

Seems like that should be a different `ExtraAccountMeta` type entirely. It should be fine to implement that as a non-breaking change and use the newer library in token-2022 for the next deployment if you really want to support it. Has anyone asked for it? I'm not even sure how the resolver would work for that. Given a huge number of accounts, how does it pick which one is the "bonus" one?

> if you are doing calculations within the program, that's a really far-out edge case in my opinion that's hard to use as a seed even natively

We need to always consider supporting DeFi or trading apps, so doing calculations shouldn't be considered an edge case. We'll just have to hope that people won't use instruction data for deriving their transfer hook accounts :grimacing:

buffalojoec:
> > That way it can be whatever address!
> 
> Seems like that should be a different `ExtraAccountMeta` type entirely. It should be fine to implement that as a non-breaking change and use the newer library in token-2022 for the next deployment if you really want to support it. Has anyone asked for it? I'm not even sure how the resolver would work for that. Given a huge number of accounts, how does it pick which one is the "bonus" one?

Agreed, it would be another `ExtraAccountMeta` type. And yeah, this was where I hit the snag, lol.

> > if you are doing calculations within the program, that's a really far-out edge case in my opinion that's hard to use as a seed even natively
> 
> We need to always consider supporting DeFi or trading apps, so doing calculations shouldn't be considered an edge case. We'll just have to hope that people won't use instruction data for deriving their transfer hook accounts 😬

But isn't the instruction data always `Execute`, ie `<discriminator (8)> + <amount (8)>`? That's the Transfer Hook interface. You couldn't use another instruction.

joncinque:
> But isn't the instruction data always Execute, ie <discriminator (8)> + <amount (8)>? That's the Transfer Hook interface. You couldn't use another instruction.

Correct. I was focusing on the `amount` part, which will be impossible to figure out off-chain for trading applications.

buffalojoec:
Closing in favor of the segmented PRs going in for this.

PR #6070, opened 2024-01-05
tlv account resolution: bump to include serde

buffalojoec:
Bumps TLC Account Resolution's patch version to include the new `serde-traits` feature. Backwards compatible without the feature, so we shouldn't need a minor release (?).


joncinque (approved):
Yep, patch bump is all good for this one!

PR #6071, opened 2024-01-07
ic; should work, barring some changes to accomodate the skeleton tests

staccDOTsol:
the CreateMint function works, untested: wrapping spl to spl22
The Wrap and Unwrap functions are still being debugged

pushing draft PR to get some input from the larger community, specifically recommendations around add'l account assertions

wondering aloud too: should we enable management of wrapped token metadata and in the case of t22 the extensions for the authority?

### DIFF
--- a/token-wrap/program/src/processor.rs
+++ b/token-wrap/program/src/processor.rs
@@ -57,6 +57,13 @@ fn process_create_mint(
 
     let rent = Rent::get()?;
 
+    let idempotent = match input[0] {
+        0 => false,
+        1 => true,
+        _ => return Err(ProgramError::InvalidInstructionData),
+    };
+
+
     
     
     assert_eq!(token_program.key, &spl_token::ID, "{}", ProgramError::IncorrectProgramId);
@@ -65,6 +72,34 @@ fn process_create_mint(
     assert_eq!(backpointer_account.owner, &solana_program::system_program::ID, "{}", ProgramError::IncorrectProgramId);
     assert_eq!(backpointer_account.key, &get_wrapped_mint_backpointer_address(unwrapped_mint_account.key), "{}", ProgramError::InvalidArgument);
     assert_eq!(wrapped_mint_account.key, &get_wrapped_mint_address(unwrapped_mint_account.key, token_program.key), "{}", ProgramError::InvalidArgument);
+
+
+    if idempotent 
+        && wrapped_mint_account.owner == token_program.key
+    {
+            
+        // Check Backpointer Account
+        let backpointer_data = Backpointer {
+            unwrapped_mint: *unwrapped_mint_account.key,
+        };
+        let data = &mut backpointer_account.data.borrow();
+        let backpointer_data_bytes = bytemuck::bytes_of(&backpointer_data);
+        if data.len() == backpointer_data_bytes.len() {
+            let mut equal = true;
+            for (i, byte) in backpointer_data_bytes.iter().enumerate() {
+                if data[i] != *byte {
+                    equal = false;
+                    break;
+                }
+            }
+            if equal {
+                return Err(ProgramError::AccountAlreadyInitialized);
+            }
+        }
+    }
+    
+
+        
     match *token_program.key {
         spl_token::ID => {
             // Handle SPL Token logic
--- a/token-wrap/program/src/processor.rs
+++ b/token-wrap/program/src/processor.rs
@@ -1,26 +1,883 @@
+//! src/processor.rs
 //! Program state processor
+//! 
+use std::convert::TryInto;
 
-use {
-    crate::instruction::TokenWrapInstruction,
-    solana_program::{account_info::AccountInfo, entrypoint::ProgramResult, pubkey::Pubkey},
-    spl_token_2022::instruction::decode_instruction_type,
+use solana_program::{
+    account_info::{next_account_info, AccountInfo},
+    entrypoint::ProgramResult,
+    program::{invoke, invoke_signed},
+    program_error::ProgramError,
+    pubkey::Pubkey,
+    rent::Rent,
+    system_instruction, program_pack::Pack, program_option::COption, sysvar::Sysvar, msg,
 };
+use spl_token::state::Mint as MintOld;
+use spl_token_2022::{state::Mint as Mint2022, instruction::decode_instruction_type};
+
+use crate::{get_wrapped_mint_authority, state::Backpointer, instruction::TokenWrapInstruction, get_wrapped_mint_authority_with_seed, get_wrapped_mint_authority_seeds, get_wrapped_mint_address, get_wrapped_mint_backpointer_address, get_wrapped_mint_backpointer_address_seeds, get_wrapped_mint_backpointer_address_with_seed, get_wrapped_mint_signer_seeds, get_wrapped_mint_address_with_seed, get_wrapped_mint_backpointer_address_signer_seeds};
+
+
 
 /// Instruction processor
 pub fn process_instruction(
-    _program_id: &Pubkey,
-    _accounts: &[AccountInfo],
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
     input: &[u8],
 ) -> ProgramResult {
-    match decode_instruction_type(input)? {
+    let instruction = decode_instruction_type(input)?;
+    msg !("instructionii {:?}", instruction);
+    match instruction {
         TokenWrapInstruction::CreateMint => {
-            unimplemented!();
+            process_create_mint(program_id, accounts, input)
         }
         TokenWrapInstruction::Wrap => {
-            unimplemented!();
+            process_wrap(program_id, accounts, input)
         }
         TokenWrapInstruction::Unwrap => {
-            unimplemented!();
+            process_unwrap(program_id, accounts, input)
         }
     }
 }
+
+/// Process 'CreateMint' instruction
+fn process_create_mint(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    input: &[u8],
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+    let funding_account = next_account_info(account_info_iter)?;
+    let wrapped_mint_account = next_account_info(account_info_iter)?;
+    let backpointer_account = next_account_info(account_info_iter)?;
+    let unwrapped_mint_account = next_account_info(account_info_iter)?;
+    let system_program = next_account_info(account_info_iter)?;
+    let token_program = next_account_info(account_info_iter)?;
+    let sysvar_rent_account = next_account_info(account_info_iter)?;
+
+    let rent = Rent::get()?;
+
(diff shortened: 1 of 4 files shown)


staccDOTsol:
todo I just noticed: add signer_seeds on CreateMint for init_mint on the spl_token_2022 match

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+    let wrapped_mint = next_account_info(account_info_iter)?;
+    let user_destination_account = next_account_info(account_info_iter)?;
+    let escrow_mint_authority = next_account_info(account_info_iter)?;
+    let unwrapped_token_program = next_account_info(account_info_iter)?;
+    let wrapped_token_program = next_account_info(account_info_iter)?;
+    msg!("unwrapped_token_program {:?}", unwrapped_token_program.key);
+    msg!("wrapped_token_program {:?}", wrapped_token_program.key);
+
+    let signer = next_account_info(account_info_iter)?;
+    let escrow_unwrapped_account = spl_token_2022::state::Account::unpack(&escrow_account.data.borrow())?;
```
joncinque, line 262: We should be able to avoid this deserialization, and let the token programs take care of the checks

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+
+    assert_eq!(escrow_mint_authority.key, &get_wrapped_mint_authority(wrapped_mint.key), "{}", ProgramError::InvalidArgument);
+    assert_eq!(wrapped_mint.key, &get_wrapped_mint_address(&escrow_unwrapped_account.mint, wrapped_token_program.key), "{}", ProgramError::InvalidArgument);
```
joncinque, line 265: This can use `unwrapped_mint.key` for the address check

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+    
+    assert_eq!(escrow_mint_authority.key, &get_wrapped_mint_authority(wrapped_mint.key), "{}", ProgramError::InvalidArgument);
```
joncinque, line 267: Probably not needed to check this here, since the token program will check this for us

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+    // Parse the amount to wrap from the input
+    let amount = unpack_amount(input[1..9].try_into().unwrap())?;
+    
+           
+    let (_, bump_seed) = get_wrapped_mint_address_with_seed(unwrapped_mint.key, wrapped_token_program.key);
```
joncinque, line 272: I don't think this is right, it should be `get_wrapped_mint_authority_with_seed`

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+    let bumps = &[bump_seed];
+    let wrapped_mint_authority_seeds: &[&[u8]]  = &get_wrapped_mint_signer_seeds(unwrapped_mint.key, wrapped_token_program.key, bumps);
```
joncinque, line 274: Same here, it should be `get_wrapped_mint_authority_signer_seeds`

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+
+
+    match *wrapped_token_program.key {
+        spl_token::ID => {
```
joncinque, line 278: No need to change the logic based on the token program here!

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+            assert_eq!(escrow_account.owner, &spl_token_2022::ID, "{}", ProgramError::IncorrectProgramId);
```
joncinque, line 279: No need to check this, since the program will manage the owner checks

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+            assert_eq!(escrow_unwrapped_account.amount, amount, "{}", ProgramError::InsufficientFunds);
```
joncinque, line 280: Same here

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+            // Transfer unwrapped tokens to the escrow account
+                 
+            let mint_to_user_account_ix = spl_token::instruction::mint_to(
+                &spl_token::ID,
+                wrapped_mint.key,
+                user_destination_account.key,
+                signer.key,
+                &[],
+                amount,
+            )?;
```
joncinque, line 293: This needs to take the wrapped mint authority as the signer

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+
+
+    match *wrapped_token_program.key {
+        spl_token::ID => {
+            assert_eq!(escrow_account.owner, &spl_token_2022::ID, "{}", ProgramError::IncorrectProgramId);
+            assert_eq!(escrow_unwrapped_account.amount, amount, "{}", ProgramError::InsufficientFunds);
+            // Handle wrapping logic for the original SPL Token program
+            msg!("1");
+
+            // Transfer unwrapped tokens to the escrow account
```
joncinque, line 284: There's currently no transfer for the unwrapped tokens, we'll definitely need that!

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+let (wrapped_mint_authority, bump_seed) = get_wrapped_mint_authority_with_seed(wrapped_token_account.key);
+let wrapped_mint_authority_seeds = get_wrapped_mint_authority_seeds(wrapped_token_account.key);
+let signer_seeds = &[
+    &wrapped_mint_authority_seeds[0][..],
+    &wrapped_mint_authority_seeds[1][..],
+    &[bump_seed],
+];
+
+
+assert!(wrapped_token_account.is_signer, "{}", ProgramError::MissingRequiredSignature);
```
joncinque, line 376: The token account shouldn't sign, no need to check this

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+assert_eq!(wrapped_token_account.owner, wrapped_token_program.key, "{}", ProgramError::IncorrectProgramId);
```
joncinque, line 377: No need to check this, the token program will take care of it

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+assert!(amount > 0, "{}", ProgramError::InvalidArgument);
+assert!(escrow_account.lamports() >= amount, "{}", ProgramError::InsufficientFunds);
```
joncinque, line 379: This is checking the lamports in the account against the amount of tokens to burn, which doesn't make sense in this context, so we can remove it

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+assert_eq!(user_unwrapped_token_account.owner, unwrapped_token_program.key, "{}", ProgramError::IncorrectProgramId);
+assert!(signer.is_signer, "{}", ProgramError::MissingRequiredSignature);
```
joncinque, line 381: No need to check these either, the token program will handle that for us

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+
+
+       
+let (_, bump_seed) = get_wrapped_mint_address_with_seed(unwrapped_mint.key, wrapped_token_program.key);
+let bumps = &[bump_seed];
+let wrapped_mint_authority_seeds: &[&[u8]]  = &get_wrapped_mint_signer_seeds(unwrapped_mint.key, wrapped_token_program.key, bumps);
```
joncinque, line 387: We're using the incorrect bump seeds here again, it should be for the mint authority, not the mint

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+
+match *wrapped_token_program.key {
+    spl_token::ID => {
```
joncinque, line 390: No need to change the logic based on the token program! This should work with any number of token programs that conform to the interface

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+        // Burn wrapped tokens from the user's account
+        let burn_wrapped_tokens_ix = spl_token::instruction::burn(
+            &spl_token::ID,
+            wrapped_token_account.key,
+            wrapped_token_account.key,
+            signer.key,
+            &[],
+            amount,
+        )?;
```
joncinque, line 399: The arguments here aren't correct, it should take the mint as the third parameter

joncinque (commented):
Thanks for taking the step to do all this work! There's a bit to fixup and clean up, and I've only looked at the processor functions for this pass. We can refine as we go -- let me know if you have any questions!

token-wrap/program/src/processor.rs
staccDOTsol: Agh. this was confusing tf out of me. late night codesessions

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
-            unimplemented!();
+            process_create_mint(program_id, accounts, input)
         }
         TokenWrapInstruction::Wrap => {
-            unimplemented!();
+            process_wrap(program_id, accounts, input)
         }
         TokenWrapInstruction::Unwrap => {
-            unimplemented!();
+            process_unwrap(program_id, accounts, input)
```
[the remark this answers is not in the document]
staccDOTsol: say again?

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+                    break;
+                }
+            }
+            if equal {
+                return Err(ProgramError::AccountAlreadyInitialized);
+            }
         }
     }
+    
+    let (_, bump_seed) = get_wrapped_mint_address_with_seed(unwrapped_mint_account.key, token_program.key);
```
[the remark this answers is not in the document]
staccDOTsol: can you elaborate?

token-wrap/program/src/processor.rs
```diff
@@ -1,26 +1,915 @@
...
+            &[wrapped_mint_authority_seeds],
+        )?;
+        },
+        _ => return Err(ProgramError::InvalidAccountData),
+    }
+
+
+    // Create and Initialize Backpointer Account
+    let backpointer_lamports = rent.minimum_balance(std::mem::size_of::<Backpointer>());
+    let (_, bump) = get_wrapped_mint_backpointer_address_with_seed(unwrapped_mint_account.key);
```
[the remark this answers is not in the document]
staccDOTsol: i'm not sure I'm following these comments. where's the other call?

PR #6085, opened 2024-01-08
token js: add variable length extension support to `getMintLen`

buffalojoec:
Some people have mentioned that calculating the expected mint size for a mint with a variable-length `TokenMetadata` extension is cumbersome. This is because if you provide `ExtensionType.TokenMetadata` to `getMintLen(..)`, it will throw an error, since it does not have a fixed size.

Currently, developers must calculate the packed length of their token metadata, get the mint length for the remaining extensions using `getMintLen(..)`, and then add 2 bytes for type, 2 bytes for length, and their token metadata packed length to the mint length, manually.

This PR proposes adding variable length extensions as an additional optional parameter to `getMintLen(..)`, which will allow developers to provide variable-length, extensions such as `TokenMetadata`, alongside their calculated extension size to get the full mint size, taking advantage of the helper!

An example of this modified helper in action can be found in the unit tests. Here's a snippet:

(Note, edited with PR feedback)

```ts
const tokenMetadataPackedLength = pack(tokenMetadata).length;
const mintLength = getMintLen([ExtensionType.TransferFeeConfig, ExtensionType.NonTransferable], {
    [ExtensionType.TokenMetadata]: tokenMetadataPackedLength,
})
```

### DIFF
--- a/token/js/src/extensions/extensionType.ts
+++ b/token/js/src/extensions/extensionType.ts
@@ -188,29 +188,26 @@ export function getAccountTypeOfMintType(e: ExtensionType): ExtensionType {
 function getLen(
     extensionTypes: ExtensionType[],
     baseSize: number,
-    variableLengthExtensions?: [ExtensionType, number][]
+    variableLengthExtensions: { [E in ExtensionType]?: number } = {}
 ): number {
-    if (extensionTypes.length === 0) {
+    if (extensionTypes.length === 0 && Object.keys(variableLengthExtensions).length === 0) {
         return baseSize;
     } else {
-        let accountLength =
+        const accountLength =
             ACCOUNT_SIZE +
             ACCOUNT_TYPE_SIZE +
             extensionTypes
                 .filter((element, i) => i === extensionTypes.indexOf(element))
                 .map((element) => addTypeAndLengthToLen(getTypeLen(element)))
-                .reduce((a, b) => a + b);
-        if (variableLengthExtensions) {
-            accountLength += variableLengthExtensions
-                .filter((element, i) => i === variableLengthExtensions.indexOf(element))
+                .reduce((a, b) => a + b, 0) +
+            Object.entries(variableLengthExtensions)
                 .map(([extension, len]) => {
-                    if (!isVariableLengthExtension(extension)) {
+                    if (!isVariableLengthExtension(Number(extension))) {
                         throw Error(`Extension ${extension} is not variable length`);
                     }
                     return addTypeAndLengthToLen(len);
                 })
-                .reduce((a, b) => a + b);
-        }
+                .reduce((a, b) => a + b, 0);
         if (accountLength === MULTISIG_SIZE) {
             return accountLength + TYPE_SIZE;
         } else {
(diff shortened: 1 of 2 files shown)


token/js/src/extensions/extensionType.ts
```diff
@@ -176,17 +185,32 @@ export function getAccountTypeOfMintType(e: ExtensionType): ExtensionType {
...
+        let accountLength =
             ACCOUNT_SIZE +
             ACCOUNT_TYPE_SIZE +
             extensionTypes
                 .filter((element, i) => i === extensionTypes.indexOf(element))
                 .map((element) => addTypeAndLengthToLen(getTypeLen(element)))
                 .reduce((a, b) => a + b);
+        if (variableLengthExtensions) {
+            accountLength += variableLengthExtensions
+                .filter((element, i) => i === variableLengthExtensions.indexOf(element))
```
joncinque, line 205: I might be missing something, but I don't think this filter will do what it intends. It'll only catch if someone provides the exact same size and extension type, but won't de-dupe based on extension type.

How about passing in a dict or object as `[key: ExtensionType]: number` instead?

token/js/src/extensions/extensionType.ts
```diff
@@ -176,17 +185,32 @@ export function getAccountTypeOfMintType(e: ExtensionType): ExtensionType {
...
     } else {
-        const accountLength =
+        let accountLength =
             ACCOUNT_SIZE +
             ACCOUNT_TYPE_SIZE +
             extensionTypes
                 .filter((element, i) => i === extensionTypes.indexOf(element))
                 .map((element) => addTypeAndLengthToLen(getTypeLen(element)))
                 .reduce((a, b) => a + b);
+        if (variableLengthExtensions) {
```
joncinque, line 203: nit / up to you: if the parameter `variableLengthExtensions` becomes a default value `= {}` rather than a nullable, then you can remove the if check and just add the next filter, which would remove this pesky branch

joncinque (commented):
This makes sense, thanks for fixing it! A few comments on the implementation, and the test looks great

token/js/src/extensions/extensionType.ts
buffalojoec: Ah, clever! Type-safe against duplicates, too!

token/js/src/extensions/extensionType.ts
```diff
@@ -195,11 +216,15 @@ function getLen(extensionTypes: ExtensionType[], baseSize: number): number {
     }
 }
 
-export function getMintLen(extensionTypes: ExtensionType[]): number {
-    return getLen(extensionTypes, MINT_SIZE);
+export function getMintLen(
+    extensionTypes: ExtensionType[],
+    variableLengthExtensions?: { [E in ExtensionType]?: number }
```
joncinque, line 221: nit: can this also take a default?
```suggestion
    variableLengthExtensions: { [E in ExtensionType]?: number } = {}
```

joncinque (approved):
Just one last little bit, looks good after that!

token/js/src/extensions/extensionType.ts
buffalojoec: Force-pushed this in 👊🏼

joncinque (approved):

PR #6088, opened 2024-01-09
token-2022: repair offchain extra metas helper

buffalojoec:
This stack introduces the necessary repairs to Token2022's extra account metas
helpers in order to solve the issue reported in #6064.

In this particular PR, Token2022's offchain helper is being repaired by first
adding the necessary steps to build an `Execute` instruction instead of
resolving on a transfer instruction. Then, this introduces a new function that
takes all of the parameters for `instruction::transfer_checked(..)`, to ensure
we are not getting the wrong instruction, as suggested by @joncinque in [this PR
comment](https://github.com/solana-labs/solana-program-library/pull/6065#discussion_r1445425286).

The commit flow is as follows:
1. Repair the original offchain helper, refactor the client and CLI, and
   refactor a transfer hook test in `program-2022-test`.
2. Add the new helper, deprecating the old one.
3. Add tests for both the deprecated and the new helper.


buffalojoec:
* `master`
  * **PR #6088** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/6088?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/6088?utm_source=stack-comment).

token/program-2022/src/offchain.rs
```diff
@@ -33,10 +36,22 @@ use {
...
+#[deprecated(
+    since = "1.1.0",
+    note = "Please use `create_transfer_instruction_with_extra_metas` instead"
+)]
 pub async fn resolve_extra_transfer_account_metas<F, Fut>(
     instruction: &mut Instruction,
     fetch_account_data_fn: F,
     mint_address: &Pubkey,
+    amount: u64,
 ) -> Result<(), AccountFetchError>
```
joncinque, line 55: If this is deprecated, no need to change this function at all, especially to introduce a breaking change!

token/program-2022/src/offchain.rs
```diff
@@ -46,14 +61,1005 @@ where
...
+///
+/// let mint = Pubkey::new_unique();
+/// let client = RpcClient::new_mock("succeeds".to_string());
+/// let mut account_metas = vec![];
+///
+/// get_extra_transfer_account_metas(
+///     &mut account_metas,
+///     |address| self.client.get_account(&address).map_ok(|opt| opt.map(|acc| acc.data)),
+///     &mint,
+/// ).await?;
```
joncinque, line 120: nit: this should probably call `create_transfer_instruction_with_extra_metas`

token/program-2022/src/offchain.rs
```diff
@@ -46,14 +61,1005 @@ where
...
+/// ```
+/// Note that this offchain helper will build a new `Execute` instruction,
+/// resolve the extra account metas, and then add them to the transfer
+/// instruction. This is because the extra account metas are configured
+/// specifically for the `Execute` instruction, which requires five accounts
+/// (source, mint, destination, authority, and validation state), wheras the
+/// transfer instruction only requires four (source, mint, destination, and
+/// authority) in addition to `n` number of multisig authorities.
+#[allow(clippy::too_many_arguments)]
+pub async fn create_transfer_instruction_with_extra_metas<F, Fut>(
```
joncinque, line 130: nit:
```suggestion
pub async fn create_transfer_checked_instruction_with_extra_metas<F, Fut>(
```

token/program-2022/src/offchain.rs
```diff
@@ -46,14 +61,1005 @@ where
...
+        source_pubkey,
+        mint_pubkey,
+        destination_pubkey,
+        authority_pubkey,
+        signer_pubkeys,
+        amount,
+        decimals,
+    )?;
+
+    if token_program_id == &crate::id() {
```
joncinque, line 156: What's this check for?

token/program-2022/src/offchain.rs
```diff
@@ -46,14 +61,1005 @@ where
...
+                &mut execute_ix,
+                fetch_account_data_fn,
+                mint_pubkey,
+                &program_id,
+            )
+            .await?;
+
+            transfer_ix
+                .accounts
+                .extend_from_slice(&execute_ix.accounts[5..]);
```
joncinque, line 187: This seems a bit confusing and error-prone, and shows that `resolve_extra_account_metas` is also a problem, since it's easy to misuse if you don't provide an `execute` instruction.

We should also deprecate `resolve_extra_account_metas` and add a new helper function in the interface that takes in the params and a mutable instruction. It adds the required accounts there, similar to what's happening here. We can call it something like `add_extra_account_metas_for_execute` or something.

What do you think?

token/program-2022/src/offchain.rs
```diff
@@ -46,14 +61,1005 @@ where
...
+        // transfer instruction from offchain
+        let mut offchain_transfer_instruction = crate::instruction::transfer_checked(
+            &spl_token_2022_program_id,
+            &source_pubkey,
+            &mint_pubkey,
+            &destination_pubkey,
+            &authority_pubkey,
+            &[],
+            amount,
+            9,
```
joncinque, line 498: not a big deal, but the decimals are given as 0 in the test mint

token/program-2022/src/offchain.rs
```diff
@@ -46,14 +61,1005 @@ where
...
+        // Still, the transfer instruction is going to be missing the
+        // the validation account at index 4
+        assert_ne!(
+            offchain_transfer_instruction.accounts,
+            offchain_execute_instruction.accounts,
+        );
+        assert_ne!(
+            offchain_transfer_instruction.accounts[4].pubkey,
+            validate_state_pubkey,
+        );
```
joncinque, line 586: I think I'm missing something -- isn't that a problem? The transfer instruction must have *all* required accounts, including the validation account, right?

token/program-2022/src/offchain.rs
```diff
@@ -46,14 +61,1005 @@ where
...
+            offchain_execute_instruction.accounts[4].pubkey,
+            validate_state_pubkey,
+        );
+        assert_eq!(
+            onchain_execute_cpi_instruction.accounts[4].pubkey,
+            validate_state_pubkey,
+        );
+
+        // The most important thing is verifying all PDAs are correct across
+        // all lists
```
joncinque, line 600: micro-nit: I think you could have a lot fewer lines of code by combining the equality checkers into one, ie:
```
let extra_metas = [extra_meta_1_pubkey, extra_meta_2_pubkey, ...];
assert_eq!(
    &offchain_transfer_instruction.accounts[4..].map(|x| x.pubkey).collect::<Vec<_>>(),
    &extra_metas,
);
```

token/program-2022/src/offchain.rs
```diff
@@ -46,14 +61,1005 @@ where
...
+            &mint_pubkey,
+            &destination_pubkey,
+            &authority_pubkey,
+            &[],
+            amount,
+            9,
+            mock_fetch_account_data_fn,
+        )
+        .await
+        .unwrap();
```
joncinque, line 894: Is this the only difference in this test? Since they're quite long, is it possible to add this instruction as another element to check against in the previous test and remove this one?

If not, can you create two `test_case`s for these? That will make the difference with these two tests clear

joncinque (commented):
The guts look great! I think we might be able to make the interface clearer, and otherwise, mostly small points

token/program-2022/src/offchain.rs
buffalojoec: The validation account should be at the 2nd last index. Sorry, I didn't articulate that in the tests, but it's in there along with the transfer hook meta.

buffalojoec: I'll add this in.

token/program-2022/src/offchain.rs
buffalojoec: Couldn't this be legacy token program? Are OK with just letting this gracefully return based on trying to find the transfer hook extension?

token/program-2022/src/offchain.rs
buffalojoec: This might eliminate the need for that first test, then.

token/program-2022/src/offchain.rs
buffalojoec: I'll give the tests a once-over and try to shorten/tighten them up. As per my other comment above, we may not even need the first.

token/program-2022/src/offchain.rs
buffalojoec: I think this makes sense. In theory, if we wanted to provide a helper that can add extra metas to _any_ instruction (which we know now is... a bit sketchy), it should probably live in `tlv-account-resolution`.

Since this is specifically a `spl-transfer-hook-interface` helper, yeah I agree it should be robust and designed specifically to create `execute` instructions.

buffalojoec:
Thanks for the review @joncinque! If you're not opposed, I think I'll rewrite some of this commit history to shuffle around changes, rather than push up several new ones. Lmk.

token/program-2022/src/offchain.rs
joncinque: legacy mints can be properly unpacked with `StateWithExtensions`, so i don't see a specific reason for this check

joncinque:
Yep, works for me!

buffalojoec:
Closing in favor of the new stack at #6099 - all comments here resolved there.

PR #6097, opened 2024-01-09
token transactions should request cu limits

taozhu-chicago:
Some of zk instructions consumes more than default instruction units (eg 200_000 CUs), transactions with these instructions will failed with error `ComputationalBudgetExceeded` since feature gate #30620 has activated everywhere, unless transactions request enough CU Limits.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -515,6 +515,16 @@ where
         let fee_payer = Some(&payer_key);
 
         {
+            // the most expensive instruction is VERIFY_TRANSFER_WITH_FEE_COMPUTE_UNITS
+            let requested_units = 500_000;
+            instructions.push(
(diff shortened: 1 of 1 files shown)


token/client/src/token.rs
```diff
@@ -514,6 +514,16 @@ where
         let payer_key = self.payer.pubkey();
         let fee_payer = Some(&payer_key);
 
+        {
+            // the most expensive instruction is VERIFY_TRANSFER_WITH_FEE_COMPUTE_UNITS
+            let requested_units = 500_000;
```
taozhu-chicago, line 519: Ideally transaction should be simulated to get required CU limits; or can manually count CUs while preparing token instructions. Setting a dummy number for now to test if this works.

PR #6099, opened 2024-01-10
transfer hook: add new offchain helper

buffalojoec:
This PR adds a new offchain helper for adding the necessary account metas for an 
`ExecuteInstruction` to the SPL Transfer Hook interface, deprecating the old 
one.

As described in #6064, the offchain helper in Token2022 was using the original 
offchain helper from the SPL Transfer Hook interface incorrectly when resolving 
extra account metas for a transfer.

In order to provide a safer, more robust helper, this new function takes the 
instruction, fetch account data function, as well as the individual arguments 
for `instruction::execute(..)`. This will help to ensure Token2022 as well as 
anyone else using the helpers from the SPL Transfer Hook interface are properly 
resolving the necessary additional accounts.

Note: Although deprecated, the original helper in the SPL Transfer Hook 
interface is not broken. It's just less safe to use than this new helper, since 
it can easily be misused.

### DIFF
--- a/token/program-2022/src/offchain.rs
+++ b/token/program-2022/src/offchain.rs
@@ -1,14 +1,12 @@
 //! Offchain helper for fetching required accounts to build instructions
 
 pub use spl_transfer_hook_interface::offchain::{AccountDataResult, AccountFetchError};
-#[allow(deprecated)]
 use {
     crate::{
         extension::{transfer_hook, StateWithExtensions},
         state::Mint,
     },
     solana_program::{instruction::Instruction, program_error::ProgramError, pubkey::Pubkey},
-    spl_transfer_hook_interface::offchain::resolve_extra_account_metas,
     std::future::Future,
 };
 
@@ -34,7 +32,6 @@ use {
 ///     &mint,
 /// ).await?;
 /// ```
-#[allow(deprecated)]
 pub async fn resolve_extra_transfer_account_metas<F, Fut>(
     instruction: &mut Instruction,
     fetch_account_data_fn: F,
@@ -49,7 +46,8 @@ where
         .ok_or(ProgramError::InvalidAccountData)?;
     let mint = StateWithExtensions::<Mint>::unpack(&mint_data)?;
     if let Some(program_id) = transfer_hook::get_program_id(&mint) {
-        resolve_extra_account_metas(
+        #[allow(deprecated)]
+        spl_transfer_hook_interface::offchain::resolve_extra_account_metas(
             instruction,
             fetch_account_data_fn,
             mint_address,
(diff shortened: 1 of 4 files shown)


buffalojoec:
* `master`
  * **PR #6099** <a href="https://app.graphite.dev/github/pr/solana-labs/solana-program-library/6099?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/6099?utm_source=stack-comment).

token/transfer-hook/interface/src/offchain.rs
```diff
@@ -68,3 +76,236 @@ where
...
+///     &destination,
+///     &authority,
+///     amount,
+///     |address| self.client.get_account(&address).map_ok(|opt| opt.map(|acc| acc.data)),
+/// )
+/// .await?;
+/// ```
+#[allow(clippy::too_many_arguments)]
+pub async fn add_extra_account_metas_for_execute<F, Fut>(
+    instruction: &mut Instruction,
```
joncinque, line 102: nit: can you add a comment about what's expected to be in this instruction? It seems like "all of the transfer accounts, none of the transfer hook accounts"

token/program-2022/src/offchain.rs
```diff
@@ -1,6 +1,7 @@
 //! Offchain helper for fetching required accounts to build instructions
 
 pub use spl_transfer_hook_interface::offchain::{AccountDataResult, AccountFetchError};
+#[allow(deprecated)]
```
joncinque, line 4: nit: rather than having one huge `allow` on the whole import statement, it might be clearer to only add it on the import for `resolve_extra_account_metas`.

But I just noticed that gets fixed in the next PR, so never mind!

joncinque (approved):
Looks great! Just a few nits to clean up, but feel free to merge after they're addressed

token/program-2022/src/offchain.rs
buffalojoec: No worries, I'll add it here anyway!

token/transfer-hook/interface/src/offchain.rs
```diff
@@ -68,3 +76,236 @@ where
...
+    program_id: &Pubkey,
```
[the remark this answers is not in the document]
buffalojoec: I think I'd rather add a comment for this, since technically this function can stand on its own aside from creating `ExecuteInstruction` from a transfer hook context. We also call it `program_id` in `instruction::execute(..)`. Wdyt?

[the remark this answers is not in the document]
joncinque: a comment works too!

joncinque (dismissed):
good-boy-helper

Bah the embed isn't working: https://media.giphy.com/media/H4zaKzL5e7ypqn10zi/giphy.gif

buffalojoec:
### Merge activity

* **Jan 11, 4:15 PM**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/6099).

PR #6106, opened 2024-01-10
pod: Add in-place stake history deserialization

joncinque:
#### Problem

The stake history sysvar is too large to deserialize on-chain.

#### Solution

Using pod types, create an in-place deserializer for `StakeHistory`. It's a bit goofy, but it should give an overall look at how it can be done. A few quirks about this, which is why it's only a draft:

* bincode serializes vector sizes as a u64, whereas most other tools do a u32, which is why there's a new type parameter on `PodSlice` and a new (goofy) conversion function into `usize` for the pod number types
* the stake program types might need to start taking in the `Pod` types for things like inflation calculation, I haven't looked that far into it

Let me know how this looks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7036,6 +7036,7 @@ name = "spl-pod"
 version = "0.1.0"
 dependencies = [
  "base64 0.21.6",
+ "bincode",
  "borsh 0.10.3",
  "bytemuck",
  "serde",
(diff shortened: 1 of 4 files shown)


joncinque:
Reopening in case it might still be useful

buffalojoec:
I may be interested in this.

2501babe:
@buffalojoec you may be interested in https://github.com/solana-labs/solana/pull/35252 also

buffalojoec:
> @buffalojoec you may be interested in https://github.com/solana-labs/solana/pull/35252 also

Ohhh yeah. I didn't realize you were tackling the sysvar, too!

PR #6107, opened 2024-01-10
token client: refactor transfer to use new offchain helper

buffalojoec:
This PR continues the necessary repairs for addressing #6064 by refactoring the
Token Client's `transfer(..)` function to use the new offchain transfer hook
helpers.

If transfer hook accounts are provided, in either case they're appended to the
instruction like before.

If no transfer hook accounts are provided:
- If decimals are provided, Token2022's offchain helper
  `create_transfer_checked_instruction_with_extra_metas(..)` is used.
- If decimals are not provided, SPL Transfer Hook interface's
  `add_extra_account_metas_for_execute(..)` is used.

In either case where no transfer hook accounts are provided, the new,
non-deprecated helpers are used.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -913,43 +913,71 @@ where
         let signing_pubkeys = signing_keypairs.pubkeys();
         let multisig_signers = self.get_multisig_signers(authority, &signing_pubkeys);
 
-        let mut instruction = if let Some(decimals) = self.decimals {
-            instruction::transfer_checked(
-                &self.program_id,
-                source,
-                &self.pubkey,
-                destination,
-                authority,
-                &multisig_signers,
-                amount,
-                decimals,
-            )?
-        } else {
-            #[allow(deprecated)]
-            instruction::transfer(
-                &self.program_id,
-                source,
-                destination,
-                authority,
-                &multisig_signers,
-                amount,
-            )?
+        let fetch_account_data_fn = |address| {
+            self.client
+                .get_account(address)
+                .map_ok(|opt| opt.map(|acc| acc.data))
         };
-        if let Some(transfer_hook_accounts) = &self.transfer_hook_accounts {
-            instruction.accounts.extend(transfer_hook_accounts.clone());
+
+        let instruction = if let Some(decimals) = self.decimals {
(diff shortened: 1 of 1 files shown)


token/client/src/token.rs
```diff
@@ -913,43 +913,71 @@ where
...
+                .get_account(address)
+                .map_ok(|opt| opt.map(|acc| acc.data))
+        };
+
+        let instruction = if let Some(decimals) = self.decimals {
+            if let Some(transfer_hook_accounts) = &self.transfer_hook_accounts {
+                let mut instruction = instruction::transfer_checked(
+                    &self.program_id,
+                    source,
+                    &self.pubkey,
```
joncinque, line 927: micro-nit: better to use `self.get_address()` for consistency, as you did in the new code

token/client/src/token.rs
```diff
@@ -913,43 +913,71 @@ where
...
+                        destination,
+                        authority,
+                        amount,
+                        fetch_account_data_fn,
+                    )
+                    .await
+                    .map_err(|_| TokenError::AccountNotFound)?;
+                }
+            };
+            instruction
```
joncinque, line 980: Maybe I'm missing something, but we can probably remove all of this -- if the decimals aren't there, then we don't do a transfer checked, which means the mint isn't passed in, so we can't use the transfer hook anyway.

joncinque (commented):
I might have misunderstood the unchecked portion, but after that, this looks good!

token/client/src/token.rs
buffalojoec: Ohhh good call, yeah we can probably do that!

buffalojoec: Wait, I'm not sure if this is accurate. The previous implementation was adding extra metas to either one. If a transfer is being requested, the mint should be passed in, no?

buffalojoec: Ahh, okay, I can update this implementation then!

joncinque (dismissed):

buffalojoec:
### Merge activity

* **Jan 11, 4:15 PM**: @buffalojoec started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/solana-labs/solana-program-library/6107).

PR #6109, opened 2024-01-10
ci: Only run one of each workflow at once for a branch / PR

joncinque:
#### Problem

It's easy for our CI machines to get bogged down if many PRs go in quickly, which often happens when we have multiple dependabot PRs in at once. This happens because CI is run at the same time for every single commit.

#### Solution

Following https://github.com/solana-labs/solana/pull/34733, limit the concurrency on PRs and branches to one per workflow type.


yihau (approved):
🪖

PR #6119, opened 2024-01-11
Transfer hook: bump versions

buffalojoec:
This PR bumps the versions of the transfer hook interface and CLI. Commits:
1. `interface`: 0.4.1 -> 0.5.0
2. `cli`: 0.1.0 -> 0.1.1 (The CLI doesn't use the helpers)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7666,7 +7666,7 @@ dependencies = [
 
 [[package]]
 name = "spl-transfer-hook-cli"
-version = "0.1.0"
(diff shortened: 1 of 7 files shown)


joncinque:
If we're bumping to 0.5, we could just remove the bad helpers beforehand

buffalojoec:
Rebased here and we should be ready to rock n' roll.

joncinque (approved):
bump

Issue #6118, opened 2024-01-11
Token Client: `create_recipient_associated_account_and_transfer` does not use transfer hook


buffalojoec:
#### Problem
The [`transfer`](https://github.com/solana-labs/solana-program-library/blob/0efab2430227e78d32d01a4434509779d5d9c3f4/token/client/src/token.rs#L905) function in the token client uses the offchain helper to add extra account metas to a transfer, should that mint have a transfer hook extension.

The [`create_recipient_associated_account_and_transfer`](https://github.com/solana-labs/solana-program-library/blob/0efab2430227e78d32d01a4434509779d5d9c3f4/token/client/src/token.rs#L960) client function does not.

#### Solution
Follow the pattern laid in `transfer` to add the helper to `create_recipient_associated_account_and_transfer`!

PR #6120, opened 2024-01-12
[token-client] Added transfer-hook compatibility to `create_recipient_associated_account_and_transfer`

tonton-sol:
Added the necessary extra metas resolution to  `create_recipient_associated_account_and_transfer` with the same pattern as `transfer` in the Token CLI as per #6118.


buffalojoec (approved):
Thanks! That was quick!

flash

tonton-sol:
Yes! I actually implemented it for personal testing yesterday!　🤣

PR #6121, opened 2024-01-12
[token-client] Add interface to add additional compute budget for transactions

samkim-crypto:
#### Problem
With the new activation of the feature `8pgXCMNXC8qyEFypuwpXyRxLXZdpM4Qo72gJ6k87A6wL - Native program should consume compute units` activated, the token-2022 tests seem to break downstream ci in the monorepo due to the token-2022 confidential transfer with fee instruction taking too many compute units.

#### Summary of Changes
- Update the token-client by adding `process_ixs_with_additional_compute_budget`, which adds an extra compute budget instruction to the transaction
- Use the new function to allocate additional compute budget for transfer with fee

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -604,20 +604,26 @@ where
         token_instructions: &[Instruction],
         signing_keypairs: &S,
     ) -> TokenResult<T::Output> {
-        self.process_ixs_with_additional_compute_budget(token_instructions, None, signing_keypairs)
+        let transaction = self
+            .construct_tx(token_instructions, None, signing_keypairs)
+            .await?;
+
+        self.client
+            .send_transaction(&transaction)
             .await
+            .map_err(TokenError::Client)
     }
 
     pub async fn process_ixs_with_additional_compute_budget<S: Signers>(
         &self,
         token_instructions: &[Instruction],
-        additional_compute_budget: Option<u32>,
+        additional_compute_budget: u32,
         signing_keypairs: &S,
     ) -> TokenResult<T::Output> {
         let transaction = self
             .construct_tx(
                 token_instructions,
-                additional_compute_budget,
+                Some(additional_compute_budget),
                 signing_keypairs,
             )
             .await?;
@@ -2655,7 +2661,7 @@ where
                 &multisig_signers,
                 proof_location,
             )?,
-            Some(TRANSFER_WITH_FEE_COMPUTE_BUDGET),
+            TRANSFER_WITH_FEE_COMPUTE_BUDGET,
             signing_keypairs,
         )
         .await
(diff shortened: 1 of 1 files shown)


joncinque (dismissed):
Just one nit, then all good!

token/client/src/token.rs
```diff
@@ -594,9 +603,23 @@ where
...
         signing_keypairs: &S,
+    ) -> TokenResult<T::Output> {
+        self.process_ixs_with_additional_compute_budget(token_instructions, None, signing_keypairs)
+            .await
+    }
+
+    pub async fn process_ixs_with_additional_compute_budget<S: Signers>(
+        &self,
+        token_instructions: &[Instruction],
+        additional_compute_budget: Option<u32>,
```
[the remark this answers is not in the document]
samkim-crypto: Yeah looking at it now, I guess it is kind of weird. I wanted this function to be a more general version of `process_ixs` which `process_ixs` can invoke, but the name is misleading. I'll fix the syntax to take in a normal `u32`.

token/client/src/token.rs
```diff
@@ -2618,7 +2647,10 @@ where
             .new_decryptable_available_balance(transfer_amount, source_aes_key)
             .map_err(|_| TokenError::AccountDecryption)?;
 
-        self.process_ixs(
+        // additional compute budget required for `VerifyTransferWithFee`
+        const TRANSFER_WITH_FEE_COMPUTE_BUDGET: u32 = 500_000;
```
joncinque, line 2651: Sorry I didn't catch this on the first review, but what about the other version of transfer with fee: `confidential_transfer_transfer_with_fee_and_split_proofs` and the parallel version?

joncinque (commented):
Sorry, just one more question. the other change looks good!

token/client/src/token.rs
samkim-crypto: Yeah, it seems like those don't seem to get caught on the CU limit. The split proofs (range proof, ciphertext validity proof, etc.) have smaller CUs since they are essentially chunks of the `VerifyTransferWithFee`. Some of those exceed 200k limit, but the CU averaging seems to allow them to pass.

The `VerifyTransferWithFee` is the culmination of all those split proofs and therefore, have huge CU requirement for which even the CU averaging is not sufficient to bring it down to 200k.

samkim-crypto: I think it might be safe to allocate additional budget for those lighter instructions as well, but perhaps it could be done in a follow-up PR to unblock https://github.com/solana-labs/solana/pull/34695 before the feature freeze.

joncinque: Ah ok, as long as the other ones work, then all good!

joncinque (approved):

PR #6122, opened 2024-01-12
token 2022 & transfer hook: drop deprecated helpers

buffalojoec:
Ahead of #6119, if we're going to do releases, we can drop the deprecated helpers altogether.

This PR removes the deprecated `offchain` and `onchain` helpers from SPL Transfer Hook interface, as well as the deprecated `offchain` helper from Token2022.

### DIFF
--- a/token/program-2022/src/offchain.rs
+++ b/token/program-2022/src/offchain.rs
@@ -11,58 +11,6 @@ use {
     std::future::Future,
 };
 
-/// Offchain helper to get all additional required account metas for a checked
-/// transfer
-///
-/// To be client-agnostic and to avoid pulling in the full solana-sdk, this
-/// simply takes a function that will return its data as `Future<Vec<u8>>` for
-/// the given address. Can be called in the following way:
-///
-/// ```rust,ignore
-/// use futures_util::TryFutureExt;
-/// use solana_client::nonblocking::rpc_client::RpcClient;
-/// use solana_program::pubkey::Pubkey;
-///
-/// let mint = Pubkey::new_unique();
-/// let client = RpcClient::new_mock("succeeds".to_string());
-/// let mut account_metas = vec![];
-///
-/// get_extra_transfer_account_metas(
-///     &mut account_metas,
-///     |address| self.client.get_account(&address).map_ok(|opt| opt.map(|acc| acc.data)),
-///     &mint,
-/// ).await?;
-/// ```
-#[deprecated(
-    since = "1.1.0",
-    note = "Please use `create_transfer_checked_instruction_with_extra_metas` instead"
-)]
-pub async fn resolve_extra_transfer_account_metas<F, Fut>(
-    instruction: &mut Instruction,
-    fetch_account_data_fn: F,
-    mint_address: &Pubkey,
-) -> Result<(), AccountFetchError>
-where
-    F: Fn(Pubkey) -> Fut,
-    Fut: Future<Output = AccountDataResult>,
-{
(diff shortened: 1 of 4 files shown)


buffalojoec:
I've also added a commit to do JavaScript here, too. But if we don't think it belongs in this PR I can roll it back!

token/program-2022/src/offchain.rs
```diff
@@ -11,58 +11,6 @@ use {
...
-///     &mut account_metas,
-///     |address| self.client.get_account(&address).map_ok(|opt| opt.map(|acc| acc.data)),
-///     &mint,
-/// ).await?;
-/// ```
-#[deprecated(
-    since = "1.1.0",
-    note = "Please use `create_transfer_checked_instruction_with_extra_metas` instead"
-)]
-pub async fn resolve_extra_transfer_account_metas<F, Fut>(
```
joncinque, line 40: Oh darn, sorry, I forgot about this helper here. We'll need to do a new major version release of token-2022.

I think that's fine, considering we'll also formally enable the token groups with 1.18, but we should be aware of that.

buffalojoec: Do you want to do that now or wait a little bit?

joncinque: we can hold off on releasing token-2022 v2 until Solana v1.18, but feel free to get all the other crates out!

buffalojoec: The client would be on the deprecated offchain helper until we released Token2022. Unless we temporarily injected the offchain helper from SPL Transfer Hook temporarily, which we could do.

buffalojoec: The client would just have to manually do this:

https://github.com/solana-labs/solana-program-library/blob/23916b24700cd222ad2397e9de8162ad6980e2e8/token/program-2022/src/offchain.rs#L104-L134

And anyone attempting a CPI to Token2022 with extra metas would need to do this setup manually as well:

https://github.com/solana-labs/solana-program-library/blob/23916b24700cd222ad2397e9de8162ad6980e2e8/token/program-2022/src/onchain.rs#L31-L79

buffalojoec: So CPIs would be a little boned until that comes out. But the client we can sneak that in.

joncinque: gotcha, we can get token-2022 v2 out sooner then, but we'll just need to clarify in the release notes that it's not going on-chain

buffalojoec:
I want to update this PR to _just_ drop the helpers in the SPL Transfer Hook interface. However, this means Token2022 wouldn't use the version in the monorepo, instead it would be fixed to `spl-transfer-hook-interface = "=0.4`. Lmk.

buffalojoec:
> > I want to update this PR to just drop the helpers in the SPL Transfer Hook interface.
> 
> Why don't we just drop them all? That'll be best for sure

Alrighty, so then this PR has all of them dropped. That means we're looking at the following releases:
- SPL Transfer Hook interface: `0.5.0`
- SPL Token JS: `0.4.0`
- SPL Token 2022: `2.0.0`

Seems I can probably cut & release the SPL Transfer Hook one, do we have any planned changes we'd like to include in the JS minor release, or should I cut that now, too? Token2022 I guess I'll wait until we have the confidential stuff wrapped?

joncinque:
That all sounds right to me! And yes, let's hold off on publishing token-2022 v2 until the last audit is done

joncinque (approved):

PR #6126, opened 2024-01-12
fix: 'owner' should be 'authority' in associated-token-account.md

mikemaccana:
The old wording seemed incorrect. Solana uses the term 'owner' to mean program owner. I believe this sentence is saying the authority on the new ATA will be the associated wallet, can someone confirm that's correct?

### DIFF
--- a/docs/src/associated-token-account.md
+++ b/docs/src/associated-token-account.md
@@ -86,5 +86,4 @@ If the associated token account for a given wallet address does not yet exist,
 it may be created by *anybody* by issuing a transaction containing the
 instruction returned by [create_associated_token_account](https://docs.rs/spl-associated-token-account/latest/spl_associated_token_account/instruction/fn.create_associated_token_account.html).
 
-Regardless of creator the new associated token account will be fully owned by
-the wallet, as if the wallet itself had created it.
+Regardless of the creator, the new associated token account's authority will be the wallet, as if the wallet itself had created the associated token account.


buffalojoec:
Nice I think this is more clear.

micro-nit: can you wrap this at 80 columns and use ticks on `AccountOwner`?

buffalojoec:
So, technically the field that represents the controlling authority of a Token Account is called `owner`.

https://github.com/solana-labs/solana-program-library/blob/23916b24700cd222ad2397e9de8162ad6980e2e8/token/program/src/state.rs#L93

However, the `SetAuthority` command used to change it is called `AccountOwner`.

https://github.com/solana-labs/solana-program-library/blob/23916b24700cd222ad2397e9de8162ad6980e2e8/token/program/src/processor.rs#L441-L462

As you can see in that snippet, it will update `owner` based on that `set_authority` call.

So, although I think it's a bit confusing, `owner` is actually probably more accurate here.

Maybe we can reword this a bit and say:

```
Regardless of the creator, the new associated token account's [`owner`](https://github.com/solana-labs/solana-program-library/blob/23916b24700cd222ad2397e9de8162ad6980e2e8/token/program/src/state.rs#L92-L93) field will be the wallet, thus it will be controlled by the wallet, as if the wallet itself had created the associated token account.
```

micro-nit: can you make sure to wrap the columns at 80?

mikemaccana:
Got it, it sounds like the real issue is the code itself has two things called owner - account owner and program owner. It might be better to switch to more distinct terms in order to avoid confusion. 

> micro-nit: can you make sure to wrap the columns at 80?

Sure! Existing code is a [mix of hard and soft wrapped](https://github.com/solana-labs/solana-program-library/blob/23916b24700cd222ad2397e9de8162ad6980e2e8/docs/src/associated-token-account.md?plain=1#L24) and I didn't want to reformat unrelated code.

buffalojoec:
> Got it! Your suggestion is definitely clearer. It sounds like the real issue is the code itself has two things called owner - account owner and program owner. It might be even better to switch to more distinct terms (rather than reuse 'owner') in order to avoid confusion. I know that sounds like significant change but if there's going to be ten times as many Solana developers in the future, it may be a good one to make sooner rather than later.

Agreed, and to make matters worse we also call it `authority` in a lot of places, too. 💀

Issue #5937, opened 2023-12-01
Add remaining CLI functionality for SPL Token Group


buffalojoec:
#### Problem
As of #5911 and #5912, the SPL Token CLI supports adding `GroupPointer` and `GroupMemberPointer` to Token2022 mints. However, there's no CLI support for adding the actual `TokenGroup` or `TokenGroupMember` extensions to mints.

#### Solution
Add the necessary commands!

tonton-sol:
Can you expand on what is needed for this? Is this not taken care of by the `--enable-group` flag?

buffalojoec:
@tonton-sol All of the Token Group support should be present in the Token Client, but we don't have it in the CLI yet. If you take a look at Token Metadata for reference, we're basically just setting up clap commands for driving the Token Client and using those functions!

https://github.com/solana-labs/solana-program-library/blob/23916b24700cd222ad2397e9de8162ad6980e2e8/token/cli/src/command.rs#L514

The CLI only has the ability to create Group and Group Member pointers. You can pass an address or use `--enable-group` or `--enable-member` to set the pointer to the mint itself. You can't run `InitializeGroup` or the others from the interface.

https://github.com/solana-labs/solana-program-library/commit/ad86abe10f83f180fba0257451450182c875d04a

tonton-sol:
Understood, I didn't catch that. I'll take a look at it!

tonton-sol:
A few questions: 

1. `spl-token initialize-group` should only invoke  `token_group_initialize_with_rent_transfer()`? and requires the mint and size?
2. `spl-token initialize-member` should only invoke  `token_group_initialize_with_rent_transfer()` and requites the mint and group-mint?
3.  Should I also add the cli functionality for updating? Or should that be in a separate PR?

tonton-sol:
Also, I want to make sure I understand the spirit of the extensions based on reading the tests. You can create a group account then initialize it with a size. That size is the max number of members, say 10. You can then create a member, then initialize it with the group-mint. This gives 1 member. You can then use the member pointer extension, point to the first member mint and create the remaining 9. If you want to create another separate but related group of 10 members, you would use a group pointer. Then you would need to create and initialize a new member mint and initialize it, and repeat the process for minting the member pointers. Is this correct?

buffalojoec:
> Also, I want to make sure I understand the spirit of the extensions based on reading the tests. You can create a group account then initialize it with a size. That size is the max number of members, say 10.

[...]

PR #6135, opened 2024-01-16
[token-CLI] Add command_initialize_group to Token Cli

tonton-sol:
I have added the necessary clap commands and functions to drive `token_group_initialize_with_rent_transfer()` from the Token CLI.  I am running into error 0xc `InvalidInstruction` and cannot find what is causing this. Is there anything obvious that I cannot see? #5937

Thank you!

Edit: To recreate this error, run
```
spl-token --program-id TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb create-token --enable-group
```
then I initialize the group using the mint address created
```
spl-token initialize-group A6iv673L7PEZDxYDwQbMZsgmS5bGzExavADQSAcKA3Tt 3   
```


tonton-sol:
@buffalojoec please let me know what you think!

buffalojoec:
> I am running into error 0xc InvalidInstruction and cannot find what is causing this. Is there anything obvious that I cannot see?

Are you running this test with Token2022's feature flag enabled for `token-group`?

tonton-sol:
Yes I believe it is enabled. And the error comes before the token-group processor checks if it is enabled.

buffalojoec:
> Yes I believe it is enabled. And the error comes before the token-group processor checks if it is enabled.

And you're using the CLI from the repo? `target/debug/spl-token`?

tonton-sol:
Yes, and I can see the new clap subcommand I added.

buffalojoec:
The subcommand is just on the CLI side. We'll have to configure the tests to use Token2022's program `.so` file with the features enabled.

To test it manually, you can do:

```
cargo build-sbf --manifest-path token/program-2022/Cargo.toml --features token-group zk-ops
solana-test-validator --bpf-program TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb target/deploy/spl_token_2022.so
```

And then we'll just have to configure that same BPF program override in the test validator config of the tests. This might be outside the scope of this PR though.

If you manage to crack it, great! If not, we can wait on this and rebase after I add that support.

tonton-sol:
I had a fundamental misunderstanding lol. The deployed devnet token-22 program doesn't have the token-group enabled. I followed the steps for testing the cli and enabled it correctly. I also wrote a test that simply creates the mint and initializes it. The test passes. What else should I add to the test? Maybe check the max_size? Since the `GroupPointer` struct does not have a max_size field, would I need to read it directly from the raw data?

tonton-sol:
Actually I think this can be done with the `spl-token-group-interface`! Would it make sense to roll `command_update_max_size` into this pr? For example it seems the metadata test includes multiple commands.

buffalojoec:
> Actually I think this can be done with the `spl-token-group-interface`! Would it make sense to roll `command_update_max_size` into this pr? For example it seems the metadata test includes multiple commands.

Maybe just try fetching the mint and unpacking `TokenGroup` and checking the values for things like update authority/mint?

tonton-sol:
Okay I believe this is ready to go!

token/cli/src/clap_app.rs
```diff
@@ -962,6 +963,52 @@ pub fn app<'a, 'b>(
...
+                )
+                .arg(
+                        Arg::with_name("max_size")
+                        .validator(is_amount)
+                        .value_name("MAX_SIZE")
+                        .takes_value(true)
+                        .required(true)
+                        .index(2)
+                        .help("The number of members in the group."),
+                    )
```
buffalojoec, line 986: Just an idea here, but what if we allowed for the keyword `max` here, and we'd default that to `U32::MAX`?

buffalojoec (commented):
Looks good! See what you think on my "max size" suggestion, but besides that I'm ready to stamp this one.

token/cli/src/clap_app.rs
tonton-sol: I think to do this we would need to create a new validator in the solana-clap-utils like `is_amount_or_max`. I am also not sure how common of a use case that would be. I think its good as is 👍 

After this, I will work my way through the rest of the commands as separate PRs. 😎

buffalojoec (approved):
Thanks!
group hug

PR #6149, opened 2024-01-18
single-pool: remove minimum delegation check

2501babe:


single-pool/program/tests/initialize.rs
```diff
@@ -56,4 +61,56 @@ async fn fail_double_init() {
...
+    )
+    .await;
+
+    let instructions = instruction::initialize(
+        &id(),
+        &accounts.vote_account.pubkey(),
+        &context.payer.pubkey(),
+        &rent,
+        minimum_delegation - 1,
+    );
```
joncinque, line 100: Maybe this is already tested elsewhere, but rather than checking against `minimum_delegation - 1`, would it be more useful to test `LAMPORTS_PER_SOL - 1` and make sure that it fails in *both* cases? Then you're sure that the pool minimum is respected

joncinque (commented):
The program-level changes look good, and the tests are really great. Just one comment on a test. And also, since you're bumping up the package, can you include the Cargo.lock change?

single-pool/program/tests/initialize.rs
2501babe: oh `get_pool_minimum_delegation` is a helper that returns `max(stake_program_minimum_delegation, LAMPORTS_PER_SOL)`, using the constant would be clearer tho

joncinque: Oh right, oops, all good then!

PR #6150, opened 2024-01-19
token-cli: Add command_update_group_max_size to token CLI

tonton-sol:
As per #5937, added command_update_group_max_size to the token CLI. This allows one to update the max_size of a token group given the update_authority.

### DIFF
--- a/token/cli/src/clap_app.rs
+++ b/token/cli/src/clap_app.rs
@@ -1033,7 +1033,7 @@ pub fn app<'a, 'b>(
                     Arg::with_name("update_authority")
                         .long("update-authority")
                         .value_name("ADDRESS")
-                        .validator(is_valid_pubkey)
+                        .validator(is_valid_signer)
                         .takes_value(true)
                         .help(
                             "Specify the update authority address. \
--- a/token/cli/src/clap_app.rs
+++ b/token/cli/src/clap_app.rs
@@ -1032,7 +1032,7 @@ pub fn app<'a, 'b>(
                 .arg(
                     Arg::with_name("update_authority")
                         .long("update-authority")
-                        .value_name("ADDRESS")
+                        .value_name("SIGNER")
                         .validator(is_valid_signer)
                         .takes_value(true)
                         .help(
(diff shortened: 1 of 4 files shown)


token/cli/src/clap_app.rs
```diff
@@ -1007,6 +1008,39 @@ pub fn app<'a, 'b>(
                         ),
                 )
         )
+        .subcommand(
+            SubCommand::with_name(CommandName::UpdateGroupMaxSize.into())
+                .about("Updates the number of member in the group.")
```
buffalojoec, line 1013: "Updates the maximum number of members for a group"

token/cli/src/clap_app.rs
```diff
@@ -1007,6 +1008,39 @@ pub fn app<'a, 'b>(
...
+                    Arg::with_name("update_authority")
+                        .long("update-authority")
+                        .value_name("ADDRESS")
+                        .validator(is_valid_pubkey)
+                        .takes_value(true)
+                        .help(
+                            "Specify the update authority address. \
+                             Defaults to the client keypair address."
+                        ),
+                )
```
buffalojoec, line 1042: This should be an `authority` keypair instead, since the authority has to sign to be able to update a group!

https://github.com/solana-labs/solana-program-library/blob/0af879a4a09c03d208d1eacea8a48aafd932b2ea/token/cli/src/clap_app.rs#L952-L959

token/cli/tests/command.rs
```diff
@@ -3842,4 +3842,27 @@ async fn group(test_validator: &TestValidator, payer: &Keypair) {
...
+    )
+    .await
+    .unwrap();
+
+    let account = config.rpc_client.get_account(&mint).await.unwrap();
+
+    let updated_mint_state = StateWithExtensionsOwned::<Mint>::unpack(account.data).unwrap();
+
+    let updated_extension = updated_mint_state.get_extension::<TokenGroup>().unwrap();
+    assert_eq!(updated_extension.max_size, new_max_size.into());
```
buffalojoec, line 3867: Just a side note: you can't do this yet until we have the other instructions, but it would be good to be able to verify the CLI response message if you try to update the max size to a value below the number of current members. Just something to keep in mind when that goes in later!

tonton-sol: I agree, I can loop back to this when I am buttoning it all up.

buffalojoec (approved):
Nice! Thanks again!

PR #6153, opened 2024-01-19
stake-pool: Ceiling all fee calculations (HAL-01)

joncinque:
#### Problem

During deposit and withdrawal functions, if the SOL or pool token amount is small, it is possible the fee calculations floor to zero due to loss of precision, preventing fee collection.

#### Solution

This scenario is not economically viable, and the cost of fee avoidance would outweigh the benefit of depositing such a small amount to the stake pool, but it's better to fix things up.

The change takes the ceiling of all fee calculations by adding `denominator - 1` to the numerator.

### DIFF
--- a/stake-pool/program/src/state.rs
+++ b/stake-pool/program/src/state.rs
@@ -942,9 +942,13 @@ impl Fee {
         if self.denominator == 0 {
             return Some(0);
         }
-        (amt as u128)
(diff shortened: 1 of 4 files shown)


2501babe (approved):

2501babe:
seems straightforward!

PR #6155, opened 2024-01-19
bump transfer hook example

buffalojoec:
Marking this as a draft for now. I want to make sure this one is timed right.

Technically, since this depends on Token2022, we should really publish Token2022 first and then update this crate's version. Although I wanted to at least check out the bump since nowhere in the example or tests are we using the helpers from Token2022, only the interface.

`cargo publish --dry-run` checks out.

Lmk.


joncinque (approved):
Up to you! if it can publish properly, we can probably ship

PR #6164, opened 2024-01-23
token-js: add `COption<Pubkey>` support to `InitializeTransferFeeConfig` instruction

### LINKED ISSUE #6113: Unexpected behavior of createInitializeTransferFeeConfigInstruction
When using createInitializeTransferFeeConfigInstruction with transferFeeConfigAuthority set to null but with a fee and/or maxFee defined:

**Expected behavior**: Either the TX fails, because you're trying to set a fee and there's no relevant authority -- OR -- the fee is set, but cannot be further changed.

**Actual behavior**: The TX containing this instruction succeeds but fee and maxFee are set to 0. 

Tested on DevNet. Specifying *any* public key (e.g. with unknown private key) makes it correctly set the fee and maxFee.

According to the documentation here, transferFeeConfigAuthority should accept either a PublicKey or null:

https://solana-labs.github.io/solana-program-library/token/js/functions/createInitializeTransferFeeConfigInstruction.html

I've also found that setting maxFee to undefined sets it to zero. Not a bug -- the docs say it must be a BigInt. Maybe more intuitive if it returns an error if undefined?

buffalojoec:
The `InitializeTransferFeeConfig` instruction in Token2022's Rust crate expects a `COption<Pubkey>`, which denotes `None` with a single `0` and `Some(Pubkey)` with a `1` followed by the public key bytes.

The layout for this instruction in Token JS - `initializeTransferFeeConfigInstructionData` - was incorrectly using 32 all-zero bytes for `None`, as one might do with an `OptionalNonZeroPubkey`. This was causing incorrect instruction data when one of the two authorities was set to `null`.

Here I've introduced a `COptionPublicKeyLayout` serialization layout to use for this kind of case. It will handle the proper `COption<Pubkey>`. encoding and decoding, so we encode and decode instructions properly!

Closes #6113

### DIFF
--- a/token/js/test/e2e-2022/transferFee.test.ts
+++ b/token/js/test/e2e-2022/transferFee.test.ts
@@ -44,18 +44,19 @@ describe('transferFee', () => {
     let sourceAccount: PublicKey;
     let destinationAccount: PublicKey;
     let mint: PublicKey;
-    let transferFeeConfigAuthority: Keypair;
-    let withdrawWithheldAuthority: Keypair;
+    let mintAuthority: Keypair;
     before(async () => {
         connection = await getConnection();
         payer = await newAccountWithLamports(connection, 1000000000);
-        transferFeeConfigAuthority = Keypair.generate();
-        withdrawWithheldAuthority = Keypair.generate();
     });
-    beforeEach(async () => {
-        const mintAuthority = Keypair.generate();
+
+    async function setupTransferFeeMint(
+        transferFeeConfigAuthority: PublicKey | null,
+        withdrawWithheldAuthority: PublicKey | null
+    ) {
         const mintKeypair = Keypair.generate();
         mint = mintKeypair.publicKey;
+        mintAuthority = Keypair.generate();
         const mintLen = getMintLen(MINT_EXTENSIONS);
         const mintLamports = await connection.getMinimumBalanceForRentExemption(mintLen);
         const mintTransaction = new Transaction().add(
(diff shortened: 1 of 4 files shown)


token/js/test/e2e-2022/transferFee.test.ts
```diff
@@ -68,203 +69,281 @@ describe('transferFee', () => {
...
-            TEST_PROGRAM_ID
-        );
-        const mintInfo = await getMint(connection, mint, undefined, TEST_PROGRAM_ID);
-        const transferFeeConfig = getTransferFeeConfig(mintInfo);
-        expect(transferFeeConfig).to.not.be.null;
-        if (transferFeeConfig !== null) {
-            expect(transferFeeConfig.withdrawWithheldAuthority).to.eql(PublicKey.default);
-        }
+
+    describe('with null authorities', () => {
```
buffalojoec, line 297: The diff for the tests looks really funky, but I've simply just refactored the setup function to be re-used down here with the provided authorities, and then just added this whole `with null authorities` section.

token/js/test/unit/transferfee.test.ts
```diff
@@ -1,98 +1,167 @@
...
-        const transferFee = {
-            epoch: 1n,
-            maximumFee: 100n,
-            transferFeeBasisPoints: 0,
-        };
-        const preFeeAmount = 100n;
-        const fee = calculateFee(transferFee, preFeeAmount);
-        expect(fee).to.eql(0n);
+describe('transferFee', () => {
+    describe('encoding/decoding `InitializeTransferFeeConfig` instructions', () => {
```
buffalojoec, line 12: Test diff here looks awful as well, but here I've just added the `encoding/decoding` section above the existing tests.

joncinque (approved):
Looks great, thanks for fixing this up!

PR #6165, opened 2024-01-23
token-cli: Add command authorize support for TokenGroup 

tonton-sol:
Added support for changing the update authority for a Token group. Added a similar test used for updating the metadata authority. 

#5937

### DIFF
--- a/token/cli/src/clap_app.rs
+++ b/token/cli/src/clap_app.rs
@@ -191,6 +191,7 @@ pub enum CliAuthorityType {
     Metadata,
     GroupPointer,
     GroupMemberPointer,
+    TokenGroup,
 }
 impl TryFrom<CliAuthorityType> for AuthorityType {
     type Error = Error;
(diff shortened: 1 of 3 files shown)


token/cli/src/clap_app.rs
```diff
@@ -218,6 +219,9 @@ impl TryFrom<CliAuthorityType> for AuthorityType {
             }
             CliAuthorityType::GroupPointer => Ok(AuthorityType::GroupPointer),
             CliAuthorityType::GroupMemberPointer => Ok(AuthorityType::GroupMemberPointer),
+            CliAuthorityType::TokenGroup => {
+                Err("Token Group update authority does not map to a token authority type".into())
```
joncinque, line 223: nit: same here
```suggestion
                Err("Group update authority does not map to a token authority type".into())
```

token/cli/src/clap_app.rs
```diff
@@ -191,6 +191,7 @@ pub enum CliAuthorityType {
     Metadata,
     GroupPointer,
     GroupMemberPointer,
+    TokenGroup,
```
joncinque, line 194: nit: can you change this to just `Group`? We don't use the `Token` prefix anywhere else
```suggestion
    Group,
```

joncinque (commented):
Thanks for your contribution, looks great! Just one nit on the naming, then this should be good to go

token/cli/src/clap_app.rs
tonton-sol: My thought was to match the extension type, such as TokenMetadata and TokenGroup. But, I do see now that you use Metadata for the CliAuthorityType.

joncinque (approved):
Looks great, thanks for your contribution again! The CI failure is unrelated, so merging this on red

Issue #6170, opened 2024-01-23
Token JS: Use `COptionPubkeyLayout` for all `COption` fields


buffalojoec:
After #6164, we now have a serialization layout for `COption<Pubkey>`, where a `None` value is represented by a single `0`, while `Some(Pubkey)` is represented by a `1` plus the public key bytes.

There are a handful more instructions in Token2022 that use `COption<Pubkey>`. Here are a few examples:

https://github.com/solana-labs/solana-program-library/blob/8b22d759fef30795eff99385f15ae1c4380eeb90/token/program-2022/src/instruction.rs#L474-L482

https://github.com/solana-labs/solana-program-library/blob/8b22d759fef30795eff99385f15ae1c4380eeb90/token/program-2022/src/instruction.rs#L549-L553

Update their JavaScript instruction builders to use the new layout!

qiweiii:
would like to work on this!

PR #6173, opened 2024-01-24
fix: Use COptionPubkeyLayout for all COption fields

qiweiii:
resolve: #6170

### DIFF
--- a/token/js/src/instructions/initializeMint.ts
+++ b/token/js/src/instructions/initializeMint.ts
@@ -1,7 +1,7 @@
 import { struct, u8 } from '@solana/buffer-layout';
 import { publicKey } from '@solana/buffer-layout-utils';
-import type { AccountMeta } from '@solana/web3.js';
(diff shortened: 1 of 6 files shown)


buffalojoec (commented):
Looks great! Do you mind just adding some test coverage to the instructions so we can assert this works properly with `null` authorities?

You can just add a few more lines to some of the existing ones like `e2e-2022/closeMint.test.ts`, etc.

qiweiii:
@buffalojoec  I am not exactly sure how this should be tested, I added a test case, is that what we expect

buffalojoec:
> @buffalojoec I am not exactly sure how this should be tested, I added a test case, is that what we expect

It actually appears the `MintCloseAuthority` and `setAuthority` have some coverage for working with null authority values. I went ahead and added the coverage for `InitializeMint` and `InitializeMint2`!

buffalojoec (approved):
Thanks very much for this!

celebrate

PR #6178, opened 2024-01-25
[token-2022] allow split confidential transfer instruction constructors to be invoked on chain

samkim-crypto:


joncinque (approved):
Looks good! I took a look through the rest of the file, and is there a reason for *any* of the instruction creators be gated on `not(target_os = "solana")`? Since none of them are performing any encryption or decryption (that I can see at least), they all seem safe.

Some of them call into the `verify_*` on the proof program, which will fail, but we could likely relax that restriction in the future. What do you think?

samkim-crypto:
> Looks good! I took a look through the rest of the file, and is there a reason for _any_ of the instruction creators be gated on `not(target_os = "solana")`? Since none of them are performing any encryption or decryption (that I can see at least), they all seem safe.
> 
> Some of them call into the `verify_*` on the proof program, which will fail, but we could likely relax that restriction in the future. What do you think?

Oh yes, you are right. All of them take in the proof data as input; the proof is not generated inside those functions. It seems like we can remove it. Let me merge this PR and then do it in a follow-up since there might be something that we are overlooking.

PR #6182, opened 2024-01-25
Update solana dependency version to allow 2.0.0

willhickey:
There are transitive dependencies from the [solana monorepo](TODO) that lead to split dependencies when the monorepo version number is bumped to 2.0. For example:
```
solana-account-decoder v2.0.0 
├── solana-config-program v2.0.0 
│   ├── solana-program-runtime v2.0.0 
│   │   ├── solana-measure v2.0.0 
│   │   │   └── solana-sdk v2.0.0 
│   │   │       ├── solana-program v2.0.0       <--------
├── spl-token v4.0.0
│   ├── solana-program v1.17.14                 <--------
```

This leads to type incompatibilities.


Anchor.toml
```diff
@@ -1,6 +1,6 @@
 [toolchain]
 anchor_version = "0.29.0"
-solana_version = "1.17.13"
+solana_version = ">=1.17.13,<=2.0.0"
```
joncinque, line 3: Do you know if Anchor supports this syntax? I don't think it does, so unless you're sure that it works, we should probably keep this as is

account-compression/programs/noop/Cargo.toml
```diff
@@ -16,4 +16,4 @@ cpi = ["no-entrypoint"]
 default = []
 
 [dependencies]
-solana-program = "1.17.13"
+solana-program = ">=1.17.13,<=2.0.0"
```
joncinque, line 19: I did a little test, and specifying it in this way means that it will only ever pull in "2.0.0" and nothing higher. Would it make more sense to relax this further as:

```suggestion
solana-program = ">=1.17.13,<=2"
```

Then it can eventually pick up v2.1

joncinque (commented):
Looks good, just a couple of questions to make sure I understand the intention

account-compression/programs/noop/Cargo.toml
willhickey: Yeah good idea. I was thinking it should be as restrictive as possible but we can always tighten the bound once we introduce an incompatibility.

joncinque (approved):
Looks great, thanks for doing this!

PR #6183, opened 2024-01-26
docs: Add "Token Extensions" in the top-level doc

joncinque:
#### Problem

There's a disconnect between the "Token Extensions" and "Token-2022" names in the docs.

#### Solution

Call out that it's also called "Token Extensions"

### DIFF
--- a/docs/src/token-2022.md
+++ b/docs/src/token-2022.md
@@ -5,8 +5,8 @@ title: Token-2022 Program
 A token program on the Solana blockchain, defining a common implementation for
 fungible and non-fungible tokens.
 
-The Token-2022 Program is a superset of the functionality provided by the
-[Token Program](token.mdx).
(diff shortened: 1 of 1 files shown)


docs/src/token-2022.md
```diff
@@ -5,8 +5,8 @@ title: Token-2022 Program
 A token program on the Solana blockchain, defining a common implementation for
 fungible and non-fungible tokens.
 
-The Token-2022 Program is a superset of the functionality provided by the
-[Token Program](token.mdx).
+The Token-2022 Program, also known as Token Extensions,  is a superset of the
```
jnwng, line 8:
```suggestion
The Token-2022 Program, also known as Token Extensions, is a superset of the
```

jnwng (approved):

PR #6187, opened 2024-01-26
token-2022: Bump to v2

joncinque:
#### Problem

There's a breaking change in the master version of token-2022, but we need to publish new crates for the solana v2 bump in #6182

#### Solution

Bump token-2022 to a new major version so that the new crates can be used.

cc @willhickey


buffalojoec (dismissed):
v2

buffalojoec:
Just a side note here: We will also want to release a new Token Client since it will have the new transfer helper as well. I'll do JavaScript now.

PR #6190, opened 2024-01-27
[token-2022] Remove build target restriction on confidential transfer instruction constructors

samkim-crypto:
#### Problem
The confidential transfer instruction constructors do not work directly with cryptographic functions (create proofs, call the encrypt function, etc.). However, the compilation target for these functions forbid them from being used on chain.

#### Summary of changes
Remove the restriction on the compilation target for these constructor functions.


joncinque (approved):
Looks great, thanks!

PR #6191, opened 2024-01-27
reduce cluttering by removing debug print command

kagren:
The ValidatorList.decode method has a debug print command in it that clutters the output whenever I use that function. Assume it is there for debugging purposes so should be safe to remove.


joncinque (approved):
Definitely there for debugging, and safe to remove.

Thanks for contributing!

PR #6199, opened 2024-01-29
[token-2022] improve condition in Account::valid_account_data

fanatid:
We can change `>=` to `>` because already checked `==` on the previous step

### DIFF
--- a/token/program-2022/src/state.rs
+++ b/token/program-2022/src/state.rs
@@ -305,10 +305,7 @@ impl GenericTokenAccount for Account {
         account_data.len() == Account::LEN && is_initialized_account(account_data)
             || (account_data.len() > Account::LEN
                 && account_data.len() != Multisig::LEN
-                && ACCOUNTTYPE_ACCOUNT
-                    == *account_data
-                        .get(spl_token::state::Account::get_packed_len())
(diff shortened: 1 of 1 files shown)


token/program-2022/src/state.rs
```diff
@@ -303,7 +303,7 @@ impl GenericTokenAccount for Account {
     fn valid_account_data(account_data: &[u8]) -> bool {
         // Use spl_token::state::Account::valid_account_data once possible
         account_data.len() == Account::LEN && is_initialized_account(account_data)
-            || (account_data.len() >= Account::LEN
+            || (account_data.len() > Account::LEN
```
joncinque, line 306: I can't comment further down, but with this change, we can even simplify to:

```
ACCOUNTTYPE_ACCOUNT == *account_data[Account::LEN]
```

What do you think?

joncinque (commented):
Thanks for your contribution! Just a tiny comment that might make this even better, let me know what you think

token/program-2022/src/state.rs
fanatid: Yeah, should be safe to change :+1:

joncinque (approved):
Beautiful, thanks for the contribution!

PR #6205, opened 2024-01-30
Added support for the increase/decrease additional stake instructions in the Python client library

kagren:
- Updated the Python stake pool client library methods increase_validator_stake and decrease_validator_stake to support the increase/decrease additional validator stake instructions using the same logic as in the corresponding methods in the JS library.
- A bugfix in the JS stake pool client code, increaseValidatorStake function

### DIFF
--- a/stake-pool/py/stake_pool/instructions.py
+++ b/stake-pool/py/stake_pool/instructions.py
@@ -1170,7 +1170,6 @@ def decrease_additional_validator_stake(params: DecreaseAdditionalValidatorStake
             AccountMeta(pubkey=params.ephemeral_stake, is_signer=False, is_writable=True),
             AccountMeta(pubkey=params.transient_stake, is_signer=False, is_writable=True),
             AccountMeta(pubkey=params.clock_sysvar, is_signer=False, is_writable=False),
-            AccountMeta(pubkey=params.stake_history_sysvar, is_signer=False, is_writable=False),
             AccountMeta(pubkey=params.system_program_id, is_signer=False, is_writable=False),
             AccountMeta(pubkey=params.stake_program_id, is_signer=False, is_writable=False),
         ],
--- a/stake-pool/py/stake_pool/instructions.py
+++ b/stake-pool/py/stake_pool/instructions.py
@@ -580,12 +580,13 @@ class IncreaseAdditionalValidatorStakeParams(NamedTuple):
 
     # Params
     lamports: int
-    """Amount of lamports to split into the transient stake account."""
+    """Amount of lamports to increase on the given validator."""
     transient_stake_seed: int
     """Seed to used to create the transient stake account."""
-
     ephemeral_stake: PublicKey
+    """The ephemeral stake account used during the operation."""
     ephemeral_stake_seed: int
+    """The seed used to generate the ephemeral stake account"""
 
 
 class DecreaseAdditionalValidatorStakeParams(NamedTuple):
@@ -610,6 +611,8 @@ class DecreaseAdditionalValidatorStakeParams(NamedTuple):
     """`[]` Clock sysvar."""
     rent_sysvar: PublicKey
     """`[]` Rent sysvar."""
+    stake_history_sysvar: PublicKey
+    """'[]' Stake history sysvar."""
     system_program_id: PublicKey
     """`[]` System program."""
     stake_program_id: PublicKey
@@ -620,10 +623,12 @@ class DecreaseAdditionalValidatorStakeParams(NamedTuple):
     """Amount of lamports to split into the transient stake account."""
     transient_stake_seed: int
     """Seed to used to create the transient stake account."""
-
     reserve_stake: PublicKey
+    """The reserve stake account to move the stake to."""
     ephemeral_stake: PublicKey
+    """The ephemeral stake account used during the operation."""
     ephemeral_stake_seed: int
+    """The seed used to generate the ephemeral stake account"""
 
 
 class InstructionType(IntEnum):
--- a/stake-pool/py/stake_pool/instructions.py
+++ b/stake-pool/py/stake_pool/instructions.py
@@ -602,16 +602,14 @@ class DecreaseAdditionalValidatorStakeParams(NamedTuple):
     """`[]` Stake pool withdraw authority."""
     validator_list: PublicKey
     """`[w]` Validator stake list storage account."""
-    reserve_stake: PublicKey
-    """`[w]` Stake pool's reserve."""
     validator_stake: PublicKey
     """`[w]` Canonical stake to split from."""
     transient_stake: PublicKey
     """`[w]` Transient stake account to receive split."""
     clock_sysvar: PublicKey
     """`[]` Clock sysvar."""
-    stake_history_sysvar: PublicKey
-    """'[]' Stake history sysvar."""
+    rent_sysvar: PublicKey
+    """`[]` Rent sysvar."""
     system_program_id: PublicKey
     """`[]` System program."""
     stake_program_id: PublicKey
--- a/stake-pool/py/stake_pool/instructions.py
+++ b/stake-pool/py/stake_pool/instructions.py
@@ -542,6 +542,88 @@ class UpdateTokenMetadataParams(NamedTuple):
     uri: str
     """URI of the uploaded metadata of the spl-token."""
 
+class IncreaseAdditionalValidatorStakeParams(NamedTuple):
+    """(Staker only) Increase stake on a validator from the reserve account."""
+
+    # Accounts
+    program_id: PublicKey
+    """SPL Stake Pool program account."""
+    stake_pool: PublicKey
+    """`[]` Stake pool."""
+    staker: PublicKey
+    """`[s]` Staker."""
+    withdraw_authority: PublicKey
+    """`[]` Stake pool withdraw authority."""
+    validator_list: PublicKey
+    """`[w]` Validator stake list storage account."""
+    reserve_stake: PublicKey
+    """`[w]` Stake pool's reserve."""
+    transient_stake: PublicKey
+    """`[w]` Transient stake account to receive split."""
+    validator_stake: PublicKey
+    """`[]` Canonical stake account to check."""
+    validator_vote: PublicKey
+    """`[]` Validator vote account to delegate to."""
+    clock_sysvar: PublicKey
+    """`[]` Clock sysvar."""
+    rent_sysvar: PublicKey
+    """`[]` Rent sysvar."""
+    stake_history_sysvar: PublicKey
+    """'[]' Stake history sysvar."""
+    stake_config_sysvar: PublicKey
+    """'[]' Stake config sysvar."""
+    system_program_id: PublicKey
+    """`[]` System program."""
+    stake_program_id: PublicKey
+    """`[]` Stake program."""
+
+    # Params
+    lamports: int
+    """Amount of lamports to split into the transient stake account."""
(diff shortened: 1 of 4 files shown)


buffalojoec:
@kagren thanks!! Do you mind splitting the JS bugfix into a separate PR for tracking, since it's not related to this PR?

stake-pool/py/stake_pool/instructions.py
```diff
@@ -1037,6 +1156,37 @@ def decrease_validator_stake(params: DecreaseValidatorStakeParams) -> Transactio
...
+            AccountMeta(pubkey=params.stake_pool, is_signer=False, is_writable=False),
+            AccountMeta(pubkey=params.staker, is_signer=True, is_writable=False),
+            AccountMeta(pubkey=params.withdraw_authority, is_signer=False, is_writable=False),
+            AccountMeta(pubkey=params.validator_list, is_signer=False, is_writable=True),
+            AccountMeta(pubkey=params.reserve_stake, is_signer=False, is_writable=True),
+            AccountMeta(pubkey=params.validator_stake, is_signer=False, is_writable=True),
+            AccountMeta(pubkey=params.ephemeral_stake, is_signer=False, is_writable=True),
+            AccountMeta(pubkey=params.transient_stake, is_signer=False, is_writable=True),
+            AccountMeta(pubkey=params.clock_sysvar, is_signer=False, is_writable=False),
+            AccountMeta(pubkey=params.system_program_id, is_signer=False, is_writable=False),
```
joncinque, line 1173: This seems to be missing the stake history sysvar
```suggestion
            AccountMeta(pubkey=params.clock_sysvar, is_signer=False, is_writable=False),
            AccountMeta(pubkey=params.stake_history_sysvar, is_signer=False, is_writable=False),
            AccountMeta(pubkey=params.system_program_id, is_signer=False, is_writable=False),
```

stake-pool/js/src/index.ts
```diff
@@ -702,20 +702,22 @@ export async function increaseValidatorStake(
...
+        transientStakeSeed: transientStakeSeed.toNumber(),
+        withdrawAuthority,
+        transientStake,
+        validatorStake,
+        validatorVote,
+        lamports,
+        ephemeralStake,
+        ephemeralStakeSeed,
+      }),
+    );
```
joncinque, line 720: lol oops, thanks for noticing and fixing this. As @buffalojoec pointed out, can you move this fix to a separate PR?

joncinque (commented):
Thanks for your contribution! The changes mostly look good, just one missing account on the decrease instruction.

To make sure that this works, how about adding an optional ephemeral stake seed parameter to the increase and decrease actions `actions.py` in https://github.com/solana-labs/solana-program-library/blob/ab3908b0a5620abb3dd6f5c6d856019897ef14cc/stake-pool/py/stake_pool/actions.py#L482 -- when the parameter is included, the instruction becomes increase or decrease additional stake.

Then you can use that new parameter in the tests at https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/py/tests/test_a_time_sensitive.py -- that way we're sure that the new instruction bindings work.

What do you think?

kagren:
@joncinque sure I will create a separate PR for the JS bug fix, and will implement the changes as you suggest

joncinque:
Looks like you'll need to reformat the code too, see the output on `flake8`, and maybe `mypy` while you're at it:

```
check_dirs=(
  "bot"
  "spl_token"
  "stake"
  "stake_pool"
  "system"
  "tests"
  "vote"
)
flake8 "${check_dirs[@]}"
mypy "${check_dirs[@]}"
```

kagren:
Will fix the last tests tomorrow

kagren:
@joncinque need some help to understand why the DecreaseAdditionalValidatorStake instruction fails. I am not able to issue a call to DecreaseAdditionalValidatorStake, it always fails with "stake account with transient stake cannot be merged". I did some debugging and in the stake-pool program, in the process_decrease_validator_stake, when stake::merge is called, the transient account (transient_stake_account_info) has both effective and deactivating stake -- which is why the merge call fails as it is not possible to merge a stake account that has both an effective and either activating or deactivating amount.

Any guidance here is much appreciated!

stake-pool/py/tests/test_a_time_sensitive.py
```diff
@@ -51,21 +68,47 @@ async def test_increase_decrease_this_is_very_slow(async_client, validators, pay
...
+    resp = await async_client.get_account_info(validator_list_address, commitment=Confirmed)
+    data = resp['result']['value']['data']
+    validator_list = ValidatorList.decode(data[0], data[1])
+    for validator in validator_list.validators:
+        assert validator.transient_stake_lamports == decrease_amount // 2 + stake_rent_exemption
+        assert validator.active_stake_lamports == increase_amount-decrease_amount//2+minimum_amount+stake_rent_exemption
+
+    print("Waiting for epoch to roll over")
+    await waiter.wait_for_next_epoch(async_client)
+    await update_stake_pool(async_client, payer, stake_pool_address)
```
joncinque, line 94: Is this the part that should be removed in order to make sure that the decrease additional bit is actually happening?

joncinque (commented):
`DecreaseAdditionalValidatorStake` does work as advertised! It's possible to merge two activating or deactivating stakes. "Transient" stake in the stake program refers to a stake undergoing a multi-epoch activation or deactivation, which happens whenever more than 8% of total stake is activating or deactivating.

You can see the decrease Rust tests at https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/tests/decrease.rs and maybe see what's going wrong. From just reading the code, this looks correct once you remove the extra wait.

stake-pool/py/tests/test_a_time_sensitive.py
kagren: If I remove that then the call to DecreaseAdditionalValidatorStake fails with the following message:

'...Program Stake11111111111111111111111111111111111111 invoke [2]', 'Checking if destination stake is mergeable', 'stake account with transient stake cannot be merged ...'

I have also tried changing so that the deactivating stake is < 8% of the total stake, but still get the same error. Will spend some more time debugging to figure out what is going wrong (and use the Rust test as a reference).

bji:
I do not believe that the DecreaseAdditionalValidatorStake instruction works despite the fact that the unit test succeeds.  The stake program does not allow a stake account that has active stake and deactivating stake to be merged into; it's very clear about this both in comments and in the code of the MergeKind::get_if_mergeable function.

IncreaseAdditionalValidatorStake works because the stake pool's transient stake account after an increase instruction has zero active stake (since the transient stake account was created from reserve SOL as a new stake account) and activating stake, so it's fine.

But if the transient stake account were created by a Decrease instruction, then it will have active stake (having been split off of an active stake pool vote account) and deactivating stake (since the Decrease put the new stake account in a deactivating state).  According to the comments and logic of MergeKind::get_if_mergeable, this transient stake account can never be merged into, so subsequent DecreaseAdditionalValidatorStake instructions fail.

bji:
Oh another thing.  Is fixing the stake program in the cards?  I can't think of a reason that two deactivating stake accounts ought not to be mergeable into one, so even outside of the stake pool program, supporting this action (unless there is a reason I am not aware of that it shouldn't be allowed) could have benefits to other programs or use cases in addition to the stake pool program.

kagren:
> I am a bit embarrassed to admit that you are right, and that `DecreaseAdditionalValidatorStake` does not work. I forgot that we didn't fix the stake program to work properly when merging two deactivating stakes.
> 
> We could fix this in the stake pool program by reactivating the transient stake, merging the activated ephemeral stake, and then deactivating the merged whole. That feels pretty hacky though, and requires passing in the validator account to do the reactivation, which we don't even have during "decrease". Which means that we'd have to create a new instruction to do this correctly.
> 
> We could also modify the stake program to properly merge valid deactivating stakes, which adds one more special case to consider for merge. I don't love either solution, so suggestions are very welcome.
> 
> My apologies for misunderstanding the situation, I'll think about this some more on my side. I created #6275 to track it.
> 
> In the meantime, for this PR, feel free to remove the test for decrease-additional and add a comment saying that it doesn't work as intended

@joncinque done

joncinque:
> Oh another thing. Is fixing the stake program in the cards? I can't think of a reason that two deactivating stake accounts ought not to be mergeable into one

That would probably be the best option. Our plan right now is to convert the stake program to run as a BPF program, so once that's done we can consider making this change.

stake-pool/py/stake_pool/instructions.py
```diff
@@ -543,6 +543,94 @@ class UpdateTokenMetadataParams(NamedTuple):
...
+
+    # Params
+    lamports: int
+    """Amount of lamports to split into the transient stake account."""
+    transient_stake_seed: int
+    """Seed to used to create the transient stake account."""
+    reserve_stake: PublicKey
+    """The reserve stake account to move the stake to."""
+    ephemeral_stake: PublicKey
+    """The ephemeral stake account used during the operation."""
```
joncinque, line 629: nit: Same here, can you put these two pubkeys with the `Accounts` in the list?

stake-pool/py/stake_pool/instructions.py
```diff
@@ -543,6 +543,94 @@ class UpdateTokenMetadataParams(NamedTuple):
...
+    stake_program_id: PublicKey
+    """`[]` Stake program."""
+
+    # Params
+    lamports: int
+    """Amount of lamports to increase on the given validator."""
+    transient_stake_seed: int
+    """Seed to used to create the transient stake account."""
+    ephemeral_stake: PublicKey
+    """The ephemeral stake account used during the operation."""
```
joncinque, line 587: micro-nit: can you put this in the list with the rest of the `Accounts` at its correct location?

stake-pool/py/tests/test_a_time_sensitive.py
```diff
@@ -51,21 +68,31 @@ async def test_increase_decrease_this_is_very_slow(async_client, validators, pay
     for validator in validator_list.validators:
         assert validator.last_update_epoch != 0
         assert validator.transient_stake_lamports == 0
-        assert validator.active_stake_lamports == increase_amount + minimum_amount
+        assert validator.active_stake_lamports == increase_amount + minimum_amount + stake_rent_exemption
+
+    def decrease(validator, decrease_amount, ephemeral_stake_seed=None):
+        return decrease_validator_stake(async_client, payer, payer, stake_pool_address, validator,
```
joncinque, line 74: nit: I'm not sure I understand the need for this function -- can you remove it and just use `decrease_validator_stake` directly?

stake-pool/py/tests/test_a_time_sensitive.py
```diff
@@ -51,21 +68,31 @@ async def test_increase_decrease_this_is_very_slow(async_client, validators, pay
...
+                                        decrease_amount, ephemeral_stake_seed=ephemeral_stake_seed)
 
     # decrease from all
     futures = [
-        decrease_validator_stake(async_client, payer, payer, stake_pool_address, validator, decrease_amount)
+        decrease(validator, decrease_amount // 2)
```
joncinque, line 79: Can you revert this back to the full decrease amount?

joncinque (commented):
Just some final nits, then we can get this in, thanks for your patience!

kagren:
> Just some final nits, then we can get this in, thanks for your patience!

Done @joncinque

joncinque (approved):
Thanks for your contribution!

thanks

Issue #5884, opened 2023-11-22
token-2022: Consider prohibiting creating non-transferable accounts *without* immutable ownership


joncinque:
#### Problem

It's possible to create a token account on a mint with the non-transferable extension *without* immutable ownership enabled. Immutable ownership is enforced during *minting*, so it's impossible to mint a non-transferable token to an account without immutable ownership.

This can be confusing, because it seems like it might be possible to have non-transferable tokens in an account with transferable ownership.

#### Solution

Consider throwing an error when trying to initialize a token account on a non-transferable mint *without* immutable ownership.

buffalojoec:
So, if I'm understanding this correctly:
- In order to mint tokens to a non-transferable account, that destination must have an immutable owner
- In order to transfer tokens from a non-transferable account, you can't

What would be the reason to allow creating a non-immutable owner token account on a non-transferable mint? I would say this only makes sense if the non-transferable extension can be removed, which at the moment it does not appear is the case.

joncinque:
> What would be the reason to allow creating a non-immutable owner token account on a non-transferable mint?

I can't remember why we did it that way. It looks like it was just done like that in https://github.com/solana-labs/solana-program-library/pull/2912 without any other questions.

> In order to transfer tokens from a non-transferable account, you can't

Also, lol

quminghaonanya:
Can I try this issue? I'm going to

quminghaonanya:
Can I have a try on this issue,
I will return the NonTransferableNeedsImmutableOwnership error when trying to initialize an account on a non-transferable mint without immutable ownership.

joncinque:
Yep that's the idea, and no need to ask for permission, go for it! We'll review your PR when it comes in

joncinque:
Fixed in #6206

PR #6206, opened 2024-01-31
token-2022: initialize an account on a non-transferable mint without immutable ownership

quminghaonanya:
This PR improves the token-2022 account initialisation addressed by issue (#5884).
when trying to initialize an account on a non-transferable mint,  Now It will check whether the account has immutable ownership:
1. if yes, pass the check.
2. if no, return NonTransferableNeedsImmutableOwnership error.

### DIFF
--- a/token/program-2022-test/tests/non_transferable.rs
+++ b/token/program-2022-test/tests/non_transferable.rs
@@ -10,7 +10,10 @@ use {
     },
     spl_token_2022::{
         error::TokenError,
-        extension::{transfer_fee::TransferFee, ExtensionType},
+        extension::{
+            immutable_owner::ImmutableOwner, transfer_fee::TransferFee, BaseStateWithExtensions,
+            ExtensionType,
+        },
     },
     spl_token_client::token::{ExtensionInitializationParams, TokenError as TokenClientError},
 };
@@ -40,6 +43,14 @@ async fn transfer() {
         .unwrap();
     let alice_account = alice.pubkey();
 
+    // immutable ownership is added to alice's account during initialization
+    token
+        .get_account_info(&alice_account)
+        .await
+        .unwrap()
+        .get_extension::<ImmutableOwner>()
+        .unwrap();
+
     token
         .create_auxiliary_token_account_with_extension_space(
             &bob,
@@ -50,8 +61,8 @@ async fn transfer() {
         .unwrap();
     let bob_account = bob.pubkey();
 
-    // mint fails because the account does not have immutable ownership
-    let error = token
+    // mint to alice should be successful
+    token
         .mint_to(
             &alice_account,
             &mint_authority.pubkey(),
@@ -59,19 +70,8 @@ async fn transfer() {
             &[&mint_authority],
         )
         .await
-        .unwrap_err();
+        .unwrap();
 
-    assert_eq!(
-        error,
-        TokenClientError::Client(Box::new(TransportError::TransactionError(
-            TransactionError::InstructionError(
-                0,
-                InstructionError::Custom(TokenError::NonTransferableNeedsImmutableOwnership as u32)
-            )
-        )))
-    );
-
-    // mint succeeds if the account has immutable ownership
     token
         .mint_to(
             &bob_account,
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -812,6 +812,9 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
             ExtensionType::TransferFeeAmount => {
                 self.init_extension::<TransferFeeAmount>(true).map(|_| ())
             }
+            ExtensionType::ImmutableOwner => {
+                self.init_extension::<ImmutableOwner>(true).map(|_| ())
+            }
             ExtensionType::NonTransferableAccount => self
                 .init_extension::<NonTransferableAccount>(true)
                 .map(|_| ()),
@@ -1133,6 +1136,7 @@ impl ExtensionType {
                 }
                 ExtensionType::NonTransferable => {
                     account_extension_types.push(ExtensionType::NonTransferableAccount);
+                    account_extension_types.push(ExtensionType::ImmutableOwner);
                 }
                 ExtensionType::TransferHook => {
                     account_extension_types.push(ExtensionType::TransferHookAccount);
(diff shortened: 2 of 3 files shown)


token/program-2022/src/processor.rs
```diff
@@ -7309,6 +7316,54 @@ mod tests {
...
+        let mut mint_account = SolanaAccount::new(Rent::default().minimum_balance(mint_len),
+                               mint_len,
+                               &program_id);
+        let mut rent_sysvar = rent_sysvar();
+
+        // create a non-transferable mint
+        assert_eq!(
+            Ok(()),
+            do_process_instruction(initialize_non_transferable_mint(&program_id, &mint_key).unwrap(),
+            vec![&mut mint_account, &mut rent_sysvar],
```
joncinque, line 7343: The rent sysvar isn't required for this instruction, it only needs the mint
```suggestion
            vec![&mut mint_account],
```

joncinque (commented):
The change and test look great! Just one little nit.

To get this through CI, you'll need to run `./cargo-nightly fmt --all` at the root of the repo, to format the code consistently.

After that, you should also run `./cargo-nightly clippy -Zunstable-options --all-targets` to be sure there aren't any clippy errors.

Finally, there will be a failure in the existing non-transferable test since it's testing the behavior of minting to an account without immutable ownership at https://github.com/solana-labs/solana-program-library/blob/f62f8cf94c829545fff03a0b42fd08013ed83a4c/token/program-2022-test/tests/non_transferable.rs#L53-L72 -- you can just remove that check

token/program-2022/src/processor.rs
quminghaonanya: removed

quminghaonanya: Thank you for the suggestions.
as for the lint check, after I ran the ./cargo-nightly clippy -Zunstable-options --all-targets command, It showed  2 errors that refer to a commented-out code, I supposed this would not impact the  CI.

quminghaonanya: Seem like I broke more than one test, Will take a look later :)

joncinque: Ah yes, a few tests depend on the old behavior unfortunately, so these are legitimate test failures. Let me know if you're having a hard time with these and I can help out

quminghaonanya: I've fixed all the failed tests in the token-2022, could u help take a look over the changes and check whether it looks good, thank you

token/program-2022/src/extension/mod.rs
```diff
@@ -1133,6 +1136,7 @@ impl ExtensionType {
                 }
                 ExtensionType::NonTransferable => {
                     account_extension_types.push(ExtensionType::NonTransferableAccount);
+                    account_extension_types.push(ExtensionType::ImmutableOwner);
```
joncinque, line 1139: The changes here are a bit invasive, since it's going further than checking the extensions at init, and rather adding the immutable ownership extension by force for all non-transferable accounts. This should be fine, but let me think about it some more.

joncinque (commented):
I think the change is ok, but let me sleep on it to make sure

joncinque (commented):
Ok, I'm convinced that this is ok! To properly cover the new behavior, let's update the tests and make a few more little changes

token/program-2022/src/processor.rs
```diff
@@ -157,6 +157,13 @@ impl Processor {
         {
             msg!("Warning: Mint has a permanent delegate, so tokens in this account may be seized at any time");
         }
+
+        if mint.get_extension::<NonTransferable>().is_ok()
+            && account.get_extension::<ImmutableOwner>().is_err()
+        {
+            return Err(TokenError::NonTransferableNeedsImmutableOwnership.into());
+        }
```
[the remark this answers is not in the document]
quminghaonanya: Yes, I agree. 
But if we remove the check, we will return the InvalidAccountData error if the account doesn't provide enough space to store the ImmutableOwnership extension while initializing a non-transferable account, 
For this scenario, I think the InvalidAccountData error is less explicit than the NonTransferableNeedsImmutableOwnership error. 
Is it necessary to give a more specific error?

token/program-2022/src/extension/mod.rs
```diff
@@ -825,6 +825,9 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
             ExtensionType::AccountPaddingTest => {
                 self.init_extension::<AccountPaddingTest>(true).map(|_| ())
             }
+            ExtensionType::ImmutableOwner => {
+                self.init_extension::<ImmutableOwner>(true).map(|_| ())
+            }
```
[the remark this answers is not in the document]
quminghaonanya: Sure, I will reorder this code.

token/program-2022/src/processor.rs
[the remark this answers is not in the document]
joncinque: I would say that since we're changing the behavior to automatically initialize the immutable owner extension, the new error is actually less clear. The true error is now not having a big enough account. So I think that behavior is fine

token/program-2022-test/tests/non_transferable.rs
```diff
@@ -50,27 +50,6 @@ async fn transfer() {
...
-
-    assert_eq!(
-        error,
-        TokenClientError::Client(Box::new(TransportError::TransactionError(
-            TransactionError::InstructionError(
-                0,
-                InstructionError::Custom(TokenError::NonTransferableNeedsImmutableOwnership as u32)
-            )
-        )))
-    );
```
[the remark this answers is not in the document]
quminghaonanya: done

token/program-2022/src/processor.rs
```diff
@@ -7309,6 +7316,60 @@ mod tests {
         assert_eq!(account_account, account3_account);
     }
 
+    #[test]
+    fn initialize_account_on_non_transferable_mint() {
+        let program_id = crate::id();
+        let account = Pubkey::new_unique();
+        let mut account_account = SolanaAccount::new(
+            account_minimum_balance(),
+            Account::get_packed_len(),
```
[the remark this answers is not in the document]
quminghaonanya: done

joncinque (approved):
Looks really great, the change is so elegant in the end!

thanks

PR #6219, opened 2024-02-03
fix: correcting import in transfer hook js example

dustinsanders:
Correcting `transferCheckedWithHook` to `transferCheckedWithTransferHook` in `/examples/transferHook.ts`


joncinque (approved):

PR #6221, opened 2024-02-04
Bump SPL crate versions

willhickey:
Bump crate versions in preparation for tagging new releases so the monorepo dependencies can be updated to SPL versions that will use 2.0.0 for transitive dependencies.  [Discord discussion](https://discord.com/channels/428295358100013066/910937142182682656/1200210293112979626)

### DIFF
--- a/libraries/discriminator/Cargo.toml
+++ b/libraries/discriminator/Cargo.toml
@@ -1,5 +1,3 @@
-
-
 [package]
 name = "spl-discriminator"
 version = "0.1.1"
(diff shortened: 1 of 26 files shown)


libraries/discriminator/Cargo.toml
```diff
@@ -1,6 +1,8 @@
+
+
```
joncinque, line 2: Is this whitespace intentional? If not, can you remove it?

joncinque (approved):
Just one little comment, feel free to merge after addressing it. Thanks for taking care of this!

PR #6224, opened 2024-02-05
Bugfix JS client stake pool code increaseValidatorStake and decreaseValidatorStake

kagren:
When the increase/decrease additional stake instructions are used, the transient stake seed needs to remain the same as in the previous call to increase/decrease stake.


joncinque (commented):
Looks good to me! You'll just need to run `pnpm lint:fix` to get the formatting correct and commit up those changes, and then we can get this in

joncinque (approved):
Perfect, thanks for the contribution!

PR #6226, opened 2024-02-06
Replace BN.toBuffer with BN.toArrayLike

kagren:
BN.toBuffer is not available in the browser. Replaced with .toArrayLike which provides the same functionality and works consistently both in nodejs and browser.


joncinque (approved):
Beautiful, thanks for fixing this!

PR #6237, opened 2024-02-08
fix dependency update for new version

2501babe:
the regex in `update-solana-dependencies.sh` doesnt match the range dependency we current have, so running the `patch-crates.io` appears to succeed but when you build a project it doesnt use the local dependency

this changes it from "match maybe an equals sign then a single old version number, replace with the new version number preceded by whatever preceded the old one" to "match maybe eq/gt/lt/gte/lte then an old version number then whatever other nonsense comes until the end quote, replace with equals the new version number"

this still assumes if `Cargo.toml` has a version `>=1.2.3,<4.5.6` then `solana-version.sh` contains `1.2.3`

### DIFF
--- a/update-solana-dependencies.sh
+++ b/update-solana-dependencies.sh
@@ -84,6 +84,6 @@ crates=(
 
 set -x
 for crate in "${crates[@]}"; do
-  sed -E -i'' -e "s:(${crate} = \")(=?)${old_solana_ver}\".*:\1\2${solana_ver}\":" "${tomls[@]}"
-  sed -E -i'' -e "s:(${crate} = \{ version = \")(=?)${old_solana_ver}(\".*):\1\2${solana_ver}\3:" "${tomls[@]}"
+  sed -E -i'' -e "s:(${crate} = \")[=<>]*${old_solana_ver}[^\"]*\".*:\1=${solana_ver}\":" "${tomls[@]}"
+  sed -E -i'' -e "s:(${crate} = \{ version = \")[=<>]*${old_solana_ver}[^\"]*(\".*):\1=${solana_ver}\2:" "${tomls[@]}"
 done
--- a/update-solana-dependencies.sh
+++ b/update-solana-dependencies.sh
@@ -84,6 +84,6 @@ crates=(
 
 set -x
 for crate in "${crates[@]}"; do
-  sed -E -i'' -e "s:(${crate} = \")[=<>]*${old_solana_ver}[^\"]*\".*:\1=${solana_ver}\":" "${tomls[@]}"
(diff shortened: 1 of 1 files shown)


update-solana-dependencies.sh
```diff
@@ -84,6 +84,6 @@ crates=(
 
 set -x
 for crate in "${crates[@]}"; do
-  sed -E -i'' -e "s:(${crate} = \")(=?)${old_solana_ver}\".*:\1\2${solana_ver}\":" "${tomls[@]}"
-  sed -E -i'' -e "s:(${crate} = \{ version = \")(=?)${old_solana_ver}(\".*):\1\2${solana_ver}\3:" "${tomls[@]}"
+  sed -E -i'' -e "s:(${crate} = \")[=<>]*${old_solana_ver}[^\"]*\".*:\1=${solana_ver}\":" "${tomls[@]}"
+  sed -E -i'' -e "s:(${crate} = \{ version = \")[=<>]*${old_solana_ver}[^\"]*(\".*):\1=${solana_ver}\2:" "${tomls[@]}"
```
joncinque, line 88: This script is also used whenever we update to a new version of the monorepo, so we'll need to retain the grouping. I tried this out and it seemed to work -- what do you think?
```suggestion
  sed -E -i'' -e "s:(${crate} = \")([=<>]*)${old_solana_ver}([^\"]*)\".*:\1\2${solana_ver}\3\":" "${tomls[@]}"
  sed -E -i'' -e "s:(${crate} = \{ version = \")([=<>]*)${old_solana_ver}([^\"]*)(\".*):\1\2${solana_ver}\3\4:" "${tomls[@]}"
```

joncinque (commented):
Thanks for pointing this out, I totally forgot about it.

update-solana-dependencies.sh
2501babe: oh i was only thinking about its use for test environment setup. is this the behavior we want in that case? eg if we go from 1.17.17 to 2.0.0 this would change `>=1.17.17,<=2` to `>=2.0.0,<=2`

joncinque: Oh right, that's a good point... we'll still need to do a few upgrades for the 1.18 release line, so we definitely want to keep supporting that behavior for a bit of time in this script.

I did a test with this change and a monorepo on 1.18, and the SPL build uses the correct crates. So even though it'll look silly for some time when the monorepo goes to 2.0.0, it'll maintain the behavior we want.

As an alternative, we could add another param to this script to force the solana dependencies to `=${solana_ver}`, but I don't think it's necessary. What do you think?

2501babe: ok yea we can use your version and just change the script again later. i dont have a usecase for forcing =, its not necessary, it just didnt occur to me we use the script for committed version updates so keeping the prefix that was there is correct behavior

PR #6239, opened 2024-02-09
Update token dependencies

willhickey:
In preparation for bumping the monorepo version to 2.0.0, relax some of the internal dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6650,7 +6650,7 @@ dependencies = [
  "num-traits",
  "solana-program",
  "spl-token 4.0.1",
- "spl-token-2022 2.0.0",
+ "spl-token-2022 2.0.1",
  "thiserror",
 ]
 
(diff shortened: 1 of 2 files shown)


CriesofCarrots (approved):
lgtm. I think we also need to bump this crate to v2.0.1 for release, right?

CriesofCarrots:
I'm actually surprised we haven't run into this before. @joncinque , have you seen any conflict issues trying to publish spl-token-2022 in the past? The token-group-interface and token-metadata-interface dependencies are probably pretty new to the monorepo, but memo has been in there forever.

CriesofCarrots (approved):
Wouldn't hurt to get @joncinque 's sanity check, if he's around

joncinque:
Sorry for the late response, but it looks good!

And I think we got lucky beforehand because cargo does support multiple concurrent major versions in some situations. So it's valid for token-2022 to use memo v4 and the monorepo to use memo v3, but it's not valid for monorepo to require `=4.0.0` but for token-2022 to use `4.0.1`

CriesofCarrots:
> And I think we got lucky beforehand because cargo does support multiple concurrent major versions in some situations. So it's valid for token-2022 to use memo v4 and the monorepo to use memo v3, but it's not valid for monorepo to require `=4.0.0` but for token-2022 to use `4.0.1`

Oh, cargo, you make no sense to my feeble human brain sometimes

PR #6277, opened 2024-02-20
get 'map_transaction_error' return value consistent

yihau:
the details in the code comment 🙈


SebastianBor (approved):

joncinque:
@yihau will the sdk change break all downstream users of this error variant? If so, we should highlight it in the changelog. I can put in a PR for that

PR #6278, opened 2024-02-20
Upgrade to Solana 1.18.2

samkim-crypto:
- Upgraded solana version to 1.18.2
- Upgraded rust version to 1.76.0 in line with solana 1.18.2
- Updated borsh to 1.2.1 in line with solana 1.18.2
- The borsh function `try_to_vec` was deprecated with version 1.x, so I replaced this function with `borsh::to_vec` throughout the repo
- The solana clap-v3 utils deprecated some validator functions, which causes warnings in some cli tools in the repo. Removing the deprecated functions from the existing cli tools will require more involved changes, so I decided to add `allow![(deprecated)]` on these tools and defer resolving the deprecations to follow-up PRs.
  - https://github.com/solana-labs/solana-program-library/issues/6305
  - https://github.com/solana-labs/solana-program-library/issues/6306
  - https://github.com/solana-labs/solana-program-library/issues/6307
- Enums with `repr(C)` in solana version 1.17 and 1.18 seem to have different representations in memory (https://discord.com/channels/428295358100013066/967028962746327060/1210701837007724564) causing errors in some `token-lending` tests. I updated the enums in token-lending to `repr(u32)`, which should have the same memory representation as in `repr(C)` with solana 1.17.


token-lending/program/src/pyth.rs
joncinque, line 1: Just so I'm sure I understand -- with 1.17, it was implicitly `repr(u32)`, correct? If so, can we do that instead of `repr(u8)`? I'm worried of some strange issue when trying to deserialize the pyth types otherwise

joncinque (commented):
Just one little question on the change to the enums in token-lending, but the rest looks great!

joncinque (approved):

Issue #6282, opened 2024-02-22
small consistency refactor needed


MohammedAlabd:
https://github.com/solana-labs/solana-program-library/blob/6002a1d2617a9b9ab783093ef52c342b6c278cdf/token-metadata/js/src/state.ts#L42C1-L42C18

this should be 
```ts
export function  pack(meta: TokenMetadata): Uint8Array {
```

not 

```ts
export const  pack = (meta: TokenMetadata): Uint8Array => {
```

this is how it is being done in other places such as the in [instruction.ts](https://github.com/solana-labs/solana-program-library/blob/master/token-metadata/js/src/instruction.ts)

joncinque:
It's a good point, we should probably enforce it directly in the linter

FaybianB:
@MohammedAlabd @joncinque Hello, may I be assigned this issue?

joncinque:
@FaybianB I won't assign the issue, but feel free to put in a PR!

FaybianB:
@joncinque Done!

MohammedAlabd:
there are some small stuff in other places as well

would be cool if one PR could address them all

joncinque:
> there are some small stuff in other places as well

PRs are always welcome! It would probably be easiest to do this at the eslint level with `allowArrowFunctions` set to false https://eslint.org/docs/latest/rules/func-style#allowarrowfunctions

FaybianB:
I can address this in the same PR.

> > there are some small stuff in other places as well
> 
> PRs are always welcome! It would probably be easiest to do this at the eslint level with `allowArrowFunctions` set to false https://eslint.org/docs/latest/rules/func-style#allowarrowfunctions

PR #6283, opened 2024-02-23
Refactor Function Declaration

FaybianB:
Refactors function declaration to use named function declaration syntax instead of arrow function declaration system to remain consistent.

Resolves: #6282


joncinque (commented):
This is failing CI at the moment because you'll need to remove the semi-colon on line 50 -- you can run `pnpm lint:fix` to pass the checks

joncinque (approved):
Thanks for your contribution!

PR #6284, opened 2024-02-23
chore: update to new-ish bubblegum repo

jim4067:


joncinque (approved):
Thanks for fixing this!

PR #6289, opened 2024-02-23
type-length-value-js: add docs configs

buffalojoec:
Simply adds typedoc configs to Type Length Value, adds the necessary `devDependencies`, and sorts the scripts alphabetically.


joncinque (approved):
sort

Issue #6175, opened 2024-01-24
token-js: Add TokenGroup support


joncinque:
#### Problem

The TokenGroup, TokenGroupMember, GroupPointer, and GroupMemberPointer extensions are all in the token-2022 program, and the two pointer extensions are live on all networks, but they aren't supported in `@solana/spl-token`.

The `TokenGroup` and `TokenGroupMember` have a JS library at https://github.com/solana-labs/solana-program-library/tree/master/token-group/js

#### Proposed Solution

Similar to #5667, add support for these extensions by incorporating `@solana/spl-token-group` into `@solana/spl-token`

buffalojoec:
The pointers have been added! Now we just need support for the Token Group interface, similarly to Token Metadata, by making use of the existing `@solana/spl-token-group` package!

qiweiii:
I can add that!

ErwanTouba:
Hello there, i'm having an issue and after googling my error i landed here.
Might be of interest for you:

    "@solana/spl-token": "^0.4.2",
    "@solana/web3.js": "^1.91.1",




`✔ Vite server hmr 9 files in 19.54ms

 WARN  [Vue Router warn]: uncaught error during route navigation:


 ERROR  The requested module '@solana/spl-token-group' does not provide an export named 'TOKEN_GROUP_MEMBER_SIZE'

  TOKEN_GROUP_MEMBER_SIZE,
  ^
  
  SyntaxError: The requested module '@solana/spl-token-group' does not provide an export named 'TOKEN_GROUP_MEMBER_SIZE'
  at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
  at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)


 WARN  [Vue Router warn]: uncaught error during route navigation:


 ERROR  The requested module '@solana/spl-token-group' does not provide an export named 'TOKEN_GROUP_MEMBER_SIZE'

  TOKEN_GROUP_MEMBER_SIZE,
  ^
  
  SyntaxError: The requested module '@solana/spl-token-group' does not provide an export named 'TOKEN_GROUP_MEMBER_SIZE'
  at ModuleJob._instantiate (node:internal/modules/esm/module_job:124:21)
  at async ModuleJob.run (node:internal/modules/esm/module_job:190:5)


 WARN  Cannot stringify arbitrary non-POJOs SyntaxError

[...]

PR #6295, opened 2024-02-25
feat: add token group extension in token

qiweiii:
As mentioned in #6175, add support for token group extension to `@solana/spl-token`

Todo
- [x] e2e tests
- [x] unit tests


single-pool/js/packages/modern/.gitignore
buffalojoec, line 1: What's this little guy doing in here? 🙃

token/js/src/extensions/groupMember/actions.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param confirmOptions         Options for confirming the transaction
+ * @param programId              SPL Token program account
+ *
+ * @return Signature of the confirmed transaction
+ */
+export async function tokenGroupMemberInitialize(
+    connection: Connection,
+    payer: Signer,
+    mint: PublicKey,
+    memberMint: PublicKey,
```
buffalojoec, line 31: This should be `member` and `mint`. Looks like you have two mints here.

token/js/src/extensions/groupMember/actions.ts
```diff
@@ -0,0 +1,53 @@
...
+    confirmOptions?: ConfirmOptions,
+    programId = TOKEN_2022_PROGRAM_ID
+): Promise<TransactionSignature> {
+    const [memberMintAuthorityPublicKey, signers] = getSigners(memberMintAuthority, multiSigners);
+
+    const transaction = new Transaction().add(
+        createInitializeMemberInstruction({
+            programId,
+            member: mint,
+            memberMint,
```
buffalojoec, line 45: Same comment as above!

token/js/src/extensions/group/state.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { Mint } from '../../state/mint.js';
+import { ExtensionType, getExtensionData } from '../extensionType.js';
+
+/** Buffer layout for de/serializing a TokenGroup extension */
+export const TokenGroupLayout = struct<TokenGroup>([
+    publicKey('updateAuthority'),
+    publicKey('mint'),
+    u32('size'),
+    u32('maxSize'),
+]);
```
buffalojoec, line 14: You don't need to redefine a layout here! The interface JS library supports serialization.

https://github.com/solana-labs/solana-program-library/blob/7e022ac1b2ac527d6607712521685e64d0ff50f8/token-group/js/src/state/tokenGroup.ts#L46-L61

token/js/src/extensions/groupMember/state.ts
```diff
@@ -0,0 +1,30 @@
...
+import type { TokenGroupMember } from '@solana/spl-token-group';
+import type { Mint } from '../../state/mint.js';
+import { ExtensionType, getExtensionData } from '../extensionType.js';
+
+/** Buffer layout for de/serializing a TokenGroupMember extension */
+export const TokenGroupMemberLayout = struct<TokenGroupMember>([
+    publicKey('mint'),
+    publicKey('group'),
+    u32('memberNumber'),
+]);
```
buffalojoec, line 13: Same with this layout!
https://github.com/solana-labs/solana-program-library/blob/7e022ac1b2ac527d6607712521685e64d0ff50f8/token-group/js/src/state/tokenGroupMember.ts#L30-L37

token/js/src/extensions/index.ts
```diff
@@ -2,6 +2,8 @@ export * from './accountType.js';
 export * from './cpiGuard/index.js';
 export * from './defaultAccountState/index.js';
 export * from './extensionType.js';
+export * from './group/index.js';
+export * from './groupMember/index.js';
```
buffalojoec, line 6: What do you think about just combining the two directories `group` and `groupMember` into one directory `tokenGroup`? That way everything for the Token Group interface is all in one place.

token/js/src/extensions/group/actions.ts
```diff
@@ -0,0 +1,130 @@
...
+ * @param mintAuthority    Mint Authority
+ * @param updateAuthority  Update Authority
+ * @param maxSize          Maximum number of members in the group
+ * @param multiSigners     Signing accounts if `authority` is a multisig
+ * @param confirmOptions   Options for confirming the transaction
+ * @param programId        SPL Token program account
+ *
+ * @return Signature of the confirmed transaction
+ */
+export async function tokenGroupInitializeGroup(
```
buffalojoec, line 29: Can we also get some instructions for "with rent transfer"? Even though `TokenGroup` and `TokenGroupMember` are fixed-size extensions, they are initialized after the mint is initialized, so it needs to be reallocated and thus requires some additional lamports. Updating the max size and authority has no rent effect.

token/js/test/e2e-2022/group.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+        const mintInfo = await getMint(connection, mint.publicKey, undefined, TEST_PROGRAM_ID);
+        const group = getTokenGroupState(mintInfo);
+        expect(group).to.deep.equal({
+            updateAuthority: updateAuthority.publicKey,
+            mint: mint.publicKey,
+            size: 0,
+            maxSize: 10,
+        });
+    });
```
buffalojoec, line 115: I think you could just compare this to the `tokenGroup` constant you set above.

buffalojoec (commented):
Looking great!

I left a few suggestions and revisions, and it would be great to have a unit test similar to that of `TokenMetadata` for creating groups and members!

Thanks for contributing! 🔥

token/js/src/extensions/index.ts
qiweiii: Sure, I put them in different dir coz I saw these are 2 extensions

single-pool/js/packages/modern/.gitignore
qiweiii: When i run build, this package create files inside `dist`, lots of changed files need to exclude before push, so I added this. Maybe put this in a separate PR?

token/js/src/extensions/group/state.ts
qiweiii: I created this for geting the `TOKEN_GROUP_SIZE` on the next line, not very familiar with all this, is there other ways to get the size?

qiweiii: maybe `tokenGroupCodec.fixedSize`?

token/js/test/e2e-2022/groupMember.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        };
+
+        await tokenGroupMemberInitializeWithRentTransfer(
+            connection,
+            payer,
+            memberAddress,
+            mint.publicKey,
+            mintAuthority.publicKey,
+            groupAddress,
+            updateAuthority.publicKey,
```
qiweiii, line 87: No sure why this test fails with `Missing signature for public key ` for `updateAuthority`

single-pool/js/packages/modern/.gitignore
buffalojoec: Yeah although it's annoying and a super minor change, it should prob be a separate commit.

token/js/src/extensions/index.ts
buffalojoec: Makes sense, but I'd lean on combining them if it's all the same to you!

token/js/test/e2e-2022/groupMember.test.ts
buffalojoec: Yeah both the member's mint authority and the group's update authority have to sign here.

https://github.com/solana-labs/solana-program-library/blob/e651623033fca7997ccd21e55d0f2388473122f9/token-group/js/src/instruction.ts#L128

You also need to create the group mint first.

qiweiii:
Hi, I was having trouble fixing the groupMember test. I may not be able to finish this soon. It would be great if you could help commit to this branch directly

buffalojoec (approved):
Handoff!

handoff

Issue #6300, opened 2024-02-26
Fix import statement in the extensions.mdx documentation file


MohammedAlabd:
https://github.com/solana-labs/solana-program-library/blob/9ddfe54cc051759f1c619aecf7ba31d93f28d846/docs/src/token-2022/extensions.mdx#L1273

```ts
import {
    createMint,
    createEnableCpiGuardInstruction,
    createInitializeAccountInstruction,
    disableCpiGuard,
    enableCpiGuard,
    getAccountLen,
    ExtensionType,
    TOKEN_2022_PROGRAM_ID,
} from '../src';
```

the import here should be coming from `@solana/spl-token`

there are more than one import in this files that needs to be changes

PR #6301, opened 2024-02-26
Doc(extensions.mdx): correct import statment

MohammedAlabd:
change the import from `../src` to `@solana/spl-token` because users will try to use these scripts in their own projects and these imports needs to come from the library itself

Related to issue:  https://github.com/solana-labs/solana-program-library/issues/6300


joncinque (approved):
Great idea, thanks!

PR #6316, opened 2024-03-01
token-2022: Update base instruction processor, token-metadata, and token-group to use in-place deserialization

joncinque:
#### Problem

Although almost all of the extensions in token-2022 use Pod-style in-place deserialization, the base type (Account and Mint) is still deserialized onto the stack, modified, then written back into the account data. This operation can use up a lot of compute units.

#### Solution

Move the base types of `StateWithExtensions` to be `Pod`-eligible and then use those new `Pod` types in the instruction processors.

It's a bit of a ride to get there, but all of the commits in this PR aim to convey the story:

* 1b80d0f666a45ea66302225a6d0569a11095000e: update the baseline CU usage numbers
* 32d199e9251d3c5d9bcf24097e69b004490cd6a5: refactor the mutable side of `StateWithExtensionsMut` into a trait, to be implemented in a new `Pod`-compatible version of `StateWithExtensions` and `StateWithExtensionsMut`
* 37b60a93539999a1f3bb4e5d537947752bc2b984: introduce a new `PackedSizeOf` trait, since `BaseState` requires that all types implement `Pack`. We want `Pod` types to be usable too, so we introduce a trait that just requires a `LEN` to be declared, which can be used by `Pod` and `Pack` types
* e10b210416178f95d5b6cceac7ffa84ea7e75024: Add the `Pod` versions of `StateWithExtensions` and `StateWithExtensionsMut`
* 83fa5f5aa8adc454d22ecb6779ec20b022a2758d: Add `Pod` versions of the base token account types
* aa8342c5e918b19c604eda61f8425b84860b489d: On the `Pod` library, add const-compatible creators (to be used for the tests in the extensions/mod.rs)
* 2867025e835c2402775c5fcc0456024c16bae0d1: Add new versions of the `alloc_and_serialize_*` functions to work with `Pod` instead of `Pack`. This is total copy-pasta, but making `unpack` work on a trait got extremely hairy. The idea is to introduce these functions, and then remove the old ones when their users in the token-group and token-metadata extensions have updated
* cf28922f45b04d42d9626fe2d710379fa855e882: Add some test helpers -- we want all of the tests in extension/mod.rs to use the Pod types, so this will help a lot
* 20eff9c368cd7c501b50a0b942abe9c65be5ac82: Convert all the tests! Two errors that are returned change in the tests, because the order of checks differs between the `Pack` and `Pod` implementations

From there, we have everything in place! So we add a new instruction count test for `transfer_checked` in 1610ab107776c4d7e0aedc8b72994780ebadf676, then we get to work converting everything. The results are pretty awesome: for the most part, instructions take up half the CUs from before.

A lot of these can be certainly broken up, so let me know if it makes more sense to reduce this PR, maybe by splitting out the changes to `spl-pod` and only including everything up to 20eff9c368cd7c501b50a0b942abe9c65be5ac82 (inclusive). I wanted to see the impact of the change.


CriesofCarrots:
Erm, could you update the description with a line that summarizes what this PR is actually doing? I'm guessing it's implementing Pod-style in-place deserialization for base account types based on the problem statement, but I don't know for sure without grokking all the intermediate steps.

> let me know if it makes more sense to reduce this PR

Yes, please.

PR #6317, opened 2024-03-01
[token-js] Update build instructions for token js

samkim-crypto:
#### Problem
The build instructions for the token js library in the readme file is outdated:
- it suggests to use `npm` instead of `pnpm`
- it suggests to build in `solana-program-library/token/js`, but the build only works at the root of the repository

#### Summary of changes
Updated the instructions to use `pnpm` and to also build at the root of the repository.


joncinque (approved):
Looks great, thanks!

PR #6325, opened 2024-03-04
lockfile: update `mio` to `0.8.11`

buffalojoec:
Fixes the audit vulnerability finding for `mio`, a dependency of `tokio`.


joncinque (approved):
Thanks for jumping on this!

PR #6328, opened 2024-03-04
pod: Add const-compatible creators for primitives

joncinque:
#### Problem

While working on converting token-2022 to `Pod` types in #6316, I need to create `Pod` types in a const context, but that wasn't possible because `From` cannot be used in `const` contexts.

#### Solution

Add a `from_primitive` function on the numbers and `from_bool` for `PodBool`.

Along with that, expose the underlying data stored, so that you can initialize a `PodU64` with `PodU64([0,0,0,0,0,0,0,0])`


buffalojoec (approved):
primitive

Side note: I'm just now realizing we tightly coupled these integer `Pod` types to Little-Endian. Maybe we should document that somewhere.

PR #6329, opened 2024-03-04
token-2022: Refactor `StateWithExtensionsMut` functionality into trait

joncinque:
#### Problem

As part of moving token-2022 to `Pod`-style deserialization in #6316, we'll eventually need to make new `PodStateWithExtensions*`, in which the `BaseState` itself is a Pod type.

That's easy for the non-mutable case, since we have `BaseStateWithExtensions` that neatly encompasses all of the functionality on a trait. We don't have a mutable version of `BaseStateWithExtensions`.

#### Solution

Create `BaseStateWithExtensionsMut`. This essentially moves all of the common functionality of `StateWithExtensionsMut` into a trait requiring only two functions:

* `get_tlv_data_mut`: just the mutable verison of `get_tlv_data` in `BaseStateWithExtensions`
* `get_account_type_mut`: some way to access the account type part of the slice in a mutable way, so that it can be written to during initialization

This PR contains no functional changes, just moving functionality to the new trait.


token/program-2022/src/extension/mod.rs
```diff
@@ -836,24 +755,135 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
     /// state initialization
     /// Noops if there is no room for an extension in the account, needed for
     /// pure base mints / accounts.
-    pub fn init_account_type(&mut self) -> Result<(), ProgramError> {
-        if !self.account_type.is_empty() {
-            if let Some(extension_type) = self.get_first_extension_type()? {
+    fn init_account_type(&mut self) -> Result<(), ProgramError> {
+        let first_extension_type = self.get_first_extension_type()?;
```
CriesofCarrots, line 759: Why did you move this expression? It seems heavier than getting account_type and checking whether it's empty.

token/program-2022/src/extension/mod.rs
```diff
@@ -836,24 +755,135 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
...
+impl<'a, S: BaseState> BaseStateWithExtensionsMut<S> for StateWithExtensionsMut<'a, S> {
+    fn get_tlv_data_mut(&mut self) -> &mut [u8] {
+        self.tlv_data
+    }
+    fn get_account_type_mut(&mut self) -> &mut [u8] {
+        self.account_type
+    }
+}
+
+fn unpack_type_and_tlv_data<S: BaseState>(
```
CriesofCarrots, line 850: It would be rather nice to dedupe this and the following fn. Could have a common method that takes a "check" closure? Something like:
```
fn generic_unpack_type_and_tlv_data_but_with_better_name<S: BaseState, F>(
    rest: &mut [u8],
    check_fn: F,
) -> Result<(&mut [u8], &mut [u8]), ProgramError>
where F: Fn(AccountType) -> Result<(), ProgramError>
 {
        if let Some((account_type_index, tlv_start_index)) = type_and_tlv_indices::<S>(rest)? {
        // type_and_tlv_indices() checks that returned indexes are within range
        let account_type = AccountType::try_from(rest[account_type_index])
            .map_err(|_| ProgramError::InvalidAccountData)?;
        check_fn(account_type)?;
        let (account_type, tlv_data) = rest.split_at_mut(tlv_start_index);
        Ok((
            &mut account_type[account_type_index..tlv_start_index],
            tlv_data,
        ))
    } else {
        Ok((&mut [], &mut []))
    }
}

fn unpack_type_and_tlv_data<S: BaseState>(
    rest: &mut [u8],
) -> Result<(&mut [u8], &mut [u8]), ProgramError> {
    generic_unpack_type_and_tlv_data_but_with_better_name(rest, check_account_type)
}


fn unpack_uninitialized_type_and_tlv_data<S: BaseState>(
    rest: &mut [u8],
) -> Result<(&mut [u8], &mut [u8]), ProgramError> {
    generic_unpack_type_and_tlv_data_but_with_better_name(
        rest,
        |account_type: AccountType| -> Result<(), ProgramError> {
            if account_type != AccountType::Uninitialized {
                Err(ProgramError::InvalidAccountData)
            } else {
                Ok(())
            }
    })
}
```
(I'm sure that code is missing some `S` generic handling.)

CriesofCarrots (commented):
One small question, one suggestion for deduping

token/program-2022/src/extension/mod.rs
joncinque: We're fetching the account type, which means borrowing `self` as mutable, and holding the resulting reference, so we can't call `self` as immutable in `get_first_extension_type`, so we get a compilation error if we don't move the call.

Maybe I was being a bit lazy, so if you can think of a slicker way to do this, I'm all for it!

token/program-2022/src/extension/mod.rs
joncinque: Works for me, thanks!

joncinque: I went with the lowest-hanging fruit name of `unpack_type_and_tlv_data_with_check`, let me know how that looks

token/program-2022/src/extension/mod.rs
```diff
@@ -836,24 +755,139 @@ impl<'data, S: BaseState> StateWithExtensionsMut<'data, S> {
...
+        Ok((&mut [], &mut []))
+    }
+}
+
+fn unpack_type_and_tlv_data<S: BaseState>(
+    rest: &mut [u8],
+) -> Result<(&mut [u8], &mut [u8]), ProgramError> {
+    unpack_type_and_tlv_data_with_check::<S, _>(rest, |account_type| {
+        check_account_type::<S>(account_type)
+    })
```
CriesofCarrots, line 877:
```suggestion
    unpack_type_and_tlv_data_with_check::<S, _>(rest, check_account_type::<S>)
```
Does the compiler not like this simplified version?
Based on [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6e106d01792c188e130832bb36da5cc5), I would expect it to be okay. I'm actually a little surprised clippy doesn't recommend it.

joncinque: Oh yeah, that is much better, done!

CriesofCarrots (approved):

PR #6335, opened 2024-03-05
Governance: Remove legacy CreateGovernance instructions

SebastianBor:
#### Summary

Remove legacy instructions to create asset specific governances: `CreateProgramGovernance`, `CreateTokenGovernance` and `CreateMintGovernance`. The instructions are leftovers from the `spl-governance` inception time when the governance accounts were top level accounts and `Realm` account didn't exist. The asset specific accounts are no longer required because the generic `Governance` account can handle all the use cases. 

#### Backward compatibility 
The change is backward compatible from the state point of view and all the asset specific `Governance` accounts created in previous versions are still supported.

Any program or client using the assets specific instructions to create `Governance` accounts must switch to `CreateGovernance` instruction and explicitly transfer the relevant authorities to the new `Governance`. 

#### Further work 
There is no concept of `governed_account` any longer  and in a follow up PR the field on `Governance` account will be renamed to `governance_seed` to indicate its actual meaning.


joncinque (approved):
Looks great!

PR #6337, opened 2024-03-05
token-2022: Add pod-compatible mint / account / multisig types

joncinque:
#### Problem

In order to use the Pod-compatible types in #6336, we need Pod-compatible base types.

#### Solution

Mostly copy-paste of the types in `state.rs`, but only using types that align to 1. Also introduces `PodCOption<T>` for a `Pod`-compatible `COption`

These will eventually be used for the extension tests.


token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,99 @@
+//! Rewrites of the base state types represented as Pods
+
+use {
+    crate::{instruction::MAX_SIGNERS, state::PackedSizeOf},
+    bytemuck::{Pod, Zeroable},
+    solana_program::{program_pack::IsInitialized, pubkey::Pubkey},
+    spl_pod::primitives::PodU64,
+};
+
+/// Mint data stored as a Pod type
```
CriesofCarrots, line 10: Maybe link to the original type doc?
It would also be nice to have a unit test or something that ensures the Pod structs can never diverge from the original type, perhaps by comparing packed versions?

token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Optional authority used to mint new tokens. The mint authority may only
+    /// be provided during mint creation. If no mint authority is present
+    /// then the mint has a fixed supply and no further tokens may be
+    /// minted.
+    pub mint_authority: PodCOption<Pubkey>,
+    /// Total supply of tokens.
+    pub supply: PodU64,
+    /// Number of base 10 digits to the right of the decimal place.
+    pub decimals: u8,
+    /// Is not 0 if this structure has been initialized
```
CriesofCarrots, line 23: Just "not 0"? Or "exactly 1"?
Also, don't we have a PodBool in spl-pod? Any reason not to use that?

token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// The mint associated with this account
+    pub mint: Pubkey,
+    /// The owner of this account.
+    pub owner: Pubkey,
+    /// The amount of tokens this account holds.
+    pub amount: PodU64,
+    /// If `delegate` is `Some` then `delegated_amount` represents
+    /// the amount authorized by the delegate
+    pub delegate: PodCOption<Pubkey>,
+    /// The account's state, stored as u8
```
CriesofCarrots, line 50: Link to AccountState doc? Not clear what these u8s will mean.

token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Multisignature data.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct PodMultisig {
+    /// Number of signers required
+    pub m: u8,
+    /// Number of valid signers
+    pub n: u8,
+    /// If not 0, this structure has been initialized
```
CriesofCarrots, line 79: Ditto about non-0 and PodBool

token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,99 @@
...
+    const SIZE_OF: usize = std::mem::size_of::<Self>();
+}
+
+/// COption<T> stored as a Pod type
+#[repr(C, packed)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+pub struct PodCOption<T: Pod> {
+    option: [u8; 4],
+    value: T,
+}
```
CriesofCarrots, line 99: Any value in adding this to spl-pod?

token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Optional authority to freeze token accounts.
+    pub freeze_authority: PodCOption<Pubkey>,
+}
+impl IsInitialized for PodMint {
+    fn is_initialized(&self) -> bool {
+        self.is_initialized != 0
+    }
+}
+impl PackedSizeOf for PodMint {
+    const SIZE_OF: usize = std::mem::size_of::<Self>();
```
CriesofCarrots, line 34: While this expression is simple enough, maybe use `spl_pod::pod_get_packed_len::<Self>()` for consistency? Don't feel too strongly, though.

token/program-2022/src/pod.rs
joncinque: Yeah sure, can do!

token/program-2022/src/pod.rs
joncinque: Will do

token/program-2022/src/pod.rs
joncinque: I'd prefer to keep it here since we have the `OptionalNonZero*` types in spl-pod, which are more space-efficient and more strictly `Pod` types. I'm worried it would create confusion by having this type in there too

token/program-2022/src/pod.rs
joncinque: That's a good point, will do!

token/program-2022/src/pod.rs
joncinque: Certainly, done!

token/program-2022/src/pod.rs
joncinque: Will do

joncinque:
All points should be addressed now, thanks for your careful review!

token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// Number of base 10 digits to the right of the decimal place.
+    pub decimals: u8,
+    /// If `true`, this structure has been initialized
+    pub is_initialized: PodBool,
+    /// Optional authority to freeze token accounts.
+    pub freeze_authority: PodCOption<Pubkey>,
+}
+impl IsInitialized for PodMint {
+    fn is_initialized(&self) -> bool {
+        self.is_initialized == PodBool::from_bool(true)
```
CriesofCarrots, line 38:
```suggestion
        self.is_initialized.into()
```
Better?

It would actually be kind of cute if PodBool implemented `PartialEq<bool> for PodBool` so that this line could be:
`self.is_initialized == true`
I can't really find a way to simplify it further to just `self.is_initialized`, though. There's a trait for `Not`, but not one for `Is` aside from PartialEq/Eq.

token/program-2022/src/pod.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// Number of valid signers
+    pub n: u8,
+    /// If `true`, this structure has been initialized
+    pub is_initialized: PodBool,
+    /// Signer public keys
+    pub signers: [Pubkey; MAX_SIGNERS],
+}
+impl IsInitialized for PodMultisig {
+    fn is_initialized(&self) -> bool {
+        self.is_initialized == PodBool::from_bool(true)
```
CriesofCarrots, line 121: Here too, if you like my suggestion above.

CriesofCarrots (dismissed):
Looks great, thanks for all the polish.
One last nit, take it or leave it.

token/program-2022/src/pod.rs
joncinque: ah yeah, much better! We might need to address the `PartialEq` / `Eq` question in a follow-up PR, but your suggestion is much clearer

PR #6343, opened 2024-03-06
token-2022: Update alloc_and_serialize to use Pod types

joncinque:
#### Problem

As part of the transition to using `Pod` types with #6316, the `alloc_and_serialize` extension helpers need to use `Pod` types to avoid more expensive deserialization.

#### Solution

Use `PodStateWithExtensions*` in the `alloc_and_serialize*` functions.

I spent a long time trying to do this by adding an `unpack` function to the `BaseStateWithExtensions` traits, but because `unpack` takes in a `&[u8]`, and `StateWithExtensions` holds onto a reference, the lifetimes were an absolute nightmare. Because of that, I decided to just transition the whole function to use only `Pod` types.

I don't know why the alloc functions were `pub` to begin with, so I've also made them `pub(crate)`. This will require a major version bump.

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -1398,7 +1398,7 @@ impl Extension for AccountPaddingTest {
 /// NOTE: Since this function deals with fixed-size extensions, it does not
 /// handle _decreasing_ the size of an account's data buffer, like the function
 /// `alloc_and_serialize_variable_len_extension` does.
-pub fn alloc_and_serialize<S: BaseState + Pack, V: Default + Extension + Pod>(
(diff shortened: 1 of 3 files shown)


CriesofCarrots (approved):
I feel like there ought to do the unpack thing you mentioned, but this seems straightforward enough. :shipit:

joncinque:
I'm sure there is, but I swear I tried everything that I knew, from higher-rank trait bounds (didn't work because `StateWithExtensions` doesn't implement the trait for all lifetimes), to specifying dependencies between the trait lifetime and the struct lifetime (ie. `'b: 'a`) but it still told me the reference didn't last long enough

PR #6349, opened 2024-03-07
Update old spl-tlv-account-resolution in Cargo.lock

CriesofCarrots:
Duplicate version of `spl-tlv-account-resolution` (which doesn't accept the range of solana-program versions) is breaking agave downstream CI builds when we attempt to bump to v2: https://github.com/anza-xyz/agave/pull/121

I am guessing this fix should have been part of #6221, but I am not sure why cargo did not resolve it automatically. I hope we're not going to have to make manual lock fixes all the time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7450,20 +7450,6 @@ dependencies = [
 
 [[package]]
 name = "spl-tlv-account-resolution"
-version = "0.5.1"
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Hm weird, thanks for fixing this up!

PR #6367, opened 2024-03-08
Governance: Remove ProposalTransaction hold_up_time

SebastianBor:
### Summary 

Remove `hold_up_time` for individual `ProposalTransactions` and use a single `Governance` level `hold_up_time` for all transactions within a `Proposal` instead.

The individual hold up times for transactions haven't been used and if the delay logic, in addition to the `Governance` defined time lock,  is required then it can be implemented by other means like inserting instructions with a delay guard  or a wrapping instructions with a configured delay.


joncinque (approved):
Simple and clean, looks great!

PR #6370, opened 2024-03-08
token-2022: Update transfer to use Pod types (uses half the CUs!)

joncinque:
#### Problem

All of the pod types are ready to be used in the instruction processors, but they're not being used anywhere.

#### Solution

Use pod types in `transfer`, which brings CUs down from 2988 to 1642 for unchecked transfers, and from 3839 to 1960 for checked transfers! This is probably the biggest win from our efforts.

### DIFF
--- a/token/program-2022/src/extension/memo_transfer/mod.rs
+++ b/token/program-2022/src/extension/memo_transfer/mod.rs
@@ -3,8 +3,7 @@ use serde::{Deserialize, Serialize};
 use {
     crate::{
         error::TokenError,
-        extension::{BaseStateWithExtensions, Extension, ExtensionType, StateWithExtensionsMut},
-        state::Account,
+        extension::{BaseState, BaseStateWithExtensions, Extension, ExtensionType},
     },
     bytemuck::{Pod, Zeroable},
     solana_program::{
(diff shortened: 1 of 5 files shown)


token/program-2022/src/extension/memo_transfer/mod.rs
```diff
@@ -33,7 +32,7 @@ impl Extension for MemoTransfer {
 }
 
 /// Determine if a memo is required for transfers into this account
-pub fn memo_required(account_state: &StateWithExtensionsMut<Account>) -> bool {
+pub fn memo_required<BSE: BaseStateWithExtensions<S>, S: BaseState>(account_state: &BSE) -> bool {
```
CriesofCarrots, line 35: BaseState must be Account specifically, though, right?

token/program-2022/src/processor.rs
```diff
@@ -378,25 +381,31 @@ impl Processor {
                     account_info_iter.as_slice(),
                 )?
             }
-            (COption::Some(ref delegate), _) if cmp_pubkeys(authority_info.key, delegate) => {
+            (
+                PodCOption {
+                    option: PodCOption::<Pubkey>::SOME,
+                    value,
```
CriesofCarrots, line 387: nit: I think this was clearer in the original code:
```suggestion
                    value: delegate,
```

token/program-2022/src/extension/memo_transfer/mod.rs
joncinque: Now that we have `PodAccount`, it can be either `Account` or `PodAccount`.

I can add a check to make sure that `S::ACCOUNT_TYPE` is `AccountType::Account` though to give at least a runtime check, but it should be redundant considering we're fetching the `MemoTransfer` extension, which is only allowed on `Account`s. What do you think?

CriesofCarrots: Yeah, I guess it's fine. I just didn't like the feel of losing compile-time type safety, especially on a public method. I suppose the alternative is to add different Account and Mint traits, implemented by `Account`+`PodAccount` and `Mint`+`PodMint` respectively. That seems like far too much work.

CriesofCarrots (approved):

PR #6371, opened 2024-03-08
docs: Add note on on local spl-stake-program deployment

exdx:
I lost several hours trying to figure this out. It may help someone else to have this info blurb be in the docs.

### DIFF
--- a/docs/src/stake-pool/quickstart.md
+++ b/docs/src/stake-pool/quickstart.md
@@ -49,6 +49,9 @@ All of these scripts can be run against devnet, testnet, or mainnet-beta, but
 to allow for more experimentation, we will setup a local validator with some
 validator vote accounts using `setup-test-validator.sh`.
 
+> Note: When running locally, be sure to build and deploy the spl-stake-pool program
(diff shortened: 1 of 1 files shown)


joncinque (commented):
Thanks for your contribution! Is the normal line to pull the program from devnet with `-c` not working?

I know the program is out of date on devnet, but it should be much easier for people running the script to just use what's already deployed rather than force them to build the program.

exdx:
> Thanks for your contribution! Is the normal line to pull the program from devnet with `-c` not working?
> 
> I know the program is out of date on devnet, but it should be much easier for people running the script to just use what's already deployed rather than force them to build the program.

Agreed

It's interesting -- it was not working for me. I tried to run `setup-stake-pool.sh` and it would consistently give an error that the program is missing. 

Looking at the documentation of `-c`:
```
    -c, --clone <ADDRESS>...
            Copy an account from the cluster referenced by the --url argument the genesis configuration. If the ledger
            already exists then this parameter is silently ignored
```

I think it's possible when running locally that there is a ledger present already and this parameter is ignored. I can confirm.

joncinque:
> I think it's possible when running locally that there is a ledger present already and this parameter is ignored. I can confirm.

That's likely the situation you're in. Maybe we can change the script to use `--url mainnet-beta` so it's always copying the most up-to-date program. What do you think?

PR #6373, opened 2024-03-09
refactor: Optimize code based on cargo clippy suggestions

gcmutator:


joncinque (commented):
Thanks for your contribution! It looks like the changes in your refactor haven't been propagated everywhere, causing CI to fail -- would you mind fixing up all uses of `to_u128`?

joncinque:
It looks like it's still failing -- you'll also need to modify all places that call `to_u128` in `token-swap/program/src/processor.rs` since you've changed the return type, otherwise the code will fail to compile. You can run `cargo test` from `token-swap/program` to make sure that everything works.

PR #6374, opened 2024-03-10
correct capitalization: change 'token-2022' to 'Token-2022'

jarrensj:


joncinque (approved):
Thanks for contributing!

PR #6376, opened 2024-03-11
[token-cli] Upgrade to clap-v3

samkim-crypto:
[fd6ee0e](https://github.com/solana-labs/solana-program-library/pull/6376/commits/fd6ee0efbf94d39b9ad151e9fed74c14f9488a4f): Upgrade to clap version `3.2.23` and change dependency `solana-clap-utils` --> `solana-clap-v3-utils`

[145285f](https://github.com/solana-labs/solana-program-library/pull/6376/commits/145285f3b3266b3b0ec473d22b66afb78773607c), [515ddc4](https://github.com/solana-labs/solana-program-library/pull/6376/commits/515ddc4c5636684dc9f878c0a0022a4f5b45804e), [ad1fa87](https://github.com/solana-labs/solana-program-library/pull/6376/commits/ad1fa8755e2e654e2cdb90badade35ea6be5b9c5): Lifetimes were removed from `ArgMatches`, `Arg`, and `App` with clap-v3 (https://github.com/clap-rs/clap/pull/4223, https://github.com/clap-rs/clap/issues/2150, https://github.com/clap-rs/clap/issues/1041), so they were removed. 

[e6f43c6](https://github.com/solana-labs/solana-program-library/pull/6376/commits/e6f43c6b67e2dcc05be441558b9dcb338d36a2e1): `min_values` and `max_values` now take `usize` instead of `u64`, so I updated these inputs.

[7f40b38](https://github.com/solana-labs/solana-program-library/pull/6376/commits/7f40b38c0d107a67f5e0b3675c2340c895005f33): `short` now takes `char` instead of string, so I updated these inputs.

[97b288c](https://github.com/solana-labs/solana-program-library/pull/6376/commits/97b288c0ca42ebb404433666d844f458094b69d5): The requirement on validator functions have been updated from `F: Fn(String) -> ...` to `F: FnMut(&str) -> ...`, so I updated the syntax for custom validator functions.

[a7c5773](https://github.com/solana-labs/solana-program-library/pull/6376/commits/a7c57731705ade0381e1b3f9e65bacb0d7393b5c): `ArgMatches::subcommand` now returns `Option<(&str, &ArgMatches)>` instead of `(&str, Option<&ArgMatches>)`, so I updated these.

[79e11b6](https://github.com/solana-labs/solana-program-library/pull/6376/commits/79e11b6de800c662058b2d22e1f9144d5c508c97): `possible_values` is now more generic and does not automatically infer a string type for its inputs, so I updated to specifically cast to string.

[d2a1271](https://github.com/solana-labs/solana-program-library/pull/6376/commits/d2a1271954c34e10a2fcdd0fad30fb991be7a0fb): Some validation functions have been deprecated in clap-v3-utils. However, removing these deprecated functions will require another set of large changes, so I added `#![allow(deprecated)]` for now. I created https://github.com/solana-labs/solana-program-library/issues/6393 for this and will update in a subsequent PR.

[cb8cc28](https://github.com/solana-labs/solana-program-library/pull/6376/commits/cb8cc2802f2229b54cad5e90dad71a50647252a2): This part is a little bit tricky, but the validator syntax changed from clap-v2 to clap-v3.
- In clap-v2, a custom validator function had to be of type `Fn(String) -> Result<...>`
- In clap-v3, a custom validator function has to be of type `FnMut(&str) -> Result<...>`
The existing custom validator functions have the following syntax:
```
fn custom_validate<T>(string: T) -> Result<...>
where
    T: AsRef<str> + Display
```
In clap-v2, adding `Arg::validator(custom_validate)` was not a problem because the generic `T` could be inferred to be `String`. However, in clap-v3, the generic `T` must be inferred to be `&str` and since `&str` is a borrowed type,`&'a str` for *some* fixed lifetime `'a`. However, `Arg::validator` requires `custom_validate` to be of *any* lifetime, which causes rust to complain. It seems like adding a level of indirection by wrapping the custom validator function inside a closure allows the lifetime associated with `T: &'a str` to be flexibly chosen on the spot is a workaround this, so I made these changes in this commit.

[0daf1a9](https://github.com/solana-labs/solana-program-library/pull/6376/commits/0daf1a90a4c9a6ded039df7b0d64e22d3d035440): Clap-v3 requires positional parameters either all or none of the arguments, so I added indices to some arguments.

[ff00bfd](https://github.com/solana-labs/solana-program-library/pull/6376/commits/ff00bfd4a1dbbeae2d609c28efa6fe30deaf1676): There is currently a useless alias `owner` for specifying `KEYPAIR`s in some subcommands. If the alias `owner` is used to specify a keypair, the argument is interpreted as `OWNER_ADDRESS` (from `owner_address_arg()`), so it can never really be invoked. I removed this alias since clap-v3 complains.

[0afb444](https://github.com/solana-labs/solana-program-library/pull/6376/commits/0afb44499b4b30af64cd6e5f1c3afc0384e0126b): In clap-v3, positional arguments can't have long or short names (which makes sense!), so I removed these.

[b613066](https://github.com/solana-labs/solana-program-library/pull/6376/commits/b613066ec45f7b57571c78d354f6d7cb3a5155fa): In clap v2, `value_of` and `is_present` returns `None` if the argument id is not previously specified in `Arg`. In contract, in clap v3, `value_of` panics if the argument id is not specified previously as `Arg` (see https://github.com/solana-labs/solana/pull/33184). I added custom logic to account for this difference in versions. In some places, these changes are a bit clunky, but cleaning them up will require larger changes, so I will do it in subsequent PRs.

[ac068bf](https://github.com/solana-labs/solana-program-library/pull/6376/commits/ac068bfd6be94a2e2f9841822149fcefa2f7467b): This commit is related to the previous commit. The functions `signer_from_path` and `signer_from_path_with_config` in `solana-clap-v3-utils` panics early in a special case that we cannot easily account for from the caller side. I added custom versions of these functions that specifically checks for this special case.

[a863f08](https://github.com/solana-labs/solana-program-library/pull/6376/commits/a863f08d452f912cc8ec5795270e2e657c7bfeda): `cmd.args([...])` mutates `cmd` to include the specified arguments. Currently, `cmd.args(args).assert()..stderr("error: Could not find config file `~/nonexistent/config.yml`\n");` and `cmd.assert().code(1).failure();` adds `args` to `cmd` twice, so I fixed this.


2501babe:
i took a first read through and it seems pretty straightforward. i was wondering about the removal of the `--owner` aliases tho. is this because v3 doesnt support that? will that break existing commands that people might be using, or is there something im missing there?

samkim-crypto:
Thanks for taking a look! and sorry the details were a bit terse with the issue surrounding `--owner`.

The issue with `--owner` alias is that this conflicts with the `owner_address_arg`. If you look at the `Unwrap`, `Close`, and `WithdrawWithheldTokens` commands, there is `.alias("owner")`, but also `.arg(owner_address_arg())`, which both specify an argument `--owner`. If an `--owner` argument is provided, then clap interprets it as a pubkey argument from `.arg(owner_address_arg())` instead of as an alias for a keypair. This makes the `.alias("owner")` useless. 

Clap-v2 did not really complain about this (the alias was just useless). But clap-v3 is smart enough to catch this and gives an error when invoked, which is why I removed the alias. This should not break existing commands because it was a useless alias from the start.

cbates8979:
I haven't been able to access my solan funds at they never came to my phantom wallet I even check Solflare but I added them to address beginning DYOC

Tyman14888 (commented):
99Z2vPMxbtaHfrJX7Xf2W2iaNimcgAv2X5QSSUhXjosL

token/cli/src/clap_app.rs
```diff
@@ -802,7 +839,7 @@ pub fn app<'a, 'b>(
                         .number_of_values(1)
                         .conflicts_with("transfer_fee")
                         .requires("transfer_fee_basis_points")
-                        .validator(is_amount)
+                        .value_parser(clap::value_parser!(f64))
```
2501babe, line 842: can this parse a number without a decimal?

token/cli/src/clap_app.rs
```diff
@@ -1396,19 +1434,19 @@ pub fn app<'a, 'b>(
                 .arg(
                     Arg::with_name("expected_fee")
                         .long("expected-fee")
-                        .validator(is_amount)
+                        .value_parser(clap::value_parser!(f64))
```
2501babe, line 1437: same here, not sure if these are supposed to stay `is_smount`

token/cli/src/clap_app.rs
samkim-crypto: Yes you are completely right! The validation `is_amount` should actually be replaced with using the `Amount` [type](https://github.com/anza-xyz/agave/blob/master/clap-v3-utils/src/input_parsers/mod.rs#L142), but I thought making this change in this PR would add on a whole new set of changes, so I just tried using the built-in clap value parser without thinking through it carefully. Thanks for the catch!

2501babe (approved):

PR #6377, opened 2024-03-11
chore: add package-lock.json & yarn.lock in gitignore

feikerwu:
chore: add **/package-lock.json to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ target
 bin
 config.json
 node_modules
-./package-lock.json
+**/package-lock.json
 hfuzz_target
 hfuzz_workspace
 **/*.so
@@ -14,3 +14,4 @@ docker-target
 .idea
 .coderrect
 .turbo
+./yarn.lock


joncinque (commented):
Thanks for contributing! What prompted this change exactly? This repo doesn't use yarn anymore, and the `package-lock.json` is intentionally committed in the `docs` folder to make sure we use the right dependencies, so if anything, we should *remove* the package-lock.json line here

feikerwu:
> Thanks for contributing! What prompted this change exactly? This repo doesn't use yarn anymore, and the `package-lock.json` is intentionally committed in the `docs` folder to make sure we use the right dependencies, so if anything, we should _remove_ the package-lock.json line here

Thank you for your reply. I included package-lock.json because I saw that SPL uses pnpm in many js-libraries. Some contributors will use yarn or npm as the default package manager, so I added package-lock.json in gitignore.

In addition, I think we should unify all package managers in all js-libraries

feikerwu:
I've noticed a yarn.lock file in the managed-token/sdk directory and a package-lock.json in the docs directory. Considering our use of pnpm as the package manager, it might be beneficial to remove these files and add them to .gitignore.

joncinque:
Ah I see, thanks for the info! The managed-token sdk is unmaintained, so it's not included by design. And the docs will eventually be moved out of this repo, so it's also meant to have a different package-lock.json

Either way, I still don't see why we would add them to .gitignore. If someone is using the incorrect package manager, it should be as clear as possible to them. If it's in the .gitignore, then it would be harder for them to find out.

PR #6378, opened 2024-03-11
fix(typo): typo

feikerwu:


joncinque (approved):
Thanks for noticing and fixing

PR #6388, opened 2024-03-11
js: Update token-group / token-metadata to use more idiomatic encoders

joncinque:
#### Problem

As pointed out in the comments in https://github.com/solana-labs/solana-program-library/pull/6358, the token group and metadata JS libraries do some slightly non-idiomatic things with the new web3.js.

#### Solution

Implement those suggestions:

* update to using just `@solana/codecs`
* change version to just `2.0.0-preview.1`
* eliminate `packInstruction` for more idiomatic encoder creator
* use a real `PublicKey` encoder rather than a buffer

Unfortunately, we can't totally get rid of `Buffer` because it's required in the `data` part of the `TransactionInstruction` in old web3.js, but now its use is more isolated.

### DIFF
--- a/token-group/js/package.json
+++ b/token-group/js/package.json
@@ -47,11 +47,8 @@
         "@solana/web3.js": "^1.91.0"
     },
     "dependencies": {
-        "@solana/codecs-core": "2.0.0-preview.1.20240308041540.58947c719d61c07ce3b1ac04edf51a2b8fa4a3eb",
-        "@solana/codecs-data-structures": "2.0.0-preview.1.20240308041540.58947c719d61c07ce3b1ac04edf51a2b8fa4a3eb",
-        "@solana/codecs-numbers": "2.0.0-preview.1.20240308041540.58947c719d61c07ce3b1ac04edf51a2b8fa4a3eb",
-        "@solana/codecs-strings": "2.0.0-preview.1.20240308041540.58947c719d61c07ce3b1ac04edf51a2b8fa4a3eb",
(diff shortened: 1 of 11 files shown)


token-group/js/package.json
```diff
@@ -47,11 +47,8 @@
         "@solana/web3.js": "^1.91.0"
     },
     "dependencies": {
-        "@solana/codecs-core": "2.0.0-preview.1.20240308041540.58947c719d61c07ce3b1ac04edf51a2b8fa4a3eb",
-        "@solana/codecs-data-structures": "2.0.0-preview.1.20240308041540.58947c719d61c07ce3b1ac04edf51a2b8fa4a3eb",
-        "@solana/codecs-numbers": "2.0.0-preview.1.20240308041540.58947c719d61c07ce3b1ac04edf51a2b8fa4a3eb",
-        "@solana/codecs-strings": "2.0.0-preview.1.20240308041540.58947c719d61c07ce3b1ac04edf51a2b8fa4a3eb",
-        "@solana/options": "2.0.0-preview.1.20240308041540.58947c719d61c07ce3b1ac04edf51a2b8fa4a3eb",
+        "@solana/codecs": "2.0.0-preview.1",
+        "@solana/options": "2.0.0-preview.1",
```
lorisleiva, line 51: I know it's not super intuitive but `@solana/codecs` also contains `@solana-options`. We should probably rename the latter `@solana/codecs-options` because 70% of the package is codecs anyway.

lorisleiva (approved):
Looks great!

token-group/js/package.json
joncinque: Ah ok, will do in follow-up work!

PR #6389, opened 2024-03-11
js: Remove direct usage of `@solana/options`

joncinque:
#### Problem

Per https://github.com/solana-labs/solana-program-library/pull/6388#discussion_r1519883981, we're using `@solana/options` by name, which is also included in `@solana/codecs`

#### Solution

Use everything from `@solana/codecs` directly, and remove the dependency on `@solana/options`


lorisleiva (approved):
Perfect 👌

PR #6396, opened 2024-03-12
Extensions doc metadata js example

MohammedAlabd:
adding js examples for the metadata part of the token22 extension


MohammedAlabd:
not sure if you have other plans for it but I was working on a course on that extension and I figured that I could drop these here quickly.

So sorry if I am stepping on somebody else's work

joncinque (commented):
This is excellent work, thanks so much! I confirm that it all works in my local testing.

We try to keep these examples in line with the `examples` at https://github.com/solana-labs/solana-program-library/tree/master/token/js/examples -- would you mind combining all of these into a file called `metadata.ts` containing all of these cases?

MohammedAlabd:
@joncinque Yes sure I will do that tomorrow

MohammedAlabd:
@joncinque can you take another look please?

joncinque (approved):
Perfect, thanks again!

PR #6406, opened 2024-03-12
token-2022: Update burn to use Pod types

joncinque:
#### Problem

The Pod types exist in token-2022, but burn doesn't use them

#### Solution

Use Pod types in burn in token-2022


CriesofCarrots (approved):
Lgtm, although I assume this needs a rebase on #6405

joncinque:
Looks like git figured it out!

PR #6409, opened 2024-03-12
token-2022: Update initialize_immutable_owner to use Pod types

joncinque:
#### Problem

The Pod types exist in token-2022, but initialize_immutable_owner doesn't use them

#### Solution

Use Pod types in initialize_immutable_owner in token-2022

Side note: I thought it was very cool that this only takes one line


CriesofCarrots (approved):
This one is my favorite so far :)

PR #6414, opened 2024-03-13
token-2022: Finish updating base processors to Pod types

joncinque:
#### Problem

There are still a few base instruction processors using the non-Pod types.

#### Solution

Update the last few: `withdraw_excess_lamports`, `initialize_non_transferable_mint`, and `initialize_permanent_delegate`.

With this, we can close #6316 !


CriesofCarrots (approved):
Looks good! Woot!

Issue #6418, opened 2024-03-13
token-2022: Convert extensions to use Pod state types


joncinque:
#### Problem

`Pod` versions of the base states in `token-2022` exist in the repo and are being used in all of the base instruction processors, but the extensions aren't using them yet.

#### Solution

Start using the `Pod` equivalents of `Mint` / `Account` / `Multisig` in the extension instruction processors. The extensions include:

- [x] confidential transfers
- [x] confidential transfers with fee
- [x] cpi guard
- [x] default account state
- [x] group member pointer
- [x] group pointer
- [x] interest bearing mint
- [x] memo transfer
- [x] metadata pointer
- [x] token group
- [x] token metadata
- [x] transfer fee
- [x] transfer hook

And also the `Multisig` usage in the base instruction processor needs to use `PodMultisig`.

joncinque:
All done!

PR #6432, opened 2024-03-13
token-2022: Update multisig usage to PodMultisig

joncinque:
#### Problem

As part of #6418, the multisig instructions and validators do not use Pod types.

#### Solution

Use them.

This one had me slightly worried since it requires holding onto the mutable reference to the data, but I didn't find any potential double-borrow situations since we're only ever looking at the keys and signer status of other accounts.


CriesofCarrots (approved):
> This one had me slightly worried since it requires holding onto the mutable reference to the data, but I didn't find any potential double-borrow situations since we're only ever looking at the keys and signer status of other accounts.

Yep, nice

PR #6441, opened 2024-03-15
WIP: Init With Root

austbot:
This pr serves as another attempt to get init with root in. This PR is wip as tests are still being written. But would like to have buy in on the approach. 

The approach is as follows, you can init a tree with a root and the proof of the first filled leaf, leaf at index 0. For huuuuuuuuuge trees there is a buffer mechanism to allow you to fill the canopy with more proofs than would fit in a transaction and then you can init.

Im hoping minimize code changes and work within the existing framework, ensure compatibility with indexers.


account-compression/programs/account-compression/src/lib.rs
```diff
@@ -173,69 +202,137 @@ pub mod spl_account_compression {
...
-    //         initialize_with_root,
-    //         root,
-    //         leaf,
-    //         &proof,
-    //         index
-    //     )?;
-    //     wrap_event(change_log.try_to_vec()?, &ctx.accounts.log_wrapper)?;
-    //     update_canopy(canopy_bytes, header.max_depth, Some(change_log))
-    // }
+    /// For extremley large trees that cannot be initialized with root in a single transaction,
```
ngundotra, line 205: trees over height *12

*provide a specific height recommendation here

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -70,6 +72,33 @@ pub struct Initialize<'info> {
...
 }
 
+/// Context for filling a proof buffer
+#[derive(Accounts)]
+pub struct FillProofBuffer<'info> {
+    /// Proof Buffer
+    pub payer: Signer<'info>,
+    pub proof_buffer: Signer<'info>,
+    pub system_program: Program<'info, System>,
+}
```
ngundotra, line 82: Missing close buffer instruction in case user funds get trapped

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -173,69 +202,137 @@ pub mod spl_account_compression {
...
-    //     )?;
-    //     wrap_event(change_log.try_to_vec()?, &ctx.accounts.log_wrapper)?;
-    //     update_canopy(canopy_bytes, header.max_depth, Some(change_log))
-    // }
+    /// For extremley large trees that cannot be initialized with root in a single transaction,
+    /// this allows a user to fill a buffer of proofs that can be used to initialize the tree with roots.
+    pub fn fill_proof_buffer(
+        ctx: Context<FillProofBuffer>,
+        max_depth: u32,
+        partial_proof: Vec<[u8; 32]>,
```
ngundotra, line 210: I think this should just be Vec<u8>. Throw validation error on init or when buffer is full if final length is not multiple of 32

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -173,69 +202,137 @@ pub mod spl_account_compression {
...
+                u32::from_le_bytes(
+                    buffer[0..4]
+                        .try_into()
+                        .map_err(|_| AccountCompressionError::InvalidProofBuffer)?,
+                ),
+                u32::from_le_bytes(
+                    buffer[4..8]
+                        .try_into()
+                        .map_err(|_| AccountCompressionError::InvalidProofBuffer)?,
+                ),
```
ngundotra, line 254: Feels like a decent idea to just give this buffer a header & version it. In case people want to use u64s later.

ngundotra (changes requested):
Can you uncommit the lint & do in a separate PR? Makes it difficult to see your changes.

Left a bunch of comments

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -461,10 +558,11 @@ pub mod spl_account_compression {
         // Close merkle tree account
         // 1. Move lamports
         let dest_starting_lamports = ctx.accounts.recipient.lamports();
-        **ctx.accounts.recipient.lamports.borrow_mut() = dest_starting_lamports
+        let tree_lamports = dest_starting_lamports
             .checked_add(ctx.accounts.merkle_tree.lamports())
             .unwrap();
-        **ctx.accounts.merkle_tree.lamports.borrow_mut() = 0;
+        **ctx.accounts.recipient.try_borrow_mut_lamports()? = tree_lamports;
+        **ctx.accounts.merkle_tree.try_borrow_mut_lamports()? = 0;
```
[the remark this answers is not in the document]
austbot: consitency, also my editor didnt like this

account-compression/programs/account-compression/src/lib.rs
austbot: k gud plan

account-compression/programs/account-compression/src/lib.rs
austbot: gud idea

account-compression/sdk/tests/utils.ts
```diff
@@ -115,3 +115,60 @@ export async function createEmptyTreeOnChain(
...
+    depthSizePair: ValidDepthSizePair,
+    firstLeaf: Buffer,
+    manifestUrl: string,
+    payer: Keypair,
+    proofBuffer?: PublicKey,
+    proofs?: Buffer[],
+    provider: AnchorProvider,
+    root: Buffer
+};
+export async function createTreeWithRoot(
```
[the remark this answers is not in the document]
austbot: yeah tests on the way wanted feedback on approach

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -173,69 +202,137 @@ pub mod spl_account_compression {
...
+            invoke(
+                &create_ix,
+                &[
+                    ctx.accounts.payer.to_account_info(),
+                    ctx.accounts.proof_buffer.to_account_info(),
+                    ctx.accounts.system_program.to_account_info(),
+                ],
+            )?;
+            buffer[0..4].copy_from_slice(&max_depth.to_le_bytes());
+            buffer[4..8].copy_from_slice(&max_depth.to_le_bytes());
```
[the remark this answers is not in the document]
austbot: yep

account-compression/programs/account-compression/src/lib.rs
austbot: k

account-compression/programs/account-compression/src/lib.rs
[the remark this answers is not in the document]
austbot: actually no, remaning starts at the number of needed parts of the proof which is equal to depth

account-compression/programs/account-compression/src/lib.rs
austbot: these should be smaller even , like max this value can be is 30, so I disagree unless we plan to add more cmt tree sizes.

account-compression/programs/account-compression/src/lib.rs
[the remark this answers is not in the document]
ngundotra: Curious what error the editor throws, but for now can we just leave this part alone?

account-compression/programs/account-compression/src/lib.rs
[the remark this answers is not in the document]
ngundotra: ah I see. okay, if u change ix data to Vec<u8> instead of Vec<[u8; 32]>, then this will have to change to `2 << max_depth` tho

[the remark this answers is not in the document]
austbot: No this will always be equal to ma x when initializing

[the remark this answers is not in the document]
austbot: max equals depth not max leaves ill make the clearer

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -173,69 +212,158 @@ pub mod spl_account_compression {
...
+    // The indexers for the specific tree use case are in charge of handling the indexing or shadow indexing of the tree.
+    // Shadow indexing is a technique that allows for the indexing of a tree incrementally with update operations.
+    pub fn init_merkle_tree_with_root(
+        ctx: Context<InitializeWithRoot>,
+        max_depth: u32,
+        max_buffer_size: u32,
+        root: [u8; 32],
+        leaf_index: u32,
+        leaf: [u8; 32], // the last leaf in the tree
+        manifest_url: String,
```
StanChe, line 306: Great work, @austbot, this should unlock projects in rolling out more compression. As for the manifest_url,  wouldn't it be more convenient to leave this logic to the caller of the `init_merkle_tree_with_root` like the bubblegum contract? As devs may want to enreach the relevent events with other information for the indexers as well, wdyt?

austbot: yeah this makes sense, ill remove it from here.

account-compression/sdk/src/constants/index.ts
```diff
@@ -9,8 +9,8 @@ export const SPL_NOOP_PROGRAM_ID = new PublicKey(SPL_NOOP_ADDRESS);
  * creating a new {@link ConcurrentMerkleTreeAccount}.
  */
 export type DepthSizePair = {
-    maxDepth: number;
     maxBufferSize: number;
+    maxDepth: number;
 };
```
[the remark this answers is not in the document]
austbot: prolly just automated lint shizz

account-compression/sdk/idl/spl_account_compression.json
```diff
@@ -570,12 +673,22 @@
...
+    },
+    {
+      "code": 6009,
+      "name": "ProofIndexOutOfBounds",
+      "msg": "Proof index of is out of bounds"
+    },
+    {
+      "code": 6010,
+      "name": "InvalidProofBuffer",
+      "msg": "Invalid proof buffer"
```
danenbm, line 685: Looks like IDL and SDK need to be regenerated after latest changes

account-compression/sdk/src/generated/types/ChangeLogEvent.ts
```diff
@@ -32,20 +31,22 @@ export type ChangeLogEventRecord = {
...
-    ],
-]) as beet.FixableBeet<ChangeLogEvent>;
+  [
+    'V1',
+    new beet.FixableBeetArgsStruct<ChangeLogEventRecord['V1']>(
+      [['fields', beet.tuple([changeLogEventV1Beet])]],
+      'ChangeLogEventRecord["V1"]'
+    ),
+  ],
+]) as beet.FixableBeet<ChangeLogEvent, ChangeLogEvent>
```
danenbm, line 52: Nit: Has the Solita autogenerated code been run through a different linter?  It seems so because afaik Solita generates with commas and semicolons.  I would recommend regenerating the Solita TS and not linting it, in order to keep it consistent with other generated Solita code and minimize changed files for the next PR.  (I do agree the 2-space indent belongs, and I think Solita does that by default).

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -173,69 +189,59 @@ pub mod spl_account_compression {
...
+        root: [u8; 32],
+        leaf: [u8; 32], // the first leaf
+        leaf_index: u32,
+    ) -> Result<()> {
+        require_eq!(
+            *ctx.accounts.merkle_tree.owner,
+            crate::id(),
+            AccountCompressionError::IncorrectAccountOwner
+        );
+        let mut merkle_tree_bytes: std::cell::RefMut<'_, &mut [u8]> =
```
danenbm, line 208: Why is the type added for `merkle_tree_bytes`?  Couldn't it just be `let mut merkle_tree_bytes = ctx.accounts.merkle_tree...`"

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -69,6 +71,20 @@ pub struct Initialize<'info> {
...
+    /// CHECK: This account will be zeroed out, and the size will be validated
+    pub merkle_tree: UncheckedAccount<'info>,
+
+    /// Authority that controls write-access to the tree
+    /// Typically a program, e.g., the Bubblegum contract validates that leaves are valid NFTs.
+    pub authority: Signer<'info>,
+
+    /// Program used to emit changelogs as cpi instruction data.
+    pub noop: Program<'info, Noop>,
+}
```
danenbm, line 87: Can we just use `Initialize` as its a duplicate struct?

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -173,69 +189,59 @@ pub mod spl_account_compression {
...
-    // }
+    // Creates a new merkle tree with an existing root in the buffer.
+    // The indexers for the specific tree use case are in charge of handling the indexing or shadow indexing of the tree.
+    // Shadow indexing is a technique that allows for the indexing of a tree incrementally with update operations.
+    pub fn init_merkle_tree_with_root(
+        ctx: Context<InitializeWithRoot>,
+        max_depth: u32,
+        max_buffer_size: u32,
+        root: [u8; 32],
+        leaf: [u8; 32], // the first leaf
```
danenbm, line 200: Nit: comment or param name should be `rightmost_leaf` imo as "first leaf" would only be rightmost if it was empty?

danenbm (commented):
Looking good!  Left a few small comments.  Note CI is failing now.

account-compression/programs/account-compression/src/lib.rs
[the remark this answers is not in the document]
danenbm: I don't think this was resolved?

danenbm:
> Can you uncommit the lint & do in a separate PR? Makes it difficult to see your changes.
> 
> Left a bunch of comments

+1 to undoing the lint.  In my review I suggested linting in a way that matches Solita output as well (I think this linting differs slightly from Solita generated output, but I could be mistaken)

StanChe:
Unfortunately this change doesn't take into account canopy. If the tree account is created with space for canopy, and the init with root method is called, the proofs provided to modify the assets will be shorter and will only represent the proof under the canopy. Only the last leaf will have the canopy cached properly, while the rest of the canopy will be considered empty. Thus the proofs under any other then the last canopy node will be invalid. I could take over from here and provide an option to init the canopy before the root is provided and also verify the canopy is properly initialized (has all the nodes to the left of the last leaf) before allowing the initialization.

account-compression/sdk/src/generated/types/ChangeLogEvent.ts
austbot: they asked me to remove the lint

danenbm: Yeah got it, I think that @ngundotra just didn't want to see all the lint changes as part of this PR, but @StanChe and I were looking into how to actually get the format that is checked in.

It looks like by default the prettierrc.yaml file is not used, so you have to do something like this:
 ```
 pnpm solita
 pnpm lint:fix
 pnpm prettier --config prettierrc.yaml  --write idl/spl_account_compression.json
```

I still see some slight differences in the output and had some error messages in the output, but it did work. 
 @StanChe might have gotten cleaner results but wasn't sure of his local setup.
 
Having said all this, I think the bigger issue is the missing canopy issue described below in @StanChe 's comment.  This could be a blocker for the PR and I think @StanChe is looking into a solution to initialize the canopy on a tree that uses one.

PR #6459, opened 2024-03-19
stake-pool: supporting AddValidatorToPool in js library

kaplanmaxe:
This PR adds support for the `AddValidatorToPool` instruction for the stake pool program in the javascript library. Removing a validator has not been implemented in the spirit of keeping the PR as small as possible. If accepted, I can also implement support for that instruction as well.

My guess is that this instruction was never prioritized in the javascript library due to the `spl-stake-pool` package mostly being used in frontend applications. This instruction is helpful to have though for users who are using nodejs, hence why I added support for it.

The code was successfully run on mainnet with tx hash `4QnuyTWTYYTPX1pr8sw5haFUp96PnNXCe4VWbWZVzN9PcRvhDkTaGpxKvpRC2wETC23VrgoSBdDC1UhZ7s3vfoM6` targetting Edgevana's stake pool.

### DIFF
--- a/stake-pool/js/src/index.ts
+++ b/stake-pool/js/src/index.ts
@@ -649,6 +649,55 @@ export async function withdrawSol(
   };
 }
 
+export async function addValidatorToPool(
+  connection: Connection,
+  stakePoolAddress: PublicKey,
+  validatorVote: PublicKey,
+  seed?: number,
+) {
+  const stakePoolAccount = await getStakePoolAccount(connection, stakePoolAddress);
+  const stakePool = stakePoolAccount.account.data;
+  const { reserveStake, staker, validatorList } = stakePool;
+
+  const validatorListAccount = await getValidatorListAccount(connection, validatorList);
+
+  const validatorInfo = validatorListAccount.account.data.validators.find(
+    (v) => v.voteAccountAddress.toBase58() == validatorVote.toBase58(),
+  );
+
+  if (validatorInfo) {
+    throw new Error('Vote account is already in validator list');
+  }
+
+  const withdrawAuthority = await findWithdrawAuthorityProgramAddress(
+    STAKE_POOL_PROGRAM_ID,
+    stakePoolAddress,
+  );
+
+  const validatorStake = await findStakeProgramAddress(
+    STAKE_POOL_PROGRAM_ID,
+    validatorVote,
+    stakePoolAddress,
+    seed,
+  );
+
+  const instructions: TransactionInstruction[] = [
+    StakePoolInstruction.addValidatorToPool({
+      stakePool: stakePoolAddress,
+      staker: staker,
+      reserveStake: reserveStake,
+      withdrawAuthority: withdrawAuthority,
+      validatorList: validatorList,
+      validatorStake: validatorStake,
+      validatorVote: validatorVote,
+    }),
+  ];
+
+  return {
+    instructions,
+  };
+}
+
 /**
  * Creates instructions required to increase validator stake.
  */
--- a/stake-pool/js/src/instructions.ts
+++ b/stake-pool/js/src/instructions.ts
@@ -36,7 +36,8 @@ export type StakePoolInstructionType =
   | 'IncreaseAdditionalValidatorStake'
   | 'DecreaseAdditionalValidatorStake'
   | 'DecreaseValidatorStakeWithReserve'
-  | 'Redelegate';
+  | 'Redelegate'
+  | 'AddValidatorToPool';
 
 // 'UpdateTokenMetadata' and 'CreateTokenMetadata' have dynamic layouts
 
@@ -91,6 +92,10 @@ export function tokenMetadataLayout(
 export const STAKE_POOL_INSTRUCTION_LAYOUTS: {
   [type in StakePoolInstructionType]: InstructionType;
 } = Object.freeze({
+  AddValidatorToPool: {
+    index: 1,
+    layout: BufferLayout.struct<any>([BufferLayout.u8('instruction'), BufferLayout.u32('seed')]),
+  },
   DecreaseValidatorStake: {
     index: 3,
     layout: MOVE_STAKE_LAYOUT,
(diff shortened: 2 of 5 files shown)


stake-pool/js/src/index.ts
```diff
@@ -649,6 +649,58 @@ export async function withdrawSol(
...
+  connection: Connection,
+  stakePoolAddress: PublicKey,
+  validatorVote: PublicKey,
+  seed?: number,
+) {
+  const stakePool = await getStakePoolAccount(connection, stakePoolAddress);
+
+  const validatorList = await getValidatorListAccount(
+    connection,
+    stakePool.account.data.validatorList,
```
buffalojoec, line 662: nit: since you do this long property access a few times in here, why not simplify with something like this?

```ts
const stakePoolAccount = await getStakePoolAccount(connection, stakePoolAddress);
const stakePool = stakePoolAccount.account.data;
const { reserveStake, staker, validatorList } = stakePool;

const validatorListAccount = await getValidatorListAccount(connection, validatorList);
```

stake-pool/js/src/index.ts
```diff
@@ -649,6 +649,58 @@ export async function withdrawSol(
...
+      withdrawAuthority: withdrawAuthority,
+      validatorList: stakePool.account.data.validatorList,
+      validatorStake: validatorStake,
+      validatorVote: validatorVote,
+    }),
+  );
+
+  return {
+    instructions,
+  };
```
buffalojoec, line 701: Since we don't need to dynamically push anything else into this list, why not just return the single-element list as-is, without doing `instructions.push()`?

stake-pool/js/src/utils/program-address.ts
```diff
@@ -28,9 +28,14 @@ export async function findStakeProgramAddress(
   programId: PublicKey,
   voteAccountAddress: PublicKey,
   stakePoolAddress: PublicKey,
+  seed?: BN,
```
buffalojoec, line 31: Hm, does this have to be `BN` here? The seed itself for a stake address is `u32` so `number` is enough.

stake-pool/js/src/utils/program-address.ts
```diff
@@ -28,9 +28,14 @@ export async function findStakeProgramAddress(
...
 ) {
   const [publicKey] = await PublicKey.findProgramAddress(
-    [voteAccountAddress.toBuffer(), stakePoolAddress.toBuffer()],
+    [
+      voteAccountAddress.toBuffer(),
+      stakePoolAddress.toBuffer(),
+      seed ? seed.toArrayLike(Buffer, 'le', 4) : Buffer.alloc(0),
```
buffalojoec, line 37: nit-up-to-you: I think a `u32` encoder from `buffer-layout` would work here, too.

stake-pool/js/test/instructions.test.ts
```diff
@@ -99,6 +137,50 @@ describe('StakePoolProgram', () => {
     expect(instruction2.keys[10].pubkey).toEqual(payload.depositAuthority);
   });
 
+  describe('addValidatorToPool', () => {
+    const validatorList = mockValidatorList();
+    const decodedValidatorList = ValidatorListLayout.decode(validatorList.data);
+    const voteAccount = decodedValidatorList.validators[0].voteAccountAddress;
+    // decodedValidatorList.validators.shift();
```
buffalojoec, line 144: Should this comment come out?

buffalojoec (commented):
This is looking really good to me! I've just added some small comments.

stake-pool/js/test/instructions.test.ts
kaplanmaxe: Good catch, thanks!

stake-pool/js/src/utils/program-address.ts
kaplanmaxe: Changed

stake-pool/js/src/index.ts
kaplanmaxe: Agreed and done

stake-pool/js/src/index.ts
kaplanmaxe: Agreed and done

stake-pool/js/src/utils/program-address.ts
kaplanmaxe: I chose to keep this just because it matches the format of the functions below it. But let me know if that's a big deal. Was trying to make minimal amount of changes to this file just due to the sensitivity of it (and i'm a first time contriubtor :) )

stake-pool/js/src/instructions.ts
```diff
@@ -376,10 +381,61 @@ export type UpdateTokenMetadataParams = {
...
 
+export type AddValidatorToPoolParams = {
+  stakePool: PublicKey;
+  staker: PublicKey;
+  reserveStake: PublicKey;
+  withdrawAuthority: PublicKey;
+  validatorList: PublicKey;
+  validatorStake: PublicKey;
+  validatorVote: PublicKey;
+  seed?: number | BN;
```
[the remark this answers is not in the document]
kaplanmaxe: done

kaplanmaxe:
Thanks a lot for your review and commenets @buffalojoec ! I greatly appreciate it. I've addressed your comments and squashed my commits

stake-pool/js/src/index.ts
```diff
@@ -649,6 +649,55 @@ export async function withdrawSol(
...
+  const withdrawAuthority = await findWithdrawAuthorityProgramAddress(
+    STAKE_POOL_PROGRAM_ID,
+    stakePoolAddress,
+  );
+
+  const validatorStake = await findStakeProgramAddress(
+    STAKE_POOL_PROGRAM_ID,
+    validatorVote,
+    stakePoolAddress,
+    seed || undefined,
```
buffalojoec, line 681: Sorry, must have missed this on my first pass.

```suggestion
    seed,
```

stake-pool/js/src/utils/program-address.ts
```diff
@@ -28,9 +28,14 @@ export async function findStakeProgramAddress(
...
   voteAccountAddress: PublicKey,
   stakePoolAddress: PublicKey,
+  seed?: number,
 ) {
   const [publicKey] = await PublicKey.findProgramAddress(
-    [voteAccountAddress.toBuffer(), stakePoolAddress.toBuffer()],
+    [
+      voteAccountAddress.toBuffer(),
+      stakePoolAddress.toBuffer(),
+      seed ? new Uint8Array([seed]) : Buffer.alloc(0),
```
buffalojoec, line 37: This won't give us four LE bytes for a `u32`. I think what you had originally was better, or `u32.encode()` from `buffer-layout`.

buffalojoec: Or even better: `buffer.writeUInt32LE()`.

You could do something like:

```ts
      seed ? Buffer.alloc(4).writeUint32LE(seed) : Buffer.alloc(0),
```

kaplanmaxe: Ahh sorry btw. I had submitted my fix before I saw your comment. Let me know if you prefer that

kaplanmaxe: `Buffer.alloc(4).writeUint32LE(seed)` returns `number` so it would need to be converted if you prefer that.

buffalojoec: I'm good with `toArrayLike`!

buffalojoec (approved):
Thanks!!

steak

PR #6473, opened 2024-03-21
token js: add missing "mint" param to docs for approveChecked

PBillingsby:
The documentation on the `approveChecked` function in `token/js/src/actions/approveChecked.ts` is missing the mint param.


buffalojoec (dismissed):
Thanks!

PR #6481, opened 2024-03-22
token-js: bump to 0.4.2 for release

buffalojoec:
#### Problem

Thanks to @qiweiii, the `TokenGroup` and `TokenGroupMember` extensions are now available in `@solana/spl-token`, but we need to publish a release with the new features!

#### Solution
Ship it!

### DIFF
--- a/token/js/package.json
+++ b/token/js/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@solana/spl-token",
     "description": "SPL Token Program JS API",
-    "version": "0.4.1",
+    "version": "0.4.2",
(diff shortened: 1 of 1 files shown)


joncinque (approved):
ship

Eianaftyio:
git diff [<options>] [<commit>] [--] [<path>…]
git diff [<options>] --cached [--merge-base] [<commit>] [--] [<path>…]
git diff [<options>] [--merge-base] <commit> [<commit>…] <commit> [--] [<path>…]
git diff [<options>] <commit>…<commit> [--] [<path>…]
git diff [<options>] <blob> <blob>
git diff [<options>] --no-index [--] <path> <path>

PR #6482, opened 2024-03-22
stake-pool-js: Remove `divideBnToNumber`, use BNs more

joncinque:
#### Problem

According to reports, `divideBnToNumber` is causing some issues for big numbers that fall outside of the safe js `number:

```
Error: Number can only safely store up to 53 bits
    at assert (bn.js:7:21)
    at BN.toNumber (bn.js:548:7)
    at divideBnToNumber (index.browser.esm.js:774:70)
    at calcPoolTokensForDeposit (index.browser.esm.js:754:23)
    at prepareWithdrawAccounts (index.browser.esm.js:722:42)
    at async Module.withdrawStake (index.browser.esm.js:1757:35)
```

#### Solution

Just kick out `divideBnToNumber` and use correct floor and ceiling divisions as needed.

### DIFF
--- a/stake-pool/js/src/index.ts
+++ b/stake-pool/js/src/index.ts
@@ -60,7 +60,7 @@ export interface StakePoolAccount {
 export interface WithdrawAccount {
   stakeAddress: PublicKey;
   voteAddress?: PublicKey;
-  poolAmount: number;
+  poolAmount: BN;
 }
 
 /**
@@ -354,7 +354,7 @@ export async function withdrawStake(
   validatorComparator?: (_a: ValidatorAccount, _b: ValidatorAccount) => number,
 ) {
   const stakePool = await getStakePoolAccount(connection, stakePoolAddress);
-  const poolAmount = solToLamports(amount);
+  const poolAmount = new BN(solToLamports(amount));
 
   if (!poolTokenAccount) {
     poolTokenAccount = getAssociatedTokenAddressSync(stakePool.account.data.poolMint, tokenOwner);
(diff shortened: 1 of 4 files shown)


stake-pool/js/src/index.ts
```diff
@@ -452,10 +452,10 @@ export async function withdrawStake(
 
     const availableForWithdrawal = calcLamportsWithdrawAmount(
       stakePool.account.data,
-      stakeAccount.lamports - MINIMUM_ACTIVE_STAKE - stakeAccountRentExemption,
+      new BN(stakeAccount.lamports - MINIMUM_ACTIVE_STAKE - stakeAccountRentExemption),
```
segfaultdoc, line 455: how do we handle negative ints?

joncinque: Unless otherwise specified, I believe `BN` does signed numbers: https://github.com/indutny/bn.js/?tab=readme-ov-file#prefixes

And if you mean for this particular line, it shouldn't ever go negative since we're checking the existence of the account earlier, and the stake pool program guarantees the account has a certain minimum.

But I can add a defensive check to make sure it's positive here, if you like

stake-pool/js/src/utils/stake.ts
```diff
@@ -186,35 +185,25 @@ export async function prepareWithdrawAccounts(
...
-  if (denominator.isZero()) {
-    return 0;
+    return new BN(0);
   }
-  const quotient = numerator.div(denominator).toNumber();
-  const rem = numerator.umod(denominator);
-  const gcd = rem.gcd(denominator);
-  return quotient + rem.div(gcd).toNumber() / denominator.div(gcd).toNumber();
+  // ceiling the calculation by adding denominator - 1 to the numerator
+  return numerator.add(denominator).sub(new BN(1)).div(denominator);
```
ebatsell, line 206: a test to make sure the ceil math works would be good

joncinque: Looking back through, I've realized that this is inconsistent with the program, and will revert: https://github.com/solana-labs/solana-program-library/blob/c38a1b5d6a1a6049b5844aef8e50e889e18918ed/stake-pool/program/src/state.rs#L178-L189

joncinque: Weirdly enough, the result is unused where we `Math.ceil`, so this was never important

joncinque:
I received reports over the weekend that this works, so merging it in

Issue #6338, opened 2024-03-06
Confidential Transfer Support in @solana/spl-token


ixmorrow:
I have been messing with confidential transfers on the client side recently and ran into some issues with the `@solana/spl-token` API that I'd like to highlight here.

1. [`getMintLen()`](https://solana-labs.github.io/solana-program-library/token/js/functions/getMintLen.html) does not return the correct byte size for a mint with the confidential transfer extension.
* This method seems to be the suggested way for determining how much space to allocate when creating an account with extensions from the client side. Using this function with the [ExtensionType.ConfidentialTransferMint type](https://solana-labs.github.io/solana-program-library/token/js/enums/ExtensionType.html#ConfidentialTransferMint) as input returns a size of 267 bytes.
* 267 bytes does not seem to be the correct size for a mint with the ConfidentialTransfer extension. Attempting to initialize a mint account with 267 bytes allocated will result in an error when you eventually call the [`InitializeMint` ](https://github.com/solana-labs/solana-program-library/blob/d72289c79a04411c69a8bf1054f7156b6196f9b3/token/program-2022/src/processor.rs#L54) instruction on the Token22 program.
  * An `InvalidAccountData` error is thrown because the[ program checks that the size of the given account matches exactly what is expected for the given extension.](https://github.com/solana-labs/solana-program-library/blob/d72289c79a04411c69a8bf1054f7156b6196f9b3/token/program-2022/src/processor.rs#L77)
* The correct amount of bytes is 235. Allocating a buffer of exactly 235 bytes when creating the account will pass this check.
  * You can even test this by running the commandline example in the Confidential Transfer [quick start guide](https://spl.solana.com/confidential-token/quickstart) and inspecting the transaction in an explorer.
``` bash
spl-token --program-id TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb create-token --enable-confidential-transfers auto
```
* Here is an [example transaction on Devnet.](https://explorer.solana.com/tx/FHwhr65hYaPBfwNx5LM4cXKSnovSGxG5njZENm58a86GVKuKGngTn42SBSUeGXDXLSrEmhxHtRTiXXQ84LFxGdf?cluster=devnet) You can see the account is 235 bytes long.

[...]

PR #6487, opened 2024-03-25
[token-js] Fix confidential transfer extensions account length

samkim-crypto:
#### Problem
The confidential transfer extension in token-js has not been updated for a while. Currently, the wrong lengths for the confidential transfer mint and accounts are hard-coded in. Although confidential transfers are not yet supported in token-js, this could be quite confusing (e.g. https://github.com/solana-labs/solana-program-library/issues/6338).

#### Summary of changes
Hard-code the mint and account lengths with the correct number. These should be replaced with the `confidentialTransferMint.span` and `confidentialTransferAccount.span` as soon as wasm support for zk-token-sdk lands.


joncinque (approved):
Thanks for fixing this! I confirm that the numbers are correct from my own local testing

Issue #6492, opened 2024-03-26
token-cli: Add priority fee support


joncinque:
#### Problem

With all the congestion on mainnet, it's tough for people to land transactions using the token-cli, since it doesn't expose the ability to add priority fees.

#### Proposed Solution

There are a few phases to this:

- [x] #6494
- [x] #6493
- [x] #6500
- [x] #6501

joncinque:
This is all done

PR #6493, opened 2024-03-26
token-client: Add compute unit price and limit fields

joncinque:
#### Problem

The token-cli doesn't support priority fees.

#### Solution

As the first part of #6492, add the ability to specify a compute unit limit on the token-client directly.

Note: Some of the parts of setting the compute unit limit by hand in the tests can be removed once we dynamically calculate the compute unit limit based on simulation, for part 2 of #6492


CriesofCarrots (approved):
So much cleaner too :)

PR #6494, opened 2024-03-26
token-client: Refactor `Message` creation to use the same `instructions` everywhere

joncinque:
#### Problem

Currently, the token client uses `Message::new_with_nonce` for nonce transactions, which means that we need to do some copy-pasta on the `instructions` before simulating to find the compute unit limit.

#### Solution

Extract the part from `Message::new_with_nonce` that adds an instruction to the transaction.

This actually fixes a bug where `Message::new_with_nonce` was taking `token_instructions` instead of `instructions`.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -541,32 +541,25 @@ where
             );
         }
 
-        let (message, blockhash) =
-            if let (Some(nonce_account), Some(nonce_authority), Some(nonce_blockhash)) = (
-                self.nonce_account,
-                &self.nonce_authority,
-                self.nonce_blockhash,
-            ) {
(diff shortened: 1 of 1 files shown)


token/client/src/token.rs
```diff
@@ -541,32 +541,25 @@ where
...
         }
 
-        let (message, blockhash) =
-            if let (Some(nonce_account), Some(nonce_authority), Some(nonce_blockhash)) = (
-                self.nonce_account,
-                &self.nonce_authority,
-                self.nonce_blockhash,
-            ) {
-                let mut message = Message::new_with_nonce(
-                    token_instructions.to_vec(),
```
CriesofCarrots, line 551: 🤕 Nice catch

PR #6499, opened 2024-03-26
stake-pool-cli: Add support for priority fees

joncinque:
#### Problem

Due to network congestion mainnet and the lack of priority fee support, stake pool users can't interact with their pool on mainnet.

#### Solution

A few things to provide a full solution:

* Add `--with-compute-unit-price` flag to set micro-lamports per CU for all transactions in a command
* Add `--with-compute-unit-limit` flag to set the number of CUs for all transactions in a command
* If `--with-compute-unit-limit` is not set, simulate the transaction and use the number of compute units reported

As part of this, the idea is to funnel all transaction creation through `checked_transaction_with_signer` or `checked_transaction_with_signers_and_additional_fee`, which will do the simulation work and add the compute budget instructions.

Since there is some refactor, it might be easier to go through this commit by commit.

### DIFF
--- a/stake-pool/cli/src/client.rs
+++ b/stake-pool/cli/src/client.rs
@@ -153,9 +153,6 @@ pub(crate) fn get_all_stake(
 
 /// Helper function to add a compute unit limit instruction to a given set
 /// of instructions
-///
-/// Returns true if the instruction was added, false if it wasn't. The false
-/// case is used for offline signing, where we cannot access the network.
 pub(crate) fn add_compute_unit_limit_from_simulation(
     rpc_client: &RpcClient,
     instructions: &mut Vec<Instruction>,
--- a/stake-pool/cli/src/client.rs
+++ b/stake-pool/cli/src/client.rs
@@ -7,7 +7,11 @@ use {
         rpc_config::{RpcAccountInfoConfig, RpcProgramAccountsConfig},
         rpc_filter::{Memcmp, RpcFilterType},
     },
-    solana_program::{borsh1::try_from_slice_unchecked, program_pack::Pack, pubkey::Pubkey, stake},
+    solana_program::{
+        borsh1::try_from_slice_unchecked, hash::Hash, instruction::Instruction, message::Message,
+        program_pack::Pack, pubkey::Pubkey, stake,
+    },
+    solana_sdk::{compute_budget::ComputeBudgetInstruction, transaction::Transaction},
     spl_stake_pool::{
         find_withdraw_authority_program_address,
         state::{StakePool, ValidatorList},
@@ -15,7 +19,7 @@ use {
     std::collections::HashSet,
 };
 
-type Error = Box<dyn std::error::Error>;
+pub(crate) type Error = Box<dyn std::error::Error>;
 
 pub fn get_stake_pool(
     rpc_client: &RpcClient,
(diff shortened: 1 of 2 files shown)


stake-pool/cli/src/client.rs
```diff
@@ -146,3 +150,38 @@ pub(crate) fn get_all_stake(
         .map(|(address, _)| address)
         .collect())
 }
+
+/// Helper function to add a compute unit limit instruction to a given set
+/// of instructions
+///
+/// Returns true if the instruction was added, false if it wasn't. The false
+/// case is used for offline signing, where we cannot access the network.
```
2501babe, line 158: think this function was adapted from elsewhere and not the comment? since it doesnt return a bool and this cli doesnt appear to have offline signing

stake-pool/cli/src/client.rs
```diff
@@ -146,3 +150,38 @@ pub(crate) fn get_all_stake(
...
+    let simulation_result = rpc_client.simulate_transaction(&transaction)?.value;
+    let units_consumed = simulation_result
+        .units_consumed
+        .ok_or("No units consumed on simulation")?;
+    // Overwrite the compute unit limit instruction with the actual units consumed
+    let compute_unit_limit = u32::try_from(units_consumed)?;
+    instructions
+        .last_mut()
+        .expect("Compute budget instruction was added earlier")
+        .data = ComputeBudgetInstruction::set_compute_unit_limit(compute_unit_limit).data;
```
2501babe, line 185: should we add a bit for fudge factor in case cu cost is nondeterministic? not sure how likely this is. if theres a complex struct that needs to be parsed and it changes before the transaction lands, it could be possible in theory

stake-pool/cli/src/main.rs
```diff
@@ -1999,6 +2004,15 @@ fn main() {
...
         )
+        .arg(compute_unit_price_arg().validator(is_parsable::<u64>))
+        .arg(
+            Arg::with_name(COMPUTE_UNIT_LIMIT_ARG.name)
+                .long(COMPUTE_UNIT_LIMIT_ARG.long)
+                .takes_value(true)
+                .value_name("COMPUTE-UNIT-LIMIT")
+                .help(COMPUTE_UNIT_LIMIT_ARG.help)
+                .validator(is_parsable::<u32>)
+        )
```
2501babe, line 2015: these need to be `.global(true)`, they dont show up for subcommands when running the cli

stake-pool/cli/src/client.rs
joncinque: Oops, yep that was some bad copy-pasta, removed that bit

stake-pool/cli/src/client.rs
joncinque: Yeah I've thought about this a bit... we could do this through a new variant of `--with-compute-unit-limit` where you can say something like `SIMULATED+10` which means "add 10 to whatever the simulation results say". I'm working on something for the token cli towards that direction.

My thinking was to start with this, and once people complain about it, we can add it.

stake-pool/cli/src/main.rs
joncinque: Ah yes, thanks! It turns out I misunderstood how `global` worked this whole time

stake-pool/cli/src/client.rs
2501babe: if it was a matter of poor ux for a new feature i would agree, but this

```rust
    if let Some(compute_unit_limit) = config.compute_unit_limit {        
        instructions.push(ComputeBudgetInstruction::set_compute_unit_limit(        
            compute_unit_limit,        
        ));        
    } else {                       
        add_compute_unit_limit_from_simulation(        
            &config.rpc_client,        
            &mut instructions,                                                      
            &config.fee_payer.pubkey(),        
            &recent_blockhash,        
        )?;        
    }
```

adds a simulated compute limit to transactions when the flag _isnt_ used, making existing flows more brittle. perhaps this was unintentional?

joncinque: It was intentional, but then I've gone back the other way with #6550 -- what do you think about that kind of approach?

TLDR you have `--with-compute-unit-limit`, you can specify `SIMULATED` or a number. If it's not specified, it doesn't use anything. But if you specify `--with-compute-unit-price` without `--with-compute-unit-limit`, it fails. What do you think about that approach?

2501babe: i like the design you have there for sure!

joncinque: Ok I'll do that here too then

joncinque:
Ok, I've updated this to be similar to the token CLI work, let me know how it looks!

2501babe:
missed your latest push, tested create pool by hand and with no flags it doesnt add compute ixns, with budget flag it demands limit, budget with limit simulated works, budget with limit with number works, budget alone works

2501babe (approved):

PR #6500, opened 2024-03-26
token-client: Use transaction simulation results for the compute unit limit

joncinque:
#### Problem

As part of #6492, the token-cli doesn't support priority fees. Currently, the token-client allows for setting a compute unit price and limit, but that requires users to know the compute units used by their transaction.

#### Solution

Transaction simulation gives the compute units used by a transaction, so instead of asking users to provide a number, we can be a bit smarter and use simulation results to set the compute unit limit used by a transaction.

We also need to support offline signing, which cannot simulate transactions. In this PR, the token client fails to extract the compute units used by the transaction after the offline "simulation", and thus uses the default limit of 200k * num_top_level_instructions.

A user-provided value for `compute_unit_limit` always takes precedence, so offline signing can use a hardcoded number if a compute unit limit is desired.

Note: the range proofs for confidential transfers are so close to the transaction size limit that we can't add any other instructions to the transaction, so those are created and sent without the helpers

### DIFF
--- a/token/client/src/client.rs
+++ b/token/client/src/client.rs
@@ -20,7 +20,12 @@ pub trait SendTransaction {
 
 /// Basic trait for simulating transactions in a validator.
 pub trait SimulateTransaction {
-    type SimulationOutput;
+    type SimulationOutput: SimulationResult;
+}
+
+/// Trait for the output of a simulation
+pub trait SimulationResult {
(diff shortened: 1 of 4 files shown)


CriesofCarrots (approved):
Lgtm!

PR #6507, opened 2024-03-27
bump: Bump everything up for new token cli

joncinque:
#### Problem

We need to get a new token cli release out to go with the next 1.18 release, since it includes priority fees.

#### Solution

Bump everything needed for token-cli:

* token-cli
* token-client
* token-2022 (major because of breaking changes)
* associated-token-account (major because of new borsh version)
* pod (needed for new constructors)

I was worried that the major version bump on token-2022 would be an absolute horror, but it was pretty ok!

### DIFF
--- a/token/transfer-hook/cli/Cargo.toml
+++ b/token/transfer-hook/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 authors = ["Solana Labs Maintainers <maintainers@solanalabs.com>"]
-description = "SPL-Token Command-line Utility"
+description = "SPL Transfer Hook Command-line Utility"
 edition = "2021"
 homepage = "https://spl.solana.com/token"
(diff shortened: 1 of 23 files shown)


token/transfer-hook/cli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 authors = ["Solana Labs Maintainers <maintainers@solanalabs.com>"]
-description = "SPL-Token Command-line Utility"
+description = "SPL Transfer Hook Command-line Utility"
```
buffalojoec, line 3: 👀

buffalojoec (approved):
bump

joncinque:
Bah never mind, borsh messes this up so it's going to require a lot more bumps

PR #6513, opened 2024-03-28
discriminator: Bump to 0.2.0

joncinque:
#### Problem

The discriminator library is still using borsh 0.10, which is causing very strange failures when trying to publish.

#### Solution

Bump to borsh 1 and its version to 0.2.0


joncinque:
Test error is unrelated, being fixed in #6514

PR #6515, opened 2024-03-28
Bump all crates for token-cli release

joncinque:
#### Problem

The bumps for #6507 ended up restricting the available versions too much, making it impossible to publish spl-token-cli.

#### Solution

Relax all of the dependencies and bump all the required crates:

* pod
* tlv
* tlv-account-resolution
* token-group-interface
* token-metadata-interface
* token-2022
* token-client
* transfer-hook-interface
* associated-token-account
* discriminator


joncinque:
Never mind, this didn't work

PR #6523, opened 2024-03-29
token-client: Make `Token` properly `Send + Sync`

### LINKED ISSUE #6478: `Token` type is `!Send + !Sync`

joncinque:
#### Problem

It's not possible to use token-client's `Token` in a detached async context because not all the types contained are `Send + Sync`. See #6510 for more info.

#### Solution

Add `Send + Sync` to all types contained within `Token` and all type parameters. This is a pretty invasive change and breaks current CLI users, but we can consider doing this in a breaking release.

Fixes #6478

cc @1500256797

### DIFF
--- a/token/client/src/client.rs
+++ b/token/client/src/client.rs
@@ -192,7 +192,7 @@ pub type ProgramClientResult<T> = Result<T, ProgramClientError>;
 
 /// Generic client interface for programs.
 #[async_trait]
-pub trait ProgramClient<ST>
+pub trait ProgramClient<ST>: Send + Sync
 where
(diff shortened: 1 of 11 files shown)


celernetworkcoloros:
Token type is !Send + !Sync

PR #6527, opened 2024-04-01
fix: ignore single pool dist

qiweiii:
when run build in project root, `single-pool`'s build results are in `dist` directory, they are not ignored, I need to revert them everytime, to avoid this, just add gitignore for them.


joncinque (approved):
Thanks for contributing!

Issue #6542, opened 2024-04-03
single-pool: Remove usage of stake::config


joncinque:
#### Problem

The `stake::config` account has been fully and removed with https://github.com/anza-xyz/agave/pull/470, but the single pool program still uses and relies on it.

#### Solution

Once the 2.0 crates are available and SPL uses them, remove all references to and usage of `stake::config`.

PR #6543, opened 2024-04-03
single-pool: Allow stake config to be anything in tests

joncinque:
#### Problem

The stake config is getting removed with https://github.com/anza-xyz/agave/pull/470, but this causes the single pool tests to fail, since it's expecting that the program will *fail* if any account (including stake config), is arbitrarily changed.

#### Solution

The full fix will come with https://github.com/solana-labs/solana-program-library/issues/6542, but to unblock the PR on Agave, just allow that account to be anything.


joncinque:
Merging early to unblock

PR #6549, opened 2024-04-04
Remove usage of VoteStateVersions::vote_state_size_of

AshwinSekar:
The feature flag `vote_state_add_vote_latency` has been activated on all clusters, so the old vote state version is no longer supported. 
As such, consumers should use `VoteState::size_of()` instead of `VoteStateVersions::vote_state_size_of(true)`.

Unblocks the cleanup of this feature flag https://github.com/anza-xyz/agave/pull/468


joncinque (approved):
Thanks for taking care of this! cc @2501babe for awareness

PR #6550, opened 2024-04-04
token-cli: Fix offline signing by making simulated compute unit limit explicit

joncinque:
#### Problem

https://github.com/solana-labs/solana-program-library/pull/6501 broke offline signing -- by default, the token client sets the compute unit limit based on simulation. If it's using offline signing, it'll silently swallow the error and pop the compute unit limit instruction.

So when we offline-sign a transaction, it won't have the compute unit limit instruction, and when we broadcast it, we'll simulate and add the compute unit limit instruction, causing a pre-signer error.

This was the classic overly-clever-but-brittle solution. I discovered it while applying the token-cli changes to the Solana CLI.

#### Solution

Be more explicit:

* don't simulate by default
* allow users to specify `--with-compute-unit-limit SIMULATED` if they want that
* error if someone sets `--with-compute-unit-price` *without* `--with-compute-unit-limit`. This one might be contentious, and I could be talked down to making it a warning. But since we've only just introduced the flag, I think it's better to do the right thing and force people to be explicit
* improve the offline signing test by actually sending the transaction so we catch this in the future

Let me know how you like the `ComputeUnitLimit`, because I'll likely do the same exact thing for the Solana CLI. Except there, `--with-compute-unit-price` has been around for some time, so I would show a warning if someone specifies price without a limit.

### DIFF
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -1,8 +1,5 @@
 use {
-    crate::clap_app::{
-        parse_compute_unit_limit, Error, COMPUTE_UNIT_LIMIT_ARG, COMPUTE_UNIT_PRICE_ARG,
-        MULTISIG_SIGNER_ARG,
-    },
+    crate::clap_app::{Error, COMPUTE_UNIT_LIMIT_ARG, COMPUTE_UNIT_PRICE_ARG, MULTISIG_SIGNER_ARG},
     clap::ArgMatches,
     solana_clap_utils::{
         input_parsers::{pubkey_of_signer, value_of},
@@ -286,12 +283,29 @@ impl<'a> Config<'a> {
                 (default_program_id, false)
             };
 
+        // need to specify a compute limit if compute price and blockhash are specified
+        if matches.is_present(BLOCKHASH_ARG.name)
+            && matches.is_present(COMPUTE_UNIT_PRICE_ARG.name)
+            && !matches.is_present(COMPUTE_UNIT_LIMIT_ARG.name)
+        {
+            eprintln!(
+                "error: need to set {} if {} and {} are set",
+                COMPUTE_UNIT_LIMIT_ARG.name, COMPUTE_UNIT_PRICE_ARG.name, BLOCKHASH_ARG.name,
+            );
+            exit(1);
+        }
+
         let nonce_blockhash = value_of(matches, BLOCKHASH_ARG.name);
         let compute_unit_price = value_of(matches, COMPUTE_UNIT_PRICE_ARG.name);
-        let compute_unit_limit = matches
-            .value_of(COMPUTE_UNIT_LIMIT_ARG.name)
-            .map(|x| parse_compute_unit_limit(x).unwrap())
-            .unwrap_or(ComputeUnitLimit::Default);
+        let compute_unit_limit = value_of(matches, COMPUTE_UNIT_LIMIT_ARG.name)
+            .map(ComputeUnitLimit::Static)
+            .unwrap_or_else(|| {
+                if nonce_blockhash.is_some() {
+                    ComputeUnitLimit::Default
+                } else {
+                    ComputeUnitLimit::Simulated
+                }
+            });
         Self {
             default_signer,
             rpc_client,
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -288,11 +288,14 @@ impl<'a> Config<'a> {
             && matches.is_present(COMPUTE_UNIT_PRICE_ARG.name)
             && !matches.is_present(COMPUTE_UNIT_LIMIT_ARG.name)
         {
-            eprintln!(
-                "error: need to set {} if {} and {} are set",
-                COMPUTE_UNIT_LIMIT_ARG.name, COMPUTE_UNIT_PRICE_ARG.name, BLOCKHASH_ARG.name,
-            );
-            exit(1);
+            clap::Error::with_description(
+                &format!(
+                    "Need to set `{}` if `{}` and `--{}` are set",
+                    COMPUTE_UNIT_LIMIT_ARG.long, COMPUTE_UNIT_PRICE_ARG.long, BLOCKHASH_ARG.long,
+                ),
+                clap::ErrorKind::MissingRequiredArgument,
+            )
+            .exit();
         }
 
         let nonce_blockhash = value_of(matches, BLOCKHASH_ARG.name);
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -1,5 +1,8 @@
 use {
-    crate::clap_app::{Error, COMPUTE_UNIT_LIMIT_ARG, COMPUTE_UNIT_PRICE_ARG, MULTISIG_SIGNER_ARG},
+    crate::clap_app::{
+        parse_compute_unit_limit, Error, COMPUTE_UNIT_LIMIT_ARG, COMPUTE_UNIT_PRICE_ARG,
+        MULTISIG_SIGNER_ARG,
+    },
     clap::ArgMatches,
     solana_clap_utils::{
         input_parsers::{pubkey_of_signer, value_of},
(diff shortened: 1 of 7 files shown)


CriesofCarrots:
Sorry for the slow review here! I'm a little confused. If it's expected that a user would use `SIMULATED` with offline signing, it seems like the `--sign-only` command would need to return the results of simulation so the values could be made specific in the subsequent submission command. Otherwise, there's a risk the signature is invalid if those values change between signing and submission (as you note in the help text). This is a bit gross to me, although we could let the user take that risk (but places more support burden on us).

If it's *not* expected that a user would use `SIMULATED` with offline signing, I think we need to prevent it. And then I don't think the keyword really offers a lot of value, and we should default to simulation in the online case, and require offline signing to be explicit from the get-go. (This is an existing approach/pattern in solana-cli, incidentally, eg split-stake: https://github.com/anza-xyz/agave/blob/180a186c7da1b8d6f5e93ec2cbc5bf2ea17db2cb/cli/src/stake.rs#L551-L560)

joncinque:
> If it's not expected that a user would use SIMULATED with offline signing, I think we need to prevent it

That's the way I was thinking about it. We could eventually create a way for someone to simulate a transaction just to get the compute units and pass it in explicitly for offline signing, but that would be a separate PR.

> And then I don't think the keyword really offers a lot of value, and we should default to simulation in the online case, and require offline signing to be explicit from the get-go

Just so I'm sure I understand: require `compute_unit_limit` if both `sign_only` and `compute_unit_price` are set, and then any instruction can also use `compute_unit_limit`. That would fix the issue.

The problem I'm running into is when to use the simulated compute units. Currently, the code *always* simulates, but that's an issue when one of the signers is a pre-signer. So we're left with only simulating for compute units if `compute_unit_price` is set and `compute_unit_limit` isn't, which will work in all cases.

However, this means it isn't possible to specify, "please simulate my transaction and use the compute unit limit from the result", you're required to also set a compute unit price. Someone could get around this by specifying a compute unit price of 0, but that seems confusing. So that's why I added the `SIMULATED` keyword.

Do you think that's a worthwhile use case? I think it is, since the scheduler should have an easier time packing transactions with (low) compute unit limits. But maybe I'm wrong.

CriesofCarrots:
> Just so I'm sure I understand: require `compute_unit_limit` if both `sign_only` and `compute_unit_price` are set, and then any instruction can also use `compute_unit_limit`. That would fix the issue.

Yep, this is what I meant. I can't quite tell from your comment, what do you think of that approach?

joncinque:
> I can't quite tell from your comment, what do you think of that approach?

Sorry, I'm having a hard time articulating this. Your idea fixes the issue, but there's a knock-on effect: it *requires* someone to specify a compute unit price in order to use the simulated compute unit limit.

I wanted to allow people to specify that they want to use simulated compute units without specifying a compute unit price, which is why I added the `SIMULATED` variant. Do you think that's worth it?

CriesofCarrots:
> it requires someone to specify a compute unit price in order to use the simulated compute unit limit.

I must be being stupid, because I don't follow why this is required. Let me re-read the code, more closely, and then reply again.

token/cli/src/config.rs
```diff
@@ -282,7 +288,10 @@ impl<'a> Config<'a> {
 
         let nonce_blockhash = value_of(matches, BLOCKHASH_ARG.name);
         let compute_unit_price = value_of(matches, COMPUTE_UNIT_PRICE_ARG.name);
-        let compute_unit_limit = value_of(matches, COMPUTE_UNIT_LIMIT_ARG.name);
+        let compute_unit_limit = matches
+            .value_of(COMPUTE_UNIT_LIMIT_ARG.name)
+            .map(|x| parse_compute_unit_limit(x).unwrap())
+            .unwrap_or(ComputeUnitLimit::Default);
```
CriesofCarrots, line 294: So this is my thought for how to handle without a special "SIMULATED" input:

```suggestion
        let compute_unit_limit = matches
            .value_of(COMPUTE_UNIT_LIMIT_ARG.name);
        if compute_unit_limit.is_none() && compute_unit_price.is_some() && (sign_only || nonce_blockhash.is_some()) {
            <return clap error that compute_unit_limit must be specified> // Possibly this if case could be elimintated by using clap `requires()` configs in the App
        }
        let compute_unit_limit = compute_unit_limit
            .map(|x| parse_compute_unit_limit(x).unwrap())
            .unwrap_or_else(|| if sign_only || nonce_blockhash.is_some() { ComputeUnitLimit::Default } else { ComputeUnitLimit::Simulate });
```
(Sorry about the poor formatting.)
Wdyt?

joncinque: That works for me! Just so I'm sure I understand, this still allows for `SIMULATED` as an option, but is smarter about when to default to simulating transactions, correct?

What's the rationale for checking `sign_only`? Is the idea to support offline signing *without* a nonce? That will still fail to broadcast since it'll try to simulate during the actual send by default.

CriesofCarrots: > this still allows for `SIMULATED` as an option, but is smarter about when to default to simulating transactions, correct?

No, my thinking was to eliminate the `SIMULATED` input option, but keep the `ComputeUnitLimit::Simulate` variant to use for the "right" transactions.

> What's the rationale for checking sign_only? Is the idea to support offline signing without a nonce? That will still fail to broadcast since it'll try to simulate during the actual send by default.

The rationale was to prevent simulation for both the offline steps, ie: (1) signing command, which has `--sign-only`; and (2) the send-tx-with-presigner command, which must have `--blockhash` ([link](https://github.com/anza-xyz/agave/blob/4672efa1c1cedcf0f034a62b8b72fb1aa0603e3a/clap-utils/src/offline.rs#L53)).
But looking at the clap-utils code again, I see checking `--blockhash` should be sufficient for both cases, since `--sign-only` does require `--blockhash` as well. Sorry about making that suggestion more complicated than it needs to be.

joncinque: Thanks for the clarification, works for me!

joncinque:
I added some more testing and integrated the changes your requested. Unfortunately, clap v2 doesn't expose a way to say "require this arg if these two other args are set", [required_if](https://docs.rs/clap/2.34.0/clap/struct.Arg.html#method.required_if) only checks if an arg is set to a particular value, and [requires_all](https://docs.rs/clap/2.34.0/clap/struct.Arg.html#method.requires_all) is not conditional

token/cli/src/config.rs
```diff
@@ -280,9 +283,29 @@ impl<'a> Config<'a> {
...
+        // need to specify a compute limit if compute price and blockhash are specified
+        if matches.is_present(BLOCKHASH_ARG.name)
+            && matches.is_present(COMPUTE_UNIT_PRICE_ARG.name)
+            && !matches.is_present(COMPUTE_UNIT_LIMIT_ARG.name)
+        {
+            eprintln!(
+                "error: need to set {} if {} and {} are set",
+                COMPUTE_UNIT_LIMIT_ARG.name, COMPUTE_UNIT_PRICE_ARG.name, BLOCKHASH_ARG.name,
+            );
+            exit(1);
```
CriesofCarrots, line 295: Nit: this works, but it might be nice to use a clap error to emulate clap `requires` handling with the error output. Similar to what we do here: https://github.com/anza-xyz/agave/blob/ccb0b55d82fa40041cfc4d346ff50486c7195b7a/validator/src/main.rs#L911-L916
Wdyt?

CriesofCarrots (commented):
Looks great! Thanks for the go-around on this.
One nit.
Also, is the CI failure expected? It doesn't look related, but did happen twice.

token/cli/src/config.rs
joncinque: Great idea, done!

joncinque:
>  Also, is the CI failure expected? It doesn't look related, but did happen twice.

It was a legit failure, the override limit that was set for `set-interest-rate` was being clobbered

CriesofCarrots (approved):
Woot! :shipit:

Issue #6577, opened 2024-04-14
Optimize Initialization Process in Concurrent Merkle Tree


silentEAG:
**Description:**

In the file `libraries/concurrent-merkle-tree/src/concurrent_merkle_tree.rs` at line 109, an `empty_node_cache` is defined but it is immutable. Even though `empty_node_cached` generates `rightmost_proof` and `path`, it cannot utilize the cache when `cache[target] == EMPTY`. My understanding is that `empty_node_cache` should be able to store the results from previous levels and use them for generating the current level, thereby avoiding recursive calls.

**Verification:**
To verify the issue, insert `println!("level: {}", level)` in the `empty_node_cached` func to output the current level. Then, run the following test code:

```rust
#[tokio::test(flavor = "multi_thread")]
async fn verify_it() {
    let mut tree1 = ConcurrentMerkleTree::<DEPTH, BUFFER_SIZE>::new();
    let _ = tree1.initialize().unwrap();
}
```

For better observation of the output, insert a line after each node generation:

```rust
for (i, node) in path.iter_mut().enumerate() {
	*node = empty_node_cached::<MAX_DEPTH>(i as u32, &empty_node_cache);
	println!("");
}
```

When running the `cargo test` with stdout, you will observe:

```text
level: 0

level: 1
level: 0

level: 2
level: 1
level: 0

level: 3
level: 2
level: 1
level: 0

level: 4
level: 3
level: 2
level: 1
level: 0

...
```

This pattern introduces an O(n(n+1)/2) computational overhead which is unnecessary.

PR #6578, opened 2024-04-14
fix `empty_node_cache` use for cache

silentEAG:
Try to solve #6577 
update  `empty_node_cache` promptly.


ngundotra:
ty for posting this, this will save a bunch of CU on mainnet!

How did you find?

silentEAG:
Hello! I am a smart contract auditor from Sec3. During an audit, I needed to study the theory and implementation of the Solana Concurrency Merkle Tree. While reading the code implementation, I stumbled upon this small issue that I hope can help SolanaLabs :)

I noticed that there are four instances in the code where `empty_node_cached` is used directly without utilizing `empty_node_cache`. Two of these in `initialize` can indeed use the cache, and they have already been addressed by the previous PR. The remaining two instances are:

1. [concurrent_merkle_tree.rs#L172](https://github.com/solana-labs/solana-program-library/blob/master/libraries/concurrent-merkle-tree/src/concurrent_merkle_tree.rs#L172): This does not involve a loop calculation, so it doesn't need to use the cache and can be ignored.
2. [concurrent_merkle_tree.rs#L277](https://github.com/solana-labs/solana-program-library/blob/master/libraries/concurrent-merkle-tree/src/concurrent_merkle_tree.rs#L277): This involves a loop calculation and could potentially benefit from calculating the cache first and then iterating. However, in most cases, the `intersection` is relatively small, so the performance gain in terms of CU would be insignificant. I think this can also be ignored.

In my latest commit, I have fixed the issues identified by cargo clippy. Thank you for your review!

ngundotra:
Going to merge & redeploy with this in mid June when back.

ngundotra:
Closing loop on this - this is getting merged in a slightly different form into #6764

PR #6588, opened 2024-04-17
chore: fix some comments

hidewrong:


buffalojoec (approved):
Thanks! That's a lot of `doesn't`s!

count

PR #6592, opened 2024-04-17
program error: bump patch version

buffalojoec:
Bumps the patch version of program error after #6590.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7015,7 +7015,7 @@ dependencies = [
  "solana-program",
  "solana-program-test",
  "solana-sdk",
- "spl-program-error 0.4.0",
+ "spl-program-error 0.4.1",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


joncinque (commented):
Unless they need the new version, can we keep the downstream deps on 0.4.0? I got screwed pretty badly by the patch releases last time and wasted a whole day getting the deps to work out for token-cli because of patch bumps.

buffalojoec:
> Unless they need the new version, can we keep the downstream deps on 0.4.0? I got screwed pretty badly by the patch releases last time and wasted a whole day getting the deps to work out for token-cli because of patch bumps.

Oh yeah sure, but I think they'd have to get modified to say `"4.0"` since they use `path`, right?

joncinque:
As long as we keep it as `0.4.0` everywhere, it should pull in either. At least, it worked with my local testing

joncinque (approved):
Thanks!

PR #6596, opened 2024-04-18
chore: remove repetitive words

careworry:


buffalojoec (approved):
bullseye

governance/program/src/error.rs
```diff
@@ -357,8 +357,8 @@ pub enum GovernanceError {
     #[error("Invalid proposal options")]
     InvalidProposalOptions, // 583
 
-    /// Proposal is not not executable
-    #[error("Proposal is not not executable")]
```
steveluscher, line 361: 🤣

PR #6604, opened 2024-04-22
Add the missing single quota mark in words

forcedebug:
Add the missing single quota mark in words


joncinque (approved):
Thanks for fixing the typos!

PR #6611, opened 2024-04-22
Feature Gate: Drop from SPL

buffalojoec:
#### Problem
Now that Feature Gate has been evicted from SPL and lives alongside its
Core BPF brethren in the Solana Program org, there's no need to maintain the
program here in SPL.

#### Solution
Sayonara!


joncinque (approved):
We can probably outright remove this one since it's pretty new and was never deployed, but we'll have to think about the best move for the other ones.

Looking at what metaplex did, I see that they used a README that just points to the new repo: https://github.com/metaplex-foundation/metaplex-program-library/tree/master/token-metadata

That seems pretty reasonable to me. What do you think?

buffalojoec:
> We can probably outright remove this one since it's pretty new and was never deployed, but we'll have to think about the best move for the other ones.
> 
> Looking at what metaplex did, I see that they used a README that just points to the new repo: https://github.com/metaplex-foundation/metaplex-program-library/tree/master/token-metadata
> 
> That seems pretty reasonable to me. What do you think?

Yeah, I think that's probably the best way to do it! A monorepo of README's!

PR #6614, opened 2024-04-22
name-service-js: add `.prettierignore` and `.eslintignore`

buffalojoec:
#### Problem
When linting against `name-service-js`, the build target directory, `lib`, can get
included in linting, since there's no ignore rules set for prettier.

#### Summary of Changes
Add some ignore rules!

### DIFF
--- a/name-service/js/.eslintignore
+++ b/name-service/js/.eslintignore
@@ -1,0 +1,5 @@
+docs
+lib
+test-ledger
+
+package-lock.json
--- a/name-service/js/.prettierignore
+++ b/name-service/js/.prettierignore
@@ -1,0 +1,5 @@
+docs
+lib
+test-ledger
+
+package-lock.json


name-service/js/.eslintignore
```diff
@@ -0,0 +1,5 @@
+docs
+lib
+test-ledger
+
+package-lock.json
```
joncinque, line 5: there's no package-lock.json, so maybe we don't need this line?

name-service/js/.prettierignore
```diff
@@ -0,0 +1,5 @@
+docs
+lib
+test-ledger
+
+package-lock.json
```
joncinque, line 5: there's no package-lock.json, so maybe we don't need this line?

joncinque (approved):
Thanks for doing this! Just a little suggestion

name-service/js/.eslintignore
buffalojoec: We use this in token-js as well, for anyone who might do `npm i`. I figured best to follow pattern?

joncinque: ah gotcha, that's fine then, let's leave it in

PR #6635, opened 2024-04-25
stake-pool-cli: Default to simulated compute units

joncinque:
#### Problem

#6550 fixed the token-cli for offline signing, but still defaults to using simulated compute units, while #6499 in the stake-pool-cli forces people to specify that they want to use `SIMULATED` compute units, and otherwise adds nothing. This is inconsistent.

#### Solution

Make the stake pool CLI consistent with the token CLI by defaulting to simulated compute units. On the other hand, this PR allows people to specify that they want `DEFAULT` compute units.

Let me know what you think!

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -114,19 +114,21 @@ const COMPUTE_UNIT_LIMIT_ARG: ArgConstant<'static> = ArgConstant {
     name: "compute_unit_limit",
     long: "--with-compute-unit-limit",
     help: "Set compute unit limit for transaction, in compute units; also accepts \
-        keyword SIMULATED to use compute units from transaction simulation prior \
-        to sending. Note that SIMULATED may fail if accounts are modified by another \
-        transaction between simulation and execution.",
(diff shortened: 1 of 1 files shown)


2501babe:
i like the latest ux the most! unfortunately i ran into two issues:
* now, proving no compute flags at all introduces a simulated compute budget; i believe this should be default ie no instruction
* there is no way to use a simulated budget with a price because the budget flag is required and `SIMULATE` is no longer a valid option

this is what i believe the grammar should look like:
* no price flag, no budget flag -> no instructions
* no price flag, budget flag with default -> no instructions
* no price flag, budget flag with number -> fixed budged instruction
* price flag, no budget flag -> price instruction, simulated budget instruction
* price flag, budget flag with default -> price instruction, no budget instruction
* price flag, budget flag with number -> price instruction, fixed budget instruction

budget options remain number or `DEFAULT`. remove the requirement for a budget flag if price flag is given

ive pushed a commit that implements this logic, lmk what you think!

joncinque:
That makes sense to me, thanks for doing the work! I'll do the same thing with the monrepo CLI

PR #6636, opened 2024-04-25
token-client: Bump for token-cli release

joncinque:
#### Problem

We want to publish a new token-cli, but token-client had a breaking change with the new `ComputeUnitLimit` enum.

#### Solution

Bump to 0.10.0 for the next release.


Tq101510:
Issues[crypto.com]()

PR #6643, opened 2024-04-26
transfer-hook-cli: patch token arg

buffalojoec:
#### Problem
The arg for the token mint address in the Transfer Hook CLI's `create-extra-metas`
and `update-extra-metas` commands was searching for `program_id`. This causes
the program ID to be passed to the transfer hook program, when instead it should
be the mint address!

#### Summary of Changes
Fix the parameter parsers!


joncinque (approved):

PR #6665, opened 2024-04-30
token-group-js: Update to tp3

joncinque:
#### Problem

web3.js TP3 is out, but the token-group JS library isn't using it.

#### Solution

Upgrade to the new lib and use it. Let me know if I'm following the right conventions! I looked at example usage from the repo directly for inspiration.


lorisleiva (dismissed):
Perfect! 👌

PR #6686, opened 2024-05-05
Update instruction.rs

SynergyCipher:
Handle potential serialization errors gracefully rather than using unwrap(), which could cause the program to panic at runtime.


buffalojoec:
@SynergyCipher thanks for the contribution! However, the change you've added is still a panic.

PR #6701, opened 2024-05-07
add cargo script

buffalojoec:
#### Problem
We've got a nice little script for using the correct version of `cargo +nightly`,
but no such script for stable!

#### Summary of Changes
Add the stable script!


buffalojoec:
> Since we have `rust-toolchain.toml`, you just need to run `cargo`

Bahh I was wondering if it was necessary. Ok then!

PR #6702, opened 2024-05-07
token-group: hash error codes

buffalojoec:
#### Problem
We decided a while back to use an error code discriminator, derived from the
name of the error enum, to offset error codes that may arise in Token-2022.

We applied this concept to a few libraries and the Token Metadata interface, but
not to Token Group.

#### Summary of Changes
Apply hashed error code discriminators to SPL Token Group error codes.

### DIFF
--- a/token-group/interface/src/error.rs
+++ b/token-group/interface/src/error.rs
@@ -3,7 +3,7 @@
 use spl_program_error::*;
 
 /// Errors that may be returned by the interface.
-#[spl_program_error]
+#[spl_program_error(hash_error_code_start = 3_406_457_176)]
 pub enum TokenGroupError {
     /// Size is greater than proposed max size
     #[error("Size is greater than proposed max size")]


joncinque (approved):
Thanks! Where did the number come from?

buffalojoec:
> Thanks! Where did the number come from?

I had to look up how we did it, but basically `program-error` will hash the enum name, and if you provide a number that isn't correct it yells at you.

This may seem weird, but originally we decided that this hashing and discriminator assignment would happen behind the scenes, but we wanted a way for the source code to reflect that actual discriminator, so we added the intentional compile error.

https://github.com/solana-labs/solana-program-library/blob/29f96a2ae8f263125f905a6a064698f9e01e5323/libraries/program-error/derive/src/macro_impl.rs#L149-L177

joncinque:
Ah right, thanks for the reminder!

PR #6703, opened 2024-05-07
token-cli: update member `GROUP_TOKEN_ADDRESS` arg

buffalojoec:
#### Problem
The two arguments for `initialize-member` on the Token CLI are `token` and `group_token`.
However, when you print the help you'll see `<TOKEN_MINT_ADDRESS> <TOKEN_MINT_ADDRESS>`.
These should be a bit more clear.

#### Summary of Changes
Change the group token address help print-out to `<GROUP_TOKEN_ADDRESS>`.


joncinque (approved):
Thanks!

PR #6704, opened 2024-05-07
token-2022: Fix pre fee calculation edge cases

### LINKED ISSUE #6451: [transfer-fee] calculate_inverse_fee  logic error when transfer_fee_basis_points is MAX_FEE_BASIS_POINTS
when transfer_fee_basis_points is MAX_FEE_BASIS_POINTS,  the function should return the maximum fee. Otherwise, the following test will fail.
```
    #[test]
    fn calculate_fee_exact_out_max11() {
        let transfer_fee = TransferFee {
            epoch: PodU64::from(0),
            maximum_fee: PodU64::from(5_000),
            transfer_fee_basis_points: PodU16::from(MAX_FEE_BASIS_POINTS),
        };

        assert_eq!(
            5_000,
            transfer_fee.calculate_inverse_fee(u64::MAX / 2).unwrap()
        );

        assert_eq!(5_000, transfer_fee.calculate_inverse_fee(1).unwrap());
    }
```

joncinque:
#### Problem

As mentioned in #6451, the pre fee amount calculation doesn't properly handle certain edge cases. If the fee is 100%, we return zero, which is flat-out wrong.

#### Solution

Return `maximum_fee + post_fee_amount` if fee is 100%. I changed this to use a match instead to make it easier to read, I hope it looks ok!

Fixes #6451

### DIFF
--- a/token/program-2022/src/extension/transfer_fee/mod.rs
+++ b/token/program-2022/src/extension/transfer_fee/mod.rs
@@ -90,20 +90,24 @@ impl TransferFee {
     pub fn calculate_pre_fee_amount(&self, post_fee_amount: u64) -> Option<u64> {
         let maximum_fee = u64::from(self.maximum_fee);
         let transfer_fee_basis_points = u16::from(self.transfer_fee_basis_points) as u128;
-        if transfer_fee_basis_points == 0 {
-            Some(post_fee_amount)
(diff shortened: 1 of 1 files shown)


samkim-crypto (approved):
Great! I think this covers all the edges that I can think of at least. The match is also easier to read as well!

PR #6709, opened 2024-05-08
Refactor: Remove duplicated build command in token-swap script

1997roylee:


buffalojoec (approved):
I can't believe NPM didn't barf on this. Thanks for spotting!

PR #6725, opened 2024-05-14
ci: upgrade cache action to version with node20

cario-dev:
Node16 is deprecated and Actions using it will stop being used 3rd of June
https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

Remove warnings like this https://github.com/solana-labs/solana-program-library/actions/runs/9073098915


joncinque (approved):
Thanks for the contribution!

PR #6764, opened 2024-05-27
Account compression: Init with root allowing trees with a canopy

StanChe:
# Initializing a tree with a root
This PR modifies the `account-compression` program allowing the creation of the trees that were prepared offline. It allows creating trees with or without a canopy.
Verification of the validity, events, any additional checks and fields (URLs, hashes, etc.), and any additional permissions for the created tree are left to the discretion of the calling program.

## The workflow

```mermaid
sequenceDiagram
autonumber
actor TC as Tree creator program
participant AC as account-compression
activate TC
TC ->> AC: prepare_batch_merkle_tree(tree_account, max_depth, max_buffer_size)
AC -> AC: initializes the header, but not the tree
loop For a tree with a canopy
TC ->> AC: append_canopy_nodes(tree_account, start_index, canopy_nodes)
end
TC ->> AC: init_prepared_tree_with_root(tree_account, root, rightmost_leaf, rightmost_index, proof)
AC -> AC: verifies the canopy is valid
AC -> AC: initializes the tree
```

The separation of a single `init_with_root` is based on the fact that some trees to be created might require knowing the pubkey of the tree, like with the bubblegum, where asset ids are derived from the tree pubkey and the index in that tree. Thus the user should first create the tree account and assign all the required permissions and then may take time to prepare the tree itself.

If the tree has a canopy - all canopy leaves should be initialized, otherwise the finalization will fail.

This change is inspired by #6441


ngundotra:
Will look at this in mid June when back from OOO. Is there context behind why this needs to get merged into SPL Account Compression, rather than being done in a separate fork of the program?

StanChe:
Integrating this feature into the standard library will benefit existing projects already utilizing account compression by enabling them to extend their contracts with rollups and batch creation of trees. This will reduce the number of transactions needed for tree creation, thereby alleviating network congestion. The batch creation method ensures that the tree is in the same state as it would be with individual appends, maintaining consistency. The prepared but not finalized tree will be unusable by other existing methods until it's finalized. The responsibility of validating the tree remains with the calling program, while this change guarantees that the tree initialized with a root has a valid canopy. This modification provides a tool for the batch creation of compressed accounts without imposing specific requirements on the caller and leaving the indexing up to their logic.

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -173,69 +177,149 @@ pub mod spl_account_compression {
...
+        header.assert_valid_authority(&ctx.accounts.authority.key())?;
+        // assert the tree is not initialized yet, we don't want to overwrite the canopy of an
+        // initialized tree
+        let merkle_tree_size = merkle_tree_get_size(&header)?;
+        let (tree_bytes, canopy_bytes) = rest.split_at_mut(merkle_tree_size);
+        // ensure the tree is not initialized, the hacky way
+        require!(
+            tree_bytes.iter().all(|&x| x == 0),
+            AccountCompressionError::TreeAlreadyInitialized
+        );
```
danenbm, line 252: Should something like this also be in `prepare_tree()`?  In the `Initialize` account validation struct used for `prepare_tree()`, it indicates the account will be zeroed out but I don't actually see where that is done.  If it is a gap, its probably an existing gap with `init_empty_merkle_tree()` as well.

account-compression/sdk/tests/accountCompression.test.ts
danenbm, line 1: Great tests!

danenbm (commented):
I suspect we should fix the stack overflow thing.  Looks like this program was last deployed about 4 months ago, maybe it wasn't an issue then, but I worry it could cause runtime issues now, even if the tests pass not sure if failures would be predictable.

Recently we fixed a similar stack overflow issue that started causing test failures in mpl-core-candy-machine by simply heap-allocating a large stack variable: https://github.com/metaplex-foundation/mpl-core-candy-machine/pull/9/commits/9575c8b8a83114c6256583a74c2ea686d62d3072

account-compression/programs/account-compression/src/lib.rs
danenbm: Ok we see the zero anchor validation, not an issue

StanChe:
@danenbm thanks for the feedback. Decided to mode the fix of the stack overflow into a separate PR in order to simplify this one and to mix everything up.

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -173,69 +177,149 @@ pub mod spl_account_compression {
...
+        let mut merkle_tree_bytes = ctx.accounts.merkle_tree.try_borrow_mut_data()?;
+
+        let (header_bytes, rest) =
+            merkle_tree_bytes.split_at_mut(CONCURRENT_MERKLE_TREE_HEADER_SIZE_V1);
+        // the header should already be initialized with prepare_tree
+        let header = ConcurrentMerkleTreeHeader::try_from_slice(header_bytes)?;
+        header.assert_valid_authority(&ctx.accounts.authority.key())?;
+        let merkle_tree_size = merkle_tree_get_size(&header)?;
+        let (tree_bytes, canopy_bytes) = rest.split_at_mut(merkle_tree_size);
```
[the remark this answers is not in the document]
StanChe: I've tried to support the existing code style as close to the original as possible. In a general case, most of the existing methods may be refactored to use that same helper method. In my opinion that may be done as a separate effort that doesn't modify the functionality and only refactores the code.

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -177,69 +182,146 @@ pub mod spl_account_compression {
...
+    /// nodes. The canopy is used to decrease the size of the proof required to update the tree.
+    /// If the tree is expected to have a canopy, it needs to be prefilled with the necessary nodes.
+    /// There are 2 ways to initialize a merkle tree:
+    /// 1. Initialize an empty tree
+    /// 2. Initialize a tree with a root and leaf
+    /// For the former case, the canopy will be empty which is expected for an empty tree. The
+    /// expected flow is `init_empty_merkle_tree`. For the latter case, the canopy should be
+    /// filled with the necessary nodes to render the tree usable. Thus we need to prefill the
+    /// canopy with the necessary nodes. The expected flow for a tree without canopy is
+    /// `prepare_tree` -> `init_merkle_tree_with_root`. The expected flow for a tree with canopy
```
ngundotra, line 196: Fix comments. This should be `finalize...` ix

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -177,69 +182,146 @@ pub mod spl_account_compression {
...
+        header.serialize(&mut header_bytes)?;
+        let merkle_tree_size = merkle_tree_get_size(&header)?;
+        let (_tree_bytes, canopy_bytes) = rest.split_at_mut(merkle_tree_size);
+        check_canopy_bytes(canopy_bytes)
+    }
+
+    /// This instruction pre-initializes the canopy with the specified leaf nodes of the canopy.
+    /// This is intended to be used after `prepare_tree` and in conjunction with the
+    /// `init_merkle_tree_with_root` instruction that'll finalize the tree initialization.
+    pub fn append_canopy_nodes(
```
ngundotra, line 232: What is supposed to happen if there is an error appending nodes to the tree?

How is the manager of the tree expected to delete the tree & reclaim lamports if this step is done in error?

ngundotra (changes requested):
This is really great PR overall. Great job extending existing design patterns (as bad as they are 😅). 

High level requests:
- There should be an easy way to discriminate normal trees from batch-init'd trees. When we talked previously, I suggested adding a new header version & a flag in that header. 

- What is the recourse for errors in batch-init'd trees? Which steps of the process are reversible or explicitly immutable? For the immutable steps, can we have some explicit documentation & also add helper methods to the SDK prevent immutable errors?

account-compression/programs/account-compression/src/lib.rs
StanChe: Updated the method doc to reflect the already filled nodes may be replaced with new nodes before calling `finalize_merkle_tree_with_root`. Also updated the `close_empty_tree` method to allow removal of the prepared tree.

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -193,9 +193,9 @@ pub mod spl_account_compression {
...
     /// filled with the necessary nodes to render the tree usable. Thus we need to prefill the
     /// canopy with the necessary nodes. The expected flow for a tree without canopy is
-    /// `prepare_tree` -> `init_merkle_tree_with_root`. The expected flow for a tree with canopy
+    /// `prepare_tree` -> `finalize_merkle_tree_with_root`. The expected flow for a tree with canopy
     /// is `prepare_tree` -> `append_canopy_nodes` (multiple times until all of the canopy is
-    /// filled) -> `init_merkle_tree_with_root`. This instruction initializes the tree header
+    /// filled) -> `finalize_merkle_tree_with_root`. This instruction initializes the tree header
     /// while leaving the tree itself uninitialized. This allows distinguishing between an empty
     /// tree and a tree prepare to be initialized with a root.
     pub fn prepare_tree(
```
ngundotra, line 201: Can this method be renamed to `init_batch_merkle_tree` to match the other init method?

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -246,6 +257,7 @@ pub mod spl_account_compression {
 
         let header = ConcurrentMerkleTreeHeader::try_from_slice(header_bytes)?;
         header.assert_valid_authority(&ctx.accounts.authority.key())?;
+        header.assert_is_batch_initialized()?;
```
ngundotra, line 260: Can you add a unit test that checks that `append_canopy_nodes` and `finalize_merkle_tree_with_root` cannot be called on batch initialized trees?

account-compression/programs/account-compression/src/lib.rs
ngundotra: Also, what happens if this method is called on an already-existing tree?

account-compression/programs/account-compression/src/lib.rs
```diff
@@ -544,8 +564,11 @@ pub mod spl_account_compression {
         let merkle_tree_size = merkle_tree_get_size(&header)?;
         let (tree_bytes, canopy_bytes) = rest.split_at_mut(merkle_tree_size);
 
-        let id = ctx.accounts.merkle_tree.key();
-        merkle_tree_apply_fn_mut!(header, id, tree_bytes, prove_tree_is_empty,)?;
+        // Check if the tree is either empty or is batch initialized and not finalized yet.
+        if !header.get_is_batch_initialized() || !tree_bytes.iter().all(|&x| x == 0) {
```
ngundotra, line 568: How does `!tree_bytes.iter().all(|&x| x == 0)` check that a batch initialized tree is not finalized yet?

ngundotra (changes requested):
What happens if I init a batch tree with `prepare_tree` and then attempt to append to it?

I think that another byte of padding is needed to store `is_initialized` for batch init'd trees. Then the tree can be checked quickly before executing replace or append. I'm open to more efficient ways of doing this - just wanted to offer my suggestion.

Again, great job with this. Almost there

account-compression/programs/account-compression/src/lib.rs
StanChe: > Can this method be renamed to `init_batch_merkle_tree` to match the other init method?


If we take a look at the account structure it may be represented as `| header | tree body | canopy |`. The existing `init_empty_merkle_tree` does the full initialization - first the `header` is initialized, then the `tree body`, and finally the `canopy` size is checked. This makes the tree usable by the append methods. 
For a batch initialized tree the actual initialization takes several steps. First the `prepare_tree` initializes just the `header` and checks the canopy size leaving the rest of the account zero'd. Then the `canopy` is set using `append_canopy_nodes`. And finally the `finalize_merkle_tree_with_root` initializes the `tree body`. 
Until the `finalize_merkle_tree_with_root` is called the tree remains unusable by any other method - the `init_empty_merkle_tree` will fail because the header is initialized and any modification will fail because the tree itself is not initialized (the `self.is_initialized()` check will fail). Only the newly added `append_canopy_nodes` and `finalize_merkle_tree_with_root` as well as the modified `close_empty_tree` can operate over a tree in such a state.
That was also the reason the method naming is different to `init_` as it leaves the tree in a not fully initialized state, but with only the initialized header (thus prepared).
I'm considering to rename the `finalize_merkle_tree_with_root` into `init_prepared_tree_with_root` and `prepare_tree` into `prepare_batch_merkle_tree`. This will keep the naming concise in terms the tree is fully initialized after the `init_xxx` methods. Will also update the comments accordingly. Wdyt?


> Also, what happens if this method is called on an already-existing tree?

Adding a test to showcase this.

account-compression/programs/account-compression/src/lib.rs
StanChe: I believe the `it('Should fail to double finalize the tree', async () => {` and `it('Should fail to replace a canopy node for a finalised tree', async () => {` are already covering the case for a finalized tree. For a regular (not batch-initialized) tree added some more tests

StanChe:
Also noted the following stack overflow error occurring on build, fix it as well
```
Error: Function _ZN23spl_account_compression23spl_account_compression22init_empty_merkle_tree17hbc30209ef0f8a270E Stack offset of 4928 exceeded max offset of 4096 by 832 bytes, please minimize large stack variables
```

ngundotra (changes requested):
Looks great. Almost ready to push to devnet for testing.

Two more comments:
1) Rebase on top of master
2) Address some small docs + typos

Once this is done, I'll add a PR that updates versions for & publishes the crates & js sdk.

account-compression/programs/account-compression/src/canopy.rs
```diff
@@ -128,3 +131,431 @@ pub fn fill_in_proof_from_canopy(
...
+#[inline(always)]
+fn get_value_for_node<const N: usize>(
+    node_idx: usize,
+    level: u32,
+    canopy: &[Node],
+    empty_node_cache: &mut [Node; N],
+) -> Node {
+    if canopy[node_idx - 2] != EMPTY {
+        return canopy[node_idx - 2];
+    }
```
ngundotra, line 266: If it wasn't for your extensive tests, I definitely would have thought this would be a footgun. 

Converting node indexing to canopy indexing almost always throws me off

StanChe:
FYI, had to bump up node to 20.5 for 2 js workflows that were failing on node 16. Looks like eslint became incompatible somehow. Used 20.5 as the latest used in some other flows, 18 should work as well.

PR #6805, opened 2024-06-05
Add Create & Update Token Commands to Stake-Pool CLI

sense112:
Adding the following commands to the `spl-stake-pool` rust cli:

- `create_token_metadata`
- `update_token_metadata`

All arguments have been catered for with friendly descriptions.

### DIFF
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -499,7 +499,7 @@ fn create_token_metadata(
     uri:String) -> CommandResult {
     let stake_pool = get_stake_pool(&config.rpc_client, stake_pool_address)?;
 
-    let mut signers = vec![config.fee_payer.as_ref(), config.staker.as_ref()];
+    let mut signers = vec![config.fee_payer.as_ref(), config.manager.as_ref()];
     let instructions = vec![
         spl_stake_pool::instruction::create_token_metadata(
             &spl_stake_pool::id(),
--- a/stake-pool/cli/src/main.rs
+++ b/stake-pool/cli/src/main.rs
@@ -526,7 +526,7 @@ fn update_token_metadata(
     uri:String) -> CommandResult {
     let stake_pool = get_stake_pool(&config.rpc_client, stake_pool_address)?;
 
-    let mut signers = vec![config.fee_payer.as_ref(), config.staker.as_ref()];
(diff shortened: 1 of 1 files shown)


stake-pool/cli/src/main.rs
```diff
@@ -491,6 +491,59 @@ fn command_create_pool(
...
 
+fn create_token_metadata(
+    config: &Config,
+    stake_pool_address: &Pubkey,
+    name: String, 
+    symbol:String, 
+    uri:String) -> CommandResult {
+    let stake_pool = get_stake_pool(&config.rpc_client, stake_pool_address)?;
+
+    let mut signers = vec![config.fee_payer.as_ref(), config.staker.as_ref()];
```
joncinque, line 502: The manager should be the one to sign this
```suggestion
    let mut signers = vec![config.fee_payer.as_ref(), config.manager.as_ref()];
```

stake-pool/cli/src/main.rs
```diff
@@ -491,6 +491,59 @@ fn command_create_pool(
...
+
+fn update_token_metadata(
+    config: &Config,
+    stake_pool_address: &Pubkey,
+    name: String, 
+    symbol:String, 
+    uri:String) -> CommandResult {
+    let stake_pool = get_stake_pool(&config.rpc_client, stake_pool_address)?;
+
+    let mut signers = vec![config.fee_payer.as_ref(), config.staker.as_ref()];
```
joncinque, line 529: Same here, this should be the manager
```suggestion
    let mut signers = vec![config.fee_payer.as_ref(), config.manager.as_ref()];
```

joncinque (commented):
Thanks so much for your contribution! There's just two small things to fixup.

In order to pass CI, the code has to be formatted per rustfmt, so could you run `./cargo-nightly fmt --all` from the root of the repo and commit the change?

And finally, I don't think this will be an issue, but to be safe, can you also make sure this passes clippy? You can run that from the root of the repo with:

```
$ ./cargo-nightly clippy -Zunstable-options --all-targets --features test-sbf -- --deny=warnings --deny=clippy::arithmetic_side_effects
```

Once that's all done, this should be good to merge!

sense112:
Took me some time to get my environment setup correctly to run these commands.

Thanks for the comments! I have run the format command and clippy seems happy :)

image

joncinque (approved):
Great job, thanks for your contribution!

PR #6812, opened 2024-06-06
transfer-hook-example: Only allow one mint to initialize

joncinque:
#### Problem

The transfer hook example program allows any mint to initialize extra account metas. If there are other accounts managed by the transfer hook program, it might be possible for an attacker to create their own extra account metas account using accounts from another mint.

For example, mint A uses the extra accounts metas account B, which declares usage of an account C, owned by the transfer hook program. An attacker creates a mint D with extra account metas E, and also declares usage of account C. Now two mints might be modifying the same account.

#### Solution

While this attack is hypothetical and not possible given the simple transfer hook program here, let's model good default behavior and limit the transfer hook example to only allow account metas for one mint.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7914,6 +7914,7 @@ dependencies = [
  "spl-tlv-account-resolution 0.6.3",
  "spl-token-2022 3.0.2",
  "spl-token-client",
+ "spl-transfer-hook-example",
  "spl-transfer-hook-interface 0.6.3",
  "strum 0.26.2",
  "strum_macros 0.26.4",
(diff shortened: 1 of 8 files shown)


buffalojoec:
> For example, mint A uses the extra accounts metas account B, which declares usage of an account C, owned by the transfer hook program. An attacker creates a mint D with extra account metas E, and also declares usage of account C. Now two mints might be modifying the same account.

What exactly is the attack here? What if both mints need to use some central PDA - say a configuration PDA - under this program?

joncinque:
> What exactly is the attack here?

Essentially, if a program allows multiple mints *and also* has data accounts, it's easy for a second mint to abuse the data accounts from the first mint, if the addresses are not PDAs off of the mint account. It's a bit of a longshot, but for safety, we should be getting people to default their transfer-hook programs to only work with one mint.

> What if both mints need to use some central PDA - say a configuration PDA - under this program?

If two mints require a central PDA, then the program author wants the program to support multiple mints, which means they'll need to be careful in what kind of accounts are allowed. If they're designing something like that, they should know to carefully remove the single-mint check

buffalojoec (approved):

buffalojoec:
Great, thanks for laying it all out!

PR #6827, opened 2024-06-10
account-compression: fix for a stack overflow issue on macro usage

StanChe:
# Account compression stack overflow fix
## The problem
Building the current version of account-compressions on the master branch gives the following error. 

```
Error: Function _ZN23spl_account_compression23spl_account_compression12replace_leaf17hdc3dd750827dcd7bE Stack offset of 4976 exceeded max offset of 4096 by 880 bytes, please minimize large stack variables
Error: Function _ZN23spl_account_compression23spl_account_compression11verify_leaf17h04e381cb5ac1c2baE Stack offset of 4912 exceeded max offset of 4096 by 816 bytes, please minimize large stack variables
Error: Function _ZN23spl_account_compression23spl_account_compression16insert_or_append17hf55fb617fa2a0f55E Stack offset of 4952 exceeded max offset of 4096 by 856 bytes, please minimize large stack variables
```

This indicates a potential stack overflow issue. The underlying issue is with the usage of the macros with huge match states. The macro is essentially inlined by the compiler into the calling code and rust seems to not reuse the same stack frame for different cases reserving a separate frame for every case. 

## The solution
First off we introduce additional wrapper methods that will call the macros. This will ensure the macro is inlined only in this wrapper method, and not in the calling code. Next we add structures for arguments for "heavy" methods that contain proofs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6907,7 +6907,7 @@ dependencies = [
 
 [[package]]
 name = "spl-concurrent-merkle-tree"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "bytemuck",
  "rand 0.8.5",
--- a/account-compression/programs/account-compression/Cargo.toml
+++ b/account-compression/programs/account-compression/Cargo.toml
@@ -21,8 +21,8 @@ default = []
 anchor-lang = "0.29.0"
 bytemuck = "1.13"
 solana-program = ">=1.18.11,<=2"
-spl-concurrent-merkle-tree = { version = "0.2.0", path = "../../../libraries/concurrent-merkle-tree", features = [
-  "sol-log",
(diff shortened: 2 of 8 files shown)


danenbm (commented):
LGTM!

ngundotra:
@StanChe thanks for fixing this! This is really smart. Looks good to me on technical front.

@joncinque This PR changes the methods available on the `ConcurrentMerkleTree` struct in `spl-concurrent-merkle-tree` crate to avoid stack issues. Should the `spl-concurrent-merkle-tree` version be bumped to `0.3.0` or `0.2.1` ?

joncinque:
Looks like this changes the function signatures, so you'll definitely need to roll 0.3.0 for the change

ngundotra (dismissed):

PR #6847, opened 2024-06-14
TLV Account Resolution: Add support for keys in data

### LINKED ISSUE #6845: transfer-hook: Unable to use seed to generate an ExtraAccountMeta that isn't a PDA
**Problem**

Currently, to generate a non-PDA account (a regular account) for a transfer hook, you can only use the `ExtraAccountMeta::new_with_pubkey` method. This method requires a static public key as input, which means the public key must be predetermined and unchanging.

However, there are scenarios where we might want to generate a non-PDA account dynamically. This means creating an account whose public key is determined at runtime based on other account data in the instruction.

For example, 
`authority` that can be derived from the `source_fee_account` data cannot be created using `ExtraAccountMeta::new_with_pubkey`

```
pub struct TransferHookCtx<'info> {
    #[account(
        token::mint = mint,
        token::authority = owner,
    )]
    pub source_token: InterfaceAccount<'info, TokenAccount>,
    /// CHECK:
    pub mint: AccountInfo<'info>,
    #[account(
        token::mint = mint,
    )]
    pub destination_token: InterfaceAccount<'info, TokenAccount>,
    /// CHECK: source token account owner, can be SystemAccount or PDA owned by another program
    pub owner: UncheckedAccount<'info>,
    /// CHECK: ExtraAccountMetaList Account,
    #[account(
        seeds = [b"extra-account-metas", mint.key().as_ref()], 
        bump
    )]
    pub extra_account_meta_list: UncheckedAccount<'info>,
    pub associated_token_program: Program<'info, AssociatedToken>,
    pub token_program: Interface<'info, TokenInterface>,
    pub system_program:  Program<'info, System>,
    #[account(mut)]
    pub source_fee_account: Account<'info, FeeAccount>,
    #[account(
        mut,
        seeds = [b"pda_authority", mint.key().as_ref()], 
        bump
    )]
    /// CHECK: 
    pub pda_authority: SystemAccount<'info>,
    /// CHECK: 
    pub authority: AccountInfo<'info>,
    #[account(
        init_if_needed,
	payer = payer,
	associated_token::authority = authority,
	associated_token::mint = mint,
    )]
    pub token_account: InterfaceAccount<'info, TokenAccount>
}
```

buffalojoec:
#### Problem
Currently, TLV Account Resolution (and Transfer Hook) doesn't support extracting
account keys from data - including instruction or account data.

#### Summary of Changes
Introduce the `PubkeyData` configuration for doing exactly that, then add support for
it to `ExtraAccountMeta` and `ExtraAccountMetaList`.

cc @jychab

Closes #6845

### DIFF
--- a/libraries/tlv-account-resolution/src/account.rs
+++ b/libraries/tlv-account-resolution/src/account.rs
@@ -4,11 +4,13 @@
 //! collection of seeds
 
 use {
-    crate::{error::AccountResolutionError, seeds::Seed},
+    crate::{error::AccountResolutionError, key_data::KeyData, seeds::Seed},
     bytemuck::{Pod, Zeroable},
     solana_program::{
-        account_info::AccountInfo, instruction::AccountMeta, program_error::ProgramError,
-        pubkey::Pubkey,
+        account_info::AccountInfo,
+        instruction::AccountMeta,
+        program_error::ProgramError,
+        pubkey::{Pubkey, PUBKEY_BYTES},
     },
     spl_pod::primitives::PodBool,
 };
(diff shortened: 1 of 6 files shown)


libraries/tlv-account-resolution/src/key_data.rs
```diff
@@ -0,0 +1,183 @@
...
+
+#[cfg(feature = "serde-traits")]
+use serde::{Deserialize, Serialize};
+use {crate::error::AccountResolutionError, solana_program::program_error::ProgramError};
+
+/// Enum to describe a required key stored in some data.
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
+pub enum KeyData {
```
joncinque, line 15: nit: Since we use `new_with_pubkey` elsewhere for the raw pubkey data, how about renaming all instances of `KeyData` to `PubkeyData`? I find the term "key" to be a little less clear

libraries/tlv-account-resolution/src/key_data.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /// Packs a key data configuration into a 32-byte array, filling the rest
+    /// with 0s.
+    pub fn pack_into_address_config(key_data: &Self) -> Result<[u8; 32], ProgramError> {
+        let mut packed = [0u8; 32];
+        let tlv_size = key_data.tlv_size() as usize;
+        key_data.pack(&mut packed[..tlv_size])?;
+        Ok(packed)
+    }
```
joncinque, line 89: It's nice when the more complicated variant is already done, and this can be simple :sweat_smile:

libraries/tlv-account-resolution/src/account.rs
```diff
@@ -66,6 +68,48 @@ where
...
+            Ok(Pubkey::new_from_array(
+                account_data[arg_start..arg_end].try_into().unwrap(),
+            ))
+        }
+    }
+}
+
 /// `Pod` type for defining a required account in a validation account.
 ///
 /// This can either be a standard `AccountMeta` or a PDA.
```
joncinque, line 115: I can't comment further down, but can you specify that `discriminator` can also refer to dynamic pubkey data on line 120-122?

joncinque (commented):
Great job! And thanks for the fast turnaround -- just a couple of nits, the logic and tests are excellent

libraries/tlv-account-resolution/src/account.rs
buffalojoec: Ahh yep I missed this documentation.

libraries/tlv-account-resolution/src/key_data.rs
buffalojoec: Sure, I like that better!

libraries/tlv-account-resolution/src/key_data.rs
buffalojoec: For real! We didn't even need `unpack_address_config` for this one.

buffalojoec: The most beautiful part was not having to change anything in `ExtraAccountMetaList`. 😎

joncinque (approved):
key

PR #6861, opened 2024-06-17
token-2022: [L-03] Don't require memo for confidential self-transfer

joncinque:
#### Problem

> Description: In regular instructions (Transfer, TransferChecked, and TransferCheckedWithFee)  a self-transfer does not require a memo even if the destination is extended with MemoTransfer while in ConfidentialTransferInstruction::Transfer, self-transfer always does. 

> Recommendation: Make regular and confidential instructions behave uniformly with respect to MemoTransfer.

#### Solution

Don't require a memo for confidential self-transfers


CriesofCarrots:
I tend to think even a self-transfer should require a memo, if the MemoTransfer extension is set (ie. this was my mistake). Thoughts?

joncinque:
Yeah, that does make more sense. I'll close this and go the other route -- thanks for chiming in!

joncinque:
Closing in favor of #6865

PR #6862, opened 2024-06-17
token-2022: [L-02] Check for non-transferable extension consistently

joncinque:
#### Problem

According to the Certora audit findings:

> Description: Regular instructions (Transfer, TransferChecked, and TransferCheckedWithFee) determine that an account is non-transferable by checking for presence of NonTransferableAccount extension in the source account. ConfidentialTransferInstruction::Transfer, instead, checks that the mint has the NonTransferable mint extension.

> Recommendation: This is not an issue at the moment because every account of NonTransferable mint has a NonTransferableAccount extension. We recommend that all instructions use NonTransferableAccount to make the code more uniform and to prevent any deviation in the future should the invariant no longer be maintained.

#### Solution

We have to use the account check for the extension in non-confidential transfers, since the mint might not always be present, so switch to checking the extension on the account in confidential transfers too.

### DIFF
--- a/token/program-2022/src/extension/confidential_transfer/processor.rs
+++ b/token/program-2022/src/extension/confidential_transfer/processor.rs
@@ -3,7 +3,7 @@
 use crate::extension::transfer_hook;
 #[cfg(feature = "zk-ops")]
 use {
-    crate::extension::non_transferable::NonTransferable,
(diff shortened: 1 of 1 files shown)


samkim-crypto (approved):
Looks good to me! It was my bad from the beginning 🙏

PR #6865, opened 2024-06-17
token-2022: [L-03] Require memo on self-transfer with memo-transfer

joncinque:
#### Problem

> Description: In regular instructions (Transfer, TransferChecked, and TransferCheckedWithFee)  a self-transfer does not require a memo even if the destination is extended with MemoTransfer while in ConfidentialTransferInstruction::Transfer, self-transfer always does. 

> Recommendation: Make regular and confidential instructions behave uniformly with respect to MemoTransfer.

#### Solution

The opposite of #6861: require a memo for non-confidential self-transfers

### DIFF
--- a/token/program-2022/src/processor.rs
+++ b/token/program-2022/src/processor.rs
@@ -451,6 +451,9 @@ impl Processor {
         // This check MUST occur just before the amounts are manipulated
         // to ensure self-transfers are fully validated
         if self_transfer {
+            if memo_required(&source_account) {
+                check_previous_sibling_instruction_is_memo()?;
+            }
             return Ok(());
         }
 
--- a/token/program-2022-test/tests/memo_transfer.rs
+++ b/token/program-2022-test/tests/memo_transfer.rs
@@ -75,6 +75,23 @@ async fn test_memo_transfers(
     let bob_state = token.get_account_info(&bob_account).await.unwrap();
     assert_eq!(bob_state.base.amount, 0);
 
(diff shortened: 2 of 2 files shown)


token/program-2022/src/processor.rs
```diff
@@ -451,6 +451,9 @@ impl Processor {
         // This check MUST occur just before the amounts are manipulated
         // to ensure self-transfers are fully validated
         if self_transfer {
+            if memo_required(&source_account) {
+                check_previous_sibling_instruction_is_memo()?;
+            }
```
CriesofCarrots, line 456: Does it maybe make sense to just move the existing check from L469 before this block so we don't have to duplicate?
https://github.com/solana-labs/solana-program-library/blob/d103601f6f83d26e30189959763da1c894ec1d44/token/program-2022/src/processor.rs#L453-L455

joncinque: We can't simply move the check up unfortunately. We borrowing the destination account data as mutable to manipulate its data and check for the memo requirement, but if it's a self-transfer, we do a double-mutable borrow between the source and destination account data and panic.

I couldn't think of a better way to get the check done unfortunately than to duplicate the check on the source account. Do you have any suggestions?

CriesofCarrots: Ugh, that's annoying. I despise the brittleness of having 2 memo checks in different spots.
But I haven't come up with anything better so far. 😞

CriesofCarrots (approved):

PR #6874, opened 2024-06-18
token-2022: [I-03] Clarify `calculate_inverse_fee` math

joncinque:
#### Problem

> Description: The function calculate_inverse_fee is not exactly an inverse operation of calculate_fee. That is, it is not the case that calculate_inverse_fee(x + calculate_fee(x)) == calculate_fee(x).
> Recommendation: Document that calculate_inverse_fee is not an exact inverse and instead that only the relationship calculate_fee(x) <= calculate_inverse_fee(x + calculate_fee(x)) holds in order to avoid confusion with the potential users of calculate_inverse_fee.

#### Solution

Do the recommended thing. I don't think the math in the comment is correct, so I've added a test to make sure the relationship is correct.

### DIFF
--- a/token/program-2022/src/extension/transfer_fee/mod.rs
+++ b/token/program-2022/src/extension/transfer_fee/mod.rs
@@ -113,6 +113,15 @@ impl TransferFee {
     }
 
     /// Calculate the fee that would produce the given output
+    ///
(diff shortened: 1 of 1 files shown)


samkim-crypto (approved):
Looks good to me!

Issue #6879, opened 2024-06-19
Feature Proposal: Confidential Mint & Burn Extension


abcalphabet:
**Proposed feature**

Token-2022 adds the token-extension of confidential transfers. In order to increase or decrease the supply of a token in the confidential balance space the tokens have to be first minted normally and then deposited into the confidential balance for the former and withdrawn from the confidential balance and then burned in the unencrypted balance space for the latter. 

If the purpose of a token is to only operate in the confidential balance space, it should be possible to skip the deposit and withdraw steps and directly mint into or burn out of the confidential balance.

**Why is this needed**

One possible application confidential token balances is the representation of RWAs via a token. At [Iron](https://iron.xyz) we’re using Token Extensions to issue tokenized bank deposits which comes with a wide range of regulatory requirements. In this more regulated environment there are two relevant considerations to minting and burning confidentially:

1. Only the issuer should be able to discern (and share with authorities when necessary) the balance and transaction amounts of individual holders
2. The supply of the token should always represent the amount of the RWA held, i.e. mint on inflows and burn on outflows

The currently available functionality of going via the unencrypted space to change the supply goes against point 1 as on on each inflow tokens have to be minted without encryption and then deposited into the confidential balance of the token account. An alternative to this would be to premint a large amount of tokens into a "treasury" account and then send those out confidentially as inflows happen, this would however obviously be in violation of point 2. The issue for burns is similar with either burning unencrypted to trigger / represent an outflow or sending the tokens back to a pooled account instead of that.

**Proposed solution**

A `confidential-mint-burn` extension adding a `ConfindentialMint` and a `ConfidentialBurn` instruction. This extension would add a separate mint-authority used for minting confidential tokens to the token's mint account.

The `ConfindentialMint` instruction can reuse most of the `Deposit` logic to add tokens to the confidential balance, but requires an additional proof that the supplied `ElGamalCiphertext`s represent a 48-bit number, so as to properly fit into `pending_balance_hi` and `pending_balance_lo`.

The `ConfidentialBurn` can also reuse some existing code, but that of the `Transfer` instead of the `Withdraw` instruction. To achieve a confidential burn, and guarantee correctness of the confidential balance before / after the burn, the source-side logic of the transfer can be used. This provides a simple way of generating the correct proofs and applying them in a proven way. The destination side of the logic and proofs would obviously not be used, as a burn can be represented as a transfer to /dev/null.

**Other considerations**

[...]

PR #6881, opened 2024-06-19
Confidential mint-burn extension

abcalphabet:
WIP implementation for the feature proposed in https://github.com/solana-labs/solana-program-library/issues/6879

**How to run**

Start up a local validator with the bpf compiled from this branch and the zk-token-sdk enabled on it.

Build the cli on the branch and run the following

```
solana-keygen new -o key.json;

export SPL_TOKEN_CLI=$PWD"/target/debug/spl-token"
export SIGNER_KEYPAIR='key.json'
export SIGNER_PUBKEY="`solana-keygen pubkey $SIGNER_KEYPAIR`"

export TOKEN_ADDR=$($SPL_TOKEN_CLI create-token --program-id TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb \
    --enable-confidential-transfers auto \
    --enable-confidential-mint-burn \
    --decimals 2 --output json \
    --mint-authority $SIGNER_KEYPAIR \
    | jq -r .commandOutput.address)

$SPL_TOKEN_CLI create-account $TOKEN_ADDR \
    --fee-payer $SIGNER_KEYPAIR --owner $SIGNER_KEYPAIR
export ATA_ADDR=$($SPL_TOKEN_CLI accounts $TOKEN_ADDR \
    --owner $SIGNER_KEYPAIR --output json | jq -r .accounts[0].address)
$SPL_TOKEN_CLI configure-confidential-transfer-account --address $ATA_ADDR \
    --fee-payer $SIGNER_KEYPAIR \
    --owner $SIGNER_KEYPAIR

$SPL_TOKEN_CLI mint-confidential-tokens $TOKEN_ADDR 2 \
    --address $ATA_ADDR --fee-payer $SIGNER_KEYPAIR \
    --owner $SIGNER_KEYPAIR
$SPL_TOKEN_CLI confidential-balance $ATA_ADDR --authority $SIGNER_KEYPAIR;
$SPL_TOKEN_CLI apply-pending-balance --address $ATA_ADDR \
    --fee-payer $SIGNER_KEYPAIR --owner $SIGNER_KEYPAIR
$SPL_TOKEN_CLI burn-confidential-tokens $TOKEN_ADDR 1 \
    --address $ATA_ADDR --fee-payer $SIGNER_KEYPAIR \
    --owner $SIGNER_KEYPAIR
```

**Implementation notes**

~~This PR uses the proofs used for the transfers, obviously only proofs for two Ciphertexts (`mint_to / burn_from` and `auditor`) are necessary, so there's some inefficiency there as the transfer proofs create a third one on top of that. (Ab)Using these transfer proofs does however mean that the logic can work on it's own without additional changes to the `solana-zk-token-sdk`.~~ After some discussion on the corresponding github issue, this has been resolved

### DIFF
--- a/token/program-2022/src/extension/confidential_mint_burn/processor.rs
+++ b/token/program-2022/src/extension/confidential_mint_burn/processor.rs
@@ -150,30 +150,24 @@ fn process_confidential_mint(
     accounts: &[AccountInfo],
     data: &MintInstructionData,
 ) -> ProgramResult {
-    println!("MINT 1");
     let account_info_iter = &mut accounts.iter();
     let token_account_info = next_account_info(account_info_iter)?;
     let mint_info = next_account_info(account_info_iter)?;
-    println!("MINT 2");
 
     check_program_account(mint_info.owner)?;
     let mint_data = &mut mint_info.data.borrow_mut();
     let mut mint = PodStateWithExtensionsMut::<PodMint>::unpack(mint_data)?;
     let mint_authority = mint.base.mint_authority;
-    println!("MINT 3");
 
     let auditor_elgamal_pubkey = mint
         .get_extension::<ConfidentialTransferMint>()?
         .auditor_elgamal_pubkey;
-    println!("MINT 4");
     if let Ok(extension) = mint.get_extension::<PausableConfig>() {
         if extension.paused.into() {
             return Err(TokenError::MintPaused.into());
         }
     }
-    println!("MINT 5");
     let mint_burn_extension = mint.get_extension_mut::<ConfidentialMintBurn>()?;
-    println!("MINT 6");
 
     let proof_context = verify_mint_proof(
         account_info_iter,
@@ -181,17 +175,14 @@ fn process_confidential_mint(
         data.ciphertext_validity_proof_instruction_offset,
         data.range_proof_instruction_offset,
     )?;
-    println!("MINT 7");
 
     check_program_account(token_account_info.owner)?;
     let token_account_data = &mut token_account_info.data.borrow_mut();
     let mut token_account = PodStateWithExtensionsMut::<PodAccount>::unpack(token_account_data)?;
-    println!("MINT 8");
 
     let authority_info = next_account_info(account_info_iter)?;
     let authority_info_data_len = authority_info.data_len();
     let authority = mint_authority.ok_or(TokenError::NoAuthorityExists)?;
-    println!("MINT 9");
 
     Processor::validate_owner(
         program_id,
@@ -200,7 +191,6 @@ fn process_confidential_mint(
         authority_info_data_len,
         account_info_iter.as_slice(),
     )?;
-    println!("MINT 10");
 
     if token_account.base.is_frozen() {
         return Err(TokenError::AccountFrozen.into());
@@ -216,7 +206,6 @@ fn process_confidential_mint(
         token_account.get_extension_mut::<ConfidentialTransferAccount>()?;
     confidential_transfer_account.valid_as_destination()?;
 
-    println!("MINT 12");
     if proof_context.mint_pubkeys.destination != confidential_transfer_account.elgamal_pubkey {
         return Err(ProgramError::InvalidInstructionData);
     }
@@ -226,7 +215,6 @@ fn process_confidential_mint(
             return Err(ProgramError::InvalidInstructionData);
         }
     }
-    println!("MINT 13");
 
     let proof_context_auditor_ciphertext_lo = proof_context
         .mint_amount_ciphertext_lo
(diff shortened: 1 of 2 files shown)


samkim-crypto:
@abcalphabet, there has been some work on refactoring the zk logic in the `confidential-transfer` submodule. I added the necessary zkp for confidential mint and burn in https://github.com/solana-labs/solana-program-library/pull/6955. Can you rebase and use the logic from the `confidential-transfer-proof-extraction` module?

samkim-crypto:
Also, is it okay if you split the PR into
1. token22 program and program-test
2. token-client and cli
components? The program part requires careful attention and it would be easier to focus solely on the program first.

abcalphabet:
Hey @samkim-crypto, will rebase your changes and use that logic. Splitting it into two PRs also makes sense, will get on that as well.

abcalphabet:
@samkim-crypto two questions:

## Splitting the PRs

The tests in `program-2022-test` generally use the token client to call the logic to be tested, so splitting the tests and the client changes would require some major work making only the mint-burn tests work without that. Would a PR with just the program changes and one with the client+cli+tests be okay for you?

## Auditability of instructions

I think I forgot to mention this on the issue discussing the feature:

Currently the auditor ciphertexts are only included in the context state accounts which makes using them in practice somewhat complicated since in order to continuously monitor transactions one needs to watch transactions on the token, get the ciphertext pubkey from the instruction, fetch all transactions on the ciphertext pubkey, find the proof instruction and extract the auditor ciphertext from there. This PR includes the auditor ciphertexts in the instruction itself so that token transactions (just mints and burns atm really) can be continuously audited by simply streaming transactions involving the specific token mint.

Does this make sense to you?

If it does make sense I guess putting that logic in a seperate issue / PR adding this across the board (including in the normal `confidential-transfer` extension) is probably better

abcalphabet:
@samkim-crypto Another question regarding the proof generation logic you added:

When generating the mint proofs there is an assumption that the mint would have a `PodAeCiphertext` decryptable supply and associated `AeKey`. How can we guarantee that this decryptable supply is actually always correct in practice? Since only we who have the key can mint it's updated fine on mints, but as any token owner can burn there's no way to update the decryptable supply on burns. 

On the issue we had discussed just using a `PodElGamalCiphertext` for the supply, but that was before the equality proof was added to the mint proofs. We could still use that and decrypt it to get the supply as u64 for the proof, but that obviously has the potential to make the proof data generation quite slow once the supply gets into higher 64-bit territory.

Do you have any ideas on how to properly do this? Or am I just missing something obvious

samkim-crypto:
Yes that is a good point. I forgot to point this out, but the mint should have a decryptable balance `PodAeCiphertext` of the total supply. Unlike an ElGamal ciphertext, this ciphertext should be easily decryptable even if the encrypted number is large.

As you suggested though, the decryptable balance cannot always encrypt the correct balance in practice. The important point here is that the an ElGamal ciphertext is hard to decrypt when the encrypted value is large, but it is homomorphic, so it can still be decrypted relatively easily with the help of the decryptable ciphertext. Let's denote the current ElGamal ciphertext encrypting the supply as `ct_current`.
1. Decrypt the decryptable ciphertext to find the old (outdated) supply `x_old`
2. Encrypt (encode) `x_old` as an ElGamal ciphertext (e.g. like how it is done for `Deposit`). Let's denote this `ct_old`.
3. Compute `ct_old - ct_current` using ciphertext arithmetic. Let's denote this ciphertext to be `ct_diff`.
4. Decrypting `ct_diff` will give the burn amount that is not taken into account in the decryptable ciphertext. Let's denote this value `x_diff`.
5. Then we know that `x_current = x_old - x_diff` is the most up to date supply. At this point, the decryptable balance can be updated to encrypt `x_current`.

The point here is that in step 4, `ct_diff` will not contain the full supply, but just the burn amounts that happened since the last time decryptable balance was updated. So as long as the decryptable balance is updated relatively frequent, the ElGamal ciphertext that encrypts the supply will not be too expensive to decrypt. Does this make sense?

abcalphabet:
Yup, makes perfect sense. Thanks for the explanation!

abcalphabet:
@samkim-crypto changed everything to use the new proof logic.

Seperate PR with just the token changes here: https://github.com/solana-labs/solana-program-library/pull/7319

samkim-crypto:
@kilogold This extension is already added to the token program (https://github.com/solana-labs/solana-program-library/pull/7319). We have not yet added this to the token-client yet (and hence why this PR is not yet closed), but that should follow very soon.

@abcalphabet Thanks for the contribution to the token program! I can go ahead and add the relevant components to the token-client (following the commits here) and to the `program-2022-tests` later this week. If you want to take a shot at the client too, then feel free to let me know!

PR #6883, opened 2024-06-19
ensure construct_tx uses the latest blockhash before returning

yihau:
### Problem

I tried to bump curve25519-dalek on agave and found this one: https://github.com/anza-xyz/agave/actions/runs/9579725243/job/26412781469?pr=1693

Screenshot 2024-06-19 at 23 01 38

it died on here

https://github.com/solana-labs/solana-program-library/blob/d9a6ee8db65167098b654b300ac23abc08fd8a7d/token/program-2022-test/tests/confidential_transfer.rs#L2959-L2980

if I didn't miss anything, here will be the only place to send txs
https://github.com/solana-labs/solana-program-library/blob/d9a6ee8db65167098b654b300ac23abc08fd8a7d/token/client/src/token.rs#L2994-L3024

all of them will finally called
 
https://github.com/solana-labs/solana-program-library/blob/d9a6ee8db65167098b654b300ac23abc08fd8a7d/token/client/src/token.rs#L571

I suspect that when construct_tx returns those txs, blockhashs have expired. (I couldn't reproduce it locally...)

https://github.com/solana-labs/solana-program-library/blob/d9a6ee8db65167098b654b300ac23abc08fd8a7d/token/client/src/token.rs#L598-L614

### Summary of Changes

try to fetch the latest blockhash again before signing

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -595,23 +595,24 @@ where
 
         instructions.extend_from_slice(token_instructions);
 
-        let blockhash = if let (Some(nonce_account), Some(nonce_authority), Some(nonce_blockhash)) = (
-            self.nonce_account,
-            &self.nonce_authority,
-            self.nonce_blockhash,
-        ) {
-            let nonce_instruction = system_instruction::advance_nonce_account(
-                &nonce_account,
-                &nonce_authority.pubkey(),
-            );
-            instructions.insert(0, nonce_instruction);
-            nonce_blockhash
-        } else {
-            self.client
-                .get_latest_blockhash()
-                .await
-                .map_err(TokenError::Client)?
-        };
+        let mut blockhash =
+            if let (Some(nonce_account), Some(nonce_authority), Some(nonce_blockhash)) = (
+                self.nonce_account,
+                &self.nonce_authority,
+                self.nonce_blockhash,
+            ) {
+                let nonce_instruction = system_instruction::advance_nonce_account(
+                    &nonce_account,
+                    &nonce_authority.pubkey(),
+                );
+                instructions.insert(0, nonce_instruction);
+                nonce_blockhash
+            } else {
+                self.client
+                    .get_latest_blockhash()
+                    .await
+                    .map_err(TokenError::Client)?
+            };
 
         if let Some(compute_unit_price) = self.compute_unit_price {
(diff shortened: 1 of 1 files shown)


token/client/src/token.rs
```diff
@@ -595,23 +595,24 @@ where
...
+                    &nonce_authority.pubkey(),
+                );
+                instructions.insert(0, nonce_instruction);
+                nonce_blockhash
+            } else {
+                self.client
+                    .get_latest_blockhash()
+                    .await
+                    .map_err(TokenError::Client)?
+            };
```
joncinque, line 615: The token-client is also used by the CLI, where we shouldn't be over-fetching blockhashes. Rather than fetch it twice, if you really think this is the issue, we should just move the whole blockhash fetch to later, and change this to 
```suggestion
            if let (Some(nonce_account), Some(nonce_authority), Some(nonce_blockhash)) = (
                self.nonce_account,
                &self.nonce_authority,
                self.nonce_blockhash,
            ) {
                let nonce_instruction = system_instruction::advance_nonce_account(
                    &nonce_account,
                    &nonce_authority.pubkey(),
                );
                instructions.insert(0, nonce_instruction);
            }
```

And then later on, get the hash all at once with my other suggestion. What do you think?

token/client/src/token.rs
```diff
@@ -636,6 +637,14 @@ where
...
         }
 
+        // ensure the blockhash is still the latest one
+        if self.nonce_blockhash.is_none() {
+            blockhash = self
+                .client
+                .get_latest_blockhash()
+                .await
+                .map_err(TokenError::Client)?
+        }
```
joncinque, line 647: To go along with my other suggestion, we can change this to:

```suggestion
        let blockhash = if let Some(blockhash) = self.nonce_blockhash {
            blockhash
        } else {
            self.client
                .get_latest_blockhash()
                .await
                .map_err(TokenError::Client)?
        };
```

joncinque (commented):
A few comments to make this better, but I'm not sure if this is really the issue. Has it appeared on any other CI runs?

yihau:
ah. okay! after some tracing I think the curve25519 version bumping increased the executing time 😢 I will try to ignore these failing tests on the upstream for this known issue 😞

PR #6886, opened 2024-06-19
[libraries/pod]: add `PodOption` type

febo:
### Problem

Optional values usually require an extra `byte` to indicate whether the value is present or not. This can affect the alignment of a type, which is not ideal.

### Solution

This PR adds a `PodOption`, which is a generic type that can be used as a `Pod` to wrap a type as a `Option`. It works over types that implement a `Nullable` trait in order to indicate whether its value represent a `None` or not.

For example, a `PodOption<u8>` assumes the a `0` value is a `None` and any value different than `0` is `Some`.

```rust
/// Trait for types that can be `None`.
///
/// This trait is used to indicate that a type can be `None` according to a
/// specific value.
pub trait Nullable: Pod {
    /// Indicates whether the value is `None` or not.
    fn is_none(&self) -> bool;
}

impl Nullable for u8 {
    fn is_none(&self) -> bool {
        *self == 0
    }
}
```

The PR includes implementation for all integer types and `Pubkey`.


libraries/pod/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 pub mod bytemuck;
 pub mod error;
 pub mod optional_keys;
+pub mod pod_option;
```
joncinque, line 6: nit: can this file be called `option`? otherwise `spl_pod::pod_option::PodOption` has the word `Pod` a few too many times :sweat_smile:

libraries/pod/src/pod_option.rs
```diff
@@ -0,0 +1,218 @@
+//! Generic `Option` that can be used as a `Pod`s for types that can have
+//! a `None` value.
```
joncinque, line 2: nit: just to make this abundantly clear
```suggestion
//! Generic `Option` that can be used as a `Pod` for types that can have
//! a designated `None` value.
//!
//! For example, a 64-bit unsigned integer can designate `0` as a `None` value.
//! This would be equivalent to
//! [`Option<NonZeroU64>`](https://doc.rust-lang.org/std/num/type.NonZeroU64.html)
//! and provide the same memory layout optimization.
```

libraries/pod/src/pod_option.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    #[inline]
+    pub fn to_option(&self) -> Option<&T> {
+        if self.0.is_none() {
+            None
+        } else {
+            Some(&self.0)
+        }
+    }
```
joncinque, line 44: I always forget what's possible with `From` and refs, but can this be done through some form of `impl From<PodOption<T>>` instead?

libraries/pod/src/pod_option.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    #[inline]
+    pub fn to_option_mut(&mut self) -> Option<&mut T> {
+        if self.0.is_none() {
+            None
+        } else {
+            Some(&mut self.0)
+        }
+    }
```
joncinque, line 53: Same here, can we use some form of `From` instead? Maybe we can get quite clever like https://doc.rust-lang.org/std/option/enum.Option.html#impl-From%3C%26mut+Option%3CT%3E%3E-for-Option%3C%26mut+T%3E

libraries/pod/src/pod_option.rs
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+    };
+}
+
+impl_int_nullable!(u8);
+impl_int_nullable!(u16);
+impl_int_nullable!(u32);
+impl_int_nullable!(u64);
+impl_int_nullable!(u128);
```
joncinque, line 93: I've only just found all of these implementations in bytemuck already -- maybe we can use them instead? https://docs.rs/bytemuck/latest/bytemuck/trait.PodInOption.html

joncinque (commented):
Looks really good! Just a few small things on my side

libraries/pod/src/pod_option.rs
```diff
@@ -0,0 +1,218 @@
...
+
+/// Implementation of `Nullable` for `Pubkey`.
+///
+/// The implementation assumes that the default value of `Pubkey` represents
+/// `None`.
+impl Nullable for Pubkey {
+    fn is_none(&self) -> bool {
+        self == &Pubkey::default()
+    }
+}
```
joncinque, line 103: This is really neat and shows how easy this is to use

buffalojoec (commented):
The API itself is clean, but I'm generally concerned with defaulting to zero values as `None` for integers.

I wonder if instead we should architect this type around "nonzero" integers, so the distinction is clear?

https://docs.rs/bytemuck/latest/bytemuck/trait.ZeroableInOption.html

libraries/pod/src/pod_option.rs
febo: Oh, this is interesting. We could use them. The only comment that I have is that we will have two "types" of optionals:
* `PodOption<T>`: when T is not one of the `NonZero*` variants
* `Option<T>`: when T is one of the `NonZero*` variants

We could still `impl` the `Nullable` trait on the `NonZero*` so you could use `PodOption` with them as well. It does not prevent someone from using `Option<NonZero*>` but gives the alternative to have all `PodOption` instead. 

In summary, we don't `impl` the `Nullable` trait on primitives, only on the `NonZero*` variants. If you want to have a `PodOption` with a primitive value, you need to use the `NonZero*` variant, which makes it explicit that zero is not a valid value.

joncinque: To be honest, I typically prefer less code, so I would recommend using `Option<NonZero*>` over potentially confusing people with two ways of doing things.

febo: Good point!

libraries/pod/src/pod_option.rs
```diff
@@ -0,0 +1,218 @@
...
+
+unsafe impl<T: Nullable> Pod for PodOption<T> {}
+
+unsafe impl<T: Nullable> Zeroable for PodOption<T> {}
+
+impl<T: Nullable> PodOption<T> {
+    #[inline]
+    pub fn new(value: T) -> Self {
+        Self(value)
+    }
```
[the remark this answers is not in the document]
febo: I removed the `fn new(...)` in favour of implementing `impl<T: Nullable> From<T> for PodOption<T>`. This way you could rely directly on the value of `T` – e.g.:
* `PodOption::from(sysvar::ID) <- Some(...)`
* `PodOption::from(Pubkey::default()) <- None`

libraries/pod/src/pod_option.rs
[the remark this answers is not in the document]
febo: Good point, removed in favour of using `From`.

libraries/pod/src/pod_option.rs
```diff
@@ -0,0 +1,218 @@
...
+        if self.0.is_none() {
+            None
+        } else {
+            Some(&self.0)
+        }
+    }
+
+    #[inline]
+    pub fn to_option_mut(&mut self) -> Option<&mut T> {
+        if self.0.is_none() {
```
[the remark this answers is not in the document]
febo: Yeah, it feels that we need to improve a bit the way you can check/get the value of `T`. This is a bit verbose:
```rust
assert_eq!(Into::<Option<Pubkey>>::into(none_pubkey), None);
```
I could not find a way around it without having to create a variable first (which is not bad, but also not great):
```rust
let none_pubkey_option: Option<Pubkey> = none_pubkey.into();
```

It would be nice to have a simple way to do a `match` over a `PodOption`.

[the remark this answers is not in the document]
febo: @buffalojoec After our chat, I had a closer look at the `NonZero` struct and followed a bit the pattern there. It has a `get` method to get the contained value, so I added a `fn get(self) -> Option<T>` to return the value as an `Option`.

https://doc.rust-lang.org/std/num/struct.NonZero.html

It uses `from(T) -> NonZero<T>` methods and also a `new(T) -> NonZero<T>`, similar to the one that we had before but I think the `from(...)` should be sufficient.

libraries/pod/src/pod_option.rs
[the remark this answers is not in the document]
buffalojoec: Cool, this sounds good to me!

libraries/pod/src/option.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl<'a, T: Nullable> From<&'a PodOption<T>> for Option<&'a T> {
+    fn from(from: &'a PodOption<T>) -> Self {
+        if from.0.is_none() {
+            None
+        } else {
+            Some(&from.0)
+        }
+    }
+}
```
buffalojoec, line 67: In my opinion I'd get rid of this trait implementation and instead add a const (if possible) `as_ref` method to the type directly.

https://doc.rust-lang.org/src/core/option.rs.html#681-686

With the setup we have here, `const` might not be possible.

https://github.com/anza-xyz/agave/blob/8db8e60c48ab064c88a76013597f99c9eb25ed74/sdk/program/src/program_option.rs#L125-L131

libraries/pod/src/option.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl<'a, T: Nullable> From<&'a mut PodOption<T>> for Option<&'a mut T> {
+    fn from(from: &'a mut PodOption<T>) -> Self {
+        if from.0.is_none() {
+            None
+        } else {
+            Some(&mut from.0)
+        }
+    }
+}
```
buffalojoec, line 77: Same here, I'd instead opt for a `get_mut(&mut self) -> Option<&mut T>` on the type directly.

https://doc.rust-lang.org/src/core/option.rs.html#703-708

libraries/pod/src/option.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl<'a, T: Nullable> From<&'a PodOption<T>> for COption<&'a T> {
+    fn from(from: &'a PodOption<T>) -> Self {
+        if from.0.is_none() {
+            COption::None
+        } else {
+            COption::Some(&from.0)
+        }
+    }
+}
```
buffalojoec, line 97: No need for this, since the user can just call `.into().as_ref()`!

https://github.com/anza-xyz/agave/blob/8db8e60c48ab064c88a76013597f99c9eb25ed74/sdk/program/src/program_option.rs#L126-L131

libraries/pod/src/option.rs
```diff
@@ -0,0 +1,140 @@
...
+
+impl<'a, T: Nullable> From<&'a mut PodOption<T>> for COption<&'a mut T> {
+    fn from(from: &'a mut PodOption<T>) -> Self {
+        if from.0.is_none() {
+            COption::None
+        } else {
+            COption::Some(&mut from.0)
+        }
+    }
+}
```
buffalojoec, line 107: No need for this, since the user can just call `.into().as_mut()`!

https://github.com/anza-xyz/agave/blob/8db8e60c48ab064c88a76013597f99c9eb25ed74/sdk/program/src/program_option.rs#L146-L151

libraries/pod/src/option.rs
```diff
@@ -0,0 +1,140 @@
...
+///
+/// This can be used when a specific value of `T` indicates that its
+/// value is `None`.
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub struct PodOption<T: Nullable>(T);
+
+impl<T: Nullable> PodOption<T> {
+    /// Returns the contained value as an `Option`.
+    pub fn get(self) -> Option<T> {
```
buffalojoec, line 34: This is totally up to you, but might I suggest `as_option()` over `get()`?

buffalojoec (commented):
This is on the right track, @febo! I've just added some suggestions on the trait implementations.

I still think you can add to this type's API to make its use a little bit more ergonomic. I still believe it should at least offer `is_some()` and `is_none()` directly on the type.

Up to you, but I think `COption` does a great job of providing an API similar to Rust's `Option`, which makes it feel more comfortable to use.

https://github.com/anza-xyz/agave/blob/master/sdk/program/src/program_option.rs

libraries/pod/src/option.rs
buffalojoec: Going off my other comment, we could also name it `as_option_mut`. What do you think?

libraries/pod/src/pod_option.rs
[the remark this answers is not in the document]
buffalojoec: Sounds good to me. I like the internal getter, I just tossed in a suggestion on the naming. I know you pulled `get` from `NonZero`, but let me know what you think.

libraries/pod/src/option.rs
febo: I added `as_ref` and `as_mut`, which return read-only / mutable references.

libraries/pod/src/option.rs
febo: Kept this one as `get()` to match the `NonZero` API.

febo:
> This is on the right track, @febo! I've just added some suggestions on the trait implementations.
> 
> I still think you can add to this type's API to make its use a little bit more ergonomic. I still believe it should at least offer `is_some()` and `is_none()` directly on the type.
> 
> Up to you, but I think `COption` does a great job of providing an API similar to Rust's `Option`, which makes it feel more comfortable to use.
> 
> https://github.com/anza-xyz/agave/blob/master/sdk/program/src/program_option.rs

@buffalojoec I cleaned up the API based on your review, let me know what do you think.

buffalojoec (approved):
Looks great! Ship it!

libraries/pod/src/option.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn from(value: T) -> Self {
+        PodOption(value)
+    }
+}
+
+impl<T: Nullable> From<Option<T>> for PodOption<T> {
+    fn from(from: Option<T>) -> Self {
+        match from {
+            Some(value) => PodOption(value),
+            None => PodOption(T::default()),
```
joncinque, line 83: This assumes that `T::default()` is the `None` value, which is not guaranteed by the trait. I understand the desire to make the value flexible, so to make that behavior less error-prone, how about changing the trait to something like:

```
pub trait Nullable: Pod + PartialEq + Sized {
    /// Value indicating the `None` to check against
    const NONE: Self; 
    /// Indicates whether the value is `None` or not.
    fn is_none(&self) -> bool {
        self == &Self::NONE
    }
```

joncinque (commented):
Just a few flyby comments that I hope will remove some footguns

libraries/pod/src/option.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+impl<T: Nullable> From<COption<T>> for PodOption<T> {
+    fn from(from: COption<T>) -> Self {
+        match from {
+            COption::Some(value) => PodOption(value),
+            COption::None => PodOption(T::default()),
+        }
+    }
+}
```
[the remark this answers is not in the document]
febo: Good call!

libraries/pod/src/option.rs
febo: The only caveat is that in this case `NONE` value has to be a `const`. I don't think is a big deal, but perhaps we can consider adding a method to the trait to specify the `NONE` value.

* Using `const NONE`:
  ```rust
  impl Nullable for Pubkey {
      const NONE: Self = Pubkey::new_from_array([0u8; PUBKEY_BYTES]);
  }
  ```
* Using `none()` method:
  ```rust
  impl Nullable for Pubkey {
      fn none() -> Self {
          Pubkey::default()
      }
  }
  ```

PR #6889, opened 2024-06-20
audit: Add curve25519-dalek RUSTSEC to ignore list

joncinque:
#### Problem

Just like https://github.com/anza-xyz/agave/pull/1786, SPL is failing the audit check in CI.

#### Solution

Until we can upgrade to curve25519-dalek v4, ignore it


joncinque:
Merging early since it passed the audit step

Issue #6897, opened 2024-06-24
`2.0.0` `solana-program` release causing issues due to `<=2.0.0` on patch upgrades.


jacobcreech:
### Problem

`solana-program` upgraded to `2.0.0` on Friday, leading to a number of builds across the ecosystem because of the `<=2.0.0` in a patch version upgrade on https://github.com/solana-labs/solana-program-library/pull/6182. 

The problematic dependencies are as follows:

`spl-pod` version `0.1.1` -> Previous version `0.1.0` did not have `<=2.0.0`. [Upgrade on patch broke some builds](https://x.com/samvwaal/status/1804742697934262441)
`spl-token-metadata-interface` version `0.2.1` -> Previous version `0.2.0` did not have `<=2.0.0`. [Upgrade on patch broke some builds](https://x.com/cryptopapi997/status/1804418386602135584/photo/1)
`spl-associated-token-account` version `2.3.1` -> Previous version `2.3.0` did not have `<=2.0.0`.
`spl-token` version `4.0.1` -> Previous version `4.0.0` did not have `<=2.0.0`.
`spl-token-group-interface` version `0.1.1` -> Previous version `0.1.0` did not have `<=2.0.0`.

These are the ones I can find at this time. This affects anyone using `anchor-spl` right now, as well as some projects that don't use Anchor but pull in `spl-token` `4.0.1` while using `solana-program` 1.X.

Relevant Anchor issue: https://github.com/coral-xyz/anchor/issues/3044

### Suggested Solution
Purge the versions causing problems, re-release them without the `<=2.0.0`, upgrade and release with `<=2.0.0` as a "major" version if not already available. As mentioned [here](https://github.com/coral-xyz/anchor/issues/3044#issuecomment-2184964762)

joncinque:
I ran into a lot of these issues while publishing the last spl-token-cli, so there are compatible major versions that already exist in spl:

* spl-associated-token-account -- v3 exists, so we can yank 2.3.1
* spl-token -- might need v5, then we can yank 4.0.1
* spl-token-group-interface -- v0.2 exists, so we can yank 0.1.1
* spl-token-metdata-interface -- v0.3 exists, so we can yank 0.2.1
* spl-pod -- v0.2 exists, so we can yank 0.1.1
* spl-token-2022 -- is ok

There are some more:

* spl-program-error -- yank 0.3.1
* spl-discriminator -- yank 0.1.1
* spl-type-length-value -- yank v0.3.1
[...]

PR #6908, opened 2024-06-24
deps: Upgrade to Solana v2

joncinque:
#### Problem

As part of fixing up #6897, we need to upgrade SPL to use all the v2 crates.

#### Solution

Taking this commit by commit:

* First we just update everything to v2
* Next we update the lockfile. There's some manual work here to force the SPL dependencies from the *monorepo* to use v2 to avoid issues
* Update the main toolchain version
* Update the nightly toolchain version
* Fix issues that arise from `cargo check`
* Fix issues that arise from `cargo clippy`
* Undo the upgrade on account-compression, since it uses anchor
* Fixup the token-2022 serde build, which requires a protobuf compiler
* Set the version properly everywhere
* Fix the twoxtx build
* Actually, don't set the solana-install version since 2.0.0 doesn't exist yet
* Nevermind, it does work with a new url!

This should be pretty straightforward, but let me know if there's any issues!


samkim-crypto (approved):
Thanks for the thorough PR description!

PR #6915, opened 2024-06-25
[confidential-transfer] Add transfer split proof generation

samkim-crypto:
#### Summary of Changes
Added the confidential transfer proof generation logic in a separate crate in SPL. Since the ciphertext arithmetic will need to be implemented in SPL as well, I think it makes to eventually refactor the `encryption` module out into a separate crate, but I kept it in the proofs crate for now.

The main logic mainly follows the split proof generation in the token-2022 crate. This PR includes the transfer proof logic for now. The transfer with fee and withdraw proof generation logic will be included in follow-ups.


Cargo.toml
```diff
@@ -60,6 +60,7 @@ members = [
   "token/transfer-hook/cli",
   "token/transfer-hook/example",
   "token/transfer-hook/interface",
+  "token/confidential-transfer/proofs",
```
joncinque, line 63: are there going to be other confidential transfer crates? if so, this is fine, but if not, can we just put this in `token/confidential-transfer-proofs`?

joncinque (commented):
Seeing all of the target_os specific stuff, it makes me think: how would it look to split this all into two crates, one for off-chain, and one for on-chain? the off-chain crate can depend on the on-chain one. Is that what you meant by refactoring the encryption bits?

If it's too annoying, don't worry about it, but since we're creating the new crates now, it could be a good time to split this up.

Otherwise, everything looks great!

token/confidential-transfer/proofs/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "spl-confidential-transfer-proofs"
```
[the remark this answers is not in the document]
samkim-crypto: okay, will make the change!

Cargo.toml
samkim-crypto: Yeah the plan was to also have the encryption crate for the ciphertext ops. The `encryption` module in the proofs crate will then refactor into that crate.

token/confidential-transfer/proof-generation/src/transfer.rs
```diff
@@ -0,0 +1,205 @@
...
+        zk_elgamal_proof_program::proof_data::{
+            BatchedGroupedCiphertext3HandlesValidityProofData, BatchedRangeProofU128Data,
+            CiphertextCommitmentEqualityProofData,
+        },
+    },
+};
+
+/// The padding bit length in range proofs that are used for a confidential
+/// token transfer
+#[cfg(not(target_os = "solana"))]
```
joncinque, line 22: Can you remove the `target_os = "solana"` cfgs in this file?

token/confidential-transfer/proof-extraction/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+repository = "https://github.com/solana-labs/solana-program-library"
+license = "Apache-2.0"
+edition = "2021"
+
+[dependencies]
+solana-zk-sdk = "2.0.0"
+thiserror = "1.0.61"
+
+[dev-dependencies]
+spl-token-confidential-transfer-proof-generation = { version = "0.1.0", path = "../proof-generation" }
```
joncinque, line 15: This is up to you can be done later, but you could put all the tests in a separate crate to avoid having the extraction crate (the on-chain one) depend on the off-chain crate here.

joncinque (commented):
Just one last thing, a suggestion, then this is good to go!

token/confidential-transfer/proof-extraction/Cargo.toml
samkim-crypto: Good idea! I refactored the tests into a separate crate.

joncinque (approved):
Looks great, thanks for integrating all the feedback!

PR #6945, opened 2024-06-28
[confidential-transfer] Add transfer with fee proof generation and extraction

samkim-crypto:
#### Problem
The logic to generate proof data for transfer with fee have not been added to the repo yet.

#### Summary
Add the logic to generate the proof data. This proof is quite technical, but it essentially follows the existing logic in [token client](https://github.com/solana-labs/solana-program-library/blob/master/token/client/src/proof_generation.rs).

### DIFF
--- a/token/confidential-transfer/proof-extraction/src/transfer_with_fee.rs
+++ b/token/confidential-transfer/proof-extraction/src/transfer_with_fee.rs
@@ -1,0 +1,324 @@
+use {
+    crate::{
+        encryption::{PodFeeCiphertext, PodTransferAmountCiphertext},
+        errors::TokenProofExtractionError,
+    },
+    bytemuck::bytes_of,
+    solana_curve25519::{
+        ristretto::{self, PodRistrettoPoint},
+        scalar::PodScalar,
+    },
+    solana_zk_sdk::{
+        encryption::pod::{
+            elgamal::{PodElGamalCiphertext, PodElGamalPubkey},
+            pedersen::PodPedersenCommitment,
+        },
+        zk_elgamal_proof_program::proof_data::{
+            BatchedGroupedCiphertext2HandlesValidityProofContext,
+            BatchedGroupedCiphertext3HandlesValidityProofContext, BatchedRangeProofContext,
+            CiphertextCommitmentEqualityProofContext, PercentageWithCapProofContext,
+        },
+    },
+};
+
+const MAX_FEE_BASIS_POINTS: u64 = 10_000;
+
+/// The transfer public keys associated with a transfer with fee.
+pub struct TransferWithFeePubkeys {
+    /// Source ElGamal public key
+    pub source: PodElGamalPubkey,
+    /// Destination ElGamal public key
+    pub destination: PodElGamalPubkey,
+    /// Auditor ElGamal public key
+    pub auditor: PodElGamalPubkey,
+    /// Withdraw withheld authority public key
+    pub withdraw_withheld_authority: PodElGamalPubkey,
+}
+
+/// The proof context information needed to process a [Transfer] instruction
+/// with fee.
+pub struct TransferWithFeeProofContext {
+    /// Group encryption of the low 16 bits of the transfer amount
+    pub ciphertext_lo: PodTransferAmountCiphertext,
+    /// Group encryption of the high 48 bits of the transfer amount
+    pub ciphertext_hi: PodTransferAmountCiphertext,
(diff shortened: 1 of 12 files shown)


token/confidential-transfer/proof-generation/src/transfer_with_fee.rs
```diff
@@ -0,0 +1,340 @@
...
+        range_proof_data,
+    ))
+}
+
+fn calculate_fee(transfer_amount: u64, fee_rate_basis_points: u16) -> Option<(u64, u64)> {
+    let numerator = (transfer_amount as u128).checked_mul(fee_rate_basis_points as u128)?;
+
+    // Warning: Division may involve CPU opcodes that have variable execution times.
+    // This non-constant-time execution of the fee calculation can theoretically
+    // reveal information about the transfer amount. For transfers that invole
```
joncinque, line 312: micro-nit
```suggestion
    // reveal information about the transfer amount. For transfers that involve
```

token/confidential-transfer/proof-extraction/src/transfer_with_fee.rs
```diff
@@ -0,0 +1,324 @@
...
+        &commitment_to_ristretto(transfer_amount_commitment_lo),
+        &commitment_to_ristretto(transfer_amount_commitment_hi),
+    )
+    .ok_or(TokenProofExtractionError::CurveArithmetic)?;
+    let transfer_fee_basis_points_scalar = u16_to_scalar(transfer_fee_basis_points);
+    let scaled_transfer_amount_point =
+        ristretto::multiply_ristretto(&transfer_fee_basis_points_scalar, &transfer_amount_point)
+            .ok_or(TokenProofExtractionError::CurveArithmetic)?;
+
+    const MAX_FEE_BASIS_POINTS: u64 = 10_000;
```
joncinque, line 309: nit: You can remove this and use the one defined at the top of the file instead

joncinque (approved):
Looks good! Just tiny things

token/confidential-transfer/proof-tests/tests/proof_test.rs
```diff
@@ -53,3 +57,77 @@ fn test_proof_validity(spendable_balance: u64, transfer_amount: u64) {
...
+fn test_transfer_with_fee_correctness() {
+    test_transfer_with_fee_proof_validity(0, 0, 0, 0);
+    test_transfer_with_fee_proof_validity(0, 0, 0, 1);
+    test_transfer_with_fee_proof_validity(0, 0, 1, 0);
+    test_transfer_with_fee_proof_validity(0, 0, 1, 1);
+    test_transfer_with_fee_proof_validity(1, 0, 0, 0);
+    test_transfer_with_fee_proof_validity(1, 1, 0, 0);
+
+    test_transfer_with_fee_proof_validity(100, 100, 5, 10);
+    test_transfer_with_fee_proof_validity(100, 100, 5, 1);
```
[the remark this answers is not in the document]
samkim-crypto: No, there is no specific reason. I had a bunch of tests and it seemed like the test was becoming quite slow and the transfer amount was already somewhat tested for transfer without fee tests, so I thought I would just include the smaller ones. Now, I did add the tests for big numbers and used `(5, 10)` and `(5, 1)` for the `(basis point, max fee)` parameters. I can also add more edge case fee parameters too if you see fit.

token/confidential-transfer/proof-generation/src/transfer_with_fee.rs
```diff
@@ -0,0 +1,340 @@
...
+    fee_rate_basis_points: u16,
+    maximum_fee: u64,
+) -> Result<
+    (
+        CiphertextCommitmentEqualityProofData,
+        BatchedGroupedCiphertext3HandlesValidityProofData,
+        PercentageWithCapProofData,
+        BatchedGroupedCiphertext2HandlesValidityProofData,
+        BatchedRangeProofU256Data,
+    ),
```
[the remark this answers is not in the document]
samkim-crypto: Yeah, I think this is a good point and I think we can certainly organize it using a structs. Since this issue is shared with transfer without fee, I'll address it in a separate PR.

token/confidential-transfer/proof-extraction/src/transfer_with_fee.rs
samkim-crypto: Yep, thanks for the catch. Fixed!

joncinque (approved):

PR #6951, opened 2024-06-30
[token-upgrade] Remove clap deprecated functions

samkim-crypto:
#### Problem
The token-upgrade cli tool still uses some deprecated functions from clap-v2.

#### Summary
Updated the deprecated functions from clap-v2. Most of the changes should be straightforward:
- Replaced `Arg::multiple` with `ArgAction::Append`, which should be functionally identical
- Replaced `is_present` with `try_contains_id`. The old `is_present` panics if the arg is not previously known/defined while `try_contains_id` gives an error. Otherwise, the two functions are identical.
- Replaced `try_pubkey_of` with `SignerSource::try_get_pubkey`. The `try_pubkey_of` internally tries to parse `Pubkey` from `SignerSource`, which unfortunately, solana-clap-v3-utils does not yet support. The `SignerSource::try_get_pubkey` parses `SignerSource` correctly and then extracts `Pubkey` from it.

The only non-trivial change would be replacing `value_of`. This should technically be updated to `try_get_one::<T>(...)` or `get_one::<T>(...)`, but I replaced it with a convenience function `try_get_signer`, which was added with `solana-clap-v3-utils 2.0`. I ended up removing the use of `DefaultSigner` and just added the logic to parse signers directly.

The feature `deprecated` in clap v3 allows us to toggle warnings from the use of deprecated functions. The deprecated functions are all removed from this crate, but other crates in the repo still uses some deprecated functions, so I removed this feature from clap for now.

This is an analogous PR to https://github.com/solana-labs/solana-program-library/pull/6952.

### DIFF
--- a/token-upgrade/cli/src/main.rs
+++ b/token-upgrade/cli/src/main.rs
@@ -3,7 +3,7 @@ use {
     solana_clap_v3_utils::{
         input_parsers::{
             parse_url_or_moniker,
-            signer::{try_pubkey_of, SignerSource, SignerSourceParserBuilder},
+            signer::{SignerSource, SignerSourceParserBuilder},
         },
         input_validators::normalize_to_url_if_moniker,
         keypair::{signer_from_path, signer_from_source, SignerFromPathConfig},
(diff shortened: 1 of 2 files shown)


token-upgrade/cli/src/main.rs
```diff
@@ -336,7 +334,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
                     .value_parser(SignerSourceParserBuilder::default().allow_all().build())
                     .value_name("MULTISIG_SIGNER")
                     .takes_value(true)
-                    .multiple(true)
+                    .action(ArgAction::Append)
```
joncinque, line 337: The documentation seems to imply that we should use `multiple_occurrences` instead, ie:

> Arg::multiple_occurrences in favor of Arg::multiple_values (positional), ArgAction::Append (flag with value), or ArgAction::Count (flag without value)

from https://github.com/clap-rs/clap/pull/3797, and:

> Replace Arg::multiple(bool) with Arg::multiple_values / Arg::multiple_occurrences

from https://github.com/clap-rs/clap/releases/tag/v3.0.0

Is there a reason to use your approach instead?

joncinque (commented):
Just the one question, but the rest looks good to me!

token-upgrade/cli/src/main.rs
samkim-crypto: Yeah `multiple` was deprecated in favor of `multiple_occurrences` in `v3.0.0`, but `multiple_occurrences` was deprecated in favor of `ArgAction` in `v3.2.0` (https://github.com/clap-rs/clap/releases/tag/v3.2.0) 😅.

joncinque: Haha I had a feeling, thanks for the info!

joncinque (approved):

PR #6952, opened 2024-06-30
[transfer-hook] Remove clap deprecated functions

samkim-crypto:
#### Problem
The transfer hook cli tool still uses some deprecated functions from clap-v2.

#### Summary
Updated the deprecated functions from clap-v2. Most of the changes should be straightforward:
- Replaced `Arg::with_name` to `Arg::new`, which is functionally identical
- Replaced `Arg::multiple` with `ArgAction::Append`, which should be functionally identical
- Replaced `is_present` with `try_contains_id`. The old `is_present` panics if the arg is not previously known/defined while `try_contains_id` gives an error. Otherwise, the two functions are identical.

The only non-trivial change would be replacing `value_of`. This should technically be updated to `try_get_one::<T>(...)` or `get_one::<T>(...)`, but I replaced it with a convenience function `try_get_signer`, which was added with `solana-clap-v3-utils 2.0`. I ended up removing the use of `DefaultSigner` and just added logic to parse signers directly.

I also used `SignerSource::try_get_pubkey` to remove some verbose parsing code (from https://github.com/solana-labs/solana-program-library/pull/6625).

The feature `deprecated` in clap v3 allows us to toggle warnings from the use of deprecated functions. The deprecated functions are all removed from this crate, but the other crates in the repo still uses some deprecated functions, so I removed this feature from clap for now.

### DIFF
--- a/token/transfer-hook/cli/Cargo.toml
+++ b/token/transfer-hook/cli/Cargo.toml
@@ -9,7 +9,7 @@ repository = "https://github.com/solana-labs/solana-program-library"
 version = "0.2.0"
 
 [dependencies]
-clap = { version = "3", features = ["cargo"] }
+clap = { version = "3", features = ["cargo", "deprecated"] }
 futures-util = "0.3.30"
 solana-clap-v3-utils = "2.0.0"
 solana-cli-config = "2.0.0"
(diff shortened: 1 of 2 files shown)


buffalojoec (approved):
Looks good!

PR #6953, opened 2024-06-30
token-2022: adjust `PodAccount` initialized check

buffalojoec:
#### Problem
The method to determine if a particular `PodAccount` is initialized simply checks
whether or not the `state` field (`AccountState`) is _not_ equal to zero.

However, this means that any number besides `0` is considered to be an initialized
`PodAccount`. This allows arbitrary data with size 165 to be considered a valid token
account. For example, `&[5; 165]` registers as an initialized token account.

#### Summary of Changes
Adjust the `is_initialized` method to explicitly check the two initialized `AccountState`
variants, rather than `!= 0 `.

### DIFF
--- a/token/program-2022/src/extension/mod.rs
+++ b/token/program-2022/src/extension/mod.rs
@@ -1512,7 +1512,9 @@ mod test {
             pubkey::Pubkey,
         },
         spl_pod::{
-            bytemuck::pod_bytes_of, optional_keys::OptionalNonZeroPubkey, primitives::PodU64,
+            bytemuck::pod_bytes_of,
+            optional_keys::OptionalNonZeroPubkey,
+            primitives::{PodBool, PodU64},
         },
         transfer_fee::test::test_transfer_fee_config,
     };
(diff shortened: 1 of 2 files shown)


token/program-2022/src/extension/mod.rs
```diff
@@ -1679,6 +1719,89 @@ mod test {
...
+        // all 5's - not a valid `AccountState`
+        let mut buffer = vec![5; BASE_ACCOUNT_LENGTH];
+        assert_eq!(
+            PodStateWithExtensions::<PodAccount>::unpack(&buffer),
+            Err(ProgramError::UninitializedAccount)
+        );
+        assert_eq!(
+            PodStateWithExtensionsMut::<PodAccount>::unpack(&mut buffer),
+            Err(ProgramError::UninitializedAccount)
+        );
```
buffalojoec, line 1749: This test case targets the exact problem.

joncinque (approved):

PR #6979, opened 2024-07-04
ci: update pnpm setup

liby:
Upgrading the `pnpm/action-setup` to v4 can prevent encountering these failed situations:

  - https://github.com/element-plus/element-plus/actions/runs/9788077597/job/27027285287
  - https://github.com/yazi-rs/yazi-rs.github.io/actions/runs/9786469651/job/27021395652

Upstream: https://github.com/pnpm/action-setup/issues/135#issuecomment-2206861174

### DIFF
--- a/.github/workflows/pull-request-account-compression.yml
+++ b/.github/workflows/pull-request-account-compression.yml
@@ -86,9 +86,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ env.NODE_VERSION }}
-      - uses: pnpm/action-setup@v2
-        with:
-          version: 8
+      - uses: pnpm/action-setup@v4
       - uses: actions/cache@v4
         with:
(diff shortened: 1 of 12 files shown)


joncinque (commented):
Thanks for your contribution! How about just removing the:

```
with:
  version: 8
```

Bit everywhere since the version is specified in the package.json?

liby:
> Thanks for your contribution! How about just removing the:
> 
> ```
> with:
>   version: 8
> ```
> 
> Bit everywhere since the version is specified in the package.json?

That's better. According to the [document description](https://github.com/pnpm/action-setup#install-only-pnpm-with-packagemanager) of `pnpm/action-setup`
> Omit version input to use the version in the [packageManager field in the package.json](https://nodejs.org/api/corepack.html).

joncinque (approved):
Thanks again for fixing this up!

PR #7003, opened 2024-07-10
temporary downgrade for spl-token dependency

DemidovVladimir:


joncinque:
Hi there, did you mean to put in this PR? I don't understand the purpose. It also seems to use much older dependencies.

PR #7005, opened 2024-07-10
ATA: Extract associated-token-account-client crate from associated-token-account crate

kevinheavey:
## Problem
People often import spl-associated-token-account just to derive an associated token address. This makes the ecosystem dependency hell and bloat problem much worse.

## Changes
- Make a new spl-associated-token-address that contains just the `get_associated_token_address*` functions.
- Re-export these functions in spl-associated-token-account with a deprecation notice

~This branches off #6989 so that should be merged first~


buffalojoec:
@kevinheavey looks like this one is also going to be closed, as per the solution proposed in #6989? Fwiw, we are close to said clients.

buffalojoec:
> This basically is the solution proposed in #6989 no?

Oops, I don't think I was following the bouncing ball over there.

This is looking pretty good, but I believe @joncinque had some additional thoughts so I'll let him weigh in on this one.

joncinque:
@kevinheavey I worked this a bit to be a more generic client crate that exposes the instruction creators and the address derivation functions. This way, we remove a dependency from token-client to the ata program, which means that I don't have to bump and publish the ata program every time there's a new version of token-2022, which always felt wrong.

Let me know what you think!

associated-token-account/client/src/address.rs
```diff
@@ -0,0 +1,69 @@
...
+    get_associated_token_address_and_bump_seed(
+        wallet_address,
+        token_mint_address,
+        &crate::id(),
+        token_program_id,
+    )
+    .0
+}
+
+/// For internal use only.
```
kevinheavey, line 54: Maybe a `#[doc(hidden)]` here?

kevinheavey:
@joncinque lgtm, full send

token/cli/tests/command.rs
```diff
@@ -167,7 +167,7 @@ async fn new_validator_for_test() -> (TestValidator, Keypair) {
             upgrade_authority: Pubkey::new_unique(),
         },
         UpgradeableProgramInfo {
-            program_id: spl_associated_token_account::id(),
+            program_id: spl_associated_token_account_client::id(),
```
buffalojoec, line 170: This is mostly a semantics thing, but do you think it looks a little strange to be using "client" here? It's implied that the ID is for the program that this package is a client to, but technically `client::id()` can also mean the client's ID.

What if instead it was `spl_associated_token_account_client::program::id()` or also `spl_associated_token_account_client::PROGRAM_ID`?

token-upgrade/cli/Cargo.toml
```diff
@@ -20,6 +20,7 @@ solana-logger = "2.0.3"
 solana-remote-wallet = "2.0.3"
 solana-sdk = "2.0.3"
 spl-associated-token-account = { version = "4.0.0", path = "../../associated-token-account/program", features = ["no-entrypoint"] }
+spl-associated-token-account-client = { version = "1.0.0", path = "../../associated-token-account/client" }
```
buffalojoec, line 23: Here you added the client but did not remove the program import. I think we can lose the program!

associated-token-account/program/Cargo.toml
```diff
@@ -12,11 +12,11 @@ no-entrypoint = []
 test-sbf = []
 
 [dependencies]
-assert_matches = "1.5.0"
 borsh = "1.5.1"
 num-derive = "0.4"
 num-traits = "0.2"
 solana-program = "2.0.3"
+spl-associated-token-account-client = { version = "1.0.0", path = "../client" }
```
buffalojoec, line 19: Interesting. Is depending on its own client an anti-pattern in the program clients world?

buffalojoec (dismissed):
Looking great! Are you planning to replace this with a generated client in the future (say, version 2.0.0)? A lot of the interface can be reusable, but it will also break. An alternative would be to try t to hand-write an IDL now and generate now.

associated-token-account/program/Cargo.toml
joncinque: In most cases, probably. In a world of generated clients, the instruction creators don't need to exist on the program crate at all, and should be exposed only on the client crate. In this case, however, we want to avoid breaking people that currently get the instruction creators from `spl-associated-token-account`.

For PDA creators, we might have to live with some replicode between the client and program crates.

token/cli/tests/command.rs
joncinque: we can do `program::id()` if you think that's clearer. What does "client's ID" refers to though?

token-upgrade/cli/Cargo.toml
joncinque: ah right, thanks!

token/cli/tests/command.rs
buffalojoec: I think it's a bit clearer yeah. It doesn't refer to anything, for me and you, it just might be confusing to people, especially newer-to-Solana devs. This is one of the first programs you encounter.

buffalojoec (approved):

associated-token-account/program/Cargo.toml
buffalojoec: Sounds reasonable, especially for the hand-rolled client here.

PR #7006, opened 2024-07-10
token-2022: add token 2022 argument to token cli for easy use

Woody4618:
Instead of copying the token program id of the token22 program one can now just write: 
```
create-token --token-program-2022
or
create-token -2
```

### DIFF
--- a/token/cli/src/clap_app.rs
+++ b/token/cli/src/clap_app.rs
@@ -657,12 +657,6 @@ pub fn app<'a, 'b>(
                         ),
                 )
                 .arg(
-                    Arg::with_name("program_2022")
-                        .value_name("TOKEN_22_PROGRAM_ID")
-                        .short("token22")
-                        .long("TokenProgram22")
-                        .help("Use token extension program token 2022 program id: TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb"))
-                .arg(
                     Arg::with_name("mint_authority")
                         .long("mint-authority")
                         .alias("owner")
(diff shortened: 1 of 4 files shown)


Woody4618:
Im also not 100% sure if restrict_to_program_id should be true or false. My guess reading line 2215 command_accounts is that it should be true.

token/cli/src/clap_app.rs
```diff
@@ -693,6 +693,13 @@ pub fn app<'a, 'b>(
                             "Enable the mint authority to close this mint"
                         ),
                 )
+                .arg(
+                    Arg::with_name("token_program_2022")
+                        .value_name("token-program-2022")
+                        .short("2")
+                        .long("token-program-2022")
+                        .takes_value(false)
+                        .help("Use token extension program token 2022 with program id: TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb"))
```
buffalojoec, line 702: This is in the wrong spot. The `--program-id` option lives under `app`:

https://github.com/solana-labs/solana-program-library/blob/d2d1fcd4402a8179fa57efbcba34f3273dfb9ce8/token/cli/src/clap_app.rs#L594-L603

Let's move it over there so it can apply to other commands besides just `create-token`, and be sure to use `.conflicts_with` to make sure someone can't provide this arg and `--program-id` together.

Also, what do you think about just `--program-2022`? Nice & short.

buffalojoec (commented):
This is a handy enhancement, but let's make it even more beneficial to users with a couple suggestions!

Thanks for rolling this in!

token/cli/src/clap_app.rs
Woody4618: Updated :) I added it only to the create-token because all others resolve the id automatically. But makes sense to have it global nevertheless :)

buffalojoec (dismissed):
Thanks!!

token/cli/src/clap_app.rs
```diff
@@ -591,6 +591,15 @@ pub fn app<'a, 'b>(
                 .possible_values(&["json", "json-compact"])
                 .help("Return information in specified output format"),
         )
+        .arg(
+            Arg::with_name("program_2022")
+                .short("2")
```
buffalojoec, line 596: This PR is ready to go in except for this. We really don't want to use a number for a short arg like this, since it manifests as `<command> -2` and that could easily be mistaken for a numerical negative two.

Can we just drop the `.short` here?

Woody4618: Sure 🫡
done

buffalojoec (approved):
Thanks!

PR #7021, opened 2024-07-17
[confidential-transfer] Organize transfer proof data using structs

samkim-crypto:
#### Problem
The constructor functions for proof data required for confidential transfer and transfer with fee returns a tuple of proof data structs. It would be cleaner and more readable to organize the return type using structs instead.

#### Summary of changes
Re-organize the proof data using structs.


joncinque (dismissed):
thanks for fixing this up!

PR #7026, opened 2024-07-18
[confidential-transfer] Add confidential transfer ciphertext arithmetic crate

samkim-crypto:
#### Problem
The syscall ElGamal ciphertext arithmetic functions (the logic in `solana_zk_token_sdk::zk_elgamal::ops`) have been removed in `solana-zk-sdk` as they are related specifically to the token program with plans to be moved into the spl repo.

#### Summary of changes
Added the confidential transfer `ciphertext-arithmetic` crate that implements the syscall ElGamal ciphertext arithmetic functions. 

This is essentially a copy of the logic from `solana_zk_token_sdk::zk_elgamal::ops` except for the following functions:
- `elgamal_ciphertext_to_ristretto(...)`: In the `zk-token-sdk`, `(`PodPedersenCommitment, PodDecryptHandle)` implemented `TryFrom` for `PodElGamalCiphertext`. However, this was removed in `zk-sdk` for simplicity. The `elgamal_ciphertext_to_ristretto(...)` was added as a function to directly convert from these two types. The return type is actually `(PodRistrettoPoint, PodRistrettoPoint)` because `PodRistrettoPoint` is actually the type that is used in the syscall functions.
- `ristretto_to_elgamal_ciphertext(...)`: This function is the opposite/complement of the function above. Unfortunately, there is one complication with this function. In `solana-zk-token-sdk`, the `PodElGamalCiphertext` is defined as `PodElGamalCiphertext(pub [u8; 64])`. However, in `solana-zk-sdk`, we have `PodElGamalCiphertext(pub(crate) [u8; 64])` and it does not expose a constructor that allows us to create `PodElGamalCiphertext` from array bytes directly. As a workaround, I ended up converting array bytes into base64 strings and then converting it to `PodElGamalCiphertext`, since `PodElGamalCiphertext` does implement `FromStr`... I think we can use this workaround for now until the next major agave release where we add a proper constructor in `solana-zk-sdk`.

### DIFF
--- a/token/confidential-transfer/ciphertext-arithmetic/src/lib.rs
+++ b/token/confidential-transfer/ciphertext-arithmetic/src/lib.rs
@@ -49,7 +49,8 @@ pub fn multiply(
     ))
 }
 
-/// Compute `left_ciphertext + (right_ciphertext_lo + 2^16 * right_ciphertext_hi)`
+/// Compute `left_ciphertext + (right_ciphertext_lo + 2^16 *
+/// right_ciphertext_hi)`
 pub fn add_with_lo_hi(
     left_ciphertext: &PodElGamalCiphertext,
     right_ciphertext_lo: &PodElGamalCiphertext,
@@ -78,7 +79,8 @@ pub fn subtract(
     ))
 }
 
-/// Compute `left_ciphertext - (right_ciphertext_lo + 2^16 * right_ciphertext_hi)`
+/// Compute `left_ciphertext - (right_ciphertext_lo + 2^16 *
+/// right_ciphertext_hi)`
 pub fn subtract_with_lo_hi(
     left_ciphertext: &PodElGamalCiphertext,
     right_ciphertext_lo: &PodElGamalCiphertext,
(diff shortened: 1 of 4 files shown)


token/confidential-transfer/ciphertext-arithmetic/src/lib.rs
```diff
@@ -0,0 +1,334 @@
...
+) -> PodElGamalCiphertext {
+    let mut ciphertext_bytes = [0u8; 64];
+    ciphertext_bytes[..32].copy_from_slice(bytes_of(commitment));
+    ciphertext_bytes[32..64].copy_from_slice(bytes_of(handle));
+    // Unfortunately, the `solana-zk-sdk` does not exporse a constructor interface
+    // to construct `PodRistrettoPoint` from bytes. As a work-around, encode the
+    // bytes as base64 string and then convert the string to a
+    // `PodElGamalCiphertext`.
+    let ciphertext_string = STANDARD.encode(ciphertext_bytes);
+    FromStr::from_str(&ciphertext_string).unwrap()
```
joncinque, line 157: Is it worth adding the constructor in future work? This seems a bit strange

joncinque (approved):
Just the one question for future work. We could add a `From<[u8; LEN]>` implementation for all of them. Doesn't need to be done immediately, of course

token/confidential-transfer/ciphertext-arithmetic/src/lib.rs
samkim-crypto: Yes, we should add the constructor on the zk-sdk side. I'l create an issue on the SPL side to use the new constructor once it lands.

PR #7033, opened 2024-07-19
stake-pool: Remove redelegate instruction from program and JS

joncinque:
#### Problem

This is required to land https://github.com/anza-xyz/agave/pull/2213, since the stake pool program uses the instruction.

The stake pool program uses the stake redelegate instruction, but it won't ever be released.

#### Solution

Remove the stake pool's redelegate instruction processor and deprecate the instruction creator. In the JS package, also remove the instruction creator.


2501babe (approved):

2501babe:
love to see it

PR #7055, opened 2024-07-26
[token-2022] Add support for reading proofs from record accounts

samkim-crypto:
#### Problem
The token22 program does not yet have support for reading zk proofs from record program.

#### Summary of Changes
Added support for reading zk proofs from the record program.

[7f510f4](https://github.com/solana-labs/solana-program-library/pull/7055/commits/7f510f402504f2f52aaf707023ba99e6af7b0533): Added support for the record program in the program. The main changes here are twofold:
- Updated the `decode_proof_instruction_context` function so that if the instruction data associated with the zk instruction is exactly 5 bytes, then read the proof from the supplied record account instead of reading it from instruction data. In the process, I changed the output type of the function to `Result<U, ProgramError>` instead of `Result<&U, ProgramError>` to simplify over-complication with lifetimes. All the functions that use `decode_proof_instruction_context` seem to immediately deref the context type anyway.
- Extended the `ProofLocation` type to have support for `RecordAccount`. Instead of having three variants of `InstructionOffset`, `ContextStateAccount`, and `RecordAccount`, I decided to overload `InstructionOffset` to support `RecordAccount` since this simplifies logic in some places.

[b245e14](https://github.com/solana-labs/solana-program-library/pull/7055/commits/b245e14ea71d866f8cf2441d462a2a48adb7627d): Added functions to create and close record accounts for confidential transfers. The main interesting part here is the `calculate_record_max_chunk_size` function, which provides the maximum chunk size for a proof record instruction to fit inside a single transaction. I included this function in `token.rs`, but I can move this to somewhere if suggested.

[92caf61](https://github.com/solana-labs/solana-program-library/pull/7055/commits/92caf61ae56b2cc480ab6b948b325726d1fec07e): For each of the instructions that require zkp's I added `...with_record_account()` test variants to make sure that the record account works. Exceptions are the transfer (with/without fee) instruction. I omitted adding tests for these instructions because I will need to heavily modify/remove these instructions in a follow-up PR.


token/client/src/temp_client.rs
joncinque, line 1: Was this meant to be included?

token/program-2022/src/lib.rs
```diff
@@ -121,6 +121,14 @@ pub fn check_zk_token_proof_program_account(zk_token_proof_program_id: &Pubkey)
     Ok(())
 }
 
+/// Checks that the supplied program ID is correct for the spl record program
+pub fn check_spl_record_program_account(spl_record_program_id: &Pubkey) -> ProgramResult {
+    if spl_record_program_id != &spl_record::id() {
+        return Err(ProgramError::IncorrectProgramId);
+    }
+    Ok(())
+}
```
joncinque, line 130: Do we need to enforce this? If not, it might be simpler to remove this check and the dependency from token-2022 to the record program

joncinque (commented):
I looked through a bit, and before I keep going, I wanted to check if we could somehow combine the proof source to be one of:

* the instructions sysvar
* context state account
* a record account

This might require updating `proof_instruction_offset` to be a bit clearer -- maybe we make it something like:

```
enum ProofLocation {
    ContextStateAccount
    InstructionOffset(i8),
    AccountData(u32),
}
```

Although it'll cause churn, I hope it'll be easier to understand -- what do you think?

token/client/src/temp_client.rs
samkim-crypto: Oh oops! I don't know how that got included 😅

token/program-2022/src/lib.rs
samkim-crypto: Yeah I thought about this as well. We don't necessarily need the proof to come from the record program, but I thought I would add it for safety. You are right. I think it will be simpler remove it and remove the dependency on the record program.

samkim-crypto:
> I looked through a bit, and before I keep going, I wanted to check if we could somehow combine the proof source to be one of:
> 
> * the instructions sysvar
> * context state account
> * a record account
> 
> This might require updating `proof_instruction_offset` to be a bit clearer -- maybe we make it something like:
> 
> ```
> enum ProofLocation {
>     ContextStateAccount
>     InstructionOffset(i8),
>     AccountData(u32),
> }
> ```
> 
> Although it'll cause churn, I hope it'll be easier to understand -- what do you think?

Yeah the current version is the result of a number of attempts to simplify things, but it is still a bit confusing. The complication is that even if the proof is read from an account, you still need the sysvar to access the zk-proof instruction in the transaction. Unlike the case of context state accounts, if the proof is stored in a record account, it is not yet verified, so an accompanying zk instruction has to be checked for the correct proof type. Furthermore, the record account address also has to be included because you still need access the proof to use it in the token-2022 processor logic.

What I initially had was something like the following
```
enum ProofLocation {
    ContextStateAccount(Pubkey),
    InstructionData(InstructionOffset),
    AccountData(InstructionOffset, Pubkey), // need both the offset for the accompanying zk instruction and address for the record account
}

type InstructionOffset = i8;
```
Then I combined `InstructionData` and `AccountData` together to simplify the logic in instruction generation, but looking at it now, maybe the three-variant enum above is the clearest way to define things. What do you think?

joncinque:
Ah gotcha, I knew I was missing something obvious, thanks for the explanation. In that case, it looks like you've done the best thing! I'll take a full look on Monday

token/program-2022-test/tests/confidential_transfer_fee.rs
```diff
@@ -612,6 +613,193 @@ async fn confidential_transfer_withdraw_withheld_tokens_from_mint() {
...
+            &withdraw_withheld_authority.pubkey(),
+            None,
+            None,
+            &withdraw_withheld_authority_elgamal_keypair,
+            alice_meta.elgamal_keypair.pubkey(),
+            &new_decryptable_available_balance.into(),
+            &[&withdraw_withheld_authority],
+        )
+        .await
+        .unwrap();
```
joncinque, line 708: Is this needed for the test? Or just to make sure that there's nothing on the mint?

joncinque (commented):
Just a couple of small things -- once I understood how it was working, it was pretty straightforward!

token/program-2022-test/tests/program_test.rs
```diff
@@ -42,7 +42,13 @@ pub struct TestContext {
...
 impl TestContext {
     pub async fn new() -> Self {
-        let program_test = ProgramTest::new("spl_token_2022", id(), processor!(Processor::process));
+        let mut program_test =
+            ProgramTest::new("spl_token_2022", id(), processor!(Processor::process));
+        program_test.add_program(
+            "spl_record",
+            spl_record::id(),
+            processor!(spl_record::processor::process_instruction),
+        );
```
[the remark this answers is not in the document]
samkim-crypto: Oh great idea! This seems to work!

token/program-2022/src/proof.rs
```diff
@@ -2,30 +2,59 @@
...
+    // the ZK ElGamal proof program.
+    if instruction.data.len() == INSTRUCTION_DATA_LENGTH_WITH_RECORD_ACCOUNT {
+        let record_account = next_account_info(account_info_iter)?;
+
+        // first byte is the proof type
+        let start_offset = u32::from_le_bytes(instruction.data[1..].try_into().unwrap()) as usize;
+        let end_offset = start_offset
+            .checked_add(std::mem::size_of::<T>())
+            .ok_or(ProgramError::InvalidAccountData)?;
+        let raw_proof_data = &record_account.data.borrow()[start_offset..end_offset];
```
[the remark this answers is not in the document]
samkim-crypto: Yes, thanks for the catch!

token/program-2022-test/tests/confidential_transfer_fee.rs
samkim-crypto: Yeah, this was inherited from withdraw withheld tokens from mint test without using record account to make sure that the command works even when there is no withheld tokens. For this specific test, we are focusing only on making sure that the record program works, so yes we should remove it. Thanks for the catch!

.github/workflows/pull-request-token.yml
```diff
@@ -246,6 +248,7 @@ jobs:
         run: |
           cargo build-sbf --manifest-path token/program-2022/Cargo.toml
           cargo build-sbf --manifest-path instruction-padding/program/Cargo.toml
+          cargo build-sbf --manifest-path record/program/Cargo.toml
```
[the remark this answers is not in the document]
samkim-crypto: Yes good point. Your suggestion of adding `prefer_bpf(false)` worked!

token/client/src/proof_generation.rs
```diff
@@ -32,6 +33,11 @@ use {
     },
 };
 
+pub enum ProofAccount {
+    ContextAccount(Pubkey),
+    RecordAccount(Pubkey),
```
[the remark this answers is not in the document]
samkim-crypto: Good idea! Added.

token/client/src/token.rs
```diff
@@ -1916,10 +1920,23 @@ where
...
+                }
+                ProofAccount::RecordAccount(record_account) => ProofLocation::InstructionOffset(
+                    1.try_into().unwrap(),
+                    ProofData::RecordAccount(
+                        record_account,
+                        RecordData::WRITABLE_START_INDEX as u32,
+                    ),
+                ),
+            }
         };
```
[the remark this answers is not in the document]
samkim-crypto: Yep, refactored into a separate function.

.github/workflows/pull-request-token.yml
joncinque, line 1: Can you revert these changes?

joncinque (approved):
Just one last little thing, then this can go in!

.github/workflows/pull-request-token.yml
samkim-crypto: Sorry I forgot to turn off the auto formatter...

joncinque (approved):

PR #7076, opened 2024-07-30
[libraries/pod]: improve `PodOption` type

febo:
### Problem

The `PodOption` type currently has 2 limitations:
1. It does not handle correctly the conversion of `Option::Some(T::default())` to `PodOption` since `T::default()` corresponds to the `None` value. As a result, it will convert a `Some` to a `None` value.
2. It always assume that `T:default()` is the `None` value, which might not be true in all cases and could be a limitation.

### Solution

To address (1), replace the `From` implementation for a `TryFrom`, where the conversion will fail when trying to convert a `Option::Some(T::default())` value to `PodOption`.

To address (2), add a new associated constant `NONE` to the `Nullable` trait instead of relying on the `default` value. Types implementing the trait can then specify the value that represents `None`.

### DIFF
--- a/libraries/pod/src/option.rs
+++ b/libraries/pod/src/option.rs
@@ -8,21 +8,26 @@
 
 use {
     bytemuck::{Pod, Zeroable},
-    solana_program::{program_option::COption, pubkey::Pubkey},
+    solana_program::{program_error::ProgramError, program_option::COption, pubkey::Pubkey},
 };
 
 /// Trait for types that can be `None`.
 ///
 /// This trait is used to indicate that a type can be `None` according to a
 /// specific value.
-pub trait Nullable: Default + Pod {
+pub trait Nullable: PartialEq + Pod + Sized {
     /// Indicates whether the value is `None` or not.
-    fn is_none(&self) -> bool;
+    fn is_none(&self) -> bool {
+        self == &Self::none()
+    }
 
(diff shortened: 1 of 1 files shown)


libraries/pod/src/option.rs
```diff
@@ -76,32 +93,33 @@ impl<T: Nullable> From<T> for PodOption<T> {
...
-            Some(value) => PodOption(value),
-            None => PodOption(T::default()),
+impl<T: Nullable> TryFrom<Option<T>> for PodOption<T> {
+    type Error = ProgramError;
+
+    fn try_from(value: Option<T>) -> Result<Self, Self::Error> {
+        match value {
+            Some(value) if value.is_some() => Ok(PodOption(value)),
+            None => Ok(PodOption(T::NONE)),
+            _ => Err(ProgramError::InvalidArgument),
```
joncinque, line 103: nit: this is totally fine, but how about making this exhaustive to be safe?
```suggestion
            Some(value) if value.is_none() => Err(ProgramError::InvalidArgument),
            Some(value) => Ok(PodOption(value)),
            None => Ok(PodOption(T::NONE)),
```

joncinque (commented):
Looks good! Just a couple of style nits

libraries/pod/src/option.rs
febo: Nice!

joncinque (approved):
Looks great

PR #7099, opened 2024-08-01
transfer-hook: Relax requirement of validation account

joncinque:
#### Problem

It's possible to write a perfectly good transfer hook without the extra account metas, and just the transfer hook program. For example, if you add a field to the metadata portion of the mint, a transfer hook program can check only that, so it has no need for anything else.

#### Summary of changes

Relax the requirement that the extra account metas account is provided. This was surprisingly straightforward, but maybe a little brittle, so let me know what you think.

### DIFF
--- a/token/program-2022-test/tests/transfer_hook.rs
+++ b/token/program-2022-test/tests/transfer_hook.rs
@@ -45,6 +45,15 @@ pub fn process_instruction_fail(
     Err(ProgramError::InvalidInstructionData)
 }
 
+/// Test program to succeed transfer hook, conforms to transfer-hook-interface
+pub fn process_instruction_success(
+    _program_id: &Pubkey,
+    _accounts: &[AccountInfo],
+    _input: &[u8],
+) -> ProgramResult {
+    Ok(())
+}
+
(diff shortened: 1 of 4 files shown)


token/program-2022-test/tests/transfer_hook.rs
```diff
@@ -913,3 +922,61 @@ async fn success_confidential_transfer() {
...
+    let token_context = context.token_context.take().unwrap();
+
+    let amount = 10;
+    let (alice_account, bob_account) =
+        setup_accounts(&token_context, Keypair::new(), Keypair::new(), amount).await;
+
+    // only add the transfer hook program id, nothing else
+    let token = token_context
+        .token
+        .with_transfer_hook_accounts(vec![AccountMeta::new_readonly(program_id, false)]);
```
buffalojoec, line 969: Maybe someday this can come out, too!

https://github.com/solana-foundation/solana-improvement-documents/pull/163

buffalojoec (approved):
Yep, this makes a lot of sense to me. Nice idea! Everything looks great.

PR #7101, opened 2024-08-01
spl-token-cli: support compute price + limit with native tokens

0xcypher02:
**Problem**

Currently the `spl-token wrap` and `spl-token unwrap` commands ignore the `--compute-budget-limit` and price arguments. With the current congestion on Solana, this makes it difficult to land transactions.

**Solution**

Do not ignore these values with native mints and instead replicate the `config_token_client` within `native_token_client_from_config` to properly check for priority fees and limits.

### DIFF
--- a/token/cli/src/command.rs
+++ b/token/cli/src/command.rs
@@ -195,6 +195,14 @@ fn native_token_client_from_config(
         config.fee_payer()?.clone(),
     );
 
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Nice catch, thanks for your contribution!

wrap

PR #7106, opened 2024-08-04
tlv: fix typos, update example code in README.md

jacksondoherty:


libraries/type-length-value/README.md
```diff
@@ -45,57 +44,63 @@ impl SplDiscriminate for MyOtherPodValue {
...
+    other_value1.data = 2;
+    
+    // Let's do it again, since we can now have repeating values!
+    let (other_value2, _) = state.init_value::<MyOtherPodValue>(true).unwrap();
+    assert_eq!(other_value2.data, 10);
+    // Update it in-place
+    other_value2.data = 4;
+    
+    // Later on, to work with it again, since we did _not_ allow repeating entries,
+    // we can just get the first value we encounter.
```
buffalojoec, line 84: Do you mind updating this comment to mention that specifically the `MyPodValue` type is the one we didn't allow repeats with? I think just rewording it is good.

buffalojoec (commented):
Looks great, thanks! Just one request for updating the comment, even though I know you didn't add it!

libraries/type-length-value/README.md
jacksondoherty: Done!

buffalojoec (approved):
Thanks!

PR #7112, opened 2024-08-05
program-error: Add option to specify solana_program crate

joncinque:
#### Problem

There are going to be a lot of issues with people seeing errors on
spl_program_error derivation. When a build pulls in two version of
solana_program, the macro uses `::solana_program`, which now becomes
ambiguous.

#### Summary of changes

I'm not sure if this is a good idea, but this PR gives the ability
to specify which version of `solana_program` to use when deriving the
various traits on your error type.

Borsh has this same functionality, and it's saved us when pulling in
multiple versions of borsh in the SDK.

Note: this PR defaults to `solana_program` instead of
`::solana_program`, which might cause downstream issues.

### DIFF
--- a/libraries/program-error/derive/src/macro_impl.rs
+++ b/libraries/program-error/derive/src/macro_impl.rs
@@ -51,33 +51,27 @@ impl MacroType {
 /// Builds the implementation of
 /// `Into<solana_program::program_error::ProgramError>` More specifically,
 /// implements `From<Self> for solana_program::program_error::ProgramError`
-pub fn into_program_error(
-    ident: &Ident,
-    solana_program_crate: &SolanaProgram,
-) -> proc_macro2::TokenStream {
+pub fn into_program_error(ident: &Ident, import: &SolanaProgram) -> proc_macro2::TokenStream {
     let this_impl = quote! {
-        impl From<#ident> for #solana_program_crate::program_error::ProgramError {
+        impl From<#ident> for #import::program_error::ProgramError {
             fn from(e: #ident) -> Self {
-                #solana_program_crate::program_error::ProgramError::Custom(e as u32)
+                #import::program_error::ProgramError::Custom(e as u32)
             }
         }
     };
-    solana_program_crate.wrap(this_impl)
+    import.wrap(this_impl)
 }
 
 /// Builds the implementation of `solana_program::decode_error::DecodeError<T>`
-pub fn decode_error(
-    ident: &Ident,
-    solana_program_crate: &SolanaProgram,
-) -> proc_macro2::TokenStream {
+pub fn decode_error(ident: &Ident, import: &SolanaProgram) -> proc_macro2::TokenStream {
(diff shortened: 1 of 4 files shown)


buffalojoec (commented):
Ahh clever, this could be useful. `num-derive` also lets you do this with `num-traits`, but it looks like it defaults to expecting the dependency to be in the crate's manifest.

https://github.com/rust-num/num-derive/blob/7cc33515dd2ae0eb43c5795c50ce49c554e8ba02/src/lib.rs#L152

However, they make use of a clever little trick - documented as being plucked from serde - to register the user's import as a local `_num_traits` module.

https://github.com/rust-num/num-derive/blob/7cc33515dd2ae0eb43c5795c50ce49c554e8ba02/src/lib.rs#L184-L191

Maybe we can do it this way for the default? I think that should avoid the ambiguous glob problem from https://github.com/solana-labs/solana-program-library/pull/6590. What do you think?

buffalojoec:
Additionally, now that I'm looking through the `num-traits` macro attribute, we could also maybe offer that as a passthrough arg so that we're not blocking access to the `#[num-traits]` attribute for downstream users, since we're leveraging it for each derive.

https://github.com/solana-labs/solana-program-library/blob/97c040009bdae72f8f63e5b920a4cd9ff99cbe09/libraries/program-error/derive/src/macro_impl.rs#L138

urb82:
It looks like you are encountering an issue related to the `spl_program_error` derivation when a build pulls in two versions of `solana_program`, causing ambiguity with the `::solana_program` macro.

To address this issue, you may need to consider the following steps:

1. **Dependency Management**: Ensure that your project's dependencies are correctly managed to avoid conflicting versions of `solana_program` being pulled in. You may need to review your dependencies and their versions to resolve any conflicts.

2. **Namespace Resolution**: When there are conflicting versions of a dependency, namespace resolution can become ambiguous. In this case, the `::solana_program` macro may be referencing a symbol that exists in multiple versions of `solana_program`, leading to ambiguity.

3. **Dependency Resolution Strategy**: Consider using dependency resolution strategies such as dependency pinning, dependency constraints, or dependency isolation to manage conflicting dependencies effectively.

4. **Update Code**: If possible, update your code to explicitly reference the correct version of `solana_program` to resolve the ambiguity. This may involve modifying the code that uses the `::solana_program` macro to specify the version or namespace explicitly

By addressing these points and potentially making adjustments to your dependency management and codebase, you can mitigate the issues related to the `spl_program_error` derivation and ambiguity with the `::solana_program` macro.

joncinque:
Thanks for the feedback! Let me know how this looks.

> Additionally, now that I'm looking through the num-traits macro attribute, we could also maybe offer that as a passthrough arg so that we're not blocking access to the #[num-traits] attribute for downstream users, since we're leveraging it for each derive.

Definitely makes sense, but I'd prefer to keep that for future work if that's ok with you

PR #7118, opened 2024-08-06
Fix `verify_transfer_with_fee_proof` broken by Rust 1.79.0

LucasSte:
**Problem**

Rust does not have a stable ABI, so Rust programs are subject to changes in their memory layout. Rust 1.79.0 will break function `verify_transfer_with_fee_proof` with the following error:

```
Error: Function _ZN14spl_token_20229extension21confidential_transfer12verify_proof30verify_transfer_with_fee_proof17h830056199538c531E Stack offset of 4512 exceeded max offset of 4096 by 416 bytes, please minimize large stack variables. Estimated function frame size: 4576 bytes.
```

This will make the following tests fail:

```
    confidential_transfer_transfer_with_fee_and_split_proof_context
    confidential_transfer_transfer_with_fee_and_split_proof_context_in_parallel
```

**Solution**

I split the function in two new methods.

### DIFF
--- a/token/program-2022/src/extension/confidential_transfer/verify_proof.rs
+++ b/token/program-2022/src/extension/confidential_transfer/verify_proof.rs
@@ -290,209 +290,15 @@ pub fn verify_transfer_with_fee_proof(
     fee_parameters: &TransferFee,
 ) -> Result<Option<TransferWithFeeProofContextInfo>, ProgramError> {
     if proof_instruction_offset == 0 && split_proof_context_state_accounts {
-        let equality_proof_context_state_account_info = next_account_info(account_info_iter)?;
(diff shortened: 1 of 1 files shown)


samkim-crypto:
@LucasSte, this function is actually on the way out with some of the new upcoming refactoring changes in the SPL repo. Can you give me a couple days to have these changes put in first?

LucasSte:
> @LucasSte, this function is actually on the way out with some of the new upcoming refactoring changes in the SPL repo. Can you give me a couple days to have these changes put in first?

Yes, no problem. Let me know when you are ready.

joncinque:
@LucasSte, I think @samkim-crypto has finished his work here -- do the tests pass if you use SPL's current master?

LucasSte:
All tests are passing now. Thanks!

PR #7130, opened 2024-08-11
token-group: Increase max size type to u64

### LINKED ISSUE #7061: Feature Request: Use `u64` for Token-Group Interface `max_size`
Although they don't exist today, there are conceivable NFT projects that could have billions of members, especially with compression. For example, a project I'm working on called [Garden](https://www.gardenbrowser.com/) will have a total possible supply of ~64B, which doesn't fit in `u32`. Unless there is a significant benefit to using `u32` that I'm not aware of, we should use the maximum type here.

jacksondoherty:
Resolves #7061

### Breaking Changes
- `token-group/js`: `bigint` used for `maxSize` and `memberNumber`
- `token/js`: `bigint` used for `maxSize`

### Notes
- I didn't update version numbers in `package.json` or the `pnpm-lock.yaml` file, it appears you all have a process for that.
- @buffalojoec I'm not sure if I ran into the issue you mentioned regarding the "extension state layout's alignment." The e2e tests appear to be working, but if you can elaborate a bit more on this I'm happy to make sure it's covered!


buffalojoec (approved):
Thanks @jacksondoherty - the changes look great! For the libraries, we'll follow the expected release cadence to publish the new changes, but we need to sort out what to do about the on-chain program (Token-2022) itself, since the feature for `token-group` was planned to be activated soon. There's a few things in-flight we need to sort out, then I can circle back here to merge this and get it in.

Thanks for throwing this together!

jacksondoherty:
Awesome, sounds good! Selfishly, I'm hoping this makes it on-chain as early as possible for my own projects. Let me know if there is any more help you need regarding things in-flight!

buffalojoec:
@jacksondoherty going in, thanks again!

PR #7143, opened 2024-08-13
Adding Curve Option When Creating ATA

QiLOL:
### Problem
In certain scenarios, such as with a DEX, it may be necessary to create an associated token account for a PDA (Program Derived Address). However, the current implementation of `createAssociatedTokenAccount` in the `@solana/spl-token` library has the `allowOffCurve` option set to false by default while `getAssociatedTokenAddressSync`.
(Reference: [GitHub link](https://github.com/solana-labs/solana-program-library/blob/master/token/js/src/actions/createAssociatedTokenAccount.ts#L29))

To create an ATA for a PDA, a custom implementation is required.

### Solution
This PR aims to enhance code reusability by allowing users to set the `allowOffCurve` option when using `createAssociatedTokenAccount`.

### DIFF
--- a/token/js/src/actions/createAssociatedTokenAccount.ts
+++ b/token/js/src/actions/createAssociatedTokenAccount.ts
@@ -28,7 +28,13 @@ export async function createAssociatedTokenAccount(
     associatedTokenProgramId = ASSOCIATED_TOKEN_PROGRAM_ID,
     allowOwnerOffCurve = false,
 ): Promise<PublicKey> {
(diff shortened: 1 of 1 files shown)


buffalojoec (approved):
Lgtm, thanks!

buffalojoec (changes requested):
Whoops sorry. Looks like a formatting issue.

QiLOL:
> Whoops sorry. Looks like a formatting issue.

Thank you, I just updated it; please check!

buffalojoec (approved):

PR #7144, opened 2024-08-13
Add tagging automation for crates

buffalojoec:
#### Problem
Whenever we increment versions for crates in SPL, we have to remember to manually create tags for these new packages on the repository, and commit those tags to the remote. It would be nice if this was just a part of CI.

#### Summary of Changes
Add two new scripts and a GitHub action:

- `ci/detect-changed-manifests.sh`: Detects changes to a manifest's version and returns a list of all manifests whose version changed.
- `ci/tag-releases.sh`: Creates a new Git tag for each changed manifest.
- `workflows/release.yml`: Workflow for `master` to create new tags and commit them to the repository whenever manifest versions change in a merged PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,23 @@
+name: Release
+
+on push:
+  branches:
+    - master
+  paths-ignore:
+    - 'docs/**'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Create tags
+        id: create_tags
+        run: |
+          ci/tag-new-crates.sh
+          git push --tags
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/ci/detect-changed-manifests.sh
+++ b/ci/detect-changed-manifests.sh
@@ -1,0 +1,25 @@
+# Source this file.
+#
+# Detects any manifests whose versions have changed.
+# Does so by comparing the current version in the Cargo.toml with the previous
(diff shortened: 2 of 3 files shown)


buffalojoec:
@joncinque The next step after this change would be to roll something similar to the monorepo's [`order-crates-for-publishing.py`](https://github.com/anza-xyz/agave/blob/master/ci/order-crates-for-publishing.py) and use it in the `release.yml` workflow to not only tag the repo but also publish the crate. If we wanted to get even more interesting, we could add automation to publish releases via [GitHub's API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release) as well.

Let me know how far you want to go with this effort!

joncinque (commented):
I'm not sure we should be automatically tagging as soon as a crate is bumped. If we accidentally bump too early or anything like that, then we can get into a bad situation.

If we do some automation for publishing, it would be nice to have something similar to the counter shank example at https://github.com/solana-program/counter-shank/blob/main/.github/workflows/publish-rust-client.yml and https://github.com/solana-program/counter-shank/blob/main/scripts/client/publish-rust.mjs

I need to test and see how `cargo release` handles cargo workspaces, but it would be really neat to have a manually triggered workflow that:

* bumps the version everywhere (specified as a workflow parameter)
* runs fmt / clippy / tests / dry-run-publish to make sure it's ok
* creates the tag and pushes it
* creates a GitHub release on that tag
* publishes the crate to crates.io

And we could use this type of automation for the SDK monorepo. What do you think?

buffalojoec:
I just tested `cargo release` and it seems to work fine in our workspace. I hadn't looked at these publish scripts nor `cargo release` before. Looks like everything I've PR'ed here and in #7142 is covered by them.

PR #7152, opened 2024-08-15
Add check withdraw-withheld-tokens has sources or include mint

### LINKED ISSUE #7042: token-cli: withdraw-withheld-tokens silently fails
Copied from https://github.com/anza-xyz/agave/issues/2224 cc @tapirtoken

#### Problem

If you look at this transaction:

4N1GvNc27DFhKK7Y8dcjMhrDuze4LMfM1qhoG6nkFG2EWTMPEcuevXVGEXmwt6gR55VEskQ9X37Z8RKkP1HhgL8i

You'll notice that instruction 5 (on Solscan) is HarvestWithheldTokensToMint. However, the Withdraw Authority is Null so this operation should normally fail. But this is an exceptional case in which the tokens are being immediately burned (by instruction 6) in the same atomic transaction, so the harvest is allowed in order to permit the wallet owner to recover the rent.

Nevertheless we wanted to make absolutely sure that there was no way to withdraw withheld tokens, so we tried to withdraw our own:

spl-token withdraw-withheld-tokens 8QhMP4HVfUd78hSjwzRXmifYCWAYB2nYiM6Emd7fTstd  --fee-payer AirDpo61FVsNDoWSBoBX3K4gcBiTVnkfZRAZk7bKfRtX.json --url mainnet-beta --output json --program-id TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb --withdraw-withheld-authority AirDpo61FVsNDoWSBoBX3K4gcBiTVnkfZRAZk7bKfRtX.json

This command just terminates as though it succeeded. It definitely didn't succeed because our Withheld Amount on explorer.solana.com is unchanged:

8QhMP4HVfUd78hSjwzRXmifYCWAYB2nYiM6Emd7fTstd

This is correct behavior but withdraw-withheld-tokens should have returned an error message.

#### Proposed Solution

Just check if the operation succeeded and report to the user.

zzau13:
Just check both conditions and return error. Clap doesn't have the method for check that, so I do check in command logic.
Fixes #7042

### DIFF
--- a/token/cli/src/clap_app.rs
+++ b/token/cli/src/clap_app.rs
@@ -2271,6 +2271,13 @@ pub fn app<'a, 'b>(
                 )
                 .arg(owner_address_arg())
                 .arg(multisig_signer_arg())
+                .group(
+                    ArgGroup::with_name("source_or_mint")
+                        .arg("source")
+                        .arg("include_mint")
+                        .multiple(true)
+                        .required(true)
+                )
         )
         .subcommand(
             SubCommand::with_name(CommandName::SetTransferFee.into())


joncinque (approved):
Thanks for your contribution!

PR #7164, opened 2024-08-17
[token-2022] Remove unnecessary split proof generation logic

samkim-crypto:
#### Problem
Currently, the token-client handles to create confidential transfer instructions that require split zkp's always generate the proof data. However, this is wasteful in that if all of the proofs are pre-verified using a context-state account or already generated and stored in a record account, then there is no need to generate the split proofs.

#### Summary of Changes
Update the client logic so that for the special case of when the proof accounts are all `Some(...)`, we skip proof generation.

### DIFF
--- a/token/client/src/token.rs
+++ b/token/client/src/token.rs
@@ -2083,19 +2083,26 @@ where
             WithdrawAccountInfo::new(confidential_transfer_account)
         };
 
-        let WithdrawProofData {
-            equality_proof_data,
-            range_proof_data,
-        } = account_info
-            .generate_proof_data(withdraw_amount, elgamal_keypair, aes_key)
-            .map_err(|_| TokenError::ProofGeneration)?;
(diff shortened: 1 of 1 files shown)


token/client/src/token.rs
```diff
@@ -2483,39 +2507,67 @@ where
...
+            equality_proof_account,
+            transfer_amount_ciphertext_validity_proof_account,
+            percentage_with_cap_proof_account,
+            fee_ciphertext_validity_proof_account,
+            range_proof_account,
+        ]
+        .iter()
+        .all(|proof_account| proof_account.is_some())
+        {
+            // is all proofs come from accounts, then skip proof generation
```
joncinque, line 2526: nit:
```suggestion
            // if all proofs come from accounts, then skip proof generation
```

joncinque (approved):
Looks great!

Issue #7059, opened 2024-07-28
token-cli: Unable to use the token cli to transfer tokens that have the transfer-fee and transfer-hook extensions. 


tonton-sol:
I am trying to test using the transfer-fee extension in conjunction with the transfer-hook extension. The transfer-hook is the provided example program given in this repository. The transfer hook extra account metas account is initialized but empty. I am running this command:

```bash
spl-token transfer --expected-fee <EXPECTED_FEE> <TOKEN_MINT_ADDRESS> <TOKEN_AMOUNT> <RECIPIENT_WALLET_ADDRESS>
```

It fails with error: `Program TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb failed: custom program error: 0x7dc8348c` My suspicion is that TransferCheckedWithFee is not correctly resolving the extra account metas. 

I have already confirmed that this same transfer-hook program works and successfully transfers a token with only the transfer-hook extension.  

Solution: Fix TransferCheckedWithFee to correctly resolve the required extra account metas.

joncinque:
Thanks for your report! This is indeed an issue.

It looks like we'll need to do a few things to fix this properly.

* in token-2022, add an `offchain` helper (and probably an `onchain` helper while we're at it) which also takes in the expected fee, something like `create_transfer_checked_with_fee_instruction_with_extra_metas`, which could resolve to some common helper function with https://github.com/solana-labs/solana-program-library/blob/7d35ad2ed2bc1e6c7f2a841526711c47f5093438/token/program-2022/src/offchain.rs#L37
* in token-client, use that offchain helper in `transfer_with_fee` https://github.com/solana-labs/solana-program-library/blob/7d35ad2ed2bc1e6c7f2a841526711c47f5093438/token/client/src/token.rs#L1149, very similar to how the `transfer` function works https://github.com/solana-labs/solana-program-library/blob/7d35ad2ed2bc1e6c7f2a841526711c47f5093438/token/client/src/token.rs#L992
* token-cli already uses `transfer_with_fee`, so that should Just Work afterwards

tonton-sol:
I will start on this!

joncinque:
Resolved by #7171, thanks!

PR #7171, opened 2024-08-20
[token-cli] Add support for transfer-hook account resolution for transfers with the transfer-fee extension.

tonton-sol:
Per #7059 

Adds a new offchain helper `offchain::create_transfer_checked_with_fee_instruction_with_extra_metas`.
Adds a new onchain helper `onchain::invoke_transfer_checked_with_fee`

Adds logic to `transfer_with_fee` in token/client/src/token.rs to add the extra-account-metas using the offchain helper if the token uses the transfer-hook extension.

### DIFF
--- a/token/program-2022-test/tests/transfer_hook.rs
+++ b/token/program-2022-test/tests/transfer_hook.rs
@@ -39,7 +39,7 @@ use {
 };
 
 const TEST_MAXIMUM_FEE: u64 = 10_000_000;
-const TEST_FEE_BASIS_POINTS: u16 = 250;
+const TEST_FEE_BASIS_POINTS: u16 = 100;
 
 /// Test program to fail transfer hook, conforms to transfer-hook-interface
 pub fn process_instruction_fail(
@@ -138,6 +138,58 @@ pub fn process_instruction_swap(
         remaining_accounts,
         1,
         9,
+        &[],
+    )?;
+
+    Ok(())
+}
+
+// Test program to transfer two types of tokens with transfer hooks at once with
+// fees
+pub fn process_instruction_swap_with_fee(
+    _program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    _input: &[u8],
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+
+    let source_a_account_info = next_account_info(account_info_iter)?;
+    let mint_a_info = next_account_info(account_info_iter)?;
+    let destination_a_account_info = next_account_info(account_info_iter)?;
+    let authority_a_info = next_account_info(account_info_iter)?;
+    let token_program_a_info = next_account_info(account_info_iter)?;
+
+    let source_b_account_info = next_account_info(account_info_iter)?;
+    let mint_b_info = next_account_info(account_info_iter)?;
+    let destination_b_account_info = next_account_info(account_info_iter)?;
+    let authority_b_info = next_account_info(account_info_iter)?;
+    let token_program_b_info = next_account_info(account_info_iter)?;
+
+    let remaining_accounts = account_info_iter.as_slice();
+
+    onchain::invoke_transfer_checked_with_fee(
+        token_program_a_info.key,
+        source_a_account_info.clone(),
+        mint_a_info.clone(),
+        destination_a_account_info.clone(),
+        authority_a_info.clone(),
+        remaining_accounts,
+        1_000_000_000,
+        9,
+        10_000_000,
+        &[],
+    )?;
+
+    onchain::invoke_transfer_checked_with_fee(
+        token_program_b_info.key,
+        source_b_account_info.clone(),
+        mint_b_info.clone(),
+        destination_b_account_info.clone(),
+        authority_b_info.clone(),
+        remaining_accounts,
+        1_000_000_000,
+        9,
+        10_000_000,
         &[],
     )?;
 
@@ -973,7 +1025,7 @@ async fn success_transfers_with_fee_using_onchain_helper() {
     let mint_a = mint_a_keypair.pubkey();
     let mint_b_keypair = Keypair::new();
     let mint_b = mint_b_keypair.pubkey();
-    let amount = 10;
+    let amount = 10_000_000_000;
 
     let transfer_fee_config_authority = Keypair::new();
     let withdraw_withheld_authority = Keypair::new();
@@ -985,7 +1037,7 @@ async fn success_transfers_with_fee_using_onchain_helper() {
     program_test.add_program(
         "my_swap",
         swap_program_id,
-        processor!(process_instruction_swap),
+        processor!(process_instruction_swap_with_fee),
     );
     add_validation_account(&mut program_test, &mint_a, &program_id);
     add_validation_account(&mut program_test, &mint_b, &program_id);
--- a/token/program-2022-test/tests/transfer_hook.rs
+++ b/token/program-2022-test/tests/transfer_hook.rs
@@ -966,6 +966,150 @@ async fn success_transfers_using_onchain_helper() {
 }
 
 #[tokio::test]
+async fn success_transfers_with_fee_using_onchain_helper() {
+    let authority = Pubkey::new_unique();
+    let program_id = Pubkey::new_unique();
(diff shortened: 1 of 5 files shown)


tonton-sol:
What tests if any needed to be added?

buffalojoec (commented):
Nice!! The changes look great to me.

For the tests, you'll want to test the helpers by adding a few cases to the Token-2022 tests under `token/program-2022-test` in the `transfer_hook` test module.

The idea is to basically make copies of `success_transfer` and `success_transfers_using_onchain_helper`, but add the transfer fee config to the mint. The former is going to test with the transfer hook example program, which should still work fine for this case, since it's really just going to test that the proper extra metas are loaded and passed to the example program in `token/transfer-hook/example`. The latter is using a mocked-up processor in the test module (see [this line here](https://github.com/solana-labs/solana-program-library/blob/42d465ab07574079cc207ff6418f3de4faac4a6b/token/program-2022-test/tests/transfer_hook.rs#L713-L717)).

Then maybe add a block in the Token CLI's `transfer_hook` tests to ensure we can parse a transfer instruction with a fee and all the extra transfer hook accounts as arguments.

If you have any questions, let me know, and thanks!

tonton-sol:
@buffalojoec I am working in `token/program-2022-test/transfer_hook.rs` and I noticed that rust-analyser/cargo is not working in this directory. Is this a known thing?

tonton-sol:
And actually, does it make sense to add the combination test here in `token/program-2022-test`? Shouldn't there then be tests for every arbitrary combination of extensions?

buffalojoec:
> @buffalojoec I am working in `token/program-2022-test/transfer_hook.rs` and I noticed that rust-analyser/cargo is not working in this directory. Is this a known thing?

Yeah, you need to add `test-sbf` to Rust Analyzer's default features. Even after doing that, I don't think it ever _really_ worked for me. Sometimes I just comment out the `#[cfg(feature = "test-sbf")]` at the top, handle all of my compile errors, then add it back.

> And actually, does it make sense to add the combination test here in token/program-2022-test? Shouldn't there then be tests for every arbitrary combination of extensions?

I didn't want to dump that whole workload on you. 😅

tonton-sol:
Looking at the rest of the tests, I actually think its okay to have these mixed extension tests in `token/program-2022-test/transfer_hook.rs` because of the special treatment needed for the extra metas. So I will go ahead and add tests for the w/ fee conditions. 

Also there are no tests here for  `spl_token_2022::offchain`, only `spl_transfer_hook_interface`. Is this by design? Would it even make sense to test the higher level helper functions here?

tonton-sol:
> Then maybe add a block in the [Token CLI's transfer_hook tests](https://github.com/solana-labs/solana-program-library/blob/42d465ab07574079cc207ff6418f3de4faac4a6b/token/cli/tests/command.rs#L3813-L3852) to ensure we can parse a transfer instruction with a fee and all the extra transfer hook accounts as arguments. 

I am having trouble trying to add the arguments for the expected fee to the transfer test. I have confirmed that I can successfully create the token with both the transfer-hook and transfer-fee extensions, so this is not a problem. Could this be caused by '--sign-only` and doing  `process_command()` in offline mode?

I assume it's a catch-22 because we need the associated token account to be created meaning it cannot be done in offline mode. If it is online mode, the transfer-hook program would need to be deployed and initialized. Is this correct?

buffalojoec:
> > Then maybe add a block in the [Token CLI's transfer_hook tests](https://github.com/solana-labs/solana-program-library/blob/42d465ab07574079cc207ff6418f3de4faac4a6b/token/cli/tests/command.rs#L3813-L3852) to ensure we can parse a transfer instruction with a fee and all the extra transfer hook accounts as arguments.
> 
> I am having trouble trying to add the arguments for the expected fee to the transfer test. I have confirmed that I can successfully create the token with both the transfer-hook and transfer-fee extensions in the test, so this is not a problem. Could this be caused by `--sign-only` and doing `process_command()` in offline mode?
> 
> I assume it's a catch-22 because we need the associated token account to be created meaning it cannot be done in offline mode. If it is online mode, the transfer-hook program would need to be deployed and initialized. Is this correct?

Yeah that's correct. The test I linked is sending the command to create the token, then disable the transfer hook (set program to `None`), but the middle bit where made the permalink is doing `--sign-only` so that it doesn't actually invoke the client but it's testing the parsing of the arguments and validation against the mint's extensions.

Maybe duplicate that test and set the token up with a fee config, similar to [this test here](https://github.com/solana-labs/solana-program-library/blob/42d465ab07574079cc207ff6418f3de4faac4a6b/token/cli/tests/command.rs#L2312-L2325), and then add the fee argument to the parsing test block?

tonton-sol:
Okay I figured it out! In sign-only + offline mode you need to explicitly pass the `--program-id` or else it will default to Tokenkeg. This is a problem for the transfer-fee extension but not the transfer-hook extension because `transfer_checked()` [checks if it is either Tokenkeg or Token2022](https://github.com/solana-labs/solana-program-library/blob/ab830053c59c9c35bc3a727703aacf40c1215132/token/program-2022/src/instruction.rs#L1637)  while `transfer_checked_with_fee()` [checks if it is Token2022](https://github.com/solana-labs/solana-program-library/blob/ab830053c59c9c35bc3a727703aacf40c1215132/token/program-2022/src/extension/transfer_fee/instruction.rs#L298).  I guess this is just a weird quirk of testing this way.

tonton-sol:
I think this is good to go!

token/program-2022-test/tests/transfer_hook.rs
```diff
@@ -261,6 +266,63 @@ async fn setup(mint: Keypair, program_id: &Pubkey, authority: &Pubkey) -> TestCo
...
+fn test_transfer_fee_config() -> TransferFeeConfig {
+    let transfer_fee = test_transfer_fee();
+    TransferFeeConfig {
+        transfer_fee_config_authority: COption::Some(Pubkey::new_unique()).try_into().unwrap(),
+        withdraw_withheld_authority: COption::Some(Pubkey::new_unique()).try_into().unwrap(),
+        withheld_amount: 0.into(),
+        older_transfer_fee: transfer_fee,
+        newer_transfer_fee: transfer_fee,
+    }
+}
```
buffalojoec, line 324: nit: I don't think these functions are really necessary. `test_transfer_fee` is only ever called by `test_transfer_fee_config`, and `test_transfer_fee_config` is just called once by the test you've added.

I'd suggest just initializing the `TransferFeeConfig` object locally within the test, and shedding these helpers.

token/program-2022-test/tests/transfer_hook.rs
```diff
@@ -821,6 +965,150 @@ async fn success_transfers_using_onchain_helper() {
...
+    let transfer_fee_basis_points = TEST_FEE_BASIS_POINTS;
+    let maximum_fee = TEST_MAXIMUM_FEE;
+
+    let swap_program_id = Pubkey::new_unique();
+    let mut program_test = setup_program_test(&program_id);
+    program_test.add_program(
+        "my_swap",
+        swap_program_id,
+        processor!(process_instruction_swap),
+    );
```
buffalojoec, line 989: This test isn't actually testing the new on-chain helper for `invoke_transfer_checked_with_fee`, since it's calling the swap program processor defined earlier in this file, which does a swap by calling `invoke_transfer_checked`. It's just invoking `TransferChecked` and debiting the base fee.

You can just create another processor with the new helpers to test them, though!

buffalojoec (commented):
Nice! This is super close, just a nit and comment about the one test, then we can ship.

token/program-2022-test/tests/transfer_hook.rs
tonton-sol: Good catch, thank you!

buffalojoec (approved):
Nice work!!

tonton-sol:
Will this auto close and not be added? 😅

buffalojoec:
> Will this auto close and not be added? 😅

Whoops, sorry! I thought you could merge after I approved.

tonton-sol:
I'd need to be employed to do that right? 😉

PR #7179, opened 2024-08-22
[token-client] remove `curve25519-dalek` dependency

samkim-crypto:
#### Problem
Sorry, the split proof generation logic has been removed from token-client, but I forgot to remove the `curve25519-dalek` dependency.

#### Summary of Changes
Remove `curve25519-dalek` dependency in `Cargo.toml`.


joncinque (approved):
Ah great, thanks!

PR #7181, opened 2024-08-22
add short createAssociatedTokenAccountIdempotentInstruction

### LINKED ISSUE #7178: token-js: `createAssociatedTokenAccountIdempotentInstruction` shouldn't require client to pass in `associatedToken`
The handy [`createAssociatedTokenAccountIdempotentInstruction`](https://github.com/solana-labs/solana-program-library/blob/69257a6959b4abac239235092679e8044fb18d39/token/js/src/instructions/associatedTokenAccount.ts#L48C17-L65) allows you create instructions that idempotently create associated token accounts. One thing that annoys me about it is that it requires me to pass in the associated token account, like so:
```
...
.preInstructions([
        createAssociatedTokenAccountIdempotentInstruction(
          this.provider.publicKey,
          token.getAssociatedTokenAddressSync(
            tokenMint,
            pdaOwner,
            true
          ),
          pdaOwner,
          tokenMint
)]
...
```
I don't see the purpose of this given that associated token accounts can always be derived by the `getAssociatedTokenAddressSync`. Maybe this function was written before `getAssociatedTokenAddressSync` existed? In any case, it would be nice if this parameter was made optional or removed entirely.

zzau13:
Fixes #7178


token/js/src/instructions/associatedTokenAccount.ts
```diff
@@ -64,6 +65,31 @@ export function createAssociatedTokenAccountIdempotentInstruction(
...
+    payer: PublicKey,
+    owner: PublicKey,
+    mint: PublicKey,
+    allowOwnerOffCurve = true,
+    programId = TOKEN_PROGRAM_ID,
+    associatedTokenProgramId = ASSOCIATED_TOKEN_PROGRAM_ID,
+) {
+    const associatedToken = getAssociatedTokenAddressSync(mint, owner, allowOwnerOffCurve);
+
+    return createAssociatedTokenAccountIdempotentInstruction(payer, owner, mint, programId, associatedTokenProgramId);
```
joncinque, line 90: This isn't using the associated token account address that was just derived, so it's incorrect. You'll need to do:
```suggestion
    return createAssociatedTokenAccountIdempotentInstruction(payer, associatedToken, owner, mint, programId, associatedTokenProgramId);
```

token/js/src/instructions/associatedTokenAccount.ts
```diff
@@ -64,6 +65,31 @@ export function createAssociatedTokenAccountIdempotentInstruction(
...
+ * @param payer                    Payer of the initialization fees
+ * @param owner                    Owner of the new account
+ * @param mint                     Token mint account
+ * @param allowOwnerOffCurve       Allow the owner account to be a PDA (Program Derived Address)
+ * @param programId                SPL Token program account
+ * @param associatedTokenProgramId SPL Associated Token program account
+ *
+ * @return Instruction to add to a transaction
+ */
+export function createAssociatedTokenAccountIdempotentInstructionShort(
```
joncinque, line 80: nit: can we rename this to something a bit clearer? maybe `createAssociatedTokenAccountIdempotentInstructionWithDerivation`?

token/js/src/instructions/associatedTokenAccount.ts
```diff
@@ -64,6 +65,31 @@ export function createAssociatedTokenAccountIdempotentInstruction(
     );
 }
 
+/**
+ * Construct a createAssociatedTokenAccountIdempotentInstructionShort instruction
```
joncinque, line 69: nit: let's keep the comment correct
```suggestion
 * Derive the associated token account and construct a CreateAssociatedTokenAccountIdempotent instruction
```

joncinque (commented):
Thanks for your contribution! A couple of things to fixup. Could you also add a little unit test to make sure that the output is exactly the same between the new function and the old function?

PR #7183, opened 2024-08-22
add derivation of createAssociatedTokenAccountIdempotentInstruction

### LINKED ISSUE #7178: token-js: `createAssociatedTokenAccountIdempotentInstruction` shouldn't require client to pass in `associatedToken`

zzau13:
Fixes #7178

### DIFF
--- a/token/js/src/instructions/associatedTokenAccount.ts
+++ b/token/js/src/instructions/associatedTokenAccount.ts
@@ -1,6 +1,7 @@
 import type { PublicKey } from '@solana/web3.js';
 import { SystemProgram, TransactionInstruction } from '@solana/web3.js';
 import { ASSOCIATED_TOKEN_PROGRAM_ID, TOKEN_PROGRAM_ID } from '../constants.js';
+import { getAssociatedTokenAddressSync } from '../state/mint.js';
 
 /**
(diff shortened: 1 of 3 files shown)


joncinque (commented):
Thanks for your contribution! All of the code changes look good, I just have a question on the comment that was added

joncinque (approved):
Thanks for the contribution!

PR #7184, opened 2024-08-23
CI: Add manual workflow to publish crates

joncinque:
#### Problem

There are a lot of crates in SPL, and in the future SDK monorepo, so it would be extremely useful to have a manual GitHub Actions workflow to test a package, bump its version everywhere, publish it, and create a release on GitHub.

#### Summary of changes

Following the excellent scripts at the
[counter-shank repo](https://github.com/solana-program/counter-shank/blob/main/.github/workflows/publish-rust-client.yml), port the zx bits to bash, and do mostly the same thing, except allowing for multiple packages to be published. Thanks @lorisleiva!

The tests before publish are just the steps from `pull-request.yml`.

You can also see my testing around with this at https://github.com/joncinque/counter-shank

### DIFF
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -1,0 +1,203 @@
+name: Publish Rust Crate
+
+on:
+  workflow_dispatch:
+    inputs:
+      package_path:
+        description: Path to directory with package to release
+        required: true
+        type: string
+      level:
+        description: Level
+        required: true
+        default: patch
+        type: choice
+        options:
+          - patch
+          - minor
+          - major
+          - rc
+          - beta
+          - alpha
+          - release
+          - version
+      version:
+        description: Version (used with level "version")
+        required: false
+        type: string
+      dry_run:
+        description: Dry run
+        required: true
+        default: true
+        type: boolean
+      create_release:
+        description: Create a GitHub release
+        required: true
+        type: boolean
+        default: true
+
+jobs:
+  rustfmt:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


yihau (approved):
👑

.github/workflows/publish-rust.yml
```diff
@@ -0,0 +1,203 @@
...
+          token: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+        if: ${{ env.token == '' }}
+        run: |
+          echo "The CARGO_REGISTRY_TOKEN secret variable is not set"
+          echo "Go to \"Settings\" -> \"Secrets and variables\" -> \"Actions\" -> \"New repository secret\"."
+          exit 1
+
+      - name: Set Git Author
+        run: |
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
buffalojoec, line 173: Is this always the same number, or does it depend on the action? Might not even matter here, just curious.

buffalojoec (approved):
Lgtm!!

.github/workflows/publish-rust.yml
lorisleiva: I'm not 100% sure but I think this number is some optional ID you can add to distinguish separate GH bots. So technically we could remove that altogether but also having a fixed identifier in case we need a separate bot in the future may be helpful.

joncinque: According to https://github.com/orgs/community/discussions/26560, this makes it so that the commit is properly tied to the github actions bot, using its user id, which is also reported at https://api.github.com/users/github-actions%5Bbot%5D

PR #7207, opened 2024-08-28
Bump token-metadata-js and token-group-js

### LINKED ISSUE #7202: Request: Release new spl-token-metadata to npm
The latest version of the JS package spl-token-metadata published to NPM (v0.1.4) is out of date with the current state of the package in the repo.

Notably, since the last released version, there have been changes to reflect API changes to @solana/codecs.

Would it be possible to publish a new npm package? I also noticed the package.json still says v0.1.3, happy to help get the process going with a PR, if that'd help.

buffalojoec:
#### Problem
As pointed out in #7202, `@solana/spl-token-metadata` depends on a newer version of `@solana/codecs`, which some developers would like to use. Additionally, `@solana/spl-token-group` also depends on this new codecs API, so it makes sense to bump them both together!

We can wait to publish a new patch for `@solana/spl-token` until that's been requested as well, if so desired.

#### Summary of Changes
Bump the patch versions of both `@solana/spl-token-metadata` and `@solana/spl-token-group`.

At one point some of the JS packages fell out of sync with their published versions, so token-metadata skips one here. We rectified a few of these with https://github.com/solana-labs/solana-program-library/pull/7000, but looks like metadata didn't get a release during that one.

Closes #7202

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -698,11 +698,11 @@ importers:
         specifier: ^0.2.0
         version: 0.2.0
       '@solana/spl-token-group':
-        specifier: ^0.0.5
+        specifier: ^0.0.6
         version: link:../../token-group/js
       '@solana/spl-token-metadata':
-        specifier: ^0.1.3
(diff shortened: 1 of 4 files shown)


joncinque (approved):
Thanks!

PR #7216, opened 2024-08-28
CI: Install build deps before publish

joncinque:
#### Problem

The publish step can fail because the protobuf compiler is missing: https://github.com/solana-labs/solana-program-library/actions/runs/10596818547/job/29366085606

#### Summary of changes

Install build deps during publish step


joncinque:
Merging early since the change has no impact on CI

PR #7220, opened 2024-08-30
Increase dev mode opt-level for curve25519-dalek crate

samkim-crypto:
#### Problem
The curve25519-dalek crate could be quite slow on development mode relative to release mode. This could result in some tests failing (e.g. https://github.com/anza-xyz/agave/pull/2252).

#### Summary of Changes
Bump up the opt-level for the crate in Cargo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,13 @@
 [profile.dev]
 split-debuginfo = "unpacked"
 
+# The curve25519-dalek crate uses the `simd` backend by default in v4 if
+# possible, which has very slow performance on some platforms with opt-level 0,
+# which is the default for `dev` and `test` builds. This slowdown causes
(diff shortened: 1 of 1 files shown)


joncinque (commented):
Looks good to me! Can you just add a comment about why this is included? Maybe something like:

```
curve25519-dalek uses the `simd` backend by default in v4 if possible, which has very slow performance on some platforms with opt-level 0, which is the default for `dev` and `test` builds. This slowdown causes certain interactions in the solana-test-validator, such as verifying ZK proofs in transactions, to take much more than 400ms, creating problems in the testing environment.
To give better performance in solana-test-validator during tests and dev builds, we override the opt-level to 3 for the crate.
```

joncinque (approved):

PR #7233, opened 2024-09-03
Publish type-length-value v0.6.0

joncinque:
#### Problem

There was an issue in the publish for TLV 0.6.0, which means that the crate was published, but the version bump didn't go through: https://github.com/solana-labs/solana-program-library/actions/runs/10595201490/job/29360836916

#### Summary of changes

Do the version bump.


joncinque:
Failure being addressed in #7234, merging on red

PR #7234, opened 2024-09-03
ci: Remove all references to twoxtx

joncinque:
#### Problem

Now that the curve25519-dalek upgrade has landed on agave with https://github.com/anza-xyz/agave/pull/2252, the twoxtx build is failing https://github.com/solana-labs/solana-program-library/actions/runs/10682051895/job/29607007216?pr=7233, which makes sense since SPL isn't using the newest curve25519-dalek crate yet.

#### Summary of changes

As we discovered last year, all of the token-2022 tests are going through banks-client / banks-server, which doesn't take into account size limits for transactions. This means that we don't need to do a special 2xTx build at all.

Remove all references to twoxtx.


joncinque:
Merging to unblock CI

PR #7235, opened 2024-09-03
ci: Add git tag / commit message with cargo release

joncinque:
#### Problem

The publish workflow is great, but it does some work by hand, namely creating the git commit and tag. `cargo release` can do this, but we currently don't use it because we aren't customizing it.

#### Summary of changes

Add configuration options to the top-level Cargo.toml and let cargo release do everything for us.

Side note: this *does not* resolve the issue of publishing a crate and then failing to push the git commit. `cargo release` is just doing the steps in series, so if `git push` fails after `cargo publish`, you're in the same situation as now. This just makes the scripts and workflow a bit simpler

### DIFF
--- a/.github/workflows/publish-rust.yml
+++ b/.github/workflows/publish-rust.yml
@@ -205,10 +205,6 @@ jobs:
 
           ./ci/publish-rust.sh "${{ inputs.package_path }}" $LEVEL $OPTIONS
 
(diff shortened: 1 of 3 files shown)


lorisleiva (approved):
Oh nice! It's a shame it doesn't do the whole thing atomically but it's clearly an improvement regardless.

PR #7260, opened 2024-09-12
Revert ESLint plugin bump

### LINKED ISSUE #7259: CI: ESLint upgrade breaks linting on Node.js 16.x
#### Problem

Our JS test jobs for pull requests are using Node.js 20.5, but the workflows on `master` use 16.x.

This difference in Node.js versions breaks linting after the latest ESLint upgrade (#7257, #7258).

#### Proposed Solution

If we want PRs to test JS against Node.js 20.5 but keep `master` on 16.x, then we revert #7257 and #7258, since the newer ESLint uses `structuredClone()`, which is [only available > Node,js 17.x](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone#browser_compatibility).

Otherwise, we can increment the minimum supported Node version, but I don't think this makes sense to do for just linting. The libraries are still compatible with 16.x.

buffalojoec:
#### Problem

The latest ESLint plugin bump from dependabot broke CI for JavaScript jobs.

#### Summary of Changes

Revert it.

Closes #7259

### DIFF
--- a/account-compression/sdk/package.json
+++ b/account-compression/sdk/package.json
@@ -69,7 +69,7 @@
     "@types/jest": "^29.5.12",
     "@types/node": "^22.5.4",
     "@types/node-fetch": "^2.6.11",
-    "@typescript-eslint/eslint-plugin": "^8.5.0",
+    "@typescript-eslint/eslint-plugin": "^8.4.0",
     "@typescript-eslint/parser": "^8.4.0",
     "eslint": "^8.57.0",
     "eslint-config-turbo": "^2.1.1",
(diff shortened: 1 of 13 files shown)


joncinque (approved):
Thanks for fixing this up! I didn't realize we were using different versions between the overall JS job and the per-program JS job

buffalojoec:
> Thanks for fixing this up! I didn't realize we were using different versions between the overall JS job and the per-program JS job

Do you want to be using different versions? I think it makes sense to actually matrix 16 and up for at least the JS PRs maybe? This way, any time a JS lib is changed (including dependabot) we get the whole compatibility run-through.

joncinque:
Yeah sure, it couldn't hurt to do 16, 18, and 20

PR #7266, opened 2024-09-12
CI: add matrix for node versions to JS PRs

buffalojoec:
#### Problem
As outlined in #7259, we aren't testing JavaScript PRs against all versions of Node.js that the packages are documented to support. 

#### Summary of Changes
Add relevant Node.js versions to the `pull-request-js.yml` matrix.


joncinque (approved):
Looks great, thanks!

PR #7271, opened 2024-09-13
Fix Transfer amount for 1 token in token-2022.md

0xAsten:


docs/src/token-2022.md
```diff
@@ -43,7 +43,7 @@ Token-2022 are a strict superset of Token.
 ### Instructions
 
 Token-2022 supports the exact same instruction layouts as Token, byte for
-byte. For example, if you want to transfer 100 tokens on a mint with 2 decimals,
+byte. For example, if you want to transfer 1 tokens on a mint with 2 decimals,
 you create a `TransferChecked` instruction, with this byte-represented data:
```
joncinque, line 47: I see what you mean here... since this can still be ambiguous, how about full clarity and changing the number?
```suggestion
byte. For example, if you want to transfer 0.75 tokens in UI amount, on a mint with 2 decimals,
then the transfer amount is 75 tokens. You create a `TransferChecked` instruction, with
this byte-represented data:
```

0xAsten:
> Thanks for the suggestion! Since I find the new wording ambiguous too, how about we do both?

LGTM

joncinque (approved):
Thanks for the contribution!

PR #7284, opened 2024-09-18
Update a typo in presentation.md

AbhisekBasu1:
What needs to change - 

Are you aware that it's 2023?

Reason - 

It's not 2023 anymore. 

Fix - 

Changed 2023 to 2024. 

PS. This was very fun to read. Good job!

### DIFF
--- a/docs/src/token-2022/presentation.md
+++ b/docs/src/token-2022/presentation.md
@@ -25,7 +25,7 @@ title: Presentation
 
 ---
 
-### Are you aware that it's 2023?
+### Are you aware that it's 2024?
 
 Yes.
 
--- a/docs/src/token-2022/presentation.md
+++ b/docs/src/token-2022/presentation.md
@@ -25,7 +25,7 @@ title: Presentation
 
 ---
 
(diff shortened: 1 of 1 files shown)


docs/src/token-2022/presentation.md
```diff
@@ -25,7 +25,7 @@ title: Presentation
 
 ---
 
-### Are you aware that it's 2023?
+### Are you aware that it's 2024?
```
joncinque, line 28: Since it'll soon be 2025, how about we future-proof this :smile: 
```suggestion
### Are you aware that it's not 2022?
```

joncinque (commented):
Haha I'm glad you enjoyed it! Thanks for pointing this out -- let me know how my suggestion looks.

docs/src/token-2022/presentation.md
AbhisekBasu1: Great idea, should've thought of this earlier 😅

joncinque (approved):
Thanks for your contribution!

PR #7324, opened 2024-10-03
[@solana/spl-stake-pool] export the StakePoolLayout

tomjohn1028:
Necessary to load individual StakePools by their pubkey (possibly across different stake pool program deployments) and decode.

### DIFF
--- a/stake-pool/js/src/index.ts
+++ b/stake-pool/js/src/index.ts
@@ -46,7 +46,7 @@ import BN from 'bn.js';
 export type { StakePool, AccountType, ValidatorList, ValidatorStakeInfo } from './layouts';
 export { STAKE_POOL_PROGRAM_ID } from './constants';
 export * from './instructions';
-export { StakePoolLayout } from './layouts';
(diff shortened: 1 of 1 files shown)


stake-pool/js/src/index.ts
```diff
@@ -46,6 +46,7 @@ import BN from 'bn.js';
 export type { StakePool, AccountType, ValidatorList, ValidatorStakeInfo } from './layouts';
 export { STAKE_POOL_PROGRAM_ID } from './constants';
 export * from './instructions';
+export { StakePoolLayout } from './layouts';
```
buffalojoec, line 49: If we're going to make this one public, might as well do `ValidatorStakeInfoLayout` and `ValidatorListLayout` as well.

buffalojoec (commented):
Wfm. Any opposition to making the layouts public here @joncinque ?

joncinque (approved):
Looks good to me (with Joe's question)!

joncinque (approved):
Thanks for the contribution!

joncinque:
Merging early since the stake pool tests passed

PR #7341, opened 2024-10-08
[token-2022] Add elgamal registry account

samkim-crypto:
#### Problem
Currently, it is not possible to use associated token accounts for confidential transfers because in order to initialize a confidential transfer account, the owner must initialize it with an ElGamal public key.

It would be nice to have a special program "registry" program where a user can store an ElGamal public key that it can use for any mint with confidential transfers enabled.

#### Summary of Changes
- I created a simple ElGamal registry program where a user can store its ElGamal public key in a PDA account that is derived from the user's wallet address - [ba60c3d](https://github.com/solana-labs/solana-program-library/pull/7341/commits/ba60c3d2dc3b3365fe3e9819dce67e75d4bb4676). The program has two instructions. The `CreateRegistry` instruction derives the PDA, invokes the system program to create the account, and then initializes the registry account. The `UpdateRegistry` instruction checks the owner signature and updates the old ElGamal key with a new key. I believe the `UpdateRegistry` instruction does not need to check the PDA address, but please correct me if I am overlooking something 🙏 .
  - In the process, I refactored the proof extraction logic that lived in `program-2022` to `confidential-transfer-proof-extraction` module since this will be shared by both program-2022 and ElGamal registry program - [eef0666](https://github.com/solana-labs/solana-program-library/pull/7341/commits/eef066656694bb866b161625281506f50d970e21)
- I added a new `ConfigureAccountWithRegistry` instruction in the confidential transfer extension which is identical to the `ConfigureAccount` instruction, but skips the signature and zk proof verification - [3104459](https://github.com/solana-labs/solana-program-library/pull/7341/commits/310445934de5dbad3f3cea262e16403b08c11201). I believe the token program itself just need to check the provided registry account is owned by the registry program and does not need to necessarily check the validity of the PDA address, but please correct me if I am overlooking something 🙏 . Another thing to note is the fact that on `ConfigureAccountWithRegistry` instruction, the program initializes the confidential transfer account with default maximum pending credits counter and also `AeCiphertext::default()` (all zero ciphertext). Unlike ElGamal ciphertexts, the all-zero symmetric ciphertext is not really valid, but I added this logic to simplify the program at the cost of increasing the complexity slightly on the client side (client can just interpret all zero ciphertext as encrypting 0). An alternative would be to add a valid zero-ciphertext field to the ElGamal registry account, but I thought this would make things a bit more complicated.
- Added support for `ConfigureAccountWithRegistry` instruction in the token-client - [1a8dc29](https://github.com/solana-labs/solana-program-library/pull/7341/commits/1a8dc290ebc172d94d8ce05f452216a16e3a1a01)
- Added tests to `program-2022-test` - https://github.com/solana-labs/solana-program-library/pull/7341/commits/48624649a2a574726ad46e977283ca963207db55


token/confidential-transfer/elgamal-registry/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "spl-elgamal-registry"
+version = "0.1.0"
+description = "Solana ElGamal Registry"
```
joncinque, line 4: nit: can we spell out this is a program?
```suggestion
description = "Solana ElGamal Registry Program"
```

token/confidential-transfer/elgamal-registry/src/instruction.rs
```diff
@@ -0,0 +1,196 @@
...
+    },
+    spl_token_confidential_transfer_proof_extraction::{ProofData, ProofLocation},
+};
+
+#[derive(Clone, Debug, PartialEq)]
+#[repr(u8)]
+pub enum RegistryInstruction {
+    /// Initialize an ElGamal public key registry.
+    ///
+    /// 0. `[writeable, signer]` The funding account (must be a system account)
```
joncinque, line 20: micro-nit typo
```suggestion
    /// 0. `[writable, signer]` The funding account (must be a system account)
```

Also, we've been trying to avoid having payer accounts passed to the program, and instead assume that there's enough lamports to cover rent-exemption already. This way, the program just needs to allocate + assign.

token/confidential-transfer/elgamal-registry/src/instruction.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// 1. `[writable]` The account to be created
+    /// 2. `[]` The wallet address (will also be the owner address for the
```
joncinque, line 22: Shouldn't this be a signer? Otherwise it's possible to create someone else's registry entry.

Side note: since this will require a signature from the owner, it means that SPL Token multisigs are not supported. This might be totally fine though

token/confidential-transfer/elgamal-registry/src/instruction.rs
```diff
@@ -0,0 +1,196 @@
...
+    };
+    append_zk_elgamal_proof(registry_instruction, proof_location)
+}
+
+/// Create a `RegistryInstruction::UpdateRegistry` instruction
+pub fn update_registry(
+    registry_account: &Pubkey,
+    owner: &Pubkey,
+    proof_location: ProofLocation<PubkeyValidityProofData>,
+) -> Result<Vec<Instruction>, ProgramError> {
```
joncinque, line 134: nit: It's up to you, but there's an inconsistency between the two instruction creators. The first derives the address for you, but this one doesn't. I would say it's probably easier to have them all derive the address for you, so this doesn't need to take the `registry_account`

token/confidential-transfer/elgamal-registry/src/instruction.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+}
+
+/// Takes a `RegistryInstruction` and appends the pubkey validity proof
+/// instruction
+fn append_zk_elgamal_proof(
+    registry_instruction: Instruction,
+    proof_data_location: ProofLocation<PubkeyValidityProofData>,
+) -> Result<Vec<Instruction>, ProgramError> {
```
joncinque, line 176: nit: the function doesn't use the registry instruction, so how about this just takes a `&mut Vec<Instruction>` and optionally appends the proof instruction?

token/confidential-transfer/elgamal-registry/src/lib.rs
```diff
@@ -0,0 +1,28 @@
+mod entrypoint;
+pub mod instruction;
+pub mod processor;
+pub mod state;
+
+use solana_program::pubkey::Pubkey;
+
+/// Seed for the ElGamal registry program-derived address
+pub const REGISTRY_ADDRESS_SEED: &[u8] = "elgamal-registry".as_bytes();
```
joncinque, line 9: nit: you can use the `b` prefix to avoid `as_bytes()`
```suggestion
pub const REGISTRY_ADDRESS_SEED: &[u8] = b"elgamal-registry";
```

token/confidential-transfer/elgamal-registry/src/processor.rs
```diff
@@ -0,0 +1,191 @@
...
+    space: usize,
+    owner: &Pubkey,
+    system_program: &AccountInfo<'a>,
+    new_pda_account: &AccountInfo<'a>,
+    new_pda_signer_seeds: &[&[u8]],
+) -> ProgramResult {
+    if new_pda_account.lamports() > 0 {
+        let required_lamports = rent
+            .minimum_balance(space)
+            .max(1)
```
joncinque, line 149: Why is the `max(1)` needed?

token/confidential-transfer/elgamal-registry/src/processor.rs
```diff
@@ -0,0 +1,191 @@
...
+            &system_instruction::assign(new_pda_account.key, owner),
+            &[new_pda_account.clone(), system_program.clone()],
+            &[new_pda_signer_seeds],
+        )
+    } else {
+        invoke_signed(
+            &system_instruction::create_account(
+                payer.key,
+                new_pda_account.key,
+                rent.minimum_balance(space).max(1),
```
joncinque, line 179: Same here, why the `max(1)`?

token/program-2022/src/lib.rs
```diff
@@ -130,3 +129,13 @@ pub fn check_system_program_account(system_program_id: &Pubkey) -> ProgramResult
     }
     Ok(())
 }
+
+/// Checks if the supplied program ID is that of the ElGamal registry program
+pub fn check_elgamal_registry_program_account(
```
joncinque, line 134: nit: can this be `pub(crate)`?

token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -472,6 +472,29 @@ pub enum ConfidentialTransferInstruction {
...
+    ///
+    /// An `ElGamalRegistry` account is valid if it shares the same owner with
+    /// the token account. If a valid `ElGamalRegistry` account is provided,
+    /// then the program skips the verification of the ElGamal pubkey
+    /// validity proof as well as the token owner signature.
+    ///
+    /// Accounts expected by this instruction:
+    ///
+    ///   * Single owner/delegate
+    ///   0. `[writeable]` The SPL Token account.
```
joncinque, line 490: micro-nit typo
```suggestion
    ///   0. `[writable]` The SPL Token account.
```

token/confidential-transfer/elgamal-registry/src/processor.rs
```diff
@@ -0,0 +1,191 @@
...
+/// Processes `CreateRegistry` instruction
+pub fn process_create_registry_account(
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    proof_instruction_offset: i64,
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+    let funding_account_info = next_account_info(account_info_iter)?;
+    let elgamal_registry_account_info = next_account_info(account_info_iter)?;
+    let wallet_account_info = next_account_info(account_info_iter)?;
```
joncinque, line 35: I don't feel too strongly about it, but we could potentially add a check that this is a system account to avoid creating registry accounts for token accounts. On the flipside, maybe people learned their lesson when deriving ATAs :sweat_smile:

token/confidential-transfer/proof-extraction/src/lib.rs
joncinque, line 1: nit: can you refactor the logic here into a separate file like `location.rs` or `data.rs` or `instruction.rs`?

token/program-2022-test/tests/confidential_transfer.rs
```diff
@@ -2810,3 +2812,122 @@ async fn confidential_transfer_transfer_with_fee_and_memo_option(
...
+    let alice_account_keypair = Keypair::new();
+    let alice_token_account = alice_account_keypair.pubkey();
+    token
+        .create_auxiliary_token_account_with_extension_space(
+            &alice_account_keypair,
+            &alice_token_account,
+            vec![ExtensionType::ConfidentialTransferAccount],
+        )
+        .await
+        .unwrap();
```
joncinque, line 2912: Oh gosh, this makes me realize something we haven't thought about yet. Should we make `reallocate` on token-2022 permissionless? Otherwise, it's not possible to configure someone else's account.

If we want to go with the safer change, we can have `configure_with_registry` optionally reallocate the account.

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -121,13 +162,25 @@ fn process_configure_account(
...
+                account_info_iter.as_slice(),
+            )?;
+        }
+        ElGamalPubkeySource::ElGamalRegistry(elgamal_registry_account) => {
+            // if ElGamal registry was provided, then just verify that the owners of the
+            // registry and token accounts match, then skip the signature
+            // verification check
+            if elgamal_registry_account.owner != *authority_info.key {
+                return Err(TokenError::OwnerMismatch.into());
+            }
```
joncinque, line 181: You'll also need to check that `authority_info.key == token_account.base.owner`, otherwise you can provide someone else's registry!

joncinque (commented):
Looks pretty good overall! Mostly small things, let me know what you think

token/confidential-transfer/proof-extraction/Cargo.toml
```diff
@@ -10,7 +10,9 @@ edition = "2021"
 [dependencies]
 bytemuck = "1.19.0"
 solana-curve25519 = "2.0.3"
+solana-program = "2.0.3"
 solana-zk-sdk = "2.0.3"
+spl-pod = "0.4.0"
```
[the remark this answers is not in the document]
samkim-crypto: Oops thanks for the catch!

token/confidential-transfer/elgamal-registry/src/instruction.rs
samkim-crypto: Oh you are completely right. For some reason I thought the owner can always update the registry account with its own key, but obviously the owner may not want to have want a registry. Fixed!

Yeah I thought about the Token multisigs when I was implementing this, but I think this should be fine.

token/confidential-transfer/elgamal-registry/src/instruction.rs
samkim-crypto: Yep, good idea!

token/confidential-transfer/elgamal-registry/src/instruction.rs
samkim-crypto: Good idea! Maybe this function can be refactored into the `proof-generation` crate and shared by both registry and token22 instruction constructors, but that is probably outside the scope of this PR.

token/confidential-transfer/elgamal-registry/src/processor.rs
samkim-crypto: Oh oops yeah I was going to ask this in the PR description, but I completely forgot. I got this function from the [ATA program](https://github.com/solana-labs/solana-program-library/blob/master/associated-token-account/program/src/tools/account.rs#L31) and it has this max value, which I wasn't sure what it was achieving.

token/confidential-transfer/proof-extraction/src/lib.rs
samkim-crypto: Yep, good idea!

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: Oh yeah I think this should actually be `elgamal_registry_account.owner != token_account.base.owner`. The authority pubkey does not necessarily need to be provided in the list of accounts since it is not providing any more info and is not a signer. I removed the authority pubkey from the instruction, so let me know what you think.

token/program-2022-test/tests/confidential_transfer.rs
samkim-crypto: Oh yeah completely missed that. It seems that if all accounts are to be rent-exempt, then there seems to be not much harm in making it permissionless? I'll think about it today and let's discuss tomorrow.

rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.78.0"
+channel = "1.79.0"
```
joncinque, line 2: Is this bumped intentionally?

token/program-2022/src/extension/confidential_transfer/processor.rs
joncinque: Looks great!

token/program-2022-test/tests/confidential_transfer.rs
joncinque: There is one slightly strange attack -- if we're competing in a trading environment, I could reallocate your token account to be as big as possible, which would count against the loaded data costs of your transaction, and try to make your transaction fail.

I'd probably go with special-casing this behavior and only reallocating for the confidential token account extension to be totally safe, but maybe I'm being overly cautious.

token/confidential-transfer/elgamal-registry/src/processor.rs
joncinque: Ah, I wonder if that's there while testing with 0 rent. I would go with removing the logic here, since I hope our testing tools are better now and never provide 0 rent.

token/confidential-transfer/elgamal-registry/src/processor.rs
joncinque: Actually, sorry, I thought about it some more, and I don't think we should have the system program check. This won't be like creating ATAs for other users, since the user needs to do some work to upload their elgamal key. Let's keep it more general

token/confidential-transfer/elgamal-registry/src/instruction.rs
```diff
@@ -0,0 +1,197 @@
...
+    },
+    spl_token_confidential_transfer_proof_extraction::instruction::{ProofData, ProofLocation},
+};
+
+#[derive(Clone, Debug, PartialEq)]
+#[repr(u8)]
+pub enum RegistryInstruction {
+    /// Initialize an ElGamal public key registry.
+    ///
+    /// 0. `[writable, signer]` The funding account (must be a system account)
```
joncinque, line 20: I think the other part of my comment got lost under the typo suggestion -- could we remove this and have the client function automatically transfer the rent-exempt lamports into the account so that the user does not need to pass a mutable signer to the program?

rust-toolchain.toml
samkim-crypto: Oops, that accidentally got in. I will revert.

token/program-2022-test/tests/confidential_transfer.rs
samkim-crypto: Oh yes I forgot that there was a `MAX_LOADED_ACCOUNTS_DATA_SIZE_BYTES` field. Yeah in this case, it seems safest to just make reallocation be permissionless for confidential extensions.

token/confidential-transfer/elgamal-registry/src/instruction.rs
samkim-crypto: Okay, I updated the program so that the registry program just allocates and assigns ownership of the PDA account. Please take a look!

token/program-2022-test/tests/confidential_transfer.rs
samkim-crypto: Okay, I added the option to reallocate when configuring confidential transfer accounts with registry.

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -92,26 +98,127 @@ fn process_update_mint(
...
+}
+
+fn reallocate_for_configure_account_with_registry<'a>(
+    token_account_info: &AccountInfo<'a>,
+    payer_info: &AccountInfo<'a>,
+    system_program_info: &AccountInfo<'a>,
+) -> ProgramResult {
+    let mut current_extension_types = {
+        let token_account = token_account_info.data.borrow();
+        let account = StateWithExtensions::<Account>::unpack(&token_account)?;
```
joncinque, line 153: nit: can you use `PodStateWithExtensions::<PodAccount>` instead for better CU usage?

token/program-2022/src/extension/confidential_transfer/processor.rs
```diff
@@ -92,26 +98,127 @@ fn process_update_mint(
...
+
+    if reallocate {
+        let payer_info = next_account_info(account_info_iter)?;
+        let system_program_info = next_account_info(account_info_iter)?;
+        reallocate_for_configure_account_with_registry(
+            token_account_info,
+            payer_info,
+            system_program_info,
+        )?;
+    }
```
joncinque, line 127: We can maybe make this processing logic more clever. Rather than always trying pull out the payer / system program info and then deciding to reallocate, the instruction can *always* try to reallocate, and only if extra lamports are needed, try to extract the payer and system program. What do you think?

joncinque (commented):
Looking really close! Sorry, I did one last pass over the whole thing and found a couple more bits

token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -484,15 +484,22 @@ pub enum ConfidentialTransferInstruction {
...
+    ///
     /// Accounts expected by this instruction:
     ///
     ///   * Single owner/delegate
     ///   0. `[writable]` The SPL Token account.
     ///   1. `[]` The corresponding SPL Token mint.
     ///   2. `[]` The ElGamal registry account.
+    ///   3. `[signer, writable]` (Optional) The payer account to fund
+    ///      reallocation
+    ///   4. `[]` (Optional) System program for reallocation funding
```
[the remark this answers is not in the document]
samkim-crypto: Yep, I did put that in right after you told me that is a bad a pattern 😅 , but I did it after I saw that we do that same thing with the `Reallocate` instruction.

token/program-2022/src/extension/confidential_transfer/instruction.rs
```diff
@@ -669,6 +676,17 @@ pub struct TransferWithFeeInstructionData {
...
+/// Data expected by
+/// `ConfidentialTransferInstruction::ConfigureAccountWithRegistry`
+#[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
+#[derive(Clone, Copy, Debug, PartialEq, Pod, Zeroable)]
+#[repr(C)]
+pub struct ConfigureAccountWithRegistryInstructionData {
+    /// Reallocate token account if it is not large enough for the
+    /// `ConfidentialTransfer` extension.
+    pub reallocate_account: PodBool,
+}
```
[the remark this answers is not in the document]
samkim-crypto: Yeah that is a good idea!

token/program-2022/src/extension/confidential_transfer/processor.rs
samkim-crypto: I thought about this as well since that logic will be cleaner. The reason why I went with the current implementation is that if the user does not want to reallocate the account, then we would be consuming unnecessary compute units by unpacking the account and calculating the needed account length. What do you think?

token/confidential-transfer/elgamal-registry/src/processor.rs
```diff
@@ -0,0 +1,166 @@
...
+    let (elgamal_registry_account_address, bump_seed) =
+        get_elgamal_registry_address_and_bump_seed(wallet_account_info.key, program_id);
+    if elgamal_registry_account_address != *elgamal_registry_account_info.key {
+        msg!("Error: ElGamal registry account address does not match seed derivation");
+        return Err(ProgramError::InvalidSeeds);
+    }
+
+    let elgamal_registry_account_seeds: &[&[_]] = &[
+        REGISTRY_ADDRESS_SEED,
+        &wallet_account_info.key.to_bytes(),
```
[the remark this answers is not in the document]
samkim-crypto: Good point!

token/program-2022/src/extension/confidential_transfer/processor.rs
joncinque: Hm, that's a good point on the CUs... could you do a little test to see how many CUs the extra logic takes up? I'm hoping it's not too many, but there is a vector allocation, so who knows

PR #7342, opened 2024-10-08
stake-pool-js: Bump to release

joncinque:
#### Problem

#7324 exposed some needed types, but it's not released yet.

#### Summary of changes

Bump the patch version to release a new version.


joncinque:
Merging early since stake pool tests passed

PR #7356, opened 2024-10-16
Set patch version to only use solana-program <2

ngundotra:
Patching this to unblock MPLX development


danenbm (commented):
Thank you! LGTM

danenbm (commented):
Actually I think you want to update `version = "0.1.0"` to `version = "0.1.1"` in merkle tree reference?

ngundotra:
@danenbm good catch, thank you!

PR #7360, opened 2024-10-16
Hardcode the discriminators so that you don't have to compute them at runtime

### LINKED ISSUE #7351: @solana/spl-token library doesn't work in Vite project
When I install `"@solana/spl-token": "^0.4.8"` in vite project and run `vite build`, I see this error.

../../node_modules/@solana/spl-type-length-value/lib/esm/splDiscriminate.js (1:9): "createHash" is not exported by "__vite-browser-external", imported by "../../node_modules/@solana/spl-type-length-value/lib/esm/splDiscriminate.js".

1: import { createHash } from 'crypto';
            ^
2: export const splDiscriminate = (discriminator, length = 8) => {
3:     const digest = createHash('sha256').update(discriminator).digest();

It's because Vite doesn't support built-in node.js module like `crypto`.
https://www.npmjs.com/package/crypto

Can you update your package to include the polyfills as default?

Thanks

steveluscher:
1. There is no need to compute these well-known values at runtime
2. Adding insult to injury, computing them requires a SHA-256 _polyfill_ for browsers

Resolves #7351.

### DIFF
--- a/token-group/js/package.json
+++ b/token-group/js/package.json
@@ -46,10 +46,10 @@
         "@solana/web3.js": "^1.95.3"
     },
     "dependencies": {
-        "@solana/codecs": "2.0.0-rc.1",
-        "@solana/spl-type-length-value": "0.1.0"
+        "@solana/codecs": "2.0.0-rc.1"
     },
     "devDependencies": {
+        "@solana/spl-type-length-value": "0.2.0",
         "@solana/web3.js": "^1.95.3",
         "@types/chai": "^5.0.0",
         "@types/mocha": "^10.0.9",
--- a/token-group/js/package.json
+++ b/token-group/js/package.json
@@ -1,7 +1,7 @@
 {
     "name": "@solana/spl-token-group",
     "description": "SPL Token Group Interface JS API",
-    "version": "0.0.6",
+    "version": "0.0.7",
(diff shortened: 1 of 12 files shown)


token-group/js/package.json
```diff
@@ -43,11 +43,11 @@
         "watch": "tsc --build --verbose --watch tsconfig.all.json"
     },
     "peerDependencies": {
+        "@solana/spl-type-length-value": "0.2.0",
```
steveluscher, line 46: This PR won't pass until this is published.

token-group/js/package.json
```diff
@@ -46,10 +46,10 @@
         "@solana/web3.js": "^1.95.3"
     },
     "dependencies": {
-        "@solana/codecs": "2.0.0-rc.1",
-        "@solana/spl-type-length-value": "0.1.0"
+        "@solana/codecs": "2.0.0-rc.1"
     },
     "devDependencies": {
+        "@solana/spl-type-length-value": "0.2.0",
```
steveluscher, line 52: This PR won't pass until this is published.

joncinque (commented):
The changes look good here! Nice job reusing the existing tests to check that the hard-coded discriminants are correct

joncinque (dismissed):
Thanks for fixing all this!

joncinque (approved):

PR #7363, opened 2024-10-17
[account compression] Fix for closing trees on mainnet

### LINKED ISSUE #7357: [Bug] Access violation in stack frame
I am trying to close an empty tree (7MjVK2Y3MCd5ZJme7npcVPJkgtb6pZj2ztkWQXjeSECR) through CPI and keep get this stack frame violation issue

image

ngundotra:
* add security_txt
* format the repo better so it can actually support `solana-verify` building
* Fix #7357


sultandilaram:
@ngundotra 
Thanks for quick support, I was doing same thing by moving the macro call to an inline(never) function but it was already done in master branch

image

PR #7378, opened 2024-10-22
Add retry logic in getOrCreateAssociatedTokenAccount function

dimitrisniras:
Adding retry logic for the second `getAccount` call in `getOrCreateAssociatedTokenAccount` function, since sometimes is failing with a `TokenAccountNotFoundError` error.


token/js/package.json
```diff
@@ -33,6 +33,7 @@
         "clean": "shx rm -rf lib **/*.tsbuildinfo || true",
         "build": "tsc --build --verbose tsconfig.all.json",
         "postbuild": "shx echo '{ \"type\": \"commonjs\" }' > lib/cjs/package.json",
+        "prepare": "npm run build",
```
buffalojoec, line 36: Why add this?

buffalojoec (commented):
Hey @dimitrisniras thanks for the contribution. Are you sure you're not having this issue because of the `confirmOptions` being too laxed? Is there any code you have to reproduce the issue?

PR #7397, opened 2024-10-28
single-pool, stake-pool: mpl-token-metadata supports token-2022

aoikurokawa:
Since mpl-token-metadata program supports token-2022, single-pool and stake-pool would use this program.


joncinque:
Thanks for your contribution! What's the motivation for this change?

Looking at the implementation of the new create_v1 instruction at: https://github.com/metaplex-foundation/mpl-token-metadata/blob/7c95f9948d7fed8db9947188b8040f0d560ac2d3/programs/token-metadata/program/src/processor/metadata/create.rs#L34

And the old one at: https://github.com/metaplex-foundation/mpl-token-metadata/blob/7c95f9948d7fed8db9947188b8040f0d560ac2d3/programs/token-metadata/program/src/processor/metadata/create_medatata_accounts_v3.rs#L12

I'm not immediately seeing a benefit. Stake pool and single pool mints are both for fungible tokens, and the mint is already created by the time these instructions are called, so the extra logic in `create_v1` doesn't seem to bring anything extra.

On top of that, the change to use the new metadata instruction is a breaking change to the single-pool and stake-pool program interfaces, so we'd have to create new instructions to make the change safely.

aoikurokawa:
Since I saw these comments in the test code and because the `mpl_token_metadata` program supports token-2022 now, so I made this change. 

https://github.com/solana-labs/solana-program-library/blob/da493575da94e6a53c3fb128c9dec87b193aa75f/stake-pool/program/tests/create_pool_token_metadata.rs#L39-L42

I agree that this PR does not add significant new features and will break the interface. Therefore, I will probably close this PR.

joncinque:
Ah I see, in that case, it might be worth updating the binary at https://github.com/solana-labs/solana-program-library/blob/master/stake-pool/program/tests/fixtures/mpl_token_metadata.so with the newest version and then enabling that test case. What do you think?

aoikurokawa:
I have checked mpl_token_metadata code. It seems `create_metadata_accounts_v3` does not support token-2022...
 
https://github.com/metaplex-foundation/mpl-token-metadata/blob/2878f777648deccd6230e45b6f081f5ed742dbef/programs/token-metadata/program/src/processor/mod.rs#L172-L180

When I call the program, the program returns `MetadataError::InstructionNotSupported`

https://github.com/metaplex-foundation/mpl-token-metadata/blob/2878f777648deccd6230e45b6f081f5ed742dbef/programs/token-metadata/program/src/error.rs#L603-L605

I think we have to use new instruction `Create`.

joncinque:
Ah I see, that's a bit annoying... in that case, we'll need to have two additional instructions for metadata. Let's leave the single-pool program alone, since it doesn't use token-2022, meaning that the existing metadata instructions still work.

For the stake-pool program, since it supports token-2022, we would need the new create and update metadata instructions. On the flip side, the existing metadata instructions were only added for stake pools that *didn't* create their metadata before creating the pool, maybe we can solve this problem on the tooling side instead of the program.

In the CLI, during `create-pool`, we can add arguments for specifying the metadata info and metadata update authority. If these arguments are not provided and the stake pool mint uses token-2022, the CLI returns an error saying "please specify the metadata or specify `--allow-unspecified-metadata`". That second flag bypasses the check.

What do you think about the general approach?

aoikurokawa:
Thank you for your suggestion! I think your suggestion seems to work.
Let me try to code it!

PR #7418, opened 2024-10-30
slashing: add duplicate block proof verification

AshwinSekar:
Initial change for duplicate block slashing as part of https://github.com/solana-foundation/solana-improvement-documents/pull/204

Users should create a proof account with data storage of at least 2512 bytes (2 * PACKET_DATA_SIZE + struct overhead) and populate it with the 2 shreds that comprise the proof in the format of `DuplicateBlockProofData`.

The program supports 1 instruction `DuplicateBlockProof` which takes a proof account, an offset to read from, the slot in which a duplicate block was broadcasted, and the pubkey of the offending leader.

After verification and deserialization of the proof account, we check if any of the following conditions are met which invalidate the proof:
- Shreds are not for the slot specified
- Shreds are for differing shred versions
- Shreds are in the legacy shred format

We then verify if the proof is valid. A proof consisting of shred1 and shred2 is valid if any of the following conditions are met:
- shred1 and shred2 are part of the same FEC set but have different merkle roots
- shred1 and shred2 are part of different FEC sets, however their erasure config indicates that the sets overlap
- shred1 and shred2 are for the same index with the same shred type and their payloads differ
- shred1 and shred2 are data shreds with different indices and indicate a last in slot violation. This means one shred has the LAST_SHRED_IN_SLOT flag set, however the other shred has a higher index.
- shred1 and shred2 are part of the same FEC set however they have conflicting erasure configs

At the moment we only log the results of the proof verification. Future work will store these results in a context state account for use by the runtime.

Signature verification is not implemented in this change. A future change will use instruction introspection to verify that both shreds were signed by the `node_pubkey`.


slashing/program/src/duplicate_block_proof.rs
```diff
@@ -0,0 +1,830 @@
...
+        msg!(
+            "Valid merkle root conflict for fec set {}, {:?} vs {:?}",
+            shred1.fec_set_index()?,
+            shred1.merkle_root()?,
+            shred2.merkle_root()?
+        );
+        return Ok(());
+    }
+
+    // Overlapping fec set check
```
AshwinSekar, line 97: Overlapping fec sets are net yet implemented on the detection side however I will work on it soon https://github.com/solana-labs/solana/issues/33037

slashing/program/src/duplicate_block_proof.rs
```diff
@@ -0,0 +1,830 @@
...
+    if shred2.slot()? != slot {
+        msg!(
+            "Invalid proof for different slots {} vs {}",
+            shred1.slot()?,
+            slot,
+        );
+        return Err(SlashingError::SlotMismatch);
+    }
+
+    if shred1.version()? != shred2.version()? {
```
AshwinSekar, line 72: This is slightly different from the gossip verification. We allow for shred versions that are not the most current to slash validators that broadcasted duplicate blocks prior to the most recent cluster restart.

slashing/program/src/duplicate_block_proof.rs
```diff
@@ -0,0 +1,830 @@
...
+        msg!(
+            "Invalid proof for different shred versions {} vs {}",
+            shred1.version()?,
+            shred2.version()?,
+        );
+        return Err(SlashingError::InvalidShredVersion);
+    }
+
```
AshwinSekar, line 80: The other difference from gossip is that we do not verify that the `node_pubkey` was supposed to be the leader for `slot` via the leader schedule.

This is important on the gossip side because without this verification anyone could censor a leader in fork choice by submitting a duplicate shred proof for the leader's slot. However for slashing since signing these shreds will get you slashed I believe it is not necessary to incur the overhead of checking the leader schedule.

slashing/program/src/shred.rs
```diff
@@ -0,0 +1,567 @@
...
+impl Shred {
+    const SIZE_OF_CODING_PAYLOAD: usize = 1228;
+    const SIZE_OF_DATA_PAYLOAD: usize =
+        Self::SIZE_OF_CODING_PAYLOAD - Self::SIZE_OF_CODING_HEADERS + SIZE_OF_SIGNATURE;
+    const SIZE_OF_CODING_HEADERS: usize = 89;
+    const SIZE_OF_DATA_HEADERS: usize = 88;
+
+    pub(crate) fn new_from_payload(payload: Vec<u8>) -> Result<Self, SlashingError> {
+        match Self::get_shred_variant(&payload)? {
+            ShredVariant::LegacyCode | ShredVariant::LegacyData => Err(SlashingError::LegacyShreds),
```
AshwinSekar, line 108: Legacy shreds are dropped on all clusters so no reason to accept them for slashing proofs.

slashing/program/src/duplicate_block_proof.rs
```diff
@@ -0,0 +1,830 @@
...
+///       have the same `shred_type`
+///     - If `shred1` and `shred2` share the same index they must be not have
+///       equal payloads excluding the retransmitter signature
+///     - If `shred1` and `shred2` do not share the same index and are data
+///       shreds verify that they indicate an index conflict. One of them must
+///       be the LAST_SHRED_IN_SLOT, however the other shred must have a higher
+///       index.
+///     - If `shred1` and `shred2` do not share the same index and are coding
+///       shreds verify that they have conflicting erasure metas
+fn check_shreds(slot: Slot, shred1: &Shred, shred2: &Shred) -> Result<(), SlashingError> {
```
AshwinSekar, line 53: if we want to simplify control flow here we could have the instruction attach an enum specifying which case to check.
the detection logic already can tell which case was violated.

slashing/program/Cargo.toml
joncinque, line 1: This is my fault for the slow review, but SPL is now using all of the v2.1 crates -- could you upgrade to those? It should be straightforward.

After this lands, we should be able to change this program using the component SDK crates instead of solana-program :rocket:

joncinque (commented):
Really great work! The base logic looks solid, and the next steps for signature verification and creating an account are very clear.

Most of the comments are pretty minor, to make this more performant.

slashing/program/src/duplicate_block_proof.rs
joncinque: Mm yeah, that all makes sense. It might be worth making that clear in the SIMD.

slashing/program/src/shred.rs
```diff
@@ -0,0 +1,567 @@
...
+            next_shred_index,
+            next_code_index, // next_code_index
+            merkle_variant,
+            &ReedSolomonCache::default(),
+            &mut ProcessShredsStats::default(),
+        )
+    }
+
+    #[test]
+    fn test_solana_shred_parity() {
```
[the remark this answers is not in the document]
joncinque: The monorepo would use it too. I can get a PR together for that split-up if you like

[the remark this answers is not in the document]
AshwinSekar: I'm okay with that approach. @behzadnouri wdyt about a `solana-shred` crate that copies the data structures and `layout` functions for use in `solana_ledger` as well as the slashing program onchain?

We can also make this split after this PR lands.

slashing/program/src/instruction.rs
```diff
@@ -0,0 +1,146 @@
...
+        /// Offset into the proof account to begin reading, expressed as `u64`
+        offset: u64,
+        /// Slot for which the violation occured
+        slot: Slot,
+        /// Identity pubkey of the Node that signed the duplicate block
+        node_pubkey: Pubkey,
+    },
+}
+
+impl SlashingInstruction {
```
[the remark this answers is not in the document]
AshwinSekar: very cool, I added that here d19b90a100d30dbf6129b8a1c73462da9def8fd3

ended up copying over some of the helpers because I figured adding a dependency on the token program is not great?

slashing/program/src/duplicate_block_proof.rs
```diff
@@ -0,0 +1,830 @@
...
+        // node_pubkey.verify_strict(merkle_root, signature).
+        // We will verify that the pubkey merkle root and signature match the shred and
+        // that the verification was successful.
+        let shred1 = Shred::new_from_payload(self.shred1)?;
+        let shred2 = Shred::new_from_payload(self.shred2)?;
+        check_shreds(slot, &shred1, &shred2)
+    }
+}
+
+/// Check that `shred1` and `shred2` indicate a valid duplicate proof
```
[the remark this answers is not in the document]
AshwinSekar: Good catch I forgot. We only allow 1 epoch for statue of limitations, I think that's plenty of time. Addressed in 9b822d6829023aa146c975a29efc4ed94a80bcca

slashing/program/src/processor.rs
```diff
@@ -0,0 +1,68 @@
...
+
+fn verify_proof_data<'a, T>(slot: Slot, pubkey: Pubkey, proof_data: &'a [u8]) -> ProgramResult
+where
+    T: SlashingProofData + Deserialize<'a>,
+{
+    if proof_data.len() < T::PROOF_TYPE.proof_account_length() {
+        return Err(ProgramError::InvalidAccountData);
+    }
+    let proof_data: T =
+        bincode::deserialize(proof_data).map_err(|_| SlashingError::DeserializationError)?;
```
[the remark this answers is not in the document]
AshwinSekar: Got it, moved to zerocopy instead of bincode a3d74ab885e19fbd8910618a7a4e1d208a1787d5

slashing/program/src/duplicate_block_proof.rs
```diff
@@ -0,0 +1,830 @@
...
+    solana_program::{clock::Slot, msg, pubkey::Pubkey},
+};
+
+#[derive(Deserialize, Serialize)]
+pub struct DuplicateBlockProofData {
+    #[serde(with = "serde_bytes")]
+    pub shred1: Vec<u8>,
+    #[serde(with = "serde_bytes")]
+    pub shred2: Vec<u8>,
+}
```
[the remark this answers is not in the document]
AshwinSekar: Rolled my own without avoiding a copy here a3d74ab885e19fbd8910618a7a4e1d208a1787d5
Format is 
* len of shred1
* shred 1 bytes
* len of shred2
* shred 2 bytes
And then custom pack and unpack fns

slashing/program/src/shred.rs
```diff
@@ -0,0 +1,567 @@
...
+}
+
+impl Shred {
+    const SIZE_OF_CODING_PAYLOAD: usize = 1228;
+    const SIZE_OF_DATA_PAYLOAD: usize =
+        Self::SIZE_OF_CODING_PAYLOAD - Self::SIZE_OF_CODING_HEADERS + SIZE_OF_SIGNATURE;
+    const SIZE_OF_CODING_HEADERS: usize = 89;
+    const SIZE_OF_DATA_HEADERS: usize = 88;
+
+    pub(crate) fn new_from_payload(payload: Vec<u8>) -> Result<Self, SlashingError> {
```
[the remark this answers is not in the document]
AshwinSekar: Not quite a Pod since I stored some of the commonly used fields `shred_type, proof_size` and the shred format flags in the struct. Wanted to avoid deserializing them multiple times.

But in the spirit of your suggestion I changed it to accept a pointer so that we can avoid copying the payload. We'll ultimately need to copy in order to compare the fields when checking the proof, but this avoids copying unnecessary bits.

slashing/program/src/processor.rs
```diff
@@ -13,21 +13,28 @@ use {
...
     T: SlashingProofData + Deserialize<'a>,
 {
-    if proof_data.len() < T::PROOF_TYPE.proof_account_length() {
-        return Err(ProgramError::InvalidAccountData);
+    // Statue of limitations is 1 epoch
+    let clock = Clock::get()?;
+    let Some(elapsed) = clock.slot.checked_sub(slot) else {
+        return Err(ProgramError::ArithmeticOverflow);
+    };
+    if elapsed > DEFAULT_SLOTS_PER_EPOCH {
```
joncinque, line 33: Can you use the leader schedule sysvar instead of hard-coding the slot-length? In case we ever want to change this, we'll be screwed. You can just `EpochSchedule::get()`

slashing/program/src/processor.rs
```diff
@@ -66,3 +73,86 @@ pub fn process_instruction(
...
+            clock::{Clock, Slot, DEFAULT_SLOTS_PER_EPOCH},
+            program_error::ProgramError,
+            signature::Keypair,
+            signer::Signer,
+        },
+        std::sync::Arc,
+    };
+
+    const SLOT: Slot = 53084024;
+    static mut CLOCK_SLOT: Slot = SLOT;
```
joncinque, line 96: This is fine for just one test, but in case we need to, we can do proper locking on this variable, like at https://github.com/solana-labs/solana-program-library/blob/4b914915f45054c7fbc3be42c57031a1eaf7cb89/token/program/src/processor.rs#L1060-L1063

slashing/program/src/processor.rs
```diff
@@ -66,3 +73,86 @@ pub fn process_instruction(
...
+
+    fn generate_proof_data(leader: Arc<Keypair>) -> Vec<u8> {
+        let mut rng = rand::thread_rng();
+        let (slot, parent_slot, reference_tick, version) = (SLOT, 53084023, 0, 0);
```
joncinque, line 100: nit: how about setting `parent_slot` as just `SLOT - 1` for clarity?

slashing/program/tests/duplicate_block_proof.rs
```diff
@@ -179,13 +189,14 @@ pub(crate) fn new_rand_shreds<R: Rng>(
...
     let mut context = program_test().start_with_context().await;
+    setup_clock(&mut context).await;
 
     let authority = Keypair::new();
     let account = Keypair::new();
 
     let mut rng = rand::thread_rng();
     let leader = Arc::new(Keypair::new());
-    let (slot, parent_slot, reference_tick, version) = (53084024, 53084023, 0, 0);
+    let (slot, parent_slot, reference_tick, version) = (SLOT, 53084023, 0, 0);
```
joncinque, line 200: nit: same here `SLOT - 1` is a bit easier to read for the parent slot for all the tests in this file

slashing/program/src/duplicate_block_proof.rs
```diff
@@ -1,22 +1,40 @@
...
+/// Proof of a duplicate block violation
+pub struct DuplicateBlockProofData<'a> {
+    /// Shred signed by a leader
+    pub shred1: &'a [u8],
+    /// Conflicting shred signed by the same leader
+    pub shred2: &'a [u8],
+}
+
+impl<'a> DuplicateBlockProofData<'a> {
+    #[allow(dead_code)]
```
joncinque, line 24: nit: is this needed?

slashing/program/src/duplicate_block_proof.rs
```diff
@@ -30,6 +48,32 @@ impl SlashingProofData for DuplicateBlockProofData {
...
+    where
+        Self: Sized,
+    {
+        if data.len() < LENGTH_SIZE {
+            return Err(SlashingError::ProofBufferTooSmall);
+        }
+        let (length1, data) = data.split_at(LENGTH_SIZE);
+        let shred1_length = try_from_bytes::<PodU32>(length1)
+            .map_err(|_| SlashingError::ProofBufferDeserializationError)?;
+        let (shred1, data) = data.split_at(u32::from(*shred1_length) as usize);
```
joncinque, line 71: To avoid a panic, can this return an error if `data.len() < shred1_length`?

slashing/program/src/processor.rs
```diff
@@ -36,7 +35,7 @@ where
     }
 
     let proof_data: T =
-        bincode::deserialize(proof_data).map_err(|_| SlashingError::ShredDeserializationError)?;
+        T::unpack(proof_data).map_err(|_| SlashingError::ShredDeserializationError)?;
```
joncinque, line 38: Beautiful!

slashing/program/src/shred.rs
```diff
@@ -333,20 +332,17 @@ impl Shred {
         {
             return false;
         }
-        fn get_payload(shred: &Shred) -> &[u8] {
+        fn get_payload<'a>(shred: &'a Shred<'a>) -> &'a [u8] {
```
joncinque, line 335: nit: does the shred reference need to have `'a` lifetime? Since the payload is coming from inside the shred, I'm thinking we can have `(shred: &'b Shred<'a>) -> &'a [u8]`. Is that correct?

slashing/program/src/state.rs
```diff
@@ -58,10 +58,15 @@ impl From<u8> for ProofType {
 
 /// Trait that proof accounts must satisfy in order to verify via the slashing
 /// program
-pub trait SlashingProofData {
+pub trait SlashingProofData<'a> {
```
joncinque, line 61: I've gotten burned by lifetimes in traits in the past, since they can make things complicated or totally unworkable without a lot of Rust magic. For example, I don't think you can have a type implement `SlashingProofData` where it copies the underlying data to own it.

I think we'll be ok in this case since it'll just be for unpacking things in a program, but something to be careful about.

joncinque (commented):
This is looking really close! A couple last bits on my side, then we can get this in

Cargo.lock
joncinque, line 1: Oh sorry, one last bit, can you revert these changes so that the PR only picks up the new crate that you're adding? There shouldn't be too many other changes past that

slashing/program/src/state.rs
AshwinSekar: Guess it forces us to zerocopy everything in the future?

slashing/program/src/duplicate_block_proof.rs
AshwinSekar: Nope it's not, removed

Cargo.lock
AshwinSekar: Done 74cdd12c92ad12956161de82ae28696505167c3a

slashing/program/src/shred.rs
AshwinSekar: No it doesn't you're absolutely right 9174be41bec85352c21555fcf671daff13e3c249

slashing/program/src/duplicate_block_proof.rs
AshwinSekar: Yep good call c018720853875cf9ef88675c68f210733738a775

slashing/program/tests/duplicate_block_proof.rs
AshwinSekar: reorganized into test constants here 880d9c0e784174a65d3fc9a8530905455db0de22

slashing/program/src/processor.rs
AshwinSekar: nice ya that's less hacky 5210f9cee99d136f7061f453c7efa171a09a4166

slashing/program/src/processor.rs
AshwinSekar: Good call 09235dac6e62961e673d281fc36063514550116e

joncinque (approved):
Looks really great, thanks for integrating all the feedback!

And sorry, I didn't mean to totally revert Cargo.lock -- can you just run `cargo tree` from the root and commit that? Once it passes CI, all good from my side

joncinque (approved):
Great work!

PR #7419, opened 2024-10-30
Lock to web3.js 1.x for compatibility

steveluscher:
Read why, here: https://github.com/solana-labs/solana-web3.js/issues/3498


joncinque (approved):
Thanks!

PR #7421, opened 2024-10-31
Extract spl-math-utils from spl-math

kevinheavey:
Crates that are using `spl-math` as a library don't need the parts that use `solana_program` .

This PR pulls the math parts into a separate `spl-math-utils` crate and updates `spl-token-swap` to use that instead of `spl-math`. That's the only SPL crate that uses `spl-math`.

This PR also removes `spl-math` from the `spl-stake-pool` deps but `spl-stake-pool` wasn't using `spl-math` in the first place


kevinheavey:
Clippy errors are unrelated to this PR:

```
error: doc list item without indentation
  --> token/program-2022/src/extension/confidential_mint_burn/instruction.rs:80:9
   |
80 |     ///   3.. `[signer]` Required M signer accounts for the SPL Token Multisig
```

libraries/math-utils/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "spl-math-utils"
```
joncinque, line 2: Can we have this crate be `spl-math` and change the one with the program to `spl-math-example`?

joncinque (commented):
Great change, thanks! Since it's more of a mistake that spl-math included a whole a program, can we rename the program bit instead?

libraries/math-utils/Cargo.toml
kevinheavey: Done. Set the spl-math version back to 0.3 and set spl-math-example version to 0.1

joncinque (approved):
Thanks!

PR #7424, opened 2024-10-31
CI: Fix clippy for confidential mint/burn

joncinque:
#### Problem

The PR for the upgrade for v2.1 and confidential mint/burn went in at the same time, but there were some improper doc comments in the confidential mint/burn, which breaks the new clippy version used with v2.1.

#### Summary of changes

Update the doc comments to unbreak master.


joncinque:
Even though the job says it failed, the clippy command succeeded -- there might be an issue with the cache, so just going to merge this on red

PR #7425, opened 2024-10-31
remove solana-program from spl-pod

kevinheavey:
This required replacing `#[spl_program_error]` with inline impls. I think it is hard or perhaps impossible to decouple `spl-program-error` from `solana_program` without breaking changes.

For now there will still be an indirect solana-program dep via solana-zk-sdk

### DIFF
--- a/libraries/pod/Cargo.toml
+++ b/libraries/pod/Cargo.toml
@@ -15,13 +15,20 @@ borsh = ["dep:borsh"]
 borsh = { version = "1.5.1", optional = true }
 bytemuck = { version = "1.19.0" }
 bytemuck_derive = { version = "1.8.0" }
+num-derive = "0.4"
+num-traits = "0.2"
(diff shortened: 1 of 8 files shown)


libraries/pod/Cargo.toml
```diff
@@ -15,13 +15,20 @@ borsh = ["dep:borsh"]
...
+solana-program-error = "2.1.0"
+solana-program-option = "2.1.0"
+solana-pubkey = "2.1.0"
 solana-zk-sdk = "2.1.0"
-spl-program-error = { version = "0.5.0", path = "../program-error" }
+thiserror = "1.0"
 
 [dev-dependencies]
 serde_json = "1.0.132"
+solana-program = "2.1.0"
```
joncinque, line 31: It looks like we're only using a particular pubkey in a test, so this should be easy to remove, but it can be done in follow-up work

PR #7429, opened 2024-10-31
remove solana-program from spl-memo

kevinheavey:
Breaking change: replaces the solana_program re-export with re-exports of the smaller crates


joncinque (approved):
Looks great, thanks!

PR #7430, opened 2024-10-31
remove solana-program from token-group-interface

kevinheavey:
Updated imports and replace spl_program_error usage with inline impls

### DIFF
--- a/token-group/interface/Cargo.toml
+++ b/token-group/interface/Cargo.toml
@@ -9,12 +9,21 @@ edition = "2021"
 
 [dependencies]
 bytemuck = "1.19.0"
+num-derive = "0.4"
+num-traits = "0.2"
+solana-decode-error = "2.1.0"
+solana-instruction = "2.1.0"
+solana-msg = "2.1.0"
 solana-program = "2.1.0"
+solana-program-error = "2.1.0"
+solana-pubkey = "2.1.0"
 spl-discriminator = { version = "0.3.0" , path = "../../libraries/discriminator" }
 spl-pod = { version = "0.4.0", path = "../../libraries/pod", features = ["borsh"] }
 spl-program-error = { version = "0.5.0" , path = "../../libraries/program-error" }
+thiserror = "1.0"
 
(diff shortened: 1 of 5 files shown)


token-group/interface/src/state.rs
joncinque, line 1: Is there a reason that you didn't update the usage of `solana_program` at the top of this file?

token-group/interface/Cargo.toml
```diff
@@ -9,12 +9,20 @@ edition = "2021"
 
 [dependencies]
 bytemuck = "1.19.0"
+num-derive = "0.4"
+num-traits = "0.2"
+solana-decode-error = "2.1.0"
+solana-instruction = "2.1.0"
+solana-msg = "2.1.0"
 solana-program = "2.1.0"
```
joncinque, line 17: It looks the only usage of solana-program can be removed from state.rs, see my other comment

joncinque (commented):
Looks good, but let's finish the job!

joncinque (approved):

joncinque:
CI error is being fixed in #7432

token-group/interface/src/error.rs
```diff
@@ -1,13 +1,18 @@
...
+use {
+    solana_decode_error::DecodeError,
+    solana_msg::msg,
+    solana_program_error::{PrintProgramError, ProgramError},
+};
 
 /// Errors that may be returned by the interface.
-#[spl_program_error(hash_error_code_start = 3_406_457_176)]
+#[repr(u32)]
+#[derive(Clone, Debug, Eq, thiserror::Error, num_derive::FromPrimitive, PartialEq)]
```
buffalojoec, line 11: Why not just update the macro?

kevinheavey: I tried, but that macro is quite tied to the assumption that there is a single solana_program crate that we import, and I didn't find a way to update it without breaking everything

joncinque: This is my fault, remember #7112?

PR #7431, opened 2024-10-31
remove solana-program from token-metadata-interface

kevinheavey:
Updated some imports and replaced spl-program-error with inlining

### DIFF
--- a/token-metadata/interface/Cargo.toml
+++ b/token-metadata/interface/Cargo.toml
@@ -12,17 +12,26 @@ serde-traits = ["dep:serde", "spl-pod/serde-traits"]
 
 [dependencies]
 borsh = "1.5.1"
+num-derive = "0.4"
+num-traits = "0.2"
 serde = { version = "1.0.214", optional = true }
+solana-borsh = "2.1.0"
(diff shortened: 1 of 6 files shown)


token-metadata/interface/Cargo.toml
```diff
@@ -12,17 +12,26 @@ serde-traits = ["dep:serde", "spl-pod/serde-traits"]
...
 [dependencies]
 borsh = "1.5.1"
+num-derive = "0.4"
+num-traits = "0.2"
 serde = { version = "1.0.214", optional = true }
+solana-borsh = "2.1.0"
+solana-decode-error = "2.1.0"
+solana-instruction = "2.1.0"
+solana-msg = "2.1.0"
 solana-program = "2.1.0"
```
joncinque, line 22: Can we go with the same approach here and just re-export the crates that it uses? It looks like that's the only usage of solana-program in the crate, so we can remove it after that

joncinque (commented):
This also contains the token-group changes, so we can land the token-group first, rebase, then land this. Looks great other than the last remaining usage of solana-program!

joncinque (approved):

PR #7436, opened 2024-11-01
instruction-padding: Remove solana-program dependency

joncinque:
#### Problem

The instruction-padding program still has an explicit dependency on solana-program, but it isn't needed.

#### Summary of changes

Use the component crates instead. Some of the syscall consts aren't available yet outside of solana-program, but it was easy to add a static assertion for those.

The program build is _lightning fast_ with this!

### DIFF
--- a/instruction-padding/program/src/instruction.rs
+++ b/instruction-padding/program/src/instruction.rs
@@ -2,14 +2,25 @@
 
 use {
     num_enum::{IntoPrimitive, TryFromPrimitive},
-    solana_program::{
-        instruction::{AccountMeta, Instruction},
-        program_error::ProgramError,
-        pubkey::Pubkey,
-        syscalls::{MAX_CPI_ACCOUNT_INFOS, MAX_CPI_INSTRUCTION_DATA_LEN},
-    },
+    solana_instruction::{AccountMeta, Instruction},
+    solana_program_error::ProgramError,
+    solana_pubkey::Pubkey,
     std::{convert::TryInto, mem::size_of},
 };
+
+const MAX_CPI_ACCOUNT_INFOS: usize = 128;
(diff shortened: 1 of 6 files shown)


instruction-padding/program/src/lib.rs
```diff
@@ -2,5 +2,4 @@ mod entrypoint;
 pub mod instruction;
 pub mod processor;
 
-pub use solana_program;
```
buffalojoec, line 5: Breaking change, right?

instruction-padding/program/src/instruction.rs
```diff
@@ -2,15 +2,26 @@
...
+#[cfg(test)]
+static_assertions::const_assert_eq!(
+    MAX_CPI_ACCOUNT_INFOS,
+    solana_program::syscalls::MAX_CPI_ACCOUNT_INFOS
+);
+#[cfg(test)]
+static_assertions::const_assert_eq!(
+    MAX_CPI_INSTRUCTION_DATA_LEN,
+    solana_program::syscalls::MAX_CPI_INSTRUCTION_DATA_LEN
+);
```
buffalojoec, line 23: Nice!

buffalojoec (dismissed):
⚡️ ⚡️

instruction-padding/program/src/lib.rs
joncinque: Ah oops! I thought I pushed up the latest change, which re-exports the new solana crates. But yes, this is otherwise breaking

buffalojoec: I guess if it's a breaking change anyway, do we want to still have the re-exports? Or should we just do away with them?

joncinque: It can't hurt in case someone needs to access it

PR #7437, opened 2024-11-01
CI: Purge runner during clippy step for publish

joncinque:
#### Problem

The clippy step is running out space during the publish workflow.

#### Summary of changes

Same as #7432, but for the publish workflow.


joncinque:
CI's unnecessary, merging early

PR #7442, opened 2024-11-01
transfer-hook-interface: Remove solana-program dependency

joncinque:
#### Problem

The transfer-hook-interface pulls in solana-program, but it doesn't need to.

#### Summary of changes

Use the component crates. The only breaking change is the re-exports. It still uses a dev-dependency on solana-program since the system program id isn't available yet outside of solana-program.


joncinque:
Oof, never mind, the CI failure here is very legit.

The token-2022 program tests rely on using the "native builtin" hack, where solana-program-test turns a provided function pointer into a program processor. This relies on the syscall stubs, which are not used in the new `solana_cpi` crate.

To properly test all the transfer hook behavior, we'll need to turn all of the test processors into built BPF programs.

buffalojoec:
> To properly test all the transfer hook behavior, we'll need to turn all of the test processors into built BPF programs.

What's the plan for this, create a "test programs" directory? Could get a little messy.

joncinque:
Yeah that's what I've been doing. A directory of test programs, with built programs in tests/fixtures. I was thinking that we'll need to do this no matter what, if we ever want to move away from solana-program-test, so it's probably worth biting the bullet. It comes out to 4 extra test programs. Other than the extra directories and workspace entries, do you have any other concerns?

buffalojoec:
> Other than the extra directories and workspace entries, do you have any other concerns?

No, I don't think so, it was mainly the workspace members and determining how often we need to build new fixtures, which shouldn't be often.

token/program-2022-test/transfer-hook-test-programs/downgrade/src/lib.rs
```diff
@@ -0,0 +1,42 @@
+//! Program implementation
+
+use ::{
```
buffalojoec, line 3: nit: did you mean to include this extra `::` ?

buffalojoec (dismissed):
Nice work! This came out cleaner than I expected!

token/program-2022-test/transfer-hook-test-programs/downgrade/src/lib.rs
joncinque: oops, sure didn't thanks!

buffalojoec (approved):

PR #7443, opened 2024-11-01
associated-token-account-client: Remove solana-program

joncinque:
#### Problem

The ATA client crate still uses solana-program, but it doesn't need it.

#### Summary of changes

Remove its usage, but it's still needed in dev-dependencies until the system program client has been extracted.

### DIFF
--- a/associated-token-account/client/Cargo.toml
+++ b/associated-token-account/client/Cargo.toml
@@ -8,6 +8,10 @@ license = "Apache-2.0"
 edition = "2021"
 
 [dependencies]
+solana-instruction = { version = "2.1.0", features = ["std"] }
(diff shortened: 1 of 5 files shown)


associated-token-account/client/Cargo.toml
```diff
@@ -8,6 +8,10 @@ license = "Apache-2.0"
 edition = "2021"
 
 [dependencies]
+solana-instruction = { version = "2.1.0", features = ["std"] }
+solana-pubkey = { version = "2.1.0", features = ["curve25519"] }
```
febo, line 12: ❤️

febo (approved):
Great!

PR #7445, opened 2024-11-01
record: Remove dependency on solana-program

joncinque:
#### Problem

It's possible to write programs without solana-program, but spl-record still has a dependency on it.

#### Summary of changes

Mostly straightforward, replacing the crates with their components. I also removed a dependency on spl-pod, because that brings in solana-zk-sdk, which does require solana-program currently.

Compile times on my local machine go down from ~16s to less than ~4s :partying_face:


samkim-crypto (approved):
Looks good to me! `solana-program` is now removed from `zk-sdk` as well (https://github.com/anza-xyz/agave/pull/3409), but that is not included in 2.1.

PR #7449, opened 2024-11-03
[token-cli] Parse compute unit limit with the proper name

samkim-crypto:
#### Problem
It seems that we define `compute_unit_limit` argument in the clap app, but try to parse it as `compute_unit_price`.

#### Summary of Changes
Parse `compute_unit_limit` properly.


joncinque (approved):
Oops, nice find!

PR #7456, opened 2024-11-04
Adds --blockhash arg to Close

brooksprumo:
I'd like the `--blockhash` arg on the Close instruction. This is done by calling `.offline_args()`.

<details><summary>output from `spl-token close --help`</summary>
<p>

```
spl-token-close 
Close a token account

USAGE:
    spl-token close [OPTIONS] [--] [TOKEN_MINT_ADDRESS]

ARGS:
    <TOKEN_MINT_ADDRESS>    Token of the associated account to close. To close a specific
                            account, use the `--address` parameter instead

OPTIONS:
        --address <TOKEN_ACCOUNT_ADDRESS>
            Specify the token account to close [default: owner's associated token account]

        --blockhash <BLOCKHASH>
            Use the supplied blockhash

    -C, --config <PATH>
            Configuration file to use

        --close-authority <KEYPAIR>
            Specify the token's close authority if it has one, otherwise specify the token's owner
            keypair. This may be a keypair file or the ASK keyword. Defaults to the client keypair.

        --dump-transaction-message
            Display the base64 encoded binary transaction message in sign-only mode

        --fee-payer <KEYPAIR>
            Specify the fee-payer account. This may be a keypair file, the ASK keyword
            or the pubkey of an offline signer, provided an appropriate --signer argument
            is also passed. Defaults to the client keypair.

    -h, --help
            Print help information

        --multisig-signer [<MULTISIG_SIGNER>...]
            Member signer of a multisig account

        --nonce <PUBKEY>
            Provide the nonce account to use when creating a nonced
            transaction. Nonced transactions are useful when a transaction
            requires a lengthy signing process. Learn more about nonced
            transactions at https://docs.solanalabs.com/cli/examples/durable-nonce

        --nonce-authority <KEYPAIR>
            Provide the nonce authority keypair to use when signing a nonced transaction

        --output <FORMAT>
            Return information in specified output format [possible values: json, json-compact]

        --owner <OWNER_ADDRESS>
            Address of the primary authority controlling a mint or account. Defaults to the client
            keypair address.

    -p, --program-id <ADDRESS>
            SPL Token program id

        --program-2022
            Use token extension program token 2022 with program id:
            TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb

        --recipient <REFUND_ACCOUNT_ADDRESS>
            The address of the account to receive remaining SOL [default: --owner]

        --sign-only
            Sign the transaction offline

        --signer <PUBKEY=SIGNATURE>
            Provide a public-key/signature pair for the transaction

    -u, --url <URL_OR_MONIKER>
            URL for Solana's JSON RPC or moniker (or their first letter): [mainnet-beta, testnet,
            devnet, localhost] Default from the configuration file.

    -v, --verbose
            Show additional information

        --with-compute-unit-limit <COMPUTE-UNIT-LIMIT>
            Set compute unit limit for transaction, in compute units.

        --with-compute-unit-price <COMPUTE-UNIT-PRICE>
            Set compute unit price for transaction, in increments of 0.000001 lamports per compute
            unit.
```

</p>
</details>

### DIFF
--- a/token/cli/tests/command.rs
+++ b/token/cli/tests/command.rs
@@ -3129,14 +3129,16 @@ async fn do_offline_multisig_transfer(
             .await
             .unwrap();
 
-        let program_client: Arc<dyn ProgramClient<ProgramRpcClientSendTransaction>> = Arc::new(
-            ProgramOfflineClient::new(blockhash, ProgramRpcClientSendTransaction),
-        );
+        let offline_program_client: Arc<dyn ProgramClient<ProgramRpcClientSendTransaction>> =
+            Arc::new(ProgramOfflineClient::new(
+                blockhash,
+                ProgramRpcClientSendTransaction,
+            ));
         let mut args = vec![
             "spl-token".to_string(),
             CommandName::Transfer.as_ref().to_string(),
             token.to_string(),
-            "10".to_string(),
+            "100".to_string(),
             destination.to_string(),
             "--blockhash".to_string(),
             blockhash.to_string(),
@@ -3166,7 +3168,7 @@ async fn do_offline_multisig_transfer(
             args.push("--with-compute-unit-limit".to_string());
             args.push(10_000.to_string());
         }
-        config.program_client = program_client;
+        config.program_client = offline_program_client;
         let result = exec_test_cmd(&config, &args).await.unwrap();
         // the provided signer has a signature, denoted by the pubkey followed
         // by "=" and the signature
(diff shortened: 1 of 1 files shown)


joncinque (commented):
The change looks good to me! The `close` implementation already takes into account the `sign_only` field, so this should be the only change needed.

Once you've had a chance to try this locally just to double-check, we can merge it. Thanks for your contribution!

joncinque:
> However, it returned an error, even though the transaction was sent, and the token account was closed... I'm not sure if this is related or not; do you know?

Hm, that typically means that there was an issue confirming the transaction. I wonder if there's a bug in `confirm_transaction_with_spinner`, specifically this bit that might fail, especially with the nonce blockhash: https://github.com/anza-xyz/agave/blob/b46c87ea538a1508ed4ae36f8dbf6c75dd57c883/rpc-client/src/nonblocking/rpc_client.rs#L1090

We might need to create the RPC client with timeouts in token-cli, similar to what we're doing in the main CLI.  I put in #7484 for that -- let me know what you think!

joncinque:
> I didn't use a nonce in my testing, just a recent blockhash I got from getLatestBlockhash. Do you think that code you highlighted would still fail in this circumstance?

Hm, it would only fail if the transaction got in right before the blockhash expired. Anyway, I've tested this locally and it works without my PR, so let's merge this in

joncinque:
Ok cool, let me know how this looks!

PR #7465, opened 2024-11-06
CI: Fixup patch crates script

joncinque:
#### Problem

The locations of crates in the monorepo was moved to the `sdk/` dir in https://github.com/anza-xyz/agave/pull/3498, but the patch script is still looking in the wrong location. This is why the downstream job was disabled in that PR.

#### Summary of changes

Get the patch script up to date with the new locations of the crates.


Lichtso:
This broke upstream monorepo, see: https://github.com/anza-xyz/agave/actions/runs/11727101736/job/32682222360?pr=3520

joncinque:
Ah darn, sorry about that, we can't support both versions at once unless we have different patch scripts for master and v2.1. I'll disable the downstream jobs in v2.1 for now

PR #7480, opened 2024-11-11
[program-2022] Add auditor ciphertexts to instruction data

samkim-crypto:
#### Problem
Currently, confidential transfer, mint and burn instructions do not include the transfer amount ciphertext as part of the token instruction data. This ciphertext is included as part of the proof that is either included in the Zk ElGamal proof program instruction, record account, or context state account.

From the perspective of a real world auditor who must constantly be monitoring activities in a particular mint to decrypt transactions, it must access the ciphertexts indirectly by first looking up the token instruction data and then if proof accounts are used, look up the corresponding record account or context state account, parse these accounts, and then get the transfer amount ciphetext. It would make life much simpler for the auditor if the token instruction data simply includes the auditor part of the ciphertext.

#### Summary of Changes
Added the auditor ciphertext to the confidential transfer, mint, and burn instruction data. I think the program changes should be straightforward. The only tricky part is coming up with the most logical interface on the client side. I updated the transfer, mint, and burn token-client functions to take in the auditor ciphertext as input, but I am open to suggestions on making the client functions a bit simpler.


token/client/src/token.rs
```diff
@@ -2201,6 +2201,8 @@ where
         ciphertext_validity_proof_account: Option<&ProofAccount>,
         range_proof_account: Option<&ProofAccount>,
         transfer_amount: u64,
+        transfer_amount_auditor_ciphertext_lo: Option<&PodElGamalCiphertext>,
+        transfer_amount_auditor_ciphertext_hi: Option<&PodElGamalCiphertext>,
```
joncinque, line 2205: I was having a hard time following when these would be used, so let me know if I've understood. From what I'm seeing, they're only used if all of the equality / ciphertext validity / range proofs are `ProofAccount`s, which makes sense since the function doesn't have access to the ciphertexts otherwise.

In that case, how about creating a new type for the ciphertext validity proof, which is like `ProofAccount`, but also holds onto the auditor ciphertexts? That seems like the clearest option to me, because otherwise this logic is pretty confusing.

token/client/src/token.rs
```diff
@@ -2261,6 +2263,33 @@ where
...
+                let transfer_amount_auditor_ciphertext_lo = proof_data
+                    .context_data()
+                    .grouped_ciphertext_lo
+                    .try_extract_ciphertext(2)
+                    .map_err(|_| TokenError::ProofGeneration)?;
+                let transfer_amount_auditor_ciphertext_hi = proof_data
+                    .context_data()
+                    .grouped_ciphertext_hi
+                    .try_extract_ciphertext(2)
+                    .map_err(|_| TokenError::ProofGeneration)?;
```
joncinque, line 2277: If I've understood this correctly, the proof data is created in this function, right before, so would it be possible for this error to happen ever? We might prefer an `unwrap()` in that case. Or maybe even simpler, we can return them from the earlier block that generates the proof data

token/client/src/token.rs
```diff
@@ -2261,6 +2263,33 @@ where
...
+                )
+            } else {
+                // the validity proof data is always generated unless
+                // `transfer_amount_auditor_ciphertext_lo` and
+                // `transfer_amount_auditor_ciphertext_hi` are `Some`, so it is
+                // safe to unwrap
+                (
+                    *transfer_amount_auditor_ciphertext_lo.unwrap(),
+                    *transfer_amount_auditor_ciphertext_hi.unwrap(),
+                )
```
joncinque, line 2290: I don't think this is guaranteed by the function (unless we add a new type as I suggested), so we should return an error if either is None

token/program-2022/src/extension/confidential_mint_burn/processor.rs
```diff
@@ -210,20 +210,36 @@ fn process_confidential_mint(
         }
     }
 
+    let proof_context_auditor_ciphertext_lo = proof_context
+        .mint_amount_ciphertext_lo
+        .try_extract_ciphertext(2)
+        .map_err(|e| -> TokenError { e.into() })?;
```
joncinque, line 216: nit-but-it's-up-to-you: I typically find it easier to read `map_err(TokenError::from)`

joncinque (commented):
the guts look great! Just a nit on the API for token-client and a nit on converting into `TokenError`

joncinque (dismissed):
Looks really good! Just one last nit, up to you

joncinque (approved):

PR #7509, opened 2024-11-20
stake-pool: Downgrade dependencies for fixed release

joncinque:
#### Problem

The current spl-stake-pool crate doesn't always work due to its dependency on spl-math which still uses the v1 Solana crates.

#### Summary of changes

Since the spl-math dependency wasn't actually needed, it was removed in #7421. So just to get a new patch release out, downgrade the local dependencies.

This will fail CI, but after it lands we can tag the release, and revert this PR.

cc @billythedummy

### DIFF
--- a/stake-pool/program/Cargo.toml
+++ b/stake-pool/program/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "spl-stake-pool"
-version = "2.0.0"
+version = "2.0.1"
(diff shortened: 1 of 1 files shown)


Ownerbusinessjjfm:
Is my coin was being bridge or tap under my wallet ID...as I see the transaction was being made.....transfer to other wallet...is my account all if it was compromised...

PR #7511, opened 2024-11-20
token-2022: Add scaled amount extension

joncinque:
#### Problem

The interest-bearing extension is useful for tokens that accrue in value constantly, but many "rebasing" tokens on other blockchains employ a different method of updating the number of tokens in accounts.

Rather than setting a rate and allowing the number to change automatically over time, they set a scaling factor for the tokens by hand.

#### Summary of changes

Add a new `ScaledUiAmount` extension to token-2022 for doing just that. This is essentially a simplified version of the interest-bearing extension, where someone just sets a scaling value into the mint directly. The scale has no impact on the operation of the token, just on the output of `amount_to_ui_amount` and `ui_amount_to_amount`.

cc @jnwng

### DIFF
--- a/token/program-2022-test/tests/scaled_ui_amount.rs
+++ b/token/program-2022-test/tests/scaled_ui_amount.rs
@@ -33,7 +33,7 @@ use {
 
 #[tokio::test]
 async fn success_initialize() {
-    for (scale, authority) in [
+    for (multiplier, authority) in [
         (f64::MIN_POSITIVE, None),
         (f64::MAX, Some(Pubkey::new_unique())),
     ] {
@@ -41,7 +41,7 @@ async fn success_initialize() {
         context
             .init_token_with_mint(vec![ExtensionInitializationParams::ScaledUiAmountConfig {
                 authority,
-                scale,
+                multiplier,
             }])
             .await
             .unwrap();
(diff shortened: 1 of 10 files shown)


joncinque:
I got feedback that issuers want to be able to set the multiplier in the future, so I'll implement that and then re-open this PR

joncinque:
Ok, this should be ready for another look!

token/program-2022/src/extension/scaled_ui_amount/instruction.rs
```diff
@@ -0,0 +1,139 @@
...
+#[cfg_attr(feature = "serde-traits", serde(rename_all = "camelCase"))]
+#[derive(Clone, Copy, Debug, PartialEq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum ScaledUiAmountMintInstruction {
+    /// Initialize a new mint with scaled UI amounts.
+    ///
+    /// Fails if the mint has already been initialized, so must be called before
+    /// `InitializeMint`.
+    ///
+    /// Fails with any number less than 0.
```
samkim-crypto, line 31: Correct me if I am wrong, but 0 multiplier is also not allowed, right? Should we also make "number" be a little more specific like "multiplier"?

token/program-2022/src/extension/scaled_ui_amount/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+        let multiplier = if unix_timestamp >= self.new_multiplier_effective_timestamp.into() {
+            self.new_multiplier
+        } else {
+            self.multiplier
+        };
+        f64::from(multiplier) / 10_f64.powi(decimals as i32)
+    }
+
+    /// Convert a raw amount to its UI representation using the given decimals
+    /// field Excess zeroes or unneeded decimal point are trimmed.
```
samkim-crypto, line 67:
```suggestion
    /// Convert a raw amount to its UI representation using the given decimals
    /// field. Excess zeroes or unneeded decimal point are trimmed.
```

token/program-2022/src/extension/scaled_ui_amount/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+
+/// `UnixTimestamp` expressed with an alignment-independent type
+pub type UnixTimestamp = PodI64;
+
+/// `f64` type that can be used in `Pod`s
+#[cfg_attr(feature = "serde-traits", derive(Serialize, Deserialize))]
+#[cfg_attr(feature = "serde-traits", serde(from = "f64", into = "f64"))]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Pod, Zeroable)]
+#[repr(transparent)]
+pub struct PodF64(pub [u8; 8]);
```
samkim-crypto, line 24: Was there a reason we define this type here rather than in `spl_pod::primitives` like the other primitive pod types?

token/program-2022-test/tests/scaled_ui_amount.rs
```diff
@@ -0,0 +1,343 @@
...
+        extension::{scaled_ui_amount::ScaledUiAmountConfig, BaseStateWithExtensions},
+        instruction::{amount_to_ui_amount, ui_amount_to_amount, AuthorityType},
+        processor::Processor,
+    },
+    spl_token_client::token::{ExtensionInitializationParams, TokenError as TokenClientError},
+    std::{convert::TryInto, sync::Arc},
+};
+
+#[tokio::test]
+async fn success_initialize() {
```
samkim-crypto, line 35: Just for completeness, should we add a test to invoke the newly added error variant `TokenError::InvalidScale` by trying to initialize with an invalid scale like 0?

samkim-crypto (commented):
Looks really good to me! Just minor questions and comments regarding the docs and tests.

token/program-2022/src/extension/scaled_ui_amount/mod.rs
joncinque: Yeah, two reasons:

1. We should discourage floats in on-chain programs, so if we put it in a general library people might feel compelled to use it
2. Whenever there's a change to spl-pod, it's very annoying to publish all the new crates

token/program-2022-test/tests/scaled_ui_amount.rs
joncinque: ah yes, good idea, I forgot to add that, thanks!

token/program-2022/src/extension/scaled_ui_amount/instruction.rs
joncinque: Correct, I initially did allow 0, but then I learned about subnormal floats https://en.wikipedia.org/wiki/Subnormal_number -- I'll fix up this documentation

joncinque:

token/program-2022/src/extension/scaled_ui_amount/mod.rs
samkim-crypto: Okay, makes sense to me!

samkim-crypto (approved):
Great! Everything looks find to me!

PR #7523, opened 2024-11-25
token-client: Check simulation results for error

joncinque:
#### Problem

If the simulation for compute unit limit fails in the token-cli, it may continue to try to broadcast a transaction since the error falls through. Although that transaction should fail too, at least we don't run the risk of broadcasting something known to be bad.

#### Summary of changes

Add an `err()` function on the simulation results, and abort if the simulation for compute units fails.


Ownerbusinessjjfm:
What shall I do?

PR #7540, opened 2024-11-27
token-2022: Take decimals into account for UI amount

joncinque:
#### Problem

The interest-bearing and scaled UI amount extensions don't take into account the mint decimals when printing the number, and they don't properly trim afterwards. This can be confusing.

#### Summary of changes

Update the UI amount conversion to take into account the decimals of precision, and then actually trim. Note that a lot of calculations become more precise because of this change!


gitteri (commented):
Nice!

samkim-crypto (approved):

PR #7541, opened 2024-11-27
add JS helpers for interest bearing tokens' UIAmounts

gitteri:
This adds four new helper functions to allow developers to better interact with interest bearing tokens. 
1. amountToUiAmountForMintWithoutSimulation - converts a number/bigint amount for a given mint at the current time to a UiAmount
2. amountToUiAmountWithoutSimulation - converts a given number/bigint amount and detailed information about the mint's interest rates to a UiAmount
3. uiAmountToAmountForMintWithoutSimulation- converts a given ui amount string for a given mint at the current time to amount
4. uiAmountToAmountWithoutSimulation - converts a given ui amount string and detailed information about the mint's interest rates to amount

UiAmounts are provided with the token's decimals as its precision and truncated instead of rounded to avoid rounding issues.
Amounts are processed in the base unit of the token (if the token has 2 decimals 1.23 is represented as 123)


token/js/src/extensions/interestBearingMint/state.ts
```diff
@@ -6,9 +6,9 @@ import { ExtensionType, getExtensionData } from '../extensionType.js';
 
 export interface InterestBearingMintConfigState {
     rateAuthority: PublicKey;
-    initializationTimestamp: bigint;
+    initializationTimestamp: number;
     preUpdateAverageRate: number;
-    lastUpdateTimestamp: bigint;
+    lastUpdateTimestamp: number;
```
gitteri, line 11: these were changed because the buffer encoder does not work with bigints. numbers should be fine for timestamps and this doesn't seem to impact any other current usage

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+ * @param amount         Amount of tokens to be converted to Ui Amount
+ * @param programId      SPL Token program account (default: TOKEN_PROGRAM_ID)
+ *
+ * @return Ui Amount generated
+ */
+export async function amountToUiAmountForMintWithoutSimulation(
+    connection: Connection,
+    mint: PublicKey,
+    amount: string,
+    programId = TOKEN_PROGRAM_ID,
```
joncinque, line 115: Up to you, but since this takes the mint's public key, we can use the account owner instead of forcing people to pass this

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+): Promise<string> {
+    Decimal.set({ toExpPos: 24, toExpNeg: -24 })
+    const mintInfo = await getMint(connection, mint, 'confirmed', programId);
+    const amountDecimal = new Decimal(amount.toString());
+    const decimalsFactor = new Decimal(10).pow(mintInfo.decimals);
+
+    if (programId.equals(TOKEN_PROGRAM_ID)) {
+        console.log('amountDecimal', amountDecimal.toString(), 'mintInfo', mintInfo);
+        return amountDecimal.div(decimalsFactor).toString();
+    }
```
joncinque, line 125: Is this needed? The next branch will cover this case since `Tokenkeg...` mints don't have any extensions

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+export async function amountToUiAmountForMintWithoutSimulation(
+    connection: Connection,
+    mint: PublicKey,
+    amount: string,
+    programId = TOKEN_PROGRAM_ID,
+): Promise<string> {
+    Decimal.set({ toExpPos: 24, toExpNeg: -24 })
+    const mintInfo = await getMint(connection, mint, 'confirmed', programId);
+    const amountDecimal = new Decimal(amount.toString());
+    const decimalsFactor = new Decimal(10).pow(mintInfo.decimals);
```
joncinque, line 120: nit: Can we move these lines into the `if (!interestBearingMintConfigState)` branch?

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+        console.log('amountDecimal', amountDecimal.toString(), 'mintInfo', mintInfo);
+        return amountDecimal.div(decimalsFactor).toString();
+    }
+
+    const interestBearingMintConfigState = getInterestBearingMintConfigState(mintInfo);
+    if (!interestBearingMintConfigState) {
+        return amountDecimal.div(decimalsFactor).toString();
+    }
+
+    const currentTime = Math.floor(Date.now() / 1000); // Convert to seconds
```
joncinque, line 132: Up to you, but it would be more correct to use the unix timestamp reported by the sysvar clock: https://docs.anza.xyz/runtime/sysvars/#clock, but that would also require fetching another account

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+    // Calculate scaled amount with interest rounded down to the nearest unit
+    const decimalsFactor = new Decimal(10).pow(decimals);
+    const scaledAmountWithInterest = new Decimal(amount).times(totalScale).div(decimalsFactor).toDecimalPlaces(decimals, Decimal.ROUND_DOWN);
+
+    return scaledAmountWithInterest.toString();
+}
+
+/**
+ * Convert amount to UiAmount for a mint with interest bearing extension without simulating a transaction
+ * This implements the same logic as the CPI instruction available in /token/program-2022/src/extension/interest_bearing_mint/mod.rs
```
joncinque, line 102: nit: this is actually reimplementing the logic from `process_amount_to_ui_amount`

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+
+    // Calculate scaled amount with interest rounded down to the nearest unit
+    const decimalsFactor = new Decimal(10).pow(decimals);
+    const scaledAmountWithInterest = new Decimal(amount).times(totalScale).div(decimalsFactor).toDecimalPlaces(decimals, Decimal.ROUND_DOWN);
+
+    return scaledAmountWithInterest.toString();
+}
+
+/**
+ * Convert amount to UiAmount for a mint with interest bearing extension without simulating a transaction
```
joncinque, line 101: nit: this works for any mint!
```suggestion
 * Convert amount to UiAmount for a mint without simulating a transaction
```

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+    // Calculate total scale
+    const totalScale = preUpdateExp.times(postUpdateExp);
+
+    // Calculate original principle by dividing the UI amount (principle + interest) by the total scale
+    const originalPrinciple = uiAmountScaled.div(totalScale);
+    return BigInt(originalPrinciple.trunc().toString()); 
+}
+
+/**
+ * Convert a UI amount with interest back to the original UI amount without interest
```
joncinque, line 208: nit: since this works more generally, how about this?
```suggestion
 * Convert a UI amount back to the raw amount
```

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+ * @param decimals                 Number of decimals of the mint
+ * @param currentTimestamp         Current timestamp in seconds
+ * @param lastUpdateTimestamp      Last time the interest rate was updated in seconds
+ * @param initializationTimestamp  Time the interest bearing extension was initialized in seconds
+ * @param preUpdateAverageRate     Interest rate in basis points (1 basis point = 0.01%) before last update
+ * @param currentRate              Current interest rate in basis points
+ * 
+ * @return Amount scaled by accrued interest as a string with appropriate decimal places
+ */
+
```
joncinque, line 64: nit: can you remove the empty line?

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+ * @param decimals                  Number of decimals for the mint
+ * @param currentTimestamp          Current timestamp in seconds
+ * @param lastUpdateTimestamp       Last time the interest rate was updated in seconds
+ * @param initializationTimestamp   Time the interest bearing extension was initialized in seconds
+ * @param preUpdateAverageRate      Interest rate in basis points (hundredths of a percent) before the last update
+ * @param currentRate              Current interest rate in basis points
+ * 
+ * @return Original amount (principle) without interest
+ */
+
```
joncinque, line 172: nit: can you remove the empty line?

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+    const preUpdateExp = preUpdateExponent.exp();
+
+    // Calculate post-update exponent
+    const postUpdateTimespan = new Decimal(currentTimestamp).minus(lastUpdateTimestamp);
+    const postUpdateNumerator = new Decimal(currentRate).times(postUpdateTimespan);
+    const postUpdateExponent = postUpdateNumerator.div(new Decimal(SECONDS_PER_YEAR).times(ONE_IN_BASIS_POINTS));
+    const postUpdateExp = postUpdateExponent.exp();
+
+    // Calculate total scale
+    const totalScale = preUpdateExp.times(postUpdateExp);
```
joncinque, line 200: nit: up to you, but all of this can get factored into a helper function to be reused for both conversions

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+    const originalPrinciple = uiAmountScaled.div(totalScale);
+    return BigInt(originalPrinciple.trunc().toString()); 
+}
+
+/**
+ * Convert a UI amount with interest back to the original UI amount without interest
+ * 
+ * @param connection     Connection to use
+ * @param mint           Mint to get decimals from
+ * @param uiAmount       UI Amount (principle plus continuously compounding interest) to be converted back to original principle
```
joncinque, line 212: nit: same here, since it's more general
```suggestion
 * @param uiAmount       UI Amount to be converted back to raw amount
```

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+    uiAmount: string,
+    programId = TOKEN_PROGRAM_ID,
+): Promise<bigint> {
+    Decimal.set({ toExpPos: 24, toExpNeg: -24 })
+    const mintInfo = await getMint(connection, mint, 'confirmed', programId);
+    const uiAmountScaled = new Decimal(uiAmount).mul(new Decimal(10).pow(mintInfo.decimals));
+
+    if (programId.equals(TOKEN_PROGRAM_ID)) {
+        return BigInt(uiAmountScaled.trunc().toString());
+    }
```
joncinque, line 230: Same as with the other function, I think we can remove this

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+ */
+export async function uiAmountToAmountForMintWithoutSimulation(
+    connection: Connection,
+    mint: PublicKey,
+    uiAmount: string,
+    programId = TOKEN_PROGRAM_ID,
+): Promise<bigint> {
+    Decimal.set({ toExpPos: 24, toExpNeg: -24 })
+    const mintInfo = await getMint(connection, mint, 'confirmed', programId);
+    const uiAmountScaled = new Decimal(uiAmount).mul(new Decimal(10).pow(mintInfo.decimals));
```
joncinque, line 226: nit: same as with the other function, this can get moved under `if (!interestBearingMintConfigState)`

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+    if (programId.equals(TOKEN_PROGRAM_ID)) {
+        return BigInt(uiAmountScaled.trunc().toString());
+    }
+
+    const interestBearingMintConfigState = getInterestBearingMintConfigState(mintInfo);
+    if (!interestBearingMintConfigState) {
+        return BigInt(uiAmountScaled.trunc().toString());
+    }
+
+    const currentTime = Math.floor(Date.now() / 1000); // Convert to seconds
```
joncinque, line 237: up to you: same here, can fetch the clock for the cluster time

token/js/src/extensions/interestBearingMint/state.ts
joncinque: Sorry, what do you mean about the buffer encoder not working? We use bigints in a lot of the extensions and they seem to work. I'd prefer to leave this as is

pnpm-lock.yaml
joncinque, line 1: Can you revert any changes that aren't necessary for the PR? It seems like we're just adding sinon and decimal.js currently (although I hope we don't need the second one)

joncinque (commented):
Really great work! Mostly small things to fixup, but the logic is there and the tests look good

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +31,189 @@ export async function amountToUiAmount(
...
+ * @param uiAmount       UI Amount (principle plus continuously compounding interest) to be converted back to original principle
+ * @param programId      SPL Token program account (default: TOKEN_PROGRAM_ID)
+ * 
+ * @return Original UI Amount (principle) without interest
+ */
+export async function uiAmountToAmountForMintWithoutSimulation(
+    connection: Connection,
+    mint: PublicKey,
+    uiAmount: string,
+    programId = TOKEN_PROGRAM_ID,
```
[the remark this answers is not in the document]
gitteri: yep good call. I'll update this

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -1,7 +1,10 @@
+import Decimal from 'decimal.js';
```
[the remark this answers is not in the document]
joncinque: Rust has the same behavior, so maybe that's actually ok?

```
    println!("{}", 0.1 + 0.2);
```

Gives

```
0.30000000000000004
```

token/js/src/actions/amountToUiAmount.ts
gitteri: sure. it makes mocks a bit more complicated but let's give it a try

token/js/src/extensions/interestBearingMint/state.ts
gitteri: Specifically if you use a bigint it as an issue when trying to encode this state due to this function that's in the buffer encoding library:
```
function divmodInt64(src) {
    const hi32 = Math.floor(src / V2E32); // cannot mix bigint and numbers here
    const lo32 = src - (hi32 * V2E32);
    return { hi32, lo32 };
}
```

gitteri: I believe to fix this we'd have to fix the buffer encoding library. I took a look through all other state type definitions I could find and didn't see any other bigints

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +30,237 @@ export async function amountToUiAmount(
...
+    preUpdateAverageRate: number,
+    currentRate: number,
+): string {
+    // Calculate pre-update exponent
+    // e^(preUpdateAverageRate * (lastUpdateTimestamp - initializationTimestamp) / (SECONDS_PER_YEAR * ONE_IN_BASIS_POINTS))
+    const preUpdateExp = calculateExponentForTimesAndRate(initializationTimestamp, lastUpdateTimestamp, preUpdateAverageRate)
+
+    // Calculate post-update exponent
+    // e^(currentRate * (currentTimestamp - lastUpdateTimestamp) / (SECONDS_PER_YEAR * ONE_IN_BASIS_POINTS))
+    const postUpdateExp = calculateExponentForTimesAndRate(lastUpdateTimestamp, Number(currentTimestamp), currentRate)
```
joncinque: nit: `Number` shouldn't be needed, unless I've missed something
```suggestion
    const postUpdateExp = calculateExponentForTimesAndRate(lastUpdateTimestamp, currentTimestamp, currentRate)
```

joncinque (commented):
Looking really close! Once the encoding bit is solved and these nits are fixed, it should be good to go

token/js/src/actions/amountToUiAmount.ts
```diff
@@ -28,3 +30,242 @@ export async function amountToUiAmount(
...
+
+    // Calculate post-update exponent
+    const postUpdateExp = calculateExponentForTimesAndRate(lastUpdateTimestamp, currentTimestamp, currentRate);
+
+    // Calculate total scale
+    const totalScale = preUpdateExp * postUpdateExp;
+
+    // Calculate original principal by dividing the UI amount (principal + interest) by the total scale
+    const originalprincipal = uiAmountScaled / totalScale;
+    return BigInt(Math.trunc(originalprincipal));
```
joncinque, line 229:
```suggestion
    const originalPrincipal = uiAmountScaled / totalScale;
    return BigInt(Math.trunc(originalPrincipal));
```

joncinque (commented):
One last one, then we can merge!

joncinque (dismissed):
Thanks for all your patience and your work! Once it passes CI, I'll merge it in

joncinque (approved):

PR #7565, opened 2024-12-08
[program-2022, token-client] Add confidential mint and burn support

samkim-crypto:


samkim-crypto:
Subsumed by https://github.com/solana-labs/solana-program-library/pull/7570 and https://github.com/solana-labs/solana-program-library/pull/7575. The token-client updates will follow in https://github.com/solana-labs/solana-program-library/pull/6881.

PR #7575, opened 2024-12-11
[program-2022] Add functions to generate mint/burn proofs from account info

samkim-crypto:
#### Problem
The confidential mint/burn extension and the confidential transfer extension have some inconsistent ways of generating proofs from account data.

#### Summary of Changes
[6b073ce](https://github.com/solana-labs/solana-program-library/pull/7575/commits/6b073ce08bdf6b6bd00e6d0ac4ec041aeca13969): All proof data types (`TransferProofData`, `TransferWithFeeProofData`, `BurnProofData`, ...) just contains the proof components while the `MintProofData` contains the new decryptable supply component. The decryptable supply ciphertext could be generated together with the proofs, but I think it is slightly more natural to generate it separately from account data like in the other confidential transfer instructions. In this commit, I removed the decryptable supply component from `MintProofData`. Sorry I should have noticed this before in the review 🙏 for https://github.com/solana-labs/solana-program-library/pull/7319.

[9c80b2c](https://github.com/solana-labs/solana-program-library/pull/7575/commits/9c80b2cf0c49923eecf41b1bc36f67a79f51d126): Added functions to generate mint and burn proof data from account infos. This would simplify the client for confidential mint and burn.

[19b7525](https://github.com/solana-labs/solana-program-library/pull/7575/commits/19b7525d565da77ca44c64d5efff68f496eeecef): The order of the parameters of type `ElGamalKeypair` and `AeKey` are inconsistent between the account info functions of confidential transfer and confidential mint/burn, so I switched these. I also renamed `decrypt_current_supply` with `decrypted_current_supply` to be consistent with `decrypt_current_balance` of confidential transfer.

I apologize for the breaking changes, but I think this should really do it for the program 🙏 .

### DIFF
--- a/token/program-2022/src/extension/confidential_mint_burn/account_info.rs
+++ b/token/program-2022/src/extension/confidential_mint_burn/account_info.rs
@@ -182,8 +182,8 @@ impl BurnAccountInfo {
         burn_amount: u64,
         source_elgamal_keypair: &ElGamalKeypair,
         aes_key: &AeKey,
-        auditor_elgamal_pubkey: Option<&ElGamalPubkey>,
         supply_elgamal_pubkey: &ElGamalPubkey,
+        auditor_elgamal_pubkey: Option<&ElGamalPubkey>,
     ) -> Result<BurnProofData, TokenError> {
         let current_available_balance_ciphertext = self
             .available_balance
--- a/token/program-2022/src/extension/confidential_mint_burn/account_info.rs
+++ b/token/program-2022/src/extension/confidential_mint_burn/account_info.rs
@@ -1,11 +1,16 @@
 use {
     super::ConfidentialMintBurn,
-    crate::error::TokenError,
+    crate::{
+        error::TokenError,
+        extension::confidential_transfer::{
+            ConfidentialTransferAccount, DecryptableBalance, EncryptedBalance,
+        },
+    },
     bytemuck::{Pod, Zeroable},
     solana_zk_sdk::{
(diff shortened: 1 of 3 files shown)


token/program-2022/src/extension/confidential_mint_burn/account_info.rs
```diff
@@ -102,4 +111,98 @@ impl SupplyAccountInfo {
...
+
+    /// Create a burn proof data that is split into equality, ciphertext
+    /// validity, and range proof.
+    pub fn generate_split_burn_proof_data(
+        &self,
+        burn_amount: u64,
+        source_elgamal_keypair: &ElGamalKeypair,
+        aes_key: &AeKey,
+        auditor_elgamal_pubkey: Option<&ElGamalPubkey>,
+        supply_elgamal_pubkey: &ElGamalPubkey,
```
joncinque, line 186: nit: the auditor pubkey is the last field in `generate_split_mint_proof_data`. The aes key is also later typically, but I think this looks better -- what do you think?
```suggestion
        aes_key: &AeKey,
        supply_elgamal_pubkey: &ElGamalPubkey,
        auditor_elgamal_pubkey: Option<&ElGamalPubkey>,
```

samkim-crypto: Oh nice catch. Yes, it does look better. Thanks!

joncinque (approved):

PR #7583, opened 2024-12-16
Update name-service.md to includes example services and reword for X

metasal1:
Updated mentions of Twitter to includes X

Added examples of current offerings

- SNS
- AllDomains


joncinque (commented):
Thanks for your contribution! We don't advertise for ecosystem projects in documentation, but I'll be happy to merge the other changes

Issue #7590, opened 2024-12-16
`spl-token` CLI ignores the Solana config's `confirmation` field


acheroncrypto:
### Problem

`spl-token` CLI loads the Solana config:

https://github.com/solana-labs/solana-program-library/blob/9d467db8a93034fb2e61857343a1e76e9794f804/token/cli/src/config.rs#L87-L96

but it hardcodes the commitment to be `confirmed` during RPC client creation (instead of respecting the specified commitment):

https://github.com/solana-labs/solana-program-library/blob/9d467db8a93034fb2e61857343a1e76e9794f804/token/cli/src/config.rs#L103-L108

### Solution

Use the `confirmation` field from the Solana config.

PR #7591, opened 2024-12-16
token-cli: Use the `commitment` field from Solana config

acheroncrypto:
### Problem

`spl-token` CLI doesn't respect the `commitment` field of Solana config when creating the RPC client.

See https://github.com/solana-labs/solana-program-library/issues/7590 for more details.

### Summary of changes

Use the `confirmation` field from the Solana config when creating the RPC client.

**Note:** This won't affect any users (unless they've overridden the `commitment` field) because [`confirmed` is already the default commitment of Solana config](https://github.com/anza-xyz/agave/blob/9e59baae79570628163d95a169e40f11b1db5109/cli-config/src/config.rs#L88).

Resolves https://github.com/solana-labs/solana-program-library/issues/7590

### DIFF
--- a/token/cli/src/config.rs
+++ b/token/cli/src/config.rs
@@ -100,10 +100,12 @@ impl<'a> Config<'a> {
                 .unwrap_or(&cli_config.json_rpc_url),
         );
         let websocket_url = solana_cli_config::Config::compute_websocket_url(&json_rpc_url);
+        let commitment_config = CommitmentConfig::from_str(&cli_config.commitment)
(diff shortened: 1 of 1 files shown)


joncinque (approved):
Thanks for noticing and fixing this! Since the token-cli doesn't accept a commitment parameter otherwise, it makes sense to just read whatever is in the config file.

And if there's an invalid commitment in the config file, then the base Solana CLI falls back to confirmed, so this change is totally consistent with the base CLI.

PR #7627, opened 2025-01-05
Fix #7549

### LINKED ISSUE #7549: [token-js] InitializeMint without FreezeAuthority breaks decodeInitializeMintInstruction
Example tx: `2LQT3oZSoAyrzAC59wfMfbiZtF74YPcw5rQuaAmBu1FhuVe1f3CZoms8wKETJYVx8q9Bi8wT8v337iMDg4PJysF2`

The problem is caused by exact span check [here](https://github.com/solana-labs/solana-program-library/blob/d544111b4cc193394b9bb00ac3e1482bb08a5fa0/token/js/src/instructions/initializeMint.ts#L96)

Proposed solution: check if length of the data is equal to 35 (data with empty freeze authority) or 67 (freeze authority exists) bytes

Danver97:
As explained in #7549, `InitializeMint` and `InitializeMint2` without FreezeAuthority break `decodeInitializeMintInstruction` and `decodeInitializeMint2Instruction` respectively.

Since buffer layouts apparently don't support optional fields, this is currently a workaround. Indeed, this solution won't scale well when more than one optional field is present as it would require implementing the same logic for each combination of presence of each optional field.

The actual solution would require supporting optional `Layout`s in `buffer-layouts`, adapt the `Structure` layout to account for it and easily provide a list of spans, then re-implement `COptionPublicKeyLayout` with it and check in `decodeInitializeMintInstruction` and `decodeInitializeMint2Instruction` if the length of the buffer is in one of the values of spans provided by `Structure`.

Overall, I'd recommend not allowing optional fields, but rather have then set to a null value like 0 or -1.

### DIFF
--- a/token/js/src/extensions/transferFee/instructions.ts
+++ b/token/js/src/extensions/transferFee/instructions.ts
@@ -1,4 +1,4 @@
-import { struct, u16, u8 } from '@solana/buffer-layout';
+import { Layout, struct, u16, u8 } from '@solana/buffer-layout';
 import { u64 } from '@solana/buffer-layout-utils';
 import type { AccountMeta, Signer, PublicKey } from '@solana/web3.js';
 import { TransactionInstruction } from '@solana/web3.js';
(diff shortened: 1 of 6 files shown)


Danver97:
@joncinque Can you please review this PR?

